PR #1260, opened 2015-01-09
Supporting two-word country in Google Geocoding verification.

### LINKED ISSUE #1259: As a user, I'd like to be able to use multi-word countries when using Google Geocoding address validation
We are not encoding the country name. As a result, it fails if the user enters something like "United States" instead of "US." We just get a "400: Bad Request" as response.

ceiroa:
This pull request can be merged now, @kbromer.

# Warning

# Info
* Google Geocoding Address Verification now supports State and Country picklists.

# Issues
Fixes #1259.


ceiroa:
Hey @davidhabib, could you take a look at this and merge when you have a moment? It's a tiny change. Apparently Google Geocoding expects spaces to be replaced with plus signs, and we were not doing that for the country name. (I think doing URL encoding and replacing them with %20 would work too, but the documentation is not very clear on that, and the example they show uses plus sign.)

PR #1265, opened 2015-01-14
Feature/1261 multicurrency recdon bug

### LINKED ISSUE #1261: Error: Attempt to de-reference a null object
Hi,

On manually running the Recurring donation batch from NPSP Setting tab we are getting 
error: "Attempt to de-reference a null object" in the npsp error logs.

We have simple multi-currency configuration in our org with a single conversion rate.

We are trying to run this manually as the Paid Amount field on Recurring Donation is not getting calculated correctly.

image

Total No. of Paid installments is 13 and installment amount is 100
so we are expecting Paid Amount to be $1300
Note: Both Donation and Recurring Donation have same currency code (AUD).

Can you please help on this.

Thanks

njjc:
# Warning
# Info
# Issues
Fixes #1261


ceiroa:
@njjc ~ Could you please make sure this pull request is flagged as passing build before I merge it? (You could just manually rerun the feature branch.)

PR #1274, opened 2015-01-19
Renaming the opportunity field sets that were mistakenly pulled from the...

### LINKED ISSUE #1271: As a developer, I need to add back (with different names) the opportunity bde fieldsets, which got removed by mistake from the packaging org
Release 3.30 removed the Opportunity BDE_Entry_FS and BDE_List_FS fieldsets. We need to put them back, although they'll need to be renamed, as the platform does not let us add them back  with the same name. Code references to them will need to be rewritten to work with the old and new names.

The Account AllAccounts view was also removed, but this one does not need to be put back.

ceiroa:
... packaging org. Removing the account listview that was also pulled. Adding hard references to all field sets, so they cannot be removed without us noticing.

# Warning

# Info
* Due to a packaging issue, it was necessary to rename the BDE_Entry_FS and BDE_List_FS fieldsets in Opportunity. The new names are BDE_Entry_FS_v2 and BDE_List_FS_v2. Once you install this version, you will have both sets of fieldsets. If you customized them, you'll probably want to reapply those customizations to the new ones. Once these are applied, the old fieldsets can be deleted.

# Issues
Fixes #1271.


ceiroa:
@kbromer or @davidhabib, could one of you take a look at this asap? We need it to get dev build job working again. Also, anything else to add to the info message?

PR #1279, opened 2015-01-20
Feature/1243 primary contact hh converter

### LINKED ISSUE #1243: As a user of the Household model converter, I'd like the primary contact to be set by contact created date if no giving information is available
Probably just need to order this query: 

https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/CONV_Account_Conversion_BATCH.cls#L78

njjc:
# Warning
# Info
# Issues
Fixes #1243


njjc:
@ceiroa  this pull req is ready to go

ceiroa:
I'll wait until it's flagged as passing build...

ceiroa:
So the only difference is the "order by" clause? How does this fix the issue?

kbromer:
The issue was the random ordering of Contacts that determined the head of Household if no donation info was available.  This would put them in CreatedDate order in that scenario

PR #1282, opened 2015-01-20
Docs/1178 javadocs code comments

davidhabib:


davidhabib:
@jlantz this branch is synced to dev, and I see no merge conflicts, but it won't enable merge pull request.  why not?  how do we resolve this?

davidhabib:
@ceiroa Can you accept this.  i fixed the merge conflict, and it is mostly ApexDoc changes.

davidhabib:
@ceiroa @njjc I manually kicked off a feature build for this, so once it completes successfully, anyone feel free to accept this (after the quick review).  thanks.

src/classes/ADDR_IValidator.cls
```diff
@@ -35,19 +35,43 @@
...
     String getDefaultURL();
     
-    /** Return true if batch address validation is supported. If not, the batch address validation button
-    will be disabled. **/
+    /** 
+    * @description returns the help text to be shown when the user selects the service from the drop-down. It
+    * may specify things such as the required fields. 
+    * @return string
+    */
+    String getHelpText();
```
kbromer: Here, this is new - so it need to be implemented by everything that implements this ADDR_IValidator interface

davidhabib: Sounds like I need to do another pull from dev. My problem was I couldn't
even get the build going on that branch manually. I hadn't heard from Jason
that he fixed anything. Maybe another intermittent issue, or he forgot to
tell me it was fixed.  Thx.



On Jan 25, 2015, at 10:40 AM, Kevin Bromer <notifications@github.com> wrote:

In src/classes/ADDR_IValidator.cls
<https://github.com/SalesforceFoundation/Cumulus/pull/1282#discussion_r23506370>
:

>      String getDefaultURL();
>
> -    /** Return true if batch address validation is supported. If not, the batch address validation button
> -    will be disabled. **/
> +    /**
> +    * @description returns the help text to be shown when the user selects the service from the drop-down. It
> +    * may specify things such as the required fields.
> +    * @return string
> +    */
> +    String getHelpText();

Here, this is new - so it need to be implemented by everything that
implements this ADDR_IValidator interface

—
Reply to this email directly or view it on GitHub
<https://github.com/SalesforceFoundation/Cumulus/pull/1282/files#r23506370>.

davidhabib:
@kbromer @ceiroa @njjc Ok, I fixed the one merge issue I had missed, and got the build to run against this branch, and it passes!  so someone please review and merge, so we can get rid of this problematic branch!  thanks.

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -32,7 +32,8 @@
 * @date 2014
 * @group Settings
 * @description Test Class for the Install Script for the NPSP package. 
-*/@isTest
+*/
+@isTest
```
njjc: Do we have to change our test methods to use the testMethod annotation instead of @isTest for apexDocs purposes?

davidhabib: if you want them to filter by the testMethod pseudo-scope, yes.  There
still is the outstanding issue to detect @isTest, but I don't see that
happening anytime soon.  that said, this is lower in priority (in my
opinion) vs getting the descriptions in there.

Dave

PR #1289, opened 2015-01-21
#1241. support for New custom setting Household Account Addresses Disabl...

### LINKED ISSUE #1218: As a user, I would like to see if my custom naming formats are invalid
Currently, if you use a custom naming format, you have to move the cursor to another field and then come back to see how it's going to look. A person that enters three invalid formats and then saves (without moving to any other field) will not see any "invalid field" message:

screen shot 2014-12-16 at 10 06 32 am

### LINKED ISSUE #1241: As a user, I would like to be able to turn Address Management off for household accounts, so I can save data storage space.
https://powerofus.force.com/0D58000001jODza

davidhabib:
# Warning

# Info
* There is a new custom setting on the Address Settings page, Household Account Addresses Disabled.  This allows organizations who have data storage concerns to still use the Household Account model, but not use the new Address management features in NPSP 3.

# Issues
Fixes #1241 
Fixes #1218

### DIFF
--- a/src/classes/STG_PanelHouseholds_CTRL.cls
+++ b/src/classes/STG_PanelHouseholds_CTRL.cls
@@ -44,12 +44,18 @@ public with sharing class STG_PanelHouseholds_CTRL extends STG_Panel {
     public boolean isRunningBatch { get; set; }
 
     public PageReference activateHouseholdNaming() {
-        isRunningBatch = true; 
         boolean isActivation = (UTIL_CustomSettingsFacade.getHouseholdsSettings().npo02__Advanced_Household_Naming__c == false);
                 
         STG_Panel.stgService.stgHH.npo02__Advanced_Household_Naming__c = true;
+
+        // make sure any custom name formats are valid.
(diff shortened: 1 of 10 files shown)


src/classes/STG_PanelHouseholds_CTRL.cls
```diff
@@ -67,7 +94,11 @@ public with sharing class STG_PanelHouseholds_CTRL extends STG_Panel {
             ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.ERROR, system.Label.stgValidationHHAccountHHRules));
             return null;
         }
-        
+
+        // make sure any custom name formats are valid.
+		if (!isValidNameFormats())
```
ceiroa: Should we also be adding an error message to the page here?

ceiroa: Never mind, I see it's being done in isValidFormats itself.

ceiroa:
It all looks good to me. I'll merge it as soon as it's flagged as passing.

PR #1300, opened 2015-01-23
BUG/1285 As a user, changing an opportunity from a null amount to a positive amount should create a payment.

bermager:
Hello


mrbelvedere:
Your branch does not meet the feature branch naming requirements.  The name must use the format feature/NNN-description-of-feature where NNN is a valid GitHub issue number from the main repository.  Please rename your branch and submit a new pull request.

PR #1313, opened 2015-01-26
Feature/1239 bde checkboxes

### LINKED ISSUE #1239: Checkbox loses its default value of 'checked' in Batch Data Entry
I have a custom field that is a checkbox and the default value is set to checked.  The Batch Data Entry module does not keep the checkbox set to default - even when using the push pin feature.

screen shot 2014-12-28 at 9 29 55 pm
screen shot 2014-12-28 at 9 29 33 pm

davidhabib:
# Warning

# Info

# Issues
Fixes #1239


davidhabib:
@ceiroa @njjc @kbromer this pull request is good to go!

PR #1317, opened 2015-01-27
Affilations bug fix and ApexDoc work

### LINKED ISSUE #1315: BUG: Affiliations not working correctly in the 1:1 model
Hub post: https://powerofus.force.com/0D58000001oFO4q

note that Judi Sohn has duplicated the problem.

davidhabib:
# Warning

# Info

# Issues
Fixes  #1315

### DIFF
--- a/src/classes/ERR_Handler.cls
+++ b/src/classes/ERR_Handler.cls
@@ -188,6 +188,7 @@ public with sharing class ERR_Handler {
 	    	}
 	                
 	    	if(UTIL_CustomSettingsFacade.getErrorSettings().Store_Errors_On__c) {
+	    	   UTIL_Debug.debug('**** Storing errorRecords: ' + errorRecords);
 	    	   Database.insert(errorRecords, false);
 	    	}
 	    	//ERR_Notifier takes care of checking the settings itself.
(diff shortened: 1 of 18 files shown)


davidhabib:
@njjc @ceiroa @kbromer This pull request is ready for review and accepting.  It is mostly just ApexDoc updates, but it also contains a few line fix in ACCT_IndividualAccounts_TDTM.cls which fixes the bug where Affiliations were not being created in the 1:1 model.

src/classes/ERR_Handler.cls
```diff
@@ -188,6 +188,7 @@ public with sharing class ERR_Handler {
 	    	}
 	                
 	    	if(UTIL_CustomSettingsFacade.getErrorSettings().Store_Errors_On__c) {
+	    	   UTIL_Debug.debug('**** Storing errorRecords: ' + errorRecords);
```
kbromer: We want these debugs in?

davidhabib: I found them useful, to figure out what was going wrong in this unexpected scenario.  I see no harm in having them given the many other debug output strings we have.  remember, they are not output normally since the setting is off.

kbromer: We need to look at the overhead at some point, but youre right, this probably won't be the straw that hurts our camel

njjc: We disabled debug statements in the managed package context, unless you set
a setting, so we're not affecting performance for that anymore. Maybe test
performance a bit, but I think that's negligible.

PR #1320, opened 2015-01-27
Feature/run parallel tests 2

### LINKED ISSUE #406: As a developer, I would like to be able to run all tests in parallel
This would speed development. Right now we have to run then sequentially or they will fail due to errors accessing/modifying custom settings

ceiroa:
This is ready to be merged now. 

# Warning

# Info

# Issues
Fixes #406.


kbromer:
Has the code coverage remained constant throughout the changes?

ceiroa:
The code coverage is at 82%, according to my dev console. Not sure where we were before, but it should not have affected it much. We are still testing all the different settings configuration, it's just that we have not saving the settings to the DB during the tests.

screen shot 2015-01-28 at 9 13 57 am

ceiroa:
@davidhabib ~ could you take a look at this pull request? I forgot to ask you during the scrum.

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -94,7 +94,7 @@ private class ACCT_IndividualAccounts_TEST {
     	//Set Individual (Bucket) settings
     	npe01__Contacts_And_Orgs_Settings__c contactsSettings = UTIL_CustomSettingsFacade.getOrgContactsSettings();
     	contactsSettings.npe01__Account_Processor__c = CAO_Constants.BUCKET_PROCESSOR;
-        update contactsSettings;
+
```
davidhabib: is the return value from getOrgContactSettings an object that all code will use the same instance of during tests?

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -217,9 +214,7 @@ public with sharing class STG_InstallScript_TEST {
         npe04Settings.npe4__DISABLE_RelationshipCM_trigger__c = flagValue;
         npe04Settings.npe4__DISABLE_RelationshipContacts_trigger__c = flagValue;
         npe04Settings.npe4__DISABLE_Relationships_trigger__c = flagValue;
-        settingsToUpdate.add(npe04Settings);
-        
-        update settingsToUpdate;
+        settingsToUpdate.add(npe04Settings);
```
davidhabib: why did you leave in this call to settingsToUpdate.add(), but removed it from other places in this test class?

src/classes/STG_PanelRD_CTRL.cls
```diff
@@ -58,12 +58,16 @@ public with sharing class STG_PanelRD_CTRL extends STG_Panel {
...
         	List<SObject> toUpdate = new List<SObject>();
-        	if(STG_Panel.stgService.stgRD != null)
+        	
+        	if(STG_Panel.stgService.stgRD != null && !Test.isRunningTest()) {
         		toUpdate.add(STG_Panel.stgService.stgRD);
+        	}
+        	
         	if(rdSched != null) {
                 rdSched.Active__c = !isRDScheduleDisabled;
                 toUpdate.add(rdSched);
```
davidhabib: does this toUpdate().add need to be under a !isRunningTest()?

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -456,13 +493,12 @@ public without sharing class UTIL_CustomSettingsFacade {
         householdsSettings.npo02__Household_Creation_Excluded_Recordtypes__c = mySettings.npo02__Household_Creation_Excluded_Recordtypes__c;
         householdsSettings.npo02__Household_Mailing_List_ID__c = mySettings.npo02__Household_Mailing_List_ID__c;
         householdsSettings.npo02__Schedule_Job_Limit__c = 25;
-        upsert householdsSettings;
+        orgHouseholdsSettings = householdsSettings;
```
davidhabib: why doesn't this do the upsert under an !isTestRunning() like other places do?

src/classes/ACCT_IndividualAccounts_TEST.cls
ceiroa: Yes. The object will stay in memory, and can be modified there.

src/classes/STG_PanelRD_CTRL.cls
ceiroa: Well, this is interesting... If we were following the pattern we are following everywhere else, we would have needed to wrap the " update toUpdate" (just a little lower in the code) statement in !isRunningTest. We would also have had to wrap the saveAll method in saveAll with that condition.

However, these were not giving row lock errors, so I didn't add it. It seems like the test runner is treating controllers different from other tests, as saving settings from them don't produce these row lock errors.

We are calling saveSettings from a controller in quite a few places without having to do anything about it:

screen shot 2015-01-29 at 9 50 06 am

I would leave it as is (why put the condition if we don't need it), but we can talk about it when we meet.

src/classes/UTIL_CustomSettingsFacade.cls
ceiroa: The upsert was removed.

src/classes/STG_PanelRD_CTRL.cls
ceiroa: I think it's worth noting about this that I didn't just go and wrapped every DML statement on custom settings in a !Test.isRunningTest() condition. I did it one by one as I was getting a row locking error for each. Which means I only did it for those that were completely necessary.

PR #1322, opened 2015-01-28
Cuz it looks so much cooler in red.

ceiroa:
Polishing doorknobs.


kbromer:
Now this is a pull request I can get behind.

PR #1341, opened 2015-02-03
Adding links to existing posts.

ceiroa:
@njjc Just adding the links to couple of existing articles, so I don't forget about them. (If you are doing this somewhere else we can delete this branch and pull request.)


ceiroa:
@njjc ~ This is ready to be merged. Unfortunately, it doesn't solve the issue with some tests failing randomly if we run all tests in parallel. But I don't want to keep blocking the feature builds. So I have changed the feature org back to running tests sequentially, and @jlantz and I have talked about creating a separate feature org where we can keep investigating this issue.

PR #1347, opened 2015-02-05
Feature/1286 store errors checkbox

### LINKED ISSUE #1286: As an admin, I expect error email notifications to be sent out, even if the errors are not being stored
After some of the refactoring work done a few releases ago, the content of the error notifications (either email or chatter) is pulled from the errors table. If there are no error records, no notifications are sent.

It might be enough with putting in some JQuery in the Error Settings page that automatically checks the "Store errors" checkbox if a user checks the "Send error notifications" checkbox.

### LINKED ISSUE #1321: We have lost filling the Zipcode Endpoint textbox automatically when a user selects SmartyStreets as the address verification service
When the drop-down changes to SmartyStreets, both endpoint values should be filled. Maybe we lost it when we re-did theUI.

ceiroa:
@jvarese, @njjc - For issue #1321, I have actually completely removed the "Zipcode Verification URL" textbox from the UI. We were still showing it for the 3 address verification services, even though only SmartyStreets makes use of it. I think removing it makes it less confusing for users. I'm instead just reading the value directly in the class that sends out the request. Let me know what you think.  

(Following this idea, I guess we could also remove the "Address Verification URL" completely from the UI. The only advantage of having it there is in case a user needs to modify it because it has changed, or something like that. Though I don't think it's very likely... But I guess this would be a different story.)

# Warning

# Info

# Issues
Fixes #1286.
Fixes #1321.


jvarese:
Nice job @ceiroa . I wholeheartedly approve of removing everything that a user doesn't need for a particular service.

PR #1348, opened 2015-02-06
#1323. In some orgs, Affiliations are not created with an Account's Primary Contact field is specified

### LINKED ISSUE #1323: BUG: In some orgs, Affiliations are not created with an Account's Primary Contact field is specified.
hub thread https://powerofus.force.com/0D58000001mccZQ

Kalman has granted login access to orgId 00Dj0000000HmbW

I have verified the Affiliation is not created because of the following line 58 in AFFL_Affiliations_TDTM

        } else if (objResult == Schema.Sobjecttype.Account && !hasRunForAccounts) {

I have verified in the debug log that the objResult == Schema.Sobjectype.Account is failing.  How can this be?  The Account trigger passes in Schema.Sobjecttype.Account which makes it to this run method which then compares against the same thing.  Note that Schema.Sobjectype.Account is an object.  Are there some scenarios where the platform may give you different instances of the describe object?  Maybe we should be doing 

if (objResult.getLabel() == 'Account') instead?

davidhabib:
# Warning
# Info
# Issues
Fixes #1323

### DIFF
--- a/src/classes/AFFL_Affiliations_TDTM.cls
+++ b/src/classes/AFFL_Affiliations_TDTM.cls
@@ -47,15 +47,15 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
             
         // dispatch to the correct handler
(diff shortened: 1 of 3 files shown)


davidhabib:
@ceiroa or @njjc ready for review.  details in the bug report #1323.

PR #1367, opened 2015-02-09
Updating README.md with information about code contributor's guide and a...

njjc:
...pex docs.


njjc:
@jlantz whenever you're ready. Also, it seems like the status links are broken at the top of the readme.

PR #1371, opened 2015-02-10
Adding Data Import to the API.

ceiroa:
What do you think, @davidhabib?

# Warning

# Info
* NPSP data import exposed through the API.

# Issues
Fixes @1324.

### DIFF
--- a/ApexDocContent/API.htm
+++ b/ApexDocContent/API.htm
@@ -12,7 +12,7 @@
 	
 		<p>The classes involved with the public interface include:</p>
 		<ul>
-		    <li>DBI_DataImport_API</li>
+		    <li>BDI_DataImport_API</li>
 			<li>ERR_Handler_API</li>
 			<li>TDTM_Config_API</li>
 			<li>UTIL_CustomSettings_API</li>
--- a/ApexDocContent/API.htm
+++ b/ApexDocContent/API.htm
@@ -3,6 +3,7 @@
 		<h2>Nonprofit Starter Pack API (Application Programming Interface)</h2>
 		<p>The public NPSP API is intended for external use outside the managed package, and includes classes that expose functionality around:</p>
 		<ul>
+		    <li>Data Import</li>
 			<li>Error Handling</li>
 			<li>Table-Driven Trigger Management (TDTM)</li>
 			<li>Custom Settings</li>
@@ -11,6 +12,7 @@
 	
 		<p>The classes involved with the public interface include:</p>
 		<ul>
+		    <li>DBI_DataImport_API</li>
 			<li>ERR_Handler_API</li>
 			<li>TDTM_Config_API</li>
 			<li>UTIL_CustomSettings_API</li>
--- a/src/classes/BDI_DataImport_API.cls
+++ b/src/classes/BDI_DataImport_API.cls
@@ -45,6 +45,7 @@ global with sharing class BDI_DataImport_API {
     */
     global static ID importData() {
     	Data_Import_Settings__c diSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
+    	Boolean batchRunning = false;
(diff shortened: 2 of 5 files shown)


src/classes/BDI_DataImport_API.cls
```diff
@@ -0,0 +1,102 @@
...
+            }
+            
+            BDI_DataImport_BATCH bdi = new BDI_DataImport_BATCH();
+            apexJobId = Database.executeBatch(bdi, integer.valueOf(diSettings.Batch_Size__c));
+            return apexJobId;
+
+        } catch (Exception ex) {
+            // no reason to log the error if we haven't started the batch,
+            // since it is just a validation error that we are displaying on the page.
+            if(apexJobId != null)
```
davidhabib: checking apexJobId is incorrect, since it is uninitialized and would never be set if we hit the exception.  the original code set a variable detecting that it had made it thru validation and was calling executeBatch.

ApexDocContent/API.htm
```diff
@@ -11,6 +12,7 @@
 	
 		<p>The classes involved with the public interface include:</p>
 		<ul>
+		    <li>DBI_DataImport_API</li>
```
davidhabib: It should be BDI_DataImport_API, not DBI_

src/classes/BDI_DataImport_API.cls
```diff
@@ -0,0 +1,102 @@
...
+
+     private class BDIException extends Exception {}
+     
+    /*******************************************************************************************************
+    * @description Imports all DataImport records that have not been previously imported. Their status is changed
+                    to imported, once the process fishishes.
+    * @return ID The ID of the batch job that performs the data import.
+    */
+    global static ID importData() {
+    	Data_Import_Settings__c diSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
```
davidhabib: I'm uncomfortable that you are now getting the settings from the facade, rather than the version already stored on the page.  are you 100% sure this will be the same object, and thus you haven't accidentally broken the pages settings?

src/classes/BDI_DataImport_CTRL.cls
```diff
@@ -86,72 +84,23 @@ public with sharing class BDI_DataImport_CTRL {
...
-	                // done an upsert on it.  This failure only occurs if there is no org defined
-	                // custom setting instance.  The workaround is to catch the upsert failing, and
-	                // just clear the Id and do an insert.
-	                diSettings.Id = null;
-	                insert diSettings;
-	            }
-            }
-            
+    	if(!Test.isRunningTest())
+    	   upsert diSettings;
```
ceiroa: Does this address your concert regarding reading settings directly from the Facade in the DataImport_API class?

src/classes/BDI_DataImport_API.cls
```diff
@@ -0,0 +1,104 @@
...
+            
+            batchRunning = true;
+            BDI_DataImport_BATCH bdi = new BDI_DataImport_BATCH();
+            apexJobId = Database.executeBatch(bdi, integer.valueOf(diSettings.Batch_Size__c));
+            return apexJobId;
+
+        } catch (Exception ex) {
+            // no reason to log the error if we haven't started the batch,
+            // since it is just a validation error that we are displaying on the page.
+            if(batchRunning)
```
ceiroa: @davidhabib ~ I'm back to using a local boolean that gets set before we call executeBatch. How does that look? What I want to make sure of is that we don't need to pass any unnecessary parameters to the importData method.

davidhabib: that works.  thanks.

src/classes/BDI_DataImport_CTRL.cls
davidhabib: great!

PR #1372, opened 2015-02-11
Feature/1318 state country picklists

### LINKED ISSUE #1318: As a user, I do not want an address record to be created if no address information is entered in an account, but State and Country picklists are enabled and a default Country value is selected
With state and country picklists enabled and a default country value selected, the tests newAddrForOrgAccounts and updateOrgAccounts in ADDR_Addresses_TEST currently fail.

Taking the example of updateOrgAccounts, if picklists are enabled it fails because address records are created when the accounts are created. This happens because of the default value automatically given to the account country. 

Setting ADDR_Addresses_TDTM.hasRunAddrTrigger to false before doing the account update should not be necessary, because no address records should have been created on insert. If we to set it, then we get an error because twice the number of address records get created (1 per account insert, and 1 per account update).

We should consider whether a new address record should actually be created if the only address information in the existing record is the country. Maybe we should update the existing record instead, specially if the country does not change.

### LINKED ISSUE #1319: As a developer, I do not want tests in HH_ManageHHAccount_TEST to fail if Country and State picklist are enabled, and a default Country is selected
The following tests in HH_ManageHHAccount_TEST fail in that case:

	testRemoveContactActiveAddress -- System.JSONException: Cannot deserialize instance of ShippingAddress from FIELD_NAME value city at [line:1, column:15386]
		Stack trace: (System Code)
		Class.HH_ManageHousehold_EXT.unpackState: line 319, column 1
		Class.HH_ManageHousehold_EXT.remoteRemoveMember: line 361, column 1
		Class.HH_ManageHHAccount_TEST.testRemoveContactActiveAddress: line 223, column 1

	testRemoveReAdd -- System.JSONException: Cannot deserialize instance of ShippingAddress from FIELD_NAME value city at [line:1, column:12189]
		Stack trace: (System Code)
		Class.HH_ManageHousehold_EXT.unpackState: line 319, column 1
		Class.HH_ManageHousehold_EXT.remoteRemoveMember: line 361, column 1
		Class.HH_ManageHHAccount_TEST.testRemoveReAdd: line 601, column 1

davidhabib:
# Warning

# Info

# Issues
* Fixes #1318 
* Fixes #1319


davidhabib:
@ceiroa @njjc This pull request is ready.  The two fixes are for bugs uncovered when an org had State & Country Picklists enabled, and had a default country set.

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -108,12 +108,19 @@ public with sharing class ADDR_Account_TDTM extends TDTM_Runnable {
...
-        return (
+        boolean isSpecified = 
            acc1.BillingStreet != null ||
            acc1.BillingCity != null ||
            acc1.BillingState != null ||
            acc1.BillingPostalCode != null ||
-           acc1.BillingCountry != null);
+            // note that we decided to ignore country so that a default value won't create unnecessary address objects
+           (acc1.BillingCountry != null && !ADDR_Addresses_TDTM.isStateCountryPicklistsEnabled);
+
```
ceiroa: In this second comparison, we want to see if there is a state code, when the picklists are enabled, even if nothing else is specified (or even if country was set automatically)? The way this second condition is written, separate from the others, makes the logic a little hard to understand. Why do we need to check it only if nothing else is set? Can we make it part of the first condition? Also, I think the same comment is there twice.

davidhabib: the first case is looking at Billing Country.  the second set is looking at the state code which must be done dynamically.  yes, the comment is twice on purpose to explain these two special cases.

ceiroa: The second comment also says "country". Shouldn't it say "state"? 
We can leave the code like that (wouldn't stop the pull request for this or the comment), but I think it'd be easier to understand if we had this also included, with another OR, in the first condition.

davidhabib: Nope.  we are looking at StateCode.  we are avoiding looking at CountryCode.

davidhabib:
@ceiroa I redid the if statements per your request.

ceiroa:
@davidhabib I'm ready to merge it, but it's not flagged as passing build.

davidhabib:
just sent email to Jason.  the feature build isn't automatically starting.


davidhabib:
@jlantz @ceiroa this has now run through the feature build and passed, but the checkmark is updating on this listing.  why not?

PR #1381, opened 2015-02-12
Feature/1316 nic apexdocs

njjc:


src/classes/AFFL_Affiliations_UTIL.cls
```diff
@@ -1,48 +0,0 @@
...
-*
-* @description  
-*/
-
-public with sharing class AFFL_Affiliations_UTIL {
-
-    public static npe5__Affiliations_Settings__c affiliationsSettings;
-    
-    //DEPRECATED - DO NOT USE!!! Use method in UTIL_CustomSettingsFacade instead.
-    //We are keeping this class just because without it we cannot deploy to the packaging org.
```
njjc: @ceiroa Is this still the case? We can't delete this class?

njjc: @davidhabib

ceiroa: I'm not sure. My guess is that I removed it once, and the dev flow failed because of this. I guess we can try again, unless @jlantz has any concerns.

njjc:
@jlantz is this a known issue?

PR #1382, opened 2015-02-13
Feature/1310 dynamic namespacing

### LINKED ISSUE #1310: As a developer, convert our namespace handling to not assume npsp, but use whatever packaging org namespace we are in.
Here are the code snippets I mentioned for dynamically looking up namespace:

Our initial Apex code implementation of this revolves around this utility class.

```apex
public with sharing abstract class ApplicationUtilities 
{
    public static String getNamespace()
    {   	
        if (plainNamespace == null)
        {
            String withDotNotation = ApplicationUtilities.class.getName();  
        
            if (withDotNotation.contains('.'))
            {
                plainNamespace = withDotNotation.substringBefore('.');
            }
            else
            {
                plainNamespace = '';
            }
        }
        return plainNamespace;   
    }
    
    public static String getNameSpacePrefix(); // snipped -- returns with  '__' or '' -- similar to above
    public static String getNamespaceDotNotation(); // snipped -- returns with '.' or ''  -- similar to above
}
```
Then anywhere with a Dynamic Query looks up field values through:
```
account.get(ApplicationUtilities.getNameSpacePrefix()+'CustomField__c');
```
or if you need a Class Name through a string:
```
String className = ApplicationUtilities.getNamespaceDotNotation()+'PolicyController';
```
Then one trickier place is OnClick JavaScript buttons. This of course would fail with multiple ApplicationUtilities classes. 

```javascript
{!REQUIRESCRIPT("/soap/ajax/30.0/connection.js")}
{!REQUIRESCRIPT("/soap/ajax/30.0/apex.js")}
var namespace_prefix = ''; 
var gettingnamespace = sforce.connection.query("SELECT NamespacePrefix FROM ApexClass where Name = 'ApplicationUtilities' LIMIT 1"); 
var getname = gettingnamespace.getArray("records"); 

if(getname.length > 0) 
{ 
        namespace_prefix = getname[0].NamespacePrefix; 
}

if (namespace_prefix==null || namespace_prefix=='')
	window.location = 'apex/InsuranceSelectionUI?accountId=' + '{!Account.Id}';
else
	window.location = 'apex/'+namespace_prefix+'__InsuranceSelectionUI?accountId=' + '{!Account.Id}';
```
In any Visualforce page / component we just always pass the namespace from the controller and then use '{!namespacePrefix}' where necessary. Which works for visualforce and JavaScript in the page.

Controller:
```apex
public String namespacePrefix 
{
	get
	{
		return ApplicationUtilities.getNameSpacePrefix();
	} 
}
```
Page:
```
$scope.namespacePrefix = '{!namespacePrefix}';
```
I think those are the really common scenarios, and all seem to work very well for us. As I said before, Flows are the one place where we are actively looking into potential solutions. Let me know if I missed any others. 

Thanks again!

Adam Rutland
Vlocity

njjc:
# Warning
# Info
We now support dynamic namespacing, allowing our unmanaged code to be used in an instance with any namespace. Buttons are still hard coded to the NPSP namespace. 
# Issues
Fixes #1310


njjc:
This is good to go, fully tested in an org with a new namespace and with no namespace.
@davidhabib @ceiroa

src/classes/BDI_DataImport_BATCH.cls
```diff
@@ -239,8 +239,20 @@ public with sharing class BDI_DataImport_BATCH implements Database.Batchable<sOb
...
+                    //also removes tokens from actual detected namespace, if we happen to be
+                    //in a non-npsp namespace
+                    if (UTIL_Namespace.getNamespace() != 'npsp') {
+                        string strFieldOld = strField;
+                        strField = UTIL_Namespace.StrTokenRemoveNSPrefix(strField);
+                        
+                        //if we have a detected namespace, and removed a namespace from the current field
+                        //add the current namespace back in, allowing this code to run in any namespace
+                        if (strField != strFieldOld && UTIL_Namespace.getNamespace() != '') {
+                            strField = UTIL_Namespace.StrAllNSPrefix(strField);
```
davidhabib: The namespace of the field could be to any of the old packages as well, or a 3rd party package.  Won't this code end up changing it to the cumulus test namespace, which is incorrect?

njjc: It removes the NPSP namespace, and any current active namespace. So if you have an org with the namespace "footest", it will remove npsp__ and footest__. It won't remove npe01, or any other namespace.

njjc: Although now that you mention it, maybe we should just remove NPSP__ and not any other namespace.

davidhabib: i'm wondering about line 252 where you re-prefix the string.  won't this mess up?

njjc: It only applies to strings that have changed, i.e. had the npsp namespace prefix removed.

njjc: We're removing the npsp__ namespace from the help text if we're not in the npsp namespaced package. Then if we're in another namespace, we're adding the namespace back to the ones we just removed npsp__ from so the field describes are handled correctly.

davidhabib: looks to me like you are removing namespaces to all fields if running in the non-npsp namespace.

njjc: @davidhabib should we schedule a time to talk this through after sprint review?

PR #1402, opened 2015-02-19
Feature/1391 bdi

### LINKED ISSUE #1391: BUG: Attemping to Import Records with Custom Field as Unique ID Generating Field Matching Error with NPSP Data Importer
https://powerofus.force.com/0D58000001s09fE

davidhabib:
# Warning

# Info

# Issues
Fixes #1391

### DIFF
--- a/src/classes/BDI_DataImport_BATCH.cls
+++ b/src/classes/BDI_DataImport_BATCH.cls
@@ -119,7 +119,12 @@ public with sharing class BDI_DataImport_BATCH implements Database.Batchable<sOb
             if (isCustomIdInContactDatatypeString == null) {
                 isCustomIdInContactDatatypeString = false;
                 if (isCustomIdInContactMatchRules) {
-                    if (UTIL_Describe.getFieldType('Contact', diSettings.Contact_Custom_Unique_ID__c) == 'STRING')   
+                    Schema.Displaytype dt = UTIL_Describe.getFieldDisplaytype('Contact', diSettings.Contact_Custom_Unique_ID__c);
+                    if (dt == Schema.Displaytype.String || 
+                        dt == Schema.Displaytype.TextArea || 
+                        dt == Schema.Displaytype.Id ||
+                        dt == Schema.Displaytype.URL ||
+                        dt == Schema.Displaytype.Reference)
                         isCustomIdInContactDatatypeString = true;
                 }
             }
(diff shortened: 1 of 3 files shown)


davidhabib:
@ceiroa This is ready for review.  thanks.

src/classes/BDI_DataImport_BATCH.cls
```diff
@@ -465,8 +504,15 @@ public with sharing class BDI_DataImport_BATCH implements Database.Batchable<sOb
                 setEmail.add(di.Contact1_Work_Email__c);            
             if (di.Contact1_Alternate_Email__c != null)
                 setEmail.add(di.Contact1_Alternate_Email__c);            
-            if (isCustomIDInContactMatchRules && di.get('Contact1_' + diSettings.Contact_Custom_Unique_ID__c + strTestContactUniqueIdSuffix) != null)
-                setCustomID.add(string.valueOf(di.get('Contact1_' + diSettings.Contact_Custom_Unique_ID__c + strTestContactUniqueIdSuffix)));
+            if (isCustomIDInContactMatchRules) { 
+                object val = di.get('Contact1_' + diSettings.Contact_Custom_Unique_ID__c + strTestContactUniqueIdSuffix);
```
ceiroa: For my understanding, could you explain why are we adding __c only if we are running tests?

ceiroa:
Just a minor observation (not a reason to block the merge); the code for accounts and contacts seems almost identical. It would be nice if it could be combined in common functions that take as parameters the strings that are different in each case.

src/classes/BDI_DataImport_BATCH.cls
davidhabib: It's a hack.  In the non-test scenario, the settings for the Account and Contact Custom Unique ID field is always a custom field, and the setting would contain the developer name of the field (and thus have __c).  But for my tests to try this functionality, I can't create any custom fields, and thus have to use some Standard Salesforce fields on either Contact or Account.  Thus I have to put the Standard field name in the setting, and use that when copying to the Contact or Account.  But that standard field which is also on the BDI object has the __c since the BDI is a custom object.  So it's a hack to say, if we are running tests, and the setting to use a custom id field is set, then we know it's a standard field against contact or account, but we still need to add the __c to access the field in the BDI object.

davidhabib: Hold off on accepting the pull request, and I'll update the code with some better comments this weekend or Monday.  I'm not sure why I didn't fully comment this when I originally wrote it.  A bit sloppy!

davidhabib:
@ceiroa OK, I cleaned up the code on this pull request.  thanks for the feedback.  once it passes build, you can merge.  thanks.

PR #1432, opened 2015-03-03
more ApexDoc for Recurring Donations.

davidhabib:


davidhabib:
@njjc This pull request is ready for review and merge. thanks.

src/classes/RD_AddDonationsBTN_CTRL.cls
```diff
@@ -30,13 +30,26 @@
 /**
 * @author Salesforce.com Foundation
 * @date 2012 (2.2)
+* @group Recurring Donations
```
njjc: We need a group-content tag as well.

src/classes/RD_RecurringDonations.cls
```diff
@@ -30,6 +30,10 @@
 /**
 * @author Salesforce.com Foundation
 * @date 2012 (2.0)
+*
```
njjc: I've been not leaving any empty lines here, I think we should follow that for consistency.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -1741,8 +1741,10 @@ public class RD_RecurringDonations_TEST {
         controller.saveCISettings();
     }
 *************************************************************************/
-   
-   // from RD_RecurringDonations_SCHED  
+
+   /**
+   * @description from RD_RecurringDonations_SCHED
```
njjc: This description seems incomplete.

njjc:
@davidhabib made some notes, let me know when they're fixed and I'll merge!

src/classes/RD_AddDonationsBTN_CTRL.cls
davidhabib: only one file in the group should include the group content tag, and I believe I have it in RD_RecurringDonations.cls

ceiroa: "Should" or "needs to"? It's ok if more than one has it, right?

davidhabib:
@njjc items addressed.  ready for merge.  build is rerunning.

src/classes/RD_AddDonationsBTN_CTRL.cls
davidhabib: if you have multiple, probably the last one parsed wins.  so it's a "should".

njjc: Interesting, I wasn't aware of that.

PR #1436, opened 2015-03-04
Feature/1353 google geo strips info

### LINKED ISSUE #1355: As an admin, I don't need to see the Address Verification Endpoint for each service
We can just remove the textbox that we auto-populate, and get the validator to pull the endpoint directly from each class.

We can also deprecate the settings (endpoint and zipcode endpoint).

Also, we shouldn't show in view mode the auth credentials (ID or Token) that don't need to have a value. For example, when in in edit mode, if a user selects Google Geocoding, we hide the Auth Token field. However, we show in in View mode, and also when the user clicks edit (before making any changes to the drop-down).

### LINKED ISSUE #1408: Address Verification Settings page is hiding the textbox where user needs to enter Google Geocoding Token
The field our Google validator is pulling the token from is not the one linked to the textbox that is displayed when Google is selected from the dropdown. Thus user gets the "MISSING AUTHID OR ENDPOINT" message.

We need to test if the same is happening to Cicero.

### LINKED ISSUE #1415: Fatal error should not occur when the size of the address verification response is larger than the max size of the API Response field
This happens when Google thinks the address is ambiguous. When should trim the response to make it fit.

screen shot 2015-02-24 at 11 20 04 am

### LINKED ISSUE #1435: Google address verification strips last 4 digits from a 9-digit zipcode
From https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe&fId=0D58000001qKr8p

Amanda Byrne says:

Regarding Google Address verification: Now that we have address verification, we've noticed some less than ideal behavior. Many of our addresses have the 9 digit zip-code. Google address verification is stripping the extra digits off and leaving only the 5 digit version.

ceiroa:
# Warning

# Info

# Issues
Fixes #1435
Fixes #1408 
Fixes #1415 
Fixes #1355


ceiroa:
@davidhabib ~ Since you and I had already talked about what to do with the endpoint field in the custom setting, I think it'd be good that you review this. At the end I decided to still read from the setting, which would allow a user to manually modify the endpoint. This could be useful in case they wanted to change it to a test endpoint. But I still removed it from the UI for simplicity. Then I had to make sure that the endpoint was being stored in the settings when the user saves them, even if it's not getting it from a text field.

I would recommend looking at the diff with the w=1 url param, since the vf page has quite a few space changes.

I made the changes to the tests (putting the error number in the address instead of the endpoint) when I thought I was going to remove the endpoint field. Now that we are still using it, these are unnecessary, but it didn't seem worth it to work on reverting them.

I have split the info truncation issue into two, and only the one about zip codes goes in this pull request.

PR #1456, opened 2015-03-14
As a contributor, I want to have a simple way to submit a contribution to the NPSP

mrbelvedere:
Testing adding some new configuration to the package.
Add 2 new validation rules on Contact to require values in Home/Mobile phone if the Preferred Phone is set to those values.

# Critical Changes

# Changes

# Issues


davidhabib:
@force2b Michael, is this something you are really trying to get into NPSP?  @njjc and I have actually been talking about removing or changing these validation rules in the template, because they cause lots of trouble with integrations that have no knowledge of these NPSP specific fields.  So I don't think we want to start including them!

jlantz:
@force2b was demoing the mrbelvedere contribution tool for me.  Closing this pull request.

PR #1461, opened 2015-03-17
Feature/1353 google geo strips info

### LINKED ISSUE #1353: Google and Cicero address verification strip apartment number
From https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe&fId=0D58000001qKr8p

Amanda Byrne says:

Regarding Google Address verification: Now that we have address verification, we've noticed some less than ideal behavior:

1. Many of our addresses have the 9 digit zip-code. Google address verification is stripping the extra digits off and leaving only the 5 digit version.

2. Apartment and unit numbers are usually removed during verification. This is a big problem.

We've noticed that the API returns the apartment number as a 'subpremise' like so:
{ 
"long_name" : "117", 
"short_name" : "117", 
"types" : [ "subpremise" ] 
}, 

But I don't know if it would be reliable to make any 'subpremise' an apartment number- it sounds like a pretty generic field name- like a 'scrap left over' field.

There is an NPSP field for Mailing Address 2, which we've never used- we enter the street and Apt or unit # in the Mailing Street. However, the verification process did not add the 'subpremise' there.

Sadly, we've decided to turn off address verification. So promising.

njjc:
# Warning
# Info
- Address verification now saves address before verification in "Pre-Verification Address" field on the Address object.
- Address verification with Google and Cicero now removes anything in the street address after a comma from the request, and adds it back after. This allows for adding a subpremise after the comma, e.g. "Apt. 5" or "Suite 200", without it being lost in verification.

# Issues
Fixes #1353


njjc:
@ceiroa could you make sure I fixed the correct webservices? I only did Cicero and Google per a note I took from you that I believe said only to do those two. Also, I couldn't test the subpremise addition/removal for Google as I couldn't understand why our test streets have API codes for the streets. We can discuss in the morning before you merge.

ceiroa:
The build failed, so that will need to be addressed first. 

Second, we need bullet points in the Info section to avoid all text being crumpled together for the release notes.

You are correct that, afaik, only Cicero and Google need the fix. As far as I've seen, SmartyStreets actually gives you the apartment number back, if you sent it in the request. But I would do some manual testing against all 3 real services to confirm. 

The test addresses have http codes in the street field just for the class that sends the mock responses to know which response to send.  Take a look at lines 143 to 155 of ADDR_MockHttpRespGenerator_TEST.

davidhabib:
I think it would be worth calling out in the release notes (pull request comments) that people might want to consider adding the new Pre-Verification Address field to the page layout for Address objects (and we should make sure we do so in the package and the DOT).

njjc:
@davidhabib I thought we decided as a group to leave that field off the layout.

ceiroa:
I think Dave is correct...

On Tue, Mar 17, 2015 at 11:31 AM, Nicolas Campbell <notifications@github.com

njjc:
You mean you want the pre-verification address field on the layout? That's not what we decided when we discussed the feature with Kevin, but I'm fine with doing that.

njjc:
@ceiroa This is ready to go. We decided to leave the field off the layout as originally planned, to not worry the average user with the technical stuff but to allow an admin to find the info if needed. I did some manual testing and it looks good.

ceiroa:
@njjc ~ Did you confirm through manual testing that the fix does not need to be added to SmartyStreets?

src/classes/ADDR_GoogleGeoAPI_Validator.cls
```diff
@@ -99,10 +99,23 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
...
+        
+        if (!String.isBlank(a.MailingStreet__c))
+            mailingStreet += a.MailingStreet__c.trim();
+        if (!String.isBlank(a.MailingStreet2__c))
+            mailingStreet += ' ' + a.MailingStreet2__c.trim();
+
+        //if our address has a comma, split it off and append it after verification
+        if (mailingStreet.contains(',')) {
+            list<string> split = mailingStreet.split(',',2);
+            address_request_string += 'search_address=' + split[0].replace(' ', '+') + '&';
```
ceiroa: Did you manually test this works? I'm not sure Google takes ampersands, or only plus signs as separators.

njjc:
@ceiroa Yes, it keeps the subpremise but removes the comma, which is fine by me.

src/classes/ADDR_GoogleGeoAPI_Validator.cls
njjc: Ah, you're right, let me switch it back and run a test.

njjc:
@ceiroa Merged in google change, verified all 3 services work with subpremise after comma. The only issue currently is that the pre-verification field contains <br> tags instead of new line characters, do you think that's worth fixing?

ceiroa:
@njjc ~ Yeah, I think we should replace those. It should be an easy fix, I would think.

njjc:
@ceiroa Really ready this time :)

ceiroa:
I just realized you added a test to verify the new functionality with Google, but I don't see one for Cicero. Should we have one too?

njjc:
I had already modified the test for cicero.

PR #1481, opened 2015-03-25
Feature/1455 always rollup hh account

### LINKED ISSUE #1370: Average Gift calculating a zero incorrectly
For some but not all Accounts the Average Gift field on Accounts is showing a $0 when multiple gifts have been given and tallied in the other rollup fields.

For example a record will show:
Total Gifts	
$252,650.00
First Gift Date	
1/1/1992
Largest Gift	
$250,000.00
Last Gift Date	
12/15/1997
Smallest Gift	
$25.00
Best Gift Year	
1997
Average Gift	
$0.00

I have confirmed this issue in two separate instances both running NPSP 3.34.
In one account this affects 17k records out of 33K that should have an average gift >0
In the other account this affects 1k records of 2.5k that should have an average gift>0

### LINKED ISSUE #1458: As a user, I would like to know that the Account Converter will leave my old Household objects around, so I don't get confused and think it didn't work.
https://powerofus.force.com/0D58000001wmuwl

### LINKED ISSUE #1460: As a user, when viewing the Address object, I would expect the Household Account field to just be labeled Account, since NPSP now supports organizational accounts for addresses as well.

njjc:
# Warning
# Info
# Issues
Fixed #1370
Fixed #1458 
Fixed #1460


njjc:
Ready for review! @davidhabib

PR #1484, opened 2015-03-26
Removing controller check for inactive accounts/cons/opps/tasks

### LINKED ISSUE #1438: As a user on Spring 15, the Household Model converter shouldn't ask me to verify inactive user records are transferred since its no longer a requirement.
https://success.salesforce.com/ideaview?id=08730000000Br8MAAS

https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/CONV_Account_Conversion_CTRL.cls#L234

kbromer:
# Warning
# Info
- Removes the requirement that Accounts/Opptys be owned by active users.  Requirement remains for Tasks/Contacts.  This accommodates changes made in Salesforce Spring 15 release

# Issues
Fixes #1438

### DIFF
--- a/src/classes/CONV_Account_Conversion_CTRL.cls
+++ b/src/classes/CONV_Account_Conversion_CTRL.cls
@@ -231,29 +231,28 @@ public with sharing class CONV_Account_Conversion_CTRL {
             ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.WARNING, errorString));
         }
 
-        /* Spring 15 removed the requirement that users must be active
+
         List<User> inactiveUsers = new List<User>();
         inactiveUsers = [select id from User where isActive = false order by LastModifiedDate desc limit 9999];
-        Integer accountCount = [select count() from Account where OwnerId IN :inactiveUsers limit 200];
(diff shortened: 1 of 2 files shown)


njjc:
I didn't know this was in Spring '15, super excited about it.

kbromer:
I need to confirm the precise behavior still, it seems like a subset of objects, but I can't get a clean answer on which ones they are!!

kbromer:
Need to modify my request here based on release notes: 

"Previously, only administrators were able to edit accounts, opportunities, and custom object records
that are owned by inactive users. With Spring ’15, administrators and all users with the create or
edit permission can create or edit accounts, opportunities, and custom object records that are
owned by inactive users. For example, you can create an account and assign an inactive user as the
record owner. Or you can edit the Account Name field on an opportunity record that is owned
by an inactive user to change its parent account."

Sounds like tasks/contacts would still need to be owned by active users

kbromer:
Okay @njjc this should be good.  Removed the check for Accounts/Opptys, kept it for Tasks/Contacts per the release notes

njjc:
Funny that they only went by half measures.

PR #1485, opened 2015-03-26
Feature/1462 health check

### LINKED ISSUE #1365: As a developer, add a test case for the Household Naming specification validation rules when saving the NPSP settings page

### LINKED ISSUE #1379: As a user, warn me in Health Check if both One-to-One and Household Account models specify the same recordtype in settings.
This is the 2nd time this week this has come up. What do you think about adding some text to the NPSP Settings > People page to tell folks that both Household and One to One shouldn't use the same record type? And maybe having Health Check look for this?

### LINKED ISSUE #1405: Data Importer: when creating a Campaign, Active=False
Issue: When Data Importer creates a Campaign, it sets Campaign.Active=False.  This makes the Campaign and other data invisible in SF.

Proposed solution: DI should set Active=True for created Campaigns.

### LINKED ISSUE #1462: As a user, if Health Check hits a Salesforce error while testing saving an opp, provide some helpful text in addition to the salesforce message.
Health Check special cases some errors, but if it doesn't recognize the exception, it is only putting in the exception message.  It should also add text about what it was doing and what the user should look into.  Also, this test is done by Recurring Donations, but it really is about saving Opportunities, not RD's, so change the topic name.    
Hub thread: https://powerofus.force.com/0D58000001wnBix

davidhabib:
# Warning
# Info
# Issues
Fixes #1462 
Fixes #1379 
Fixes #1365 
Fixes #1405


davidhabib:
@njjc This one is ready for review and merge.  It includes updating all the metadata files to the correct format.  thanks.

davidhabib:
I just accepted the pull request with Judi's work, which also included all the metadata file changes, so hopefully the merge into this branch will work and this pull request will be greatly reduced!

davidhabib:
@njjc Can you please review and merge.  I'd like to create a new branch which has this work in it.  thanks!

njjc:
@davidhabib done!

PR #1502, opened 2015-04-14
Feature/1488 addr verification fixes

### LINKED ISSUE #1476: BUG: Including # in an address (such as Apt #301) causes an invalid response error from Cicero
Not sure if we need to encode #'s and other special characters, or just remove them.  But they currently will cause the validation to fail.

### LINKED ISSUE #1488: BUG: Google geocoding completely deleting street portion of address
With this address:

1203 Kenray Loop
Springfield, OR 97477

[Per this hub thread.](https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe&fId=0D58000001zsPjC&emtm=1427757534469&s1nid=0DB800000004Fa7&OpenCommentForEdit=1&fromEmail=1&emkind=chatterCommentNotification&s1ext=0&s1oid=00D00000000hWLM&s1uid=00580000006ktn4)

### LINKED ISSUE #1491: BUG: Addresses sub-tab in NPSP Settings errors on save
On the People > Address Settings tab of NPSP settings, if you edit and then attempt to save, 
AND Enable Automatic Verification is FALSE
AND Verification Service is NULL
then the settings are not saved and a "Settings not saved: Attempt to de-reference a null object" error is shown under the Save/Cancel buttons.

Selecting Google or Cicero while leaving the Enable Automatic Verification = FALSE allows a save, even with a NULL Auth Token. Trying to save with SmartyStreets selected gives verification errors on Auth ID and Auth Token. Note also that selecting SmartyStreets doesn't cause the Auth ID field to appear until a save is attempted.

njjc:
# Warning
# Info
# Issues
Fixes #1491.
Fixes #1488.
Fixes #1476.

### DIFF
--- a/src/classes/ADDR_GoogleGeoAPI_Validator.cls
+++ b/src/classes/ADDR_GoogleGeoAPI_Validator.cls
@@ -38,13 +38,15 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
     private String END_POINT;
     private Double TIMEOUT;
     private List<Address__c> checked_addresses;
+
+    private Addr_Verification_Settings__c settings;
     
     /*******************************************************************************************************
     * @description Provides the called with the service name.
     * @return String The name of the address validation service.
     */
     public String getServiceName() {
-    	return 'The Google Geocoding API';
+        return 'The Google Geocoding API';
     }
     
     /*******************************************************************************************************
@@ -52,7 +54,7 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
     * @return String The URL to send requests to.
     */
     public String getDefaultURL() {
-    	return 'https://maps.googleapis.com/maps/api/geocode/json?';
+        return 'https://maps.googleapis.com/maps/api/geocode/json?';
     }
     
     /*******************************************************************************************************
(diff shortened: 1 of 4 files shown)


njjc:
@ceiroa Ready for review. I recommend using the [w=1](https://github.com/SalesforceFoundation/Cumulus/pull/1502/files?w=1) parameter to review without the whitespace changes.

ceiroa:
So weird, it seems like I cannot create inline comments in the pull request. I was just going to point out that in line 220 of ADDR_GoogleGeoAPI_Validator you check for the result size, but in line 168 you don't. I was wondering if it matters.

src/classes/ADDR_GoogleGeoAPI_Validator.cls
```diff
@@ -161,24 +165,24 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
 
                 String verified_street = '';
                 String verified_premise = '';
-                if (googresp.status == 'OK' && googresp.results.size() == 1){
+                if (googresp.status == 'OK' && googresp.results.size() == 1 && !(settings.Reject_Ambiguous_Addresses__c && googresp.results[0].partial_match)){
```
njjc: I think inline comments only work if you don't have w=1 set.

Line 168 is all ANDs, so the googresp.results.size() has to be 1.

njjc: @ceiroa

PR #1503, opened 2015-04-15
Avoid attempting email sends in unit tests

### LINKED ISSUE #1486: Avoid attempting email sends in unit tests
Often, when running NPSP tests in a sandbox, I get errs because usually non-system email is turned off in sandbox, and you guys have tests that attempt to send it.  IMHO it's best to not attempt email sends in tests.  It never sends, but it does still throw an err if it's disabled.

Example from today:

Apex Test Result Detail

```
Time Started    3/27/2015 5:29 PM
Class   ADDR_Validator_TEST
Method Name test400
Pass/Fail   Fail
Error Message   System.EmailException: SendEmail failed. First exception on row 0; first error: NO_MASS_MAIL_PERMISSION, Single email is not enabled for your organization or profile. Single email must be enabled for you to use this feature.: []
Stack Trace (npsp)
```

Looking at that test method, it doesn't do any email sends itself.  Guessing it's some error handling routine that's doing it.  Anyway, I would humbly recommend looking for sendEmail() methods, and wrapping them in

```
 if (!Test.isRunningTest())
```

its not that this err is such a big deal, many folks will understand why it's happening.  but it does mean that when we're debugging it's one more thing we have to look at, and could conceivably be obfuscating a real error.

Thanks!

mrbelvedere:
I made the edits that @msholtz suggested and the tests run fine now without any failures even with the email delivery settings on "System Emails Only".

# Critical Changes

No critical changes. 

# Changes

No changes necessary. 

# Issues

Fixes #1486

### DIFF
--- a/src/classes/ADDR_Validator_TEST.cls
+++ b/src/classes/ADDR_Validator_TEST.cls
@@ -127,119 +127,127 @@ public with sharing class ADDR_Validator_TEST {
     }
     
     static testMethod void test400() {
-        if (strTestOnly != '*' && strTestOnly != 'test400') return;
-        
-        Addr_Verification_Settings__c settings = createDefaultSettings();
-        Test.setMock(HttpCalloutMock.class, new ADDR_MockHttpRespGenerator_TEST());
-              
-        Contact contact = new contact(firstname='Test', lastname='Person'); 
-        contact.MailingStreet = '400 error';
-        contact.MailingCity = 'chicago';
-        contact.MailingState = 'Illinois';
-        contact.MailingPostalCode = '60606';
-        contact.MailingCountry = 'United States';
-        
-        Test.StartTest();
-        insert contact;
-        ADDR_Addresses_TDTM.hasRunAddrTrigger = false;
-        Test.StopTest();
-        
-        //check household got created for contact
-        contact = [select Account.Id from Contact where Id = :contact.Id];
-        Id accountId = contact.Account.Id;
-        System.assertNotEquals(null, accountId);
-
-        Address__c[] av = [select Verified__c, Verification_Status__c from Address__c where Household_Account__c = :contact.AccountId];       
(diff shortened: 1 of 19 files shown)


ceiroa:
I think only the line that actually attempt to send the email should be wrapped with the !Test.isRunningTest() clause. The condition could be added to https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/ERR_Notifier.cls#L127.

Also, it seems like a number of classes has been added to the package. I believe these are classes we had at some point, but were deleted from the package.

src/classes/ADDR_Validator_TEST.cls
```diff
@@ -127,119 +127,127 @@ public with sharing class ADDR_Validator_TEST {
...
-        //check household got created for contact
-        contact = [select Account.Id from Contact where Id = :contact.Id];
-        Id accountId = contact.Account.Id;
-        System.assertNotEquals(null, accountId);
-
-        Address__c[] av = [select Verified__c, Verification_Status__c from Address__c where Household_Account__c = :contact.AccountId];       
-        System.assertEquals(1, av.size());
-        System.assertEquals(false, av[0].Verified__c);
-        System.assertEquals('Bad input. Required fields missing from input or are malformed.', av[0].Verification_Status__c);
+    	if (!Test.isRunningTest()) {
```
njjc: Code inside testMethods only runs in a test context. Wrapping the entire test contents in !Test.isRunningTest() causes these test methods to never run.

mscholtz:
Yes, what @ceiroa suggested is what I was suggesting, not what I see in the commits above.  And FWIW, I'm still seeing these errs today in a client org, so looks like this hasn't made it into production yet?  Thanks!

ceiroa:
@mscholtz ~ Nope, this pull request was not accepted. The original issue is still open.

PR #1505, opened 2015-04-17
Feature/1406 ids on error msg

### LINKED ISSUE #1406: As an admin, I'd like error messages to include the user and org ids
For those admins that manage more than one org, it is useful to know which org the error is coming from.

We can have a message similar to the one Salesforce adds at the beginning of an error message:

Apex script unhandled exception by user/organization: xxxxxxxxxxxxxxxx/xxxxxxxxxxxxx

ceiroa:
# Warning
# Info
# Issues
Fixes #1406.

### DIFF
--- a/ApexDocContent/Allocations.htm
+++ b/ApexDocContent/Allocations.htm
@@ -7,8 +7,7 @@
 			<li>Manage Allocations page</li>
 			<li>Allocations rollups</li>
 			<li>Make Default Allocations feature</li>
-			<li>Multi currency support
-			<li>
+			<li>Multi currency support<li>
 		</ul>
 	
 		<p>The classes involved with Allocations trigger support include:</p>
--- a/ApexDocContent/Allocations.htm
+++ b/ApexDocContent/Allocations.htm
@@ -7,7 +7,7 @@
 			<li>Manage Allocations page</li>
 			<li>Allocations rollups</li>
 			<li>Make Default Allocations feature</li>
-			<li>Multi currency support<li>
+			<li>Multi currency support</li>
 		</ul>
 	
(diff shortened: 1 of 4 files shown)


ceiroa:
@njjc ~ ready for review.

ApexDocContent/Allocations.htm
```diff
@@ -7,8 +7,7 @@
 			<li>Manage Allocations page</li>
 			<li>Allocations rollups</li>
 			<li>Make Default Allocations feature</li>
-			<li>Multi currency support
-			<li>
+			<li>Multi currency support<li>
```
njjc: The closing tag isn't /closing

src/classes/ERR_Notifier.cls
```diff
@@ -146,7 +146,8 @@ public with sharing class ERR_Notifier {
         mail.setReplyTo('donotreply@salesforcefoundation.org');
         mail.setSenderDisplayName('Nonprofit Starter Pack');
         mail.setSubject('NPSP Error');
-        string bodyString = '';
+        string bodyString = 'Organization: ' + UserInfo.getOrganizationName() + '(' + UserInfo.getOrganizationId() + ')';
+        bodyString += 'User: ' + UserInfo.getUserName() + '(' + UserInfo.getUserId() + ')';
```
njjc: Do you have an example of what this looks like? Is this on a single line with no space between the ) and User:?

njjc: There are \n later on, we might need one here.

ceiroa:
Ok, I added the / for the closing tag, and a new line after the one with org info. Let me know how this looks.

PR #1518, opened 2015-04-22
Changing opp naming tests to not rely on an active opportunity record ty...

njjc:
...pe for the current profile.


njjc:
Tested this with and without active profile record types, works in both cases.

njjc:
@davidhabib ready for review

PR #1527, opened 2015-04-30
to fix a timing issue in a test

davidhabib:


davidhabib:
@njjc passes build so ready for you!

PR #1528, opened 2015-04-30
contact merge was incorrectly renaming one2one accounts.

davidhabib:


davidhabib:
@njjc THis is ready for review.  fixes the one random test error with addresses (by removing assert).  also found and fixed a bug where one2ones got renamed incorrectly.  thanks.

PR #1529, opened 2015-05-04
Feature/634 opportunity naming

### LINKED ISSUE #634: As a user, I'd like to be able to configure the naming of Opportunities being created as part of Recurring Donation so I can conform to whatever standards my organization uses.
Currently the name of an Opportunity generated from a recurring donation is generated within the code, and is rather inflexible.
-----
***NOTE: We may want to consider a complete overhaul of Opportunity naming, maybe in a similar fashion as we did for Household naming***
-----
The code currently concatenates the Donor name, a Custom label normally set to "Donation", the index of the donation and the date.
https://github.com/SalesforceFoundation/Cumulus/blob/2537e769f22c71951ccf64759b104e0b1bc57d09/src/classes/RD_RecurringDonations.cls#L401

This setup is fairly inflexible
- The format of the name cannot be changed (e.g. brackets around number etc.)
- The data in the name cannot be re-ordered
- The format of the name cannot be varied between different recurring donations

I would suggest that the name of the Opportunity be derived from the name of the Recurring Donation. It could be implemented such as:

```
<Recurring Donation Name> (X) DD/MM/YYYY
```

The format could be described by a custom label using apex's string.format()

```
{0} ({1}) {2}
```

This would mean that it could be varied by the users by naming the recurring donation appropriately. Administrators can also define the merge order / brackets around index etc.

Alternatively more complex formatting could be used on an org level basis configurable by custom labels / custom settings, or on a per record basis which may require a wizard to help users populate the template field.

njjc:
# Warning
# Info
# Issues
Fixes #634

### DIFF
--- a/src/classes/ACCT_AccountMerge_TDTM.cls
+++ b/src/classes/ACCT_AccountMerge_TDTM.cls
@@ -1,0 +1,119 @@
+/*
+    Copyright (c) 2015 Salesforce.com Foundation
+    All rights reserved.
+    
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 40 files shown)


njjc:
@davidhabib ready for review, thanks!

PR #1530, opened 2015-05-05
Adding opp name setting validation, fixed formula name to correctly key off of open ended status.

njjc:
# Warning
# Info
- Opportunity Naming is now out of beta! If you already created opportunity name settings, you'll have to edit each setting, select the correct "Attribution" setting, and save the setting.

# Issues


njjc:
@davidhabib Ready for review, added validation and a test.

PR #1532, opened 2015-05-05
Feature/1412 addr verification status

### LINKED ISSUE #1475: As a developer, add State Upper and State Lower to the Address object, and have the Cicero validator populate them.
Request from Cicero, who have received these requests from our customers.

We have had good leads and free trial sign ups from new Salesforce users. A few questions we've had: Is it possible on your end to add other districts besides congressional to Salesforece using Cicero. Meaning, would it be possible for you to expose those results so users could consume them? They are already paying for the API call which provides this info, so it might be useful for them to include. 

About additional fields to return: We provide additional districts beyond congressional in our API response. This documentation page http://cicero.azavea.com/docs/district_type.html#cicero.webservice.models.DistrictType defines each district type we can provide. We thought the following would be most useful in addition to congressional districts in Salesforce: State Lower, State Upper, Local (though Local refers to local council districts for only the 100 cities which we track). Please disregard the _2010 districts.  Those were a hold over during redistricting process to capture new districts before the came into effect, all main district types are populated with the current and active districts.

My response to Cicero (Sarah Cordivano):
I think adding State Upper and State Lower to our base address object should be fine.  I'm less included to include Local, since I don't even know how to really explain it.  (When I look at Local for my sample address, it get several LOCAL records with subtype COUNTY, which I will add to our code to map into the Address's County field).  I'll add this work item to our list, and it should get worked upon in 2-6 weeks.

### LINKED ISSUE #1501: As a user, I'd like the lat and long address info to be pushed from the address record to the Account.BillingLatitude  and Account.BillingLongitude
These fields may have been added together with the new geolocation-related functionality included in Spring 15. 

Looks like you can create Account/Contact list views based on geolocation (e.g. within X miles of ____" but that requires the geocoding to be included on the Account/Contact record.

njjc:
# Warning
# Info
- Cicero address verification now supports State Upper and State Lower districts.
- Address verification that provides geolocation now fills in geolocation coordinates for Accounts and Contacts that use that address.

# Issues
Fixes #1475 
Fixes #1501


njjc:
@davidhabib Ready for review. Thanks!

njjc:
@davidhabib ready for review.

PR #1535, opened 2015-05-05
Moving batch kickoff out of try/catch, moving error notification email into install script class due to sharing privs.

njjc:


src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -40,7 +40,7 @@ public class RD_RecurringDonations_TEST {
     * @description if you only want to run one test in this class, fill in its name here.
     * if you want to run all tests, then use '*'
     */
-    private static string strTestOnly = '*';    
+    private static string strTestOnly = 'testInstallmentNumbers';
```
davidhabib: dont' you want this back to *

PR #1556, opened 2015-05-14
Feature/685 lead convert

### LINKED ISSUE #685: Bug: Lead Conversion doesn't create household record (unless being merged into an existing contact).
Daniel Trimble (KIDANGO INC)
 
Bug with #NPSP Lead Conversion?

The NPSF lead conversion intervention process worked pretty well, but there seems to be a glaring bug. Am I the only one who's experiencing this?

1. Created a new lead. "George Washington", with a Company of "Self"
2. From lead record, clicked Convert
-----> NPSF conversion process, specified to create a new contact, not create an opportunity, and not email the owner
-----> NPSF converted the lead to a contact with his own individual account. It did NOT create a household.
3. From lead record, created a new lead. "Mr. George Washington", with a Company of "Self"
4. From lead record, clicked Convert
-----> NPSF conversion process, specified to merge with the existing George Washington contact, not create an opportunity, and not email the owner
-----> NPSF merged the lead data correctly. And this time it DID create the household

It seems the lead conversion process only creates the household record if it's merging the lead into an existing contact. Otherwise, I have to manually create the household.

### LINKED ISSUE #965: Lead Conversion page uses "Opportunity" as label instead of "Donation"
The use of "Opportunity" instead of "Donation" might be confusing to some users who are not familiar with standard Salesforce architecture. I recommend using the object label {!$ObjectType.Opportunity.Label} on the VF page. This same is true on the Lead Settings page.

https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/pages/LD_LeadConvertOverride.page

### LINKED ISSUE #1471: BUG: Lead Convert button doesn't allow for organization account matching anymore
Hub post: https://powerofus.force.com/0D58000001xrX3K

I confirmed in an instance running NPSP 2 with Convert overrride. If a company is entered, the standard lead conversion comes up. If Self is entered the override comes up.

In NPSP 3 org, the override comes up regardless of what is in the Company name field (happens with both the npe01 version and the npsp version of the VF page).

However, upon conversion the Account is created as an organization account. But it's not giving the user the option to match to an existing Account as a standard lead conversion would do. I know there's plans for enhancement around Lead conversion, but this is a bit of a regression from NPSP 2.

### LINKED ISSUE #1554: Address Validation Error during Contact Merge
user reporting that they are hitting address validation errors during contact merge even though address validation is off.  https://powerofus.force.com/0D580000027Fxdu

Sounds like there is two issues:

1. if validation is off, this should not happen.

2. if validation is on, contact merge will be doing its fixups in a future, and thus cannot call validation.

davidhabib:
# Warning
# Information
* NPSP's Lead Convert page has been enhanced to support specifying an Account model Account and Affiliation, when the lead has an Organization account in its Company field.  This gives more flexibility to convert leads with companies, into the account model you want.

# Issues
Fixes #1471 
Fixes #965 
Fixes #685 
Fixes #1554

### DIFF
--- a/src/classes/ADDR_Validator_TDTM.cls
+++ b/src/classes/ADDR_Validator_TDTM.cls
@@ -73,8 +73,10 @@ public with sharing class ADDR_Validator_TDTM extends TDTM_Runnable {
                     ADDR_Validator.verifyRecordsFuture(new Map<Id, SObject>(newList).keyset());
                     alreadyRun = true;
                 } else if(system.isFuture()) {
+                    UTIL_Debug.debug('****ADDR_Validator_TDTM called from future, addresses not verified.');
+                    /*We don't want to log an error here. Instead, we'll just do a debug statement.
                 	ERR_Handler.processError(new ADDR_Validator_Exception('ADDR_Validator_TDTM called from future'), 
-                	                           ERR_Handler_API.Context.ADDR);
+                	                           ERR_Handler_API.Context.ADDR);*/
                 } else if(system.isBatch()) {
                 	UTIL_Debug.debug('****ADDR_Validator_TDTM called from batch, addresses not verified.');
                     /*We don't want to log an error here. Instead, we'll just do a debug statement.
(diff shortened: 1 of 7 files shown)


davidhabib:
@njjc ready for review and merge.  let me know if you want to go through it together.  thanks!

PR #1563, opened 2015-05-19
Feature/685 lead convert

davidhabib:
@njjc just renaming the two lead convert labels.


davidhabib:
@jlantz when do you plan on cutting the release?  this simple two label fix needs to be included, but won't make it thru the build for another hour or so it looks like.  shall we just merge it now?

davidhabib:
@jlantz this build keeps breaking, but it has nothing to do with the 2 label changes!

PR #1565, opened 2015-05-19
Feature/1412 addr verification status

njjc:


njjc:
@davidhabib ready for review

davidhabib:
not fair!  I put my pull request in 2 hours before you, but the build keeps breaking on it!  even though all it is two label changes!

PR #1573, opened 2015-05-21
#1531. Handle deleting Address objects, and updating the contact and …

### LINKED ISSUE #1531: Primary address deletion behavior is not handled.
It appears that when deleting a primary address directly from the related list, the address is still retained as the Billing Address on the Account post-deletion, even though the address record itself has been removed. 

Consider one of the following deletion behaviors: 
1.  Select the most recently available previous address (if one exists) from the list of addresses on the Account
2.  Remove the address values from the Shipping address fields and any associated Contacts

davidhabib:
# Warning
# Information
# Issues
Fixes #1531

### DIFF
--- a/src/classes/ADDR_Addresses_TDTM.cls
+++ b/src/classes/ADDR_Addresses_TDTM.cls
@@ -97,70 +97,70 @@ public with sharing class ADDR_Addresses_TDTM extends TDTM_Runnable {
             }
         } else {  
 
-        integer i = 0;        
-        for (SObject so : listNew) {
-            Address__c addr = (Address__c)so;
-            Address__c addrOld = (listOld != null) ? (Address__c)listOld[i] : null;
-            
-            // BEFORE INSERT
-	        if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
-	           // when an address is marked Default, we update its latest date fields
-	           if (addr.Default_Address__c) { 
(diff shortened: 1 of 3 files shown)


davidhabib:
@njjc ready for your review and merge.

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -234,6 +243,13 @@ public with sharing class ADDR_Addresses_TDTM extends TDTM_Runnable {
 	                }      
 	            }
             }
+            
+            // finally, if no default found, create an empty default
```
njjc: Why are we creating an empty default?

davidhabib: to support the scenario where the user is deleting the last address object.
 this will cause the account and contact address fields to get cleared,
which was reported as a bug (that they didn't get cleared).


njjc: Wouldn't it be better to clear them another way, so that we don't have empty address objects hanging around eating up data storage?

davidhabib: it's just in memory.  it never gets saved.

njjc: Ahhh.

PR #1576, opened 2015-05-26
#1443. create Account address object on Insert

### LINKED ISSUE #1443: When manually converting to the new Household Account model, addresses don't populate Address object
One of our clients wasn't using the Household object; so, the conversion utility for the new Household Account model didn't work. When we switched to the new Household Account model and manually created the new Household Accounts, the address on the contact didn't populate the Address object. Here's how we recreated the issue:

- Starting NPSP Settings:
  - Account Model = One-to-One
  - Household Account Record Type = Individual

- Created test contact with address and verified One-to-One Account created.

- Changed NPSP Settings:
  - Account Model = Household Account
  - Household Account Record Type = Household Account

- Ran Jitterbit query on the Contact object to extract:
  - AccountID
  - ID
- Cleared AccountID column in CSV & used Jitterbit to update the Contact with the null value in the AccountID field.
- This successfully created a new Household Account for the test Contact. The Contact's address was added to the new Household Account, but not to the Address object on the Account.

davidhabib:
# warning
# information
# issues
Fixes #1443

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -547,24 +547,23 @@ public with sharing class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
         }
         
         if (accountInserts.size() > 0) {
-            // DJH: with current TDTM design, we can't use dmlWrapper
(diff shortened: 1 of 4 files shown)


davidhabib:
@njjc  build is running.  ready for review.

PR #1599, opened 2015-06-09
#1591. remove now unnecessary soql that hit limits with over 100k opps during Contact merge

### LINKED ISSUE #1591: Contact Merge query can hit limits if org has > 100k Opportunities
Cascase Aids Alliance hits a runtime soql error when trying to merge contacts.  Here is the hub post: https://powerofus.force.com/0D580000028RIkb

the problem is due to the following query in CON_ContactMerge_TDTM.run()

list<Opportunity> listOpp = [select Id, npe01__Contact_Id_For_Role__c from Opportunity 
                where npe01__Contact_Id_For_Role__c in :listConId];

Unfortunately, npe01__Contact_Id_For_Role__c is just a text field and not a real lookup, so it doesn't get indexed.  not sure what our workaround is.

davidhabib:
# Warning

# Info

# Issues
Fixes #1591


davidhabib:
@njjc build is running.  ready for review.  thanks.

PR #1614, opened 2015-06-22
Affiliation and Account Converter fixes

### LINKED ISSUE #1557: If the account model is set to One to One, Organization Affiliation record not created based on Primary Affiliation
Replicated this in 3.38 org - POU thread https://powerofus.force.com/0D580000027Fx0n

### LINKED ISSUE #1590: Account Converter should copy contact's primary address type to address objects and not make work address the default
hub thread here:  https://powerofus.force.com/0D580000027H65N

relevant part:

Caroline Renard (Technology Interpreter for Arts Groups)
It would be useful if the Primary Address Type on Contact was applied during conversion. Have just converted an org with several contacts that had been directly related to organizations (so the address on their Contact record was "Work") and the HH address was blank. Now that they are converted, in the new Household Accounts, the same address appears as "Home" and Default in the Address object. I've used Apsona to update the Address Type on the Address object back to "Work" and to uncheck the Default, but it would be nice not to have to do this bit of post-processing - and it would have completely confused my client if they had noticed (or NOT noticed!) before I did. Thanks!

davidhabib:
# Warning

# Info

# Issues
Fixes #1557 
Fixes #1590


davidhabib:
@njjc ready for review.  build is running.

Issue #1584, opened 2015-05-29
System.QueryException on contact merge


AshwinPandole:
@njjc 
Hi Nicolas,

Getting below exception while doing contact merge

![image](https://cloud.githubusercontent.com/assets/3069322/7877119/73569048-05f3-11e5-969a-611be16404b5.png)

below is the log while exception was inserted.

![image](https://cloud.githubusercontent.com/assets/3069322/7877133/953a40d8-05f3-11e5-85ec-d7a37ea23d01.png)

can you please assist.

judisohn:
Hi @AshwinPandole - it would be great if you could start a post on the Power of Us Hub in the NPSP group detailing a little bit more about your instance and situation. Do you have additional custom code or other apps? What version of NPSP? Account model? How many opportunities do these contacts have? Etc.

Our typical process is to collaborate with you and the community to troubleshoot errors in the Hub. Then once determined that it is indeed a general product bug/issue and the problem is identified in a reproducible and actionable way, it's updated here. 

Thanks!!

davidhabib:
same as #1591?

davidhabib:
fixed in NPSP version 3.41 which will be released to sandboxes hopefully on 6/10.

PR #1616, opened 2015-06-22
Feature/1550 primary donor

### LINKED ISSUE #1564: BUG: Relationships - set to List Setting - when Status changes to Former from Current, Type is set to Neutral value of row
Hub thread: https://powerofus.force.com/0D580000027HLFb

Replicated in test org. When changing status, on Save the Type is updated to the row's Neutral value.

### LINKED ISSUE #1615: Fix Opportunity Naming to use OCR and less system resources
We can't assume everyone has Primary Contact populated as we made it also.
Also only do one opportunity update in the opp after insert trigger to save system resources.

njjc:
# Warning
# Info
# Issues
Fixes #1615


src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -190,7 +211,7 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
 
         //Update existing CRs.
         if (!CRUpdates.isEmpty()) {
-            dmlWrapper.objectsToUpdate.addAll((List<SObject>)CRUpdates);
+            update CRUpdates;
```
njjc: Had to do these directly so they exist for opportunity naming code.

src/classes/OPP_OpportunityNaming.cls
```diff
@@ -306,11 +278,28 @@ public with sharing class OPP_OpportunityNaming implements OPP_INaming {
         for (string field : getConFields(fields)) {
             result += field + ', ';
         }
-        result = result.substring(0, result.length()-2) + ' FROM OpportunityContactRole';
+        result = result.substring(0, result.length()-2) + ' FROM OpportunityContactRoles';
```
davidhabib: isn't the object called OpportunityContactRole?

njjc: This was a query and is now a subquery, the subquery relationship name has an s at the end.

src/classes/REL_Relationships_TDTM.cls
```diff
@@ -218,8 +210,7 @@ public with sharing class REL_Relationships_TDTM extends TDTM_Runnable {
         for(npe4__Relationship__c r : Relationships) {
         	
             //SWA 2009-05-13 removed a duplicate check for change to npe4__Status__c from this OR statement
-            if ((r.npe4__Status__c != oldRelationships[counter].npe4__Status__c 
-            || r.npe4__Type__c != oldRelationships[counter].npe4__Type__c 
+            if ((r.npe4__Type__c != oldRelationships[counter].npe4__Type__c
```
davidhabib: why do we no longer need to compare Status changes?

njjc: https://github.com/SalesforceFoundation/Cumulus/issues/1564

davidhabib: but does this mean making one former will not update the reciprocal to former?  I thought #1584 was about us messing up the gender rather than leaving it alone.

njjc: Yeah, it looks like you're right. I'm not sure how to fix it then.

davidhabib: well, either back out this fix and leave the bug open, or we can try to do the correct fix.  I looked a little at the code, and I don't fully understand how the rels originally had gender specific roles, but on re-evaluation, they get reset to neutral.  is it only one of the contacts has gender specified, and the rels were originally started from that gendered contact, but the update to former is happening on the ungendered contact's rel?

njjc: I think it's when you manually specify the correct type on creation and haven't worried about specifying genders on the contacts. Recalculating it always sets it as neutral. Not sure how we can understand which case we're in, though.

davidhabib: what if we only updated type on the reciprocal rel, if type was changed on the first rel.  ie, if the rel being updated hasn't changed its type, then we avoid calling the type routine.

PR #1622, opened 2015-06-24
Fixing #1621

### LINKED ISSUE #1621: New Opportunity for household's non primary contact: record is read only
https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe&fId=0D58000002AFily&fromEmail=1

njjc:
# Warning
# Info
# Issues
Fixes #1621

### DIFF
--- a/src/classes/OPP_OpportunityContactRoles_TDTM.cls
+++ b/src/classes/OPP_OpportunityContactRoles_TDTM.cls
@@ -54,7 +54,7 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
             for (sObject dmlObj : dmlWrapper.objectsToUpdate) {
                 id objId = (id) dmlObj.get('id');
                 if (mapOppNaming.containsKey(objId)) {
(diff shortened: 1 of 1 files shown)


src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -188,10 +188,6 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
                 //check if our contact lookup field is incorrect and update it
                 if (thisOpp.Primary_Contact__c != cr.ContactId) {
                     
-                    //update existing opp for OPP_OpportunityNaming call
```
davidhabib: why did you add this last pull request, and why now must it go?  I was assuming it was needed to keep thisOpp in sync.

njjc: It was from before I rewrote the naming to just use contact roles. It didn't actually work and we have no way of testing the way the "New Contact Donation" button creates contact roles except manually. I made a story to make a selenium test for this button since we can't test it in apex.

PR #1631, opened 2015-06-29
Feature/email thank yous

davidhabib:
# Warning

# Info
- NPSP now includes a feature to send Acknowledgment Emails to donors thanking them for their donation.  Full details on configuring the feature, and using the feature, will be documented on the Power of Us Hub.

# Issues


davidhabib:
build system doesn't currently handle the email templates in the unfiled public email templates folder, so do not merge until we resolve this.

PR #1644, opened 2015-07-06
Feature/1479 sharing

### LINKED ISSUE #1479: remove "with sharing" from core classes that don't need it (AFTER security review is complete)
There are many classes in NPSP that have "with sharing" declaration that IMHO don't warrant it, and I humbly suggest that you review where this is really needed (VF controllers, etc)  and where it's not (most trigger code IMHO).  

This has caused me several problems with writing code that will work w/ NPSP3.  In every case this is code that worked fine w/ NPSP2, and we made expected adjustments, but hard-to-debug permissions errors have gotten in the way (hard to debug because occurring on managed NPSP code).

One situation was posted on the PUB here (although discussion petered out because I didn't have it diagnosed quite properly): https://powerofus.force.com/0D58000001PuE6G

Another that just arose was this:

* client has Sharing Settings on opp set to Read Only
* we have a custom Before Insert trigger on opp that resets the owner
* an integration (IATS Brickwork) inserts opps from donation forms via a Sites guest user account
* worked fine in NPSP2
* in NPSP3, was throwing a "insufficient access rights on cross-reference ID" error
* through copious testing, was able to figure out that the error was happening here: https://github.com/SalesforceFoundation/Cumulus/blob/2d8fc54af04102c9aa84b68380eedd5587d6b95e/src/classes/TDTM_TriggerHandler.cls#L149
* my understanding is that after the before trigger changes the owner of the opp, the Guest User account no longer has permissions to edit it.  when your trigger fires and calls this class (declared as "with sharing") a DML of some sort that touches opps is no longer allowed and throws this err
* If I comment out the resetting of Owner, or if I change the Sharing Settings to Read/Write, the error doesn't occur

My understanding of the TDTM_TriggerHandler class is limited, but it doesn't seem like the sort of thing that needs to be "with sharing".  Most triggers should just run when the conditions are met, regardless of which user put the data in.

Furthermore, it leads to the kind of errors and difficult debugging that I've been dealing with today.  I think you don't want your users to have to assign additional permissions to user accounts simply to prevent core NPSP  triggers from throwing errors.

Thanks! 
@davidhabib @ceiroa

njjc:
# Warning
# Info
# Issues
Fixes #1479


src/classes/PMT_PaymentCreator.cls
```diff
@@ -32,7 +32,7 @@
 * @date 2011 (2.1)
 * @description Controls the creation of payment objects from Opportunity trigger
 */
-public without sharing class PMT_PaymentCreator {
+public with sharing class PMT_PaymentCreator {
```
davidhabib: since this class is used by the trigger, shouldn't it be empty to inherit the trigger's without sharing?

njjc: We decided it didn't need it as it only deals with payments which are
detail records of opportunities and inherit the opp's sharing rules.

mscholtz:
@davidhabib @njjc 
Happy to see this going in!  Any ETA on when this will be released?  Did it already go pass security review?

njjc:
Should make it into a sandbox push today. Still in security review, but explained these away.

PR #1661, opened 2015-07-22
Store Geolocation information correctly

cdcarter:
Google connector is currently storing the latitude and in both the lat and long fields.


cdcarter:
Hey folks, hope your off-site is going well. I can probably supply a failing test class tomorrow, but I'm having some struggles getting a real dev environment set up. As you can see, this code is pretty simple and the situation currently isn't tested at all.

kbromer:
Looks good to me, thx for the fix!

PR #1700, opened 2015-09-02
fixes #1698 by first gathering related Contact Ids and using them as …

### LINKED ISSUE #1698: Delete of Contact that has Relationships causes non-selective query exception when Relationships table has 100k rows
Reproduction: 
In an org with 100k rows in the npe4__Relationship__c table, delete a Contact that has a Relationship.

Looks to be caused by [this query](https://github.com/SalesforceFoundation/Cumulus/blob/bb7c84b5739ad52d3a0f34d3068a103b19caa733/src/classes/REL_Relationships_Con_TDTM.cls#L143) in **REL_Relationships_Con_TDTM.cls**:  `[Select Id from npe4__Relationship__c where npe4__RelatedContact__c = null]`

Error Message:
System.QueryException - Non-selective query against large object type (more than 100000 rows). Consider an indexed filter or contact salesforce.com about custom indexing.
Even if a field is indexed a filter might still not be selective when:
1. The filter value includes null (for instance binding with a list that contains null)
2. Data skew exists whereby the number of matching rows is very large (for instance, filtering for a particular foreign key value that occurs many times)

kenslewis:
…a selective filter when querying for Relationships to delete

# Warning
# Info
# Issues
Fixed #1698

### DIFF
--- a/src/classes/REL_Relationships_Con_TDTM.cls
+++ b/src/classes/REL_Relationships_Con_TDTM.cls
@@ -135,12 +135,13 @@ public class REL_Relationships_Con_TDTM extends TDTM_Runnable {
     
     /*******************************************************************************************************
     * @description Deletes Relationships when a Contact is deleted
+    * @param contacts the Map of Contacts from trigger oldMap. 
     * @return dmlWrapper Relationships to delete.
     ********************************************************************************************************/
     public static DmlWrapper deleteEmptyRelationships(Map<Id, Contact> contacts) {
         DmlWrapper dmlWrapper = new DmlWrapper();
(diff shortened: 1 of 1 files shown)


davidhabib:
Thanks @kenslewis for investigating this issue and coming up with the fix.  It might not be until after Dreamforce that we are able to get this merged in.  We are in the middle of a sprint, so we are currently busy.  One of us will need to create a feature branch from your pull request, add some appropriate test code, and then get it merged in.  Thanks again!

njjc:
Already fixed and merged in #1703

PR #1721, opened 2015-09-24
Feature/data import fixes

### LINKED ISSUE #1672: Unpredictable error reporting in NPSP Data Import
When Data Import encounters an error from Salesforce in updating Account after it's created (such as after Address record is inserted due to a field that the Account address fields can't accept) the error is reported as a failure on the first record in the batch, rather than on the record where the error is actually located.

davidhabib:
# Warning

# Info

# Issues
Fixes #1672


davidhabib:
@njjc build passes!  ready for review and merge.  thanks!

PR #1736, opened 2015-10-05
RD Custom Field Mapping fix

### LINKED ISSUE #1697: Recurring Donation - Refresh Opportunities not updating with new values
If the Recurring Donation object is updated (i.e. Amount, Campaign, Custom Fields), the Refresh Opportunities button doesn't update the mapped fields in the open child Opportunities. Likewise if the instalment period is changed (for example from monthly to quarterly), the previously created monthly Opportunities still exist.

Ideally when the Refresh Opportunities button is selected, all currently open child Opportunities are deleted. Then when the action method to refresh the RD is run, new child Opportunities (with the correct schedule and field mappings) will be created in line with the current RD.

davidhabib:
# Warning

# Info

# Issues
Fixes #1697


davidhabib:
@njjc OK, it's finally building, so ready for review and merge once it passes!  thanks.

PR #1742, opened 2015-10-08
W-008087. Health Check now gives warnings if Trigger Handlers don't match NPSP defaults

davidhabib:
# Warning

# Info
* Health Check now gives warnings if Trigger Handlers don't match NPSP defaults
* Trigger Handler tab in NPSP Settings gives warning if any NPSP Trigger Handlers are marked Asynchronous.

# Fixes


davidhabib:
@njjc build passes.  ready for review and merge.  thanks.

PR #1751, opened 2015-10-15
Feature/async tdtm work

davidhabib:
This pull requests includes testing of all of the ADDR related TDTM trigger handler classes as async.

While most of it is just testing infrastructure changes and test fixes, there are a few real fixes in TDTM core code.

### DIFF
--- a/src/classes/ADDR_Account_TDTM.cls
+++ b/src/classes/ADDR_Account_TDTM.cls
@@ -32,6 +32,7 @@
 * @date 2014
 * @group Address Management
 * @description Trigger Handler on Account for Address management 
+* @async_support this trigger handler has been tested with async=true, and appears to not cause problems.
 */
 public class ADDR_Account_TDTM extends TDTM_Runnable {
 
--- a/src/classes/ADDR_Validator_TDTM.cls
+++ b/src/classes/ADDR_Validator_TDTM.cls
@@ -32,6 +32,8 @@
 * @date 2014
 * @description Class that gets called when Address__c records change and
 * begins the address verification/normalization.
+* @async_support this trigger handler DOES NOT SUPPORT ASYNC!  the validation call is a future,
+* and if we are already in a future to deal with the trigger handler, we can't call the future verification api!
 */
 public class ADDR_Validator_TDTM extends TDTM_Runnable {
 	
@@ -58,7 +60,8 @@ public class ADDR_Validator_TDTM extends TDTM_Runnable {
         
         if(!alreadyRun) {
 	        Addr_Verification_Settings__c settings = UTIL_CustomSettingsFacade.getAddressVerificationSettings();
-	        if(trigger.isBefore && settings.enable_automatic_verification__c) {        	
+	        if ((triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) &&
+	           settings.enable_automatic_verification__c) {        	
 	        	for(SObject so : newList) {
 	        		Address__c address = (Address__c)so;
 	        		if(!system.isFuture() && !system.isBatch()) {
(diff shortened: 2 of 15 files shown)


davidhabib:
@ceiroa @njjc this passes build and is ready for review.  I suggest Carlos review it, since he should see the changes done in TDTM, since he will want to migrate these to HEDAP.  thanks!

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -181,15 +182,19 @@ public with sharing class TDTM_TriggerHandler {
     
     private void runAsync(TDTM_Runnable classToRun, String classToRunName, List<Sobject> newList, 
     List<Sobject> oldList, TDTM_Runnable.Action thisAction, Schema.DescribeSobjectResult describeObj) {
-       Map<Id,Sobject> nm;
-       if(newlist != null) nm = new Map<Id,Sobject>(newlist);
-       else nm = new Map<Id,Sobject>();
+       set<Id> setNewId;
```
ceiroa: The only difference that I see in this whole section is that now we are passing nulls to runFutureNonStatic if newlist or oldlist are null.  Does new set<Id>(om.keySet()) really give you something different from om.keySet()?!?

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -32,6 +32,7 @@
 * @date 2014
 * @group Address Management
 * @description Trigger Handler on Account for Address management 
+* @async_support this trigger handler has been tested with async=true, and appears to not cause problems.
```
ceiroa: Great idea to make this an annotation.

src/classes/ADDR_Validator_TDTM.cls
```diff
@@ -58,7 +60,8 @@ public class ADDR_Validator_TDTM extends TDTM_Runnable {
         
         if(!alreadyRun) {
 	        Addr_Verification_Settings__c settings = UTIL_CustomSettingsFacade.getAddressVerificationSettings();
-	        if(trigger.isBefore && settings.enable_automatic_verification__c) {        	
+	        if ((triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) &&
```
ceiroa: So the Trigger context variable only exists in the trigger itself, not in a class called from it? I just want to understand it to make sure I don't make the wrong assumption in the future...

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -37,6 +37,36 @@
                settings and the new ones.
 */
 public without sharing class TDTM_DefaultConfig {
+    
+    public static boolean ADDR_Address_TDTM_TestAsync = false;
+    private static boolean ADDR_Address_TDTM_TestAsyncTemp;
```
ceiroa: This might just be me being slow this morning but, don't static boolean variables get initialized to false if no value is given? And if so, is there really any way that these temp variables could ever be set to true given this logic?

src/classes/TDTM_TriggerHandler.cls
davidhabib: the point is to actually pass null if newlist or oldlist are not provided by the trigger.  your code would end up passing an empty list to the trigger handler, and several would hit errors due to assuming that if oldlist was provided, that its size matched newlist's size.  the semantics of triggers are that trigger.old or trigger.new are null if they are not provided.

src/classes/ADDR_Validator_TDTM.cls
davidhabib: trigger is only valid if the trigger handler is called syncronously.  when i was testing async, you would get an exception looking at trigger since it doesn't exist in the future call!

src/classes/TDTM_DefaultConfig.cls
davidhabib: when testing a given trigger handler, you have to manually modify this file to set the value to true.  if we pursue an automated test strategy, then we will need to provide some way to set these flags from some config some way.

PR #1756, opened 2015-10-19
updated comments and labels for Asynchronous After Events and using Async testing framework.

davidhabib:


davidhabib:
@njjc mostly comments and 2 label changes for Async.  ready for review and commit once tests pass.

PR #1761, opened 2015-10-21
adding startTest()/stopTest() to high system limit tests

davidhabib:


davidhabib:
@njjc Could you review and merge when the final tests pass.  I'd like to get this into dev so that these changes will become part of the LDV test orgs (though I guess that won't change anything since they are all seeAllData false!)

src/classes/ACCT_AccountMerge_TEST.cls
```diff
@@ -181,6 +181,7 @@ private with sharing class ACCT_AccountMerge_TEST {
         listAccLoser.add(ADDR_Addresses_TEST.listAccT[2]);
         
         // merge the Accounts
+        Test.startTest();
```
njjc: No stop test here?

davidhabib: there is already a stopTest on line 188

PR #1774, opened 2015-10-23
CONV_Account_Conversion has a lot of english language

### LINKED ISSUE #1771: CONV_Account_Conversion has a lot of english language
the Household Account Conversion page has a LOT of hardcoded english language that is untranslatable/overridable.

mrbelvedere:
Per Nic, the new custom labels should NOT be protected.

This org is being affected by the weird Label reference -> ID issue we discovered at the sprint. Only the changes to CONV_Account_Conversion.page should be kept, but contributor tool doesn't let me cherry pick commits. 

# Contribution Info

This external contribution was submitted by @cdcarter and Fixes #1771

# Critical Changes

# Changes

Closes #1771 - @cdcarter

# Issues


davidhabib:
I've cloned this pull request into #1823

cdcarter:
In some quick re-reading of packaging guides -- I think that these labels SHOULD be protected. That way, they can be removed from the package some day, which I think would make @kbromer happier. ...if I understand the protected flag correctly...

davidhabib:
protected just means whether they can be translated in the org or not (at least that is my understanding).  so any ui string should be unprotected.  labels that are internal and used in code and should not be translated should be marked protected.

cdcarter:
Ah, I misunderstood. great!

PR #1800, opened 2015-11-05
Feature/1753 thousand gau

### LINKED ISSUE #1778: Contact Merge doesn't work if Contact has a lookup to Case
The merge controller https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/CON_ContactMerge_CTRL.cls#L283 expects lookup relationships to have a `Name` APIname field, but Case has `CaseNumber`

### LINKED ISSUE #1787: In LDV Org, non-selective query error from REL_Relationships_Con_TDTM.deleteEmptyRelationships()

### LINKED ISSUE #1799: Allocations Settings: Replace GAU picklist with lookup
When large numbers of GAUs exist, the picklist is unwieldy, and is also limited to 1000 values.

njjc:
# Warning
# Info
# Issues
Fixes #1787
Fixes #1778
Fixes #1799

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -117,4 +117,8 @@ If you are on a unix based system, the following commands should work:
             replace="${lt}recordTypeVisibilities>${line.separator}      ${lt}default>true${lt}/default>${line.separator}      ${lt}personAccountDefault>true${lt}/personAccountDefault>${line.separator}      ${lt}recordType>Opportunity.NPSP_Default${lt}/recordType>${line.separator}      ${lt}visible>true${lt}/visible>${line.separator}    ${lt}/recordTypeVisibilities>${line.separator}    \1" />
         
     </target>
+    <target name="postDeploy">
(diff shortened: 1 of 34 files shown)


njjc:
@davidhabib ready for review and merge

davidhabib:
@njjc I reviewed each commit, rather than the final, since it covered different areas and I was trying to avoid confusing myself.  of course this meant I may have commented on some stuff you committed and then later recommitted to fix.  Also, it seems that your fix for #1787 was ultimately undone, so that bug stays open?  if so, it should be removed from the comment header.

njjc:
I reverted the BeforeDelete fix because I couldn't parse out the contact merge deletion case, since MasterRecordID only exists in AfterDelete. I'm not 100% positive that the ALL ROWS won't cause the query to be non-selective, since the query plan tool doesn't support it, but the easiest way we have to test is through the LDV loading mechanism, so I say we put it in and assume it's fixed, and if it ends up still throwing the error I'll reopen the bug. Thanks!

PR #1804, opened 2015-11-09
#1667. Deleting contact address data generates a blank Address record fixed

### LINKED ISSUE #1667: BUG: Deleting contact address data generates a blank Address record
Hub thread: https://powerofus.force.com/0D58000002DMFU3

Reproduced in clean org. Steps to reproduce: 

1. Contact with Address record, no override
2. Edit contact, remove data from Mailing address fields
3. Save

Result is that prior address is set to non-default. New blank record is created as the default record.

Question is whether we should assume that if a contact address is removed without making it an override, does that assume that the account should not have an address as well? And if so, then there should be no default record.

davidhabib:
# Warning

# Info

# Issues
Fixes #1667


davidhabib:
@njjc this is ready for review and merge.  thanks.

PR #1807, opened 2015-11-09
soql reduction in trigger management

davidhabib:
# Warning

# Info
- reduces use of soql statements for trigger management

# Issues

### DIFF
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -73,6 +73,13 @@ public with sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGatew
     }
     
     /*******************************************************************************************************
+     * @description to allow test code to modify the saved trigger handlers and have them reread into the cache.
+     */ 
+    public static void ClearCachedTriggerHandlersForTest() {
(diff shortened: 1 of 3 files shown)


davidhabib:
@njjc could you review and merge. this passes build and contains my optimization on TDTM soqls.

src/classes/TDTM_ObjectDataGateway.cls
```diff
@@ -46,11 +46,11 @@ public with sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGatew
      * @return Boolean True is there are no TDTM records. False otherwise.
      */
     public Boolean isEmpty() {
```
njjc: I wasn't able to find where this is used, since it uses a super common method name, but shouldn't it just pull the size of  listTH to avoid a new soql query?

davidhabib: No, this needs to look directly at the saved records to decide if it should create our default ones.  If I just look at listTH.size() == 0, that will have loaded listTH with an empty list, and it won't get reset to the saved ones.  I could I guess check listTH.size() and then set it to null, but I'd rather just leave this method alone to its original code.

PR #1808, opened 2015-11-10
Fix bug where test assumes existence of UserRole

### LINKED ISSUE #1806: Test fails if org does not have 'COO' UserRole record
The test data utility method UTIL_UnitTestData_TEST.CreateNewUserForTests() references the UserRole 'COO'.  In orgs where there is no such UserRole record, any tests that use this test data method fail.

This method should not assume the existence of the 'COO' UserRole.

dospromptman:
# Warning
# Info
# Issues
Fixed #1806

### DIFF
--- a/src/classes/ERR_Handler_TEST.cls
+++ b/src/classes/ERR_Handler_TEST.cls
@@ -417,17 +417,7 @@ public with sharing class ERR_Handler_TEST {
         if (strTestOnly != '*' && strTestOnly != 'testInactiveUserEmailBehavior') return;
 
         String un = System.now().getTime() + '@testerson.com';
-        Profile p = [SELECT Id FROM Profile WHERE Name='Standard User'];
-        UserRole r = [SELECT Id FROM UserRole WHERE Name='COO'];
-        User u; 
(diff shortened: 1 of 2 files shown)


dospromptman:
I resolved the issue in UTIL_UnitTestData_Test by performing a check for an existing UserRole, and querying that one if it exists.  If it did not exist, then it creates it.

To facilitate testing, I broke the createNewUserForTests() method into one that accepts a UserRole name and one that delegates using the default 'COO'.

I wasn't sure where to add the test methods, since this class was already suffixed '_TEST', so I added the @IsTest annotation to UTIL_UnitTestData_TEST and added the test methods to that class.  If this is an issue for any reason, let me know and I can move the test methods elsewhere.

Also, as I was testing I found another part of the code that has this same issue, ERR_Handler_TEST.testInactiveUserEmailBehavior(), which tries to create a test user with IsActive set to false for testing.  I added this functionality to UTIL_UnitTestData_TEST and replaced the code in testInactiveUserEmailBehavior() with a call to the utility test data method.

PR #1810, opened 2015-11-11
Address Bugs & BDI Bugs

### LINKED ISSUE #1448: As a user, I would like new contacts with Address override set true, to not put their mailing address in the household account, since I have their work address.
hub discussion: https://powerofus.force.com/0D58000001spM5l

the discussion also raises other potential enhancements with how addresses interact based  on the primary address type, but this tasks is really just avoiding putting the contact address into the household account on creation, if address override is checked.

### LINKED ISSUE #1509: BUG: Changed Primary Address Type on contact overwritten by type in current address
Steps to reproduce:
1) Create a contact record with a current address that has a type specified.
2) Change the Primary Address Type on the contact but no other address fields.
3) Address management code sets the Primary Address Type field to equal the type on the address record.

This one could be particularly annoying for users because it looks/behaves as though changes they made are just not being saved. There should either be a new address record created, or the address type on contacts should not be controlled by the related address record.

### LINKED ISSUE #1630: NPSP Data Import should give Error if not in Household Account model, or matches existing 1:1 contacts
The data importer should detect whether the accounts that come back from new and matched contacts are Household or other, and give an error if other.  Currently, if you have existing 1:1 contacts that get matched on data import, then the failure will occur when we try to save an address to the 1:1 account.  The error is something like "Addresses can only be saved to HOusehold and Org accoutns (if enabled)".  But the real issue is that the contact's account is a 1:1 account!  we should test for this, to give a better error message.

### LINKED ISSUE #1797: BUG: If NPSP Data Import has custom checkbox field on Opportunity, an Opportunity will be created even if false/unchecked and no other Opp data mapped
POU thread: https://powerofus.force.com/0D58000002M1JfO (replicated in clean org)

davidhabib:
# Warning

# Info

# Issues
Fixes #1509 
Fixes #1448
Fixes #1797 
Fixes #1630

### DIFF
--- a/src/classes/ACCT_AccountMerge_TDTM.cls
+++ b/src/classes/ACCT_AccountMerge_TDTM.cls
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2015 Salesforce.com Foundation
+    Copyright (c) 2015 Salesforce.org
     All rights reserved.
     
     Redistribution and use in source and binary forms, with or without
@@ -10,7 +10,7 @@
     * Redistributions in binary form must reproduce the above copyright
       notice, this list of conditions and the following disclaimer in the
       documentation and/or other materials provided with the distribution.
-    * Neither the name of the Salesforce.com Foundation nor the names of
+    * Neither the name of the Salesforce.org nor the names of
       its contributors may be used to endorse or promote products derived
       from this software without specific prior written permission.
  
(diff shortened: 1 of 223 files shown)


davidhabib:
@njjc ready for review.  build in progress.  sorry that it includes multiple unrelated fixes.

PR #1827, opened 2015-11-18
Feature/opp rt no payments

davidhabib:
# Warning

# Information
* Two new custom settings have been added to the NPSP Settings, Payments page: Opp RecTypes Excluded for Payments and Opp Types Excluded for Payments.  These settings allow you to specify types and recordtypes of Opportunities to exclude from creating Payments.

# Issues


davidhabib:
@njjc ready for review and merge!  thanks!

src/classes/PMT_PaymentCreator.cls
```diff
@@ -75,6 +75,56 @@ public with sharing class PMT_PaymentCreator {
...
+        }
+        set;
+    }
+
+    /*******************************************************************************************************
+    * @description specifies whether Payments should be excluded for the Opportunity
+    * @param opp the Opportunity to test.
+    * @return boolean true if Payments excluded for the Opportunity
+    */ 
+    private static boolean isNoPaymentForOpp(Opportunity opp) {
```
njjc: I think it would be cleaner and easier to read if we integrated the npe01__Do_Not_Automatically_Create_Payment__c into this method. What do you think?

davidhabib: you still need the other uses of npe01__Do_Not_Automatically_Create_Payment__c since those are done outside of the loop and soqls.  whereas this routine is for a specific opp.  So i felt like adding npe01__Do_Not_Automatically_Create_Payment__c in this routine would be redundant.

njjc: After looking through the whole class, I'm not seeing a place where integrating npe01__Do_Not_Automatically_Create_Payment__c here just relying on this method would be a problem.

davidhabib: the point is even if you add it within this routine, you still need to keep testing it outside of this routine like the current code does.

njjc: At what line specifically do you need it outside of this routine?

njjc: Maybe we should also consider actually checking this box when the payment isn't created automatically for another reason.

davidhabib: https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/PMT_PaymentCreator.cls#L91 still needs to reference npe01__Do_Not_Automatically_Create_Payment__c.

I don't understand your last comment.

PR #1833, opened 2015-11-20
Feature/795 close opportunity when all payments received

### LINKED ISSUE #795: As a user, when I mark the last payment as paid, I'd like my Opportunity automatically updated to a Closed Won Stage, so I don't have to manually do this.
We'd probably need to a new setting for the user to specify what the Won Closed stage is that we should use.  I assume users want this and someone must have asked about it in the past?  It seems so basic to me if you are using Opps with payments.

dospromptman:
# Warning
# Info
# Issues
Fixed #795


dospromptman:
@njjc I believe the literal value stage name behavior is a platform bug.  Don't merge this until I can fix it.

PR #1842, opened 2015-11-24
Feature/with sharing changes

### LINKED ISSUE #1479: remove "with sharing" from core classes that don't need it (AFTER security review is complete)

### LINKED ISSUE #1819: Settings page error when class from other package found starting with ADDR and without no-args constructor
See https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=00580000009bi6N&tab=sfdc.ProfilePlatformFeed&fId=0D58000002MzHmu

We probably need to wrap the line that tries to instantiate the class with classType.newInstance() in a try-catch, in case the class doesn't have a no-args constructor.

davidhabib:
# Warning

# Information

# Issues
Fixes #1479 
Fixes #1819


davidhabib:
@njjc ready for your review and merge.  it's just all the with sharing changes we talked about, plus one minor bug fix.

davidhabib:
@dospromptman Nic is out this week, so could you please review and merge this pull request.  It passes build and is ready.  Let me know if you have any questions. thanks!

dospromptman:
It looks like there was much discussion that happened before about which classes to remove 'with sharing' from.  I can only confirm that 'with sharing' was removed from several classes.

I have tested and confirmed that the NPSP Settings page will not choke when finding classes named ADDR% that do not have a no-arg constructor.

Merging!

PR #1843, opened 2015-11-25
Adding expanded pattern for matching test classes

### LINKED ISSUE #1841: Some Test Classes Not Suffixed as '_TEST'
These classes have the suffix '_TEST2' instead of '_TEST', and these are both test classes:

* [RLLP_OppRollup_TEST2.cls](https://github.com/SalesforceFoundation/Cumulus/blob/001db1d/src/classes/RLLP_OppRollup_TEST2.cls)
* [STG_InstallScript_TEST2.cls](https://github.com/SalesforceFoundation/Cumulus/blob/001db1d/src/classes/STG_InstallScript_TEST2.cls)

The issue with this is that the default matching pattern used in CumulusCI to figure out what tests to run is '%_TEST'.  These classes don't match that pattern.

As far as I can tell, this matching pattern is only used when the tests are being run using the python script test running, which is used when running records with parallel running disabled.

We may want to consider renaming these classes to match the '_TEST' suffix, or changing the matching pattern used in our CumulusCI properties to run these tests.

dospromptman:
The property `cumulusci.test.namematch` is used in a SOQL LIKE query to determine which classes to run in the python test runner.

The default value for this property is `%_TEST`, but this will not match classes that end in `_TEST2`, of which there are some test classes with this pattern.

This change adds an override for the `cumulus.test.namematch` property so that the string used in the LIKE query is `%\_TEST%`.  Here, the `%` character matches zero or more of any character.  The `_` character would normally match any single character, but here it is escaped using a backslash so that it matches a literal `_` character.

Note that the definition in the property file escapes the backslash in the value, and this is expected to be unescaped to a single backslash when the property file is read by Ant.

# Warning
# Info
# Issues
Fixed #1841


dospromptman:
@jlantz can you confirm that I got the escaping right for the property file?  I think it is right, but I am not certain of exactly how the property file gets parsed and used.

davidhabib:
@jlantz Can you please review and merge this pull request, since it is about the build system?  thanks.

PR #1845, opened 2015-11-28
Splitting test classes to avoid reaching debug log size length, enabl…

njjc:
…es limits parsing from debug logs.


njjc:
@jlantz This fixes the debug log size length for some test classes. Ready for review and merge.

PR #1847, opened 2015-11-30
Feature/health check reports

### LINKED ISSUE #1629: Health Check must gracefully handle creating an Account failing
a user reported they have a validation rule on Accounts that causes Health Check to fail to run (rather than logging an error).  Hub post here: https://powerofus.force.com/0D58000002AGcF2

the culprit is STG_PanelRDCustomInstallment_CTRL.verifyRDSaving(this) which creates a temp account assuming it will always succeed.  this needs to be in a try/catch and handled gracefully with an error log entry.

### LINKED ISSUE #1782: Enhance Health Check to provide links to reports showing suspect data
At the community sprint we went through all the health check errors, and identified which we could provide a report to.  We need to have a way of providing a link to something that opens a report (and will work in Lightning).

davidhabib:
# Warning

# Information
* This release includes a new set of Health Check reports to help identify records with issues called out by Health Check.  Special thanks to Barbara Christensen for helping with the reports during the NPSP Community Sprint.

# Issues
Fixes #1629 
Fixes #1782


davidhabib:
@dospromptman This pull request passes build and is ready.  Since Nic is out of town this week, please take a look and see if you are comfortable reviewing it.  You can post any questions you have on specific lines in the files changed tab.  Or we can also meet tomorrow to go over it together if you'd prefer.  thanks.

dospromptman:
I have reviewed this code and tested this functionality as follows:

I created an account under the Household model, switched the account model to 1-1, and created a second account.
I ran the health check.
I confirmed that the error message that is reported references the "Individual Contacts by Account Type" report.

I created a Contact with no associated Account (by disabling TDTM triggers that automatically create the Account).
I ran the health check.
I confirmed that the error message that is reported references the "Contacts without Accounts" report.


I created a class that implements the `UTIL_MasterSchedulableHelper.UTIL_IRecurring` interface, but lacks a no-args constructor.
I added a record in `Schedulable__c` that references this class.
I ran the health check.
I confirmed that the class was flagged as 'invalid'
**Note: The error message contains a grammar error:** "You must update the record to have a valid Class the implements the UTIL_MasterSchedulableHelper.UTIL_IRecurring interface."


I modified the default `Trigger_Handler__c` records by:
Deleting one existing standard trigger.
Modifying one existing standard trigger by flipping the async flag.
Duplicated one existing standard trigger.
Duplicated one existing standard trigger and modified the duplicate by flipping the async flag.
I ran the health check.
I confirmed that the deleted standard trigger was reported as an error.
I confirmed that the modified standard trigger was reported as an error.
I confirmed that the duplicated and modified trigger was reported as an error, **however, the error reported was only that the trigger does not match the default setting, not that the trigger was duplicated**.


I confirmed that the following reports referenced by the updated error labels exist:
"Individual Contacts by Account Type"
"Contacts without Accounts"
"Empty Household Accounts"
"Empty Household Objects"
"Opportunities with Primary Contact Roles"

I think the anomaly with the trigger handler error reporting is fine, since the record is still flagged as an error, even if it isn't flagged twice. @davidhabib, do you agree?

davidhabib:
Wow @dospromptman, you have really gone above and beyond the call of duty!  We have not typically expected the reviewer to actually install the branch and do adhoc testing to verify each issue.  We've only required a code review, and followup with any questions.  
1. I think the issue with the errors in trigger handlers is fine, and I think we can go with it as is.  So merge the pull request if you agree.
2. We should discuss at our next Architecture review meeting how we might want to refine our current review process.  While I certainly appreciate the testing you did, it certainly is a larger investment in time to our current process and we should discuss.
Thanks!

kbromer:
:+1:

PR #1858, opened 2015-12-09
changed filters on two health check reports to avoid report data limits.

davidhabib:


davidhabib:
@jlantz These are the two report changes we need for version 3.57

davidhabib:
@jlantz I'll let you merge this pull request.  you can see from the Files Changed tab, it's just two report files.  Thanks.

PR #1867, opened 2015-12-14
As a user, I'd like to be able to run the Data Importer without having the Customize Application system permission

### LINKED ISSUE #1707: As a user, I'd like to be able to run the Data Importer without having the Customize Application system permission
BDI_DataImport_CTRL.importData() attempts to upsert the custom setting record before starting the data import batch job. If the user does not have the Customize Application permission, then they will get an error that they don't have access:
screen shot-npspdataimport

mrbelvedere:
This is a fairly brute-force way of solving the problem. There will still be an error if the user doesn't have Customize App AND there's no org instance yet, but at least the error is pushed further up the chain. 

# Contribution Info

This external contribution was submitted by @cdcarter and Fixes #1707

# Critical Changes

# Changes

With this change, non-system admins can be configured to do batch data importing. 

# Issues

### DIFF
--- a/src/labels/CustomLabels.labels
+++ b/src/labels/CustomLabels.labels
@@ -276,14 +276,6 @@
         <value>You cannot verify more than 100 addresses at a time.</value>
     </labels>
     <labels>
-        <fullName>CampaignMemberStatusOmit</fullName>
-        <categories>Opportunity, Campaign Member</categories>
-        <language>en_US</language>
-        <protected>false</protected>
-        <shortDescription>CampaignMemberStatusOmit</shortDescription>
-        <value>omit</value>
-    </labels>
-    <labels>
         <fullName>CONV_Accept_Risk</fullName>
         <language>en_US</language>
         <protected>false</protected>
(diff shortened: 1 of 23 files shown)


src/labels/CustomLabels.labels
```diff
@@ -3464,7 +3464,7 @@ Once the process is 100% complete, you can safely leave or refresh this page.</v
         <language>en_US</language>
         <protected>false</protected>
         <shortDescription>stgHelpRDAddCampaign</shortDescription>
-        <value>If selected, adds the campaign that you’ve specified for the Recurring Donation to all related (i.e. “child”) opportunities. If this option is not selected, the campaign appears on the Recurring Donation’s first opportunity only.</value>
+        <value>If selected, adds the campaign that youâve specified for the Recurring Donation to all related (i.e. âchildâ) opportunities. If this option is not selected, the campaign appears on the Recurring Donationâs first opportunity only.</value>
     </labels>
```
davidhabib: @jlantz looks like the contributor tool is not dealing well with single and double quotes, or Christian's system is doing it.

cdcarter: I was working out of the Developer Console, so nothing I did should have touched labels at all.

njjc:
Moved over to #1880 to fix metadata issues.

PR #1872, opened 2015-12-15
Adding UTIL_RecordTypes method to tell if a record type is the defaul…

njjc:
…t for the current profile. Updating Payments test to only run record type exclusion if a non-default record type is available for opportunities.


njjc:
@davidhabib ready for review and merge, this should fix the current UAT failure.

PR #1880, opened 2015-12-22
Feature/1707 data importer perms

### LINKED ISSUE #1707: As a user, I'd like to be able to run the Data Importer without having the Customize Application system permission

njjc:
# Warning
# Info
- Thanks to Christian Carter @cdcarter for the contribution allowing non-system admins to be configured to do NPSP batch data import.

# Issues
Fixes #1707


src/classes/BDI_DataImport_API.cls
```diff
@@ -37,6 +37,11 @@
 global with sharing class BDI_DataImport_API {
 
      private class BDIException extends Exception {}
+    
+	public static boolean userCanEditSettings {get {
```
njjc: @cdcarter this isn't needed in this in this class, right? It's only used in the controller?

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -35,6 +35,10 @@
 * Needs to be "without sharing" so that the Install Script can run it.
 */
 public without sharing class UTIL_CustomSettingsFacade {
+    
+   	public static boolean userCanEditDataImportSettings {get {
```
njjc: @cdcarter any reason this is in the facade? feels like it should be directly in the controller since it's only used there.

cdcarter: I pulled it into the facade without a full knowledge of how the facade wants to be structured. I don't know if there's a world where the facade cares about user object permissions.

I think it'd be safe just to be in the controller.

src/classes/BDI_DataImport_API.cls
cdcarter: Definitely not needed in this class. At first, I was checking everywhere down the chain and deleted a lot of those checks but I guess accidentally left this helper. Should only be used in controller.

BrightStepTracy:
Wow.  Thank you @cdcarter, @njjc and everyone who worked on this - it opens up a huge new horizon for using this tool!

PR #1881, opened 2015-12-22
Enhancements around Deceased contacts, fixes #1870, #1750, #1588, #1572

### LINKED ISSUE #1572: If the only member of a household/primary contact is marked as deceased, Exclude from Household Name should be unchecked
This is to avoid an "Anonymous Household." Exclude from Greetings should still be checked, opt outs should still be checked.

POU thread: https://powerofus.force.com/0D580000020pFeI

### LINKED ISSUE #1588: If the Household primary Contact is marked deceased, another Household member should be automatically elevated to be the primary Contact

### LINKED ISSUE #1750: When all household members are deceased, use different household naming convention
https://powerofus.force.com/0D58000002K4x6Y

### LINKED ISSUE #1870: If a Contact is marked as Deceased, they should no longer get soft credit for Household gifts
POU thread: https://powerofus.force.com/0D58000002Q4c7x

njjc:
# Warning
# Info
# Issues
Fixes #1870
Fixes #1750
Fixes #1588
Fixes #1572

### DIFF
--- a/src/classes/CON_DoNotContact_TDTM.cls
+++ b/src/classes/CON_DoNotContact_TDTM.cls
@@ -36,6 +36,8 @@
 */
 public class CON_DoNotContact_TDTM extends TDTM_Runnable {
     
+    private DmlWrapper dmlWrapper = new DmlWrapper();
+
     /*******************************************************************************************************
     * @description Handles deceased and do not contact fields.
     * @param listNew the list of Accounts from trigger new. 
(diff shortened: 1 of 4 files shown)


src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -64,25 +97,45 @@ public class CON_DoNotContact_TDTM extends TDTM_Runnable {
...
         markContactDoNotContact(con);
 
+        if (acc!=null && con.Deceased__c) {
+            list<Contact> contacts = (list<Contact>)acc.Contacts;
+            //if there are non-deceased members of the household
+            //Exclude from household name and update household's primary contact
+            if (!contacts.isEmpty()) {
+                excludeFromHHNaming(con);
+                if (con.Id != null && acc.npe01__One2OneContact__c == con.id) {
+                    acc.npe01__One2OneContact__c = contacts[0].id;
```
davidhabib: are you sure that acc.Contacts doesn't hold the current contact who is being marked deceased.  I saw that you query for all the household contacts who aren't deceased, but since you do this in the BeforeInsert and BeforeUpdate, I think that contact might be returned in your query?

davidhabib:
it looks like if the only remaining household member is deceased, you avoid removing them from naming.  But I thought the user story was to somehow make the name show this.  ideally customizeable, but that is probably too big a lift for now.  I thought we were maybe going to just put (deceased) after the lastname.  Sorry if I'm premature in reviewing this, or you guys decided to not do the special naming part.

njjc:
Not premature. Mary was happy with just not renaming households when the last person is marked deceased. I think we can include this as-is and discuss a new naming method as an enhancement.

src/classes/CON_DoNotContact_TDTM.cls
njjc: the "Id NOT IN :newlist" part of the query should filter them out.

PR #1908, opened 2016-01-25
Updating links in NPSP Resources (Getting Started) page

dospromptman:
Removed jumbotron styling from main heading, to reduce footprint on the page and allow more content to appear "above the fold".

Replaced "Workbook" links section with "Documentation" section, where now the icon links to the NPSP documentation and the text contains links to both the NPSP documentation and the workbook.

The existing URLs for the workbook and the data import guide were broken.  I have replaced those links with links to static pages in the hub (as opposed to directly linking to the Salesforce object for the PDF).


dospromptman:
@njjc Can you review this whenever you have time?  Hold off on merging it until Jenkins is back.

PR #1920, opened 2016-02-04
description and help text for Opp matching gift fields

davidhabib:


davidhabib:
@njjc a few missed help text and descriptions for matching gift fields.

davidhabib:
@njjc ready for review and merge.

PR #1931, opened 2016-02-08
Feature/njjc multicurrency bugfix

njjc:


src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -216,6 +205,28 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
+            npe01__Payment_Field_Mapping_Settings__c pfms = PMT_PaymentCreator.paymentMappings.get(s);
+            if (!query.contains(pfms.npe01__Opportunity_Field__c.toLowerCase())) {               
+                query += ', ' + pfms.npe01__Opportunity_Field__c;
+            }   
+        }       
+        //if currencyiso field exists add it to query for use later
+        if(Schema.sObjectType.Opportunity.fields.getMap().get('CurrencyIsoCode') != null)
+            query = query + ',CurrencyIsoCode';
+            
+        query += ' from Opportunity where id = :id';
```
davidhabib: seems pretty obscure that this will make a query string that refers to :id that the caller had better provide.  I would have preferred seeing Id passed in, and then the value of Id put in the query string to avoid this dependency.  Especially since this is listed as a public function.  Is it really private, or used elsewhere?

src/classes/PMT_PaymentWizard_TEST.cls
```diff
@@ -351,6 +362,7 @@ private class PMT_PaymentWizard_TEST {
             stagename = closedWonStage()
         );
         insert opp1;
+        list<opportunity> queryopp = [select id, currencyisocode from opportunity where id=:opp1.id];
```
davidhabib: why add this query to the test?  It isn't used anywhere that I can see.  should it?  maybe some assert dealing with currencyisocode?

PR #1954, opened 2016-02-16
Feature/njjc lightning

### LINKED ISSUE #1930: Manage Allocation: Save button no longer disabled automatically when allocated amount exceeds opp amount
**lurch: attach W-009301

njjc:
# Warning
# Info
# Issues
Fixes #1930

### DIFF
--- a/src/pages/PMT_PaymentWizard.page
+++ b/src/pages/PMT_PaymentWizard.page
@@ -1,155 +1,363 @@
-<apex:page controller="PMT_PaymentWizard_CTRL" tabStyle="Opportunity">
+<apex:page controller="PMT_PaymentWizard_CTRL" tabStyle="Opportunity" standardStylesheets="false">
+    <apex:stylesheet value="{!URLFOR($Resource.SLDS, 'assets/styles/salesforce-lightning-design-system-vf.css')}" />
+
 <!--
 Opp Payment Creator VF Page
 Copyright (c) 2008 Evan Callahan, evanc@npowerseattle.org, NPower Seattle, 403 23rd Ave S, Seattle, WA 98144
 -->
+    <style>
+        html body.sfdcBody {
+            padding: 0;
+        }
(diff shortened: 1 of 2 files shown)


src/pages/PMT_PaymentWizard.page
```diff
@@ -1,155 +1,363 @@
...
+                    {!$Label.pmtWizardWarning}
+                </p>
+                <p>
+                    {!$Label.pmtWizardNoUnpaidAmount}&nbsp;&nbsp;
+                    <a href="{!URLFOR($Action.Opportunity.View, currentOpp.id)}">
+                        {!currentOpp.Name}
+                    </a>
+                </p>
+            </div>
+            <div class="slds-grid slds-grid--align-center">
```
dospromptman: This button bumps up against the alert and the div below it.  It would probably look better if this div had some padding (slds-p-vertical--medium maybe?)

dospromptman:
It doesn't look like the code changed for this part, but it was incredibly confusing to me that if I had paid payments the button for "remove payments" doesn't actually remove the payments!  It just lets me schedule more-- but it still shows the alert and the "remove payments" button regardless.  I assume this is normal functionality, since I can't see anything in the code that would actually remove those payments or cause that notice to be removed (in the old version or this one).

Also, I had a lot of confusing moments due to the Opportunity page and the Payments related list not updating after scheduling new payments.  At one point, I clicked 'View All' from the payments related list card on the opportunity detail page and saw 4 payments when there should only have been 3.  I navigated back to the Opportunity detail page and when I did I saw the related list card refresh from 4 to 3 payments.  When I clicked 'View All' again, it showed just 3 payments.  The latency around these UI updates is *incredibly* confusing and misleading.... but there isn't much we can do about that.

dospromptman:
Confirmed that "Save & Close" button on manage allocations is disabled when allocated amount exceeds available amount.

PR #1963, opened 2016-02-19
Account to account associations

### LINKED ISSUE #1769: Account to account associations
Hello from the Seattle NPSP sprint.

We are planning to work on account-to-account associations (paralleling contact-account relationships).

This will entail creating an Associated Account join object between account records for the purpose of tracking partnered organizations within the NPSP. Including the following fields:
- Account (master-detail)
- Associated Account (lookup w/ cascade delete)
- Association (picklist)
- Reciprocal Association (picklist)
- From Date
- To Date
- Description
- Status

Creating an Account Association will trigger the creation of a reciprocal Account Association in the other direction so that both outgoing and incoming associations can be viewed in a single related list. Deleting an Account Association will delete the reciprocal. Updating an Account Association will update the Association picklists, from/to date and status on the reciprocal.

mrbelvedere:
Hello NPSP team,
This is an implementation of account-account associations as described in this doc: https://docs.google.com/document/d/11e5I3m8Km8CB1b847tT3j7-xe5Hdnj7XHGvfFzBce9k/edit?pli=1

Tracy Kronzak can provide more context and any relevant info from the Advisory Board's review of this feature.

This was mostly implemented during the NPSP sprint in Seattle and shortly thereafter, so is probably somewhat out of date with master by now...sorry!

# Contribution Info

This external contribution was submitted by @davisagli and Fixes #1769

# Critical Changes

# Changes

# Issues


judisohn:
Hi @davisagli! Thank you so much for the submission. We'll get it reviewed and considered as soon as we can. (cc @mpusto)

donners77:
Hi. I'm interested in using this feature. Any news on it?

davidhabib:
@mpusto From a dev perspective, this is about 8-13 story points to clone, update, test, and merge.  There are of course documentation costs associated with it.

mpusto:
Thank you for taking the time to submit this pull request. We've been doing a lot of reviewing and consideration, and we are not going to accept it at this time.  We want to be completely transparent with our community about the reasons why so that others understand the process.

We are not accepting this PR at this time for a few reasons.

1. We are always cautious about what we add to the codebase due to the increase in technical debt and something we call "feature bloat".  We want to ensure that we keep NPSP as "lean" as possible. This feature, while useful to some organizations, has rarely been requested by a broad swath of customers. If we are going to add something to the codebase, we want to ensure that it is broadly applicable to the thousands of customers who use our product every day.

2. If we did see a groundswell of customers requesting this feature, we would likely architect the solution differently.  in evaluating the implementation of the feature vis-a-vis previously defined requirements, we felt we may want to make different design and/or architectural decisions based on the use-cases in question

3. The current implementation of this feature creates two records for the Account to Account relationship, which can be undesirable for organizations who are concerned with storage. As mentioned previously, we try to keep NPSP "lean" and that extends to the footprint it has on data storage as well.

We greatly appreciate the contribution and we are sorry we were not able to provide a more timely response. If you have any comments or concerns, please do not hesitate to reach out to me.

Thank you again for your work and we look forward to seeing you in the community.

davisagli:
Hi, @mpusto. Thanks for the explanation. I can certainly understand the decision on the basis of not adding something when a lot of customers aren't asking for it, especially since this is functionality that could be implemented in a separate package.

I don't understand the critique of how the functionality is architected using two symmetric association objects. We chose this design partly because it's the same approach used for the contact-contact relationships that are already part of NPSP. We also recognized it as non-ideal, but as the only way we are aware of to build this functionality, due to the platform's restriction against having multiple master-detail fields to the same object type. Finally, the storage concerns are a real tradeoff, but can be documented so that orgs concerned about this can choose not to disable the feature.

You're being noncommittal about *how* the team would architect it differently, which is understandable, but if anyone has suggestions about alternate approaches, I'd be interested to hear them.

best,
David

mpusto:
We're being non-committal because we don't understand fully all of the various use cases and ways in which a feature like this would be used.  Its tough to say "we don't think there's a groundswell of support for implementing it, but we have an architecture that we like" - its sort of presuming the answer to the first question about broad applicability is more fully fleshed out.

This also began during a period of transition for NPSP. I only started in December, after you had begun work on this. Since then we have done a lot of backlog reviewing, architecture review, and long-range planning. This particular feature just isn't high on our priority list as something to incorporate and support due to the amount of maintenance, refactoring, and new features we want to support at this time with the resources we have. That priority is based on other areas having a broader impact on our community of users.

If you'd like to make your code available to anyone who would like to "add on" I recommend offering it as a repository and encouraging people to use the "[deploy from Github](https://andyinthecloud.com/2013/09/24/deploy-direct-from-github-to-salesforce/)" tool provided by Andy Fawcett.

davisagli:
Yeah, that all makes sense. I'm not complaining.

I'm not likely to personally work on taking this further (it's not scratching my own itch; just something that seemed like low-hanging fruit at the sprint), but if someone else wants to run with it that's fine with me.

PR #1965, opened 2016-02-22
Fixing #1950 and whitespace issues.

### LINKED ISSUE #1950: NPSP Settings for Address: Settings not saved: Attempt to de-reference a null object.
After checking the box to set minor changes as updates and trying to save, receive the error "Settings not saved: Attempt to de-reference a null object." Here are the steps to reproduce:

- In the Addresses section of NPSP Settings click Edit
- Check the box for "Simple Address Change Treated as Update"
- Click Save
- Settings not saved, remain in Edit screen, and have "Settings not saved: Attempt to de-reference a null object" error in red below Save button.

If I then, check the box for "Enable Automatic Verification" and select "The Google Geocoding API" or "Cicero" for the verification service, I'm able to save without error. I can't select "SmartyStreets" and save without providing an Auth Token. @njjc

njjc:
# Warning
# Info
# Issues
Fixes #1950

### DIFF
--- a/src/classes/ADDR_Validator_TEST.cls
+++ b/src/classes/ADDR_Validator_TEST.cls
@@ -61,11 +61,11 @@ public with sharing class ADDR_Validator_TEST {
     }
     
     static testMethod void testContactAddress() {
-    	if (strTestOnly != '*' && strTestOnly != 'testContactAddress') return;
-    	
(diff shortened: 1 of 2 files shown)


njjc:
This is best viewed with the ?w=1 parameter to ignore whitespace changes, as I replaced a bunch of tabs with spaces.

Issue #1851, opened 2015-12-01
As a developer, I'd like to understand the process needed to contribute to the NPSP


cdcarter:
Specifically, guidance in writing issues, guidance in what issue labels are (`community-ready`)?, an understanding of the CLA process, etc... Not the technical steps, but the open source stuff!

kbromer:
@judisohn is working on a site for this right now!

judisohn:
Hang tight @cdcarter - coming soon, I promise. This actually came out of the conversation you and I had at the sprint about increasing the transparency of the work we're doing.

cdcarter:
You all rock!

judisohn:
Glad you did. :-)

**lurch: attach W-008829

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000001hThAAI" target="_blank">W-008829</a>

cdcarter:
Just popping questions into here: is it preferred that we create issues with the contrib tool that are owned by mrbelvedere, or should we create issues with our github creds and then link contributions to them?

judisohn:
Others may have opinions, but I think it would be better to create the issue first, then link to it in the tool. @jlantz do you have a preference?

judisohn:
@cdcarter Safe to close this one yet? Especially now that we have the great new docs?

PR #1972, opened 2016-02-28
Feature/1971 double count opps

### LINKED ISSUE #1971: Double counting opportunity amounts with duplicate OCRs
Currently, if a contact holds two soft credit OCRs on an opportunity (e.g. in a peer-to-peer campaign, a contact may be both Household Member and Fundraiser, if their spouse gives to their campaign), that opportunity's value will be double counted in the soft credit rollups. 

@judisohn writes: 
> Yes, it’s expected… We don’t expect for the same contact to be listed more than once and advise against that. Probably not a bad idea to anticipate that it’s not intended and code rollups to not double count, but right now we don’t do that.

It'd be great if an opportunity was counted once and only once in soft credit rollups, regardless of number of OCRs.

cdcarter:
Hey folks, all tests are green.

In the past, opportunities that were linked to a Contact via Opportunity Contact Roles more than once would be double-counted in rollups and report totals. The fix in this release prevents this double-counting from occurring, so reports and soft credit rollups could potentially appear with different totals than in the past. The totals in an organization's reports and soft credit rollups will now be correct. To check if this fix affects your organization, there is now a Health Check (NPSP Settings-->System Tools-->Health Check) which will identify Contacts who may have had bad data in the past.

# Contribution Info

This external contribution was submitted by @cdcarter and closes #1971.

# Critical Changes

# Changes

# Warning

# Issues

Fixed #1971 - When a contact has multiple soft credit OCRs for an opportunity, that opportunity's value will now only be counted once in soft credit rollups.


judisohn:
This is awesome as always @cdcarter - thank you!! 

Now that I'm thinking about it, though, I'm wondering if deduplicating Contact Roles should be a Settings option, similar to the way it is for Relationships? Pushed unchecked, checked in new trials. The reason I'm asking is because when this is pushed, the next day those soft credit rollups are going to recalculate and may very well change results suddenly in reports/dashboards.

src/classes/RLLP_OppPartialSoftCreditRollup.cls
```diff
@@ -116,18 +115,21 @@ public class RLLP_OppPartialSoftCreditRollup {
                 newC.put('CurrencyIsoCode', c.get('CurrencyIsoCode'));
             }
             
-            Set<String> ocrIds = new Set<String>();
+            Set<Id> ocrForPSCIds = new Set<Id>();
             if (partialSoftCredits.containsKey(c.Id)) {
                 for (Partial_Soft_Credit__c psc : partialSoftCredits.get(c.Id)) {
                     newC = countPartialSoftCredits(newC, psc, currentYearStart);
-                    ocrIds.add(psc.Contact_Role_ID__c);
+                    ocrForPSCIds.add(psc.Contact_Role_ID__c);
```
davidhabib: given psc.Contact_Role_ID__c isn't actually an ID (Salesforce provides no way to specify a lookup to an OCR), doesn't this change to use a set<ID> now generate a compile error or warning?

src/classes/RLLP_OppPartialSoftCreditRollup.cls
```diff
@@ -116,18 +115,21 @@ public class RLLP_OppPartialSoftCreditRollup {
...
                 }
             }
             
+            Set<Id> processedOpportunityIds = new Set<Id>();
             if (opportunityContactRoles.containsKey(c.Id)) {
                 for (OpportunityContactRole ocr : opportunityContactRoles.get(c.Id)) {
-                    if (ocrIds.contains(ocr.Id)) continue; // partial soft credit
+                    if(ocrForPSCIds.contains(ocr.Id)) continue; // partial soft credit
```
davidhabib: our coding guidelines specify if (), not if().  but when we modify existing code, we try to keep it consistent with the body of code we are modifying.  So please use spaces after your if.

davidhabib:
@judisohn This pull request prevents multiple OCR's from the same contact giving the soft credit multiple times.  It does not do this for PSC's, however.  Is that the desired behavior?  I assume so, but want to verify since it could seem inconsistent to the user.  If we decide that PSC's should also limit to one per contact, then which PSC amount would you use?  So I think we need to verify the desired behavior before moving forward with this.

davidhabib:
Thanks @cdcarter for the bug fix!  We need to confirm desired behavior, and make a decision on the settings work for this before it goes in.  but if you are able to update with the minor changes I mentioned, great!

judisohn:
@davidhabib I just did some testing and I think this is already accounted for in PSC - it throws an error - but it *does* create the duplicate OCR. So this may not be something to worry about?

image

src/classes/RLLP_OppPartialSoftCreditRollup.cls
cdcarter: Whoops, that was just me moving too fast, not intentionally trying to shift up the style guide. Somewhat related, is there a link to your coding guidelines? Nothing [on the contributor page](http://developer.salesforce.org/Cumulus/Contributor/) but I would be happy to follow them.

src/classes/RLLP_OppPartialSoftCreditRollup.cls
cdcarter: Since the string IS a valid Id, Set<Id> will work fine even if the field they are coming from is just text. Compiles fine and passes tests.

cdcarter:
I would expect that you could have multiple PSCs for an individual, as you can be explicit about how the money is allocated/rolled up. But I would also see that as an enhancement over current behavior, whereas double counting OCRs (when no PSC is in play) feels like a bug.

src/classes/RLLP_OppPartialSoftCreditRollup.cls
cdcarter: related to #1851.

cdcarter:
Have you had a chance to confirm desired behavior here/make a decision on whether or not this needs a setting? Let me know if there are any next steps I should take.

cdcarter:
@judisohn contacted me offline and confirmed that we DO need a setting. I'll work on bringing the branch up to par to support a setting, but if the labels bug shows up again I may need some assistance adjusting the settings page.

davidhabib:
@cdcarter Please refresh my memory.  Were you hitting the blocking label/Id bug in settings?  If so, we will have to take it from here.  Also worth nothing that Settings has undergone SLDS styling and some reorganization, so if you do decide to undertake this work you would need to get fully synced up to our current Dev branch.

@mpusto If Christian isn't able to do this additional work, then I'd say put in 8 story points to have someone in dev clone this branch and get it updated, tested, and then adding in the new settings work.

PR #1984, opened 2016-03-14
Fix nav links on settings page

dospromptman:
Adding slds-grow class to nav links to allow them to be clicked anywhere
within their parent element.

Also, fixing id of Address nav link to properly show selected tab
status.


njjc:
Added changes manually to avoid a deploy.

PR #1985, opened 2016-03-14
Feature/npsp settings slds

njjc:
# Warning
# Info
- NPSP Settings has been redesigned in SLDS.
- Bulk data processing tools now have their own section in NPSP Settings.

# Issues


src/pages/STG_SettingsManager.page
```diff
@@ -1,78 +1,74 @@
-<apex:page showHeader="true" sidebar="false" controller="STG_SettingsManager_CTRL" action="{!setTriggerSchedConfig}" docType="html-5.0" >
+<apex:page showHeader="true" sidebar="false" controller="STG_SettingsManager_CTRL" action="{!setTriggerSchedConfig}" docType="html-5.0" standardStylesheets="false">
```
davidhabib: I don't like how the navbar is now the same background as the page.  It no longer looks like a separate section.  If you look at the navbar in lightning setup, that I assume you are trying to mimic, it does appear in its own section since they have a gray background on the current page they host.  I'd like us to at least add a vertical line demarcating the two.  Jade might have more input on whether some amount of difference in background should be used.

also, their highlight for an indented item still spans the whole width of the navbar.  Would be nice if you could do the same if that is relatively easy.

image

src/pages/STG_PanelHouseholds.page
```diff
@@ -106,228 +106,182 @@
...
-                        <apex:inputText value="{!stgService.stgHN.Name_Connector__c}" id="txtAnd" onchange="setExamples();" rendered="{!isEditMode}" styleClass="form-control" />
-                    </div>
-                    <div class="col-sm-offset-4 col-sm-8 help-block">
-                        <apex:outputText value="{!$Label.stgHelpNameConnector}" />
-                    </div>
+            </div>
+            <div class="slds-form-element">
+                <apex:outputLabel value="{!$ObjectType.Household_Naming_Settings__c.Fields.Household_Name_Format__c.Label}" for="slstrFormatHH" styleClass="slds-form-element__label" />
+                <div class="slds-form-element__control">
+                    <apex:outputText value="{!stgService.stgHN.Household_Name_Format__c}" rendered="{!isReadOnlyMode}" styleClass="slds-form-element__static" />
```
davidhabib: The other textbox now appears too narrow.  it used to be as wide as all the other controls.  same goes for the two remaining other textboxes.

src/pages/STG_PanelAddrVerification.page
```diff
@@ -2,19 +2,19 @@
```
davidhabib: when you go into Edit mode on the Address page, the Edit button remains visible in the header.  The save and cancel buttons do correctly display at the bottom.

src/pages/STG_PanelAddrVerification.page
```diff
@@ -196,38 +187,39 @@
...
-                        <div class="panel-body" style="padding:15px;">                 
-                            <apex:outputText escape="false" value="{!$Label.Addr_Verification_Batch_Body}" />
-                            <apex:outputText id="validatorBatchHelpText" escape="false" value="{!validatorBatchHelpText}" />
-                            <label class="checkbox" style="margin-left: 15px"> 
-                                    <apex:inputCheckbox value="{!skipPreviouslyVerifiedObjects}" />{!$Label.Addr_Skip_Verified}
-                            </label>
-                            <div style="text-align:center; margin-top: 15px;">
-                                <apex:commandButton id="batchVerify" action="{!batchVerify}" rerender="batchVerifyResult" value="Verify All Addresses" disabled="{!NOT batchVerifEnabled}" styleClass="btn-sm btn-default"/> 
+
+        <div class="slds-m-around--large" id="iBatchVerifyDiv">
```
davidhabib: The Mass Verify Existing Addresses box only displays when you pick Smarty Streets, but the verify button is disabled, and there is a message that the selected service does not support Batch Address Validation.

src/pages/STG_PanelAddrVerification.page
```diff
@@ -2,19 +2,19 @@
 
     <script>
         var j$ = jQuery.noConflict();
```
davidhabib: On all of our settings pages, the Label font is a different size than the value font, and so their baseline is not lining up.  It looks off.  Worse, those properties who are checkboxes look very off, as the checkbox is way above the baseline.  (I just put this comment on the first page)

src/pages/STG_PanelMembership.page
```diff
@@ -1,38 +1,30 @@
...
-                    <apex:outputText value="{!$Label.stgHelpMembershipRT}" />
+    <apex:form id="form" styleClass="slds-m-around--x-large">
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavDonations}" pageLabel="{!$Label.stgNavMembership}" />
+        <c:UTIL_PageMessages />
+        <div class="slds-grid slds-grid--align-center slds-grid--vertical-align-center slds-m-around--large">
+            <apex:commandButton value="{!$Label.stgBtnEdit}" status="statusLoad" action="{!editSettings}" rendered="{!isReadOnlyMode}" immediate="true" rerender="form" styleClass="slds-button slds-button--small slds-button--neutral" />
+        </div>
+        <div class="slds-form--horizontal slds-m-around--large">
+            <div class="slds-form-element">
+                <apex:outputLabel value="{!$ObjectType.npo02__Households_Settings__c.Fields.npo02__Membership_Record_Types__c.Label}" for="slMRT" styleClass="slds-form-element__label" />
```
davidhabib: clicking on this label did not select the picklist.

davidhabib: seems to happen on other multiselectpicklists, so maybe its an issue we can't solve (or need to log a separate issue for)?

src/pages/STG_PanelContactRoles.page
```diff
@@ -1,159 +1,115 @@
...
-                    <apex:outputLabel value="{!$ObjectType.npo02__Households_Settings__c.Fields.npo02__Household_Contact_Roles_On__c.Label}" for="cbxHCRO" styleClass="col-sm-4 control-label" />
-                    <div class="col-sm-8 form-control-column">
-                        <apex:inputCheckbox value="{!stgService.stgHH.npo02__Household_Contact_Roles_On__c}" disabled="{!isReadOnlyMode}" id="cbxHCRO" />
-                    </div>
-                    <div class="col-sm-offset-4 col-sm-8 help-block">
-                        <apex:outputText value="{!$Label.stgHelpHHOCROn}" />
-                    </div>
+            </div>
+            <div class="slds-form-element">
+                <apex:outputLabel value="{!$ObjectType.npo02__Households_Settings__c.Fields.Matched_Donor_Role__c.Label}" for="idDBMS" styleClass="slds-form-element__label" />
```
davidhabib: clicking on this label does not select its picklist.

src/pages/STG_PanelRD.page
```diff
@@ -1,81 +1,72 @@
...
-                <apex:commandButton value="{!$Label.stgBtnEdit}" status="statusLoad" action="{!editSettings}" rendered="{!isReadOnlyMode}" immediate="true" rerender="form" styleClass="btn-sm btn-primary" />
+    <apex:form id="form" styleClass="slds-m-around--x-large">
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavRecurringDonations}" pageLabel="{!$Label.stgNavRecurringDonations}" />
+        <p class="slds-text-body--small slds-m-top--x-small">
+            <apex:outputText value="{!$Label.npe03__RecurringDonationSettingsDescription}" escape="false" />
+        </p>
+        <c:UTIL_PageMessages />
+        <div class="slds-grid slds-grid--align-center slds-grid--vertical-align-center slds-p-around--large">
+            <apex:outputPanel layout="block" rendered="{!isReadOnlyMode}" styleClass="slds-button-group">
+                <apex:commandButton value="{!$Label.stgBtnEdit}" status="statusLoad" action="{!editSettings}" rendered="{!isReadOnlyMode}" immediate="true" rerender="form" styleClass="slds-button slds-button--small slds-button--neutral" />
```
davidhabib: when you click save or cancel, and the Edit button is redisplayed, it is clipped on its left side.  I reproduced multiple times.  not sure what is specific about this Edit button vs all our other ones!

src/pages/STG_PanelRDCustomInstallment.page
```diff
@@ -1,64 +1,67 @@
...
-                <div class="form-group">
-                    <apex:outputLabel value="{!$ObjectType.npe03__Custom_Installment_Settings__c.Fields.Name.Label}" for="tbxN" styleClass="col-md-4 control-label" />
-                    <div class="col-md-8 form-control-column">
-                        <apex:inputField value="{!ciNew.Name}" rendered="{!isEditMode}" required="false" id="tbxN" styleClass="form-control" />
-                    </div>
-                    <div class="col-md-offset-4 col-md-8 help-block">
-                        <apex:outputText value="{!$Label.stgHelpRDPeriodName}" />
+    <apex:form id="form" styleClass="slds-m-around--x-large">
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavRecurringDonations}" pageLabel="{!$Label.stgNavRDInstallmentPeriods}" />
+        <c:UTIL_PageMessages />
```
davidhabib: The messages displayed by this component are using too large a font, that seems way out of proportion to our settings page.  Look at this example and see what I mean.
image

src/pages/STG_PanelOppBatch.page
```diff
@@ -1,18 +1,19 @@
...
+        <div class="slds-text-body--small slds-m-around--medium">{!$Label.stgHelpBatchOppRollup}</div>
+        <c:UTIL_PageMessages />
+        <div class="slds-grid slds-grid--align-center slds-grid--vertical-align-center slds-p-around--large">
+            <apex:commandButton value="{!$Label.stgBtnRunBatch}" status="statusLoad" action="{!runBatch}" disabled="{!isRunningBatch}" immediate="true" rerender="form" styleClass="slds-button slds-button--small slds-button--neutral" />
+        </div>
+        <apex:outputPanel rendered="{!isRunningBatch}">
+            <c:UTIL_JobProgressLightning id="jobProgress"
+                eventTargetId="{!$Component.jobProgress}"
+                strBatchComponentLabel="{!$Label.stgLabelOppBatchProgress}"
+                cNumberOfJobs="1"
```
davidhabib: Our Opp rollups are 4 jobs not 1.  only one progress is being displayed.

src/pages/STG_PanelAlloBatch.page
```diff
@@ -1,18 +1,19 @@
...
-                <c:UTIL_JobProgress strBatchComponentLabel="{!$Label.stgLabelAlloBatchProgress}" cNumberOfJobs="1" bRerenderOnComplete="true" />
-            </apex:outputPanel>
-            
-        </apex:pageBlock>
+    <apex:form id="form" styleClass="slds-m-around--x-large">
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavBulkProcesses}" pageLabel="{!$Label.stgNavRollupAlloBatch}" />
+        <div class="slds-text-body--small slds-m-around--medium">{!$Label.stgHelpBatchAlloRollup}</div>
+        <c:UTIL_PageMessages />
+        <div class="slds-grid slds-grid--align-center slds-grid--vertical-align-center slds-p-around--large">
+            <apex:commandButton value="{!$Label.stgBtnRunBatch}" status="statusLoad" action="{!runBatch}" disabled="{!isRunningBatch}" immediate="true" rerender="form" styleClass="slds-button slds-button--small slds-button--neutral" />
```
davidhabib: button always is disabled for me, even after refreshing the page, so I can never run this batch!

njjc: Default allocations need to be enabled... the help text says so, but it could be clearer.

src/pages/STG_PanelOppBatch.page
```diff
@@ -1,18 +1,19 @@
...
-        </apex:pageBlock>
+    <apex:form id="form" styleClass="slds-m-around--x-large">
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavBulkProcesses}" pageLabel="{!$Label.stgNavRollupBatch}" />
+        <div class="slds-text-body--small slds-m-around--medium">{!$Label.stgHelpBatchOppRollup}</div>
+        <c:UTIL_PageMessages />
+        <div class="slds-grid slds-grid--align-center slds-grid--vertical-align-center slds-p-around--large">
+            <apex:commandButton value="{!$Label.stgBtnRunBatch}" status="statusLoad" action="{!runBatch}" disabled="{!isRunningBatch}" immediate="true" rerender="form" styleClass="slds-button slds-button--small slds-button--neutral" />
+        </div>
+        <apex:outputPanel rendered="{!isRunningBatch}">
+            <c:UTIL_JobProgressLightning id="jobProgress"
```
davidhabib: So something is wrong with the job progress component.  Frequently when I tried running settings batches, the page would reload (as expected) and show the outline of the widget, but it would be stuck empty with the animated cursor, until it finally (minutes later) displayed at 100%.  This was not consistent, and did not happen every time.  But I did reproduce it in most of the batch panels.  It may be we aren't getting the first call that we expect out of the batch for quite some time, but then we should change our widget so it isn't completely empty until it gets it first 5%.  I reproduced this both under LEX and Classic.  Here is an example of one that has been running now for 5 minutes!
image

src/pages/STG_PanelERR.page
```diff
@@ -1,8 +1,8 @@
 <apex:page controller="STG_PanelERR_CTRL">
```
davidhabib: while trying to test the chatter group typeahead, it didn't seem to match anything (but maybe I have no groups), but then if I saved with a bogus chatter group name, it would not save, but no error was displayed.

src/classes/STG_PanelUpdatePrimaryContact_CTRL.cls
```diff
@@ -0,0 +1,56 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT 
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2016
+* @group Settings
+* @description Controller for the Batch Rename Households panel.
```
davidhabib: wrong description for this class

src/pages/STG_PanelAllocations.page
```diff
@@ -1,113 +1,81 @@
...
     <style>
         input.lookupInput {
-            display: inline-block;
-            height: 34px;
-            padding: 6px 12px;
-            font-size: 14px;
-            line-height: 1.428571429;
-            color: #555555;
-            background-color: #ffffff;
+            background-color: white;
```
davidhabib: do we use this loookupInput css anywhere else in settings?  If so, can we create a shared settings css file to use?

src/pages/STG_PanelAddrVerification.page
```diff
@@ -42,152 +42,143 @@
...
+            });
         });
     </script>
+    <style>
+        div.slds-text-body--small ol {
+            list-style: initial;
+            list-style-type: decimal;
+        }
+    </style>
+    <apex:form id="form" styleClass="slds-m-around--x-large">
```
davidhabib: I thought when using SLDS, one needed an outer <div class="slds">, which is what I've used in the pages I've SLDS'ed so far.

src/pages/STG_PanelHouseholds.page
```diff
@@ -106,228 +106,182 @@
...
-    } 
-    
-    .bootstrap .form-control-column input[type="text"].txtOther {
-        margin-top: 12px;    
+        border-radius: 4px;
     }
-    
-</style>
 
+    .bootstrap .slds-form-element__control input[type="text"].txtOther {
```
davidhabib: I thought we removed bootstrap.  should .bootstrap be removed from this CSS?

src/pages/STG_PanelMembership.page
davidhabib: i'm guessing its because the listbox within the STG_DataBoundMultiSelect component's id is different than what we expect.  We thought it would be slMRT, but I wonder if we need to give an id to the component like "idFoo", and then use for = "idFoo.slMRT"

src/pages/STG_PanelTDTM.page
```diff
@@ -1,96 +1,88 @@
...
+                        <th scope="col"><label class="slds-truncate">{!$ObjectType.Trigger_Handler__c.Fields.Active__c.Label}</label></th>
+                        <th scope="col"><label class="slds-truncate">{!$ObjectType.Trigger_Handler__c.Fields.Asynchronous__c.Label}</label></th>
+                        <th scope="col"><label class="slds-truncate">{!$ObjectType.Trigger_Handler__c.Fields.Load_Order__c.Label}</label></th>
+                        <th scope="col"><label class="slds-truncate">{!$ObjectType.Trigger_Handler__c.Fields.User_Managed__c.Label}</label></th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <apex:repeat var="th" value="{!listTH}" >
+                        <tr>
+                            <td><div class="slds-text-body--small">{!th.Object__c}</div></td>
```
davidhabib: this table becomes quite wide.  should we consider using xxsmall text size instead?

src/pages/STG_SettingsManager.page
```diff
@@ -80,161 +76,146 @@
...
-        border-bottom: 0px;
-        border-right: 4px solid transparent;
-        border-left: 4px solid transparent;
-    }
-
-    li .caret {
-        float: right;
-        width: 0;
-        height: 0;
+        */
```
davidhabib: it looks like this whole upper section of CSS terminating on this line is commented out.  If so, shall we remove it?

davidhabib:
@njjc I've finished going through all the code and pages (the pages review was cursory, given the amount of markup changes.  i felt my time testing the pages was better spent than looking at page diffs).  Let me know when you have tackled Health Check, Error Log, and addressed the other comments, and I'll sync my branch and do more UI testing to confirm.  Thanks!

src/pages/STG_PanelMembership.page
njjc: Yeah, this is an issue with using the multiselect component: the id of the select list has to be static within the component. Nothing we can do if we want to still use a component, although I feel like that component has very limited value and we might as well stop using it at some point.

davidhabib: that component makes it easy to deal with our properties that are stored as a series of semicolon separated strings.  it allows you to use the component just like it was a native inputField without the client having to parse on load and compose on save, so i think it actually saves a bit of code and headache in each place it is used.

davidhabib:
@njjc found some issues:

- in Household page, it is too long and truncates at the bottom.  worse, in edit mode you can't even get to the household object settings section, nor see the save & cancel buttons.

- in Address page, in edit mode, if you choose Smarty Streets, the Mass Verify Existing Addresses section is truncated at the bottom of the page and you can't see its verify button, nor the save & cancel buttons.

- Relationship Reciprocal Settings also truncates at bottom in edit mode.

- Opportunity Names page, the table is truncated on the right, rather than scrolling or wordwrapping

- I notice inconsistent save & cancel buttons.  on Membership, save and cancel have space between them.  On Payments, they are a button group and combined.

- Donor Statistics gets clipped at the bottom in Edit mode.

- Contact Roles gets clipped in both modes

- so I think I figured out what's going on with the batch progress indicator.  First I did Rollup Donations, and now it correctly shows 4 progress indicators.  I did notice, however, that they jump around in order while they are updated.  After they finished, I went to the Rollup Allocations Batch page and started it.  Now it appears hung with an empty progress area.  But what's weird is when I clicked back on the Rollup Donations page, it had updated to show one progress indicator, and it showed that ALLO_Rollup_Schedule had completed!  so our progress indicator instances are conflicting with eachother.  trying any of the additional batches, and they have the same behavior; they appear stuck but the original displays that they have finished!

- Health Check. the buttons for passed and failed now appear at the very top of the box.  I think when you showed it to me, they were centered vertically, and it looked much better.

- Error Log page is truncated at the bottom.  my example had 6 errors on one page, but I can only see 3.25 of them.

- Error Notifications page is always too wide in edit mode, even when you make the browser wider, it's controls get too wide.  
image

- Trigger Configuration columns overwrite each other if the page isn't real wide. 
image

davidhabib:
@njjc checked the latest.  good improvements.  but I see one minor issue with the recent fix:

image

I can obviously live with that problem, but I'm not comfortable merging the pull request until we figure out what's going on with the batch process indicator.  if we don't fix it, we are going to get a lot of people posting on the hub that their batches are hung.  have you spoken with @dospromptman about this issue?  maybe he has some input on the cause and solution.  thanks.

njjc:
@davidhabib I believe all the truncation issues are fixed, as well as the side scrolling issue. Let's review together when you're back before final merge.

dospromptman:
The issue with the job progress component was that it was referencing certain elements that it needed to target based on html id.  In the settings page, the job progress component gets added to the page, but then hidden using CSS (not actually removed), and then another job progress component added elsewhere in the markup.  This was causing elements with duplicate ids to be added to the page, and the second component instance was accessing the elements intended for use by the first component instance.  I have updated the job progress component to use a wrapper element with a unique visualforce generated id around the elements it needs to target, and locates those elements only as children of that unique wrapper element.  A pull request is open on this branch with that fix.

dospromptman:
As for the reördering of the jobs in the list as they are running:  I have seen this behavior last night when jobs were sitting in Holding and Queued states before they moved to Processing state.  Today, I wasn't able to see the re-ordering because the jobs would start and then finish almost immediately.

Interestingly, I am sorting the queried jobs by their CreatedDate, so this is very strange that they would be returned in a different order each time.  There may be a different bug elsewhere that is causing the change in order.  I will have to investigate more.

PR #2007, opened 2016-03-21
BUG: 'Find Matching Gifts' & 'Manage Soft Credits' button - Attempt to de-reference a null object 

### LINKED ISSUE #1977: BUG: 'Find Matching Gifts' & 'Manage Soft Credits' button - Attempt to de-reference a null object
If the Amount is null, both buttons will product 'Attempt to de-reference a null object' Visualforce error. Regardless of type of Account or any other factor. Does not happen if Amount = 0

Hub thread: https://powerofus.force.com/0D58000002XPJPL

mrbelvedere:
Put in a check in the init/constructor for opp.Amount == null, and used custom labels to show a message on the page saying the amount can't be null. I did not write negative unit tests for the opp amount == null.

# Contribution Info

This external contribution was submitted by @schieco and Fixes #1977

# Critical Changes

# Changes

# Issues


davidhabib:
we've merged in the fix.  thanks.

PR #2011, opened 2016-03-21
BUG: Lead Conversion page does not respect Lead Process Status selected values

### LINKED ISSUE #1922: BUG: Lead Conversion page does not respect Lead Process Status selected values
Thread: https://powerofus.force.com/0D58000002UgML0

mrbelvedere:
Incomplete work done during the 2016 NPSP Sprint.  Still TODO:

complete coding in the lead convert controller (the status getter property), also need some logic to skip the metadata check if the object has no record type.

Remove unneeded financialforce files such as the VF pages and controllers and the JSZip components. 

# Contribution Info

This external contribution was submitted by @davidchengIC and Fixes #1922

# Critical Changes

# Changes

# Issues


davidchengIC:
@njjc  I've run into an obstacle.  The Metadata API does not expose record-type-specific values for special picklist fields like Lead Status.  I think it is because LeadStatus is an object like CampaignMemberStatus.  I can't find any other methods or APIs to expose the right data to fix the bug.

Not sure how to proceed here.  Do you want to close out the pull request?

judisohn:
Thanks @davidchengIC - I think @cdcarter found something similar when investigating this? Too bad, as I know it's a challenge that would be great to solve.

@njjc @davidhabib See note above - okay to close this or do you have other ideas?

davidhabib:
Thanks @davidchengIC for trying to get a solution here!

PR #2024, opened 2016-03-30
1st stage of LEX Manage Household UI

davidhabib:
This 1st stage includes:
- a visualforce page which just hosts a lightning component
- 3 lightning components: HH_Container, HH_Canvas, and HH_ContactCard
- the visualforce page can be bound to a action/button just like the current Manage Household page
- the page correctly loads, displays all contacts in Cards, and displays a set of fields on the Household
- the page supports drag/drop to re-order the Cards and update the Household Name and greetings
- Contact exclusion checkboxes are supported
- Household auto-naming checkboxes are supported
- correctly saves all changes to both contacts and the household 
- untested against Household objects at this point, only Household accounts
- there is a separate work item to figure out using custom labels


davidhabib:
@dospromptman @njjc Here is the first pass at the new LEX Manage Household UI Page.  this pull request is into a REVIEWED branch (not Dev)!  Not sure if you both would like to review this to learn more about Lightning Components, or just one of you.  I'm happy to give a code walk through so you aren't looking at it blind.  Let me know how you two want to proceed.  thanks!

PR #2026, opened 2016-03-30
Updating SLDS.resource to version 1.0.0

njjc:


davidhabib:
@njjc @jlantz why isn't this pull request getting marked as a build pending or passing (or failing)?

njjc:
@davidhabib looks like it was in process of building when the dev flow happened, and it hat to wait until that one was done to rebuild.

PR #2031, opened 2016-04-06
support field set of Household fields on the visualforce page

davidhabib:


davidhabib:
@dospromptman Once this finishes build, it is ready for review and merge.  It contains the work we looked at today.  As to some of the issues raised today:
1. no currency symbol.  can't fix.  this is behavior of inputField and you see it on the old page as well.
2. no percent symbol.  same as #1.
3. checkbox styling.  I couldn't get it to look like SLDS ones, which uses the outer label trick.
4. use util_pagemessages component.  done.
5. ui:message styled for SLDS.  I couldn't get it to look like SLDS, because this component renders multiple html tags and complex structure.  I think its good enough for now, and hopefully in a future release they will style it appropriately.
6. ui:buttons not styled with SLDS.  fixed (required css overrides in the component)

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -122,23 +123,28 @@
...
-		var self = this;
-        var countSuccesses = 0;
 		action.setCallback(this, function(response) {
 			var state = response.getState();
 			if (component.isValid() && state === "SUCCESS") {
                 countSuccesses++;
-                if (countSuccesses == 2)
-	                self.close(component);
+                if (countSuccesses == 2) {
+                    saveVisualforce();
```
dospromptman: Since this method is declared outside of the LC, then if this LC is used on any page that doesn't also declare this method, then this call will fail.  This means the HH_Container LC is not generally re-useable.  This may be okay, depending on intended usage.  What do you think, @davidhabib?

davidhabib: definetly ok.  we have no plans to make the whole container hostable.  maybe in the future we might decide to make the canvas and cards hostable, but not the full container.

PR #2043, opened 2016-04-12
Feature/1975 org camp mem

### LINKED ISSUE #1974: Deleting Honoree Contact causes unexpected exception
Nulling out the Honoree Contact field on Opportunity causes an TDTM_Opportunity exception and prevents saving a record if error handling is turned off:

1. NPSP Settings-->System Tools-->Error Notifications-->Disable Error Handling checked.
2. Create an Opp with Honoree Contact set to point to a Contact, save the Opp record.
3. Edit the Opp to null out the Honoree Contact field.  Try to save.  Error that prevents the record from being saved:

Error:Apex trigger npsp.TDTM_Opportunity caused an unexpected exception, contact your administrator: npsp.TDTM_Opportunity: execution of BeforeUpdate caused by: System.NullPointerException: Attempt to de-reference a null object: (npsp)

If you uncheck "Disable Error Handling", then using the steps above, the record can be saved correctly.

### LINKED ISSUE #1975: If Contact's Account is Organization Account, Campaign Member functionality does not work
Even though our best practice model is that all Contacts have a Household Account and Organizations are through Affiliation, the Campaign Member functionality should still work regardless of Account Record Type (if not, we should explicitly say this in documentation).

Steps to reproduce:

1. Create Organization Account. Create Contact directly in that Account.
1. NPSP Settings set to defaults with Campaign Member functionality enabled
2. Add Contact to Campaign, Status = Sent
3. Create Opportunity from Contact, Associate to same Campaign
4. Result: Campaign Member Status remains at Sent instead of updating to Responded as expected

If Contact not already member of Campaign, they are not added.

njjc:
# Warning
# Info
# Issues
Fixes #1974 
Fixes #1975


njjc:
@dospromptman Ready for review.

First commit fixes #1974 and adds a test.
Second commit fixes #1975. The reason why this bug was happening is a bit complicated. OPP_OpportunityContactRoles_TDTM tries to keep the Opportunity.Primary_Contact__c field in sync with the primary Opportunity Contact Role. In some cases, such as the "New Opportunity" button on the contact's related list, an opportunity contact role is created behind the scenes after Before triggers and before After triggers, in a way we can't replicate in test code. That means the Primary Contact field isn't populated on Opportunity AfterInsert, and we have to use the contact value from the OpportunityContactRole in that case.

The fix doesn't include any test changes since there's no way for us to have an OCR exist in an Opportunity After Insert context.  Let me know if you want to meet to discuss this further.

src/classes/OPP_CampaignMember_TDTM.cls
```diff
@@ -96,34 +106,68 @@ public class OPP_CampaignMember_TDTM extends TDTM_Runnable {
...
+    /*******************************************************************************************************
+    * @description Adds a Contact Id, Campaign Id, and Status to our three Lists. If the opportunity has a
     * status it is used, otherwise a default is used based on if the opportunity is closed/won or not.
     * @param Opp The current opportunity.
-    * @return void This method adds to three lists and returns nothing.
+    * @return void This method adds to three Lists and returns nothing.
     ********************************************************************************************************/
     private void addOppMemberFields(Opportunity opp) {
-        if (opp.Primary_Contact__c == null || opp.CampaignId == null)
+        if (opp.CampaignId == null || opp.Primary_Contact__c == null) {
```
dospromptman: I can't figure out why swapping these would make a difference.  Are there some implications I'm missing, here?

dospromptman:
I mostly get it, but there is still some stuff I don't understand about the explanation for #1975.  Is there a BeforeInsert/BeforeUpdate trigger somewhere that populates Primary_Contact__c from related OpportunityContactRoles?  And thats why the OCR being created after BeforeInsert means that the Primary_Contact__c field is not populated on new Opportunities?

src/classes/OPP_CampaignMember_TDTM.cls
njjc: Ha, no, I was in process of changing that method but ended up reverting it backwards.

PR #2045, opened 2016-04-12
Update STG_PanelOppNamingBatch.page

### LINKED ISSUE #2044: Bulk Data Process - "Refresh Opportunity Names" has wrong header text.
The header text on the visual force page reads "Bulk Data Processes - Rollup Donations Batch"

image

mpusto:
# Warning
# Info
# Issues
Fixes #2044


dospromptman:
I tried 6 times to push this to an org to test the update visually in a browser, but ran into the Label Bug each time.  In this case, the normal workaround for the label bug, which is to manually edit the page and replace the label, is somewhat counter productive since the update to the label is the very thing I'm trying to verify.

I have reviewed the code and it looks good to me.  This should get looked at after packaging for beta, when it goes through QA.

PR #2046, opened 2016-04-12
Updated inline help labels

jvarese:
Changed Latest Start/End Date descriptions to distinguish these fields from the Seasonal Address fields, which also appear at the bottom of this same interface.


dospromptman:
Pushed to org and verified that the updated inline help text labels appear correctly on the Add Address from related list layout.

Interestingly, @jvarese, the string was added to the XML metadata with an unencoded apostrophe character, but Salesforce tolerated this just fine, even though when it generates these files it will replace apostrophes with HTML entities.  Presumably that means apostrophe characters are okay to edit directly into meta XML files, although this is only proven for inline help text on objects, not custom labels.

PR #2069, opened 2016-05-04
Feature/lex manage household

davidhabib:
@dospromptman This pull request is ready for review.  it includes support for Address Management in the Manage Household UI.  It is made up of one new component, HH_AddressMgr, that is hosted by the HH_Container component.  It provides the following:

- displays the current household address in a tile
- provides a button in this tile to change the household address
- a modal popup form that displays all addresses used by the household, contact, and address objects
- the user can choose one of these existing addresses, and choosing the save address button pushes it to all contacts and the household (in memory).
- the modal popup form also provides a section for entering a new address instead of selecting one of the current addresses.
- on Save of the page, the code just saves the address changes to the contacts and the household (object or account).
- our Address Management triggers do the magic work of creating an Address object (if address mgmt is enabled), and updating each contact's address lookup.
- the component honors the address override checkbox on a contact, and will leave their address alone if it is set.

other non-lightning changes:

- changes to address management code's dealing with state & country picklist maps to always store upper case labels to compare against.  this was so the user typing 'california' would match even though they did not capitalize the state or country correctly.

- removing the SLDS resource loading from the pageMessage component, so it doesn't interfere with the consumer's version of SLDS.

- updated the HH Account  page layout and Manage UI field set to use the Household Phone rather than standard phone field.

Unfortunately, gitHub says this pull request cannot be automatically merged, probably because both branches contains the new SLDS static resource.  I'm not sure how to work around that.  I'm hoping you know some git magic!


src/aura/HH_AddressMgr/HH_AddressMgr.cmp
```diff
@@ -0,0 +1,159 @@
...
+    
+    <!-- markup follows... -->
+    <div class="slds-grid slds-p-around--x-small">
+        <div class="slds-card slds-p-around--none">
+            <div class="slds-card__body slds-m-around--x-small slds-p-vertical--none" >
+                <!-- address -->
+                <aura:if isTrue="{!v.addrDefault != null}">
+                    <div class="slds-text-body--small" >
+                        <ui:outputText value="{!v.addrDefault.MailingStreet__c + 
+                        	(v.addrDefault.MailingStreet2__c != null ? '\r' + v.addrDefault.MailingStreet2__c : '')}" /><br/>
```
dospromptman: I almost never see `\r` used for breaking a line.  It is usually `\n` where a character is needed to break the line.

Presumably this works as is, here, but I'm curious as to why you used `\r` instead of `\n`?  Did `\n` not work?

src/aura/HH_AddressMgr/HH_AddressMgr.cmp
```diff
@@ -0,0 +1,159 @@
...
+                        <ui:outputText value="{!v.addrDefault.MailingCountry__c}" />                                     
+                    </div>
+                </aura:if>
+            </div> 
+            <div class="slds-card__footer">
+                <div class="slds-grid slds-grid--align-center">
+                    <ui:button class="slds-button slds-button--small slds-button--neutral" 
+                               aura:id="btnChangeAddress" 
+                               buttonTitle="Change address" 
+                               label="Change address"
```
dospromptman: Do we have or should we have a story to go back and update all strings to use custom labels in the manage household components?

Part of that task will be having to figure out how to deal with the namespace resolution issue with LC labels, unfortunately.

src/aura/HH_AddressMgr/HH_AddressMgr.cmp
```diff
@@ -0,0 +1,159 @@
...
+                            <h4 class="slds-section__title-action" onclick="{!c.toggleAddrSection}">
+                                <c:svg class="slds-button__icon"
+                                       xlinkHref="/resource/SLDS/1_0_3/assets/icons/utility-sprite/svg/symbols.svg#switch" />
+                                Enter a New Address
+                            </h4>
+                        </div>
+                        <div class="slds-section__content">                                                    
+                            <div class="slds-form--stacked">
+                                <div class="slds-form-element">
+                                    <label class="slds-form-element__label" for="inputSample2">Mailing Street</label>
```
dospromptman: You may want to remove the `for` attribute of the labels in this section, since they all refer to the same non-existent element.  Alternatively, add unique ids to the input elements and use those ids for these attributes.  This is non-breaking, so probably optional to change.

src/aura/HH_AddressMgr/HH_AddressMgr.css
```diff
@@ -0,0 +1,11 @@
+.THIS li.selected {
+    background-color: token(colorBackgroundRowSelected);
```
dospromptman: Can you help me understand this?  Default design tokens are in developer preview for Spring '16, yes?  Are they released in Summer '16?  Are you running in a GS0 org, or is the developer preview enabled for everyone?  The docs say the list of default tokens for the developer preview is subject to change, but this wouldn't be the first time we aimed at a moving target.

What considerations need to be made to use tokens in this feature?

src/aura/HH_AddressMgr/HH_AddressMgrController.js
```diff
@@ -0,0 +1,55 @@
...
+
+	saveChangeAddress : function(component, event, helper) {
+        helper.setDefaultAddress(component);
+        component.set('v.showChangeAddressPopup', false);
+	},
+
+	onPressAddrTile : function(component, event, helper) {
+        var iAddr = Number(event.currentTarget.getAttribute('data-iAddr'));
+        component.set('v.iAddrSelected', iAddr);
+        component.set('v.listAddr', component.get('v.listAddr'));
```
dospromptman: What does assigning the value of `v.listAddr` to `v.listAddr` accomplish?

src/aura/HH_AddressMgr/HH_AddressMgr.cmp
davidhabib: I couldn't remember what the character was and \r popped in my head, and it obviously worked!  I later put the correct \n in related helper code (another component).  I'll change this to \n.

src/aura/HH_AddressMgr/HH_AddressMgr.cmp
davidhabib: yes I already have that backlog item in place for next sprint.

src/aura/HH_AddressMgr/HH_AddressMgr.cmp
davidhabib: probably needed for screen readers?  obviously a copy paste from sample code I missed!  I'll update it.

src/aura/HH_AddressMgr/HH_AddressMgr.css
davidhabib: I'm in a dev org I've had for years.  not GS0.  can't tell what server because I've got custom domains enabled!  I needed a way to get the default selected item color, and tokens is where I found it.  since we are using our own static resource for SLDS, I think we are safe from changes, until we upgrade of course.

src/aura/HH_AddressMgr/HH_AddressMgrController.js
davidhabib: I needed it to redisplay the list of addresses so the one I just selected will get redrawn selected.  I didn't think just setting iAddrSelected would be enough to trigger the refresh.

davidhabib:
@dospromptman OK, I committed a few minor changes based on your feedback.  so it's good to merge once the build confirms.  How do we deal with the merge conflict issue with SLDS?  Do we need to back it out of the pull request (and how does one do that?)

dospromptman:
The weird thing is there didn't seem to be a merge conflict when I performed the merge on my local machine.  Not sure why Github thought there was a conflict...

PR #2073, opened 2016-05-06
Manage HH Delete Button perms

cdcarter:
Only render Manage HHA delete button if the users has the permissions to delete households.


davidhabib:
@cdcarter thanks for the fix, but we are currently working on replacing the entire Manage Household UI page with a Lightning Component based one, which won't include a delete button in its ui.   Delete can be performed from the standard Salesforce ui.

cdcarter:
No worries, I figured that work was in progress, but wanted to alert the issue so at least the bug wouldn't be propagated further. It came up in office hours today. Thanks for the status.

PR #2075, opened 2016-05-09
Replacing deprecated Limits calls

njjc:
Replacing calls to Limits.getFutureCalls() and Limits.getLimitFutureCalls() with Limits.getAsyncCalls() and Limits.getLimitAsyncCalls() respectively.


njjc:
@dospromptman this is passing build, should be good to merge now.

Issue #1679, opened 2015-08-11
Spanish translations need correction & updating


judisohn:
Hub thread: https://powerofus.force.com/0D58000002EHfgV

- Affiliation package almost all translated, missing new fields and a bug needs to be fixed
- Households is partially translated. 
- Recurring Donations is partially translated. 
- Contacts & Organizations is partially translated. Payment fields, payment related fields in opportunities and other opportunity fields are not translated. 
- Nonprofit Starter Pack is not translated (No translations for Batch, Addresses, Settings, etc)
- Relationship Package is not translated

In this document you will see the fields that are missing for the first packages (and the Bug I mention in the Affiliation package)
https://docs.google.com/spreadsheets/d/1gbeRyUF1xFLwmdg3l1PAN-hfL37CoX2dmpEgZ9uCMV4/edit?usp=sharing

judisohn:
**lurch: add

judisohn:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000001fgvAAA" target="_blank">W-008201</a>

mrbelvedere:
Included in production release version 3.73

PR #2076, opened 2016-05-10
Spanish translation import

dospromptman:
Filtered translations from file `Bilingual_es_2016-04-14 0147.stf` and imported translations specific to NPSP namespace.

# Contribution Info

These translations were contributed and reviewed by a group of volunteer translators, including Laura Diaz, Elena Inurrategui, Rosario Gonzalez Moron, Raúl González García, Mariana Mazzeo, and María Florencia Salvador . ¡Gracias a todos!

# Critical Changes

# Changes

Spanish translations have been added to NPSP package.

# Fixes
[Issue 1679](https://github.com/SalesforceFoundation/Cumulus/issues/1679)

### DIFF
--- a/src/objectTranslations/DataImport__c-es.objectTranslation
+++ b/src/objectTranslations/DataImport__c-es.objectTranslation
@@ -18,42 +18,34 @@
         <relationshipLabel>NPSP Data Imports (Cuenta1 Importada)</relationshipLabel>
     </fields>
     <fields>
-        <help>Cuenta1.BillingCity</help>
         <label>Cuenta1 Ciudad</label>
         <name>Account1_City__c</name>
     </fields>
     <fields>
-        <help>Cuenta1.BillingCountry</help>
         <label>Cuenta1 País</label>
         <name>Account1_Country__c</name>
     </fields>
     <fields>
-        <help>Cuenta1.Name</help>
         <label>Cuenta1 Nombre</label>
         <name>Account1_Name__c</name>
     </fields>
     <fields>
-        <help>Cuenta1.Phone</help>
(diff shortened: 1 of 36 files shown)


src/objectTranslations/DataImport__c-es.objectTranslation
```diff
@@ -0,0 +1,543 @@
...
+        <label>Cuenta1 Estado de importación</label>
+        <name>Account1ImportStatus__c</name>
+    </fields>
+    <fields>
+        <label>Cuenta1 Importada</label>
+        <name>Account1Imported__c</name>
+        <relationshipLabel>NPSP Data Imports (Cuenta1 Importada)</relationshipLabel>
+    </fields>
+    <fields>
+        <help>Cuenta1.BillingCity</help>
```
davidhabib: these translations of the help text will break the NPSP data importer.  the help text is used to specify which object and field the import should map to.  So Cuenta1 needs to stay as Account1.  the easiest fix is to just strip out all help text entries for this translation file.

dospromptman: Great catch!

I have removed the translated help text from fields where the help text was a field name.

There are other translations for fields where I think it might be valuable to include the translated help text, though.  I kept some of the translated help strings in this file.  Can you look it over again and let me know if any of those would affect the data import process?

For reference, the fields where there is translated help text are:
* `ApexJobId__c`
* `Donation_Donor__c`
* `Donation_Record_Type_Name__c`
* `FailureInformation__c`
* `ImportedDate__c`
* `Status__c`

PR #2078, opened 2016-05-10
Feature/lex manage household

davidhabib:
this includes 

- minor changes to make the HH_ManageHH page shareable by both Account and Household objects.

- conversion of all hardcoded strings to use custom labels


davidhabib:
@jlantz @dospromptman @njjc Why is this pull request showing diffs for all of the label work that Nic checked in today.  Since pushes to dev the push back into all feature branches, shouldn't this diff between two feature branches not show those items?

davidhabib:
@jlantz also, how do we make a build for this, or is that not supported?

davidhabib:
@dospromptman While I find the idea of doing our own value provider intriguing, given its lack of documentation, I'm wary of investing too much time trying to get this to work at this time.  So I'd like to move forward with my simple hack for now.  Thus this pull request is ready for review and merge.  thanks.

davidhabib:
@dospromptman please review and merge today.  i have additional features I want to submit, but I want to wait for this branch to get merged first before adding to it.  thanks.

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -339,6 +339,20 @@
...
         }
         component.set('v.listCon', listCon);
-    }
-    
+    },
+
+    /*******************************************************************************************************
+	* @description fixup custom labels exposed thru $Label, so the ones from the wrong namespace are null.
+    */
+    fixupCustomLabels : function(component) {
```
dospromptman: Can you some comments inside the method explaining what this is doing?  What is the expected structure of the return value of `$A.get('$Label')`?  What is the `startsWith()` check for?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -37,6 +37,28 @@
 public class HH_Container_LCTRL {
     
     /*******************************************************************************************************
+    * @description trick the packaging spider to pickup references to custom labels only used by our
+    * lightning components.
+    */
+    public static void packageLabels() {
```
dospromptman: This looks like it only references labels that you created anew for this work?  Do you think it should contain references to ALL labels that you use?

The reason it might be better to do that is there is no guarantee that those labels are or will always be referenced elsewhere.  Other completely unrelated code could change or be removed and that would break this.

src/classes/HH_ManageHH_CTRL.cls
```diff
@@ -37,18 +37,72 @@
...
+    * @description ActionMethod to save the Household and close the page.
+    * @return PageReference the original calling page's URL.
     */
-    public HH_ManageHH_CTRL(ApexPages.StandardController controller) {
-        hh = controller.getRecord();        
-        hhId = hh.Id;
-    }    
+    public PageReference save() {
+        update hh;
+        return new PageReference(retURL);
```
dospromptman: I have had code like this flagged by Security Review, before.  I don't really know how the exploit works, but the general guidance given is to add code that forces or guarantees that the URL is a relative URL, or (even better) that the URL is on a known white list of safe URLs.

Supposedly parameter named 'retURL' is supposed to be filtered by Salesforce, but advice is not to rely on this.

https://developer.salesforce.com/page/Secure_Coding_Arbitrary_Redirect
http://salesforce.stackexchange.com/questions/51946/should-i-use-this-solution-to-remove-open-redirect-vulnerability-on-cancel

src/pages/HH_ManageHousehold.page
```diff
@@ -1,180 +1,5 @@
...
-<apex:inputField value="{!hhO.npo02__MailingPostalCode__c}" styleClass="gwHidden sfField sfField-MailingPostalCode__c" />
-<apex:inputField value="{!hhO.npo02__MailingCountry__c}" styleClass="gwHidden sfField sfField-MailingCountry__c" />
-
-<apex:pageBlock id="hhDetails" rendered="{!hhO.id!=null}" title="{!$Label.npo02__HouseholdDetails}">
-
-<apex:pageBlockSection columns="2">
-
+<apex:page standardController="npo02__Household__c" extensions="HH_ManageHH_CTRL" title="{!$Label.npo02__ManageHousehold}" standardStylesheets="false" showHeader="true" docType="html-5.0"  >
+        
+	<apex:include pageName="HH_ManageHH"/>
```
dospromptman: What is the difference between the `HH_ManageHH` and `HH_ManageHousehold` pages?  Where is one used vs the other, etc?

dospromptman:
Ok, we should revisit the label hack sooner rather than later, though.  Hopefully they fix the issue soon and we won't need our own workaround anymore.

dospromptman:
All the other cruft in this PR makes it really hard to figure out what to review.  I looked at everything, but skipped things that looked unrelated. Hopefully I didn't miss anything.

src/classes/HH_Container_LCTRL.cls
davidhabib: I only include those labels that reside in the npsp package.  the other labels are in the npo02 package, and including them here will do nothing, since they don't live in our npsp package.

dospromptman: Yes, but this list does not contain even every NPSP label you reference.  For example, `stgBtnCancel` or `bdiBtnClose`.  It also contains references to labels you don't use, like `lblStreet` (you use a label just called `Street`).

src/pages/HH_ManageHousehold.page
davidhabib: HH_ManageHH is the new visualforce page that wraps the lightning component.
HH_ManageHousehold is the old page for use with Household Objects.  It can now be just a wrapper to the shared HH_ManageHH page.
HH_ManageHHAccount is the old page for use with Household Accounts.  It will also become a wrapper over HH_ManageHH, but I am not ready to replace it yet since I still want to bring up the old ui for testing.

dospromptman: So `HH_ManageHH` is intended to be included from those two pages eventually, but right now it is only included from `HH_ManageHousehold`.  Is that accurate?

davidhabib: yes


src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: done

src/classes/HH_Container_LCTRL.cls
davidhabib: ok, i added the missing other references to npsp labels.  not sure why thought I wasn't using lblStreet, which i see still used.  I didn't encounter any markup to $Label.c.Street.

src/classes/HH_ManageHH_CTRL.cls
davidhabib: I changed it to just make the page reference from '/' + hhId, which achieves what I need and avoids this security issue.

davidhabib:
@dospromptman  OK, I've checked in the changes from the code review.  I don't understand why gitHub is showing diffs with work from other accepted pull requests that are in both branches (spanish localization, etc).  Anyways, hopefully you are able to merge this.  thanks.

src/classes/HH_Container_LCTRL.cls
dospromptman: https://github.com/SalesforceFoundation/Cumulus/pull/2078/files/afb6a9f7ab92cbcce58c8259500251e8f36e8dda#diff-06ad8a1650536fafd5d1c9efc19c965bR105 refers to `$Label.c.Street` and `$Label.npsp.Street`.  Other labels in this section refer to similarly named labels.  Maybe this were supposed to be `$Label.npsp.lblStreet`, etc.?

davidhabib: I think you are looking at an intermediary diff.  the pull request is using lblStreet, lblCity, etc.  just search in this pull request to confirm.

dospromptman: I'm looking at the diff that appears when I click the 'Files Changed' tab at the top of this pull request. What the heck??  I wonder how much of what I reviewed is out of date, then...

Why is the diff on the PR so messed up?

dospromptman: I'm going to pull this branch down and review the diff on my local machine.  Will merge after I re-review.

dospromptman:
Git was, once again, reporting that there were merge conflicts when none existed.  Something is really up with this PR....

I merged it locally.

PR #2083, opened 2016-05-13
Modify spanish translation for opp picklist value

dospromptman:
The spanish translation for the `Opportunity.Acknowledgement_Status`
picklist value 'To Be Acknowledged' was inconsistent with the spanish
translation for the same value on the
`npe01__OppPayment__c.npsp__Payment_Acknowledgement_Status` picklist
field.

This change updates the `Opportunity` picklist value to use the same
translation as the `npe01__OppPayment__c` value.


dospromptman:
@mpusto this is the change you requested via the Google Sheets comment discussion

mpusto:
Gracias @dospromptman!

PR #2088, opened 2016-05-19
Feature/lex manage household

davidhabib:
this pull requests adds support for adding an existing contact to the household, or merging a household and all of its contacts.  there is a little bit of stub code to test this functionality that will be removed when we hook up the page to use the autocomplete component.


davidhabib:
@dospromptman this is ready for your review.  gitHub no longer thinks it can't merge it, and the zillions of extra diffs are gone! yay!  maybe your deleting the branch on the server had a good side affect!

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -161,17 +161,39 @@
...
+        var callbacksWaiting = 4;
+        
+        // first need to merge any households (Accounts only) before we save contacts
+        // so we avoid deleting a household if that contact was the last one in the hh.
+        var listHHMerge = component.get('v.listHHMerge');
+        var hh = component.get('v.hh');
+        if (listHHMerge != null && listHHMerge.length > 0) {        
+            var action = component.get("c.mergeHouseholds");
+            action.setParams({ hhWinner : hh, listHHMerge : listHHMerge });
+            action.setCallback(this, function(response) {
```
dospromptman: At this point you have an identical callback function defined 4 times within this method.  Can you extract it out into its own definition (being careful to maintain closure of the `callbacksWaiting` variable) and just reference it in each `setCallback()` call?

src/aura/HH_ContactCard/HH_ContactCard.cmp
```diff
@@ -18,7 +18,7 @@
                                    xlinkHref="/resource/SLDS/1_0_3/assets/icons/standard-sprite/svg/symbols.svg#contact" />
                         </div>
                         <div class="slds-media__body">
-                            <a href="#">{!v.contact.FirstName}&nbsp;{!v.contact.LastName}</a>
+                            <a href="{! '/' + v.contact.Id}" target="_blank">{!v.contact.FirstName}&nbsp;{!v.contact.LastName}</a>
```
dospromptman: I saw your Chatter post regarding this, so I know you were having issues without the `target="\_blank"` attribute.  But, does using this attribute work in LEX?  I thought LEX gets upset when you try to pop it open in a new window.

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -427,4 +446,145 @@
...
+        action.setParams({ hhId : hhMerge.Id });
+		var self = this;
+		action.setCallback(this, function(response) {
+			var state = response.getState();            
+			if (component.isValid() && state === "SUCCESS") {
+                var listConMerge = response.getReturnValue();
+				var listCon = component.get("v.listCon");
+                
+				// move all contacts into our household
+				var cExisting = listCon.length;
```
dospromptman: could `listCon` be undefined?

src/classes/HH_Households_TDTM.cls
```diff
@@ -239,7 +239,7 @@ public without sharing class HH_Households_TDTM extends TDTM_Runnable {
             
             if (householdnameupdates.size() > 0){
                               
-                if (currentHouseholdsSettings.npo02__Async_Household_Naming__c == true)
+                if (currentHouseholdsSettings.npo02__Async_Household_Naming__c == true && !system.isFuture() && !system.isBatch())
```
dospromptman: Is this to fix a bug unrelated to the manage households UI?  If so, do you think it is better to apply it directly to dev in a separate PR, rather than tie it up with the UI work?

src/aura/HH_Container/HH_Container.cmp
```diff
@@ -195,6 +207,53 @@
...
+                            <div class="slds-x-small-buttons--horizontal">
+                                <ui:button class="slds-button slds-button--neutral" 
+                                    aura:id="btnCancelHHMerge"
+                                    buttonTitle="{!$Label.c.stgBtnCancel + $Label.npsp.stgBtnCancel}" 
+                                    label="{!$Label.c.stgBtnCancel + $Label.npsp.stgBtnCancel}" 
+                                    press="{!c.cancelMergeHHPopup}" />
+                                <ui:button class="slds-button slds-button--neutral" 
+                                    aura:id="btnAddContact"
+                                    buttonTitle="{!format($Label.c.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName) + 
+                                                 format($Label.npsp.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName)}"
```
dospromptman: I can't find `lblBtnAddContact` in the metadata or referenced in `HH_Container_LCTRL`.  Can you check to see if this label is being packaged and also double check the other labels?

src/aura/HH_Container/HH_Container.cmp
```diff
@@ -195,6 +207,53 @@
...
+                                <ui:button class="slds-button slds-button--neutral" 
+                                    aura:id="btnAddContact"
+                                    buttonTitle="{!format($Label.c.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName) + 
+                                                 format($Label.npsp.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName)}" 
+                                    label="{!format($Label.c.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName) + 
+                                                 format($Label.npsp.lblBtnAddContact, v.conAdd.FirstName + ' ' + v.conAdd.LastName)}" 
+                                    press="{!c.doAddContact}" />
+                                <ui:button class="slds-button slds-button--brand" 
+                                    aura:id="btnMergeHH"
+                                    buttonTitle="{!$Label.c.lblBtnAddAllHHMembers + $Label.npsp.lblBtnAddAllHHMembers}"
```
dospromptman: I can't find `lblBtnAddAllHHMembers` in the metadata or referenced in `HH_Container_LCTRL`.  Can you check to see if this label is being packaged and also double check the other labels?

dospromptman:
The indentation in this is all over the place.  The chaotic spacing combined with most if/else clauses omitting braces seems like playing with fire.

Can you update the spacing in these files so the indentation is consistent, and also make sure that a mix of tabs and spaces are not being used?

src/aura/HH_ContactCard/HH_ContactCard.cmp
davidhabib: This works correctly in LEX and Classic in Spring '16.  It was failing in Summer '16 (I think under both LEX and Classic), but testing again this morning it now works in Summer '16.  I'm guessing they refreshed GS0 last night!!

src/aura/HH_Container/HH_Container.cmp
davidhabib: good catch

src/aura/HH_Container/HH_Container.cmp
davidhabib: i've now added the two missing labels (not yet pushed to gitHub).

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: how can a separate shared method for the callback use callbacksWaiting without having to make it an attribute of the component, which I would not want to do?

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: no.  it is passed in when the component is instantiated, and everywhere it is used without additional checks to see if it is null.  even if they delete all members of the household or there are no members originally, it will be an array with no members.

src/classes/HH_Households_TDTM.cls
davidhabib: it is an existing bug I ran into while developing Manage Household UI.  since it has been around forever (since 2.x), I felt it wasn't critical to get in sooner, and could just go in with my work and save the time of creating yet another branch, etc.

davidhabib:
uhg.  so I developed all of these files using the Developer Console, which unfortunately doesn't have an option to save tabs as spaces, or even to specify how many spaces = 1 tab.  Guess I'll have to find a tool to clean up these files.  what a pain.

src/aura/HH_Container/HH_ContainerHelper.js
dospromptman: You can assign a function to a variable local to the save method.  I created an example of how this could be done and tagged you in it.

davidhabib:
OK @dospromptman I believe I have addressed all the issues you brought up.  Please merge away.  thanks.

dospromptman:
Github still erroneously reports merge conflict, even when there is none.  I performed the merge on my local machine.

PR #2093, opened 2016-05-25
Refactor HH_ContainerHelper callback

dospromptman:
@davidhabib I think something like this would work to refactor the callback into its own definition that can be reused.  This probably wouldn't merge cleanly because of other work you're doing on the file, but this is an example if you wanted to use it.


davidhabib:
ah, so nesting the function within the function allows it to access vars from the parent.  didn't know that.  javascript is so different in some areas than other languages I've used!  after I fix the tabs/spaces issue, I'll try this out.

davidhabib:
manually put in my branch.

PR #2096, opened 2016-05-26
Scripts for processing STF translation files

dospromptman:
This is a set of scripts for processing an STF (Salesforce Translation File) exported from an org that has the NPSP managed package installed.

The translations exported from that org will have namespaced "keys" to indicate which metadata each translation belongs to.  Additionally, translations for both the npsp namespace as well as metadata belonging to underlying packages will be included in this export.

These scripts are intended to:
* Extract the translations from the exported STF file containing namespaced keys for all packages,
* Filter out any keys that correspond to metadata that we know we never want to translate,
* For the remaining keys, identify the namespace to which the translated metadata belongs,
* Rewrite the key to be compatible with unmanaged code belonging to that namespace,
* Write the translations, with rewritten keys, out to separate files based on the org where each file is to be imported.


dospromptman:
@jlantz Can you look this over and give me any feedback about any possible issues with the code, and any other feedback?

I realize the organization of these scripts is probably somewhat unusual, so if you have better ideas on how to organize it, let me know.  Pipelining these together like this just made sense to me.  Also, I avoided using modules like `argparse` or `logger` to keep things "simple", but I'm not sure if it really ended up being any simpler.  Any feedback you have about how I could better go about this task would be appreciated!

scripts/stf/filters.json
```diff
@@ -0,0 +1,43 @@
+[
```
dospromptman: @kbromer @mpusto @davidhabib Can you please review this list of keys that are to be filtered out from the raw export file?

This is copied over from the Google sheet, more or less, but I also added 'Scontrol' related keys, since including those will cause the import to fail (the Scontrols do not and can not exist in unmanaged deployments of the underlying packages)

scripts/stf/filters.json
```diff
@@ -0,0 +1,43 @@
...
+    "PicklistValue.Contact.npe01__PreferredPhone.Other",
+    "PicklistValue.Contact.npe01__PreferredPhone.Work",
+    "PicklistValue.Contact.npe01__Preferred_Email.Alternate",
+    "PicklistValue.Contact.npe01__Preferred_Email.Personal",
+    "PicklistValue.Contact.npe01__Preferred_Email.Work",
+    "PicklistValue.Contact.npe01__SystemAccountProcessor.One-to-One",
+    "PicklistValue.Contact.npo02__SystemHouseholdProcessor.All Individual Contacts",
+    "PicklistValue.Contact.npo02__SystemHouseholdProcessor.All New or Edited Contacts",
+    "PicklistValue.Contact.npo02__SystemHouseholdProcessor.No Contacts",
+    "PicklistValue.Opportunity.npsp__Matching_Gift_Status.Potential",
```
davidhabib: why did you exclude these.  the 20 places in the code that looks at status use the hardcoded strings, which is what should be internally stored for the picklist values.  so having the user translate the localized version of the picklist value should be fine since the non-translated value is what gets saved for the picklist.  If I'm wrong, then my whole understanding of picklist values is off!

dospromptman: I removed all translations that were marked in red in the translation review spreadsheet.  Maybe the comments there will help? https://docs.google.com/spreadsheets/d/1Z9HsXtt4ioIVFm1D5CAnLcWWYSMgxCzeeJnYVBtVubI/edit

I have not tested the behavior myself, and I'm not sure if the master label for the picklist value gets used in the SOQL compare or what.

davidhabib: @dospromptman @kbromer see my comment in the spreadsheet.  i think the matching gift status picklist values can be localized, because the query should be looking at the underlying picklist value, not its label.  no?

kbromer: Added a comment to the spreadsheet - I don't think so, I think because your query is hardcoded to the english string, this will fail

kbromer: We could support it with a custom label of course, the Matching Status Value or whatever

davidhabib: so you are saying that so

dospromptman: As per your Chatter post, I have removed the 3 Matching_Gift_Status picklist translations from the filters list.

cpolcyn:
Verified using supplied STF file. Could be refactored into a single file for ease of use, so I created a work item in AA.

PR #2105, opened 2016-06-06
Feature/lex manage household

davidhabib:
@dospromptman this branch is ready for review (still queued up in the build).

This pull requests includes:
- your autocomplete components
- HH wrappers on autocomplete components where needed
- small set of enhancements to core autocomplete (we discussed)
- integration of HH_AutoComplete into Manage Household UI


davidhabib:
@dospromptman build passes so ready for review and merge!

src/aura/HH_AutoComplete/HH_AutoCompleteController.js
```diff
@@ -0,0 +1,23 @@
+({
+    handleNewContact: function(component, event, helper) {
+        var con = {'sobjectType': 'Contact'};
+        var autocomplete = component.find('autocomplete');
```
dospromptman: You're using what the user searched for as the default for the name of their new contact?  Clever.  This is one of those subtle usability things that you go out of your way to do for users but that they probably never notice except subconsciously.

src/aura/HH_AutoComplete/HH_AutoCompleteController.js
```diff
@@ -0,0 +1,23 @@
...
+        var name = autocomplete.get('v.displayValue');
+        var list = name.split(' ');
+        if (list.length == 0) {
```
dospromptman: I don't think this case is possible.  If there was just one word, it would be the only element in the array (handled in next case)

davidhabib: I wanted to protect for name being an empty string.  not sure what split would return then.

src/aura/HH_AutoComplete/HH_AutoCompleteController.js
kbromer: 👍

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -437,12 +454,34 @@
     * @description sees if the contact is NOT the lone member of a household, in which case it will bring up
     * our Merge Household popup.  if they are the lone member, it will just proceed with the HH merge.
     */
-    addOrMergeContact : function(component, conAdd, hhMerge) {
+    addOrMergeContact : function(component, event) {
+        debugger;
```
dospromptman: I know this is wip, and does have some debugging code elsewhere, but was this statement intended to be included?  If so, maybe mark it (and other debug code) as `// @debug` so we can find it later and remove it.

src/aura/autocomplete/autocomplete.cmp
```diff
@@ -3,28 +3,31 @@
 
     <aura:handler name="init" value="{!this}" action="{!c.doInit}"/>
     <aura:handler name="optionSelected" event="c:autocompleteSelectListOption" action="{!c.handleOptionSelected}"/>
+    <aura:handler name="change" value="{!v.isListVisible}" action="{!c.handleIsListVisibleChange}"/>
```
dospromptman: Instead of using a change handler to react to changes of `isListVisible`, how about just using the attribute directly in the class string of the menu div?  `<div class="{!'slds-lookup__menu' + (isListVisible ? '' : ' slds-hide')}" ... >`

This way no change handler is needed.  `helper.setListVisibility` could be changed to just set the attribute, instead of manually toggling the css class.

Are there downsides to doing it like that?  It seems better than adding a change handler, to me, but what do you think?  I'm not sure if Aura will rerender the whole menu div if the class attribute needs to be updated, or if it will just update the attribute in the DOM...

src/aura/autocomplete/autocompleteHelper.js
```diff
@@ -9,14 +9,15 @@
...
-        if (el === document.activeElement) {
+        if (component.get('v.inputFocused')) {
             component.set('v.keyword', keyword);
 
             if (!keyword) {
                 return;
             }
 
             this.setListVisibility(component, true);
+			component.set('v.showLoadingIndicator', true);
```
dospromptman: those dang tabs :(

(also on #L30)

src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
```diff
@@ -0,0 +1,84 @@
...
+    */
+    @AuraEnabled
+    public static List<ProviderResult> queryObjects(String queryValue, list<Contact> listCon) {
+        try {
+            // '%', '_', and '\' all are special characters in LIKE syntax.
+            // We escape them here by prepending those characters with a backslash
+            String preparedQueryValue = '%' + queryValue.replaceAll('([%_\\\\])', '\\\\$0') + '%';
+            
+			// use helper to get all contact fields to support custom household naming and greetings.
+            String strSoql = HH_Container_LCTRL.conSelect;
```
dospromptman: Since these fields are being sent to the client, is it necessary to perform FLS checks on the field list?

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -59,6 +59,11 @@
         action = component.get("c.getAddresses");
         action.setParams({ hhId : hhId, listAddrExisting : null });
         action.setCallback(this, function(response) {
+
+		    // tell our visualforce page we are done loading
```
dospromptman: https://www.youtube.com/watch?v=SsoOG6ZeyUI

davidhabib: @kbromer you must watch this video, and then tell us why you still want us to use spaces over tabs!!

kbromer: Picked one, liked the one that allowed maximum control. :)  But, for those with the left-pinky twitch <ahem>:

https://atom.io/packages/tabs-to-spaces
http://stackoverflow.com/questions/407929/how-do-i-change-eclipse-to-use-spaces-instead-of-tabs
https://www.sublimetext.com/docs/2/indentation.html

dospromptman: The real sin is mixing tabs and spaces... which Developer Console does in the most illogical way with no way to change the behavior.

src/aura/autocomplete/autocomplete.cmp
davidhabib: sounds good.

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: removed.

src/aura/autocomplete/autocompleteHelper.js
davidhabib: damn you dev console!

src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
davidhabib: i have a separate backlog item, that I just moved into this sprint, to add FLS where needed in the Manage Household UI.

davidhabib:
ok @dospromptman I've just updated in response to your feedback.  ready for merge.  thanks.

PR #2107, opened 2016-06-07
Address State is not resolved because of this typo

ahusanovic:
This issue is making problems with resolving of State addresses using Google Maps API. Someone misspelled tag that we get from Google and that was not resolving this issue.

Please can you we push this change to the most recent upgrade.


davidhabib:
Thank you for the fix!

ahusanovic:
Hello David,

When do you think that this change might reach the production environment? I have never contributed before with NPSP therefore I had to ask it ☺

Thank you,
Alen

davidhabib:
we are just ending our sprint today, so normally this would be released to sandboxes tomorrow, and product in one week.  But since we are in the Summer '16 upgrade window, we won't be making a release until after Summer '16 is out to all customers, so I think the last week of June.  @jlantz correct me if I got the dates wrong.

ahusanovic:
Thank you David so much!! Looking forward to see these changes on Sandbox environment.

PR #2112, opened 2016-06-09
Feature/lex manage household

davidhabib:
@dospromptman this pull request adds apex tests for all apex code used in the Manage Household UI lightning implementation.


davidhabib:
@dospromptman didn't realize you hadn't merged this branch yet.  It now also includes my FLS and permissions work.  Please review and merge (preferably today if possible).  test are in process.  thanks!

src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
```diff
@@ -53,6 +53,7 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
             String preparedQueryValue = '%' + queryValue.replaceAll('([%_\\\\])', '\\\\$0') + '%';
             
             // use helper to get all contact fields to support custom household naming and greetings.
+            // note that our helper does FLS, so we don't have to do it here.
```
dospromptman: I find it very confusing to use property getters this way.  You're using the conSelect property like a function call.. a function call that takes implicit arguments from global scope, rather than being passed in, even.

I urge you to consider changing the conSelect property to a method call with a descriptive name.  Maybe even something like `getContactSelectQueryWithFls()`?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -115,6 +142,7 @@ public class HH_Container_LCTRL {
             if (hhId == null)
                 return null;
 
+			// conSelect does the FLS check
```
dospromptman: I find it very confusing to use property getters this way.  You're using the conSelect property like a function call.. a function call that takes implicit arguments from global scope, rather than being passed in, even.

I urge you to consider changing the conSelect property to a method call with a descriptive name.  Maybe even something like `getContactSelectQueryWithFls()`?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -80,31 +83,55 @@ public class HH_Container_LCTRL {
...
+            for (String strF : listStrFld) {
+                DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(strObject, strF);
+                if (dfr == null || !dfr.isAccessible() || !dfr.isUpdateable())
+                    throw (new permsException(string.format(label.flsError, new string[]{strObject + '.' + strF})));
+                
+                strSoql += strComma + strF;
+                strComma = ',';
             }
+            
+            strSoql += ' from ' + strObject + ' where Id = :hhId';
```
dospromptman: Do you also need to add a CRUD check on the object?  Or is checking the fields enough?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -142,12 +170,24 @@ public class HH_Container_LCTRL {
...
                 list<string> listStrField = getContactFields();
                 string strSoql = 'SELECT ';
-                string strComma = '';
+                string strComma = '';                
+                set<String> setCheckUpdate = setUpdateableContactFields();
+                
                 for (string strF : listStrField) {
+
+                    // check field level security
+                    DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe('Contact', strF);
```
dospromptman: Is it possible for `strF` to be a relationship field?  For example, `Owner.Name` or something like that?  Would `UTIL_Describe.getFieldDescribe('Contact', 'Owner.Name')` fail?

src/classes/UTIL_Describe.cls
```diff
@@ -369,4 +369,26 @@ public class UTIL_Describe {
...
+    * @param listStrF the list of field names to check FLS and copy data from
+    * @return void
+    */
+    public static void copyObjectFLS(string strObject, SObject src, SObject dst, list<String> listStrF) {
+        for (String strF : listStrF) {
+            DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(strObject, strF);
+            if (dfr == null || !dfr.isCreateable() || !dfr.isUpdateable())
+                throw (new permsException(string.format(label.flsError, new string[]{strObject + '.' + strF})));
+                
+            dst.put(strF, src.get(strF));
```
dospromptman: `sObject.get()` and `sObject.put()` are fine as long as `strF` is not a relationship field

src/pages/HH_ManageHH.page
```diff
@@ -62,20 +62,30 @@
...
+                                     namespacePrefix : nsPrefix },
                                    "lightning",
                                    function(cmp) {
                                        // after lightning is loaded, do what we need to do here!
                                        
                                        // hook up our event handlers
-                                       $A.eventService.addHandler({ "event": "c:HH_HouseholdSavedEvent", "handler" : saveAndCloseVisualforce});
-                                       $A.eventService.addHandler({ "event": "c:HH_ContactAfterDeleteEvent", "handler" : contactRemoved});
-                                       $A.eventService.addHandler({ "event": "c:HH_ContainerLoadedEvent", "handler" : doneLoading});
+                                       $A.eventService.addHandler({ "event": nsPrefixEvent + ":HH_HouseholdSavedEvent", "handler" : saveAndCloseVisualforce});
```
dospromptman: does `c:eventName` not work here because it is being called from the `createComponent()` callback, or does `c:eventName` just not work for managed packages?  If it needs to be namespaced, I've got some updates to do elsewhere...

src/classes/HH_Container_LCTRL.cls
davidhabib: checking fields is enough.  if the object permissions aren't allowed, then any field check will also fail.

src/classes/HH_Container_LCTRL.cls
davidhabib: I originally did code parsing for object.field and then realized that we were not needing to support lookup field references in any of the use cases in manage household ui, so I removed it to keep the code simpler.

src/classes/UTIL_Describe.cls
davidhabib: agreed.  in my use cases, we are copying in order to do an update, and one cannot update lookup field references.

src/pages/HH_ManageHH.page
davidhabib: since this is javascript in the visualforce page, there will be no fixup/handling by salesforce of c:fooEvent, so I'm assuming we have to manually use the correct namespace at runtime.

src/classes/HH_Container_LCTRL.cls
davidhabib: agreed.  I copied and modified this from some older code in NPSP.  I'll update it to a function.

davidhabib:
@dospromptman OK, I've updated with your good suggestion about replacing conSelect with getContactSelectQueryWithFls().  I believe I've answered your other questions.  If you are ok with those changes, then I'd like you to merge this into the REVIEWED branch.  I realize it won't pass builds currently due to the permissions issue.  Since this is not going into Dev, there is no harm.  And this will allow me to continue work on other Manage HH UI work while we wait for the permissions issue to get resolved with Jason Lantz.  Thanks.

PR #2118, opened 2016-06-15
Feature/jen actionplans

jjbennett:


README.md
```diff
@@ -3,15 +3,15 @@
 #### For Users
 
 * <a href="https://powerofus.force.com" target="_blank">Ask questions or get help</a>
-* <a href="https://github.com/SalesforceFoundation/Cumulus/issues/new" target="_blank">Log a confirmed Issue or Feature Request</a>
+* <a href="https://github.com/SalesforceFoundation/Cumulus/issues/new" target="_blank">Log a confirmed Issue</a> or <a href="https://powerofus.force.com/hub-ideas" target="_blank">Feature Request</a>
```
kbromer: Thanks for changing these!

njjc: Looks like these were Judi's changes to dev, not sure why they're not in both of these branches.

njjc:
@jjbennett You'll also need to include the additions to the custom labels file; try retrievePackagedToSrc and see if it pulls them down correctly. If there are any label deletions in there, you won't want to commit those. Once the new labels file is correct, do another updatePackageXml, commit, and push. Let me know if you get stuck.

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+/**
+* @author Salesforce.org
+* @date 2016
+* @group Engagement Plans
+* @description Trigger Handler on Engagement Plan Tasks 
+*/
+public class EP_TaskDpendencyCheck_TDTM extends TDTM_Runnable {
+
+    /** @description maximum number of tasks to check in a dependency chain */
+    public static Integer maxCount = 500;
```
njjc: This can be private since it won't be accessed outside of this class.

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+    * @param listNew the list of Engagement Plan Tasks from trigger new. 
+    * @param listOld the list of Engagement Plan Tasks from trigger old. 
+    * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.). 
+    * @param objResult the describe for Engagement Plan Tasks 
+    * @return dmlWrapper.  
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> listNew, List<SObject> listOld, 
+        TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+        /*********************************************************************************************************
+        * @description A list of the tasks that require validation.
```
njjc: I don't think apexdoc handles this type of comment inside a method. For inline comments, we generally just use //

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /*******************************************************************************************************
+    * @description For the Tasks that are new or where the EP Template or Dependency has changed,
+    * prevent cross dependencies and mismatched templates
+    * @param listPSCNew the list of PSCs from trigger new. 
+    * @param listPSCOld the list of PSCs from trigger old.
+    * @return void.  
+    ********************************************************************************************************/
+    public static void validateTasks(List<Engagement_Plan_Task__c> listToValidate) {
```
njjc: This method can also be private.

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+        // A set of the Engagement Plan Template Ids related to the tasks.
+        Set<Id> templateIds = new Set<Id>();        
+        // Engagement Plan Task mapped to the template and all the template's related tasks
+        Map<Id, Engagement_Plan_Template__c> mapTaskIdToTemplate = new Map<Id, Engagement_Plan_Template__c>();
```
njjc: I would make this a map from the template id to the template object, since each task has a master-detail to the template, it will end up holding less objects in memory and be easier to grock.

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+        // Engagement Plan Task mapped to tasks that are dependent on it to check for template mismatch
+        Map<Id, List<Engagement_Plan_Task__c>> mapTaskIdToChildren = new Map<Id, List<Engagement_Plan_Task__c>>();
```
njjc: Similarly, I would just make this map Task Id to Task Object; you'll only have to hold each task object in memory once, and you can still follow the dependency path by doing map.get(epTask.Dependent_Task__c);

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,161 @@
...
+            if (mapTaskIdToChildren.containsKey(childTask.Dependent_Task__c)) {
+                mapTaskIdToChildren.get(childTask.Dependent_Task__c).add(childTask);
+            } else {
+                List<Engagement_Plan_Task__c> childTasks = new List<Engagement_Plan_Task__c>();
+                childTasks.add(childTask);
+                mapTaskIdToChildren.put(childTask.Dependent_Task__c, childTasks);
+            }
+        }
+
+        for (Engagement_Plan_Task__c newEPTask : listToValidate) {
```
njjc: I'm having a bit of trouble following the dense logic here. Some inline comments would be great, and I think simplifying the map structure as suggested above might clear this up a bit. Let me know if you want to have a call to discuss. Thanks!

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,155 @@
...
+
+                if ((newEPTask.Dependent_Task__c != null
+                    && newEPTask.Dependent_Task__c != oldEPTask.Dependent_Task__c)
+                        || newEPTask.Engagement_Plan_Template__c != oldEPTask.Engagement_Plan_Template__c) {
+                        listToValidate.add(newEPTask);
+                }
+            }
+            if (listToValidate.size() > 0) {
+                validateTasks(listToValidate);
+                System.debug('List to validate: ' + listToValidate);
```
njjc: Please remove debug statements before the final review.

njjc: In cases when you actually want a debug statement in production code, use the UTIL_Debug class instead.

src/classes/EP_TaskDependencyCheck_TDTM.cls
```diff
@@ -0,0 +1,170 @@
...
+    * query for any child tasks to verify the templates are the same
+    * @param listToValidate the list of EP Tasks that require child dependency validation. 
+    * @return void.  
+    ********************************************************************************************************/
+    private static void validateChildDependencies(List<Engagement_Plan_Task__c> listToValidate) {
+        // Engagement Plan Task mapped to tasks that are dependent on it to check for template mismatch
+        Map<Id, Engagement_Plan_Task__c> mapIdToTaskWithChildren = new Map<Id, Engagement_Plan_Task__c>(
+            [SELECT Id, 
+                (SELECT Id, Dependent_Task__c, Engagement_Plan_Template__c
+                FROM Action_Plan_Tasks__r)
```
njjc: Sorry, this is a pain, but would you mind changing the child relationship name to start with Engagement_ instead of Action_?

src/classes/EP_TaskDependencyCheck_TEST.cls
```diff
@@ -0,0 +1,155 @@
...
+            
+    /*******************************************************************************************************
+    * @description Insert a single Engagment Plan Task Dependent on an existing task.
+    */ 
+    static testmethod void testSingleEPDependentTask() {
+        List<Engagement_Plan_Template__c> templateTests = UTIL_UnitTestData_TEST.createEPTemplates(1);
+        insert templateTests;
+        List<Engagement_Plan_Task__c> taskTests = UTIL_UnitTestData_TEST.createEPTasksForTemplates(2, templateTests);
+        insert taskTests;
+        Engagement_Plan_Task__c taskTest = new Engagement_Plan_Task__c(
```
njjc: I have validation in my code to ensure that each Engagement_Plan_Task__c has one and only one lookup populated, so your tests will have to conform to that as well; you can create a dummy contact/account and set Account__c or Contact__c to on the EP Task.

njjc: Sorry, never mind, I was talking about the Engagement_Plan__c object. EP Task is fine. Please disregard the last comment :)

src/classes/EP_TaskRollup_TEST.cls
```diff
@@ -0,0 +1,201 @@
...
+            
+    /*******************************************************************************************************
+    * @description Insert a single Task related to an Engagement Plan
+    */ 
+    static testmethod void testSingleEPTask() {
+        List<Engagement_Plan_Template__c> templateTests = UTIL_UnitTestData_TEST.createEPTemplates(1);
+        insert templateTests;
+        List<Engagement_Plan_Task__c> epTaskTests = UTIL_UnitTestData_TEST.createEPTasksForTemplates(1, templateTests);
+        insert epTaskTests;
+        Engagement_Plan__c epToInsert = new Engagement_Plan__c(
```
njjc: Comment above does apply here though; one Engagement_Plan__c lookup needs to be populated to pass my validation.

njjc:
@jlantz any idea what this "[limit exceeded](https://salesforce-org.atlassian.net/builds/download/NPSP-FEAT44-JOB1/build_logs/NPSP-FEAT44-JOB1-19.log)" error in the deploy is about?

src/labels/CustomLabels.labels
```diff
@@ -573,6 +573,22 @@
...
+        <shortDescription>EPTaskDependency</shortDescription>
+        <value>dependency recursion is bad</value>
+    </labels>
+    <labels>
+        <fullName>EPTaskTemplate</fullName>
+        <categories>Engagement Plan, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>EPTaskTemplate</shortDescription>
+        <value>EP Tasks can only be dependent on other EP Tasks that belong to the same EP Template.</value>
```
njjc: Could you write out Engagement Plan Tasks for the error message?

src/labels/CustomLabels.labels
```diff
@@ -573,6 +573,22 @@
         <value>omit</value>
     </labels>
     <labels>
+        <fullName>EPTaskDependency</fullName>
+        <categories>Engagement Plan, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>EPTaskDependency</shortDescription>
+        <value>dependency recursion is bad</value>
```
njjc: Ha, this error message was a joke but I like that you took it at face value. How about: "Engagement Plan Task dependencies cannot make recursive dependency loops." For these simple edits to text strings, you can actually edit inline in github and it will create a commit for you.

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -84,6 +86,34 @@ public class UTIL_UnitTestData_TEST {
         
         return openStage;
     }
+
+    public static String getClosedTaskStatus(){
```
njjc: Generally we add to this file when the methods are broadly applicable across the test suite, I think these Task related methods will be used only for the Engagement Plans feature, so I might move them to my EP_Utils class once it's merged in.

Issue #1388, opened 2015-02-17
BUG: When an opportunity is created via Lead Conversion, duplicate primary contact roles are generated for the oppty


kbromer:
Replication instructions: 
Create a lead with a company of 'self'.  Click 'Convert', uncheck the 'Do not create Opportunity on Conversion' box, click 'Convert', note two Contact Roles for the new Contact/Oppty are created, both primary, one with the NPSP-defined default role, the other blank. 

https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=00580000007UuJC&tab=sfdc.ProfilePlatformFeed&fId=0D58000001nILd5&emtm=1424188398094&s1nid=0DB800000004Fa7&OpenCommentForEdit=1&fromEmail=1&emkind=chatterCommentNotification&s1ext=0&s1oid=00D00000000hWLM&s1uid=00580000003egB0

davidhabib:
@kbromer I cannot reproduce, and I see that our lead convert code has logic to deal with this.  but in the hub thread, you mention that you finally could reproduce it.  what is the secret?  thanks.

davidhabib:
@kbromer Never mind.  I see that I did fix it in March!

kbromer:
Way to go past-Dave!

pankaj90:
@davidhabib  I just encountered same issues with NPSP 3.45
After converting a lead and creating an Opportunity on conversion causes duplicate contact roles to be created with both marked as primary and role as blank value. Please help.

AshwinPandole:
Receiving the same issue in my org with version 3.45
We have a demo scheduled with customer shortly. Can you please assist.
Thanks a lot.

davidhabib:
several people are reporting still seeing this issue, so reopening so we can investigate again.  For those on this lists still seeing this issue, please provide details:
1. which version of NPSP?
2. which account model?
3. converting to a new contact or merging an existing contact?
4. does it happen for all lead conversions, or only some?

davidhabib:
[...]

PR #2120, opened 2016-06-16
#1388.  When an opportunity is created via Lead Conversion, remove dup OCRs

davidhabib:
## Issues
#1388


davidhabib:
@njjc this simple fix is ready for review and merge.  let me know if you want to actually talk about the full history of this.  this is definitely a fix for something we can't reproduce, but many others have.

PR #2122, opened 2016-06-20
Feature/lex manage household

davidhabib:
@dospromptman This pull requests removes the old Manage Household UI code (apex controllers, javascript resource, static css), and converts the existing HH_ManageHHAccount page to use the shared HH_ManageHH page, just like we do for Household Objects.  Please review and merge.


src/classes/HH_ManageHH_CTRL.cls
```diff
@@ -41,22 +41,57 @@ public with sharing class HH_ManageHH_CTRL {
...
     */
     public HH_ManageHH_CTRL() {
+        this(null);
+    }
+    
+    public HH_ManageHH_CTRL(ApexPages.StandardController controller) {
         namespacePrefix = UTIL_Namespace.getNamespace();
 
         Map<String, String> pageParams = ApexPages.currentPage().getParameters();
         hhId = pageParams.get('Id');
```
dospromptman: Since you have `StandardController` passed in now, you could just do `controller.getId()`

PR #2123, opened 2016-06-20
Feature/lex manage household

davidhabib:
@dospromptman Oops.  a minor fix, and then tab cleanup.  thanks!


build.xml
```diff
@@ -118,7 +118,10 @@ If you are on a unix based system, the following commands should work:
         
     </target>
     <target name="postDeploy">
+        <!--
         <echo>Sleeping for 30 seconds to allow background jobs to complete</echo>
         <sleep seconds="30" />
+        -->
+        <antcall target="deployUpdatedAdminProfile" />
```
dospromptman: Where is this target defined?

davidhabib: this is from @jlantz commit to this branch, so he'll have to answer that.

PR #2124, opened 2016-06-21
Feature/bdi checkbox

### LINKED ISSUE #2101: Data Import: Contact True checkbox fields are updated to False in import
hub thread: https://powerofus.force.com/0D58000002jsKto

Reproduce:
1. Custom checkbox on Contact, check as True
2. Create custom NPSP Data Import field for checkbox, mapped in help to contact field
3. Create NPSP Data Import record, First/Last Name only, with a donation (donation probably not necessary)
4. Process, matching on First and Last Name
5. Result: Contact's custom checkbox is now False, shouldn't have been updated

davidhabib:
## Info
The NPSP Data Importer will now only update checkbox fields when they are checked in the Data Import record.

## Issues
Fixes #2101

### DIFF
--- a/src/classes/BDI_DataImport_BATCH.cls
+++ b/src/classes/BDI_DataImport_BATCH.cls
@@ -340,6 +340,28 @@ public class BDI_DataImport_BATCH implements Database.Batchable<sObject> {
         return mapDIFieldToObjField;
     }
 
+    /*******************************************************************************************************
+    * @description Copies non-null DI field values to the destination object, given the provided mapping.
+    * also explicitly ignores boolean false's, since we can't differentiate null from false.
+    * @param di The DataImport record
+    * @param strDIField The DI field to copy from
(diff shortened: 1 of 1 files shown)


davidhabib:
@njjc this bugfix is ready for review and merge. thanks.  Passed build too!

njjc:
On other non boolean fields with a null value, I take it we're never overwriting the value?
Do you think it's worth eventually considering a setting like the data loader's [insert null values setting](https://help.salesforce.com/apex/HTViewSolution?id=000005442)? (It's a bit of a misnomer since it really only applies on update.)

davidhabib:
correct.  we ignore null value fields.  I don't think we'll ever add that option since the amount of fields on the Data Import object is very large, so users will undoubtedly always have some they leave blank.

njjc:
Right, that makes sense. Thanks.

PR #2133, opened 2016-06-30
added LockerService detection and warning code.

davidhabib:
@dospromptman this adds a simple check for LockerService and displays a warning if its activated.  I tested this under my DE org where I can turn LockerService on and off.


davidhabib:
@dospromptman converted addresses in address mgr to objects, to avoid inconsistent handling of namespace prefixing by lightning code.

davidhabib:
@dospromptman converted other typed sobjects to also be objects to avoid other namespace prefixing issues.

davidhabib:
@dospromptman this commit also now includes changing the Delete Contact functionality to be Remove (from Household).

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -1,5 +1,14 @@
 ({
     /*******************************************************************************************************
+     * @description called at onInit to detect if Locker Service is enabled, warning if so.
+     */
+    warnOnLockerService: function(component) {
+        if (component.$lskey != null) {        
+            this.displayUIMessage(component, "LockerService is activated which prevents this page from working correctly.  Under Setup, search for Critical Updates, and deactivate the critical update 'Enable Lightning LockerService Security'. ", "divUIMessageContainer");
```
dospromptman: I know that this is temporary and hopefully very short lived, so maybe it doesn't make sense to put it in a label to localize it.  But, should we get doc review on this string?

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -279,12 +272,12 @@
         action.setCallback(this, callback);
         $A.enqueueAction(action);
 
-        // delete any contacts
-        var listConDelete = component.get("v.listConDelete");
-        listConDelete = this.addPrefixToListObjectFields(namespacePrefix, listConDelete);
-        action = component.get("c.deleteContacts");
+        // remove any contacts
+        var listConRemove = component.get("v.listConRemove");
+        listConRemove = this.addPrefixToListObjectFields(namespacePrefix, listConRemove);
```
dospromptman: adding prefixes still needs to be done here?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -327,7 +329,6 @@ public with sharing class HH_Container_LCTRL {
     * @description deletes the Contacts in the database
     * @param listCon The list of Contacts to delete
     * @return void
-    */
```
dospromptman: can't we just delete the method, or will it for sure be returning in the near future?

src/classes/HH_Container_TEST.cls
```diff
@@ -133,7 +133,6 @@ private class HH_Container_TEST {
 
     /*********************************************************************************************************
     * @description Tests deleteContacts()
-    */
     public static testMethod void testDeleteContacts() {
```
dospromptman: can't we just delete the method, or will it for sure be returning in the near future?

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: correct.  short lived, so no label was intentional. 

@jvarese are you ok with the message "LockerService is activated which prevents this page from working correctly.  Under Setup, search for Critical Updates, and deactivate the critical update 'Enable Lightning LockerService Security'."?  or do you have any edits you'd like to it?

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: Yes, we are doing an upsert, and if some of the field names are incorrectly namespaced,I assume we'd get an error.  best to be consistent and always correctly fixup the field names before any server call.

src/classes/HH_Container_LCTRL.cls
davidhabib: I wanted to keep it around in case PM decides to offer both types of functionality (delete and remove).  Mary and I had talked about maybe adding a backlog item for that (though I don't recall what we decided).

src/classes/HH_Container_TEST.cls
davidhabib: same as previous comment.  it may be needed.

src/aura/HH_Container/HH_ContainerHelper.js
dospromptman: I was confused because you removed the namespace prefix processing from `HH_AddressMgrHelper` but not here.  So when sending sObjects to the server, namespaces have to be added, but because you're using plain objects within the component, the framework isn't doing anything tricky to add prefixes to the objects you query or pass in?

Since this is intentional and it works, I'm fine with it... but I don't totally understand it :)

dospromptman:
I'm merging this with the doc string review outstanding, since that can easily be managed in its own PR

Issue #1941, opened 2016-02-11
Update manage allocations page to better handle multiple currencies


njjc:
Currently we rely on the user replacing a custom label to change a currency symbol. Instead, we should rely on salesforce to display the currency using an outputfield and a dummy opportunity with currencyisocode set.

njjc:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000001qwiAAA" target="_blank">W-009327</a>

kbromer:
Clever!

mrbelvedere:
Included in beta release version 3.76 (Beta 16) and higher

mrbelvedere:
Included in production release version 3.76

PR #2136, opened 2016-07-05
Adding basic currency locale awareness to Manage Allocations page

dospromptman:
Adding org/related record currency detection to Manage Allocations
controller, along with some support for ISO code to currency symbol
mapping.

Reformatting currency input fields and totals to use SLDS 'addon'
styling, and to dynamically show detected org/related record currency
symbol.

Adding utility class for currency handling.

Removing some cruft from Manage Allocations page.

## Info
Added basic currency locale awareness to Manage Allocations page.  Now, for some commonly used currencies, the correct currency symbol will be displayed.  For multi-currency orgs, the currency ISO code of the related opportunity, recurring donation, or campaign will be displayed. 

## Issues
#1941

### DIFF
--- a/src/classes/ALLO_ManageAllocations_CTRL.cls
+++ b/src/classes/ALLO_ManageAllocations_CTRL.cls
@@ -87,6 +87,9 @@ public with sharing class ALLO_ManageAllocations_CTRL {
     /** @description Row number sent back by page for add and delete row methods.*/
     public integer rowNumber {get;set;}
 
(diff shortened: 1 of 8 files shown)


dospromptman:
@chrismcmahon just a heads up: I have not run the tests to check, but this change *might* break the UI test for the manage allocations page.

https://github.com/SalesforceFoundation/NPSP-Browser-Tests/blob/master/features/npsp_manage_allocations.feature#L22

Issue #2082, opened 2016-05-12
Add Health Check test to detect duplicate contact roles


davidhabib:
to allow users to identify opps that might be double counted in rollups and soft credit rollups.  after we release this, we can then fix the double counting issue.

davidhabib:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003n4yUAAQ" target="_blank">W-010292</a>

davidhabib:
dupe of #2022

dospromptman:
@davidhabib, I don't think this is a dupe of #2022.  Was there a different ticket that contains additional details?

davidhabib:
Oops!  I think i mixed up bugs.  I believe #2081 and #2022 are the dupes of eachother!

mrbelvedere:
Included in beta release version 3.76 (Beta 16) and higher

mrbelvedere:
Included in production release version 3.76

PR #2137, opened 2016-07-05
Duplicate Contacts and Rollup Values Health Check

dospromptman:
Adding a health check to detect the case where an opportunity has the
same contact associated via more than one OpportunityContactRole.

The reason this health check is being added is because we are changing
the behavior of how partial soft credits are counted.  In the future, we
will only credit a contact associated with an opportunity once for that
opportunity, NOT multiple times based on how many times the contact is
associated with the opportunity.

This health check brings with it some labels for displaying error,
success, and solution messages.

# Warning

# Info
We've changed the way the NPSP treats soft credit Opportunity Rollup values for Contacts that are doubly associated with an Opportunity through two different Opportunity Contact Roles. In the past these rollup values were double-counted for the Contact; now they're only counted once. We've provided a new Health Check that lets you detect Opportunities that formerly would have rolled up soft credit values twice for these double-associated Contacts.

# Issues
#2082

### DIFF
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -682,23 +682,38 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
      */
      public void verifyNoDuplicateContactRoles() {
         try {
+            Set<String> softCreditRoles = RLLP_OppRollup_UTIL.softCreditRoles;
+
             List<AggregateResult> results = [
                 SELECT
                     COUNT(Id),
                     OpportunityId,
                     ContactId
                 FROM OpportunityContactRole
+                WHERE Opportunity.IsWon = true
+                AND Role IN :softCreditRoles
                 GROUP BY OpportunityId, ContactId
                 HAVING Count(Id) > 1
                 LIMIT 1
             ];
+
+            List<String> quotedSoftCreditRoles = new List<String>();
+
+            for (String role : softCreditRoles) {
+                quotedSoftCreditRoles.add('\'' + String.escapeSingleQuotes(role) + '\'');
+            }
 
             if (results.size() > 0) {
                 createDR(
                     Label.healthLabelDuplicateContactRoles,
                     statusWarning,
                     Label.healthDetailsDuplicateContactRoles,
-                    Label.healthSolutionDuplicateContactRoles
+                    String.format(
+                        Label.healthSolutionDuplicateContactRoles,
+                        new List<String>{
+                            String.join(quotedSoftCreditRoles, ', ')
+                        }
+                    )
                 );
             } else {
                 createDR(
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -182,6 +182,7 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
         verifyOppContactRoles();
         verifyTriggerHandlers();
         verifyNoMissingOppPayments();
+        verifyNoDuplicateContactRoles();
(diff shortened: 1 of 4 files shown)


dospromptman:
Please hold off on merging until @jvarese finishes string review

dospromptman:
@mpusto I can't think of anything else to add to the release notes section, even though this might require more explanation than what I have here.  Can you think of anything I should add to the release notes section for this health check?

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -673,5 +674,47 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+     public void verifyNoDuplicateContactRoles() {
+        try {
+            List<AggregateResult> results = [
+                SELECT
+                    COUNT(Id),
+                    OpportunityId,
+                    ContactId
+                FROM OpportunityContactRole
+                GROUP BY OpportunityId, ContactId
+                HAVING Count(Id) > 1
```
davidhabib: @dospromptman @mpusto I thought we would only complain about duplicate OCR's whose role was in a soft credit role.  If we don't, we might find that we are giving this error to lots of orgs who are using some custom non-credit OCR for some sort of tracking.  I think this soql just needs to add a where clause of Role in :softCreditRoles.

src/labels/CustomLabels.labels
```diff
@@ -2054,6 +2078,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>View these Contacts by running the &lt;b&gt;Contacts without Accounts&lt;/b&gt; report in the NPSP Health Check reports folder.</value>
     </labels>
     <labels>
+        <fullName>healthSolutionDuplicateContactRoles</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>The recommended solution for when the duplicate contact roles health check fails</shortDescription>
+        <value>You should expect these contacts to have different roll-up values than before, since duplicate contacts are no longer counted twice.  To discover duplicate opportunities and contacts in your org, you can run the following SOQL query in the Developer Console: &apos;SELECT Count(Id), OpportunityId, Opportunity.Name OpportunityName, ContactId, Contact.Name ContactName FROM OpportunityContactRole GROUP BY OpportunityId, Opportunity.Name, ContactId, Contact.Name HAVING Count(Id) &gt; 1&apos;</value>
+    </labels>
```
davidhabib: I think the message might want to be specific to soft credit ocrs only.  our normal rollups run against the ocr that is marked primary, so dupes aren't an issue for them.

src/classes/STG_PanelHealthCheck_CTRL.cls
dospromptman: I coded this straight to what the rollup code checks.  There is no check against role being 'Soft Credit' or any specific role at all in the rollup code, where the deduplication takes place.

Yes, calling code may decide to filter by soft credit roles, but the rollup code does not make that distinction, so there is no guarantee that the duplicate counting issue is limited to soft credit roles.

dospromptman: Wait, this might not be correct... let me verify

dospromptman: @davidhabib, you're correct.  I will add more filters to this query (and the displayed query).  I was confusing this with a check against `IsPrimary` on OCR (which is not explicitly checked in the rollup code).

dospromptman:
I have updated the health check to include a more specific query that matches the query performed in the rollup code (`WHERE Opportunity.IsWon = true AND Role IN :softCreditRoles`).  I have also updated the strings, as per @jvarese, so the string review is complete.

PR #2148, opened 2016-07-07
Feature/lex manage household

davidhabib:
@dospromptman these simple fixes to autocomplete get it working under LockerService

### DIFF
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -106,7 +106,9 @@
 
             // tell our visualforce page we are done loading
             var event = $A.get("e.c:HH_ContainerLoadedEvent");
-            event.fire();
+            //event.fire();
+            var vfEventHandlers = component.get('v.vfEventHandlers');
+            vfEventHandlers.HH_ContainerLoadedEvent(event);
 
             var state = response.getState();
             if (component.isValid() && state === "SUCCESS") {
@@ -225,7 +227,9 @@
                     // tell our visualforce page we are done saving, so it should save
                     // its data and close the page.
                     var event = $A.get("e.c:HH_HouseholdSavedEvent");
-                    event.fire();
+                    //event.fire();
+                    var vfEventHandlers = component.get('v.vfEventHandlers');
+                    vfEventHandlers.HH_HouseholdSavedEvent(event);
                 }
             } else if (component.isValid() && state === "ERROR") {
                 self.reportError(component, response);
@@ -367,7 +371,10 @@
             event.setParams({
                 "contact": con
             });
-            event.fire();
+            //event.fire();
+            var vfEventHandlers = component.get('v.vfEventHandlers');
+            vfEventHandlers.HH_ContactAfterRemoveEvent(event);
+            
         }
     },
 
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -473,16 +473,20 @@
             // for each namespace, and property for each label.
             var lbl = $A.get('$Label');
 
-            for (var str in lbl['c']) {
+            for (var str in lbl.c) {
                 // the labels that fail to get resolved appear as
                 // "$Label namespace.foo does not exist: Field $Label namespace__foo does not exist. Check spelling."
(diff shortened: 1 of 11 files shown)


davidhabib:
@dospromptman per Doug Chasman's design, no longer hooks up visualforce event handlers to the EventService, but passes them in an object to the container instead.

also includes several Lightning CLI fixes.

This branch is ready for review and merge.  thanks.

src/aura/autocomplete/autocomplete.cmp
```diff
@@ -77,6 +77,7 @@
                                   class="{!'slds-input ' + (and(v.showPills, !empty(v.value), !empty(v.displayValue)) ? 'slds-hide' : '')}"
                                   placeholder="{!v.placeholder}"
                                   value="{!v.displayValue}"
+                                  updateOn="keyup"
```
dospromptman: what does this change, and the corresponding change in doInit, fix?

davidhabib: you were hooking up the 'input' event which is not an official Lightning event.  thus one could not look at component.get('v.value') in the event handler; it wasn't set.  So first I changed it to be the 'change' event, but turns out Lightning only calls that when the textbox loses focus, rather than on each key.  So I had to go with keyup instead.

davidhabib: this was uncovered during LockerService testing.  with LS deactivated, it worked fine.

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -106,7 +106,9 @@
 
             // tell our visualforce page we are done loading
             var event = $A.get("e.c:HH_ContainerLoadedEvent");
-            event.fire();
```
dospromptman: This isn't exactly dchasman's pattern.  Doug was setting up the actions in the container app (in this case actions you would have defined in HH_ContainerController.js) to handle events fired (via attribute shortcut to addEventHandler in hosting visualforce page).  Those controller actions would delegate to the external methods passed in to the container app from the visualforce page.

Here, you're passing in the methods but foregoing the addEventHandler step altogether.  In which case, your methods aren't actually event handlers, they're just methods you're getting a reference to and then calling.  So, the whole $A.get('e.c:HH_ContainerLoadedEvent') etc is superfluous.  Just call the method, and pass in an argument if you need to pass something in.

Not sure if your more direct way has any other implications.

src/aura/autocomplete/autocomplete.cmp
dospromptman: well that is interesting. not sure why they would lock down the 'input' event.  but in any case, keyup is different than input.  keyup works if the user is using a keyboard to enter text.  it may or may not work if the user pastes text or text is entered in any other way.  have you tested pasting text?

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: Doug's pattern is basically, pass an object with callback functions to a lightning component, so that component can call your visualforce javascript code.  The fact that he put it on the app component, and I am putting it on my main Lightning component is a design decision I made to be more direct and consistent with my Lightning container component being the brains behind the whole page.  In fact, it was the one firing the three events that the visualforce page needed.  I decided to keep the same event prototype to be consistent with lightning events, so my visualforce handlers could also use the same event object to receive parameters.  So I want to continue with the design I have.

src/aura/autocomplete/autocomplete.cmp
davidhabib: if they paste with the mouse, then they must hit enter to get it to requery, which seems reasonable enough to me.

src/aura/HH_Container/HH_ContainerHelper.js
dospromptman: You have also removed the firing of the event, so there are no "handlers" of the event at all.  I'm okay if you want to keep it, but it seems like a convoluted way to pass an argument to a method to me.

davidhabib: basically i'm trying to keep it matching what would have been the correct event based design, if Lightning allowed visualforce pages to register their event handlers.  so yes, I prefer to keep it as is.

PR #2157, opened 2016-07-12
Feature/lex manage household

davidhabib:
@dospromptman This fixes a namespace fixup bug I uncovered when I added a namespace to my local DE org.  It showed up because any object like a contact that contained a subobject, like Account, wouldn't get the subfields namespace fixed up, and so we wouldn't see the correct data.  In practical terms, when looking at Contacts for add or merge, if we are in the HH model, I look at Contact.Account.Number_of_Household_Members__c, but in my contact object, which got fixed up, Account still held a field called djhcumulus2__Number_of_Household_Members__c.  

Please review and merge once it passes build.  thanks.

### DIFF
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -350,7 +350,7 @@
             } else {
                 con.npo02__Household__c = null;
             }
-            con.hhId = null;
+            con.HHId__c = null;
             
             var listConRemove = component.get('v.listConRemove');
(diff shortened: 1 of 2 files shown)


davidhabib:
@dospromptman hold off.  I think it may have one issue I'm testing now.  thanks.

davidhabib:
@dospromptman OK, I resolved the issue (you can't do typeof foo === "object" without making sure foo isn't null!)  please review and merge when passes build.

PR #2159, opened 2016-07-13
Feature/jen preserve contact address

### LINKED ISSUE #2111: moving contact to new household account should preserve their mailing address
hub thread: https://powerofus.force.com/0D58000002kgcIb

jjbennett:
# Warning
# Info
# Issues
Fixes #2111


jjbennett:
@davidhabib would you mind reviewing this address fix pull request?

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -101,8 +113,8 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
                     if (con.is_Address_Override__c)
                         // track which address's need their dates updated
                         mapAddrIdIsOverride.put(con.Current_Address__c, true);
-                } else if (isContactAddressEmpty(con)) {
-                	// if the contact has no address specified, pick up the hh default.
+                } else if (isContactAddressEmpty(con) && acctHasAddress) {
```
davidhabib: given the contact address was empty, what was the harm in adding it to listConAddrReset, even if its household had no address?  did this change fix a bug?

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -135,13 +147,13 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
             	
-            	Contact conOld = (Contact)listOld[i];
-            	
             	// new highest priority! if Account changes... 
-                if (conOld != null && con.AccountId != conOld.AccountId) {
+                if (acctChanged) {
                     // if no override, refill from the Default Address
             	    if (!con.is_Address_Override__c && !isContactAddressChanged(con, conOld)) {
-                        listConAddrReset.add(con);
+                        if (acctHasAddress) { // refresh address if the contact does not have one and account does
```
davidhabib: if the acct doesn't have an address, I see you correctly avoid adding the contact to the reset address list, but don't we then want to do have a new address object created on the account, just like it was an override, and the contact added to listConCreateAddr?

src/classes/ADDR_Contact_TDTM.cls
jjbennett: Yes, this was pretty much the fix for the issue, the address reset creates a blank default address when one is not found and sets it on the contact which was what was causing it to blank out.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -188,9 +200,8 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
                     continue;
                 }
 
-                // if the address changed, remember the contact we want to add a new address for    
-                Contact conOld = (listOld != null ? (Contact)listOld[i] : null);
-                if (isContactAddressChanged(con, conOld)) 
+                // if the contacts address changed or if the contact with an address moves to a new household remember the contact
+                if (isContactAddressChanged(con, conOld) || (acctChanged))
```
davidhabib: ok, I see your adding the || acctChange will get the address created (listConCreateAddr), but wouldn't that be incorrect if the contact has no address?

src/classes/ADDR_Contact_TDTM.cls
jjbennett: I think that was what I originally did but then changed it to create on line 204 when I thought the contacts address should be added to an account the same as a new contact would. Let me take another look at this.

src/classes/ADDR_Contact_TDTM.cls
davidhabib: ah got it.  didn't realize the address reset would create a blank one if none found.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -188,9 +200,8 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
                     continue;
                 }
 
-                // if the address changed, remember the contact we want to add a new address for    
-                Contact conOld = (listOld != null ? (Contact)listOld[i] : null);
-                if (isContactAddressChanged(con, conOld)) 
+                // if the contacts address changed or if the contact with an address moves to a new household remember the contact
+                if (isContactAddressChanged(con, conOld) || (acctChanged && !isContactAddressEmpty(con)))
```
jjbennett: @davidhabib updated the condition to include a check on the contact's address.

davidhabib: @jjbennett I have to admit I'm troubled that this code might be adding a contact to listConCreateAddr twice.  If the contact gets added in 158, due to acct changing, not an override, and no address fields changed, won't the contact get added again in 204 is acctChanged and not empty address fields?

davidhabib: and I take the blame that this code is confusing!  I wrote it, and it was pretty hard to figure out the best way to structure the code and make it follow the precedence rules I came up with and deal with all the permutations of potential changes!  and of course there was no official spec!

jjbennett: Yes, but I think it is ok, because if the address is changed in the before context then it will check and deal with duplicates but if it is in the after context and not a duplicate it will add it to the address. I think this is the trickery you mentioned in your comment.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -52,18 +54,22 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
         	
         DmlWrapper dmlWrapper = new DmlWrapper();        
         list<Contact> listConAddrSet = new list<Contact>(); 
-        list<Contact> listConCreateAddr = new list<Contact>();
+        Set<Contact> setConCreateAddr = new Set<Contact>();
```
jjbennett: @davidhabib per our discussion, changed the variable to a set and converted it back before calling the method.

davidhabib:
@jjbennett @jlantz not sure why this feature build is now failing.  seems totally unrelated to the changes in this pull request.

jjbennett:
@davidhabib I spoke with @jlantz and it appears to be the limit issue with the physical delete again.

PR #2165, opened 2016-07-19
Feature/2127 psc lookup icon

### LINKED ISSUE #1942: Update opportunity naming to better handle currency fields
Currently a custom label is applied to the currency, which is OK for single currency orgs but not ideal for multicurrency.

### LINKED ISSUE #2127: Manage Soft Credit page lookup icon partially hidden
https://powerofus.force.com/0D58000002mBmcJ

njjc:
# Warning
# Info
# Issues
Fixes #1942
Fixes #2127

### DIFF
--- a/src/classes/OPP_OpportunityNaming.cls
+++ b/src/classes/OPP_OpportunityNaming.cls
@@ -410,7 +410,11 @@ public with sharing class OPP_OpportunityNaming implements OPP_INaming {
             }
         } else if (dt == Schema.DisplayType.Currency) {
             if (obj.get(fieldName.trim())!=null) {
-                result = Label.npe01.DefaultContactTransactionCurrency + String.valueOf(obj.get(fieldName.trim()));
+                String currencySymbol = UTIL_Currency.getInstance().getCurrencySymbol((Id)obj.get('Id'));
+                if (currencySymbol.length() == 3) {
+                    currencySymbol += ' ';
+                }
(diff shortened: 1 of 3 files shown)


njjc:
Before removing slds-input--small


After removing slds-input--small

src/classes/OPP_OpportunityNaming.cls
```diff
@@ -410,7 +410,11 @@ public with sharing class OPP_OpportunityNaming implements OPP_INaming {
             }
         } else if (dt == Schema.DisplayType.Currency) {
             if (obj.get(fieldName.trim())!=null) {
-                result = Label.npe01.DefaultContactTransactionCurrency + String.valueOf(obj.get(fieldName.trim()));
+                String currencySymbol = UTIL_Currency.getInstance().getCurrencySymbol((Id)obj.get('Id'));
+                if (currencySymbol.length() == 3) {
```
davidhabib: what is this for?  why is 3 special?

njjc: In multicurrency orgs, it's always an ISO code that's attached to the amount, e.g. USD 500, so it needs the space. In single currency orgs it maps to a currency symbol, e.g. $500.

njjc: Note that this doesn't yet handle which side of the number to put the currency symbol on.

PR #2169, opened 2016-07-19
Import German translation STF

dospromptman:
# Info

German translations are now available

### DIFF
--- a/src/objectTranslations/DataImport__c-de.objectTranslation
+++ b/src/objectTranslations/DataImport__c-de.objectTranslation
@@ -1,0 +1,575 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
+    <caseValues>
+        <caseType>Nominative</caseType>
+        <plural>false</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Nominative</caseType>
+        <plural>true</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Accusative</caseType>
+        <plural>false</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Accusative</caseType>
+        <plural>true</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Genitive</caseType>
+        <plural>false</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Genitive</caseType>
+        <plural>true</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Dative</caseType>
+        <plural>false</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <caseValues>
+        <caseType>Dative</caseType>
+        <plural>true</plural>
+        <value><!-- NPSP Data Import --></value>
+    </caseValues>
+    <fields>
+        <label>Account1-Importstatus</label>
+        <name>Account1ImportStatus__c</name>
+    </fields>
+    <fields>
+        <label>Account1 importiert</label>
+        <name>Account1Imported__c</name>
+        <relationshipLabel>NPSP-Datenimporte (Account1 importiert)</relationshipLabel>
+    </fields>
+    <fields>
+        <help><!-- Account1.BillingCity --></help>
+        <label>Account1 Stadt</label>
+        <name>Account1_City__c</name>
+    </fields>
+    <fields>
+        <help><!-- Account1.BillingCountry --></help>
+        <label>Account1 Land</label>
+        <name>Account1_Country__c</name>
+    </fields>
+    <fields>
+        <help><!-- Account1.Name --></help>
+        <label><!-- Account1 Name --></label>
+        <name>Account1_Name__c</name>
+    </fields>
+    <fields>
+        <help><!-- Account1.Phone --></help>
+        <label>Account1 Telefon</label>
+        <name>Account1_Phone__c</name>
+    </fields>
+    <fields>
+        <help><!-- Account1.BillingState --></help>
+        <label>Account1 Bundesstaat/Provinz</label>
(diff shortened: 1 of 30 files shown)


src/objectTranslations/DataImport__c-de.objectTranslation
```diff
@@ -0,0 +1,575 @@
...
+        <label><!-- Account2 Website --></label>
+        <name>Account2_Website__c</name>
+    </fields>
+    <fields>
+        <help><!-- Account2.BillingPostalCode --></help>
+        <label>Account2 Postleitzahl</label>
+        <name>Account2_Zip_Postal_Code__c</name>
+    </fields>
+    <fields>
+        <help>Die ApexJobID des Batchauftrags, mit dem der Datensatz verarbeitet wurde</help>
```
dospromptman: This is okay to translate, yes?

src/objectTranslations/DataImport__c-de.objectTranslation
```diff
@@ -0,0 +1,575 @@
...
+        <label>Spende Datum</label>
+        <name>Donation_Date__c</name>
+    </fields>
+    <fields>
+        <help><!-- Opportunity.Description --></help>
+        <label>Spende Beschreibung</label>
+        <name>Donation_Description__c</name>
+    </fields>
+    <fields>
+        <help>Gibt an, ob die Spende eine Einzelspende von Kontakt1 ist oder eine Organisationsspende von Account1.</help>
```
dospromptman: This is okay to translate, yes?

src/objectTranslations/DataImport__c-de.objectTranslation
```diff
@@ -0,0 +1,575 @@
...
+        <label>Spende Startdatum der Mitgliedschaft</label>
+        <name>Donation_Membership_Start_Date__c</name>
+    </fields>
+    <fields>
+        <help><!-- Opportunity.Name --></help>
+        <label>Spende Name</label>
+        <name>Donation_Name__c</name>
+    </fields>
+    <fields>
+        <help>Der Name des Opportunity-Datensatztyps, der für die Spende verwendet werden soll Verwendet den Standarddatensatztyp, falls das Feld leer ist.</help>
```
dospromptman: This is okay to translate, yes?

src/objectTranslations/DataImport__c-de.objectTranslation
```diff
@@ -0,0 +1,575 @@
...
+        <label>Spende Phase</label>
+        <name>Donation_Stage__c</name>
+    </fields>
+    <fields>
+        <help><!-- Opportunity.Type --></help>
+        <label>Spende Typ</label>
+        <name>Donation_Type__c</name>
+    </fields>
+    <fields>
+        <help>Beschreibung der Ursache des Fehlers beim Importieren des Datenimportdatensatzes</help>
```
dospromptman: This is okay to translate, yes?

src/objectTranslations/DataImport__c-de.objectTranslation
davidhabib: yes ok

src/objectTranslations/DataImport__c-de.objectTranslation
davidhabib: ok

src/objectTranslations/DataImport__c-de.objectTranslation
davidhabib: ok

src/objectTranslations/DataImport__c-de.objectTranslation
davidhabib: ok

src/objectTranslations/Schedulable__c-de.objectTranslation
```diff
@@ -0,0 +1,92 @@
...
+        <label>Klassenname</label>
+        <name>Class_Name__c</name>
+    </fields>
+    <fields>
+        <help>Wie oft der Auftrag ausgeführt wird</help>
+        <label>Häufigkeit</label>
+        <name>Frequency__c</name>
+        <picklistValues>
+            <masterLabel>Daily</masterLabel>
+            <translation><!-- Daily --></translation>
```
davidhabib: @dospromptman remind me why we don't want these picklist value for Frequency__c translated?  Is it because our code isn't looking at the developer name (masterLabel) internally?  do we have a backlog item related to this?

dospromptman: I based the [filter list](https://github.com/SalesforceFoundation/Cumulus/blob/dev/scripts/stf/filters.json) on [the spreadsheet](https://docs.google.com/spreadsheets/d/1Z9HsXtt4ioIVFm1D5CAnLcWWYSMgxCzeeJnYVBtVubI) where Kevin reviewed each key and pointed out which values might be problematic if translated.  I, myself, have never investigated the reasons why or done any experimentation.

According to Kevin's comment for the Scheduleable__c.Frequency__c field, this value can not be translated [because the value we compare it to in the code is an Enum](https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/UTIL_MasterSchedulableHelper.cls#L88).

Many of the the "untranslatable" fields are documented in the "[Nonprofit Starter Pack and Translation](https://powerofus.force.com/articles/Resource/Nonprofit-Starter-Pack-and-Translation)" knowledge article.

However, Elena recently [posted in the hub](https://powerofus.force.com/0D58000002nmmMB) that she tried translating some of the disallowed values, and she found that the workflow rules relating to these values do still work.  (Our documentation says that translating these values will break workflow rules).

So, I suspect that _some_ of these picklist values actually can be safely translated, since I think that querying a picklist value in SOQL will return the untranslated label, [unless you specify `toLabel()`](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_tolabel.htm) for the field in the query.  The only way to be sure is to look carefully at where each of these values is being queried and used, and to experiment/test to verify the behavior of the system after translation.

I would love it if we invested some time in evaluating which fields are truly untranslatable, and then fix the code so that they are translatable.  As far as I know, there is nothing in the backlog for further investigation in this area.

davidhabib: As I thought.  So I assume your scripts kept these picklist labels on the black list.  Can you please add a backlog item to review all picklists on the blacklist, to look at our code, and confirm whether they can be translated or not, and update the blacklist as appropriate.  thanks!

PR #2172, opened 2016-07-21
Feature/manage hh fixes

### LINKED ISSUE #2160: new MHUI: don't allow adding the same contact more than once
Search for a Contact and add them as Member to the Household
Search again and add the same Contact to the Household
Save
Get errors

screen shot 2016-07-13 at 3 57 50 pm

Suggest removing Contacts already added to the Household from search results; alternately, produce an error when attempting to add the same Contact a second time.

### LINKED ISSUE #2171: New Manage Household: Adding new contact to Household Account does not update Number of Household field
STR:

* Household record with one member
* Launch Manage Household and add new Contact through field in upper right
* Save
* Result: Number of Household Members remains at 1

Repeat by adding a new Contact directly from the Contact related list on the Account and the Number of Household Members value increases as expected.

davidhabib:
This request improves autocomplete to handle data callbacks not returning in order.  It now ignores results that aren't for the most recent data request.  This has fixed the issue where you can see random results at times (no gitHub issue).  Also fixes issue #2160.

# Critical Changes

# Changes

# Issues Closed
Fixes #2160 
Fixes #2171

### DIFF
--- a/src/aura/HH_AutoComplete/HH_AutoComplete.cmp
+++ b/src/aura/HH_AutoComplete/HH_AutoComplete.cmp
@@ -5,6 +5,9 @@
     <aura:attribute name="listCon" type="Object[]" access="public" />
     <aura:attribute name="isListVisible" default="false" type="Boolean" access="public"/>    
 
+    <!-- methods -->
+    <aura:method name="clearList" action="{!c.clearList}" access="public" />
+
     <!-- events we fire -->
     <aura:registerEvent name="ContactNewEvent" type="c:HH_ContactNewEvent" />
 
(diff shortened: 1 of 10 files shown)


davidhabib:
@dospromptman ready for review and merge.  thanks.

dospromptman:
@davidhabib I don't really get how this solves #2160, but I don't want to start digging too deep into the code to find out.  Adding the number of household members somehow prevents adding duplicates in the UI?

davidhabib:
@dospromptman thanks for merging.  this branch contained fixes for two unrelated bugs.  #2160 was fixed be clearing the search results after adding the contact.  the code already did the correct work to filter out existing contacts.  it just needed to have its results cleared after adding.  the other bug, #2171 was just updating the number of household members before sending the hh to the server for update.

PR #2177, opened 2016-07-25
ManageHHUI Save fixes

### LINKED ISSUE #2174: ManageHHUI hits save errors if NPSP error logging disabled
Many of our save scenarios are hitting record lock issues on the account, as so many pieces of the save hit the account (merging, updating contacts, updating the hh from the LC component, updating the hh from the visualforce page fieldset).  Seems like UpdateNames() is usually the call that fails with the record lock.  With NPSP Error logging enabled, this error isn't treated as fatal, and just gets logged.  But with error logging disabled, this error gets propagated up to the page, and the save fails.

### LINKED ISSUE #2186: If Household object is hidden in security, new Manage Household will not work
thread: https://powerofus.force.com/0D58000002qr06e

Pretty common to disable Household object in Household Account model to avoid confusion. Manage Household should still work on a Household Account in Household Account model.

image

davidhabib:
This pull request is an attempt to reduce the number of errors encountered during save scenarios from the ManageHouseholdUI.  When dealing with lots of data changes thru merging in additional households, I've found that we were frequently getting record locking errors when trying to update the household.  With these changes the frequency has been greatly reduced, but there are still some scenarios where I believe it will happen.

@dospromptman This pull requests re-orders the save method, so we update the household first, then merge households, then save contacts, and finally the visualforce page's save to field sets.

@chrismcmahon I could really use some additional help testing out this scenario.  You need to disable NPSP Error Handling from the NPSP Settings page.  Then try out lots of different scenarios of making many edits and saving.  I can only now reproduce the problems when I start with a household with ~100 opportunities, and I then merge in a separate 10 person household.  When I try with fewer opps, or fewer contacts in the household merge, I don't seem to hit errors.  If you could report any reproduceable errors you see displayed on the page, or emailed to you thru the NPSP error mechanism, please let me know.  Happy to discuss further if that would be helpful.  thanks!

# Critical Changes

# Changes

# Issues Closed
Fixes #2174 
Fixes #2186

### DIFF
--- a/src/classes/HH_ManageHH_CTRL.cls
+++ b/src/classes/HH_ManageHH_CTRL.cls
@@ -132,25 +132,15 @@ public with sharing class HH_ManageHH_CTRL {
     }
 
     /*******************************************************************************************************
-    * @description ActionMethod to save the Household and close the page.
-    * @return PageReference the original calling page's URL.
+    * @description ActionMethod to save the Household
+    * @return null
     */
     public PageReference save() {
         try {
-            // we've already saved the data for the contacts in the household.
-            // now we only need to save the account fields an update, and we don't want
-            // any of the normal account trigger work to run again.  specifically,
-            // we may have updated the names and greetings, because the contacts have changed,
-            // but the user still may want auto-naming turned on.  our normal trigger flow
-            // would think the user updated the name from what it was, and thus it should be
-            // excluded from auto-naming!
-            HH_ProcessControl.inFutureContext = true;
             update hh;
         } catch (Exception ex) {
             ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.Error, ex.getMessage()));
-            return null;
-        }
-        
-        return new PageReference('/' + hhId);
+        }        
+        return null; 
     }
 }
--- a/src/classes/HH_ManageHH_CTRL.cls
+++ b/src/classes/HH_ManageHH_CTRL.cls
@@ -136,7 +136,21 @@ public with sharing class HH_ManageHH_CTRL {
     * @return PageReference the original calling page's URL.
     */
     public PageReference save() {
-        update hh;
+        try {
+            // we've already saved the data for the contacts in the household.
+            // now we only need to save the account fields an update, and we don't want
+            // any of the normal account trigger work to run again.  specifically,
+            // we may have updated the names and greetings, because the contacts have changed,
+            // but the user still may want auto-naming turned on.  our normal trigger flow
+            // would think the user updated the name from what it was, and thus it should be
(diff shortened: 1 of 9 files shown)


src/classes/HH_ManageHH_CTRL.cls
```diff
@@ -136,7 +136,21 @@ public with sharing class HH_ManageHH_CTRL {
...
-        update hh;
+        try {
+            // we've already saved the data for the contacts in the household.
+            // now we only need to save the account fields an update, and we don't want
+            // any of the normal account trigger work to run again.  specifically,
+            // we may have updated the names and greetings, because the contacts have changed,
+            // but the user still may want auto-naming turned on.  our normal trigger flow
+            // would think the user updated the name from what it was, and thus it should be
+            // excluded from auto-naming!
+            HH_ProcessControl.inFutureContext = true;
```
dospromptman: This all seems like the wrong way to fix this, even if it is the least intrusive.  What do you think?  Reread your code comment here; does this seem like the way the system should be working?

Is this a temporary fix, or just the way it is supposed to work from now on?

dospromptman:
It doesn't seem like re-ordering the aura remote calls should make much of a difference.  Maybe it does tend to work better now, just by chance.  Since you're doing all this work in one component controller method, would it be more helpful to send everything in just one aura remote call and let the apex method control the sequencing of how the arguments are processed?  I don't know if that would work at all, since I don't fully understand the race conditions at play here.  I'm just throwing it out there as a possible alternative.

davidhabib:
@dospromptman OK, a new design!  
1. have the visualforce page save its household field set
2. have the lightning container call a single server method to save hh, contacts, merges, etc.
3. when the lightning save completes, ask the visualforce page if it had any validation errors or save errors.
4. if no errors, let the lightning component close the page.

This design is much cleaner.  no setting of global flags.  all the lightning objects are saved in a single transaction, so they avoid race issues.  the account merge still happens with a future, so it avoids system limits.

the only disappointment was there was no clean way for my lightning component to ask the visualforce page for its field contents, so it could do the save for it directly.  since the field set can contain lookups, and multiselect picklists, it would have taken a whole bunch of javascript logic to duplicate whatever magic visualforce does to coerce the html control values into the SObject on the server.  I went down the path a little bit, before realizing it was doomed!  So we continue to let the visualforce page save its contents, but we do it first, to avoid the locking issues with any account merges.

@chrismcmahon I've now done a bunch of manual testing and think it is all working, but when this gets into dev, i'd really appreciate if you banged on it manually to make sure you can't uncover any additional save issues.  thanks!

davidhabib:
@dospromptman ready for review and merge.  thanks.

PR #2184, opened 2016-07-27
Convert setting to use record type ids instead of names

### LINKED ISSUE #1693: Support multi-language orgs in Opp Record Type settings & rollups
From owiener as a comment on issue #968 

from your post dated October 2014, 2, you said: "As for the the record type issue, the current settings are storing the record type's name (not devname), which can get translated. The code then tests against name. So if you save your membership record type setting in French, then the rollups will work correctly in a French instance. You get problems, as reported above, when the setting is saved in one language, and then the code is run from another language. The fix for this would be to save all picklist and recordtype settings by their underlying value (for picklists) and DevName (for recordtypes), and always that those in the code. Unfortunately, this would break current user's settings where the saved setting doesn't match."

Did you progress on this one ? The solution you proposed last year (ie translate the record type in french on the english translation) won't work when we use a second language. I though I could create a new record type (which would be the equivalent of the english one), then in NPSP Settings > Donations > Donor Statistics >Opp Rec Types Excluded for Contact Rollup add that new record type, and re run the rollup donations batch. But that produce a very strange result: the donation linked to the english record type are back in the contact donation total (but should be in Memberships total), and the one using the newly created record type only appears in the Household donation totals (it doesn't appear in the contact donation totals and not in the membership totals) ?

I'm sorry to bother you with this subject again but I really need a solution to work with 2 or more languages.

dospromptman:
Modify the following custom settings to store record type id, instead of record type name.  This requires changing how the settings are stored (the various STG_Panel* pages and controllers), as well as every location where they are referenced (so that the referencing code expects to receive record type ids instead of names):

Allocations_Settings__c.Excluded_Opp_RecTypes__c
npe01__Contacts_And_Orgs_Settings__c.Opp_RecTypes_Excluded_for_Payments__c
npo02__Households_Settings__c.npo02__Household_Creation_Excluded_Recordtypes__c
npo02__Households_Settings__c.npo02__Household_OCR_Excluded_Recordtypes__c
npo02__Households_Settings__c.npo02__Membership_Record_Types__c
npo02__Households_Settings__c.npo02__Excluded_Account_Opp_Rectypes__c
npo02__Households_Settings__c.npo02__Excluded_Contact_Opp_Rectypes__c
Opportunity_Naming_Settings__c.Opportunity_Record_Types__c

This change is required to allow for localization of record type names in orgs where multiple languages are in use.  Since the record type name was being stored and then converted back to a record type id at runtime, that runtime conversion could fail if the language of the running user differs from the language that the record type names were stored in.  Storing these values as record type id allows NPSP to operate independently of the language of the running user or administrator. See #1693 for more details.

# Warning

This release updates the format in which some NPSP settings are stored, in order to enable organizations with multiple languages to use translated Record Type names.  The following settings have been updated:

* NPSP Settings > People > Households > Household Object > Household Creation Excluded Record Types
* NPSP Settings > Donations > Opportunity Names > Opportunity Record Types
* NPSP Settings > Donations > Membership > Membership Record Type Names
* NPSP Settings > Donations > Payments > Excluded Opp Record Types
* NPSP Settings > Donations > GAU Allocations > Excluded Opp RecTypes
* NPSP Settings > Donations > Donor Statistics > Opportunity Rollup Record Types > Opp RecTypes Excluded for Account Rollup
* NPSP Settings > Donations > Donor Statistics > Opportunity Rollup Record Types > Opp RecTypes Excluded for Contact Rollup
* NPSP Settings > Donations > Contact Roles > Household Opportunity Contact Roles > Household OCR Excluded Record Types

 Please review these settings on the NPSP Settings page.

# Info

# Issues

Fixes #1693


dospromptman:
@jvarese can you review the message in the Warning section for the release notes?

cdcarter:
@dospromptman - learning time for me, why not just use the RTs developer name/API name? Those shouldn't be translated, but can still be used across orgs?

kbromer:
Two reasons: 

1.  Depending on where/how it was retrieved, it actually can impact the usage around translations.  The referenced issue #1693 is a good example, there are others in the backlog as well

2.  Consistency. Right now, when you write to a setting, depending on who/when/how it was implemented, it might be an ID that's expected, it might be a name, up to you to figure it out as a developer, integrator or implementer. 

This way, we're deployment, language, environment and developer agnostic in our approach.

dospromptman:
@cdcarter Using `DeveloperName` is technically possible, but would actually be a bigger change than switching to using Ids.  The reason is because the existing code uses the `getRecordTypeInfos*()` methods on the `SObjectDescribeResult` object to access record type information.  There is a method to `getRecordTypeInfosByName()` and to `getRecordTypeInfosById()`, but there isn't any method that will let you access record types by `DeveloperName` (as far as I know).  Even the `RecordTypeInfo` objects you get back from these methods do not allow you to access the `DeveloperName` of the record type (but they do let you get the display name, id, and various other properties of the record type).  So, to get a record type by `DeveloperName` would require querying the `RecordType` sObject... which is certainly possible, but somewhat different than the way the existing code works.  Switching from display name to record type id for storage made more sense in this case.

I would have loved to see more abstraction in place around custom settings and record type manipulation, but it wasn't there.  That made this a bit of a tricky refactoring, and so a less ambitious refactoring is the safer choice.  Settings that store record types are accessed directly in about 250 different places in the code, and I had to carefully inspect each reference to evaluate the impact of this switch, so it was important to me for this iteration to step lightly wherever possible.  Hopefully some day the storage and processing of custom settings can be designed in such a way that a change in format like this could be done in just a single place.

Record type ids were already being used to store some custom settings, such as the Household Account Record Type (or anywhere else an Account record type is stored in a setting), so the storage of these settings was already not completely org independent.  Is it that you find yourself more often needing to change these particular settings in an automated way across orgs that makes the usage of ids troublesome?

cdcarter:
Understood the reasoning not to take the larger lift on this refactor. Somewhere there's a grumpy post from me about the existing inconsistency in the settings too, so all your points are well taken @dospromptman & @kbromer - thanks for explaining!

> Is it that you find yourself more often needing to change these particular settings in an automated way across orgs that makes the usage of ids troublesome?

That's the issue, yep. For any number of reasons, we may want to be deploying these settings from orgs that don't share IDs (configuration was originally built in a different trial, customer is doing a full org replacement, or even standard config that we have as a partner). A pretty special-snowflake use case but now it's in y'all's brains at least!

dospromptman:
@njjc I added the install handler conversion script, so this PR should be complete, ready for review, and ready to merge.

Note code requires that it be included in release 3.78.  If this does not end up being the case, we must update the code before it can be merged/pushed.

dospromptman:
I have updated this PR after all of the review feedback:

Move install script code that calls UTIL_RecordTypeSettingsUpdate outside of try/catch block, so that any exceptions thrown during the conversion process will cause the installation to fail.  Added a return statement in the catch block, so that if any exceptions occur at an earlier point in the install process, they are still caught, and handled by sending an email, but will now prevent later install steps from executing.

Mark UTIL_RecordTypeSettingsUpdate class, getInstance(), and updateSettings() methods as globals so that they can be called from outside of the managed package (i.e. a user can run the update script from the Developer Console)

Switching from using the settings getInstance() and getList() methods to querying the settings sObjects directly.  This is because getInstance() would have returned a setting object appropriate for the running user, but we really want to update all stored settings (including those at the org and profile default level)

Removing some unit tests that relied too heavily on internal implementation details, and now stub out only the minimal set of methods to be able to perform a more end-to-end test.

Update version number checks with the expectation that this code will be included in v3.79 of Cumulus, instead of v3.78

njjc:
Manual testing look good, I'll still need to test the upgrade portion with the beta package.

cpolcyn:
@dospromptman Please update the section headers in your PR body like so:
Warning --> Critical Changes
Info --> Changes
Issues --> Issues Closed

PR #2195, opened 2016-08-02
DON'T MERGE: Changing PMT_PaymentCreator to create a payment when the opportunity …

### LINKED ISSUE #1714: Payment should be auto-created if no existing Payment record and Opportunity Amount changes to > 0 from 0 or null
This would allow compatibility with Opportunity Line Items (Products).

njjc:
…is updated from a null amount, regardless of whether the opportunity is open or closed/won.

# Warning
# Info
# Issues
Fixes #1714


njjc:
@mpusto here's the pull request for the payment amount update for closed/won opps.

njjc:
**lurch: review apex visualforce

njjc:
This is by design as we don't update anything for closed/won opportunities. Users should enter opportunities with products in an open stage, add line items, then close the opportunity.

PR #2204, opened 2016-08-08
Levels Assignment Batch

davidhabib:
This adds a new NPSP Settings page under the Bulk Data Processes menu, called Level Assignment Batch.

@njjc this is ready for review and merge once the build finishes and passes.  thanks.

### DIFF
--- a/src/classes/LVL_LevelAssignBATCH_TEST.cls
+++ b/src/classes/LVL_LevelAssignBATCH_TEST.cls
@@ -39,6 +39,7 @@ private class LVL_LevelAssignBATCH_TEST {
     private static Level__c lvlBronze { get; set; }
     private static Level__c lvlSilver { get; set; }
     private static Level__c lvlGold { get; set; }
+    private static Level__c lvlAcctZinc { get; set; }
 
     /*********************************************************************************************************
     * @description creates Bronze, Silver, and Gold Level records for test code to use.
@@ -75,7 +76,17 @@ private class LVL_LevelAssignBATCH_TEST {
             Minimum_Amount__c = 5000, 
             Maximum_Amount__c = 1000000);
         
-        insert new Level__c[]{lvlBronze, lvlSilver, lvlGold};        
+        lvlAcctZinc = new Level__c(
+            Name = 'Zinc',
+            Target__c = 'Account',
+            Source_Field__c = 'npo02__TotalOppAmount__c', 
+            Level_Field__c = 'npo02__LastMembershipLevel__c', 
+            Previous_Level_Field__c = 'npo02__LastMembershipOrigin__c', 
+            Active__c = true, 
+            Minimum_Amount__c = 100, 
+            Maximum_Amount__c = 200);
+
+        insert new Level__c[]{lvlBronze, lvlSilver, lvlGold, lvlAcctZinc};        
     }
 
 
@@ -104,6 +115,18 @@ private class LVL_LevelAssignBATCH_TEST {
         insert listCon;
     }
         
+    /*********************************************************************************************************
+    * @description creates a list of Accounts with various donation totals for test code to use.
+    */
+    private static list<Account> listAcc { get; set; }
+    static void createTestAccounts() {
+        // create & insert contact
+        listAcc = UTIL_UnitTestData_TEST.CreateMultipleTestAccounts(2, null);
+        listAcc[0].npo02__TotalOppAmount__c = 10;
+        listAcc[1].npo02__TotalOppAmount__c = 150;
+        insert listAcc;
+    }
+
     /*********************************************************************************************************
     * @description test our batch process for assigning Levels to Contacts
     */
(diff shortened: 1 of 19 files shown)


davidhabib:
not sure why this build doesn't show up on this pull request, unless its due to going from non-feature to feature.  here is the build: https://salesforce-org.atlassian.net/builds/browse/NPSP-FEAT74-2

njjc:
That must be it, it would be nice if it would show up when run manually, is that possible @jlantz?

davidhabib:
Oops, @njjc I just committed the work for the Level trigger to do validation.  I meant to keep it separate just to make it easier to review.  sorry about that.

njjc:
No worries, separate commits is fine!

davidhabib:
@njjc Just added support for Health Check of Level fields.

src/classes/LVL_Level_TDTM.cls
```diff
@@ -0,0 +1,166 @@
...
+    * @param triggerAction Which action triggered the current context. (Before or After; Insert, Update, 
+    * Delete, or Undelete.)
+    * @param objResult The field describe object for the current trigger context. 
+    * @return dmlWrapper Objects that need to be inserted or updated after the processing of the trigger.
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist, 
+        TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+            
+            // only validate before triggers
+            if (triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
```
njjc: Do you think we should handle the undelete case?

src/classes/LVL_Level_TDTM.cls
```diff
@@ -0,0 +1,166 @@
...
+                }
+                
+                // for each object, validate any updates to its ladders
+                for (String strObj : setStrObject) {
+                    
+                    // get our map of Ladder Levels
+                    map<String, list<Level__c>> mapLadderToLevels = LVL_LevelAssign_BATCH.mapLadderToLevelsForObject(strObj);
+                    
+                    // go thru our new/updated levels
+                    for (integer i = 0; i < newlist.size(); i++) {
```
njjc: Here we're looping through the entire trigger set, even though we've only got the mapLadderToLevels for one sObject type at this point. I think in this for loop you need to make sure we're only dealing with levels of the correct sObject type. Also we're running a query for each sObject, it would be preferable to run a single query for all sObject types in the trigger set, although it's not a big deal in this transaction.

src/classes/LVL_LevelAssignBATCH_TEST.cls
```diff
@@ -187,4 +187,147 @@ private class LVL_LevelAssignBATCH_TEST {
...
+
+        try {
+            Test.startTest();
+            insert new Level__c[]{lvl1, lvl2, lvl3};
+            Test.stopTest();
+            system.assert(false, 'exception should have occurred');
+        } catch(Exception ex) {
+            system.assert(ex.getMessage().contains(lvl3.Name));
+        }
+    }
```
njjc: A test for inserting levels to more than one object type is needed.

src/classes/LVL_Level_TDTM.cls
```diff
@@ -0,0 +1,166 @@
...
+        }
+        return listLvl;
+    }
+    
+    /*******************************************************************************************************
+    * @description insert the Level in the correct order for the list of Levels
+    * @param lvl The Level to insert
+    * @param listLvl The list to insert into
+    */
+    private void insertLevel(Level__c lvl, list<Level__c> listLvl) {
```
njjc: I think we should call this "addLevel" to correspond with the list methods, insert makes it feel like it's performing DML.

src/classes/LVL_Level_TDTM.cls
davidhabib: I think not.  that would prevent the undelete from working.  I think it would be better to allow them to undelete so they can then modify the undeleted level as they see fit.

src/classes/LVL_Level_TDTM.cls
davidhabib: good catch on the iterating all objects, and missing a filter for the current strObj type.  

as for each strObj doing a separate query, I think the number of unique objects will be small enough, and it keeps the code much simpler, and having to construct and maintain a map of maps.

src/classes/LVL_LevelAssignBATCH_TEST.cls
davidhabib: good point

src/classes/LVL_Level_TDTM.cls
njjc: Agreed, that works for me.

src/classes/LVL_Level_TDTM.cls
davidhabib: ok

src/classes/LVL_Level_TDTM.cls
njjc: That's a good point, it's just too bad that we can't rely 100% on the fact that overlapping levels will never exist without it.

davidhabib:
OK @njjc i've updated with your feedback and fixes.  once it passes build, you can merge away.  thanks!

PR #2206, opened 2016-08-09
Major Gift fields on Opp and Grant Deadlines rename to Deliverables

### LINKED ISSUE #1617: Suggest renaming 'Grant Deadline' object to 'Deliverable'
Brian Pickett (North Peak)
Deliverable management is a great feature (thanks @Sam Knox (501 Commons) and the SF Foundation!), though its usefulness is not limited to grants. Suggest that the language be generalized around "Deliverables" rather than made so specific with "Grant Deadlines."

davidhabib:
# Critical Changes

# Changes
* NPSP has renamed the Grant Deadlines object to be Deliverables.  The fields have also been renamed to be less Grant specific.  This was done to make this object more applicable to additional types of Opportunities, such as Major Gifts.  All of the developer/api names have stayed the same.  If you prefer sticking with the previous labels, you can rename then, following the steps outlined in this knowledge article.  https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_rename.htm&language=en_US
* The Deliverables (previously Grant Deadlines) object now has a new picklist field called Type.  You need to enable Field Level Security on this new field.
* Opportunity has 4 new fields that you need to enable Field Level Security for.  These are Ask Date, Qualified Date, Gift Strategy, and Closed Lost Reason.
* Several Opportunity fields that had Grant specific labels have been renamed to be more general for use with Grants and Major Gifts.  All of the developer/api names have stayed the same.

# Issues Closed
Fixes #1617


kbromer:
We might want to pre-empt this by pointing users to the "How to Change Labels" KB article from the release notes when this is released - thoughts?

davidhabib:
Good idea Kevin.  I added the following:

If you prefer sticking with the previous labels, you can rename then, following the steps outlined in this knowledge article. https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_rename.htm&language=en_US

kbromer:
👍

jjbennett:
@davidhabib should the new Opportunity-Major Gift Layout.layout be added to the package? I noticed that de and en translations were added for the changed Grant_Deadline object and de only for the new fields on the Opportunity object. I'm always unsure of when we need to include the translation files, is there any hard and fast rule on that?

judisohn:
For the record, we have been including page layouts in the package (Matching Gift and Membership, for example) but not the Record Types.

davidhabib:
@jjbennett 
- yes, we want the opp major gift layout added to the package, even though we won't package a record type.
- i just include the language files whenever my checkin causes new ones to be generated.  since this is usually just placeholders for new terms, not a big deal to include or not include.  but i tend to include if they have changed.
- don't merge this pull request until W-008308 Add "Type" field as part of included Grant Deadline Object fields is also done.  I'll ping you when it is ready early next week.
thanks!

davidhabib:
@jjbennett I went ahead and finished this work, so it now also has the new Type picklist field on the Deliverables object, and I've added the Major Gift opp page layout to package.xml.

Unfortunately, this is sitting in the build queue, so not sure when it will run build.  hopefully soon, and hopefully it will pass.  but no need to get this merged until next week.  I'd like to get Nic's stuff in first into dev.  thanks.

davidhabib:
@jjbennett added the missing Opp field renames.  thanks.

jjbennett:
@davidhabib looks good, I'll wait for your go ahead to merge.

PR #2214, opened 2016-08-11
Updating record type localizations query

dospromptman:
Including a query that explicitly referenced the RecordTypeLocalization
type, via the Localization relationship on the RecordType sObject, was
causing deployment errors for orgs that did not have translation
workbench enabled.  These orgs act as if RecordTypeLocalization (and the
Localization relationship on RecordType) do not exist.

I am updating the querying of record types to attempt to query record
types with translations first, but if that fails then to only query
untranslated record types.  Additionally, when iterating over the
RecordTypes returned by this query, I attempt to iterate over child
sObjects in the Localization relationship field, but if this field was
not queried, then I silently ignore this failed attempt at reading
translated values.

This should allow this code to be deployed to orgs that do not have
translation workbench enabled, and should also prevent any runtime
errors when running in orgs that do or do not have it enabled.  If
translation workbench is enabled, then it should include any record type
translations in the returned id map.


cpolcyn:
Looks good to me.

PR #2219, opened 2016-08-15
Feature/jen slds scoping

jjbennett:


src/aura/HH_Canvas/HH_Canvas.cmp
```diff
@@ -17,9 +17,9 @@
 
     <!-- markup follows... -->
 
-    <ul id="SortCanvas" class="slds-list--horizontal slds-has-cards--space" style="display: inline;">
+    <ul id="SortCanvas" class="slds-list--horizontal slds-has-dividers--around-space slds-wrap">
```
davidhabib: did you test this in the manage household ui to make sure drag/drop of the cards still works.  I only put in styles when I had no choice to get something working.

src/aura/HH_ContactCard/HH_ContactCard.cmp
```diff
@@ -9,13 +9,13 @@
     <!-- markup follows... -->
 
     <div class="slds-card slds-p-around--none">
-        <div class="slds-card__header slds-p-around--none">
+        <div class="slds-card__header slds-p-around--small">
```
davidhabib: did you do any side by side testing to make sure the end results was the same, or do I need to test this out in my own org to verify?  thanks.

src/package.xml
```diff
@@ -252,14 +252,6 @@
         <name>ApexClass</name>
     </types>
     <types>
-        <members>autocomplete</members>
```
davidhabib: whoops!  I don't think we want these removed from package.xml

src/aura/HH_Canvas/HH_Canvas.cmp
jjbennett: I did do testing but only in Chrome, the inline style was acting wonky after the updated slds, it caused the card to take the entire width of the page. When I removed it, all was well with the size but then when I added a bunch of contacts it wasn't wrapping so I added the slds-wrap. slds-has-cards--space was deprecated so switched that over to the slds-has-dividers--around.

src/aura/HH_ContactCard/HH_ContactCard.cmp
jjbennett: Yes, but never hurts to double check!

src/package.xml
jjbennett: It looks like it was listed in there 2x, it shows up again on line 365.

davidhabib:
@jjbennett so looks like slds 2 has broken the expand/collapse of sections in the Change Address component of the address manager.  You didn't change any of the markup in this area, but my use of <div class="{!v.isExistingAddrSectionOpen ? 'slds-section slds-is-open' : 'slds-section'}"> in HH_AddressMgr.cmp in line 61 and 92 is no longer working.

I verified under the debugger that the click on the section header is being handled and we toggle our state variable v.isExistingAddrSectionOpen.  But we aren't getting the hide/show we used to get when LC would rerender the section.

Here's the link to the slds doc that shows using this attribute, but of course their examples don't actually work.
http://www.lightningdesignsystem.com/components/utilities/text/#flavor-section-title-section-title-open

anyways, I could spend more time trying to figure this out, but I'm going to let you track this one down, because I don't think I'll solve it tonite in time for the build cutoff.

jjbennett:
@davidhabib I believe the issue is that toggling the visibility is now reserving the space for the hidden item. I added conditional collapsing to the section content, please take a look when you have a chance and let me know what you think.

PR #2220, opened 2016-08-16
Update record type custom settings on new install

dospromptman:
If a user installs 'npsp' package when they were previously using the
underlying packages in some other way (i.e. NPSP 2.0), then that user
might have settings that need to be updated to reflect the change in
storage format.

This commit updates the post install conditions to also run the settings
update code on a new install of NPSP.  In an org where these settings
were not populated (i.e. a brand new NPSP install), this should be a
noop.

A real world QA test for this would involve installing underlying packages,
editing settings to contain record type names, and then installing the
'npsp' package.  The custom settings should have been updated with the
corresponding record type ids.

### DIFF
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -63,7 +63,7 @@ global without sharing class STG_InstallScript implements InstallHandler {
             //We won't rethrow the exception, to avoid package failure installation due to install script error
             return;
         }
-        if (context.isUpgrade() && context.previousVersion().compareTo(new Version(3, 79)) < 0) {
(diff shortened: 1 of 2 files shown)


davidhabib:
@dospromptman @njjc @cpolcyn This pull request shows as failing build, but when I looked at the bamboo log, I couldn't easily figure out what went wrong.  looks like all the test ran successfully.  @cpolcyn can you please investigate asap, as this is a pull request we must get in for the release.  thanks.

njjc:
CPU time limit exceeded on a test class... I'll run it again to see if it happens again, maybe it's an intermittent server load issue.

cpolcyn:
Thanks @njjc for retrying that build. It passed on the 2nd attempt so this PR was merged. Now to see how the build on master (dev) goes…

PR #2221, opened 2016-08-16
Removing SLDS source from Static Resource Sources

jjbennett:


njjc:
@cpolcyn looks like the build org for this build has an expired password.

cpolcyn:
I swapped the org and retried the build, and it passed. Should be good to go!

njjc:
Thanks!

PR #2228, opened 2016-08-20
Rewrite UTIL_RecordTypes class

dospromptman:
Modify signatures of all UTIL_RecordTypes methods to accept an SObjectType instead of a string.  Modify all calling code to use SObjectType literal instead of string literal to refer to object types.

Also remove any methods not called from external code.

Modify some test classes to remove calls to methods that were being tested but never used by external code (since these methods are no longer needed)

### DIFF
--- a/src/classes/UTIL_RecordTypes.cls-meta.xml
+++ b/src/classes/UTIL_RecordTypes.cls-meta.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
-    <apiVersion>33.0</apiVersion>
+    <apiVersion>37.0</apiVersion>
     <status>Active</status>
 </ApexClass>
--- a/src/classes/UTIL_RecordTypes_TEST.cls
+++ b/src/classes/UTIL_RecordTypes_TEST.cls
@@ -40,47 +40,40 @@ private class UTIL_RecordTypes_TEST {
     // TEST
     static testmethod void testRecTypes() {
         // try bogus values
-        Id reallyBogus = UTIL_RecordTypes.GetRecordTypeId('Bogus', 'Bogus');
-        Id bogus = UTIL_RecordTypes.GetRecordTypeId('Opportunity', 'Bogus');
-        bogus = UTIL_RecordTypes.GetRecordTypeId('Contact', 'Bogus');
+        Id bogus = UTIL_RecordTypes.GetRecordTypeId(Opportunity.sObjectType, 'Bogus');
+        bogus = UTIL_RecordTypes.GetRecordTypeId(Contact.sObjectType, 'Bogus');
         bogus = UTIL_RecordTypes_API.GetRecordTypeId('Campaign', 'Bogus');
         
         // try all the functions
         set<Id> oppRtSet = UTIL_RecordTypes_API.GetRecordTypeIdSet('Opportunity', (new set<string>{'Bogus', 'Master', 'Grant'}));
-        Map<String, Id> oppTypeMap = UTIL_RecordTypes.GetRecordTypeIdMap('Opportunity');
-        List<SelectOption> rtOptions = UTIL_RecordTypes.getRecordTypesForSelectList('Opportunity');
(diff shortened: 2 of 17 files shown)


src/classes/UTIL_RecordTypes.cls-meta.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
-    <apiVersion>33.0</apiVersion>
+    <apiVersion>37.0</apiVersion>
```
dospromptman: I am using v37 of the API for the `Schema.RecordTypeInfo.isMaster()` method.  If you want to avoid mixing version numbers, then hold off on merging this until all the other classes are up to v37.

src/classes/UTIL_RecordTypes_TEST.cls
```diff
@@ -40,47 +40,40 @@ private class UTIL_RecordTypes_TEST {
     // TEST
     static testmethod void testRecTypes() {
         // try bogus values
-        Id reallyBogus = UTIL_RecordTypes.GetRecordTypeId('Bogus', 'Bogus');
```
dospromptman: I dropped this line because with the change to SObjectType in the signature, there isn't a way to pass in a "bogus" object type (which is one of the main benefits to using SObjectType instead of string literals)

src/classes/UTIL_RecordTypes_TEST.cls
```diff
@@ -40,47 +40,40 @@ private class UTIL_RecordTypes_TEST {
...
-        Id bogus = UTIL_RecordTypes.GetRecordTypeId('Opportunity', 'Bogus');
-        bogus = UTIL_RecordTypes.GetRecordTypeId('Contact', 'Bogus');
+        Id bogus = UTIL_RecordTypes.GetRecordTypeId(Opportunity.sObjectType, 'Bogus');
+        bogus = UTIL_RecordTypes.GetRecordTypeId(Contact.sObjectType, 'Bogus');
         bogus = UTIL_RecordTypes_API.GetRecordTypeId('Campaign', 'Bogus');
         
         // try all the functions
         set<Id> oppRtSet = UTIL_RecordTypes_API.GetRecordTypeIdSet('Opportunity', (new set<string>{'Bogus', 'Master', 'Grant'}));
-        Map<String, Id> oppTypeMap = UTIL_RecordTypes.GetRecordTypeIdMap('Opportunity');
```
dospromptman: I dropped this line because I removed the `GetRecordTypeIdMap` method, since it was not being used by any external code

src/classes/UTIL_RecordTypes_TEST.cls
```diff
@@ -40,47 +40,40 @@ private class UTIL_RecordTypes_TEST {
...
+        ID idGift = UTIL_RecordTypes.GetRecordTypeId(Opportunity.sObjectType, UTIL_RecordTypes.getRecordTypeNameForGiftsTests(Opportunity.sObjectType));
         
-        Set<String> setStr = new Set<String>{UTIL_RecordTypes.getRecordTypeNameForGiftsTests('Opportunity')};
-        Set<Id> setId = UTIL_RecordTypes.GetRecordTypeIdSet('Opportunity', setStr);
+        Set<String> setStr = new Set<String>{UTIL_RecordTypes.getRecordTypeNameForGiftsTests(Opportunity.sObjectType)};
+        Set<Id> setId = UTIL_RecordTypes.GetRecordTypeIdSet(Opportunity.sObjectType, setStr);
         if(idGift!=null){
             system.Assert(setId.contains(idGift));
         
-            Map<String, Id> mapRCId = UTIL_RecordTypes.GetRecordTypeIdMap('Opportunity');
```
dospromptman: I dropped this line because I removed the `GetRecordTypeIdMap` method, since it was not being used by any external code

src/classes/UTIL_RecordTypes.cls-meta.xml
njjc: Sounds good, the v37 pull request is ready, but we have to fix some build issues first.

PR #2229, opened 2016-08-22
Feature/2128 jen opp wf copy fmv amt

### LINKED ISSUE #2121: Add 'Disable Duplicate Management Rules' to Account Converter checklist
https://powerofus.force.com/0D58000002lJRWb

In the 'Important Information' column we should add a checkbox for 'All Data.com Duplication Management Rules have been disabled.'

### LINKED ISSUE #2128: fix workflow rule Opportunity Copy FMV to Amount
https://powerofus.force.com/0D58000002lIe9h

### LINKED ISSUE #2202: Account Converter not populating Household ID on Account custom field
thread: https://powerofus.force.com/0D58000002rh4FT

@kbromer I duplicated in my test instance as well. Field didn't copy - length 255, definitely wouldn't have been an ID conflict issue. Perhaps worth a deeper look?

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2128 
Fixes #2202 
Fixes #2121

### DIFF
--- a/src/pages/CONV_Account_Conversion.page
+++ b/src/pages/CONV_Account_Conversion.page
@@ -58,6 +58,14 @@
                                 </div>
 
                                 <div class="slds-form-element">
+                                    <label class="slds-checkbox" for="check-dup-disabled">
+                                        <input name="checkbox" type="checkbox" id="check-dup-disabled" />
+                                        <span class="slds-checkbox--faux"></span>
+                                        <span class="slds-form-element__label">{!$Label.CONV_All_Dup_Rules_Disabled}</span>
+                                    </label>
+                                </div>
+
+                                <div class="slds-form-element">
                                     <label class="slds-checkbox" for="check-sandbox-first">
                                         <input name="checkbox" type="checkbox" id="check-sandbox-first" />
                                         <span class="slds-checkbox--faux"></span>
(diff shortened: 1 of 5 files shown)


jjbennett:
@dospromptman would you mind reviewing this pull request, since german was active in my org the custom label was added to that translation file, but I wasn't sure if that should be committed and if so whether or not I should also make sure it is added to the other translation files.

dospromptman:
@jjbennett can you write a little bit about what #2202 entails?  I can't figure out what that ticket is for, and what part of this PR fixes it.

src/pages/CONV_Account_Conversion.page
```diff
@@ -227,10 +236,6 @@
             <div class="slds-backdrop" tabindex="-1"></div>
         </div>
 
-        <apex:form >
-            <apex:actionFunction name="convert" action="{!convert}" reRender="pageMessages" oncomplete="afterConvert()"/>
```
jjbennett: @dospromptman moving line 231 into the above form was the fix for #2202, because they were in different forms the value was never being passed into the controller.

dospromptman:
Including the label in the German language translations file is fine, and omitting it from the other files should also be fine.  I think this should not break anything, since there is no actual translation for that label.  It will, however, appear in metadata retrieves from orgs where those languages are enabled, and it will show up in the repository diff at that time.  That doesn't affect this pull request, though.

This PR looks good to me, should I merge it?

jjbennett:
Ok, thanks for looking it over and yes this is ready to go.

PR #2230, opened 2016-08-22
Level enhancements

davidhabib:
This pull requests includes:

- validation that a Level name is unique within its ladder
- improved validation when levels saved to tag the correct record with validation errors, such as overlapping another level.
- support for both Minumum Level and Maximum Level being null, meaning no lower/upper limit
- changed treatment of Maximum Level to be exclusive, leaving Min inclusive, ie,  Min <= data < Max.
- additional Level Health Check logic to detect invalid field datatypes
- Level Health Check test code

### DIFF
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -735,7 +735,7 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
      }
 
     /*********************************************************************************************************
-    * @description Verifies that all Level records refer to existing fields
+    * @description Verifies that all Level records refer to existing fields of the correct type
     */
     public void verifyLevels() {
         try {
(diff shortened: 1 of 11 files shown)


davidhabib:
@njjc this is ready for review and merge.  thanks.

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -756,12 +756,29 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
                         cErrors++; 
                         createDR(Label.healthLabelVerifyLevels, statusError, 
                             string.format(Label.healthDetailsInvalidLevelField, new string[]{lvl.Name, strField, lvl.Target__c}),
                             string.format(label.healthSolutionInvalidLevelField, new string[]{lvl.Name}));
-                    }        
+                    } 
+                    // ensure the datatype is valid
+                    else if (strField != null) {
+                        Schema.Displaytype dt = UTIL_Describe.getFieldDisplayType(lvl.Target__c, strField);
+                        if (i == 0 && !UTIL_Describe.canDisplaytypesCopy(dt, Schema.DisplayType.Double)) {
```
njjc: I'm not positive, but if a number field has 0 decimal places, could its [display type](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_enum_Schema_DisplayType.htm) be Integer? Also, couldn't this feature theoretically work fine with percent fields?

davidhabib: displaytype.double returns true for copying between all of the various numeric types.  so it covers all those cases.

njjc: Ah gotcha, sounds good. I forgot about that method.

PR #2234, opened 2016-08-23
Feature/api 37

njjc:


njjc:
@davidhabib this is ready to go, since this is a big change in how describes work, we should get this in soon so we have a week to tease out issues before release.

PR #2243, opened 2016-08-30
Feature/2173 jen svg edge ie

### LINKED ISSUE #2173: Icons not displaying in ManageHHUI and other SLDS pages under IE and Edge
due to SVG icons not supported on these browsers.  there are some workarounds we need to implement.  we are not yet sure on which of our SLDS themed visualforce pages have this issue, but assume some do.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2173


jjbennett:
@davidhabib icons now working when Locker Service is activated, except in IE.

davidhabib:
@jjbennett I'm confused.  I thought the whole point of this SVG redesign effort was to get our icons displaying in Edge and IE.  But your comment says it doesn't work in IE.  I don't recall you saying that yesterday.

jjbennett:
@davidhabib this is when Locker Service is enabled and in IE, switching to innerHTML fixes it everywhere with Locker Service enabled except IE.

davidhabib:
@jjbennett @kbromer @dospromptman @njjc I've downloaded this branch to my DE org to play with it, and as a user, I'm very unhappy with the delayed load of the icons.  It is too noticeable in my opinion, so I am uncomfortable with us going with this solution.  When searching for a contact, the hits would display, and sometimes it would take 3+ seconds for the icons to show up to add the contact.  too disruptive.

Also the fact that it does not work under IE w/ Locker Service makes this a non-starter.  Here is what I see in the console by the way:
WARNING: Callback failed: apex://djhcumulus2.UTIL_SvgIcon/ACTION$getSvgIconMarkupAura
SCRIPT5022: Error in $A.getCallback() [TypeError: Invalid calling object]

thoughts?

dospromptman:
1. Let's scrap this.
2. Implement [the latest official Lightning Component Helper](http://lightningdesignsystem.com/resources/lightning-svg-icon-component-helper/).
3. Include the svg4everybody script in Lightning Components (as described [here](http://lightningdesignsystem.com/components/icons/)).

At that point, we'd be completely SLDS compliant and be following officially recommended best practices.

davidhabib:
I agree with Jason's proposal, with one minor addition:

1. if svg4everybody doesn't work in IE11, then we change the icon on the add to household dropdown to use a white sprite on a colored background, so at least the color background will display where the button is.  Then we can at least tell users to click where the colored button is.  Today, you see nothing at all since we chose a transparent background for it.

dospromptman:
@davidhabib If you can get that to work in the MHH UI, go for it.  But, in general, I don't think we should take even one more step in attempting to work around this SLDS bug.  Supposedly this is fixed in Winter '17 (W-3178764).

PR #2246, opened 2016-08-30
Update UTIL_RecordTypes.getRecordTypeName

dospromptman:
UTIL_RecordTypes.getRecordTypeName() was returning null if the running user did not have access to the requested record type.  This was a change in behavior introduced during the UTIL_RecordTypes rewrite.

This changed behavior was causing the UTIL_RecordTypes_TEST tests to fail if run in an org where a custom Opportunity record type was created, but the running user did not have access to that record type. The test would pass if there were either no custom Opportunity record types defined, or if the user running the tests had access to those record types.

This fixes the broken behavior by returning the record type name, regardless of whether the user has access to the requested record type or not.

### DIFF
--- a/src/classes/UTIL_RecordTypes.cls
+++ b/src/classes/UTIL_RecordTypes.cls
@@ -107,11 +107,7 @@ public class UTIL_RecordTypes {
         Map<Id, Schema.RecordTypeInfo> rtiById = objectType.getDescribe().getRecordTypeInfosById();
         if (rtiById.containsKey(recordTypeId)) {
             Schema.RecordTypeInfo rti = rtiById.get(recordTypeId);
-            if (!rti.isMaster() && rti.isAvailable()) {
-                return rti.getName();
-            } else {
-                return null;
-            }
+            return rti.getName();
         } else {
             return null;
         }


src/classes/UTIL_RecordTypes.cls
```diff
@@ -107,11 +107,7 @@ public class UTIL_RecordTypes {
         Map<Id, Schema.RecordTypeInfo> rtiById = objectType.getDescribe().getRecordTypeInfosById();
         if (rtiById.containsKey(recordTypeId)) {
             Schema.RecordTypeInfo rti = rtiById.get(recordTypeId);
-            if (!rti.isMaster() && rti.isAvailable()) {
```
davidhabib: so it looks like you would not return a name if it was the Master record type.  Now you will return it.  Is that a problem?

dospromptman: No, I don't think so.

If you look at [the prior implementation of UTIL_RecordTypes](https://github.com/SalesforceFoundation/Cumulus/blob/33d9a7ea14583c151d2ca93721c37e0092146c31/src/classes/UTIL_RecordTypes.cls#L44), the `getRecordTypeName` method was inspecting the `recordTypesById` map and returning the value from there (if it existed).  Taking out these access/master checks now makes the code follow essentially the same path as before: `sObjectType.getDescribe().getRecordTypeInfosById()` followed by `containsKey(recordTypeId)`, then returning the record type name.

My error in the rewrite was that I didn't notice that `getRecordTypeName` was using the `recordTypeIds` list (which is not filtered), instead of the `recordTypesList`, which is the only one that is filtered (for master record types and available record types).

Not only that, but I am pretty sure that the master record type isn't even returned by these `getRecordTypeInfos` methods.

PR #2249, opened 2016-09-02
Adding Household field to editable Contact Fields

### LINKED ISSUE #2232: BUG: Adding Contact in Manage Household w/Household Object adds Contact to its own Household
Thread: https://powerofus.force.com/0D58000002uMbOA

Duplicated issue following described STR:

Set Account model to 1:1 and Household creation to 'All New or Edited Contacts'

1) From Jane Smith contact record, access household by clicking "Smith Household"
2) Click "Edit" button on household record
3) Use search box that reads "Find a contact or add a new contact to the household" to search for "John Smith"
4) John Smith is not found, so click "+ New Contact"
5) New Contact box appears with John Smith's name already populated. Click blue "New Contact" button
6) John Smith now appears to be included as a household member. Click blue "Save" button in upper right-hand corner.
Results: Back on the household page, there is still only one household member (Jane Smith). There is a new record for John Smith, but he's a member of a new Smith Household.

Note: The Household is named correctly if First Names are in naming formula - example 'Jane and John Smith Household' but the contacts are in separate households, so the issue is happening at some point later in the process.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2232


davidhabib:
@jjbennett I actually specifically removed checking the security on Household__c because we have many users who are using the HH Account model, and have they have removed permissions on the Household object (since they aren't using it).  Even if the user sets FLS to view and edit the Household__c field, if the underlying object does not have permissions, then FLS check will fail.  See the comment on line 186 in HH_Container_LCTRL.cls.

So I think the fix for this bug would be to only add npo02__Household__c to setUpdateableContactFields if we are editing a Household object, rather than Household Account.  If we do this, I'm not sure if you can then remove line 191        

strSoql += ', npo02__Household__c'; 

completely, or if in the HH Account model, we still need this field part of the soql.  You'll need to test this out to make the correct determination.  Thanks.

davidhabib:
@jjbennett this fix looks great.  much less intrusive than you and I had discussed several days ago.  

I wonder, however, if by putting npo02__Household__c in setUpdateableContactFields, that when we save the contacts, our code will still try to check FLS on npo02__Household__c and fail, if the user has no permissions on npo02__Household__c or Household__c.  Did you test that out?

the fix might require changing UTIL_Describe.copyObjectFLS() to not check FLS is the source field is null.

jjbennett:
@davidhabib I created a user that had access to Households but not to that specific field on Contacts and when I saved I did get an error about permission for the Contact. So the error still shows up just not on initial load of the page.

davidhabib:
yep, as I suspected.  unfortunately, in looking at UTIL_Describe.copyObjectFLS(), it is designed to also copy null columns over, so my suggestion of avoiding the FLS checks if the source is null won't work.

I think an appropriate fix would be in HH_Container_LCTRL.upsertContacts() to explicitly remove npo02_Household__c from the set, and just manually copy the value after calling UTIL_Describe.copyObjectFLS().  Kinda a hack, but I just don't see how we can easily figure out when to do FLS on npo02_Household__c.

jjbennett:
I see what you are seeing, I thought it was a good thing that the security still occurred but I am not sure if I tested the Account Household model with that user. Let me make sure, unless you would rather go this route and skip security on that field all together?

davidhabib:
We need to make sure that the page works correctly with the household account model, even if the admin removes permissions to the Household__c object and Household__c field.

And we need to make sure that the page works correctly with Household Objects (1:1 account model) if the user does have permissions.

As for the page needing to fail if the user is using Household Objects but doesn't have perms, that would be ideal, but I won't lose sleep on it if we can't easily pull it off, since the old page did no perm checking at all!

dospromptman:
I have deployed this to an org and, using the Standard Profile user, with a permission set that granted: Read access to Address__c object; Create, Read, Edit, Delete access to Household__c object; Edit access to Contact fields: Address Override, Current Address, Exclude from Household Greetings/Name.

After following the steps to reproduce that Judi reported in the related bug, it created the John Smith contact on the existing Smith Household Household__c object.

I created a different permission set with very limited permissions and performed the same steps using the Household Account model, with the result that the John Smith contact was created on the Smith Household account.

In neither case did I have to grant permissions to the Contact.Household__c field.

PR #2250, opened 2016-09-06
Feature/nic actionplans

njjc:
Ready for review and testing. I've got two issues I need help with:

1) Field validation messages for the tasks are displayed twice. Some research has shown that this should be resolved by adding apex:outputLabels for each inputField, which I have done, but the issue persists.
2) I can't get the dead simple rerender at line 63 of the page to work. It was working at first, then stopped, and I've spent forever trying to fix it and haven't been able to.

Let me know if I should split out some of the controller logic to make a smaller class. It feels a bit unwieldy.


davidhabib:
@njjc looks like you may not have checked in two files.  here are the build errors:

All Component Failures:
06-Sep-2016 08:48:45	1.  package.xml (EP_TaskDependencyCheck_TDTM) -- Error: An object 'EP_TaskDependencyCheck_TDTM' of type ApexClass was named in package.xml, but was not found in zipped directory
06-Sep-2016 08:48:45	2.  package.xml (EP_TaskDependencyCheck_TEST) -- Error: An object 'EP_TaskDependencyCheck_TEST' of type ApexClass was named in package.xml, but was not found in zipped directory

davidhabib:
- I updated package.xml to remove the renamed file references.

- Did you intend to include a Tab for Engagement Plan Templates?  I had to create my own to get going.

- Did you intend to include a page layout for Engagement Plan Templates?  I had to modify the default page layout to include all the object's fields, and also the Manage Template button.  I have not checked in this change.

- I had to update the dynamic commandButton's label from $Label.EPAddTask to EPAddDependentTask.  I wasn't sure if you forgot to include the EPAddTask label, or whether it should use EPAddDependentTask.  I have not checked in this change.

- The Assign To picklist and User lookup are not styled with SLDS in my DE org.  Also the npsp-slds styling fixup to get the lookup button inside the User lookup edit, doesn't appear to be working.

- Your tests were failing, because you were looking for the format 1.1, but you are actually using 1-1.  I've fixed up your tests and will check that simple fix in to your branch.

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+    private EP_Task_UTIL util;
+
+    /** @description Property used by the Add Dependent Task and Delete Task buttons to tell where in the heirarchy
+    * to perform the functions. */
+    public String currentTaskLevel {get; set;}
+
+    /** @description Set the select options Reminder Time field.*/
+    public List<SelectOption> reminderTimeOptions {
+        get {
+            List<SelectOption> options = new List<SelectOption>{
```
davidhabib: do these need to get localized?

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+            return options;
+        }
+        set;
+    }
+
+    /** @description Picklist options for the Task Priority field. We're using Task values directly in case they have
+    * been updated from their defaults.*/
+    public List<SelectOption> taskPriorityOptions {
+        get {
+            return listSelectOptions('Task','Priority');
```
davidhabib: need to support localization with custom label

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+        }
+    }
+
+    /** @description Picklist options for the Task Status field. We're using Task values directly in case they have
+    * been updated from their defaults.*/
+    public List<SelectOption> taskStatusOptions {
+        get {
+            return listSelectOptions('Task','Status');
```
davidhabib: localize support

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+        }
+    }
+
+    /** @description Returns a list of SelectOptions for a given object and picklist field name.
+    * @param objectName The name of the object.
+    * @param fieldName The name of the field.
+    * @return List<SelectOption> The select options for this picklist field.
+    */
+    private List<SelectOption> listSelectOptions(String objectName, String fieldName) {
```
davidhabib: I coulda sworn we had these in a UTIL class, but I can't seem to find them.  I would be good to move these to UTIL so we can share them, since they are so general.  i won't hold up the pull request for this, however.

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+        }
+
+        for (String fieldName : epTaskFields) {
+            checkFieldFLS('Engagement_Plan_Task__c', fieldName);
+        }
+    }
+
+    /** @description For a given object and field, verifies that the field is accessible and updateable,
+    * throwing an exception if it isn't.*/
+    private void checkFieldFLS(String objectName, String fieldName) {
```
davidhabib: another candidate to move into UTIL_Describe.

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+        return null;
+    }
+
+    /** @description Removes a given wrapper from the heirarchy by removing the current wrapper from the parent's list
+    * of children, and adding current children to parent's list of children. 
+    * @param wrapper The TaskWrapper to remove from the wrapper heirarchy.
+    * @return void This method updates class properties.
+    */
+
+    private void removeFromHeirarchy(TaskWrapper wrapper) {
```
davidhabib: it's spelled hierarchy!  i before e except after c!

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+            'slds-p-around--small", style="border-left: 6px solid ' + sidebarColors[colorNumber] + '"><legend id=task' + wrapper.level + 
+            ' class="slds-text-heading--small">' + Schema.Sobjecttype.Task.getLabel() + ' ' + wrapper.level + '</legend>', escape=false);
+
+        Component.Apex.OutputPanel firstRow = new Component.Apex.OutputPanel(styleClass='slds-grid', layout='block');
+        Component.Apex.OutputPanel secondRow = new Component.Apex.OutputPanel(styleClass='slds-grid', layout='block');
+        Component.Apex.OutputPanel cardFooter = new Component.Apex.OutputPanel(styleClass='slds-grid', layout='block');
+        Component.Apex.OutputPanel buttonGroup = new Component.Apex.OutputPanel(styleClass='slds-button-group slds-p-top--large', layout='block');
+
+        firstRow.childComponents.add(genericInputPanel(wrapper, accessString, 'Name', 'slds-col slds-size--1-of-4', 'slds-input'));
+        firstRow.childComponents.add(genericInputPanel(wrapper, accessString, 'Assigned_To__c', 'slds-col slds-size--1-of-4', 'lookupInput203'));
```
davidhabib: why is this one a genericInputPanel?  shouldn't it be a selectListInputPanel?

davidhabib: also, the lookup control gets obscured by the Type field.  why aren't you explicitly creating a component for it, so it has room?

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+    /** @description Generates the markup for an SLDS styled apex:inputField.
+    * @param wrapper The TaskWrapper for the current Engagement Plan Task.
+    * @param accessString The string for accessing the current place in the heirarchy using a nested expression.
+    * @param fieldName The name of the field.
+    * @param outterCss CSS markup for the field input container.
+    * @param inputCss CSS markup for the input itself.
+    * @return Component.Apex.OutputPanel OutputPanel containing the markup for an inputField.
+    */
+    private Component.Apex.OutputPanel genericInputPanel(TaskWrapper wrapper, String accessString, String fieldName, String outterCss, String inputCss) {
+        Component.Apex.OutputPanel outterPanel = fieldLabelPanel(wrapper, fieldName, outterCss);
```
davidhabib: for required fields (ie, name), ideally you would add a preceding red asterisk, to mimic how Lightning Experience displays required fields.

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -0,0 +1,629 @@
...
+    * @return Component.Apex.OutputPanel OutputPanel containing the markup for a picklist field.
+    */
+    private Component.Apex.OutputPanel selectListInputPanel(TaskWrapper wrapper, String accessString, String fieldName, String outterCss, String inputCss, String selectOptions) {
+        Component.Apex.OutputPanel outterPanel = fieldLabelPanel(wrapper, fieldName, outterCss);
+
+        Component.Apex.SelectList selectList = new Component.Apex.SelectList(
+            id = fieldName + wrapper.level,
+            value = wrapper.detail.get(fieldName),
+            styleClass = 'slds-form-element__control ' + inputCss,
+            style = 'height: calc(1.875rem + 2px);',
```
davidhabib: why not use class names for some css you put in the header of the page, rather than embedding this specific height calculation in the code.

src/classes/EP_ManageEPTemplate_CTRL.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>34.0</apiVersion>
```
davidhabib: should be api version 37 like the rest of the project

src/classes/HH_Container_LCTRL.cls
```diff
@@ -564,7 +564,7 @@ public with sharing class HH_Container_LCTRL {
                 
                 for (String strF : setFields) {
                     // check field level security
-                    DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(UTIL_Namespace.StrTokenNSPrefix('Address__c'), strF);
```
davidhabib: not sure why this is showing in your pull request since this already went into dev.  wonder if your branch is not up to date?

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+                <div class="slds-spinner--brand slds-spinner slds-spinner--medium" style="position:fixed;top:30%;" aria-hidden="false" role="alert">
+                    <div class="slds-spinner__dot-a"></div>
+                    <div class="slds-spinner__dot-b"></div>
+                </div>
+            </div>
+            <!-- PAGE HEADER -->
+            <div class="slds-page-header" role="banner">
+              <div class="slds-grid">
+                <div class="slds-col">
+                  <p class="slds-text-heading--medium">Manage Engagement Plan Template</p>
```
davidhabib: needs to become a custom label

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+                  <!--h1 class="slds-text-heading- -medium">My Expenses</h1-->
```
davidhabib: we can probably get rid of this line!

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+                        <apex:commandButton id="saveCloseBTN" styleClass="slds-button slds-button--brand" action="{!saveClose}" value="{!$Label.stgBtnSave}" immediate="false"/>
+                        <apex:commandButton id="cancelBTN" styleClass="slds-button slds-button--neutral" action="{!cancel}" value="{!$Label.stgBtnCancel}"  immediate="true"/>
+                    </div>
+                </div>
+              </div>
+            </div>
+            <!-- / PAGE HEADER -->
+            
+            <apex:outputPanel id="panelErrors">
+                <c:UTIL_PageMessages html-class="slds-grid slds-grid--align-center slds-m-bottom--small"/>
```
davidhabib: since this component does the slds styling, why did you have to add this html-class list?  I know my pages that use c:UTIL_PageMessages have never done this.

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+            </apex:outputPanel>
+
+            <!-- TEMPLATE VIEW -->
+            <form class="slds-form slds-form--horizontal">
+                <apex:outputPanel styleClass="slds-grid" id="templateForm">
+                    <div class="slds-col slds-col--padded slds-p-top--large slds-size--1-of-2">
+                        <div class="slds-form-element slds-p-top--small slds-is-required">
+                            <apex:outputLabel value="{!$ObjectType.Engagement_Plan_Template__c.Fields.Name.Label}" for="idName" styleClass="slds-form-element__label" />
```
davidhabib: add the required asterisk like I showed in a previous comment on all of the required fields on this page.

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+            
+            <apex:outputPanel id="panelErrors">
+                <c:UTIL_PageMessages html-class="slds-grid slds-grid--align-center slds-m-bottom--small"/>
+            </apex:outputPanel>
+
+            <!-- TEMPLATE VIEW -->
+            <form class="slds-form slds-form--horizontal">
+                <apex:outputPanel styleClass="slds-grid" id="templateForm">
+                    <div class="slds-col slds-col--padded slds-p-top--large slds-size--1-of-2">
+                        <div class="slds-form-element slds-p-top--small slds-is-required">
```
davidhabib: in slds 203, in the documentation I see is-required defined, but no slds-is-required.  is the doc wrong, or your code wrong?

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+                                
+                                    <apex:outputLabel value="{!$ObjectType.Engagement_Plan_Template__c.Fields.Automatically_Update_Child_Task_Due_Date__c.Label}" for="idUpdateDueDate" styleClass="slds-form-element__label" />
+                                </label>
+                            </div>
+                            <apex:outputText styleClass="slds-form-element__help" value="{!$ObjectType.Engagement_Plan_Template__c.Fields.Automatically_Update_Child_Task_Due_Date__c.inlinehelpText}" />
+                        </div>
+                    </div>
+                </apex:outputPanel>
+            </form>
+            <p class="slds-text-heading--medium">Manage Engagement Plan Tasks</p>
```
davidhabib: need custom label

src/pages/EP_ManageEPTemplate.page
```diff
@@ -0,0 +1,111 @@
...
+        <apex:actionFunction name="addDependentTaskJS" action="{!addDependentTask}" immediate="true" status="statusLoad" reRender="taskTree">
+            <apex:param name="depentendTaskParam" value="" assignTo="{!currentTaskLevel}" />
+        </apex:actionFunction>
+        <apex:actionFunction name="deleteTaskJS" action="{!deleteTask}" immediate="true" status="statusLoad" reRender="taskTree">
+            <apex:param name="deleteTaskParam" value="" assignTo="{!currentTaskLevel}" />
+        </apex:actionFunction>
+        <apex:actionFunction name="updateTemplateForm" immediate="true" status="statusLoad" reRender="theForm" />
+        <!--spinner status-->
+        <apex:actionStatus id="statusLoad"
+                    onstart="document.getElementById('divLoading').style.display = 'block';console.log('started');"
```
davidhabib: remove the console logging

src/pages/EP_ManageEPTemplate.page-meta.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexPage xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>34.0</apiVersion>
```
davidhabib: version 37

davidhabib:
I fixed the rerender issue you had with the checkbox.  you needed to change the actionSupport to have event="onchange" and immediate="false"
```
<apex:actionSupport event="onchange" immediate="false" reRender="templateForm" status="statusLoad"/>
```

davidhabib:
I'm guessing you are seeing multiple error messages displayed because you are doing addMessages(ex).  And I see that in some controllers, we've done             ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.Error, ex.getMessage()));   which would only display one message.

I think your code is actually more correct, and the messages are not exactly the same.  for example, if you add a new task and don't fill in the subject field, and then save the first error says "Error: you must enter a value" and the second one says "Error: Subject: you must enter a value".        

it's a little weird, but obviously the exception being thrown by the system has both messages in it, and I'm ok that we display both.

davidhabib:
I added the custom label for EPAddTask, and since it is only referenced dynamically, I had to add it to our list of labels to force the spider to put in the package.

davidhabib:
I've now also tested this pull request by merging into my de org with my Level code, and the AssignTo control does now display as a single lookup control.  So I assume this is correct, and that my report that it was showing a non styled select list and non styled lookup field was due to my DE org not getting the updated static resource.  Can you confirm that it should look like this?

mpusto:
Looks good to me.

src/classes/EP_ManageEPTemplate_CTRL.cls
davidhabib: you might be able to ignore my last two comments!  in my 2nd DE org, the control appears correctly as a single field.  not sure why my 1st DE org didn't display it correctly, unless the static resource failed to update for some reason.

davidhabib:
@njjc @mpusto @jjbennett @kbromer I have gone ahead and merged this into the actionplans combined branch, and will now create a new pull request to put actionplans into dev.  Even though I identified things I want @njjc to address in the pull request, such as using custom labels in a few places that hardcoded strings were used, there is nothing I see that is critical to be done before Nic gets back from vacation.  At this point I am ok with this work going into dev so we can move foward.  @njjc I do expect you to followup with all of the comments in this now closed pull request when you get back from vacation.  thanks.

PR #2286, opened 2016-09-19
Feature/jen lex navigation

jjbennett:
# Critical Changes

# Changes
When converting a Lead and viewing an existing Contact, a popup will open instead of the print view of the Contact record. The popup will display fields from the "Contact Merge Found FS" field set on the Contact object.

# Issues Closed


davidhabib:
@jjbennett I notice that in Lead Convert, the popup does not have the X close button that is supposed to live in the upper right (outside of the popup).  we should add it to be consistent with SLDS standards and the other popups in NPSP.

jjbennett:
@davidhabib I meant to mention that this morning, I actually pulled it out because it doesn't actually appear to be showing up in our modals.

davidhabib:
I see the X appearing in my popups in Manage Household, as well as in Matching Gifts (find more gifts popup).

jjbennett:
Ok, I was looking at the contact merge, but I will go back and see what I can figure out. Thanks!

jjbennett:
@davidhabib The rerender issue with the close icon was resolved by adding the "docType="html-5.0" attribute to the page.

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -179,17 +184,48 @@ public with sharing class LD_LeadConvertOverride_CTRL {
     private map<id, contact> conMap {
         get {
             if (conMap == null) {
+                Set<String> fieldSet = new Set<String>{'Id', 'Name', 'FirstName', 'LastName'};
+                String queryStr = 'SELECT ';
+                for (FieldSetMember cfs : conFieldSet)
+                   fieldSet.add(cfs.getFieldPath());
+                queryStr += String.join(new List<String>(fieldSet), ',');
+                queryStr += ' FROM Contact WHERE (LastName = \'' + l.LastName + '\' AND FirstName = \'' + l.FirstName + '\') OR (Email = \'' + l.Email + '\' AND Email != null) LIMIT 50';
```
davidhabib: sorry for not catching this last review.  you should avoid concatenating fields directly in soql strings that could potentially be malicious or ill formatted.  your dynamic soql should still use variable reference syntax.  Ie, '...where Lastname = :l.llastname...'

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -179,17 +184,48 @@ public with sharing class LD_LeadConvertOverride_CTRL {
...
+
+    /*******************************************************************************************************
+    * @description Display the selected Contact
+    */
+    public Contact getSelectedCon() {
+        if (strConId!='NEW_CONTACT' & strConId!=null && conMap.containsKey(strConId)) {
+            selectedCon = conMap.get(strConId);
+        } else if (strConId == null) {
+            selectedCon = null;
+        }
```
davidhabib: what will this return if strConId != null, but its not found in conMap?  Should it then return null?

jjbennett:
@davidhabib thanks, fixed up both issues.

PR #2298, opened 2016-09-22
Feature/2267 smarty streets endpoint

### LINKED ISSUE #2267: Update SmartyStreets Zipcode URL endpoint
hub thread: https://powerofus.force.com/0D58000002wvvVh

looks like smarty streets changed their endpoints, and our defaults in ADDR_SmartyStreets_Validator need to be updated.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2267


njjc:
I'm a bit weary of adding a new remote site for the first time. I know that on install the user has to accept the current three remote sites, so I'm wondering if this could possibly cause any issues with push upgrades. Any thoughts @jlantz?

jjbennett:
@njjc - if we change the endpoint and the remote site isn't included the users will get an error, could we maybe include instructions on adding the remote site in the critical changes section instead?

njjc:
If we can't push the new remote site that would be preferable.

njjc:
[This is the only thing I could find on it.](http://salesforce.stackexchange.com/questions/85242/adding-remote-site-to-managed-package-can-we-push-to-customers) Says it will work, but I'm not sure if we should trust the source. Do you have any other resources that could confirm for us, @jlantz?

njjc:
Jason Lantz: there's nothing special about remoteSiteSettings listed here on this page so I think we're in the clear: https://help.salesforce.com/apex/HTViewHelpDoc?id=design_your_app.htm

PR #2299, opened 2016-09-22
Feature/2292 ep required subject

### LINKED ISSUE #2292: Deleting a blank Engagement Plan Task from within Engagement Plan Manage Template requires task subject
reported in a comment here: https://powerofus.force.com/0D58000002wyimz

image

njjc:
# Critical Changes

# Changes

# Issues Closed
- Fixes #2292

### DIFF
--- a/src/classes/EP_EngagementPlans_TDTM.cls
+++ b/src/classes/EP_EngagementPlans_TDTM.cls
@@ -38,8 +38,8 @@
 public with sharing class EP_EngagementPlans_TDTM extends TDTM_Runnable {
 
     /*******************************************************************************************************
-    * @description Trigger Handler on Engagement Plans after insert; creates Tasks based on the associated 
-    * template's Engagement Plan Tasks.
+    * @description Trigger Handler on Engagement Plans; creates Tasks based on the associated 
+    * template's Engagement Plan Tasks after insert, and prevents modifying lookups after update.
     * This method inserts Task objects directly instead of using DmlWrapper, as DML options are needed on
     * certain Tasks for sending emails.
     * @param listNew the list of Engagement Plans from trigger new. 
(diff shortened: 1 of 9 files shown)


src/classes/EP_EngagementPlans_TDTM.cls
```diff
@@ -102,6 +102,19 @@ public with sharing class EP_EngagementPlans_TDTM extends TDTM_Runnable {
...
             }
         }
+
+        //validate that the user doesn't add two lookups, or modify an existing lookup to a new value
+        if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            for (Integer i = 0; i < newlist.size(); i++) {
+                Engagement_Plan__c newEP = (Engagement_Plan__c) newlist[i];
+                Engagement_Plan__c oldEP = (Engagement_Plan__c) oldlist[i];
+                Id newEpTarget = (Id) newEP.get(EP_EngagementPlans_UTIL.getTargetObjectField(newEP));
+                Id oldEpTarget = (Id) oldEP.get(EP_EngagementPlans_UTIL.getTargetObjectField(newEP));
```
davidhabib: should this be doing getTargetObjectField(oldEP) rather than newEP?

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -277,6 +277,10 @@ public with sharing class EP_ManageEPTemplate_CTRL {
             for (TaskWrapper wrapper : mapTaskWrappers.values()) {
                 Engagement_Plan_Task__c epTask = wrapper.detail;
                 epTask.Engagement_Plan_Template__c = template.id;
+                if (String.isBlank(epTask.Name)) {
+                    epTask.Name.addError(Label.NameRequired);
```
davidhabib: don't you need to throw an exception in order for it to get displayed on the page.  I thought AddError only worked inside a trigger trying to update or save an object.  But since you return null after the addError, I didn't think it would cause the error to get added to pageMessages.

src/classes/EP_EngagementPlans_TDTM.cls
njjc: Whoops, good catch.

src/classes/EP_ManageEPTemplate_CTRL.cls
njjc: In my testing, addError added a pagemessage, but did not throw an exception, and the save went through. If I threw an exception in addition to addError, there were two pagemessages, so instead I returned null and did the validation before any additional processing, which shows the message both in pagemessages and at the field level.

davidhabib: ok, I didn't know addError would internally add it to page messages.

PR #2308, opened 2016-09-26
Feature/jen lex navigation

jjbennett:
# Critical Changes

# Changes

# Issues Closed


jjbennett:
@davidhabib this was the change we spoke about this morning. Removing the showPills attribute.

PR #2311, opened 2016-09-29
Adding field set support to Manage EP Template page, adding new/edit …

njjc:
…override for EP Template.
# Critical Changes

# Changes
- Added new/edit override to use custom UI for Engagement Plan Templates.
- Added field set support to Manage Engagement Plan Template UI.

# Issues Closed

### DIFF
--- a/src/package.xml
+++ b/src/package.xml
@@ -1479,11 +1479,10 @@
         <name>CustomLabel</name>
     </types>
     <types>
-        <name>CustomLabel</name>
+        <members>CustomLabels</members>
+        <name>CustomLabels</name>
     </types>
     <types>
-        <members>Account</members>
-        <members>Activity</members>
         <members>Addr_Verification_Settings__c</members>
         <members>Address_Verification_Settings__c</members>
         <members>Address__c</members>
@@ -1491,7 +1490,6 @@
         <members>Allocations_Settings__c</members>
         <members>Batch_Data_Entry_Settings__c</members>
         <members>Batch__c</members>
-        <members>Contact</members>
         <members>DataImport__c</members>
         <members>Data_Import_Settings__c</members>
         <members>Engagement_Plan_Task__c</members>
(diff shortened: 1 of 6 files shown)


src/package.xml
```diff
@@ -1479,19 +1479,17 @@
         <name>CustomLabel</name>
     </types>
     <types>
-        <name>CustomLabel</name>
+        <members>CustomLabels</members>
+        <name>CustomLabels</name>
     </types>
     <types>
-        <members>Account</members>
```
davidhabib: were all these standard objects supposed to be deleted from package.xml?  did you remember to run ant updatePackageXml?

src/pages/EP_ManageEPTemplate.page
```diff
@@ -89,7 +105,7 @@
                     </div>
                 </apex:outputPanel>
             </form>
-            <p class="slds-text-heading--medium">Manage Engagement Plan Tasks</p>
+            <p class="slds-text-heading--medium slds-p-top--medium">Manage Engagement Plan Tasks</p>
```
davidhabib: needs to be a custom label, which reminds me that I gave a bunch of feedback in your original EP pull request that I let go in, since you were gone, but wanted you to address when you returned.  please look at the closed pull request for EP's to see what everything was.  nothing major, or I wouldn't have accepted the merge.  thanks.

src/pages/EP_ManageEPTemplate.page
```diff
@@ -35,21 +35,37 @@
...
+                                        styleClass="{!CASE(field.Type, 
+                                                'reference', 'lookupInput203 ' + field.FieldPath, 
+                                                'picklist', 'slds-select',
+                                                'multipicklist', 'slds-select', 
+                                                'boolean', 'slds-checkbox',
+                                                'textarea', 'slds-textarea',
+                                                'slds-input')}" />
+                                </div>
+                            </div>
+                        </apex:repeat>
```
davidhabib: in my use of field sets, I've typically seen if the field is required in either the DB or just in the field set (using isRequired, isDBRequired), and then added a '*' and appropriate markup to make required.  Now I realize in this pull request you are removing some of your required's, I assume to support the delete button.  But even if we don't have SLDS or visualforce enforce the requirement, should we still include the red '*' so the user believes it is required, and that our controller's save validation will check?  You can see an example of this here: https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/pages/LVL_LevelEdit.page#L142

PR #2318, opened 2016-10-04
Feature/hh naming fixes

### LINKED ISSUE #1727: Blanking out a household account naming field doesn't recalculate name
e.g. [here](https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/HH_HouseholdNaming.cls#L342) it only calculates if it's nulled AND auto naming is enabled.

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixes #1727

### DIFF
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -271,11 +271,11 @@
     updateNamingExclusions: function(component, hh) {
         var strExclusions = '';
         if (!component.get('v.isAutoName'))
-            strExclusions += 'Household__c.Name;';
+            strExclusions += 'Name;';
         if (!component.get('v.isAutoFormalGreeting'))
-            strExclusions += 'Household__c.Formal_Greeting__c;';
+            strExclusions += 'Formal_Greeting__c;';
         if (!component.get('v.isAutoInformalGreeting'))
-            strExclusions += 'Household__c.Informal_Greeting__c;';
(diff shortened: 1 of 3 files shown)


davidhabib:
@njjc 1 fix ready for you.  thx.

src/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -271,11 +271,11 @@
     updateNamingExclusions: function(component, hh) {
         var strExclusions = '';
         if (!component.get('v.isAutoName'))
-            strExclusions += 'Household__c.Name;';
```
njjc: Why this change? Haven't we always been using these weird strings? While I prefer to remove the Household__c portion, I'm worried about the possible side effects of changing this without a good mitigation strategy.

davidhabib: turns out the picklist values don't include Household__c.  the just had the field names.  The managehouseholdui code (from ryan) was setting household__c as well, but it didn't matter, because the code that looks at this field, was just doing a contains(fieldname).

So I fixed up the new managehouseholdui code to set the correct picklist values.

davidhabib:
now also includes the Level test assertion fix when contact's are modified by workflow rules during insert.

src/aura/HH_Container/HH_ContainerHelper.js
davidhabib: nope.  the code detects them by using contains(fieldname)

PR #2319, opened 2016-10-05
Adding conditions to contacts loop to avoid a different number of rec…

### LINKED ISSUE #1962: Bug when excluding record types from Householding code
I think I've found an error in `HH_Households_TDTM#insertHousehold` starting at line [371](https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/HH_Households_TDTM.cls#L371) that is evidenced when you insert multiple Contacts of different RecordTypes (some excluded, some not).

I don't have a test case for this (I found it when trying to debug a different issue), but I think you can easily generate one.

Let's say we don't Household Contacts of RecordType X, and we do household Contacts of RecordType Y, and we insert two Contacts (one of each type).  For convenience, let's call them "Noel" (No Household) and "Yasmin" (Yes Household).

If you pass in an array of `[Noel, Yasmin]` to the `insertHousehold` method, you'll end up with one entry for Yasmin's Household in `householdInserts`.  `lsr` will then have one record.  The `for` loop on line 371 will run twice.  On the first run, Noel will be assigned to Yasmin's Household.  On the second run, it will error on line 373 when we try to access `lsr[1]`.

njjc:
…ords in the trigger set vs save results. 
# Critical Changes

# Changes

# Issues Closed
- Fixes #1962


njjc:
@davidhabib this is ready for review, I couldn't write a regression test case due to the necessity of contact record types to test this.

PR #2322, opened 2016-10-10
Improving Manage EP Template UI accessibility: adding aria-labeledby …

### LINKED ISSUE #2321: Enforce Required on Priority field on Engagement Plan Task
thread: https://powerofus.force.com/0D580000030CD8v

The Visualforce page does not allow option to set Priority to None, but this can be done easily in standard edit. We should enforce that the Priority field on Engagement Plan Task is always populated, even if not editing from VF page.

njjc:
…tags for input fields with help text, replacing p tags with h1 tags for titles, and adding screen reader tags for extra context on Add Dependent Task and Delete Task buttons.
# Critical Changes

# Changes
- This release improves accessibility for Engagement Plan Templates UI.

# Issues Closed
- Fixes #2321

### DIFF
--- a/src/classes/EP_EngagementPlans_TEST.cls
+++ b/src/classes/EP_EngagementPlans_TEST.cls
@@ -65,8 +65,7 @@ private with sharing class EP_EngagementPlans_TEST {
             Reminder_Time__c = '660',
             Send_Email__c = true,
             Days_After__c = 3,
-            Type__c = 'Call',
-            Priority__c = 'Medium'
+            Type__c = 'Call'
         );
         insert epTask;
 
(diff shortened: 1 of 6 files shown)


jjbennett:
@njjc With voiceover turned on it does read the button first and then the accessibility text : "Delete Task Task 1" Maybe not having Task repeat like you were talking about earlier would be better. Should we also add a title to the page and is it possible to add a legend or something similar to each grouping?

njjc:
I think "Delete Task 1" sounds good. I'm not sure about "Add Dependent Task 1", how do you feel about "Add Dependent Task to Task 1"?

Title to the page, will do. What do you mean "legend for each grouping"? For each task? Those already have the title at the top.

jjbennett:
It didn't seem to read the title. It just told me I entered a grouping and didn't say what grouping. This is when I was editing an existing template and entered the Task 1 grouping and then the next Task 1-1 grouping.

njjc:
@jjbennett This is good to go!

PR #2325, opened 2016-10-17
Feature/2173 jen svgs

jjbennett:
# Critical Changes

# Changes

# Issues Closed


jjbennett:
@davidhabib I didn't close the icon issue since this will not completely resolve that issue, with the two outstanding bugs with svgs. Please take a look when you have time and let me know what you think, the newer version is a little smaller with some of the icons and with the address component.

src/aura/HH_Container/HH_Container.cmp
```diff
@@ -56,14 +56,13 @@
                         <nav role="navigation" class="slds-p-bottom--x-small">
                             <p id="bread-crumb-label" class="slds-assistive-text">You are here:</p>
                             <ol class="slds-breadcrumb slds-list--horizontal" aria-labelledby="bread-crumb-label">
-                                <li class="slds-breadcrumb__item slds-text-heading--label"><a class="slds-type-focus" href="{!'/' + v.hhTypePrefix}">{!$Label.c.lblHousehold}</a></li>
```
davidhabib: why did you remove slds-type-focus?  I believe that was giving me the hover hyperlink underline that no longer works.

src/aura/HH_AutoComplete/HH_AutoComplete.cmp
```diff
@@ -29,13 +29,8 @@
...
-                <ui:button class="slds-button" press="{!c.handleNewContact}">
-                    <c:svg ariaHidden="true"
-                           class="slds-icon slds-icon-text-default slds-icon--small"
-                           xlinkHref="{!$Resource.sfdo203 + '/assets/icons/utility-sprite/svg/symbols.svg#add'}"
-                           />
-                    {!$Label.npo02.NewContact}
-                </ui:button>
+                <lightning:button variant="bare" label="{!$Label.npo02.NewContact}"
+                    iconName="utility:add" iconPosition="left" onclick="{!c.handleNewContact}" />
             </div>
```
davidhabib: for some reason with your changes, the loadingIndicator spinner now displays in the middle of the main window, rather than just within the autocomplete component.  I verified this isn't happening in my other DE org on Winter '17 without your changes.  I suspect its due to moving from our scoped SLDS to the system one.  Note sure what we need to change to get the spinner in the correct location.

davidhabib:
@jjbennett don't you also need to modify the visualforce page HH_ManageHH to no longer include the scoped version of SLDS?  it looks like it is still including our scoped version:

<apex:stylesheet value="{!URLFOR($Resource.sfdo203, '/assets/styles/salesforce-lightning-design-system-vf.min.css')}" />

and 

<div class="sfdo203">

jjbennett:
@davidhabib I will double check removing it from the VF page but if I recall, it could load part of the page before the lightning and with using slds directly on the page might show up in the background without the styles.

src/aura/HH_AutoCompleteOption/HH_AutoCompleteOption.cmp
```diff
@@ -3,29 +3,14 @@
...
-                               xlinkHref="{!$Resource.sfdo203 + '/assets/icons/standard-sprite/svg/symbols.svg#household'}"
-                               />
-                        <aura:set attribute="else">                            
-                            <c:svg ariaHidden="true"
-                                   class="slds-icon slds-icon--small slds-icon-standard-contact"
-                                   xlinkHref="{!$Resource.sfdo203 + '/assets/icons/standard-sprite/svg/symbols.svg#contact'}"
-                                   />
-                        </aura:set>
-                    </aura:if>
+                    <lightning:icon iconName="{!'standard:' + (!empty(v.value.HHId__c) ? 'household' : 'contact')}"
```
jjbennett: I will double check this as well, but in my org I'm definitely seeing 2 different icons in my org.

davidhabib:
so I just checked, and this version still doesn't render icons on IE10 or Edge.  is that as expected?  I can't recall after all the back and forth we've had on this issue :-)

jjbennett:
@davidhabib they should render in edge in classic - there is a sf bug with lightning out on VF with SVGs in Lightning Experience.

jjbennett:
I am unable to continue to work on this branch. The lightning components are getting corrupt in my dev orgs, both new and old. I'm going to close this one and create a new one using appUnstyled to fix the WI17 issue with the buttons for now. Will revisit the svg issue again later.

PR #2326, opened 2016-10-17
Removing NPSP Scheduler, implementing default Scheduler.

njjc:
# Critical Changes

# Changes
- The NPSP Scheduler object has been removed and the npsp__Schedulable__c object has been deprecated. NPSP's default scheduled jobs are now scheduled individually through Salesforce. Administrators can now change the time those scheduled jobs run under Salesforce Setup.

- NPSP has a new settings group: Bulk Data Processes > Process Scheduler. This setting allows disabling NPSP from auto scheduling default recurring jobs. If you had previously disabled any recurring jobs in the npsp__Schedulable__c custom object, this setting will be enabled and those jobs will not be created.

- The NPSP Setting to Disable Error Handling has been removed.  NPSP now defers error handling and transaction control to Salesforce, and logs any errors outside of transaction control.

# Issues Closed


njjc:
@davidhabib make sure to add the w=1 param to the URL to ignore whitespace changes.

davidhabib:
@njjc now that we are supporting running hourly jobs, do we have an issue that if a job takes more than an hour to run, we will keep kicking it off, and it will never finish?  We avoided this before because our granularity was once a day, and everything finished by the next day.

njjc:
You're right that we don't have anything in place to detect the status of the jobs that have already ran, or potentially are currently running. I've added a backlog item to investigate enhancements to the scheduler to do that.

davidhabib:
waiting to accept until we hear back from @ceiroa about why our code did not schedule the main schedule jobs in the install script, but only did it when going to NPSP settings.

ceiroa:
@njjc @davidhabib I have @mentioned you in two Chatter threads where we had discussed the schedulable issues a while back. Hope this helps.

PR #2338, opened 2016-10-31
Friendly Error Messages

njjc:
# Critical Changes

# Changes
- Validation rules and required fields now display more user friendly error messages.

# Issues Closed

### DIFF
--- a/src/classes/ERR_ExceptionHandler.cls
+++ b/src/classes/ERR_ExceptionHandler.cls
@@ -37,14 +37,14 @@
 public with sharing class ERR_ExceptionHandler {
     /** @description Map of exception status codes to friendly error message.
     */
-    private static final Map<String, String> exceptionMessages = new Map<String, String>{
+    private static final Map<String, String> EXCEPTION_MESSAGES = new Map<String, String>{
         'FIELD_CUSTOM_VALIDATION_EXCEPTION' => Label.exceptionValidationRule,
         'REQUIRED_FIELD_MISSING' => Label.exceptionRequiredField
     };
 
     /** @description Map of exception status codes to the string that defines the start and end of the useful part of the message
     */
-    private static final Map<String, List<String>> subStrings = new Map<String, List<String>>{
+    private static final Map<String, List<String>> SUBSTRINGS = new Map<String, List<String>>{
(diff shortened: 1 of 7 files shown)


src/classes/ERR_ExceptionHandler.cls
```diff
@@ -0,0 +1,16 @@
+public with sharing class ERR_ExceptionHandler {
+    private static final Map<String, String> exceptionMessages = new Map<String, String>{
+        'FIELD_CUSTOM_VALIDATION_EXCEPTION' => 'A validation rule is preventing the record from saving: ',
```
njjc: The values of this map would be custom labels in a final implementation.

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -0,0 +1,16 @@
...
+        'REQUIRED_FIELD_MISSING' => 'A required field is missing: '
+    };
+
+    public static void beautifyException(Exception ex) {
+        String message = ex.getMessage();
+        for (String status : exceptionMessages.keySet()) {
+            if (message.contains(status)) {
+                message = exceptionMessages.get(status) + message.substringBetween(status, ':');
```
davidhabib: so does this end up still preserving the name of the field?  I think it does, but wanted to check.

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -107,6 +124,8 @@ public class TDTM_TriggerHandler {
                 UTIL_Debug.debugWithInfo('****Finished executing: ' + classToCall.get('Class__c'));
             }
 
+            depthCounter++;
```
davidhabib: when does this get decremented?

njjc:
This isn't quite working yet, please hold off on comments until I can fix it.

ceiroa:
Please keep me posted on when this is ready for review. I'd like to get my head wrapped around this. Thanks!

njjc:
@jjbennett this is ready for review. I couldn't think of a way to get a handled exception through a trigger in a test, so I just wrote unit tests and not end to end tests. Maybe we can brainstorm a way to do that together.

PR #2339, opened 2016-10-31
Feature/jen lex nav

jjbennett:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/ADDR_Validator.cls
+++ b/src/classes/ADDR_Validator.cls
@@ -47,11 +47,8 @@ public with sharing class ADDR_Validator {
         if (addrId != null) {
             set<id> setAddrId = new Set<Id>{addrId};
             verifyRecords(setAddrId);
-            
-            PageReference pageRef = new PageReference('/' + addrId);
-            pageRef.setRedirect(true);
-            return pageRef;
-        } else return null;
+        }
+        return null;
     }
 
     /*******************************************************************************************************
--- a/src/classes/ALLO_RollupBTN_CTRL.cls
+++ b/src/classes/ALLO_RollupBTN_CTRL.cls
@@ -50,10 +50,7 @@ public with sharing class ALLO_RollupBTN_CTRL {
         	//run rollups for this General Accouting Unit
 			ALLO_Rollup_SCHED rollup = new ALLO_Rollup_SCHED(parentId);
 			rollup.runRollups();
-
-	        PageReference pageRef = sc.view();
-	        pageRef.setRedirect(true);
-	        return pageRef;
-        } else return null;
+        }
+        return null;
     }
 }
--- a/src/classes/OPP_OpportunityNamingBTN_CTRL.cls
+++ b/src/classes/OPP_OpportunityNamingBTN_CTRL.cls
@@ -37,6 +37,7 @@
 public with sharing class OPP_OpportunityNamingBTN_CTRL {
     
     private ApexPages.StandardController ctrl;
+    public Boolean redirect {get; set;}
 
     /** @description Standard opportunity controller constructor.*/
     public OPP_OpportunityNamingBTN_CTRL(ApexPages.StandardController oppCtrl) {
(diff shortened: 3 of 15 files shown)


jjbennett:
@njjc @davidhabib Even though we are not going to merge this in right away, I wanted to see if you could review how I'm handling the LEx navigation issue when modifying the current object with a vf button so we can discuss.

src/components/UTIL_NavigateBack.component
```diff
@@ -0,0 +1,28 @@
...
+        type="Boolean"
+        default="false"
+        />
+
+    <script type="text/javascript">
+        var redirect = {!redirect};
+        if (redirect) {
+            if ((typeof sforce != 'undefined') && sforce && (!!sforce.one)) {
+                // Manage navigation in Lightning Experience & Salesforce1
+                    sforce.one.back(true);
```
jjbennett: sforce.one.back() takes a refresh parameter, which I set to true above which is what forces the page to refresh. There doesn't seem to be an equivalent refresh for any other method.

davidhabib:
@jjbennett @njjc OK, i've reviewed this and am comfortable with it being merged in after the code freeze is lifted.  thanks.  @njjc you should review as well.  i will probably need to employ the same solution in the Mass Edit Volunteer Hours page in V4S.

jjbennett:
@davidhabib @njjc since this wasn't merged yet, I just pushed the fix for the payment wizard page as well.

src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -231,9 +236,8 @@ public with sharing class PMT_PaymentWizard_CTRL {
     /*******************************************************************************************************
     * @description ActionMethod for the Calculate button that given the current opp outstanding balance, 
     * creates the specified new payments to cover the amount.
-    * @return PageReference null
     */ 
-    public pageReference calculate() {
+    public void calculate() {
```
davidhabib: why did you change all these action methods to be void, rather than PageReference and returning null?  Seems incorrect to not have the signature match what action methods are supposed to be.

jjbennett: My understanding is that if a page reference is never expected to be return the method should be void and not a null page reference?

davidhabib:
can you find that in any official doc?  i've never heard of this.


njjc:
I don't see anything in the official docs, but lots of other references say to use void if the return is always null, like [this strongly worded stackexchange post](http://salesforce.stackexchange.com/questions/63286/best-practices-page-reference-method-return-null).

src/classes/PMT_PaymentWizard_CTRL.cls
jjbennett: I'm not finding specifics in the docs around it, but it does state PageReferences should be used for:
To view or set query string parameters and values for a page
To navigate the user to a different page as the result of an action method

davidhabib:
first time i've seen it discussed.  still seems weird to me that one would not follow the official signature of what an actionMethod is, and I find it useful to see PageReference as calling out that it is called from the visualforce page, even if it will always return null.

jjbennett:
From everything I can find, it appears to be more a matter of preference. I've returned to returning null.

src/classes/ADDR_Validator.cls
```diff
@@ -47,11 +47,8 @@ public with sharing class ADDR_Validator {
         if (addrId != null) {
             set<id> setAddrId = new Set<Id>{addrId};
             verifyRecords(setAddrId);
-            
-            PageReference pageRef = new PageReference('/' + addrId);
-            pageRef.setRedirect(true);
-            return pageRef;
-        } else return null;
+        }
```
davidhabib: this page is really just for a button, I think, and it used to redirect back to the address object.  by now just returning null, I think it might land on the empty visualforce page.

src/classes/ALLO_RollupBTN_CTRL.cls
```diff
@@ -50,10 +50,7 @@ public with sharing class ALLO_RollupBTN_CTRL {
         	//run rollups for this General Accouting Unit
 			ALLO_Rollup_SCHED rollup = new ALLO_Rollup_SCHED(parentId);
 			rollup.runRollups();
-
-	        PageReference pageRef = sc.view();
-	        pageRef.setRedirect(true);
-	        return pageRef;
-        } else return null;
+        }
+        return null;
```
jjbennett: It needs to land on the page to allow the javascript to handle the redirect.

src/classes/OPP_OpportunityNamingBTN_CTRL.cls
```diff
@@ -48,14 +49,15 @@ public with sharing class OPP_OpportunityNamingBTN_CTRL {
...
-            update thisOpp;
-            
-            PageReference pageRef = new PageReference('/' + thisOpp[0].id);
-            pageRef.setRedirect(true);
-            return pageRef;
-        
-        } else return null;
+            try {
+                update thisOpp;
+                redirect = true;
```
davidhabib: this page used to redirect back to the opp page, but now I think it will end up on the empty button page.

src/classes/ADDR_Validator.cls
davidhabib: Ignore my last comment!  i forgot about the redirect work you added!!

PR #2345, opened 2016-11-03
Updating scheduler helper method to reschedule jobs owned by the NPSP…

njjc:
… user or an inactive user.
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/STG_SettingsManager_CTRL.cls
+++ b/src/classes/STG_SettingsManager_CTRL.cls
@@ -69,6 +69,18 @@ public with sharing class STG_SettingsManager_CTRL extends STG_Panel {
         installScript.runNewOrgScript();
         //Set master schedulables config
         UTIL_MasterSchedulableHelper.abortOldScheduledJobs();
-        UTIL_MasterSchedulableHelper.setScheduledJobs();
+        if (runningUserIsAdmin()) {
+            UTIL_MasterSchedulableHelper.setScheduledJobs();
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description Returns whether the running user is a System Administrator.
+    * @return void
+    */
+    private boolean runningUserIsAdmin() {
+        Profile currentUserProfile = [SELECT Id, PermissionsCustomizeApplication, PermissionsModifyAllData 
+                                        FROM Profile WHERE Id = :UserInfo.getProfileId()];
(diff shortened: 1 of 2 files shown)


src/classes/STG_SettingsManager_CTRL.cls
```diff
@@ -69,6 +69,18 @@ public with sharing class STG_SettingsManager_CTRL extends STG_Panel {
...
+            UTIL_MasterSchedulableHelper.setScheduledJobs();
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description Returns whether the running user is a System Administrator.
+    * @return void
+    */
+    private boolean runningUserIsAdmin() {
+        Profile currentUserProfile = [SELECT Id, PermissionsCustomizeApplication, PermissionsModifyAllData
```
davidhabib: is there any chance the user running this query might not have perms on Profile and it would throw an exception?  maybe we should wrap the entire query in a try/catch, and return false if it throws?

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -122,22 +122,32 @@ public without sharing class UTIL_MasterSchedulableHelper {
...
-        Set<String> scheduledJobs = new Set<String>();
-        //query for the job names of all scheduled apex
-        for (CronTrigger ct : [SELECT CronJobDetail.Name FROM CronTrigger WHERE CronJobDetail.JobType = '7']) {
-            scheduledJobs.add(ct.CronJobDetail.Name);
+        Map<String, CronTrigger> scheduledJobs = new Map<String, CronTrigger>();
+        
+        //query for all scheduled apex
+        for (CronTrigger ct : [SELECT Id, CronExpression, CronJobDetail.Name, CreatedBy.Name, CreatedBy.IsActive FROM CronTrigger WHERE CronJobDetail.JobType = '7']) {
+            scheduledJobs.put(ct.CronJobDetail.Name, ct);
         }
```
davidhabib: any chance querying CronTrigger will throw an exception we need to handle?

davidhabib: doesn't with sharing just change whether you can see a specific record (ie, private vs public vs owned by someone else), rather than changing your permissions (whether you can see a table and field).

PR #2350, opened 2016-11-07
Adding clone override handling.

njjc:
# Critical Changes

# Changes
- Manage Engagement Plan Tasks UI now supports setting the page as Clone button override. Using the overridden Clone button clones the EP Template and its EP Tasks, allowing users to easily create new templates by modifying existing ones. Admins should set the Clone button override to use this functionality.

# Issues Closed


njjc:
@jjbennett this is ready when you are.

PR #2360, opened 2016-11-14
Feature/dataimport errors

davidhabib:
There is no gitHub issue for this, as it was not user reported.  

# Critical Changes

# Changes
- When using the NPSP Data Importer, if a contact fails to create its account, NPSP will now flag just that specific contact with the error, rather than failing all contacts in the batch.

# Issues Closed


davidhabib:
@njjc ready for your review and merge.

PR #2376, opened 2016-11-18
Updating error email message and opp naming error handling.

njjc:
# Critical Changes

# Changes
- NPSP Error emails now include a friendlier and more helpful message.
- Errors with Opportunity Naming no longer cause Opportunity inserts to fail with an exception. NPSP Error messages will be sent instead.

# Issues Closed


jjbennett:
@njjc looks like the new label is missing from the package.

PR #2380, opened 2016-11-21
Feature/managehh name fixes

davidhabib:
# Critical Changes

# Changes

# Issues Closed


davidhabib:
@njjc This is just a trivial fix I found while looking at other issues.  no bug in gitHub about it.  basically, the only time the contact's household naming order gets saved is if they do drag/drop and save from the ManageHH page.  This fix just initially sets household naming order when contacts are loaded, so the order will match the naming logic.

PR #2390, opened 2016-11-22
support Clone button override for Level. #2356

### LINKED ISSUE #2356: Clone button is Editing Current Level
When a user clicks the Clone button with the Level override page set on Clone and then modifies the Level and Saves, the original Level is being saved and a new Level is not created.

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixes #2356

### DIFF
--- a/src/classes/LVL_LevelEdit_CTRL.cls
+++ b/src/classes/LVL_LevelEdit_CTRL.cls
@@ -48,6 +48,13 @@ public with sharing class LVL_LevelEdit_CTRL {
             controller.addFields(UTIL_Describe.listStrFromFieldSet(UTIL_Namespace.StrTokenNSPrefix('Level__c'), UTIL_Namespace.StrTokenNSPrefix('LevelEdit')));
         }
         lvl = (Level__c)controller.getRecord();
+        
+        //if clone param is set, this is the clone button override
+        Map<String,String> params = ApexPages.currentPage().getParameters();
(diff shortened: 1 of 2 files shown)


src/classes/LVL_LevelEdit_CTRL.cls
```diff
@@ -48,6 +48,13 @@ public with sharing class LVL_LevelEdit_CTRL {
             controller.addFields(UTIL_Describe.listStrFromFieldSet(UTIL_Namespace.StrTokenNSPrefix('Level__c'), UTIL_Namespace.StrTokenNSPrefix('LevelEdit')));
         }
         lvl = (Level__c)controller.getRecord();
+        
+        //if clone param is set, this is the clone button override
+        Map<String,String> params = ApexPages.currentPage().getParameters();
+        if (params.containsKey('clone') && params.get('clone') == '1') {
+            lvl.Id = null;
+            lvl.Name = null;
```
njjc: I kept the original name and appended " - Cloned" to make it clear what has happened.

njjc: https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/EP_ManageEPTemplate_CTRL.cls#L142

davidhabib: great idea.  i'll do that.

PR #2392, opened 2016-11-23
Default address not resetting on Seasonal End Dates #2358

### LINKED ISSUE #2358: Default address not resetting on Seasonal End Dates
Seasonal addresses are not ending on scheduled end date.

davidhabib:
I've redesigned how we do the scheduled batch job for seasonal addresses.  previously, it just looked for seasonal addresses that became active on the current day.  That's flawed for two reasons; if the job doesn't run one day, the change would not happen, and we never detected when seasonal addresses no longer should be used.

the new strategy finds all accounts who have any seasonal addresses associated with them, then calls existing logic to find each accounts current default and potential seasonal.  it then sees if the appropriate address matches what's in the account or not.  If not, it then calls existing logic to update the HH and Contacts with the appropriate address.  

# Critical Changes

# Changes

# Issues Closed
Fixes #2358

### DIFF
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -822,12 +822,15 @@
     },
 
     /*******************************************************************************************************
-     * @description adds or removes namespace prefix from the object's custom fields
+     * @description adds or removes namespace prefix from the object's custom fields.  for non-namespaced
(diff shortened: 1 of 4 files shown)


davidhabib:
oops!  I accidentally included my fix to namespace issues in Manage Household UI, rather than putting it in a separate pull request.  Given the fix is just several lines of code all isolated in one class method, I'm hoping you are OK with that @njjc .  Thanks!

PR #2393, opened 2016-11-28
Feature/1922 leads

### LINKED ISSUE #1571: Warn users in lead conversion if profile default account record type is same as Household/One to One
If default is set to same record type as is used for Household or One to One, then Affiliate accounts created in lead conversion will have that record type be treated as individual account in reporting and naming.

Could be in Health Check too, but then would have to check against all profiles?

### LINKED ISSUE #1922: BUG: Lead Conversion page does not respect Lead Process Status selected values

### LINKED ISSUE #2378: If Opp created in Lead Conversion, Opp Automatic Naming does not fire
1. Set up NPSP Opp Naming for 'Both' and select a standard naming rule for profile's default Opp record type
2. Create Lead using 'Self' as Company
3. Convert using NPSP override - uncheck 'Do not create opportunity...' box

Result: Lead is converted correctly to Contact/Household, however Opportunity has name as set in Lead Conversion dialog and is not being updated to use naming rules as expected on Opportunity save. Admin must 'Refresh Name' to update.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #1922 
Fixes #1571 
Fixes #2378

### DIFF
--- a/src/classes/LD_LeadConvertOverride_CTRL.cls
+++ b/src/classes/LD_LeadConvertOverride_CTRL.cls
@@ -311,6 +311,17 @@ public with sharing class LD_LeadConvertOverride_CTRL {
             strConId = 'NEW_CONTACT';
         if (!isSpecialAccount)
             strAfflAccountId = listSOAfflAccounts[1].getValue(); // either the first existing or the new account if none.
+
+        // warn user if default record type is the same as the HH Account Record Type
+        Id recTypeId = UTIL_CustomSettingsFacade.getContactsSettings().npe01__HH_Account_RecordTypeID__c;
+        if (recTypeId != null) {
+            sObjectType objType = Account.sObjectType;
+            String recTypeName = UTIL_RecordTypes.getRecordTypeName(objType, recTypeId);
+            if (UTIL_RecordTypes.isRecordTypeDefault(objType, recTypeName)) {
+                apexPages.addMessage(new ApexPages.Message(ApexPages.Severity.Warning,
+                        Label.leadConvertRecordTypeWarning));
+            }
+        }
     }
     
     /*******************************************************************************************************
@@ -338,7 +349,7 @@ public with sharing class LD_LeadConvertOverride_CTRL {
     	try {
 	        if (strConId == null) {
 	            apexPages.addMessage(new ApexPages.Message(ApexPages.Severity.Error,
-	                'You must select an option in the Contact box.'));
+	                label.leadConvertContactSelection));
 	            return null;
 	        }
 
--- a/src/classes/LD_LeadConvertOverride_CTRL.cls
+++ b/src/classes/LD_LeadConvertOverride_CTRL.cls
@@ -85,20 +85,6 @@ public with sharing class LD_LeadConvertOverride_CTRL {
     public boolean doNotCreateOpp { get; set; }
 
     /*******************************************************************************************************
-    * @description Shows the value of the lead status
-    */
-    public string leadConvStatus {
-    	get {
-    		if (leadConvStatus == null || leadConvStatus == '') {
-                if (lcStatuses != null && lcStatuses.size() > 0)
-                    leadConvStatus = lcStatuses[0].getValue();
-    		}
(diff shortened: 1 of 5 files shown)


jjbennett:
@njjc @davidhabib this PR is ready, Nic adding you to get your thoughts as well on how I'm resolving the available status by record type via JS on LeadConvertOverride.

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -361,7 +358,7 @@ public with sharing class LD_LeadConvertOverride_CTRL {
 	        lc.setLeadId(l.Id);
 	        if (!doNotCreateOpp && Schema.SObjectType.Opportunity.isCreateable()) lc.setOpportunityName(oppName);
 	        lc.setDoNotCreateOpportunity(!Schema.SObjectType.Opportunity.isCreateable() || doNotCreateOpp);
-	        lc.setConvertedStatus(leadConvStatus);
+	        lc.setConvertedStatus(l.Status);
```
jjbennett: The default converted status will be set on the page, the user can change it if there are multiple options. Whatever is chosen will be used.

src/classes/LD_LeadConvertOverride_TEST.cls
```diff
@@ -91,15 +91,19 @@ private class LD_LeadConvertOverride_TEST {
...
         system.assertEquals(1, ext.listSOAccounts.size());
         system.assertEquals(1, ext.listSOAfflAccounts.size());
 
         // this selects the "New" option in the Contact box
         ext.strConId = 'NEW_CONTACT';
         ext.strAccountId = 'AUTO';
+
         Test.startTest();
-        ext.convertLeadNow();
+        for (SelectOption lcStatus : ext.lcStatuses) {
```
davidhabib: I'm guessing the tests did this loop, because they didn't know what status value would actually convert the lead.  are you certain that convertLeadNow() will always succeed?

src/pages/LD_LeadConvertOverride.page
```diff
@@ -231,9 +230,18 @@
         </apex:form>
 
     </div>
-
-    <apex:outputField value="{!lead.company}" rendered="false" />
-    <apex:outputField value="{!lead.name}" rendered="false" />
-    <apex:outputField value="{!lead.Email}" rendered="false" />
-    <apex:outputField value="{!lead.OwnerId}" rendered="false"/>
+    <script>
+        window.onload = function filterStatus() {
```
davidhabib: a comment header on what this function is trying to do would be helpful.

src/classes/LD_LeadConvertOverride_TEST.cls
davidhabib: my bad!  I thought the old version of the test had this loop and you removed it!  So you are doing this because you aren't sure what is the valid converted status to use?  is there really no way other than trying them all?

src/classes/LD_LeadConvertOverride_CTRL.cls
davidhabib: got it.  this Status is coming from what the user chooses on the page.

src/classes/LD_LeadConvertOverride_TEST.cls
jjbennett: I cannot find a way since the page never really loads in the test, it doesn't know the valid values based on Visualforce filtering by recordType for us. The only way I could find to get this information in Apex was by using that metadata api workaround. I'm also working under the assumption that there won't be too many orgs set up with too many converted statuses based on record types.

davidhabib: got it.  thanks.

davidhabib:
ok, I'm merging this even though I can't get the build to run on it! (repeat [sf:deploy] Request Status: Pending for the next hour!).  I've deployed it to my own org and tested it, and code reviewed it.

PR #2400, opened 2016-12-04
Added generic cascade deletion trigger handler.

### LINKED ISSUE #2377: Recurring Donation is not deleted if Account/Contact are deleted
We are seeing users who are getting NPSP Error emails with the error "Attempt to de-reference a null object"

This appears to consistently be because the Organization and Contact are blank on a Recurring Donation record.

I don't believe this is bad data entry. When you insert a Recurring Donation without an Organization or Contact, you get an error. 

This is a consistent way to reproduce the error and may be why our users are hitting it - when an Account (or Contact) connected to the Recurring Donation is deleted, the Recurring Donation is left behind! The nightly batch is attempting to roll up that now orphaned Recurring Donation record and getting the exception.

Steps to repeat:

1. Create Organization Account (or Contact)
2. Create Recurring Donation, open, $1, Yearly Installment Period (so there's only 1 opp created)
3. Delete Organization Account (or Contact)
4. Run Recurring Donation Rollup - get 'Attempt to de-reference a null object."

Recurring Donation is not deleted as part of the cascade delete of Account/Contact (even though the Opportunities are deleted). 

Delete behavior should remove orphaned Recurring Donation, and/or those records should not be calculated in the batch update.

### LINKED ISSUE #2382: Deleting an opportunity with allocations should delete the allocation

njjc:
# Critical Changes

# Changes

# Issues Closed
- Fixed #2377 
- Fixed #2382


njjc:
@jjbennett @davidhabib I created this handler for cascade deletion over lookup fields. I know it's a standard option now, but that's only when creating new lookup fields, so we needed a way to cascade delete via lookup on objects already in our package.

I'd love to get your thoughts on [this part](https://github.com/SalesforceFoundation/Cumulus/pull/2400/files#diff-26720effda6dd6fb08588d4daf16a9c2R52). We could easily make the design user extensible by using a custom setting instead of the private map, does that seem worthwhile to users, or should we just keep it as an internal utility?

njjc:
@davidhabib Jen and I met about this, we thought it a fine idea to keep the values hard coded for now, and that we can easily extend it to a custom setting at a later date if we decide it could be useful. This is ready for review.

njjc:
Just realized that Jen was assigned in AA so switching this over to her.

src/classes/UTIL_CascadeDeleteLookups_TDTM.cls
```diff
@@ -0,0 +1,131 @@
...
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist, 
+        TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+
+        if (cascades.containsKey(objResult.getSObjectType())) {
+            for (String childLookup : cascades.get(objResult.getSObjectType())) {
+                try {
+                    cascadeDelete(newlist, oldlist, childLookup, triggerAction);
+                } catch (Exception ex) {
+                    ERR_Handler.processError(ex, 'TDTM');
```
davidhabib: this just logs the error.  wouldn't we expect this to propagate any exception instead?

davidhabib:
@njjc Nice job Nic on this.  I like the design.  Just one question on a place where you are not propagating exceptions.

src/classes/UTIL_CascadeDeleteLookups_TDTM.cls
njjc: I didn't think we'd want to throw an exception for an operation that would have succeeded but left behind orphaned records previously. Let's discuss.

njjc:
@davidhabib Jen just discovered that [cascade deleted objects don't invoke triggers](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_triggers_ignoring_operations.htm) which means, e.g. in the contact delete override that deletes the Account and relies on cascade to delete the single Contact, our trigger isn't invoked and the recurring donation related to the Contact isn't deleted since before/after delete of Contact triggers aren't invoked.

Basically this design won't work due to that issue. I'm trying to brainstorm a way to work around that.

davidhabib:
well we could certainly change CON_DeleteContactOverride_CTRL to always delete the contact first, but do we still have the problem elsewhere, where one of our Contacts, Accounts, or Opps are being deleted thru a parent's cascade?

njjc:
Yeah, exactly. This won't work on any operation where the object is being deleted through its parent's cascade.

njjc:
@jjbennett @davidhabib @mpusto While this design won't work for objects that are part of a cascade delete, it's still better than what we have now. 

I can't think of a good way to generically handle the case where the object being deleted doesn't have its triggers run due to standard cascade deletion, I need more time to think about it.

Should we put this in anyhow? It will save us some nightly recurring donations errors, and cascade delete orphaned objects it some cases.

njjc:
@jjbennett I FIGURED OUT A SOLUTION! Simple fix too! This should be good to go, I'd love to get it in before freeze.

PR #2404, opened 2016-12-05
Adding logic to make contact merge field selection ignore identical…

### LINKED ISSUE #2056: BUG: If a contact custom field with Name that matches NPSP field (after namespace) exists, error in Contact Merge and Manage Household
Thread: https://powerofus.force.com/0D58000002eAJYw

Steps to reproduce. 

Create custom Contact field - label Deceased, Name - Deceased (Deceased__c) 
Clone a contact, and then merge in Contact Merge tab

Error: Failed to query contact information. Error: System.QueryException: duplicate field selected: npsp__Deceased__c

njjc:
…API names to NPSP namespaced fields.
# Critical Changes

# Changes

# Issues Closed
- Fixed #2056


src/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -274,6 +274,14 @@ public with sharing class CON_ContactMerge_CTRL {
...
             Schema.DescribeFieldResult fResult = f.getDescribe();
             
+            // Ignore fields that have an identical API name as an npsp prefixed field
+            // This is due to a bug in dynamic SOQL, quich causes querying for two real fields, 
+            // e.g. npsp__foo__c and foo__c, to throw a duplicate field selected error
+            if (fieldName != UTIL_Namespace.StrTokenRemoveNSPrefix(fieldName) && 
+                contactFields.containsKey(UTIL_Namespace.StrTokenRemoveNSPrefix(fieldName))) {
+                continue;
+            }
+
```
davidhabib: Just so I'm clear, you are basically not adding to the soql the npsp__foo__c field if  they have an unmanaged foo__c field.  But which one will then get displayed & merged.  i think it will be their field, rather than the npsp one.  and this will happen regardless of what they have in any field set.  if they have a shadow field, it will be the one copied.  did I understand this correctly?  If so, I kinda hate that even if they put our npsp field in the field set, we will still end up only copying values between their shadow fields.  not sure on how to solve this though.  I'd prefer it to be if you have a field that tries to shadow the npsp field, we will ignore your field and use the npsp field.

njjc: My understanding is that dynamic SOQL resolves both fields, whether with or without namespace, to the NPSP namespaced field, which means no matter what we as package developers do, we can only show our version of the field in our UI. I could definitely update the logic to make that more clear.

src/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -274,6 +274,16 @@ public with sharing class CON_ContactMerge_CTRL {
...
+            // Ignore fields that have an identical API name as an npsp prefixed field
+            // This is due to a bug in dynamic SOQL, quich causes querying for two real fields, 
+            // e.g. npsp__foo__c and foo__c, to throw a duplicate field selected error.
+            // Only select the NPSP namespaced field and ignore the same field without namespace
+            if (fieldName.countMatches('__') == 1 && 
+                contactFields.containsKey(UTIL_Namespace.StrTokenNSPrefix(fieldName))) {
+                
+                continue;
+            }
+
```
davidhabib: in unmanaged code, won't this have the affect of ignoring all of our npsp fields?

njjc: You're right, I think I've finally got a good solution.

PR #2414, opened 2016-12-09
Feature/levels validation

### LINKED ISSUE #2407: User should receive an error message for Level with no min/max value
Create a Level whose values for Minimum Amount and Maximum Amount are empty. 

Attempt  to 'Clone' this Level.

Regardless of the amount set in the cloned Level, attempting to save the new Level yields error message "Error: The Minimum Amount to Maximum Amount range on Level <Foo> overlaps with the Minimum Amount to Maximum Amount range on Level <Foo - Cloned>"

It may be that we should prevent creating Levels with no min/max values.

screen shot 2016-12-06 at 2 23 25 pm

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixed #2407


davidhabib:
@njjc ready for review and merge.  thx.

PR #2420, opened 2016-12-20
Optimize trigger to not query if no RD’s referenced by the Opps in the trigger

davidhabib:
this is a non-critical improvement we can merge in any sprint.

# Critical Changes

# Changes

# Issues Closed


davidhabib:
@njjc ready for review and merge.

PR #2426, opened 2016-12-21
Prevent exception when legacy accounts have >50k contacts

### LINKED ISSUE #2425: Cannot delete Contact in Account with >50k Contacts
I have several legacy accounts with huge numbers of contacts. When users try to delete those contacts using `CON_DeleteContactOverride`, they get an exception in the form:

> npsp:Too many query rows: 50001
> Error is in expression '{!processDelete}' in component apex:page in page npsp:con_deletecontactoverride: (npsp)

## Steps to reproduce:

1. Create an Account
1. Create 60k contacts all in that account.
1. Click delete on one of those contacts.

## Expected Result:
That contact is deleted.

## Actual result:
Exception: 

> npsp:Too many query rows: 50001
> Error is in expression '{!processDelete}' in component apex:page in page npsp:con_deletecontactoverride: (npsp)

## Workaround:
Use a different override.

ckoppelman:
I have several legacy accounts with huge numbers of contacts.  When users try to delete those contacts using `CON_DeleteContactOverride`, they get an exception in the form:

> npsp:Too many query rows: 50001
> Error is in expression '{!processDelete}' in component <apex:page> in page npsp:con_deletecontactoverride: (npsp)

This patch aggregates the results instead of querying the 50k+ contacts that will never be used.

It's likely that `&& contactsInHousehold[0].get('ct') == 1` is redundant.

# Critical Changes

Fixes bug that can be duplicated as follows:

## Reproduce
1. Create an Account
1. Create 60k contacts all in that account.
1. Click delete on one of those contacts.

## Expected result:
Contact is deleted.

## Actual result:

> npsp:Too many query rows: 50001
> Error is in expression '{!processDelete}' in component <apex:page> in page npsp:con_deletecontactoverride: (npsp)

# Changes

# Issues Closed

#2425: **Cannot delete Contact in Account with >50k Contacts**

### DIFF
--- a/src/classes/CON_DeleteContactOverride_CTRL.cls
+++ b/src/classes/CON_DeleteContactOverride_CTRL.cls
@@ -63,19 +63,26 @@ public with sharing class CON_DeleteContactOverride_CTRL {
     public pageReference processDelete() {
         Contact queryContact = [SELECT Id, AccountId, Account.npe01__SYSTEMIsIndividual__c FROM Contact WHERE Id = :conId];
         string accId = queryContact.AccountId;
-        list<AggregateResult> contactsInHousehold = [SELECT COUNT(id) ct FROM Contact WHERE AccountId = :accId GROUP BY AccountId HAVING COUNT(id) = 1];
+        boolean shouldDeleteContactAlone = true;
 
-        //This contact is alone in a system account, delete the system account and allow the cascading
-        //delete to remove the contact
-        if (contactsInHousehold.size() == 1 && contactsInHousehold[0].get('ct') == 1 && queryContact.Account.npe01__SYSTEMIsIndividual__c) {
-            Account accForDelete = new Account(id=accId);
-            delete accForDelete;
+        if (queryContact.AccountId != null && queryContact.Account.npe01__SYSTEMIsIndividual__c) {
+            list<AggregateResult> contactsInHousehold = [SELECT COUNT(id) ct FROM Contact WHERE AccountId = :accId GROUP BY AccountId HAVING COUNT(id) = 1];
 
-            //if we were returning to the account we just deleted, go back to contacts home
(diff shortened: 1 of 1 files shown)


davidhabib:
@ckoppelman thanks for the pull request.  my understanding is that aggregate queries still count the number of rows they have to process over.  So even though your change is only returning one row, the fact that the underlying data has over 50k records will still be a problem.  I could be wrong, but that's my recollection.  Were you able to test your new query/code in your large org using either anonymous apex or the query editor in the dev console?

ckoppelman:
I have not yet deployed this to production (in process), but I ran it as anon Apex (well, technically, I ran the below) and was able to reproduce the error with a known mega-Account.  First I tried the simple `GROUP BY` which didn't help.  Adding the `HAVING` clause fixed it.

Separately (not addressed here), it might also be useful to ensure that the user may delete that Account on a permissions level.

ckoppelman:
@davidhabib Update - It is now deployed to production on my system and working just fine.

ckoppelman:
An even better solution is to only make that second SOQL query if `queryContact.Account.npe01__SYSTEMIsIndividual__c`.  That way you save a SOQL call.  I can modify my patch to do that if you'd like.

davidhabib:
thanks for testing and verifying it solved the problem in your org.  yes,
if you could update your pull request that would be great.  I'll then make
sure we get it assigned for further review and testing in our next sprint
(which will start in Jan), so we can release it thereafter.  thanks.


On Wed, Dec 21, 2016 at 1:31 PM, Charles Koppelman <notifications@github.com

ckoppelman:
@davidhabib done. I look forward to deleting my patch.

davidhabib:
I've cloned your pull request to a feature branch in our repository so it can run thru our build process.  I'll tag you on the new request.  thanks!

PR #2431, opened 2016-12-29
Fix RD query to be selective in LDV orgs.  Issue #2430

### LINKED ISSUE #2430: Recurring Donation non-selective query error in LDV orgs
We have a customer with over 8 million opportunities.  They are unable to successfully create Recurring Donations, because one of the queries our trigger handler uses is being deemed non-selective in this environment.  

The NPSP query is doing 

....from Opportunity where  npe03__Recurring_Donation__r.id IN ... 

which fails to use the lookup index.  by making it the following, it will use the index.

....from Opportunity where  npe03__Recurring_Donation__c IN ...

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixes #2430


davidhabib:
@njjc ready for review and merge

PR #2446, opened 2017-01-10
Data Importer fixes & enhancements

### LINKED ISSUE #2162: In Data Importer, if Campaign is specified and record fails to import, a dup Campaign record is created
Create and import a DI template file in which there is a row of data where:
- Donation Amount is specified  
- Campaign matches an existing SF Campaign
- Something about the row will cause the DI import to fail (ex: Contact First Name and Last Name will fail SF duplicate management rules).

Issue: After DI import, the failed record causes DI to create a duplicate Campaign record whose name is the value from the Campaign field in the template file.  For each failed record in a DI import, another dup Campaign record is created.

### LINKED ISSUE #2182: add Contact and Account Ids to NPSP import match criteria
https://powerofus.force.com/hub-ideas/1057

Based on this idea, allow users to match against the Salesforce Id if they have it, when using the NPSP Data Importer.

davidhabib:
# Critical Changes

# Changes

* The NPSP Data Importer now supports matching to existing Contacts, Accounts, and Household Accounts if you specify the Salesforce ID in the respective Lookup fields Contact1 Imported, Contact2 Imported, Account1 Imported, Account2 Imported, and Household Account Imported.  Note that this happens automatically if any of those lookup fields are populated; there is no picklist value to choose in Contact Custom Unique ID or Account Custom Unique ID.

# Issues Closed
Fixed #2162 
Fixed #2182


davidhabib:
@njjc This pull request is ready for review.

davidhabib:
@jessingrass @judisohn This is the first pull request for NPSP Data Importer which adds support for:
1. fixing the duplicate Campaigns bug (#2162)
2. adding support for matching existing Contacts, Accounts, and Household Accounts by filling in Contact1Imported__c, Contact2Imported__c, Account1Imported__c, Account2Imported__c, and HouseholdAccountImported__c.

Interesting testing scenarios include:
- using valid existing Contact and Account Id's, the data import matches and updates those existing records
- what happens with invalid id's?  errors, or do we ignore the match and create new records?  given you are having to store the id's into lookup fields, Salesforce may not even allow this to happen!
- if you match by Id contact1 to an existing contact1, and there is an existing contact2 in the household, do we correctly match to the existing contact2, even when no Id is specified for contact2.
- if you specify a HouseholdAccountId to use, does the household account get correctly updated with its name, greetings, and number of household members?
- if you specify an existing contact, and provide a different existing household account, does the contact get correctly moved to the other household account?  are their related objects fixed up?  basically, this should match whatever happens if you manually move a contact from one household account to another.

njjc:
This pull request looks good, I reviewed the code and did some basic tests. @judisohn is doing some more in depth testing, then I'll merge once that's finished.

PR #2457, opened 2017-01-17
Adding phone support to NPSP Data Importer

### LINKED ISSUE #1926: NPSP Data Import Tool - Add Phone Number Field(s) to Matching Rules
The Contact/Account level phone number fields can't be included in the NPSP Data Imports Matching Rules - they aren't available in the Custom Unique ID field.  Can these be included for Contacts with no email addresses/access?

davidhabib:
# Critical Changes

# Changes

* The NPSP Data Importer now supports matching to existing Contacts using Phone fields.  A new matching option is supported on the "Start Data Import" screen:  First Name, Last Name, and Phone.  Similar to the NPSP Data Importer's support for Email matching, it will use any of the phone fields specified in the Data Import record, and try to match against any of the Standard and NPSP phone fields on Contact. Note that the phone number formatting must be an exact match. For example, 555-121-1212 will not match (555) 121-1212 or 555.121.1212.

# Issues Closed
Fixed #1926


davidhabib:
@njjc this pull request is ready for review.  It includes the work from the first data importer pull request, which you should review and merge first.  then this request will only diff for the additional phone support.

davidhabib:
@jessingrass @judisohn This is the second pull request for NPSP Data Importer, and should be tested after the previous one has been completed.  This pull request adds support for matching existing contacts by Phone number, much like we do with Emails.

Interesting scenarios to test:
- try out the various 4 new matching options
- try out matching to existing contacts who have the matching phone in all the different Standard and NPSP phone fields (5 of them)
- try out matching to existing contacts who have matching phone, using all the different phone fields on the data import record (4 of them)
- does it work for both contact1 and contact2

davidhabib:
@njjc I've just committed to this pull request 1 change and 1 bug fix based on Judi's thorough testing.  So at this point it is ready for review and merge.  thanks.  Just a reminder that you want to merge pull request #2446 first.

judisohn:
Let me know if you'd like for me to take a look at this again, but sounds like you got it covered. Thanks @davidhabib

PR #2458, opened 2017-01-18
Add back empty test class to prevent permanent deletion from package

jlantz:
The UTIL_CascadeDeleteLookups_TEST class was removed from the dev branch.  When I went to create a production release, I noticed that the class would be permanently deleted from the package meaning we couldn't create it again in the future.  This PR adds back an empty class that will prevent that deletion.


jlantz:
Already handled in a separate PR.

PR #2462, opened 2017-01-18
Google Address Validator hits exception on some addresses.  Issue #2461

### LINKED ISSUE #2461: Google Address Validator hits exception on some addresses
If you have google geocoding chosen for your address verification provider, the following address will generate a null pointer exception.

image

davidhabib:
# Critical Changes

# Changes

# Issues Closed
FIxed #2461


davidhabib:
@njjc this fix is ready for review.  turns out if you were using google address verification, and you had Reject Ambiguous Addresses checked on, we would hit a null ptr exception because where were expecting a boolean partial_match in the response, but google actually returns true or null (rather than false).  so treating the null as a boolean caused a null ptr exception.  kinda feels like a platform issue, but anyways, changing the logic to explicitly test the boolean works.

davidhabib:
@njjc I added a trivial fix to opp naming due to an error someone reported in the hub.

njjc:
@jlantz this is an error I haven't seen before, thought you might like to investigate.

ckoppelman:
@davidhabib Any idea when this will be pushed to production?  I'd like to know when these error messages will end :)

judisohn:
@ckoppelman This will push to production as part of NPSP 3.90 tonight. Thanks for your patience.

PR #2465, opened 2017-01-19
Workarounds for locker service issues 

### LINKED ISSUE #2289: Winter '17, Manage Household page hits error copying address to added contact
in Winter '17, if you add an existing contact to the household with the autocomplete control, you will see the error "unable to read sobject" get displayed in the header, and the household's street will copy to the added contact, but not city, state, or zip.

davidhabib:
This pull request introduces our workaround for the LockerService issue where backing objects would not get updated.  It also fixed a few other unreported issues when LockerService is on.

# Critical Changes

# Changes

# Issues Closed
Fixed  #2289

### DIFF
--- a/src/aura/HH_Container/HH_ContainerHelper.js
+++ b/src/aura/HH_Container/HH_ContainerHelper.js
@@ -9,6 +9,15 @@
     },
 
     /*******************************************************************************************************
+     * @description This is a workaround due to a LockerService bug where any added properties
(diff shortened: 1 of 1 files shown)


davidhabib:
@jjbennett this pull request is ready.

davidhabib:
I also manually tested this branch against a GS0 org, so it seems to resolve the issue under Spring '17 as well.

PR #2466, opened 2017-01-23
Cascade Delete

### LINKED ISSUE #1812: When deleting list of relationships, deleting of reciprocal relationships fail if list contains mutually referential relationships
When deleting relationships, it is expected that the reciprocal relationship will also be deleted.

However, when deleting a list of relationships where at least two of the relationships in the list are reciprocals of each other, none of the remaining reciprocal relationships will be deleted.

This seems to be caused by an issue in REL_Relationship_TDTM where trying to delete records that are already deleted in an after delete trigger causes an ENTITY_IS_DELETED error for each of those mutually reciprocal rows when processing the delete DML in TDTM_TriggerHandler.

Since this trigger is run with rollback, the transaction is rolled back and none of the reciprocal relationships are deleted (except for those that were deleted as part of the original delete dml).

There is a unit test that demonstrates this issue here: https://github.com/SalesforceFoundation/Cumulus/blob/test-for-delete-reciprocal-relationship-bug/src/classes/REL_Relationships_TEST.cls#L864

### LINKED ISSUE #2163: NPSP Deletes Non-Empty Accounts
The cleanup code at https://github.com/SalesforceFoundation/Cumulus/blob/rel/3.76/src/classes/ACCT_IndividualAccounts_TDTM.cls#L628 deletes HH accounts if they have no more contacts or opportunities.

However, an account is not necessarily empty at that point, it may have other custom related objects still in it.

I am running into this issue during an overnight integration that sometimes shuffles contacts between households (in a weird but meaningful way). Sometimes, that can cross the batch boundary, and leave a household "empty", even though it still has lots of related information. This behavior should be able to be turned off.

### LINKED ISSUE #2377: Recurring Donation is not deleted if Account/Contact are deleted

### LINKED ISSUE #2443: Delete reciprocal Relationship records when Contact is deleted
thread: https://powerofus.force.com/0D580000039cpMZ

Issue replicated in test org.

### LINKED ISSUE #2449: When Contacts are merged with Recurring Donations, the losing Contact's Recurring Donation(s) are deleted
Thread: https://powerofus.force.com/0D58000003AUWPg

Opportunities are retained, it's only the Recurring Donation that's deleted. It's recoverable from recycle bin.

njjc:
# Critical Changes
- We have re-architected deletion of records in the NPSP to be more consistent with Salesforce behavior. Further documentation will be available soon, and posted in the HUB.

# Changes
## Deletion Behavior Changes
- When a records are deleted, we have improved the handling of Relationships, Allocations, and Recurring Donations so that they are not left orphaned. 
- NPSP no longer automatically deletes orphaned Household or 1-1 Accounts in cases of Contact deletion or Contact Merge.
- Contact Delete button override now prompts the user to confirm delete of the Household/1-1 Account if the Contact is last in the Account.

# Issues Closed
- Fixes #1812
- Fixes #2163 
- Fixes #2377 
- Fixes #2443 
- Fixes #2449


njjc:
@jessingrass here's the branch for testing. Thanks!

src/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -32,60 +32,143 @@
...
-    * contact in the account.
-    * @return pageReference Redirects to the Contacts tab.
-    */ 
-    public pageReference processDelete() {
-        Contact queryContact = [SELECT Id, AccountId, Account.npe01__SYSTEMIsIndividual__c FROM Contact WHERE Id = :conId];
-        string accId = queryContact.AccountId;
-        boolean shouldDeleteContactAlone = true;
+        queryContact = [SELECT Id, Name, AccountId, Account.Name, Account.npe01__SYSTEMIsIndividual__c FROM Contact WHERE Id = :stdController.getId()];
+
+        helpDeleteAccountInstead = String.format(Label.ConfirmDeleteAccount, new List<String>{queryContact.Name, queryContact.Account.Name});
```
davidhabib: won't queryContact.Account.Name cause an exception if Account is null, in the scenario of Contacts marked Private (where we don't create an account for them).

src/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -32,60 +32,143 @@
...
+
+        shouldDeleteContactAlone = true;
 
         if (queryContact.AccountId != null && queryContact.Account.npe01__SYSTEMIsIndividual__c) {
-            list<AggregateResult> contactsInHousehold = [SELECT COUNT(id) ct FROM Contact WHERE AccountId = :accId GROUP BY AccountId HAVING COUNT(id) = 1];
+            accountId = queryContact.AccountId;
```
davidhabib: same issue previously mentioned.  with private contacts, I don't think we can safely dereference  contact.account.foo

davidhabib: oops!  nevermind this one, since you are testing for AccountId != null first!

src/classes/CON_DeleteContactOverride_CTRL.cls
davidhabib: just manually tested with a private Contact (ie, no account), and did not run into any error.  weird.

davidhabib:
I found the following issues while adhoc testing this pull request:

	1. Have a contact with a hh account and a Closed/Won opp associated with both the contact and hh account.  Using the Contact Delete Override page, choose to delete the Account.  Page refreshes with no error, and seems to have done nothing.

	2. Have a private contact (no account), who has a Closed/Won opp associated with themselves and some other account.  Using Contact Delete Override page.  Delete contact.  Expected error about having closed/won opps.  No error shown and the contact gets deleted.

	3. Have a private contact (no account), who has an open Case.  Contact Delete Override is in place.  Delete the contact.  The Contact Delete Override page displays with nothing but the "return to Contact" button.  No error, no text.

        4. Same as #3, but contact is not private.  Contact Delete Override page displays and gives message about needing to delete the account instead.  Click the Delete Account button.  Page redisplays with no error or change.

davidhabib:
OK, just retested and all 4 of my issues were because I didn't have the updated trigger handlers in my trigger handler table.  those scenarios now work fine (other than the URL click issue, and contact.name being null).  thanks.

njjc:
@davidhabib thanks, I've updated the pull request with fixes for those two remaining issues.

PR #2474, opened 2017-01-31
Feature/jen cal fs

jjbennett:
Fixing issues in the component discovered during v4s implementation code review.

# Critical Changes

# Changes

# Issues Closed


jjbennett:
@davidhabib I've made the changes to npsp component from the comments in the v4s which is also updated with these changes.

PR #2483, opened 2017-02-06
2 fixes for Batch Data Import

### LINKED ISSUE #2479: NPSP Data Importer (3.89) creates new Account for Contact2 even if Contact1 is matched on ID
STR:

1. New NPSP Data Import record - Contact1 Imported (select existing Contact), Contact2 First Name/Contact2 Last Name (something that won't match)
2. Run utility with default matching

Expected: New Created Contact2 in same existing Household as Contact1

Result: Contact1 is Matched correctly. Contact2 is Created in its own Household.

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixed #2479

### DIFF
--- a/src/classes/BDI_DataImport_BATCH.cls
+++ b/src/classes/BDI_DataImport_BATCH.cls
@@ -461,9 +461,9 @@ public class BDI_DataImport_BATCH implements Database.Batchable<sObject> {
             setC1Id.add(di.Contact1Imported__c);
         }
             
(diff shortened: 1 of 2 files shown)


davidhabib:
@judisohn @jjbennett Here is the fix for the BDI issue Judi found, as well as one other I found (if the DI record has c1's Id, but no c1 names provided, and c2 only had a firstname, we weren't copying c1's lastname to c2 based on the saved contact.

Issue #2375, opened 2016-11-18
Deleting an Affiliation that is checked as Primary should remove the Primary Affiliation from the Contact record


judisohn:
Thread here: https://powerofus.force.com/0D580000034bdBK

judisohn:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000004DBOtAAO" target="_blank">W-016962</a>

PR #2490, opened 2017-02-13
Affiliation Delete and Undelete Handling

njjc:
# Critical Changes

# Changes
- Deleting Affiliations marked Primary now blanks out the contact's Primary Affiliation lookup.
- Undeleting Affiliations marked Primary now populates the contact's Primary Affiliation lookup. If that lookup already has a value, Primary is unchecked on the Affiliation being undeleted.

# Issues Closed


njjc:
@smarovich could you review the Changes text in this pull request? Let me know if you have any questions. Thanks!

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -464,26 +477,45 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
...
         setAllContactIds.addAll(mapConIdAccIdUnprimary.keySet());
+        setAllContactIds.addAll(mapConIdAfflUndelete.keySet());
 
         if (!setAllContactIds.isEmpty()) {
             list<Contact> conForUpdate = new list<Contact>();
             list<Contact> queryCon = [SELECT Id, Primary_Affiliation__c FROM Contact WHERE Id IN :setAllContactIds];
             
             for (Contact con : queryCon) {
-                if (mapConIdAccIdPrimary.containsKey(con.id)) {
+                if (mapConIdAccIdPrimary.containsKey(con.id) && con.Primary_Affiliation__c != mapConIdAccIdPrimary.get(con.id)) {
```
davidhabib: were we previously updating contacts even though we didn't have to?  what was the scenario?

njjc: Yep, we were updating it regardless of whether it was needed it so I added the check to avoid changing last modified date for nothing.

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -495,17 +527,16 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
 
         if (!mapConIdAccIdPrimary.isEmpty()) {
             //uncheck any existing primary affiliations
-            list<npe5__Affiliation__c> afflForUpdate = new list<npe5__Affiliation__c>();
             list<npe5__Affiliation__c> queryAffl = [SELECT Id, npe5__Contact__c, npe5__Organization__c FROM npe5__Affiliation__c WHERE npe5__Primary__c = TRUE AND npe5__Contact__c IN :mapConIdAccIdPrimary.keySet() AND ID NOT IN :mapConIdAfflIdPrimary.values()];
         
             for (npe5__Affiliation__c affl : queryAffl) {
```
davidhabib: as I've been reading more about LDV orgs and queries, I suspect that we should redo this query to remove the NOT IN check, which can't use an index, and instead just ignore the primary affiliations in our subsequent for loop.  In a LDV org with a million contacts, and 1 million+ affiliations, I think this query would time out or get an error about not being selective.  If you want to try the change, add it to this pull request.  If you'd prefer to move on, could you just add a backlog item for this change?  thanks.

njjc: Good catch, I'll remove the query filter and filter out those records manually.

judisohn:
I believe this fixes #2375 (making note to close issue).

PR #2491, opened 2017-02-14
Feature/rel field sync

tet3:
**Synchronization of custom fields across reciprocal relationship pairs**
Handled in REL_Relationships_TDTM. The synchronization of the packaged Description and Status fields is done with the same mechanism as for custom fields (made for much easier test coverage, among other things.)  The list of fields to sync is built from an AllFieldsDescribe, from which are removed: 

* Standard fields
* Packaged relationship fields which should not be synced (stored in a `final` at the top of the class) 
* fields that can't be written (formula, autonumber)
* if the custom field sync feature is turned on by checkbox custom setting, any non-npsp custom fields excluded via custom list setting
* if the custom field sync feature is off, all non-npsp custom fields are removed from the sync list

An upgrading organization that does not enable the custom field sync would still see the Description and Status fields synced (via the new field sync code). 

The list of custom fields to exclude from sync is in a new List Custom Setting, with just the API name of the field stored in the Name field of the record. The alternative approaches I considered and discussed some on the Hub, a hierarchical custom setting textarea field for either inclusion or exclusion, were likely to prove inadequate with very many fields. 

That said, I preferred a multi-select picklist for the settings UI rather than the one-at-a-time creation of records as is done for most other List settings (which have more than one meaningful field). So I based it off of the STG_DataBoundMultiSelect.component, and made adjustments in the panel controller to work with a list setting rather than a Textarea field.  ~~I'm creating and deleting setting records in the setter for the panel's List&lt;String&gt; property that's bound to the multi-select. It feels a bit odd, but works. But I'm definitely especially open to feedback on that point.~~

# Critical Changes

# Changes
## Synchronization of custom Relationship fields across reciprocal pairs
- New checkbox in NPSP Settings > Relationships, "Enable Custom Field Sync." When checked, writable custom fields on the Relationship object will be synchronized across reciprocal Relationship pairs.
- To **exclude** fields from this sync, select them in the "Relationship Sync Excluded Fields" list.  

# Issues Closed


judisohn:
Thanks @tet3 - this likely won't be able to be reviewed in the sprint that starts tomorrow, but I'll work with @mpusto and @davidhabib to see if we can get it in to the next one. Really appreciate the time you put in on this and can't wait to see how it works!

janeisaac:
Thanks @tet3 and @judisohn  This would be a very useful feature to have.

judisohn:
Hey @tet3 We just had our Sprint Planning meeting - unfortunately, this didn't make it in. The entire engineering team will be at an offsite next week, so work days are super short on this one. We hope to have it in at some point after they return, but we're likely looking into April at this point. Thanks for your patience and looking forward to seeing you soon!

janeisaac:
Thanks for the update Judi

njjc:
Thanks again @tet3 for the contribution, everything looks great and we're about ready to merge it in. I've moved it over to #2528 and will be merging it after it goes through documentation review.

dylanhansen:
Thanks @tet3 for your contribution!

PR #2504, opened 2017-02-28
FIX: Contacts could not be merged in batch/future.

stenebenau:
# Critical Changes

# Changes

# Issues Closed
2463

### DIFF
--- a/src/classes/CON_ContactMerge_BATCH.cls
+++ b/src/classes/CON_ContactMerge_BATCH.cls
@@ -34,7 +34,7 @@
 * @description Batch class to merge multiple contacts.
 */
 
-global class CON_ContactMerge_BATCH implements Database.Batchable<SObject>{
+public class CON_ContactMerge_BATCH implements Database.Batchable<SObject>{
 	
 	/*******************************************************************************************************
     * @description masterContact holds the master contact where all losers are merged into. 
@@ -51,7 +51,7 @@ global class CON_ContactMerge_BATCH implements Database.Batchable<SObject>{
     * @param master masterContact 
     * @param losers List<Id> containing all loserids.
     */
-	global CON_ContactMerge_BATCH(Contact master, List<Id> losers) {
+	public CON_ContactMerge_BATCH(Contact master, List<Id> losers) {
 		this.masterContact = master;
 		this.loserIds = losers;
 	}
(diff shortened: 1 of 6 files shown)


src/classes/CON_ContactMerge_BATCH.cls
```diff
@@ -0,0 +1,99 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Sten Ebenau
+* @date 28 February 2017
+* @group ContactMerge
+* @description Batch class to merge multiple contacts.
+*/
+
+global class CON_ContactMerge_BATCH implements Database.Batchable<SObject>{
```
davidhabib: since this batch class is really only for testing purposes, let's make it public, not global.  I assume you were not planning on using it in your solution, correct?

stenebenau:
Ah. Didn't know that was even possible. Added the public modifiers. Wasn't planning on using it in my app.

src/classes/CON_ContactMerge_BATCH.cls
stenebenau: Didn't know you could use public these days. Just added it.

judisohn:
@stenebenau Hi Sten - we have quite a bit happening right now so it may be a while until @davidhabib or another dev on our team will be able to consider this for a merge. In the meantime, would you mind dropping me a note at jsohn@salesforce.com so I have your email and can send you a CLA? Thanks!

davidhabib:
@stenebenau I'm afraid this isn't going to work.  I copied your branch so I could do more thorough testing.  Unfortunately, your test code explicitly sets a single trigger handler, ACCT_AccountMerge_TDTM, to async.  But by inserting this one trigger handler, when your test runs, none of the other trigger handlers are invoked, because with SeeAllData=false, they don't exist, and our trigger handler code only adds them for tests if no other trigger handlers exist.  The correct way to do your test setup would be:

        List<Trigger_Handler__c> listTH = TDTM_DefaultConfig.getDefaultRecords();
        for (Trigger_Handler__c th : listTH) {
            if (th.Class__c == 'ACCT_AccountMerge_TDTM')
                th.Asynchronous__c = true;
        }
        insert listTH;

But I found that this still does not work.  With our other triggers involved, there are other callers who can't do their future work, since we are inside a batch, but fail to do their update, because the account being updated is still being merged:

System.DmlException: Update failed. First exception on row 0 with id 001j0000019ZHx1AAG; first error: SELF_REFERENCE_FROM_TRIGGER, Object 001j0000019ZHx1AAG is currently in a merge operation, therefore a trigger can not update it.: []

I'm guessing you ran into this originally, and that's why you changed the ACCT_AccountMerge_TDTM trigger handler to be async.

I do not see any work around that will provide a solution.  We would need to come up with a whole new way of doing some of our work asynchronously, without using futures, to support your desire for Contact Merge working from batch.

@judisohn @mpusto I'm closing this, since I don't see how we can support this.

PR #2512, opened 2017-03-02
Engagement Plans fixes

### LINKED ISSUE #2495: If Task on record is deleted but EP Task remains, updating other task status fails with error
support case: [15733215](https://org62.lightning.force.com/one/one.app#/sObject/5000M00000gVgx4QAC/view) 

Issue replicated. Steps to repeat:

1. Create EP Template with 3 tasks. 2 dependent on 1, 3 dependent on 2 (settings on template do not matter, as long as the tasks have dependencies)
2. Create EP to a Contact (object doesn't matter)
3. On the Contact's record, delete the final task (which should be 'Waiting on Dependent Task')
4. Set status on first 'Not Started' task to 'Completed'
5. Error:

image

It may not be an uncommon use case to edit individual task records, apart from the template's more generic tasks.

### LINKED ISSUE #2510: EP: Error if non-Admin user completing task doesn't have ownership on child task
[Hub thread](https://powerofus.force.com/0D58000003GDy1i).

This happens if non-Admins are assigned EP Tasks, but then on completion the dependent tasks can't be updated to move out of Waiting status. User has to have 'Modify All' on the parent object (or be above in role hierarchy) to update tasks they are not assigned.

njjc:
# Critical Changes

# Changes

# Issues Closed
- Fixed #2495 
- Fixed #2510

### DIFF
--- a/src/classes/EP_TaskDependency_TDTM.cls
+++ b/src/classes/EP_TaskDependency_TDTM.cls
@@ -34,6 +34,9 @@
 * @group-content ../../ApexDocContent/EngagementPlans.htm
 * @description When a Task created by an Engagement Plan is given a Closed status, checks for dependent Tasks
 * and updates them by changing the status, updating the due date, and sending email.
+*
+* This class is not marked with sharing as it needs access to Tasks dependent on those in the trigger 
+* set that are potentially owned by other users.
 */
 public class EP_TaskDependency_TDTM extends TDTM_Runnable {
 
--- a/src/classes/EP_TaskDependency_TDTM.cls
+++ b/src/classes/EP_TaskDependency_TDTM.cls
@@ -82,7 +82,10 @@ public class EP_TaskDependency_TDTM extends TDTM_Runnable {
     /** @description A set of EP Task IDs in the above dependent tasks, to avoir OR in the query.*/
     private Set<Id> allDependentEPTasksForQuery {
         get {
-            return new Set<Id>().addAll(directDependentEPTasks.keySet()).addAll(indirectDependentEPTasks.keySet());
+            Set<Id> results = new Set<Id>();
+            results.addAll(directDependentEPTasks.keySet());
+            results.addAll(indirectDependentEPTasks.keySet());
+            return results;
         }
     }
 
--- a/src/classes/EP_TaskDependency_TDTM.cls
+++ b/src/classes/EP_TaskDependency_TDTM.cls
@@ -82,14 +82,7 @@ public class EP_TaskDependency_TDTM extends TDTM_Runnable {
     /** @description A set of EP Task IDs in the above dependent tasks, to avoir OR in the query.*/
     private Set<Id> allDependentEPTasksForQuery {
         get {
-            Set<Id> dependentEPTasksForQuery = new Set<Id>();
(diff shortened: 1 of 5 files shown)


src/classes/EP_TaskDependency_TDTM.cls
```diff
@@ -35,7 +35,7 @@
 * @description When a Task created by an Engagement Plan is given a Closed status, checks for dependent Tasks
 * and updates them by changing the status, updating the due date, and sending email.
 */
-public with sharing class EP_TaskDependency_TDTM extends TDTM_Runnable {
+public class EP_TaskDependency_TDTM extends TDTM_Runnable {
```
davidhabib: @njjc please put a comment here saying why we don't have with sharing.  thanks.

src/classes/EP_Task_UTIL.cls
```diff
@@ -35,7 +35,7 @@
 * @description EP Task and Task utilities for Engagement Plans. Constructor creates maps of EP Tasks and EP Templates.
 * Methods to create and update tasks, calculate 
 */
-public with sharing class EP_Task_UTIL {
+public class EP_Task_UTIL {
```
davidhabib: please comment the reason for no explicit with sharing.  thanks!

src/classes/EP_TaskDependency_TEST.cls
```diff
@@ -295,16 +295,81 @@ private with sharing class EP_TaskDependency_TEST {
...
+        System.assertEquals(2, results.size(), 'Two tasks should be created.');
+
+        //delete the dependent task
+        delete [SELECT Id FROM Task WHERE Engagement_Plan_Task__c = :dependentEpTask.id];
+
+        Task parentTask = [SELECT Id FROM Task WHERE Engagement_Plan_Task__c = :epTask.id];
+        parentTask.Status = UTIL_UnitTestData_TEST.getClosedTaskStatus();
+
+        try {
+            update parentTask;
```
jjbennett: @njjc should the Test.startTest be here vs 355?

src/classes/EP_TaskDependency_TDTM.cls
```diff
@@ -76,9 +79,26 @@ public with sharing class EP_TaskDependency_TDTM extends TDTM_Runnable {
     private Map<Id, Engagement_Plan_Task__c> directDependentEPTasks = new Map<Id, Engagement_Plan_Task__c>();
     /** @description Map of EP Tasks that are indirect dependents of Tasks in the trigger set, i.e. grandchildren. */
     private Map<Id, Engagement_Plan_Task__c> indirectDependentEPTasks = new Map<Id, Engagement_Plan_Task__c>();
+    /** @description A set of EP Task IDs in the above dependent tasks, to avoir OR in the query.*/
+    private Set<Id> allDependentEPTasksForQuery {
+        get {
+            Set<Id> dependentEPTasksForQuery = new Set<Id>();
+            for (Engagement_Plan_Task__c epTask : directDependentEPTasks.values()) {
```
jjbennett: Instead of the two loops couldn't we add the keysets to the set?

njjc: Good point, I forgot that the keys were the IDs of the values!

PR #2528, opened 2017-03-20
Relationships Field Sync

njjc:
# Critical Changes

# Changes
## Sync Custom Relationship Fields
- This feature is controlled by a new NPSP Setting, Enable Custom Field Sync. When checked, custom fields on the Relationship will be sychronized between the two contacts in a relationship pair. 
- You can exclude fields from synching by adding them to the Fields to Exclude from Sync list. This list appears when you select the Enable Custom Field Sync checkbox.

# Issues Closed


tet3:
The second point in the *Changes* section above isn't accurate - the Excluded Fields multi-select in `STG_PanelRel.page`  is not dynamically en/disabled by the Enable checkbox. I thought about that, but a) I'm not good enough with JS, and b) that didn't seem to be the standard behavior in NPSP settings - there are several places where settings that only apply in the case of another checkbox are exposed regardless of the state of that checkbox.

Also, @smarovich , I was talking to @mpusto here at the Sprint - I'm happy to help with any doc needs. It seems like understanding the changes in NPSP Settings will be helped by illustrating what happens in a reciprocal pair with the Setting on.

njjc:
Hey @tet3, I made the update myself to hide that section if the box is unchecked, felt less confusing when looking at the UI for the first time: https://github.com/SalesforceFoundation/Cumulus/pull/2528/commits/8aac949c46f2dd4ab4d1746e1ca939630755b5a5

PR #2533, opened 2017-03-21
Feature/bdi unique id fix

### LINKED ISSUE #2517: NPSP Data Import: if Contact Unique ID is number, fails with duplicate value error
[Thread.](https://powerofus.force.com/0D58000003GHMk0) [Case.](https://org62.my.salesforce.com/5000M00000gX1ag)

Steps to repeat:

1. Create custom field on Contact, type Number (18,0). Mark as External Unique ID
2. Create corresponding NPSP Data Import fields for Contact1/Contact2 Unique ID, type Number (18,0), mapped to above field
3. Create NPSP Data Import record with First Name, Last Name, Contact1 Unique ID
4. Create Contact with same Unique ID (so it will have something to expect to match against)
4. Process with Contact Unique ID field selected

Error: duplicate value found: unknown duplicates value on record with id: 003...

5. Now change Contact and NPSP Data Import fields from Number to Text data types.
6. Process Data Import record with no error.

davidhabib:
In Batch Data Import, the user can have us match by an external unique ID for contacts and accounts.  The datatype of this field can be either a string or a number.  Unfortunately, we had a bug where if it was a number, we were not consistently handling how we coerce them to strings for BDI key mapping.  So when we read the numeric field from an existing contact that had 1 stored in its numeric Unique ID, we'd get back 1 from the soql, and '1' is what was stored in the key.  But when we were reading from our DI record, even though the field was also numeric, it was coming back as 1.0, and '1.0' was stored in the key. So now we force always storing numbers as doubles, which causes them to consistently include the decimal and zero.  Exact same issue for Accounts.

Unfortunately, I cannot write tests to show this problem and prove the fix.  The current tests dealing with our external unique id matching has to use existing fields on contact and the DI record (and similarly Account and the Di record), since we can't create temporary fields in a test.  The DI record does not contain any numeric fields, only text ones. 

# Critical Changes

# Changes

# Issues Closed
fixed #2517

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -12,6 +12,9 @@ project:
         default_branch: dev
         prefix_beta: uat/
         prefix_release: rel/
+    apexdoc:
+        homepage: ApexDocContent/homepage.htm
+        banner: ApexDocContent/projectheader.htm
     dependencies:
         - namespace: npo02
           version: 3.7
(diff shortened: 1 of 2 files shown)


davidhabib:
note that I tested with an ExternalID of number(18,0) and number(4,4) to make sure the conversion to double worked for both.

PR #2536, opened 2017-03-23
Clear level when target object no longer meets any level criteria

### LINKED ISSUE #2503: When Levels are re-defined, Levels not cleared from objects
When a user re-defines a level to be a different start and end value, the Level assignment batch goes through and assigns all records to the new levels. However, if a record has an old level field populated and it now meets *no* level definition, it retains the old information. This can cause confusion by mis-labeling records. The level field should be cleared for records which no longer meet any level criteria.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2503

### DIFF
--- a/src/classes/LVL_LevelAssign_BATCH.cls
+++ b/src/classes/LVL_LevelAssign_BATCH.cls
@@ -80,11 +80,20 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
             // for each of our Ladders
             for (String strLadder : mapLadderToLevels.keySet()) {
                 list<Level__c> listLvls = mapLadderToLevels.get(strLadder);
-                
+                Id lvlIdPrev;
+                // track whether or not a match was found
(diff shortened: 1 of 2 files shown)


src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -136,7 +138,12 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
                         break;
                     }
 
-                } // levels               
+                } // levels
+                // After looping through all levels in the current ladder if no match found but level set clear level
+                if (!matchedLvl && lvlIdPrev != null) {
+                    sobj.put(strLadder, null);
+                    mapSObjUpdate.put(sobj.Id, sobj);
+                }
```
davidhabib: all looks good, except when we clear out the current level, we should store that level in the previous level on the object.

PR #2544, opened 2017-03-28
moved all BDI opportunity/payment code into a separate class

davidhabib:
@njjc this pull request just moves all of the opportunity and payment code in BDI to a separate class, in preparation for additional work.  no logic has changed.  all tests pass!  just had to make a few private utility functions in BDI_DataImport_BATCH public, so the new class could use them as well.

# Critical Changes

# Changes

# Issues Closed


davidhabib:
hi @njjc .  it would be great if you could review and merge this tomorrow, so that I'm able to start committing new work to gitHub without having it go into this request.  I guess I could create a new branch, but I'm concerned I may then mess up work I have in progress.  thanks.

PR #2551, opened 2017-04-05
Only updating installments on open opportunities

### LINKED ISSUE #2523: If recurring donation is None, and an opp is Closed Lost, any edit to other opps adds installments
[Thread](https://powerofus.force.com/0D58000003HSNmj)

Steps to repeat:

1. None Recurring Donation (associate to Contact, but probably doesn't matter)
2. Allow to generate opps
3. Set first opp to Closed Lost
4. Refresh opps. New opp created, which is expected.
5. Edit anything on new Pledged opp (such as description - doesn't matter the field)
6. Refresh opps. New opp created and installment increases. Happens now on every edit.

If there is no Closed Lost opp, this behavior (opp creation on opp edit) does not occur.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2523


jjbennett:
@davidhabib this should be good to go now. The dupe rules were being a pain but are happy now.

PR #2555, opened 2017-04-07
Updated query for batch default allocation creation.

njjc:
# Critical Changes

# Changes

# Issues Closed


src/classes/ALLO_MakeDefaultAllocations_BATCH.cls
```diff
@@ -39,43 +39,40 @@ public class ALLO_MakeDefaultAllocations_BATCH implements Database.Batchable<sOb
...
     private Allocations_Settings__c alloSettings = UTIL_CustomSettingsFacade.getAllocationsSettings();
+    
+    /** @description Which opportunity types to exclude. We need to split this into a set 
+    * to avoid a partial match, e.g. Donor could be matched to the string Matching Donor*/
+    private Set<String> oppTypesToExclude = String.isBlank(alloSettings.Excluded_Opp_Types__c) ? new Set<String>() : new Set<String>(alloSettings.Excluded_Opp_Types__c.split(';'));
+
     /** @description The query for the batch process to run on.*/
     private String query;
     
     /** @description The batch process constructor; creates the query based on type and record type exclusion settings.*/
```
jjbennett: @njjc should we update the constructor comment explaining why we are using this approach and no longer filtering via the query?

njjc: Sounds good, will do.

PR #2564, opened 2017-04-27
core logic for BDI support for Donation matching

davidhabib:
@njjc I've decided the BDI support for Opp/Pmt matching has reached a stable point and is ready for review.  Since I still need to do LDV testing, and UI work for the new options, I think we should start reviewing and merging the work into this new bdi-donations-REVIEWED branch, not dev.  I will also have Judi start some testing once I write up a beta document on the behavior.

# Critical Changes

# Changes
* The NPSP Data Importer now supports options for having Donations matched against existing Opportunities and Payments, rather than always creating new ones.  Please see <doc link here> for full details on this new feature set.

# Issues Closed


judisohn:
@davidhabib Can't wait to play with it. :-)

davidhabib:
@judisohn @jessingrass OK, I believe the BDI Donation matching is ready to be tested.  There is no UI yet, so you will have to set your options manually in custom settings.  Details in the quip doc I posted.  Note that if you run this from unmanaged code, when you click on the Start Data Import button, you will have to do the hack to fixup the URL by replace 'npsp.' with 'c.'

davidhabib:
@njjc let's schedule an hour for me to walk you through this code.  is Monday OK?

njjc:
@davidhabib sounds good, sent you an invite.

davidhabib:
@njjc just finished checking in more tests.  coverage is now up to 95% for BDI_Donations, and 92% for BDI_DataImport_BATCH.

PR #2580, opened 2017-05-15
2521: Manage Allocations bug with commas and other symbols in the numeric field

### LINKED ISSUE #2521: Manage Allocations - Summary row malfunctions for numbers with commas
Allocation amounts with commas in them, e.g. $2,500.00, are misinterpreted by Javascript as NaN, causing summary rows to not include those amounts.

[From the Hub.](https://powerofus.force.com/0D58000003H0RIl)

force2b:
Update the JS that calculates the amount values to strip out leading currency symbols and all commas from the string value

# Critical Changes

# Changes

# Issues Closed
Fixed #2521

### DIFF
--- a/src/pages/ALLO_ManageAllocations.page
+++ b/src/pages/ALLO_ManageAllocations.page
@@ -10,6 +10,8 @@
         var sumAmount;
         
         var parentAmount = {!parentAmount};
+        var decimalSeparator = "{!decimalSeparator}";
+        var thousandsSeparator = "{!thousandsSeparator}";
         var defaultEnabled = {!Settings.Default_Allocations_Enabled__c};
 
         function isLightningExperienceOrSalesforce1() {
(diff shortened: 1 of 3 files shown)


src/pages/ALLO_ManageAllocations.page
```diff
@@ -45,9 +45,12 @@
...
                     }
                 }
-                if (!isNaN($(o).val()))
-                    sumAmount += Math.round($(o).val()*100);
-               
+                var amt=$(o).val();
+                // due to how chrome handle's the global replace, have to
+                // first remove the currency symbols first and then strip out all commas.
+                if (!isNaN(amt.replace(/[\$£€]/g, '').replace(/,/g, '')))
+                    sumAmount += Math.round(amt.replace(/[\$£€]/g, '').replace(/,/g, '')*100);
```
jjbennett: The only concern I have with filtering out the comma is that in some locales the comma is used in place of the decimal. Can you try to see if there is another solution that doesn't remove the comma completely?

force2b: Good point. Completely forgot that a comma is used for that in some countries. I think there's a way I can determine the appropriate thousands delimiter (probably in Apex). If so, we can use that the JS to strip out the appropriate character to avoid removing a comma when it's actually meant to be a decimal delimiter.

Issue #2525, opened 2017-03-16
Provide transactional and user account way to disable all NPSP Triggers


davidhabib:
hub thread: https://powerofus.force.com/0D58000003HU7DH

davidhabib:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000005QCnvAAG" target="_blank">W-018116</a>

sefg:
Hi Nick, attached is the code and and some notes - I'd be happy for you to base a solution on this if it is useful. 
cheers
Stuart
[ProcessControl.zip](https://github.com/SalesforceFoundation/Cumulus/files/867893/ProcessControl.zip)

davidhabib:
this issue was addressed and should be closed.

mrbelvedere:
Included in production release rel/3.100

PR #2584, opened 2017-05-22
[2525] TDTM Transactional and persistent control to disable individual triggers

force2b:
# Critical Changes

# Changes

# Issues Closed

Addresses request in #2525 to allow specific NPSP Triggers to be disable through either Apex Code  for the current user context or persistently by adding a filter to the TDTM TriggerHandler record.

### DIFF
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -50,8 +50,9 @@ public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGa
             List<Trigger_Handler__c> result = [select Id from Trigger_Handler__c limit 1];
             rowCount = result.size();
         }
-        if(rowCount == 0)
-            return true;    
+        if(rowCount == 0) {
+            return true;
+        }
         return false;
     }
 
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -91,7 +91,7 @@ public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGa
     public static list<Trigger_Handler__c> listTH {
         get {
             if (listTH == null) {
-                listTH =  [select Asynchronous__c, Class__c, Load_Order__c, Object__c, Active__c, Usernames_to_Exclude__c
+                listTH =  [select Asynchronous__c, Class__c, Load_Order__c, Object__c, Active__c, Usernames_to_Exclude__c, 
                     Trigger_Action__c from Trigger_Handler__c order by Load_Order__c];
             }
             return listTH;
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -71,7 +71,7 @@ public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGa
             Set<String> excludedUserNames = (th.Exclude_For_These_Usernames__c != null ? 
                     new Set<String>(th.Exclude_For_These_Usernames__c.split(';')) : new Set<String>());
             if (th.Object__c == objectName && th.Trigger_Action__c.contains(strAction) && th.Active__c == true
-                    && (th.Exclude_For_These_Usernames__c == null || !excludedUserNames.contains(currUserName))) {
+                    && !excludedUserNames.contains(currUserName)) {
                 listClasses.add(th);
             }
         }
@@ -92,7 +92,7 @@ public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGa
         get {
             if (listTH == null) {
                 listTH =  [select Asynchronous__c, Class__c, Load_Order__c, Object__c, Active__c, Exclude_For_These_Usernames__c,
-                    Trigger_Action__c from Trigger_Handler__c where Active__c = true order by Load_Order__c];
+                    Trigger_Action__c from Trigger_Handler__c order by Load_Order__c];
             }
             return listTH;
         }
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -36,7 +36,7 @@
  * @description Class that queries custom object to retrieve classes to call for creation or modification
  * of a record of a specific object type.
  */
-public class TDTM_ObjectDataGateway implements TDTM_iTableDataGateway {
+public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGateway {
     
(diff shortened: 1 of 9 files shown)


src/classes/TDTM_ObjectDataGateway.cls
```diff
@@ -54,45 +54,51 @@ public class TDTM_ObjectDataGateway implements TDTM_iTableDataGateway {
...
     /*******************************************************************************************************
      * @description holds the trigger handlers cached in a list, to avoid repetitive soql calls.
      */
-    private static list<Trigger_Handler__c> listTH {
+    public static list<Trigger_Handler__c> listTH {
         get {
             if (listTH == null) {
-                listTH =  [select Asynchronous__c, Class__c, Load_Order__c, Object__c, Active__c, 
+                listTH =  [select Asynchronous__c, Class__c, Load_Order__c, Object__c, Active__c, Exclude_For_These_Usernames__c,
                     Trigger_Action__c from Trigger_Handler__c where Active__c = true order by Load_Order__c];
```
njjc: I think we should remove the soql filter for active__c since we're now removing them in getClassesToCallForObject(), that way the static collection contains all the records and not just the active ones; integrators could enable certain triggers only for certain transactions but leave them disabled by default.

src/objects/Trigger_Handler__c.object
```diff
@@ -89,6 +89,16 @@
         <unique>false</unique>
     </fields>
     <fields>
+        <fullName>Exclude_For_These_Usernames__c</fullName>
```
njjc: @smarovich we need string review for this field name and description along with the example docs that Mike forwarded you. This won't go in until next sprint so no huge rush. Thanks!

src/classes/TDTM_ObjectDataGateway.cls
```diff
@@ -54,45 +54,51 @@ public class TDTM_ObjectDataGateway implements TDTM_iTableDataGateway {
...
-     *              is sorted by load order.
+     *              is sorted by load order. Filters out in-active handlers and any where the list of 
+     *              excluded usernames includes the current context user.
      * @param objectName The name of the object.
      * @param action The action (before insert, after insert,...). 
      * @return List<SObject> List of classes to call, sorted by load order.
      **/
     public static List<SObject> getClassesToCallForObject(String objectName, TDTM_Runnable.Action action) {   
         string strAction = action.Name();
+        string currUserName = UserInfo.getUserName();
```
njjc: I think this variable is superfluous and we can just put the call directly where it's needed.

davidhabib:
@force2b the code looks good.  I do think we should add this new field to our NPSP Settings page, STG_PanelTDTM, that displays triggers and allows users to create new triggers.

src/classes/TDTM_ObjectDataGateway.cls
```diff
@@ -54,45 +54,51 @@ public class TDTM_ObjectDataGateway implements TDTM_iTableDataGateway {
...
         List<SObject> listClasses = new list<Sobject>();
         for (Trigger_Handler__c th : listTH) {
-            if (th.Object__c == objectName && th.Trigger_Action__c.contains(strAction)) 
+            Set<String> excludedUserNames = (th.Exclude_For_These_Usernames__c != null ? 
+                    new Set<String>(th.Exclude_For_These_Usernames__c.split(';')) : new Set<String>());
+            if (th.Object__c == objectName && th.Trigger_Action__c.contains(strAction) && th.Active__c == true
+                    && (th.Exclude_For_These_Usernames__c == null || !excludedUserNames.contains(currUserName))) {
```
njjc: I think we can leave out the null check and the OR since an empty set won't ever contain the current username.

src/objects/Trigger_Handler__c.object
davidhabib: @smarovich I was also going to ask for a string review.  Michael has used the developer name Exclude_For_These_Usernames__c, but the label is Do Not Run for these Users.  Typically, we try to keep the dev name and label the same.  Also, this name seems a bit longer and more phrase like than we have done in the past.  I'm not against it, but it does seem a bit different than most fields in NPSP.

src/classes/TDTM_ObjectDataGateway.cls
force2b: True. I generally think of any calls to system classes as expensive, however this is a very small recordset, so it be negligible.

src/classes/TDTM_ObjectDataGateway.cls
force2b: That will  work as long as `excludedUserNames` is initialized to an empty set rather than null if the `Exclude_For_These_Usernames__c` field is null

src/classes/TDTM_ObjectDataGateway.cls
njjc: Good point, I hope they cache it in a static var, but leaving this as is is fine :)

src/objects/Trigger_Handler__c.object
force2b: Fair point. I can put the two in sync once a final label is determined.

force2b:
@njjc and @davidhabib All changes made, less any required change to the field label/api name/help text/description from @smarovich.

On the Settings panel, I added the new field as the last column and set word-break wrapping manually via a style on the div because SLDS did not have a style class to support this specific kind of wrapping for the column. An alternative would be to go with a much wider column, but that seemed like overkill for the rare use-cases where this new field will have a value.

src/objects/Trigger_Handler__c.object
smarovich: I agree that the Developer Name and the Label should match (so it's easier to identify regardless if you're looking at one or the other). Could we go with something simpler like Developer Name: Users_to_Exclude and Label: Users to Exclude?

force2b: How about "Usernames to Exclude" (`Usernames_to_Exclude__c`)?

smarovich: Sounds good!

force2b: @smarovich Are you ok with the inline help text and description values for this new field?

davidhabib:
@force2b looks like you need to update the code to match the new field names!

force2b:
@davidhabib Actually just forgot to update the package.xml with the changed field name and refresh/commit the layout

src/objects/Trigger_Handler__c.object
force2b: Hi @smarovich. Did you happen to see my note from a couple of days ago on approving the inline help/description for this new field? Here is a direct link to the field definition in the PR: https://github.com/SalesforceFoundation/Cumulus/pull/2584/files#diff-e5e3591142b5df80da074261b1a9114d

smarovich: Looks great, @force2b. I would just make usernames lowercase.

PR #2594, opened 2017-06-07
Display unexpected exception message for Delete Data Import buttons

### LINKED ISSUE #1955: Buttons for Batch Data Import need to trap and display error messages
the shared page has apex:pageMessages, but the button handlers need to try/catch and add to pagemessages the exception.  Also worth considering making new pages for these buttons since lightning doesn't allow us to fix the namespace prefix with c in the address bar in DE orgs.

RadSFDeveloper:
# Critical Changes
- In order to enable allowed users to execute "Delete All Data Import Records" and "Delete Imported Data Import Records" operations, it is required to add selected profiles into "Enabled Profiles" at "Security" page for the new "BDI_DataImportDeleteBTN" Visualforce page.

# Changes

# Issues Closed
Fixes #1955

### DIFF
--- a/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
+++ b/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
@@ -1,0 +1,104 @@
+/*
+    Copyright (c) 2017 Salesforce.org
+    All rights reserved.
+    
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+ 
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT 
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS 
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE 
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, 
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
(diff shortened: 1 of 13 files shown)


src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -0,0 +1,104 @@
...
+
+        if (actionType == ACTION_DELETE_IMPORTED) {
+            return deleteDIRecords(false);    
+
+        } else if (actionType == ACTION_DELETE_ALL) {
+            return deleteDIRecords(true);    
+        }
+
+        String errMsg = String.isBlank(actionType)
+            ? 'Please specify an action type'
```
davidhabib: normally, we don't allow hardcoded strings, as they should be custom labels to support translation.  this is an example, however, of a use case our customers will not hit.  We control the buttons, and will be passing the correct arguments.  I'm ok if you leave this in, and prefer it without labels, as you have it.

RadSFDeveloper: That is right -- this message will be displayed only if the passed "action" page parameter is incorrect. This will not happen since we do have Data Import buttons setup correctly. This error message is not applicable to any user action on the page.

src/classes/BDI_DataImportDeleteBTN_CTRL.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>39.0</apiVersion>
```
RadSFDeveloper: @davidhabib , the new controller and VF page API versions are set to the latest API version (as indicated in the "PR Checklist" document). The package.xml version is 37.0 however. Can you please check into what version we should use for the new files?

davidhabib: good catch.  ideally they would be 37 to be consistent with the rest of our classes, and to avoid issues when classes call other classes on different versions.

RadSFDeveloper: The API version has been set to 37.0. The API version check for new classes was mentioned by Mike in the meeting today.

davidhabib:
@RadSFDeveloper I was just manually testing, and logged in my DE org with a profile without delete privileges on Data Import Records.  when I tried the button, I got this system error displayed, rather than in our new page.  any ideas why?

image

RadSFDeveloper:
@davidhabib , is the testing User profile added to the "BDI_DataImportDeleteBTN" VF page?

davidhabib:
@smarovich can you update the Critical Issues text on this pull request to be more appropriate for our users.  I was thinking something like: 

The "Delete All Data Import Records" and "Delete Imported Data Import Records" buttons now use a new visualforce page, which you must enable on the profiles of all users who need to use this functionality.

I assume we have an article we can link to that shows them how to enable visualforce pages on a profile.

thanks!

davidhabib:
@judisohn Can you own enabling this new page on the profiles in the Trial Template?  do you need us to create a work item?  thanks.

judisohn:
@davidhabib No problem. I need to make some tweaks to profiles in there anyway so I'll be sure this is done.

PR #2596, opened 2017-06-07
Resolved issue #2581 -- Update the Paid Amount for an Open Recurring …

### LINKED ISSUE #2581: If new Opp added to Open Recurring Donation, Paid Amount rollup is not calculated on edited Opportunities
[Thread.](https://powerofus.force.com/0D58000003PMOPZ)

Verified in test org. It appears that the when the Opportunity is updated to include the RD, it's not firing the trigger on RD the way it does on insert. Then the batch is not seeing that Opp in its query either?

While it's our preferred process to update an auto-created Opp rather than create a new one, this still feels like a bug to investigate.

Steps to repeat:

1. Contact with RD, Open (I did quarterly but that doesn't matter)
2. Create new Opp for Contact and populate RD lookup
3. Verify that Paid amount is updated with new opp
4. Create new Opp for Contact and leave RD lookup blank. Save.
5. Edit and add RD.
6. RD Paid amount not updated. Running Batch does not update. Refresh does not update.

image

image

joy-sfdo:
…Donation when an Opportunity is added.

# Critical Changes

# Changes

# Issues Closed
Fixes #2581

### DIFF
--- a/src/classes/RD_RecurringDonations_Opp_TDTM.cls
+++ b/src/classes/RD_RecurringDonations_Opp_TDTM.cls
@@ -58,11 +58,14 @@ public class RD_RecurringDonations_Opp_TDTM extends TDTM_Runnable {
                 //does it have a recurring donation reference?
                 //has it changed closed state, amount, or recurring donation changed?
                 Opportunity oppOld = (Opportunity)oldlist[i];
-                
-                if (o.npe03__Recurring_Donation__c != null 
-                    && (o.isClosed != oppOld.isClosed 
-                        || o.Amount != oppOld.Amount
-                        || o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c)){
+                if (o.isClosed != oppOld.isClosed 
+                    || o.Amount != oppOld.Amount
+                    || o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c){
+                    // If the Recurring Donation has changed, also update the Recurring Donation 
+                    // associated with the old (previous) version of the Opportunity
+                    if(o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c) {
+                        rdIds.add(oppOld.npe03__Recurring_Donation__c);
+                    }
                     rdIds.add(o.npe03__Recurring_Donation__c);
                 }
                 i++;
--- a/src/classes/RD_RecurringDonations_Opp_TDTM.cls
+++ b/src/classes/RD_RecurringDonations_Opp_TDTM.cls
@@ -56,13 +56,17 @@ public class RD_RecurringDonations_Opp_TDTM extends TDTM_Runnable {
             for (SObject sobj : newlist) {
                 Opportunity o = (Opportunity)sobj;
                 //does it have a recurring donation reference?
-                //has it changed closed state or amount?
+                //has it changed closed state, amount, or recurring donation changed?
                 Opportunity oppOld = (Opportunity)oldlist[i];
-                if (o.npe03__Recurring_Donation__c != null && (o.isClosed != oppOld.isClosed || o.Amount != oppOld.Amount)){
+                
+                if (o.npe03__Recurring_Donation__c != null 
+                    && (o.isClosed != oppOld.isClosed 
+                        || o.Amount != oppOld.Amount
+                        || o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c)){
                     rdIds.add(o.npe03__Recurring_Donation__c);
                 }
                 i++;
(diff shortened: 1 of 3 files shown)


src/classes/RD_RecurringDonations_Opp_TDTM.cls
```diff
@@ -56,13 +56,20 @@ public class RD_RecurringDonations_Opp_TDTM extends TDTM_Runnable {
...
-                //has it changed closed state or amount?
+                //has it changed closed state, amount, or recurring donation changed?
                 Opportunity oppOld = (Opportunity)oldlist[i];
-                if (o.npe03__Recurring_Donation__c != null && (o.isClosed != oppOld.isClosed || o.Amount != oppOld.Amount)){
+                if (o.isClosed != oppOld.isClosed 
+                    || o.Amount != oppOld.Amount
+                    || o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c){
+                    // If the Recurring Donation has changed, also update the Recurring Donation 
+                    // associated with the old (previous) version of the Opportunity
+                    if(o.npe03__Recurring_Donation__c != oppOld.npe03__Recurring_Donation__c) {
```
davidhabib: You need to make sure the RD isn't null.  adding null to rdIds will probably cause problems!

src/classes/RD_RecurringDonations_Opp_TDTM.cls
```diff
@@ -56,13 +56,20 @@ public class RD_RecurringDonations_Opp_TDTM extends TDTM_Runnable {
...
+                        rdIds.add(oppOld.npe03__Recurring_Donation__c);
+                    }
                     rdIds.add(o.npe03__Recurring_Donation__c);
```
davidhabib: similarly, you need to make sure the new opp's RD isn't null.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -1154,4 +1154,194 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
+        //--------------------------------
+        // Confirm the Opportunity has updated the npe03__Paid_Amount__c for both Recurring Donations (r1 & r2) 
+        rd1 = [select npe03__Paid_Amount__c from npe03__Recurring_Donation__c where id = :r1.id];    
+        System.assertEquals(null, rd1.npe03__Paid_Amount__c);
+
+        rd2 = [select npe03__Paid_Amount__c from npe03__Recurring_Donation__c where id = :r2.id];    
+        System.assertEquals(50, rd2.npe03__Paid_Amount__c);
+    }
+    
 }
```
davidhabib: your first test covers an opp changing from no RD to an existing RD.  your second test covers moving an opp from RD1 to RD2.  but I think you missed testing an opp changing from an RD to no RD.  not sure if that would have then flagged the issues I mentioned above, or whether we are getting luck in the code that there is no negative side effect of having nulls in our set rdIds.  we should add the test and make the code avoid this.  thanks.

src/classes/RD_RecurringDonations_Opp_TDTM.cls
joy-sfdo: Hi Dave, 

I didn't notice a problem when I set the Opp's RD to null. It updated the values on the RD accordingly. I'll test again.

src/classes/RD_RecurringDonations_Opp_TDTM.cls
joy-sfdo: Same here, setting the Opp's RD to null, removed the Opp from the RD and updated its values. I'll test again.

Thanks!

davidhabib: even if you can't reproduce a problem, it is poor programming practice to put null values onto a set, list, or map.  It can lead to bugs in soql and code that uses the collection, expecting all the values to be valid.

src/classes/RD_RecurringDonations_Opp_TDTM.cls
joy-sfdo: Hi Dave, 
I added the null check before adding to the rdIds.
Thanks!

src/classes/RD_RecurringDonations_Opp_TDTM.cls
joy-sfdo: Hi Dave,
Agreed. It can lead to bugs.
Thanks!

src/classes/RD_RecurringDonations_TEST2.cls
joy-sfdo: Hi Dave,
Added a new test case for changing from an RD to no RD.
Thanks!

PR #2604, opened 2017-06-12
Only consider active RecordTypes

mattparlane:
Hi all...

I recently added some inactive record types to assist a migration from one org to another, and I found that my tests now fail. I traced it to this line, where an inactive record type is being used. Setting the record types to active solves the problem.

Please note that I haven't actually tested this, I don't even know how. I'm not actually proposing this change, I just want to start a conversation with someone more knowledgeable than myself.

Thanks!


judisohn:
Thanks @mattparlane - are you on the Hub? If so, I think it might be best to start a conversation there in the [NPSP group](https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe) and explain the problem and then we can start an issue for investigation/change. We can reopen this pull request if it turns out that you did indeed solve the problem with this simple change. Thanks!

judisohn:
Okay @mattparlane - @force2b took a look and thinks this is indeed valid so I'm going to reopen and add to the backlog for a developer to consider and merge. Thanks for the contribution and your patience!

force2b:
We've moved this to another branch for final testing.

PR #2608, opened 2017-06-12
Feature/bdi donations

davidhabib:
@jjbennett This is the pull request to review.  Basically, it involved just some UI changes to the BDI page.  Now instead of having the settings directly on the page, I created a new visualforce component that hosts the settings.  This visualforce component is then included in the BDI_DataImport page, and later will get used when we handle the backlog item to create BDI Batches.  You'll see that some controller methods moved from BDI_DataImport_CTRL to BDI_SettingsUI_CTRL.  The BDI_DataImport page now also supports explicit going into edit mode on the settings.

# Critical Changes

# Changes

# Issues Closed


davidhabib:
@jjbennett OK, I fixed the two issues mentioned.  1) we validate Donation Matching Extension class if provided.  2) we put focus on the batch size edit control on edit mode.

src/classes/BDI_SettingsUI_CTRL.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>38.0</apiVersion>
```
jjbennett: I know the page needs to be 39 for slds but should the remaining new classes be 37?

davidhabib: good catch.  I just downgraded all the classes to 37.

PR #2609, opened 2017-06-12
Feature/joy best gift year - In Multicurrency org, Best Gift Year Total not rolling up correct value if opp/account currency does not match user currency.

### LINKED ISSUE #2570: In Multicurrency org, Best Gift Year Total not rolling up correct value if opp/account currency does not match user currency
If the user's currency (may also be related to corporate currency) is a different value than the currency on the Opp & Account, then Best Gift Year Value appears to be using the converted value of the user's currency, rather than the value of the Opportunity (and then further converting it). 

Screen shot from customer org illustrates issue based on user currency in GBP, Account in USD and a single $75K Opp in USD. 

image

joy-sfdo:
# Critical Changes

# Changes

# Issues Closed
Fixes #2570

### DIFF
--- a/src/classes/RLLP_OppRollup.cls
+++ b/src/classes/RLLP_OppRollup.cls
@@ -271,22 +271,26 @@ public without sharing class RLLP_OppRollup {
         // invokes code to get the currency of npo02__CombinedRollupFieldset__c oppty
         opptyCurrencies = (RLLP_OppRollup_UTIL.isMultiCurrency() ? RLLP_OppRollup_UTIL.rcfFindCurrency(objectList) : null);
         
-        for (sobject r : objectList){            
-            id aid = (id)(r.get('AccountId'));
-            if (aid != null){
-                updateRollupFromResult((sobject)(accountsToUpdate.get(aid)), r, opptyCurrencies);
-                
-                //calculate best gift year
-                decimal yrAmt = (decimal)r.get('TotalOppAmount');
+        for (sObject r : objectList) {
+            Id accountId = (Id)(r.get('AccountId'));
+            if (null != accountId) {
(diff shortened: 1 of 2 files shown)


davidhabib:
@joy-sfdo Nic is out for the next several weeks. so let's have @jjbennett review this pull request.  thanks!

joy-sfdo:
Hi @davidhabib, sounds good. I thought Nic's PTO started on Wednesday, I see now that it started today. I'll assign the PR to @jjbennett.

Thanks!

joy-sfdo:
@davidhabib - Thanks for assigning to @jjbennett.

src/classes/RLLP_OppRollup.cls
```diff
@@ -593,6 +599,10 @@ public without sharing class RLLP_OppRollup {
             
                 //calculate best gift year
                 decimal yrAmt = (decimal)r.get('TotalOppAmount');
+                // Handle currency conversion for the Household in a MultiCurrency org
+                if(RLLP_OppRollup_UTIL.isMultiCurrency()){
```
jjbennett: After our coding convention and whitespace talks I noticed the need for whitespace on the 3 conditionals, also I'm wondering if we'd want to move the variable declaration on 607 up so that we can use the variable vs retrieving it from the map multiple times?

joy-sfdo: Sounds good! I'll add the whitespace before the curly braces for the 3 conditionals.

Also, I like your suggestion about moving the variable declaration up instead of accessing the map multiple time. Accessing the map seems to be the pattern of that method. I wonder if there's a reason for that. I'll implement the change.

PR #2612, opened 2017-06-15
#2527 Manage household contact search does not support middle name

### LINKED ISSUE #2527: Manage Household New Contact search does not support standard Middle Name
https://powerofus.force.com/0D58000003IBdxL

force2b:
# Critical Changes

# Changes
* Add work-around support for known issue searching for Contacts when the MiddleName feature is enabled: https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM

# Issues Closed
Fixes #2527

### DIFF
--- a/src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
+++ b/src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
@@ -57,28 +57,22 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
             // note that our helper does FLS, so we don't have to do it here.
             String strSoql = HH_Container_LCTRL.getContactSelectQueryWithFls() + ' WHERE ';
 
+            Set<Id> contactIds = new Set<Id>();
             if (UTIL_Describe.isMiddleNameEnabled()) {
                 // When the Middlename field is enabled, there is a known issue (as of 6/15/2017) that
                 // prevents a query on the Name field from working properly if the record has a
-                // middlename or suffix value. To work-around this, we'll parse the queryValue into
-                // individual "words" and build a more generic query to find the records.
-                // Technically this will result in a larger result set than we really want
-                // however the only other solution would be to assume that three values will
-                // always match to first/middle/last. If the user enters two values, it could assume
-                // that matches to first + middle OR last. That gets more complicated and doesn't take
-                // into account compound firstnames such as "Anne Marie".
+                // middlename or suffix value. To work around this, use SOSL to search by name
+                // and then run a second query for the returned set of contacts.
                 // This should be considered a temporary work-around until such time as Salesforce
                 // fixes the known issue:
                 // https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM
-                strSoql += '(';
-                for (String p : preparedQueryValue.remove('%').split(' ')) {
-                    strSoql += ' FirstName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ' +
-                            ' MiddleName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ' +
-                            ' LastName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ';
+                List<List<SObject>> searchResults = [FIND :queryValue IN NAME FIELDS RETURNING Contact(Id, Name) LIMIT 100];
+                for (SObject s : searchResults[0]) {
+                    contactIds.add((Id)s.get('Id'));
                 }
-                strSoql = strSoql.removeEnd(' OR ') + ') ';
+                strSoql += ' Id in :contactIds ';
             } else {
-                strSoql += ' Name LIKE :preparedQueryValue';
+                strSoql += ' Name LIKE :preparedQueryValue ';
             }
             strSoql += ' AND Id not in :listCon LIMIT 100';
 
--- a/src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
+++ b/src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
@@ -48,7 +48,7 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
             if (listCon == null) {
                 listCon = new list<Contact>();
             }
-            
+
             // '%', '_', and '\' all are special characters in LIKE syntax.
             // We escape them here by prepending those characters with a backslash
             String preparedQueryValue = '%' + queryValue.replaceAll('([%_\\\\])', '\\\\$0') + '%';
@@ -58,7 +58,7 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
             String strSoql = HH_Container_LCTRL.getContactSelectQueryWithFls() + ' WHERE ';
 
             Set<Id> contactIds = new Set<Id>();
-            if (UTIL_Describe.isMiddleNameEnabled()) {
+            if (UTIL_Describe.isMiddleNameEnabled() && queryValue.length() > 1) {
                 // When the Middlename field is enabled, there is a known issue (as of 6/15/2017) that
                 // prevents a query on the Name field from working properly if the record has a
                 // middlename or suffix value. To work around this, use SOSL to search by name
@@ -68,7 +68,7 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
                 // https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM
                 List<List<SObject>> searchResults = [FIND :queryValue IN NAME FIELDS RETURNING Contact(Id, Name) LIMIT 100];
                 for (SObject s : searchResults[0]) {
-                    contactIds.add((Id)s.get('Id'));
+                    contactIds.add((Id) s.get('Id'));
                 }
                 strSoql += ' Id in :contactIds ';
             } else {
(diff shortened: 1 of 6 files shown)


src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
```diff
@@ -45,18 +45,43 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
...
+                    strSoql += ' FirstName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ' +
+                            ' MiddleName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ' +
+                            ' LastName LIKE \'%' + String.escapeSingleQuotes(p) + '%\' OR ';
+                }
+                strSoql = strSoql.removeEnd(' OR ') + ') ';
+            } else {
+                strSoql += ' Name LIKE :preparedQueryValue';
+            }
+            strSoql += ' AND Id not in :listCon LIMIT 100';
+
```
davidhabib: did you do any testing of this new query with QueryPlan to see if it performs much worse than the original?  I'm concerned this non-selective query will totally fail on LDV orgs.  I suspect the original one may as well; not sure if Name is indexed.  I'd like to see what QueryPlan says before we consider incorporating this.  thanks.

force2b: Thanks for the reminder on this. Meant to do that while I had DevConsole open earlier, but forgot.The ideal would be to just use the query on `Name`, but that's not possible right now due to the known issue. 

To check results, I ran the the full query in the LDV org. 

- The cost for the more complicated query came out to 0.42490 with an Operation Type of "Other".

- The cost for the simplified query on just `Name` came out to 0.21241 with an Operation Type of "Other" and an Index on the Name field

The pure query against name is definitely more selective, however I believe that the use of the `LIMIT` clause in the query will keep this from crashing an an extremely large org. Would be great to absolutely confirm that in a huge org, but for now as a comparison I ran a similar query against CampaignMember in the LDV org (424k records), filtering for `Name Like '%Raymond%'`. Without the LIMIT clause, the cost was 2.8333. With a LIMIT 100 clause the cost is only 0.000788.

davidhabib: great!  thanks for checking this.

force2b: Found the GUS work-item for this known issue and on it was a work-around I hadn't thought of -- using SOSL to query on the name. I confirmed this does work with Middle Name. To make this work, the code would have to run the SOSL first to retrieve a list of Contact Id's, and then re-query the Contact object for those Id's to return a result set to the component. Might be more efficient than the method I came up with initially.
`FIND {Raymond JOhn} IN NAME FIELDS RETURNING Contact(Id, Name)`

I can't speak on SOSL's effectiveness in an LDV org per-se. There's no query plan available for a SOSL statement, but the secondary query on Contact would definitely be indexed on Id.

What do you think?

davidhabib: since this is just a workaround, and limited to those orgs using middle name, I think going with what you have is a good route.  No need to change it all to use sosl when we think what you have will be sufficient.  thanks.

jjbennett: @davidhabib @force2b in testing this solution, the range of results is much greater than they used to be. If I search in an org with middle name enabled for someones full name it will return results for anyone that has the same last name or first name. The current behavior would only find the contact with the same name. Are we ok with the expanded results?

force2b: That is true. It's also that way with SOSL, though SOSL can return a more accurate result set. This would only be the case when the MiddleName feature is enabled, but still in a large org this could result in a lot more results. If getting a tighter result set is crucial, I suggest going with SOSL over the expanded SOQL query.

davidhabib: since their published workaround was to do name like first__last, have we tried Name like first%last?

jjbennett: I just tested Name Like first%last and that does work in dev console.

davidhabib: in the org that has Middle Name enabled?

jjbennett: Correct, I dropped the middle name from the query and used firstname%lastname and it returned the expected result.

force2b: If the user enters two terms ("Raymond John") how would the query know if that is meant to be firstname+lastname or firstname+middlename?

davidhabib: does it include any contacts that have a middle name specified.  my understanding was this bug in core prevented that from working correctly.

davidhabib: how about we just put % between all spaces in the input.  that should take care of it.

jjbennett: I did have to drop the middle name from the search so in @force2b example the system would be searching and assuming first and last name were provided.

force2b: That is correct. The core bug is that queries on name completely exclude the middle name value. So, if that value is included in the search term it will never mind a match querying against the name field. That's why I went with parsing out the terms and searching all 3 fields, or suggested going with SOSL which apparently does support Middlename.

Also, the following query returns 0 results:
`SELECT Id, Name, FirstName, MiddleName, LastName FROM Contact
WHERE Name Like '%Raymond%John%Alvarez%'`

jjbennett: @davidhabib I agree with @force2b that SOSL sounds like the most accurate way to work around this bug, do you see any issue with going that route?

davidhabib: ok, let's go ahead and give SOSL a try.  I know we use it in the Contact Merge page.  It will require the secondary soql to get the contacts, but now that I understand what Michael's solution is doing, SOSL seems worth pursuing.  thanks all.

src/classes/HH_AutoCompleteDataProvider_LCTRL.cls
```diff
@@ -45,18 +45,37 @@ public with sharing class HH_AutoCompleteDataProvider_LCTRL {
...
+            Set<Id> contactIds = new Set<Id>();
+            if (UTIL_Describe.isMiddleNameEnabled()) {
+                // When the Middlename field is enabled, there is a known issue (as of 6/15/2017) that
+                // prevents a query on the Name field from working properly if the record has a
+                // middlename or suffix value. To work around this, use SOSL to search by name
+                // and then run a second query for the returned set of contacts.
+                // This should be considered a temporary work-around until such time as Salesforce
+                // fixes the known issue:
+                // https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM
+                List<List<SObject>> searchResults = [FIND :queryValue IN NAME FIELDS RETURNING Contact(Id, Name) LIMIT 100];
```
davidhabib: do you need to do any escaping like you did for preparedQueryValue?  why does soql need that, but sosl doesn't?

force2b: Good question. Normally I'd use escapeSingleQuote() on a string, but that's not required when using a bound var (`:queryValue`). The wildcards (%) are not required for SOSL, and that replace method appeared to only be escaping slashes and percent signs, neither of which should be an issue in a SOSL query, so it didn't appear to be necessary. That said, including the replace() logic without appending the wildcards wouldn't hurt anything either.

jjbennett:
@force2b - I was getting ready to merge and wanted to give everything another once over since my focus was primarily on the SOQL. I noticed this error now occurs at the bottom of the page when Middle Name is enabled. "This page has an error. You might just need to refresh it. Error in $A.getCallback() [] Callback failed: apex://HH_AutoCompleteDataProvider_LCTRL/ACTION$queryObjects Failing descriptor: {markup://c:autocomplete}" - It seems like there needs to be at least 3 chars before a search result is returned so thinking it might be related? Also, if we are supported middle name do we want to allow the user to enter a user name on a new contact? I noticed we added it to updateable fields but not really updating it anywhere.

force2b:
Odd. Didn't see that at all yesterday. I'll review again over the weekend.

I was looking at adding it to the "new contact" part yesterday, however that get a bit more complicated. The Contact.Middlename field doesn't exist unless that feature is enabled, so we can't bind an inputField to it in the page without having the feature enabled. Technically there are probably some ways to get that to work as a dynamic bind, however I was concerned about how well LEX would handle that. Plus, Middlename is optional and I believe the New Contact pop-up doesn't even support Salutation, so it's pretty bare already.

force2b:
@jjbennett Not seeing any error when I load the Manage Households page with Middlename enabled in the org. Technically you're right though. SOSL requires at least 1 character to search. I'm just seeing a page error when I type in one character. There's clearly some browser or settings difference here, but either way I think it would be better to just not search until at least two characters have been entered.

jjbennett:
@force2b yes correct only when you start typing in the autocomplete component not on page load.

PR #2614, opened 2017-06-15
Feature/unmanagedweblinks

cdcarter:
# Critical Changes

# Changes
Use the namespace_inject feature of cumulusci to dynamically insert npsp__ into weblinks to apex pages only where necessary. This prevents unmanaged deployments from having to manually edit BDI weblinks.

# Issues Closed

--

Lets try this again.

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -34,6 +34,11 @@ tasks:
             dir: browser_tests
             subfolder: NPSP-Browser-Tests-master
 
+    deploy:
+        options:
+            namespace_inject: npsp
+            unmanaged: True
+
     deploy_dev_config:
         description: Deploys the post install configuration for an unmanaged DE org
         class_path: cumulusci.tasks.salesforce.Deploy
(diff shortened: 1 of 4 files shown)


davidhabib:
@cdcarter I'm not clear on how this proposed change might affect the dev team's unmanaged DE orgs.  Will the new files even commit to an org?  what happens when another change is made to the object that requires retrieving the metadata and committing to the project?  will the fixup be lost?  Also, who were you expecting to review this change?  thanks!

cdcarter:
@davidhabib - I assigned @jlantz as a reviewer - this was a change we discussed making offline.

that said, this would rely on the developer to use CumulusCI for all deploys, so perhaps this isn't feasible for the core definitions and we should just include this in dev_config without the namespace injection functionality. thoughts jason?

cdcarter:
closing this and the other dev_config related PR.

i'm having to make a few dev_config changes - and want to both discuss plan for merge&review w/ jason but also just consolidate them into one PR.

PR #2617, opened 2017-06-16
#2587 Fix invalid page error handling on the Manage Households page

### LINKED ISSUE #2587: Manage Household page does not fail gracefully if an invalid Primary Contact is entered
Go to a Manage Household page and put a name in Primary Contact that does not exist. Click Save. Get screenshot below.

If Primary Contact can't be found (was entered as a typo and not selected in lookup), we should show a pretty error and not this. :-)

image

force2b:
# Critical Changes

# Changes

# Issues Closed
Fixes #2587

### DIFF
--- a/src/components/UTIL_PageMessages.component
+++ b/src/components/UTIL_PageMessages.component
@@ -15,7 +15,7 @@
         </style>
     </apex:outputText>
 
-    <div class="slds" id="page_messages">
+    <div class="sfdo203 slds" id="page_messages">
         <div class="{!IF(useToast, 'slds-notify-container', '')}">
             <apex:repeat value="{!pageMessages}" var="m">
                 <div role="alert" class="slds-notify slds-notify--toast slds-theme--{!m.severityTheme}">
--- a/src/pages/HH_ManageHH.page
+++ b/src/pages/HH_ManageHH.page
@@ -11,6 +11,9 @@
                 color: red;
                 font-weight: 700;
             }
+            #contentWrapper .bPageFooter.noTableFooter {
+                display: none;
+            }
         </style>
                 
         <!-- start with a spinner here until lightning loads -->


jjbennett:
@force2b - I was getting ready to merge and wanted to give everything another once over since my focus was primarily on the SOQL. I noticed this error now occurs at the bottom of the page when Middle Name is enabled. "This page has an error. You might just need to refresh it. Error in $A.getCallback() [] Callback failed: apex://HH_AutoCompleteDataProvider_LCTRL/ACTION$queryObjects Failing descriptor: {markup://c:autocomplete}" - It seems like there needs to be at least 3 chars before a search result is returned so thinking it might be related? Also, if we are supported middle name do we want to allow the user to enter a user name on a new contact? I noticed we added it to updateable fields but not really updating it anywhere.

src/components/UTIL_PageMessages.component
```diff
@@ -15,7 +15,7 @@
         </style>
     </apex:outputText>
 
-    <div class="slds" id="page_messages">
```
davidhabib: since you are changing one of our shared components, did you test all the other pages to make sure this is OK?  I don't recall if those others are including the slds 203 resource or using another version.

force2b: Yes. I went looking for some pages that used this component and where I could generate an error or message of some kind to confirm that it was working. Of course, you know the package better than I, so maybe there are some specific pages I should look at.

Also, it hadn't occurred to me that different pages would use different versions of SLDS. I has assumed that since all apex/vf is on the same api versions, the slds version would follow the same standard. Looking at the static resources though, there is a second slds zip file. I'll look into this further to make sure that the additional scope doesn't break pages that use the other static resource.

force2b: Checked one of the other pages using the sdls static resources that also uses this component. Errors look fine.

davidhabib: thx for checking

PR #2619, opened 2017-06-19
Feature/bdi donations reviewed

davidhabib:
@judisohn @smarovich What should we put in release notes for this new big feature release?  do we have the final URL to the documentation we can point people to?  thanks.

# Critical Changes

# Changes

## Data Importer Updates

* The NPSP Data Importer now supports matching against Donations and Payments. There are several new Configuration Options that control if Donation and Payment matching is enabled, how to match, and what criteria to use when matching. The **Donation Matching Behavior** options control if the feature is enabled and how to match:

    * Do not Match - Disabled Donation and Payment matching. New records are created. This is the way the NPSP Data Importer worked previously.
    * No Match - Only import a record if it doesn't match an existing record.
    * Single Match - Only import a record if it matches a single existing record.
    * Single Match or Create - Import a record if it matches a single existing record; create a new record is no single match is found.
    * Best Match - Only import a record when it matches at least one existing record, and update the best matched record.
    * Best Match or Create - Import a record if it matches an existing record; create a new record if no match is found.

* The **Donation Matching Rule** options determine which fields to match against. The **Number of Days from Donation Date** field determines how many days from the Donation Date to consider when looking for a matching Donation or Payment. If you're working with a developer that would like to modify the matching behavior, use the **Donation Matching Implementing Class** field to specify the full developer name of an Apex class that implements the BDI_IMatchDonations interface. When left empty, the NPSP default implementation is used. For details on using the Data Importer to match against Donations and Payments, see the NPSP Administrator's Guide to Importing Donor Data (https://powerofus.force.com/articles/Resource/NPSP-Administrators-Guide-to-Importing-Donor-Data).

* We also revamped the Configuration Options on the NPSP Data Import page. The options are now grouped into sub-sections.

# Issues Closed


judisohn:
@cpolcyn heads up that I'll have work to do in the template org for this to enable to new fields and update the NPSP Data Import page layout (and will need to update the data dictionary), so please let me know when the release is pushed there. Thanks.

@smarovich In addition to the doc PDF that I know you have about ready, we'll also need to update the template (for the 'Imported' fields on Donation and Payment). @davidhabib we need it to be updated in the packaging org too, I think.

PR #2623, opened 2017-06-21
Household Naming: Group by last name when the LastName is the only token

RadSFDeveloper:
Group by unique last name when the LastName is the only token in a Household name/greeting format.
# Critical Changes

# Changes

# Issues Closed
2554


src/classes/HH_NameSpec.cls
```diff
@@ -217,76 +216,102 @@ public with sharing class HH_NameSpec implements HH_INaming {
     * return string
     */
     private static string strNameFromNameSpec(NameSpec ns, list<Contact> listCon) {
-        string strName = '';
-        integer cCon = listCon.size();          
-        integer i = 0;
-        boolean isNoLastnames = (ns.strFullNameSpec.indexOf(chLToken + 'LastName' + chRToken) == -1);
+        Integer size = listCon.size();
```
davidhabib: I'd prefer variables don't get renamed in existing code.  It makes the diff more complicated, and it changes the original author's intent.  In this case the variable name cCon represents the count of Contacts, which is more descriptive and useful in my opinion, than the generic name size.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -54,13 +54,13 @@ private with sharing class HH_NameSpec_TEST {
...
         if (strTestOnly != '*' && strTestOnly != 'OneContact') return;
         
-        list<Contact> listCon = new list<Contact>();
-        Contact c1 = new Contact (
-            Salutation = 'Skipper',
-            FirstName = 'John',
-            LastName = 'Doe'
-        );
-        listCon.add(c1);
+        Contact[] listCon = new Contact[] {
```
davidhabib: I'd prefer to not have stylistic changes made to code.  it's one thing if we have decided on a team on a new style.  but there are many areas where we have no defined style, and it is currently a matter of choice.  for those undefined styles, you should feel free to use the style you prefer in new code you are writing, but it is recommended that we leave the style alone in existing code.  this reduces the amount of code that the reviewer must review to ensure there are no problems introduced.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -95,18 +95,17 @@ private with sharing class HH_NameSpec_TEST {
...
-            FirstName = 'John',
-            LastName = 'Doe'
-        );
-        listCon.add(c1);
-        Contact c2 = new Contact (
-            FirstName = 'Jane',
-            LastName = 'Doe'
-        );
-        listCon.add(c2);
+        Contact[] listCon = new Contact[] {
```
davidhabib: another example of an unnecessary change in a test that wasn't substantively changed.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -144,18 +143,17 @@ private with sharing class HH_NameSpec_TEST {
...
-            FirstName = 'John',
-            LastName = 'Doe'
-        );
-        listCon.add(c1);
-        Contact c2 = new Contact (
-            FirstName = 'Jane',
-            LastName = 'Smith'
-        );
-        listCon.add(c2);
+        Contact[] listCon = new Contact[] {
```
davidhabib: unnecessary change.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -187,23 +185,21 @@ private with sharing class HH_NameSpec_TEST {
...
-            FirstName = 'Jane',
-            LastName = 'Doe'
-        );
-        listCon.add(c2);
-        Contact c3 = new Contact (
-            FirstName = 'Sally',
-            LastName = 'Smith'
-        );
-        listCon.add(c3);
+        Contact[] listCon = new Contact[] {
```
davidhabib: unnecessary change

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -230,7 +226,118 @@ private with sharing class HH_NameSpec_TEST {
...
+
+    /*********************************************************************************************************
+    @description
+    operation:
+        3 contacts, 2 different lastnames in household, contacts' last name is not in the sequence 
+    verify:
+        HH Name, Formal Greeting, and Informal Greeting are correct
+    **********************************************************************************************************/            
+    static testMethod void ThreeContactsTwoLastNamesNotInSequenceWhenTheNameFormatHasLastName() {
+        if (strTestOnly != '*' && strTestOnly != 'ThreeContactsTwoLastNamesNotInSequence') return;
```
davidhabib: your string doesn't match the method name, so it should be fixed, or removed.  Turns out that since the developer console and other tools now allow us to specify a test method, this strTestOnly trick is really no longer necessary.

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: These are tests that do not need individual contacts declared. Even though c1 and listCon[0] point to the same contact for example, it is easier to maintain and read the code if we add items to the list when it is constructed. It also reduces number of lines in the code.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -230,7 +226,118 @@ private with sharing class HH_NameSpec_TEST {
...
+
+    /*********************************************************************************************************
+    @description
+    operation:
+        3 contacts, 2 different lastnames in household, the contacts' last name is not in the sequence 
+    verify:
+        HH Name, Formal Greeting, and Informal Greeting are correct
+    **********************************************************************************************************/            
+    static testMethod void ThreeContactsTwoLastNamesNotInSequenceWhenTheNameFormatHasNoLastName() {
+        if (strTestOnly != '*' && strTestOnly != 'ThreeContactsTwoLastNamesNotInSequence') return;
```
davidhabib: strTestOnly string doesn't correctly match

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -239,29 +346,26 @@ private with sharing class HH_NameSpec_TEST {
...
-        );
-        listCon.add(c2);
-        Contact c3 = new Contact (
-            FirstName = 'Sally',
-            LastName = 'Smith'
-        );
-        listCon.add(c3);
+        if (strTestOnly != '*' && strTestOnly != 'ThreeContactsTwoLastnamesCustomFields') return;                
+        
+        Contact[] listCon = new Contact[] {
```
davidhabib: another unnecessary change

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: Shall we remove strTestOnly for all test methods in the class? If so, I will remove the strTestOnly from new test methods. Otherwise, will correct the name.

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -822,14 +917,14 @@ private with sharing class HH_NameSpec_TEST {
...
 
-        list<Contact> listCon = new list<Contact>();
-        Contact c1 = new Contact (
-            Salutation = 'Skipper',
-            FirstName = 'John',
-            LastName = 'Doe',
-            Primary_Affiliation__c = acc.Id
-        );
-        listCon.add(c1);
+        Contact[] listCon = new Contact[] {
```
davidhabib: another unnecessary change

src/classes/HH_NameSpec.cls
RadSFDeveloper: I will revert name back from "size" to initial "cCon".
Even though "size" might be generic name, it is used as a List method on the platform, so using a variable "size" for listCon.size() seemed ok.

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: Please see the reply above for the similar comment. The change was made to reduce number of lines in the test class especially since we do not reference Contact cX (where X=1,2,...) anywhere in the test method. Other test methods that reference individual Contacts were not updated, even though they could have accessed records as listCon[0...X-1] instead.

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: Fixed strTestOnly assignment for new test methods. Please advise if this check should be removed for all test methods in this class.

src/classes/HH_NameSpec.cls
RadSFDeveloper: Variable name update has been reverted back.

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: Fixed strTestOnly assignment for new test methods. Please advise if strTestOnly check should be removed for all test methods in this class.

judisohn:
@cpolcyn @RadSFDeveloper This isn't showing in the latest release notes. Did it trigger a beta build?

cpolcyn:
@judisohn yes you can see when David merged the Feature Test build passed. The release notes were changed after merge so did not show up in the beta notes, but will be included in the release.

PR #2630, opened 2017-07-07
Address related static flag redesign

davidhabib:
# Critical Changes

# Changes

# Issues Closed


davidhabib:
@force2b here is my rework of address's use of static flags to use a static counter than is now incremented and decremented by the address code, no longer relying on our execution context to clear the static flag.

davidhabib:
@force2b and I included a test that inserts 210 contacts, accounts, and addresses, and verified that under the old static flag code it failed, and with the new static counter code it succeeds!

PR #2632, opened 2017-07-11
2591: TDTM Static Flag Handling

### LINKED ISSUE #1699: If a WFR field update updates the Open Ended Status field on a Recurring Donation to Closed, triggers do not fire to clean up opps
https://powerofus.force.com/0D58000002FGsPq

theory to investigate from @njjc: triggers still fire a second time after workflow field updates could be in the same context and our code has a static flag set to prevent recursion so it runs it the first time, then thinks it's done and doesn't try the second time around. firing order does triggers first, then workflow rules then if the workflow has a field update, it fires triggers a second time so if those are all in the first context, first run of triggers sets static flag second run of triggers doesn't fire since the static flag is set.

### LINKED ISSUE #1960: Not all Allocations are created when open ended Recurring Donations are updated
When default allocations is enabled, there are recurring donations, and the forecast months setting is changed, then for each of the recurring donations there should be newly generated opportunities, payments, and allocations.  However, under some circumstances, not all of the allocations are being generated.

``` gherkin
Given Default Allocations are enabled
And Opportunity Forecast Months is set to 1
When I create 20 open ended recurring donations
And set the Opportunity Forecast Months to 12
Then there should be 240 Opportunity records related to the recurring donations
And there should be 240 Payment records related to the opportunities
And there should be 240 Allocation records related to the opportunities
And there should be 240 Allocation records related to the payments
```

The above scenario fails with only 200 Allocation records created, instead of the expected 240.

I have created a unit test to reproduce this bug: [RecurringDonationsAllocationLimitTest.cls](https://github.com/SalesforceFoundation/Cumulus/blob/recurring-donation-allocation-test/src/classes/RecurringDonationsAllocationLimitTest.cls)

**lurch: add

### LINKED ISSUE #2591: Allocations static flags issues.
Allocations triggers have multiple issues due to improper static flags implementation: 
- they don't run correctly after workflow field updates on opportunity
- they don't run correctly in when opportunity update DML happens before `after insert` triggers have a chance to fire

### LINKED ISSUE #2651: Cascade Delete Static Flag Issue
Cascade Delete uses a "set and forget" static flag that doesn't work with batches over 200.

force2b:
# Critical Changes

# Changes

- We redesigned how static flag variables are used by TDTM triggers for recursion control. These changes will ensure that the static flags work as expected when a single DML operation has more than 200 records. If you implemented custom code that results in insert/update of more than 200 records at a time, we recommend thorough testing in your Sandbox.
- We enhanced TDTM to cache objects in all After triggers, so that any subsequent Update trigger in the same transaction receives the most up-to-date version of the objects in the Trigger.old list. This fix addresses a known platform limitation that was preventing our trigger handlers from performing the correct logic after a field was updated by a workflow rule.

# Issues Closed

Fixes #2591 
Fixes #1960 
Fixes #1699 
Fixes #2651


judisohn:
FYI @smarovich for the release notes

judisohn:
Also need to link this to #2591?

force2b:
Yes, but this is still just a draft PR primarily for review. Not ready for QA or docs yet.

judisohn:
Gotcha, thanks!

njjc:
@force2b It looks like the most recent dev-to-feature flow caused a test to fail. Also, I was wondering if there is a high level doc explaining the which issues are being resolved and the design chosen?

force2b:
It was something I was going to discuss on the next call on this story -  this is an error I was getting sporadically in my local test runs, but I was not able to find any pattern. Probably works 8 out of 10 times. It's interesting that this is the first build failure with that error. Nothing should be sporadic in Unit Tests though, so this bothers me. It's only happening in the household naming test as well. Now that I know it's definitely not my dev org, I'm going to spend some more time trying to nail this down.

force2b:
Just to confirm that sporadic nature of this, I resubmitted the build and all the tests passed this time.

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -175,7 +172,11 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
             
             if (triggerAction == TDTM_Runnable.Action.AfterInsert)
                 copyRecurringDonationCampaignAllocations(listOppsForProcessing);
-            
+
```
force2b: @judisohn Came across this particular scenario while testing. I believe this is a bug that we should create a new issue for. What do you think?

force2b: @judisohn Ok, my notes are not showing up in the above. Here they are:

> TO-DO: If a null Opp.Amount field is populated on an Update AND the Opportunity has an attached campaign that has Allocations on it, the action below will NOT set the default allocations on the Opportunity using the allocation records from the Campaign. It will use the global default Allocation in the NPSP Settings if there is one though.

njjc: Yeah, that sounds like a bug and we should prefer the campaign allocations. Please do create an issue for it.

force2b: bug created

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -1073,8 +1073,14 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
...
         
         // if specified, also include Address Type
-        if (strAddrType != null) 
+        if (strAddrType != null) {
             addr.Address_Type__c = strAddrType;
+        } else {
+            // A null value will force a newly created Address record to use its defined default value for
+            // the picklist, which if not reflected here could cause isAddressChanged() to return true
+            // in cases when it should not.
+            addr.Address_Type__c = defaultAddressType;
```
RadSFDeveloper: @force2b , is this fix related or already fixed by https://github.com/SalesforceFoundation/Cumulus/pull/2635 ?

force2b: Good question. Definitely related, but not already fixed by that PR. 

In the method a new address SObject is created for `addrNew` by calling the `ADDR_Addresses_TDTM.copyAddressStdSObjAddr` method. That method returns a null value for AddressType if `con.npe01__Primary_Address_Type__c` is null - which it would be that early in the insert process. Your new code sets `con.npe01__Primary_Address_Type__c` to the value retrieved from the Address record, but that happens later in the flow.

force2b:
@smarovich and @judisohn This is still going through QA, but I updated the PR comments for the release notes. We can tweak as necessary to ensure a clear message is delivered during the release.

src/classes/ADDR_ProcessControl.cls
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /* @description Returns true if the recursionFlag value is greater than 1 */
+    public static Boolean checkRecursionState() {
+        UTIL_Debug.debug('>>>>>> checkRecursionState() ==> ' + recursionFlag);
+        return (recursionFlag > 0);
+    }
+
+    /*******************************************************************************************************
+    * @description Simple static flag to prevent execution of the address validation triggers
```
davidhabib: could you update the comment to be more clear.  this is about calling the address validation services, not validation within a trigger!  had me stumped for a bit why you'd add this flag as well as the separate recursionFlag.

src/classes/ADDR_ProcessControl.cls
```diff
@@ -0,0 +1,100 @@
...
+            validationRecursionFlag = 0;
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description To support DML operations with more than 200 records, use a static Set of Id's that
+    * have been passed to the verifyAddressRecords() method. This allows for a recursion check that goes
+    * by the Id's of the records as opposed to a simple boolean.
+    */
+    public static Set<Id> processedIdsForValidation = new Set<Id>();
```
davidhabib: remove since we are not using.

davidhabib: guess I was wrong, and you are using it in the address validator class.  why did you need to use this set methodology rather than a single flag/counter?

src/classes/ADDR_Validator_TDTM.cls
```diff
@@ -37,59 +37,65 @@
...
-	        	for(SObject so : newList) {
-	        		Address__c address = (Address__c)so;
-	        		if(!system.isFuture() && !system.isBatch()) {
-	        		    address.Verification_Status__c = Label.Addr_Pending_Verification;
-	        		} else {
-	        			address.Verification_Status__c = Label.Addr_Verification_Required;
-	        		}	
-        	    }
-	        }
+        if(!ADDR_ProcessControl.checkValidationRecursionState() && !recordsPreviouslyProcessed) {
```
davidhabib: why the need for both using a recursion state flag, and using the set?

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -53,34 +48,40 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
...
-            return wrapper;
-        } else if (objResult.getsObjectType() == Account.sObjectType && !hasRunForAccounts) {
-            DmlWrapper wrapper = runAccountTriggerHandler(newlist, oldlist, triggerAction, objResult);
-            hasRunForAccounts = true;
-            return wrapper;
+        if (objResult.getsObjectType() == Contact.sObjectType && AFFL_ProcessControl.doNotRunForContacts == 0) {
+            wrapper = runContactTriggerHandler(newlist, oldlist, triggerAction, objResult);
+
+        } else if (objResult.getsObjectType() == npe5__Affiliation__c.sObjectType && AFFL_ProcessControl.hasRunForAffiliations == 0) {
+            AFFL_ProcessControl.hasRunForAffiliations++;
```
davidhabib: why for addresses did you abstract the increment/decrement to methods, but with affiliations you just expose the counter?  I'm ok either way, but it seems like we should be consistent.

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -90,26 +83,27 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
 
         // dispatch to the correct handler
-        if (objResult.getsObjectType() == Allocation__c.sObjectType && !hasRunForAllocationsBefore && 
+        if (objResult.getsObjectType() == Allocation__c.sObjectType && !ALLO_ProcessControl.checkRecursionState() &&
             (triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate)) {
 
             runAllocationBeforeTriggerHandler(newlist, oldlist, triggerAction, objResult);
-            hasRunForAllocationsBefore = true;
-        } else if (objResult.getsObjectType() == Allocation__c.sObjectType && !hasRunForAllocationsAfter && 
+        } else if (objResult.getsObjectType() == Allocation__c.sObjectType && !ALLO_ProcessControl.checkRecursionState() &&
```
davidhabib: was the separation of Allocations Before and After flags not really needed?  i see you are now using the same flag/counter.

src/classes/ALLO_MakeDefaultAllocations_BATCH.cls
```diff
@@ -134,11 +134,11 @@ public class ALLO_MakeDefaultAllocations_BATCH implements Database.Batchable<sOb
         //No error handling necessary, because the only place this is called from is the widget that displays the batch
         //process status in the UI.
         if (!listAlloForInsert.isEmpty()) {
-            ALLO_Allocations_TDTM.resetFlags(true);
+            ALLO_ProcessControl.setRecursionFlag(true);
```
davidhabib: why don't these need to clear the flag after the DML?

src/classes/ALLO_MakeDefaultAllocations_BATCH.cls
```diff
@@ -134,11 +134,11 @@ public class ALLO_MakeDefaultAllocations_BATCH implements Database.Batchable<sOb
...
             insert listAlloForInsert;
         }
         if (!listAlloForUpdate.isEmpty()) {
-            ALLO_Allocations_TDTM.resetFlags(true);
+            ALLO_ProcessControl.setRecursionFlag(true);
```
davidhabib: why don't these need to clear the flag after the DML?

src/classes/BDI_DataImport_BATCH.cls
```diff
@@ -1328,8 +1328,6 @@ public class BDI_DataImport_BATCH implements Database.Batchable<sObject> {
         // now update the Contacts to create their Affiliations
 	    // unfortunately, we must manually clear these static flags since we are doing
```
davidhabib: need to remove these two lines of comments, since BDI no longer needs to clear these affiliation flags.

src/classes/HH_HouseholdNaming_BATCH.cls
```diff
@@ -85,7 +85,7 @@ public class HH_HouseholdNaming_BATCH implements Database.Batchable<SObject> {
     * @return void  
     ********************************************************************************************************/
     public void execute(Database.BatchableContext bc, Sobject[] result) {
-        HH_ProcessControl.inFutureContext = true;     
+        HH_ProcessControl.setRecursionFlag(true);
```
davidhabib: doesn't the execute method now need to clear the recursion flag at the end, rather than assuming it goes out of context?

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -531,7 +508,7 @@ private with sharing class ALLO_Allocations_TEST {
         setupSettings(new Allocations_Settings__c(Default_Allocations_Enabled__c = true, Default__c = defaultGau.id));
 
         //setting flags to not create default allocations for the new opportunities
-        ALLO_Allocations_TDTM.resetFlags(true);
+        TDTM_TriggerHandler.toggleTriggerState('Opportunity', 'ALLO_Allocations_TDTM', false);
```
davidhabib: why turning off the trigger directly, rather than continuing to use flags/counters?

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -375,10 +375,9 @@ public with sharing class LD_LeadConvertOverride_CTRL {
                 lc.setAccountID(CAO_Constants.getIndividualAccountID());
 	        }
 
-	        // avoid auto-affiliation creation
-            AFFL_Affiliations_TDTM.doNotRunForContacts = true;
-            AFFL_Affiliations_TDTM.hasRunForAccounts = true;
-            
+	        // avoid auto-affiliation creation by disabling the Affiliations Trigger for Contacts
+            TDTM_TriggerHandler.toggleTriggerState('Contact', 'AFFL_Affiliations_TDTM', false);
```
davidhabib: why turning off the trigger directly, rather than continuing to use flags/counters?

src/classes/OPP_CampaignMember_TDTM.cls
```diff
@@ -58,7 +58,7 @@ public class OPP_CampaignMember_TDTM extends TDTM_Runnable {
         
         dmlWrapper dmlWrapper = new DmlWrapper();
 
-        if (hasRun) {
+        if (recursionFlag > 0) {
```
davidhabib: seems inconsistent to not create a separate class for maintaining the flag/counter, like you did for all the other components.

src/classes/RD_ProcessControl.cls
```diff
@@ -38,11 +38,28 @@ public with sharing class RD_ProcessControl {
     /*******************************************************************************************************
     * @description flag to prevent recursion into the RD trigger handler/logic
     */ 
-    public static boolean hasRun = false;
+    private static Integer recursionFlag = 0;
```
davidhabib: so we've now encountered the recursionFlag, setRecursionFlag(), checkRecursionState() in many of our process control classes.  why not create a base process control class that these all inherit from that provides this implementation so it is truly shared.  then individual process control classes can add any additional logic or additional flags as needed.  but this way the common way is shared.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -83,7 +83,7 @@ private with sharing class RD_RecurringDonations_TEST2 {
         string closedstage = [select masterlabel from opportunitystage where isActive = true and iswon = true and isClosed = true limit 1].masterlabel;
         o.StageName = closedstage;
         o.CloseDate = system.today();
-        RD_ProcessControl.hasRun = false;
+        RD_ProcessControl.setRecursionFlag(false);
```
davidhabib: if we are now correctly incr/dec our flag in our TDTM code, why is setting the flag needed in our test code?  we've fixed this in all our other components.

src/classes/REL_ProcessControl.cls
```diff
@@ -34,5 +34,26 @@
 * @description Provides process control to prevent relationship creation recursion in triggers 
 */
 public with sharing class REL_ProcessControl {
-    public static boolean hasRun = false;
+    /*******************************************************************************************************
```
davidhabib: perfect example of being able to use a shared process control base class.

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+ *
+***/
+@isTest
+private class TDTM_DMLgt200_TEST {
+
+    /*********************************************************************************************************
+    * @description if you only want to run one test in this class, fill in its name here.
+    * if you want to run all tests, then use '*'
+    */
+    private static string strTestOnly = '*';
```
davidhabib: not a big deal, but now that salesforce allows us to individually run tests, we don't really need to use this hack anymore.  for new test classes I write, i don't put it in.  for new tests in existing classes, I still follow it to be consistent with the file.  one could probably make a case that we should then just remove it from the test class.

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+    *   NOTE: THIS WILL FAIL IN AN ORG WITHOUT THE STATIC FLAG CHANGES
+    */
+    @isTest
+    static void test_contacts_with_affiliations() {
+        if (strTestOnly != '*' && strTestOnly != 'test_contacts_with_affiliations') return;
+
+        npe5__Affiliations_Settings__c affiliationsSettingsForTests = UTIL_CustomSettingsFacade.getAffiliationsSettingsForTests(
+                new npe5__Affiliations_Settings__c(npe5__Automatic_Affiliation_Creation_Turned_On__c = true));
+
+        Id orgRTId = [SELECT Id FROM RecordType WHERE SobjectType = 'Account' AND DeveloperName = 'Organization' AND IsActive = True LIMIT 1].Id;
```
davidhabib: our tests cannot assume they have any account record types.  We tend to not specify the record type when creating test data.  And yes, if the org had the HH Account record type incorrectly set as default, the test might fail.  but that is probably a good thing, since that is an incorrect setup that will lead to problems in the org.

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+        Account orgAcct = new Account(Name='TestOrganization', RecordTypeId = orgRTId);
+        insert orgAcct;
+
+        Test.startTest();
+
+        // Create more than 200 Contact records and then validate that all of the Household Accounts were created
```
davidhabib: you might consider using UTIL_UnitTestData_TEST.CreateMultipleTestContacts() in the future.  lot's of helper methods in this class.

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+        for (Integer n=0; n<insertCnt; n++) {
+            contacts.add(new Contact(
+                    FirstName = 'TEST' + n,
+                    LastName = 'TestContact# ' + n,
+                    Title = 'TestTitle'
+            ));
+        }
+        database.insert(contacts);
+        system.AssertEquals(insertCnt, [SELECT Count() FROM Account WHERE RecordType.DeveloperName = 'HH_Account'],
+                'There should be exactly ' + insertCnt.format() + ' Household Account records');
```
davidhabib: you can't assume the org has this record type.  it is not part of the package (though we include it in the trial template).

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+    */
+    @isTest
+    static void test_contacts_with_primary_affiliations() {
+        if (strTestOnly != '*' && strTestOnly != 'test_contacts_with_primary_affiliations') return;
+
+        npe5__Affiliations_Settings__c affiliationsSettingsForTests = UTIL_CustomSettingsFacade.getAffiliationsSettingsForTests(
+                new npe5__Affiliations_Settings__c(npe5__Automatic_Affiliation_Creation_Turned_On__c = true));
+
+        Id orgRTId = [SELECT Id FROM RecordType WHERE SobjectType = 'Account' AND DeveloperName = 'Organization' AND IsActive = True LIMIT 1].Id;
+        Account orgAcct = new Account(Name='TestOrganization', RecordTypeId = orgRTId);
```
davidhabib: nix the recordtype

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+        insert orgAcct;
+
+        Test.startTest();
+
+        // Create more than 200 Contact records and then validate that all of the Household Accounts were created
+        // as well as all the related Affiliation recorsd
+        List<Contact> contacts = new List<Contact>();
```
davidhabib: consider using the UTIL_UnitTestData_TEST helper(s)

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+    * @description Test creating 200+ Contacts with Relationships
+    */
+    @isTest
+    static void test_contacts_with_relationships() {
+        if (strTestOnly != '*' && strTestOnly != 'test_contacts_with_relationships') return;
+
+        Test.startTest();
+
+        // Create more than 200 Contact records and then validate that all of the Household Accounts were created
+        List<Contact> contacts = new List<Contact>();
```
davidhabib: UTIL_UnitTestData_TEST

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+        for (Integer n=0; n<insertCnt; n++) {
+            contacts.add(new Contact(
+                    FirstName = 'TEST' + n,
+                    LastName = 'TestContact# ' + n,
+                    Title = 'TestTitle'
+            ));
+        }
+        database.insert(contacts);
+        system.AssertEquals(insertCnt, [SELECT Count() FROM Account WHERE RecordType.DeveloperName = 'HH_Account'],
+                'There should be exactly ' + insertCnt.format() + ' Household Account records');
```
davidhabib: nix the record type.  you can find them by querying on Account.
cHHAcc = [select Count() from Account 
                where npe01__SYSTEM_AccountType__c = :CAO_Constants.HH_ACCOUNT_TYPE]

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+                General_Accounting_Unit__c = gau2.Id,
+                Percent__c = 100
+        );
+        insert campAlloc;
+
+        Test.startTest();
+
+        // Create 200+ Opportunity Records
+        List<Opportunity> opps = new List<Opportunity>();
+        for (Integer n=0; n<insertCnt; n++) {
```
davidhabib: consider UTIL_UnitTestData_TEST.OppsForContactList() or one of the other helpers?

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+                MailingStreet = '1313 Mockingbird Lane',
+                MailingCity = 'Munsterville',
+                MailingState = 'Illinois',
+                MailingPostalCode = '06123',
+                MailingCountry = 'United States'
+            ));
+        }
+        database.insert(contacts);
+
+        system.AssertEquals(insertCnt, [SELECT Count() FROM Account WHERE RecordType.DeveloperName = 'HH_Account'],
```
davidhabib: see previous note on this query

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+    */
+    @isTest
+    static void test_household_account_naming() {
+        if (strTestOnly != '*' && strTestOnly != 'test_household_account_naming') return;
+
+        Test.startTest();
+
+        // Create more than 200 Contact records and then validate that all of the Household Accounts were created
+        List<Contact> contacts = new List<Contact>();
+        for (Integer n=0; n<insertCnt; n++) {
```
davidhabib: UTIL_UnitTestData_TEST.CreateMultipleTestContacts

src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -0,0 +1,465 @@
...
+                    MailingPostalCode = '95014',
+                    MailingCountry = 'United States'
+            ));
+        }
+        Test.StartTest();
+        database.insert(contacts);
+        Test.StopTest();
+
+        // check household created for contacts
+        system.AssertEquals(insertCnt, [SELECT Count() FROM Account WHERE RecordType.DeveloperName = 'HH_Account'],
```
davidhabib: fix query

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -242,4 +245,20 @@ public class TDTM_TriggerHandler {
...
     }
+
+    /*******************************************************************************************************
+    * @description Modifies the global static collection of Trigger_Handler records to disable or enable
+    * a specific TriggerHander by SObjectName and ApexClass name.
+    * @param objName SObject Name
+    * @param className Apex Class Name
+    * @param state True (enable) | False (disable)
+    */
+    public static void toggleTriggerState(String objName, String className, Boolean state) {
```
davidhabib: is this truly needed, or only used in places where you could have used existing flags/counters that were in use?  I'd prefer avoiding two different methodologies for tackling these issues if we can avoid it.

davidhabib:
@force2b @njjc I've finished my code review.  lot's of minor feedback, as well as the issue raised in michael's post in our quip document about processDML.  I'd recommend we address those issues before we have @jessingrass invest time in testing.  sound ok?

src/classes/ADDR_ProcessControl.cls
force2b: @davidhabib  The goal was to prevent the same address from being submitted to the validation more than once. If I'm not mistaken this is necessary because DML does not occur in the validator class as it would in our other triggers. Instead, it queues up an @future method to handle the validation.  In the event that more than 200 records are committed at once, the validator class would need to know which of those records were previously queued up for validation as well as handling general recursion into the validator class. Incrementing a var before calling the @future method and decrementing afterwards wouldn't help with recursion on the validation logic. I'll update the comments around this to clarify.

src/classes/AFFL_Affiliations_TDTM.cls
force2b: I considered at abstracting to a method, but because there were 3 static integers in play for Affiliations I felt it was adding more code than necessary. Plus usage for these 3 vars is limited to only the AFFL_Affilations_TDTM class. I'm happy to implement a method for these if you feel it would be more consistent.

src/classes/ADDR_Validator_TDTM.cls
force2b: Good question. I think I had left the original recursion check (changing `alreadyRun` to `checkValidationRecursionState`) and then at a later point added the the `recordsPreviouslyProcessed` verification as I got into DML > 200 tests. Testing further, it appears that the first check is not required here so I'll remove it.

src/classes/ALLO_Allocations_TDTM.cls
force2b: I know there was a reason I pulled that out, but I want to go back and re-test the logic. The performance results for the ALLO tests show an increase of one DML statement for a few of the ALLO unit tests, so I want to verify that I didn't miss something in the initial review.

src/classes/ALLO_MakeDefaultAllocations_BATCH.cls
force2b: Because these are at the end of batch iteration so there would be no additional actions performed after those dml statements (in the current execution context). Wouldn't harm anything to add them either.

src/classes/HH_HouseholdNaming_BATCH.cls
force2b: This is basically the same as with `ALLO_MakeDefaultAllocations_BATCH`. Once the batch iteration ends the execution context ends so any static vars are cleared. Doesn't hurt to reset either, though not explicitly required.

src/classes/ALLO_Allocations_TEST.cls
force2b: Mainly because it seemed more efficient and explicit in terms of objective

src/classes/LD_LeadConvertOverride_CTRL.cls
force2b: Mainly because it seemed more efficient and explicit in terms of objective

src/classes/OPP_CampaignMember_TDTM.cls
force2b: This was the example I called out in the Quip doc. My reasoning is that the static var is never referenced from outside the class, thus creating a ProcessControl class did not seem appropriate.

src/classes/RD_ProcessControl.cls
force2b: Great suggestion!

src/classes/RD_RecurringDonations_TEST2.cls
force2b: Not required. Just missed removing it.

src/classes/TDTM_DMLgt200_TEST.cls
force2b: Good point. I didn't have that in there initially, but as I started to dig into all the other unit tests I found this pattern and it seemed appropriate to add in. More than happy to remove.

src/classes/TDTM_DMLgt200_TEST.cls
force2b: I used it elsewhere, but I seem to remember deciding it didn't fit in all situations.

src/classes/TDTM_TriggerHandler.cls
force2b: It's used in a few places where a specific trigger is being disabled entirely. It's a bit different than a pure recursion check, though that can certainly work for those that check the recursion state up front (not all do). The other advantage of completely disabling the trigger is that it prevents it the class from ever being instantiated at all, which should help with performance.

force2b:
Thanks @davidhabib. I asked Jess to hold off until I'm done with the changes.

src/classes/ADDR_ProcessControl.cls
davidhabib: thanks.

src/classes/ADDR_Validator_TDTM.cls
davidhabib: do you think it's valid to just see if the first address in the newlist was already seen, and then blow off the whole set?

src/classes/AFFL_Affiliations_TDTM.cls
davidhabib: I agree it feels like overkill, but I think there is benefit in being consistent.  let's have you change these to use the methods.  thanks.

src/classes/TDTM_DMLgt200_TEST.cls
force2b: All of the helpers create one unique opp per one record of another object (Account or Contact), where here I'm just creating 210 Opportunities all linked to the same Contact. I could re-write to create 210 unique contacts first, but not sure it's necessary.

src/classes/ALLO_Allocations_TEST.cls
davidhabib: I'm torn on this.  i agree it is more explicit, but I suspect that some of the other recursive flag uses could also be accomplished this way.  I'm also concerned that people might start following this pattern and doing a set and forget, which seems dangerous given our static flag issues.  I realize this is just test code, but you do later use this method in production code.  @njjc thoughts?

src/classes/ALLO_MakeDefaultAllocations_BATCH.cls
davidhabib: I'd prefer seeing the explicit cleanup, to avoid people making assumptions about execution context, and accidentally following this pattern in non-batch scenarios.

src/classes/HH_HouseholdNaming_BATCH.cls
davidhabib: see previous comment about batch.  thanks.

src/classes/TDTM_DMLgt200_TEST.cls
force2b: Now I remember why this was problematic. That method creates contacts with the same last name, however that fails with a DuplicateCheck error. I can override the last name in a loop, but at that point I'm overriding most of the fields set in the helper method. What are the implications of updating the helper method to add the `i` counter to the end of the lastname for all contacts created?

src/classes/LD_LeadConvertOverride_CTRL.cls
davidhabib: @njjc thoughts on disabling triggers directly vs using process flags?

src/classes/OPP_CampaignMember_TDTM.cls
davidhabib: I think there is benefit in being consistent and following our new process control paradigm.  that said, you could create the process control class as an inner class private to just this outer class.

src/classes/TDTM_DMLgt200_TEST.cls
davidhabib: i'm guessing some amount of tests will break!  if you want to give it a try and see, go for it.  if it passes all tests, then its a good idea so we avoid dup errors.  but if it starts hitting asserts, I think you can move on and we'll leave it as is.  thanks.

src/classes/TDTM_DMLgt200_TEST.cls
davidhabib: got it.  ignore!  thanks.

src/classes/TDTM_TriggerHandler.cls
davidhabib: @njjc thoughts?

src/classes/TDTM_DMLgt200_TEST.cls
force2b: Actually I forgot that I can disable duplicate check with DMLOptions. That should resolve the issue.

src/classes/ALLO_Allocations_TEST.cls
njjc: I feel like this is fine since this is a test context. Incrementing the counter feels artificial; this solution does feel more explicit. Maybe we can toggle it back after the DML to not encourage set-and-forget?

src/classes/ADDR_Validator_TDTM.cls
force2b: I thought of it this way:
- The initial call to the class saved 200 of Id's being validated.
- A subsequent recursive call would re-submit one or all of those 200 records. 
- Thus if the first Id in the triggerNew collection exists in the saved list of Id's, then it's recursive and the whole batch can be ignored.

Though, I suppose there is another possible path:
- Recursion could submit a combination of previously processed Addresses _and_ one or more new addresses for validation.
- I'm not necessarily sure that is possible though given that we're controlling the address recursion. 
- Though unlikely, I suppose that a org could have a custom or managed trigger that inserts a new Address after a previously inserted/updated Address is validated. Certainly an outlier, but probably not impossible.

src/classes/RD_ProcessControl.cls
force2b: It doesn't appear that static methods are supported in Virtual/Abstract base classes, more specifically they are not inherited by the derived class. Basically, I can define a static method in the Virtual or Abstract base class, but that method is not visible in a class that Extends that base class.

force2b:
@davidhabib All changes committed include the refactor of ProcessControl classes

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -53,34 +48,40 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
...
             return null;
         }
+
+        // Process the DML
+        TDTM_TriggerHandler.processDML(wrapper);
+
+        // Decrement the recursion counters based on related processing logic
+        if (objResult.getsObjectType() == npe5__Affiliation__c.sObjectType) {
+            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.AFFL_hasRunForAffiliations, false);
+            if (TDTM_ProcessControl.checkRecursionState(TDTM_ProcessControl.flag.AFFL_hasRunForAccounts)) {
```
davidhabib: i don't understand this.  shouldn't the unsetting match the setting before the processDML?  why is this one nested and checking for the flag set on accounts and then clearing the flag on contacts?

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -90,26 +83,27 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
         }
 
         // dispatch to the correct handler
-        if (objResult.getsObjectType() == Allocation__c.sObjectType && !hasRunForAllocationsBefore && 
+        if (objResult.getsObjectType() == Allocation__c.sObjectType && !TDTM_ProcessControl.checkRecursionState(TDTM_ProcessControl.flag.ALLOC) &&
             (triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate)) {
```
davidhabib: sorry if I asked this before.  are you confident that the three separate old flags were not really needed, and that a single flag suffices?

davidhabib: @njjc thoughts (since you wrote this originally)?

davidhabib:
@force2b ok, I think I had two additional questions/comments.  once those are dealt with, then I think it will be ready for testing!  thanks Michael!

src/classes/AFFL_Affiliations_TDTM.cls
force2b: This is effectively the original logic substituting our flags. Though what I remember seeing with the logic here is that after DML, if the Affiliation trigger is running on the Affiliation object it clears the flag for hasRunForAffiliations, and if the Affiliations trigger on Account had previously run it clears the doNotRunForContacts flag because that flag is set in the Affiliations-Account logic to prevent the trigger from running on the Contact when the Affiliation is created/updated. This is one of those scenarios where a single TDTM trigger class is called by triggers on the Account, Contact and Affiliation objects.

src/classes/ALLO_Allocations_TDTM.cls
force2b: I'm reviewing these at the moment, and so far haven't found anything that appears to require these. However, they were certainly there for a reason so I'm still drilling into it.

force2b: Given the trigger and dml flow I'm not seeing why it has to track the before/after/opportunity operations separately as 3 different flags. Not that I can't reimplement that logic, but it appears to be unnecessary. I'm very curious what @njjc thinks about these 3 flags.

force2b:
@davidhabib We should be set pending @njjc's thoughts on the Affiliation trigger flags.

davidhabib:
I say you go ahead and start Jess to testing!
thanks,
Dave


src/classes/ALLO_Allocations_TDTM.cls
njjc: I've been doing some manual testing on allocations and it seems to be working well. I don't recall exactly why I split the allocations before/after flag, but I do remember having a single flag and running into some sort of bug that caused me to add the second one. It was definitely a reactive decision, not a proactive one. I think the new version is working as expected.

src/classes/ADDR_Validator_TDTM.cls
```diff
@@ -37,59 +37,72 @@
...
-	        		}	
-        	    }
-	        }
+    public override DmlWrapper run(List<SObject> newList, List<SObject> oldList,
+    TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+
+        // Look in the static list of previoysly processed Id's to determine if the record has already
+        // gone through this trigger. If so, then don't process it a second time.
+        Boolean recordsPreviouslyProcessed = false;
+        if (newList[0].Id != null && processedIdsForValidation.contains(newList[0].Id)) {
```
njjc: Isn't newList null for some triggers, e.g. AfterDelete? Couldn't this cause a null pointer exception?

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -87,13 +87,16 @@ public class TDTM_TriggerHandler {
         // Get the classes to run. Note that we are using the local object name, so it works for objects
         // from Cumulus both in the package and unmanaged dev orgs.
         List<SObject> classesToCall = dao.getclassesToCallForObject(describeObj.getLocalName(), thisAction);
-        UTIL_Debug.debug(LoggingLevel.WARN, '****Classes To Call for ' + describeObj.getLocalName() + ': ' + JSON.serializePretty(classesToCall));
+//        UTIL_Debug.debug(LoggingLevel.WARN, '****Classes To Call for ' + describeObj.getLocalName() + ': ' + JSON.serializePretty(classesToCall));
```
force2b: I had left this behind because it seemed like someone added it for other debugging that is probably not required at this point, but might be useful at some future point. I can remove it though.

src/classes/CON_ContactMerge_TEST2.cls
```diff
@@ -60,7 +60,7 @@ private class CON_ContactMerge_TEST2 {
         
         // note that we include addresses so the household pass will deal with merging household accounts with address objects.
         system.assertEquals(false, contactSettingsForTests.Household_Account_Addresses_Disabled__c);
-        system.assertEquals(false, ADDR_Addresses_TDTM.hasRunAddrTrigger);
+//        system.assertEquals(0, ADDR_Addresses_TDTM.cHasRunAddrTrigger);
```
force2b: Yes. Probably just missed this one.

src/classes/ADDR_Validator_TDTM.cls
force2b: That is true for delete triggers, though this particular trigger isn't called on a delete. Still, there might be some point in the future where it is, so it's a good idea to add the additional check.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -112,11 +112,12 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
         
         Map<Id,Account> primaryAccounts = new Map<Id,Account>([select id, npe01__One2OneContact__c, npe01__SYSTEM_AccountType__c from Account where id IN :oppAccounts]);
-        List<Contact> primaryContacts = [select AccountId,Account.Id,Account.npe01__SYSTEM_AccountType__c from Contact where Id IN :primaryContactIds];
-
-        for (Contact thisContact : primaryContacts)
-            if (thisContact.AccountId!=null)
-                contactsAndOneToOneAccounts.put(thisContact.Id,thisContact.Account);
+        if (!primaryContactIds.isEmpty()) {
+            for (Contact thisContact : [select AccountId,Account.Id,Account.npe01__SYSTEM_AccountType__c from Contact
+                    where Id IN :primaryContactIds AND AccountId != null]) {
```
force2b: Good question. The change was to only issue the query when `primaryContactIds` is not empty - saving a query in cases when that collection is empty. This sort of relates to the quip doc we were commenting on this morning. Does filtering out null for Id fields help? QueryPlan seems to indicate that it's questionable, which probably doesn't match up with the note in the Quip doc. However, I think it would be safer to remove the "AND AccountId != null" clause for now and filter those out within the for loop.

njjc: Agreed that would be safer. I know contacts with AccountId==null exist, so I don't think this would fit the description of "indexes that do not include nulls".

force2b: I believe that "indexes that do not include nulls" means that the Index itself has been updated/recreated to include nulls in it, since by default all Indexes do not include null values. An admin has to specifically request to include nulls in Indexes from Salesforce support.

njjc: Interesting, I hope we can get confirmation as to the exact meaning.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -195,40 +197,67 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
                 // note that we detect whether the address change is coming from the address trigger,
                 // in which case we don't want to create a new address.
-                if (ADDR_Addresses_TDTM.hasRunAddrTrigger) {
+                if (TDTM_ProcessControl.isRecursive(TDTM_ProcessControl.flag.ADDR_hasRunTrigger)) {
                 	//UTIL_Debug.debug('DJH**** stop Addr Trigger recursion in ADDR_Contact_TDTM');
                     continue;
                 }
+                // If the Before Insert/Update trigger on the Contact has already executed and taken action
+                // and this is an After Insert/Update trigger, simply reset the flag and exit out.
+                if (TDTM_ProcessControl.isRecursive(TDTM_ProcessControl.flag.ADDR_hasRunContactOnBefore) &&
```
force2b: @xinnanlisf This (`ADDR_hasRunContactOnBefore`) is the new flag/logic related to handling the Contact part of the address trigger. This is set/referenced in a few different places in the code base to ensure that recursion does not occur.

PR #2634, opened 2017-07-11
Limit query for default Recurring Donation RT to only active record types

### LINKED ISSUE #2633: Recurring Donations settings should not default to an inactive record type
From @mattparlane on 06/11/2017:
> 
> I recently added some inactive record types to assist a migration from one org to another, and I found that my tests now fail. I traced it to this line, where an inactive record type is being used. Setting the record types to active solves the problem.
> 
> Please note that I haven't actually tested this, I don't even know how. I'm not actually proposing this change, I just want to start a conversation with someone more knowledgeable than myself.
>

force2b:
# Critical Changes

# Changes

# Issues Closed

Fixes #2633


judisohn:
Just a note for release note review to make sure we're thanking Matt Parlane for the contribution.

PR #2635, opened 2017-07-12
Sync New Contact "Primary Address Type" with "Address Type"

### LINKED ISSUE #2199: Updating Address on Account sets 'Primary Address Type' to Home, but new contacts in HH do not have field populated
reported by Donna Whittig

Steps to repeat:
1. Household Account with members without any data in Billing Address fields.
2. Edit account to populate Address. Save.
3. Note that the 'Primary Address Type' field on Contacts are populated with 'Home' automatically.
4. Now add a new Contact to the Household (either directly in UI or Manage Household).
5. Note that the new Contact does not have the same Primary Address Type as the other members (it's blank)

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed
Fixes #2199

### DIFF
--- a/src/classes/ADDR_Addresses_TEST.cls
+++ b/src/classes/ADDR_Addresses_TEST.cls
@@ -1183,6 +1183,121 @@ public with sharing class ADDR_Addresses_TEST {
 
     /*********************************************************************************************************
     @description
+        Add an existing Contact to a Household. 
+        Pre-populate the Contact's address with the Household Address
+        in the same way Manage Household pre-populates Contact's Mailing Address with Account's Billing Address.  
+    verify:
+        The Contact's current Address is equal to the new Household Address.
+        Contact's "Primary Address Type" is the same as the Address' "Address Type".
+    **********************************************************************************************************/        
+    static testMethod void addExistingContactToNewHouseholdShouldSyncPrimaryAddressTypeWithAddressType() {
+        if (strTestOnly != '*' && strTestOnly != 'addExistingContactToNewHouseholdShouldSyncPrimaryAddressTypeWithAddressType') return;
+        
+        UTIL_CustomSettingsFacade.getContactsSettingsForTests(
+            new npe01__Contacts_and_Orgs_Settings__c (
+                npe01__Account_Processor__c = CAO_Constants.HH_ACCOUNT_PROCESSOR,
+                Simple_Address_Change_Treated_as_Update__c = true
+            )
+        );
+
+        Integer hhSize = 1;
+        Integer contactSize = 1; 
+        createHHTestData(hhSize, contactSize);
+
+        Account account = new Account(Name = 'AccountWithoutAddress');
+        insert account;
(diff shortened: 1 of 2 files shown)


src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -1183,6 +1183,121 @@ public with sharing class ADDR_Addresses_TEST {
...
+        insert account;
+        Contact contact = new Contact(LastName = 'TestPrimaryAddressTypeExistingContact', AccountId = account.Id);
+        insert contact;
+
+        Test.startTest();
+
+        ADDR_Addresses_TDTM.hasRunAddrTrigger = false;
+
+        contact.accountId = listAccT[0].Id;
+        update contact = populateContactAddress(contact, listAddrT[0]);
```
njjc: I've never seen this syntax before, could you explain the use?

src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -1579,4 +1694,99 @@ public with sharing class ADDR_Addresses_TEST {
...
+    // Helpers
+    ////////////
+
+    /*********************************************************************************************************
+    * @description Populates Contact's Mailing Address from the Address
+    * @param c Contact
+    * @param sourceAddress Address
+    * @return Contact
+    **********************************************************************************************************/
+    static Contact populateContactAddress(Contact c, Address__c sourceAddress) {
```
njjc: What's the reason for returning the contact? Can't you return void and the contact in the parameter will be updated anyway?

src/classes/ADDR_Addresses_TEST.cls
RadSFDeveloper: The "populateContactAddress" will populate Contact address from the source address and return the transformed Contact. The transformation of the Contact is returned even though the update of the address on Contact is by reference (the input argument Contact address update is visible after the method call). 
We could have "update populateContactAddress(contact, listAddrT[0]);" statement instead, but for the clarity and readability reasons, the method output is assigned to the "contact" variable. 
If you have other suggestions, please let me know.

njjc: I understand what it does, but it feels weird to do an update call and a variable assignment on the same line. It also feels a bit strange to reassign the contact variable to the same underlying contact, like we're passing around pointers for no reason. Would you mind putting the update on its own line, and give populateContactAddress a void return type? I'll merge this PR after that. Great work!

src/classes/ADDR_Addresses_TEST.cls
RadSFDeveloper: We can have void in the method signature  and the contact will be updated outside of the  "populateContactAddress" scope. I wanted to make the call more clear; I was following the rule " if a function is going to transform its input argument the transformation should appear as the return value".  However, I can easily change the method name and the unit test to execute the following two lines instead:
populateContactAddress(c, listAddrT[0]);
update c;

njjc: Interesting, I had never heard of that rule. It makes sense and does make it more clear, though, so I think you can leave it. I'll merge as is. Thanks!

PR #2644, opened 2017-07-18
BDI Batch support

davidhabib:
# Critical Changes
* Users who need to be able to run the NPSP Data Importer need to be granted permissions to view the new NPSP Data Import Batch object on their system profile.
* Field Level Security must also be enabled for the new NPSP Data Import object field, NPSP Data Import Batch.

# Changes
* The NPSP Data Import object contains a new optional lookup to a NPSP Data Import Batch object.  You may want to add the new field to the page layout for the NPSP Data Import object.

# Issues Closed


davidhabib:
@judisohn Well you found a real bug!  I had updated the tokens for the Donation Matching Behavior settings, not realizing that the visualforce component had the old values hard coded into it.  That's now fixed, as well as additional validation of all Data Import settings from both the UI and from calling data import directly.

So feel free to play with this branch.  As I mentioned, it has no UI yet, so you can only test running a specific batch by writing code to call BDI_DataImport_API.processDataImportBatches(list<ID> listBatchId)

davidhabib:
@njjc This pull request is ready for review, though I do not want it being merged until after the code freeze.  As for your comment to me about Invocable Methods, in looking at all the examples I could find, and Andy Fawcett's blog about Invocable Methods, they do tend to throw or propagate exceptions, just like any global API, rather than wrapping them and returning the exception in a returned object.  Here's his blog: https://andyinthecloud.com/2015/05/25/automating-org-setup-via-process-builder-and-metadata-api/

judisohn:
Thanks @davidhabib - is there a snippet you can give me to run anonymous to call the batch? Even just to get me started and I can edit for the specific batch.

davidhabib:
// create a batch
DataImportBatch__c batch = new DataImportBatch__c();
        batch.Name = name;
        batch.Batch_Process_Size__c = 200;
        batch.Contact_Matching_Rule__c = 'Firstname,Lastname,Email';
        batch.Donation_Matching_Rule__c = UTIL_Namespace.StrTokenNSPrefix('Donation_Amount__c') + ';' + UTIL_Namespace.StrTokenNSPrefix('Donation_Date__c');
        batch.Donation_Matching_Behavior__c = BDI_DataImport_API.BestMatchOrCreate;
        insert batch;

// now insert your Data Import records....

// invoke processing on the specific batch.
list<BDI_DataImport_API.ProcessDataImportBatchesResult> listResult = BDI_DataImport_API.processDataImportBatches(new list<ID>{batch.Id});

judisohn:
@davidhabib Thank you!!

njjc:
@davidhabib I think the invocable method return type needs a larger discussion. 

I agree that for an API called from apex, or for an invocable method called from process builder/flow, that propagating exceptions is fine.

In the case of calling the API from an external source, such as a website or standalone application,  including possible exceptions in the return value is important. The article you linked doesn't deal with that case.

For example, imagine in the current method that the batch job can't be started because a limit has been reached. An external system getting a blank response makes it difficult for them to do the right thing, in this case wait a while because there are too many batch jobs running.

While I think the utility in this current pull request is a bit limited due to the limited scope of the method, that won't necessarily be the case for future invocable methods that are consumable externally. I'd like for us to come to a generic return type that we can use going forward, instead of putting this in and having it be an outlier when we really need a composite response.

[Here's a good example](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/responses_composite.htm) of how to return values and possible exceptions. @cdcarter feels strongly that this is an important pattern for external consumption of APIs and I think we should do something of the sort going forward.

force2b:
So a response with something like:
`[{
    "apexJobId": "xxx",
    "response": "xxx"
}]`

cdcarter:
Nic and I had a brief brainstorm about this. The main point I wanted to make is that our API should fail fast to any caller - if the batch settings are invalid and no records will be processed, as a consumer of the API I want that in the initial response. 

I haven't tested the Actions API to see if it includes exception bodies in the response or if we need to handle them manually.

davidhabib:
@njjc just committed the work to annotate the exception with the batchId that caused a problem, and to rollback the transaction to stop other batch jobs that got queued.

src/classes/BDI_DataImport_API.cls
```diff
@@ -42,83 +42,69 @@ global with sharing class BDI_DataImport_API {
     * @description string values for different matching behaviors BDI can support.  couldn't use an enum
     * since we need to store it in a custom settings text field.
     */
-    global static final string DoNotMatch = 'DoNotMatch'; // Don't Match. Always create new Opportunities.
-    global static final string RequireNoMatch = 'RequireNoMatch'; // Always create.  Show error if any match.
-    global static final string RequireExactMatch = 'RequireExactMatch'; // Use exact match. Show error if there isn't an exact match or there are no matches.
-    global static final string ExactMatchOrCreate = 'ExactMatchOrCreate'; // Use exact match. Show error if there isn't an exact match. Create opportunity if there are no matches.
-    global static final string RequireBestMatch = 'RequireBestMatch'; // Use the best match. Show error if no match.
-    global static final string BestMatchOrCreate = 'BestMatchOrCreate'; // Use the best match. Create opportunity if no match.
+    global static final string DoNotMatch = 'Do Not Match'; // Don't Match. Always create new Opportunities.
```
njjc: Since these have been released in the wild, could there be any issues matching existing values?

src/classes/BDI_DataImport_BATCH.cls
```diff
@@ -39,47 +39,184 @@
...
+                    Batch_Size__c = diBatch.Batch_Process_Size__c,
+                    Contact_Custom_Unique_ID__c = diBatch.Contact_Custom_Unique_ID__c,
+                    Contact_Matching_Rule__c = diBatch.Contact_Matching_Rule__c,
+                    Donation_Matching_Behavior__c = diBatch.Donation_Matching_Behavior__c,
+                    Donation_Matching_Implementing_Class__c = diBatch.Donation_Matching_Implementing_Class__c,
+                    Donation_Matching_Rule__c = diBatch.Donation_Matching_Rule__c,
+                    Donation_Date_Range__c = diBatch.Donation_Date_Range__c
+                );
+            } else {
+                throw(new BDIException(string.format(Label.bdiContactCustomIdError, new string[]{batchId})));
```
njjc: This needs a new label, it's currently: The Contact Custom Unique ID setting field {0} must have matching fields in the NPSP Data Import object whose API Names are {1} and {2}.

src/package.xml
```diff
@@ -1664,6 +1679,12 @@
         <members>Contact-iw</members>
         <members>Contact-ja</members>
         <members>Contact-nl_NL</members>
+        <members>DataImportBatch__c-de</members>
```
njjc: @jjbennett can you remind me if we should include all these placeholders?

src/classes/BDI_DataImport_BATCH.cls
```diff
@@ -39,47 +39,184 @@
...
+            }
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description validates the data import settings.  Used by UI as well as at runtime by the batch.
+    * throws an exception if there are any invalid settings.
+    * @return void
+    */
+    public static void validateSettings(Data_Import_Settings__c diSettings) {
```
njjc: Yesterday we discussed basing the settings component on the DataImportBatch__c object because it's a superset of the custom setting. Do you want to update this method to be similar for consistency?

njjc:
@davidhabib we [hit that bug](https://mrbelvedereci.herokuapp.com/builds/5513/tests) I talked about this morning again in this PR's build...

src/package.xml
jjbennett: The consensus is to only update them when we have new translations to update. However, we are still trying to iron out the details for ignoring the changes but not removing them from the package. I wouldn't add any new English translation files though, I'm still hoping we can eliminate those from the package at some point.

src/classes/BDI_DataImport_API.cls
davidhabib: I needed to change the values since they are now also used in the batch picklist.  Thus I wanted the picklist values to use the same wording that our expanded help text values (that we will show in the UI) are using.  What will happen if any settings stored the old values is that it won't show a choice in the UI, and when they run the batch, they will get an error that the choice isn't valid.  then they can easily fix it.  Since I'm not convinced many customers have even used the new donation support, the amount of people hitting this will be quite small.

src/classes/BDI_DataImport_API.cls
```diff
@@ -42,83 +42,69 @@ global with sharing class BDI_DataImport_API {
...
+    global static final string RequireNoMatch = 'No Match'; // Always create.  Show error if any match.
+    global static final string RequireExactMatch = 'Single Match'; // Use exact match. Show error if there isn't an exact match or there are no matches.
+    global static final string ExactMatchOrCreate = 'Single Match or Create'; // Use exact match. Show error if there isn't an exact match. Create opportunity if there are no matches.
+    global static final string RequireBestMatch = ' Best Match'; // Use the best match. Show error if no match.
```
davidhabib: good catch!  thanks!

src/package.xml
```diff
@@ -1664,6 +1679,12 @@
...
+        <members>DataImportBatch__c-en_US</members>
```
njjc: @davidhabib let's remove en_US and the associated translation file

src/classes/BDI_DataImport_BATCH.cls
davidhabib: my bad!  I had added the label, but I forgot to use it!

src/classes/BDI_DataImport_BATCH.cls
davidhabib: no, I'd rather change it if needed in the new work, rather than hold this up.

src/package.xml
davidhabib: so I assume it's ok that DataImportBatch__c-de is being added.

src/package.xml
davidhabib: ok

davidhabib:
@njjc asked for changes just checked in.  we'll see if it passes build or hits that error.

davidhabib:
@njjc and it passes!

Issue #2316, opened 2016-10-03
CON_DoNotContact_TDTM resets fields changed in workflow rule field updates


njjc:
When a contact is part of a household with non deceased household members, and a workflow rule is created to uncheck "exclude from household name", the checkbox field is being overwritten by the trigger the second time it runs, i.e. the trigger context is still showing that the "deceased field" has changed, even though we're running through triggers a second time. [Reported here.](https://powerofus.force.com/0D58000002zMp66)

njjc:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000002y9shAAA" target="_blank">W-011494</a>

mbeaty-sf:
Salesforce has determined that this issue will not be explored further.

PR #2649, opened 2017-07-20
TDTM Workflow Rule changes

davidhabib:
This pull request enhanced TDTM to cache objects in all After triggers, so that any subsequent Update trigger, in the same transaction, will receive the most up to date version of the objects in the trigger old list.

This work was specifically done to work around a known platform limitation.  From the last paragraph on https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_triggers_order_of_execution.htm

"When a record is updated and subsequently triggers a workflow rule field update, Trigger.old in the last update trigger won’t contain the version of the object immediately prior to the workflow update, but the object before the initial update was made."

This standard behavior prevents our trigger handlers from doing the correct logic after a workflow rule runs, since we may not correctly detect which change just occurred.

# Critical Changes

# Changes

# Issues Closed


davidhabib:
@force2b @njjc @judisohn @jessingrass @dylanhansen 

Here is the "fix" to the workflow rule issue with triggers.  We do not want to even consider this change until we have finished dealing with our static flag pull request, and have merged that code in.  Since this is a major change with how our triggers will potentially handle workflow rules, we must exercise extreme due diligence to convince ourselves this is the correct solution.  Here are suggestions on what we can do to gain that level of confidence:

1. run our normal test suite and our LDV test suite, and compare performance/limit numbers to ensure this change hasn't affected anything in scenarios where no workflow rules are involved.

2. come up with a set of workflow rule tests that try modifying each of our objects and exhibit incorrect behavior in our current release, and then verify this branch resolves those issues.  Some example ideas:

- a workflow rule on Address that sets MailingCity to some value when MailingPostalCode = some value
- a workflow rule on Contact that sets MailingCity to some value when MailingPostalCode = some value
- a workflow rule on Account that sets BillingCity to some value when BillingPostalCode = some value
- a workflow rule on Opportunity that sets Stage to Closed / Won, when some custom field is set

Obviously the possibilities are endless, and there is no way for us to do any complete testing.  What I'm hoping we can achieve is a set of representative "reasonable" workflow rules that customers might actually try, that currently fail, that we can see if we've solved.  Given the potential effort, we should discuss our team's strategy on this, and how we want to proceed.

njjc:
This should fix #2316

judisohn:
If you need help coming up with examples to test against, let me know - I've seen many.

davidhabib:
@judisohn As I thought, you would have great knowledge of some workflow rule use cases.  That's why I'd like us to have a group discussion about who should be involved in this work, rather than assume development should do it all.  @dylanhansen @judisohn will you both be available to discuss at architecture review next thursday 7/27?

judisohn:
@davidhabib I'll be there!

PR #2652, opened 2017-07-20
Unsetting static flag at the end of cascade deletion transactions, an…

### LINKED ISSUE #2651: Cascade Delete Static Flag Issue

njjc:
…d clearing out cached results.

# Critical Changes

# Changes

# Issues Closed
- Fixes #2651


force2b:
@njjc Not sure why I missed this particular class, but thanks for catching it. One thing that I've done to verify that the flag change has worked as expected is to remove all references to it in unit tests. If the flag isn't required at all, then all the unit tests should still succeed. Once that works then the static flag var can be changed to private.

njjc:
Awesome, I'll do that! @force2b

force2b:
Odd build failure on this.

njjc:
I like the idea of using the centralized mechanism. I say don't merge this
yet and we'll decide once your work is in.

PR #2653, opened 2017-07-20
Manage Household: Add a Contact with the same Address creates multiple default Addresses

### LINKED ISSUE #2566: Manage Household: adding Contacts to same HH w/same Address creates multiple default addresses
[Hub thread.](https://powerofus.force.com/0D58000003NLLLX)

Replicated in latest build, exactly as customer describes. This only happens in Manage Household. If the Contact is reparented to Household, or if standard SFDC Account Merge is used, the Address record is not duplicated.

RadSFDeveloper:
Currently, an update of Addresses is done asynchronously (via a future Apex method) for Account Merge, Contact Merge and a Household update via Manage Household. So, due to the Address update lag, users might see Addresses that are not cleaned up if the future job does not execute relatively soon upon saving changes and closing Manage Household page. Thus, a change has been made to cleanup Addresses synchronously via Manage Household.
# Critical Changes

# Changes

# Issues Closed
Fixes #2566

### DIFF
--- a/src/classes/HH_Container_LCTRL.cls
+++ b/src/classes/HH_Container_LCTRL.cls
@@ -348,27 +348,6 @@ public with sharing class HH_Container_LCTRL {
     }
 
     /*******************************************************************************************************
-    * @description deletes the Contacts in the database
-    * @param listCon The list of Contacts to delete
-    * @return void
-    @AuraEnabled
-    public static void deleteContacts(list<Contact> listCon) {
-        try {
-            if (listCon.size() == 0)
-                return;
-            
-            // Check object permissions
-            if (!Contact.sObjectType.getDescribe().isDeletable()) {
-                throw new System.NoAccessException();    
-            }
-            delete listCon;
-        } catch (Exception ex) {
-            throw new AuraHandledException(ex.getMessage());
-        }
-    }
-    */
-
-    /*******************************************************************************************************
     * @description saves any changes to the Household in the database
     * @param hh The Household
     * @return void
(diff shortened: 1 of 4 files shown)


src/classes/HH_Container_LCTRL.cls
```diff
@@ -446,23 +425,62 @@ public with sharing class HH_Container_LCTRL {
...
+    static Boolean isNotEmpty(List<Account> accounts) {
+        return accounts != null && !accounts.isEmpty();
+    }
+
+    /*******************************************************************************************************
+    * @description Cleans up/refreshes the Addresses for specified Account Ids
+    * @param accountIds List of Account Ids
+    * @return void
+    */
+    static void cleanupAddresses(List<Id> accountIds) {
```
jjbennett: Is there any benefit to taking a list of Ids vs a single Id for the cleanupAddresses method?
Also wondering if we want to explicitly state private on these new methods?

RadSFDeveloper: The code was written so it is easy to extract and refactor if/when we decide to use a common method to process Address cleanup (see ACCT_AccountMerge_TDTM.handleAccountMergeFixups() for another call to cleanup Addresses). However, I can easily make the change to the method to accept one Account Id instead of the bulk. Please let me know.
Methods in Apex are private by default, so no need to explicitly specify it.

jjbennett: I could see in the future potentially pulling that out to a common method so that should be fine as is.
According to the doc here, http://developer.salesforce.org/Cumulus/Contributor/Coding-Conventions.html, it looks like we need to be explicit for docs to properly pickup the method. Thanks, I always did it but couldn't remember why, but was able to find that reference above.

RadSFDeveloper: Thanks! The method scope has been updated to private explicitly.

PR #2654, opened 2017-07-20
Fix unit tests failing on a org having Person Accounts enabled

RadSFDeveloper:
After "Account Record Type Defaults" setting is updated for a profile so default Record Type is Organization instead Household Account, unit test pass except two unit tests fixed in this PR.
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TEST.cls
+++ b/src/classes/ACCT_IndividualAccounts_TEST.cls
@@ -361,22 +361,6 @@ private class ACCT_IndividualAccounts_TEST {
 	        system.assertEquals(updatedContacts[0].OtherCity,updatedContacts[0].Account.ShippingCity);
         }        
     }
-    /*
-    private static testMethod void testDefault()
-    {
-       
-        boolean aDefaultIsSelected = false;
-        Schema.DescribeFieldResult F = Schema.sObjectType.Contact.fields.Salutation;
-            List<Schema.PicklistEntry> P = F.getPicklistValues();
-            for(Schema.PicklistEntry pe : P){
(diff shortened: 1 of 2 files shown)


src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1743,44 +1200,59 @@ private class ACCT_IndividualAccounts_TEST {
...
+    verify:
+        The specific system fields should not be set.
+    **********************************************************************************************************/ 
     public static testMethod void updateHHAccountRecType() {
+        Boolean isPersonAccountEnabled = UTIL_Describe.isValidField('Account', 'isPersonAccount');
 
-        List<RecordType> rectypes = [SELECT Id FROM RecordType WHERE SobjectType = 'Account' AND IsActive = true];
-        if (rectypes.size() < 2) {
+        String soql = String.format(
+            'SELECT Id{0} FROM RecordType WHERE SobjectType = \'\'Account\'\' AND IsActive = true',
```
davidhabib: out of curiosity, why did you use string.format with a parameter in the soql string, versus building it incrementally?  most code in NPSP would be:

soql = 'SELECT Id';
if (isPersonAccountEnabled)
   soql += ', isPersonType';
soql += ' FROM .....'

RadSFDeveloper: Good question. I think it is easier to read/maintain/debug the query when it is one String where variations are implemented using parameters {0}, {1}, ...
On the other side, there were different discussions of what is better wrt performance and memory fragmentation for String concat using "+" vs String.format() vs StringBuilder. The concat is re-creating a new String for every append (it is overhead). The String.format() is using less fragmentation. But, the main reason is readability and maintenance.

PR #2667, opened 2017-07-25
Intermittent assertion failure in PSC_PartialSoftCredit_TEST unit test

### LINKED ISSUE #2674: Intermittent assertion failure in PSC_PartialSoftCredit_TEST unit test
There is an intermittent unit test assertion failure in PSC_PartialSoftCredit_TEST.createDupePSCsNewOCRs() when Partial_Soft_Credit__c.Contact_Role_ID__c field is compared with an OpportunityContactRole.Id on sorted records.

Test failure details:
**PSC_PartialSoftCredit_TEST.createDupePSCsNewOCRs**
_Stack Trace:_ Class.PSC_PartialSoftCredit_TEST.createDupePSCsNewOCRs: line 408, column 1
_Message:_ System.AssertException: Assertion Failed: Expected: 00K0v000000KGoYEAW, Actual: 00K0v000000KGoaEAG

Notes:
- The issue seems to be related to retrieving records using Order By Contact_Role_Id__c field which is Text(18) (Unique Case Sensitive). As per @davidhabib research: "The following factors affect results returned with ORDER BY: Sorting is case insensitive."
- It is required to also investigate if other PSC unit tests verify Contact_Role_ID__c by comparing sorted records.

RadSFDeveloper:
Unit Test: Assert Opportunity Contact Role Ids are stored in Partial Soft Credits without relying on sorting of records
# Critical Changes

# Changes

# Issues Closed
Fixes #2674


davidhabib:
@njjc I'd like us to delay putting in a workaround until we can get a repro case to report to core.  otherwise, this bug will probably end up biting us in some other code at a later date.  Once we report to core, I'm fine putting in the workaround.

njjc:
I've spent a couple hours on this and have been unable to reproduce the failure outside of the build environment; putting in the extra debug statement to get more information about whether the issue is actually with soql ordering seems to have stopped the errors from occurring. It feels like a lot of work to chase down a reproducible case for core when the behavior isn't deterministic, and where the bug is orthogonal to what we're actually testing. I'm personally ready to give up on chasing down this bug, but was happy to fix the test case so that if it does crop up at least we won't be getting false positives. @davidhabib

src/classes/PSC_PartialSoftCredit_TEST.cls
```diff
@@ -123,19 +123,13 @@ public with sharing class PSC_PartialSoftCredit_TEST {
...
-        system.assertEquals(cOppTest, setOCRId.size());
+        system.assertEquals(cOppTest, listOCR.size());
         
-        // make sure OCR id's stored in PSCs
-        listPSC = [select Id, Contact_Role_ID__c from Partial_Soft_Credit__c];
-        system.assertEquals(cOppTest, listPSC.size());
-        for (Partial_Soft_Credit__c psc : listPSC) {
-            system.assertEquals(true, setOCRId.contains(psc.Contact_Role_ID__c));
-        }        
+        assertOpportunityContactRolesAreStoredInPartialSoftCredits(listOCR);
```
RadSFDeveloper: @njjc  Other test methods used Sets to verify OpportunityContactRole.Ids are stored in PSC, thus they are deterministic already. However, these tests were refactored to be consistent and to use the same data retrieval and verification methods throughout the test class. If there is any concern to making all unit tests uniform, I can easily revert back to the previous commit.

njjc: Looks great, thanks for fixing the same issue in other places!

PR #2669, opened 2017-07-26
Populate Previous Level Field when Level Field changes to blank

### LINKED ISSUE #2590: If Level is updated to blank value by criteria, Previous Level is not populated
[Thread.](https://powerofus.force.com/0D58000003QG85q)

Replicated. Happens when record does not meet any criteria for a Level.

Steps to repeat:

1. Custom number field on account, plus Level & Previous Level fields
2. Level with minimum value of 5, max of 10
3. Populate number field with 6, run Level - Level is correct
4. Change number to 2, run Level

The Level should appear in Previous Level and Level should be blank, instead both fields are blank.

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed
Fixes #2590

### DIFF
--- a/src/classes/LVL_LevelAssignBATCH_TEST.cls
+++ b/src/classes/LVL_LevelAssignBATCH_TEST.cls
@@ -629,4 +629,175 @@ private class LVL_LevelAssignBATCH_TEST {
         }
     }
 
+    /*********************************************************************************************************
+    @description
+        Execute Level Assign Batch on Contacts having Source Field value outside of all levels
+    verify:
+        The Level Field is changed to blank.
+        The Previous Level Field is populated with the Level Field as it was before the batch has been run.
+    **********************************************************************************************************/  
+    static testMethod void previousLevelShouldBePopulatedWhenCurrentLevelIsChangedToBlank() {
+        Level__c[] levels = new Level__c[] {
+            new Level__c(
+                Name = 'Tin',
+                Target__c = 'Contact',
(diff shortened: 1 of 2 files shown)


src/classes/LVL_LevelAssignBATCH_TEST.cls
```diff
@@ -629,4 +629,175 @@ private class LVL_LevelAssignBATCH_TEST {
...
 
+    /*********************************************************************************************************
+    @description
+        Execute Level Assign Batch on Contacts having Source Field value outside of all levels
+    verify:
+        The Level Field is changed to blank.
+        The Previous Level Field is populated with the Level Field as it was before the batch has been run.
+    **********************************************************************************************************/  
+    static testMethod void previousLevelShouldBePopulatedWhenCurrentLevelIsChangedToBlank() {
+        Level__c[] levels = new Level__c[] {
```
davidhabib: any reason you didn't use the helper createTestLevels(), like most of the other tests do?

davidhabib: never mind.  I assume it was to avoid the levels with no lower and no upper limits.

RadSFDeveloper: That is correct, already defined test levels covered all ranges so  could not create a test where the field is outside of levels. Although, I see now, I could have used a break between Gold and Platinum levels. 
On the other side, having two levels created in the test method made test clear of what data is within the limit and what is outside of levels.

PR #2680, opened 2017-07-28
DataImportBatch UI features

davidhabib:
# Critical Changes
* NPSP now includes support for Data Import Batches, which allows one to import a specific set of Data Import records, using a specific set of configuration settings.
    * Users who need to be able to run the NPSP Data Importer need to be granted permissions to view the new NPSP Data Import Batch object on their system profile.
    * Field Level Security must also be enabled for the new NPSP Data Import object field, NPSP Data Import Batch.
    * The NPSP Data Import Batches tab needs to be enabled on each profile that will use it.
    * The NPSP Data Import object contains a new optional lookup to a NPSP Data Import Batch object.  You may want to add the new field to the page layout for the NPSP Data Import object.

# Changes

# Issues Closed


davidhabib:
@njjc This pull request is now complete and ready for code review.  Even though we won't merge it until given the green light, I'd love to see the review happen this week.

@judisohn @smarovich I've updated the release notes with the critical configuration information, but we obviously need to decide about how we want to present any beta-doc, and refer to it in the release notes.  Let me know how to help.  thanks.

judisohn:
@davidhabib - @smarovich and I are going to talk tomorrow morning and come up with a doc plan for this so we're ready to go on release at the end of 2017.8a - thanks!

davidhabib:
@njjc I added a few commits to this pull request to respond to several issues judi identified in testing.  ready for review.

davidhabib:
@njjc since this pull request is not going in this release (due to lack of final documentation), I am continuing to add to it (no easy way to create separate pull requests that I am aware of).  I've just checked in support for Scheduling support.

davidhabib:
@judisohn would be great if you have time later this week to test out the batch scheduling support I've just checked into this branch.  thanks!

judisohn:
@davidhabib Will do, thanks!

src/classes/BDI_DataImportBatch_SEL.cls
```diff
@@ -51,8 +52,46 @@ public with sharing class BDI_DataImportBatch_SEL {
...
+    /*******************************************************************************************************
+    * @description queries the database for all Data Import Batch records that are marked with Scheduled
+    * processing, and have DI records to import.
+    * @return list<Id>
+    */
+    public static list<Id> selectWithScheduledProcessing() {
+        List<Id> listIdBatches = new List<Id>();
+
+        for (DataImportBatch__c diBatch : [SELECT
+            Id,
```
davidhabib: Two problems with this method.

1. in its current form, this SOQL is non-selective.  QueryPlan says it's a tablescan with cost of 1.96.  There is no index on the Process checkbox. 

2. if we decide to support a DataImportBatchName text field on the import object, then we can't use the subquery.  We'd need to find all of the batches who have unimported records (by either the batch lookup or batch name), and then query against batches that have Process checked and are in this set of batches with records to be imported.  Unfortunately, getting all the unique batch names or batch Id's requires using GROUP BY queries, and those are not going to be selective.  Just getting the unique batch Id's with:

Select NPSP_Data_Import_Batch__c from DataImport__c where NPSP_Data_Import_Batch__c <> null AND Status__c <> 'Imported' GROUP BY NPSP_Data_Import_Batch__c

is a Tablescan with cost 0.883

if we try to get both fields in a single GROUP BY query with something like:

SELECT NPSP_Data_Import_Batch__c, NPSP_Data_Import_Batch_Name__c FROM DataImport__c WHERE Status__c != 'Imported' GROUP BY NPSP_Data_Import_Batch__c, NPSP_Data_Import_Batch_Name__c 

we get a tablescan at cost 2.68

So I'm not sure how to tackle this to handle LDV orgs where query selectiveness is key.  Thoughts anyone? @njjc @joy-sfdo @force2b @RadSFDeveloper @jjbennett @cdcarter @judisohn @szeitz

davidhabib: One solution that comes to mind is instead of having a "Process with Scheduled Job" checkbox on the Batch, we could instead use a picklist which has the values "Yes" and "Completed".  When we finish running a batch, and we detect there are no failed import records for the batch, we could change the picklist to Completed.

Now instead of having to query for auto-process batches that still have unimported data import records, we could just rely on this one picklist for a single query against Batch, without having to look at Data Import records.

what do you think @judisohn @szeitz ?

cdcarter: I just asked @davidhabib a question in quip that was really leading to what he just suggested above. 

querying positively (i.e. = "Yes") will also be custom indexable by the customer if needed, as opposed to the negative (i.e. NOT "Imported")

davidhabib: should we go with field name "Process with Scheduled Job" with the values "To be processed" and "Completed"?  @smarovich any better wording you suggest for both the field name and values?

szeitz: I don't see any issues with this solution.

judisohn: @davidhabib I see you went with 'To be Imported' and 'Imported.' 

I have a test charter going on this with some questions.

davidhabib:
i'm closing this pull request, since we have outstanding spec issues, and we also don't plan on releasing this until we have all documentation in place.

PR #2681, opened 2017-07-28
Manage Household: If configuration includes MiddleName, an exception is generated at Manage Household page

### LINKED ISSUE #2658: If standard Middle Name enabled and used in Household Naming, error in Manage Household
Replicated in scratch org.

1. Enable Middle Names
2. Use this as Household Formal Greeting Formula: {!{!Salutation} {!FirstName} {!MiddleName}} {!LastName}
3. Go to Manage Household on any Account: duplicate field selected: MiddleName

image

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed
Fixes #2658

### DIFF
--- a/src/classes/HH_Container_LCTRL.cls
+++ b/src/classes/HH_Container_LCTRL.cls
@@ -242,7 +242,7 @@ public with sharing class HH_Container_LCTRL {
                     strSoql += ', MailingCountryCode, MailingStateCode ';
                 ******/
             
-            addHouseholdNamingContactFields(setContactFields);
+            setContactFields.addAll(HH_HouseholdNaming.iNaming.setHouseholdNameFieldsOnContact());
 
             return new list<String>(setContactFields);
         } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


src/classes/HH_Container_LCTRL.cls
```diff
@@ -271,13 +272,24 @@ public with sharing class HH_Container_LCTRL {
...
     }
 
     /*******************************************************************************************************
+    * @description Adds MiddleName field into the set of field names
+    * @param fieldNames Set of field names 
+    * @return void
+    */
+    private static void addContactMiddleNameField(Set<String> fieldNames) {        
+        if (UTIL_Describe.isMiddleNameEnabled()) {
+            fieldNames.add('MiddleName');
```
force2b: @RadSFDeveloper The only real change here is the switch from `Middlename` to `MiddleName`. The Set is Case Sensitive and the custom household name value is a open text field, so the User can simply enter the following value and cause this to still fail:

`{!{!Salutation} {!FirstName} {!Middlename}} {!LastName} Household`

I think the right answer here is to verify that "middlename" (in any _case_) is not already in the list of fields. Without looking at the code, I'm not 100% sure where the custom naming fields are added to the set - before or after the getContactFields() call. So either this method needs to do that check, or where ever the logic is to add the specific fields listed in the custom naming template.

RadSFDeveloper: Good question @force2b . I did consider case insensitive field names and added a method to process them as such, so Manage Household page does not generate an exception (difference can be seen when dev is compared with the first commit in this branch). However, this is not the only place when HH Naming configuration fields are used, so the exception is not local to the Managed Household page in the case configuration is not as expected. There is a backlog ticket that will ensure configuration is properly setup since the exception will be raised for any field (using firstname instead of FirstName for example). Therefore, the code to check for case insensitive names in Manage Household becomes obsolete if it is expected that field names will be as expected, ie case sensitive API names. Thus, the changes are made in the second commit of this branch to take care of the MiddleName issue for which the bug has been created.

force2b: I see your point. It does fail with `Lastname` instead of `LastName`. Certainly `MiddleName` is more correct as a field name than what I had as `Middlename`. I wonder where that existing backlog item is, or whether it's even a priority given that most Admin's will choose a pre-defined naming convention, and if they choose to override it'll default to one with the proper case. Maybe we should update the documentation to state that field names are case sensitive. What do you think @smarovich?

PR #2686, opened 2017-08-08
Recurring donations - Support option for "Always use last day of month"

### LINKED ISSUE #2034: Any Open Recurring Donation with Next Donation Date after the 28th is assumed to be last day of month
thread: https://powerofus.force.com/0D58000002ci7gK

repro steps: New open ended RD connected to a contact, Monthly Installment Period. Next Donation date on 4/28/2017. Default settings. 

First Opp is set for correct Close Date. Other 11 Opps are set to close on the last of the month. Should be 4/28, 5/28, 6/28, 7/28, etc. Instead it's 4/28, 5/31, 6/30, 7/31, etc.

force2b:
# Critical Changes

* The Recurring Donations object has a new field called **Always Use Last Day of Month**. To use this new field, administrators need to perform the following setup:
    * In the profiles for all users that need access to this field, edit Recurring Donations object settings and set the **Always Use Last Day of Month** field permission to Read Access and Edit Access.
    * Add the **Always Use Last Day of Month** field to the Recurring Donations page layout.

# Changes

* When the **Always Use Last Day of Month** field is checked on a Recurring Donation, monthly and quarterly installment Opportunities will be created with the Close Date set to the last day of the month. 
* When the **Always Use Last Day of Month** field is NOT checked on a Recurring Donation and the Next Payment date is the 28th, 29th, or 30th day of the month, new Installment Opportunities will be created with the Close Date set to that specific day. 
* A Next Payment or Date Established date of the 31st (of any month) is equivalent to having the **Always Use Last Day of Month** field checked. This new change only affects Recurring Donations with an Installment Period of either Monthly or Quarterly.

# Issues Closed

Fixes #2034

### DIFF
--- a/src/classes/RD_RecurringDonations.cls
+++ b/src/classes/RD_RecurringDonations.cls
@@ -260,39 +260,41 @@ public class RD_RecurringDonations {
     * currency iso fields.  Does not include the where clause since different callers have different
     * filtering needs.
     * @return string
-    */ 
+    */
     public static string strQueryRDNoWhere() {
-    	
-    	// these must be lowercase so set.contains() will work
+
+        // these must be lowercase so set.contains() will work
         set<string> existingFields = new set<string>{  'npe03__open_ended_status__c', 'npe03__next_payment_date__c', 'name', 'npe03__Paid_Amount__c',
-                                                       'npe03__organization__c', 'npe03__contact__c', 'npe03__installment_amount__c',
-                                                       'npe03__installments__c', 'npe03__amount__c', 'npe03__total__c', 'npe03__installment_period__c',
-                                                       'npe03__date_established__c', 'npe03__donor_name__c', 'npe03__schedule_type__c', 
-                                                       'npe03__recurring_donation_campaign__c', 'npe03__total_paid_installments__c', 'ownerid'};
-        
+                'npe03__organization__c', 'npe03__contact__c', 'npe03__installment_amount__c',
+                'npe03__installments__c', 'npe03__amount__c', 'npe03__total__c', 'npe03__installment_period__c',
+                'npe03__date_established__c', 'npe03__donor_name__c', 'npe03__schedule_type__c',
+                'npe03__recurring_donation_campaign__c', 'npe03__total_paid_installments__c', 'ownerid',
+                'always_use_last_day_of_month__c'};
+
         String queryRCD = 'select id';
         for (string s : existingFields){
-            queryRCD += ', ' + s;               
+            queryRCD += ', ' + s;
         }
-   
+
         //add any custom mapping to make sure we have the required fields
         for (string s : customFieldMappingSettings.keySet()){
-            string RDFieldName = customFieldMappingSettings.get(s).npe03__Recurring_Donation_Field__c;             
+            string RDFieldName = customFieldMappingSettings.get(s).npe03__Recurring_Donation_Field__c;
             if (!existingFields.contains(RDFieldName.toLowerCase()) && s != 'id'){
-               queryRCD = queryRCD + ',' + customFieldMappingSettings.get(s).npe03__Recurring_Donation_Field__c;
-               existingFields.add(RDFieldName.toLowerCase());   
+                queryRCD = queryRCD + ',' + customFieldMappingSettings.get(s).npe03__Recurring_Donation_Field__c;
+                existingFields.add(RDFieldName.toLowerCase());
             }
-        }       
-   
+        }
+
         //if currencyiso field exists add it to query for use later
-        if (isMultiCurrency)
+        if (isMultiCurrency) {
             queryRCD = queryRCD + ',CurrencyIsoCode';
-    
+        }
+
         queryRCD=queryRCD+' from npe03__Recurring_Donation__c';
-        
+
         return (queryRCD);
-    }   
-     
+    }
+
     /*******************************************************************************************************
     * @description Future method that creates new Opportunities for the specified Recurring Donations.
     * @param recurringDonations The list of RD's being created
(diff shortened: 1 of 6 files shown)


force2b:
@smarovich Please note the new field on the Recurring Donations object (with inline help text) as well as the detailed PR comments.

judisohn:
@jessingrass This is going to need extra testing attention. 

@force2b I have it in a scratch org and I'll start playing around with it as much as I can this week.

src/classes/RD_RecurringDonations.cls
```diff
@@ -436,7 +433,8 @@ public class RD_RecurringDonations {
                     *********************************/
                                         
                     OppCloseDate = getNextDate(OppCloseDate, r);
-                
+
+                    // TODO: Rewrite this to use the NPSP settings for Opportunity Naming
```
davidhabib: We haven't used TODO's in code, and I strongly discourage them.  If it something that needs to get done, we need to either address it now, or more typically, add it to the backlog.  In this case, I'm not even sure that Opp naming handles RD's where we want N of M included in the name.

src/classes/RD_RecurringDonations_TDTM.cls
```diff
@@ -86,6 +86,11 @@ public class RD_RecurringDonations_TDTM extends TDTM_Runnable {
                      r.npe03__Open_Ended_Status__c != system.label.npe03.RecurringDonationClosedStatus)) {
                     r.addError(system.label.npe03.RecurringDonationTooManyInstallmentsError);
                 }            
+                // Reset this field to false if the installment period is not Monthly or Quarterly
```
davidhabib: is there a case to be made that we should support last day of month for custom intervals.  For example, what if they define one for Every 2 months or every 6 months?

src/classes/RD_RecurringDonations.cls
force2b: I had seen a few TODO's in the code previously, so I thought this was the practice; though in this case I noticed it early on and added the TODO to remember to discuss this particular item.

I think you're correct that our general Opp Naming does not support "N of M" in the name. What bothered me about this is that we have a define opportunity naming pattern in settings and that is used for all new Opportunities except recurring donations. If the User runs the batch job to force renaming, it will overwrite whatever name is created here with the defined naming pattern. Seems inconsistent to me. I don't think it's something we should tackle in this story, but it might be worth adding to the backlog.

src/classes/RD_RecurringDonations_TDTM.cls
force2b: Possibly, though there are no current open issues requesting additional interval types.

davidhabib: @judisohn @szeitz any thoughts?

judisohn: @davidhabib I hadn't thought of this, but I'm with you... I think the 'last day of the month' functionality should be supported for any interval that is based on months.

force2b: Right now I think the only intervals we support based on months are Monthly and Quarterly, both of which are supported by the new feature. To my knowledge there's no current way for a customer to create a custom interval, such as "Every 2 months". It only supports the specific defined Intervals - Weekly, Monthly, Quarterly, Yearly, and Every 1st & 15th.

judisohn: @force2b It's a custom list setting - npe03__Custom_Installment_Settings__c

force2b: Thanks. I'll check it out.

force2b: @judisohn  Funny. I had this working (without realizing it) and then added logic to uncheck the "Always use last day of month" field when the installment period was not monthly or quarterly - which broke the use of that field for custom installment periods. Works now.

davidhabib:
@force2b @jessingrass let me know when testing is complete and this is ready to be merged, since I've already done the code review.  thanks.

judisohn:
@jessingrass @force2b I'm going to update my branch with the latest and do some exploratory testing on this tonight.

davidhabib:
In testing, I just created an open monthly RD with Date Established of 8/31/2017.  Now when the RD's get created, once it gets moved back due to a short month (ie, 9/30), it stays at that earlier date (so I get 10/30).  Once you roll thru Feb, now it gets pinned to the 28th.  While the fix was to improve the original fix to avoid this flaw, I thought we would still honor the original Date Established day in these scenarios?  @judisohn @force2b what is the new behavior we are trying for?

davidhabib:
Seems like this fix prevents someone from successfully specifying 29th or 30th and having it stick.  wasn't that the original problem that customers complained about?

judisohn:
@davidhabib @force2b  When I tested a date of the 31st, I thought I had observed that it stuck with the last day of all months? I'll try again first thing in the morning, but this was something I thought I had specifically checked for. If set for the 29th it would stay as the 29th for all months except February if the 'Last Day..." wasn't checked.

force2b:
@davidhabib and @judisohn  I definitely tested the 28/29/30 scenario as well. When selecting that as a date and "Use Last Day of Month" is unchecked, it should respect that day of the month for all future installments - adjusting as necessary for February. I'll retest as well.

davidhabib:
note that in my testing, I set Established Date, but left the other dates blank.  not sure if that is a nuance that caused me to see the bug, but you two to miss it.

force2b:
@davidhabib Thanks. That was the key difference.

force2b:
Fixed

judisohn:
Thanks @force2b and @davidhabib I'm testing one last time now to be sure everything appears okay, also to be sure how to communicate to affected customers how to correct their data.

PR #2693, opened 2017-08-11
If Automatic Household Naming is disabled and existing naming rules have syntax error, exception is thrown

### LINKED ISSUE #2476: If Automatic Household Naming is disabled and existing naming rules have syntax error, exception is thrown
Steps to repeat:

1. Household Naming Name rule with custom text that is invalid, such as {!{!FirstName}} {!LastName}} Household
2. Automatic Household Naming checked off
3. Edit any contact record

image

If Automatic Household Naming is disabled, any errors in the existing naming should be ignored.

RadSFDeveloper:
# Critical Changes

# Changes
When NPSP Settings > Household Naming Settings are saved, they are validated whether Automatic Household Naming is checked or not. Moreover, the settings validation will be done when NPSP Settings > Refresh Household Names button is clicked on and if the Automatic Household Naming is not checked. The Automatic Household Naming will be checked and the batch process will execute only if the Household Naming Settings are valid.

# Issues Closed
Fixes #2476

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -86,7 +86,7 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
             map<Id, Contact> mapContactIdContactOwnerChange = new map<Id, Contact>();
             map<Id, Id> mapContactIdAccountIdOldMoveOpps = new map<Id, Id>();
             map<Id, Id> mapContactIdAccountIdNewMoveOpps = new map<Id, Id>();
-            Set<String> hhNamingFields = HH_HouseholdNaming.iNaming.setHouseholdNameFieldsOnContact();
+            Set<String> hhNamingFields = getHouseholdNamingContactFields();
 
             
             if (triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) {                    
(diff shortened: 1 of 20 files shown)


RadSFDeveloper:
@smarovich , can you please verify the Release Notes Changes section?

judisohn:
@smarovich @RadSFDeveloper For the release notes, I think we can make this clear by pointing out that it applies if you select 'Other' for a name format and type in an alternate format. (Yes, it will be checked regardless but if they use a default it will pass and they'll have nothing to worry about.) :-) Maybe we can even give an example of a simple invalid one that will trigger an error, such as {!FirstName}} instead of {!FirstName} that will make it fail.

judisohn:
@smarovich We should also point out that field names are case sensitive... so FIRSTNAME will generate an error, while FirstName will not. You'll also get an error with custom_field__C vs custom_field__c.

njjc:
@RadSFDeveloper @judisohn why are field names case sensitive? That's never been the case as far as I'm aware.

RadSFDeveloper:
@njjc , if API Names are not code sensitive, then there are numerous errors as a consequence of that. For example, use firstname or middlename in the Household Name Format. Depending on the field, the error might be displayed when Save button is clicked or it can save successfully, but the error is shown when you save a Contact or add a new Contact via Manage Households. Related issue was fixed here (see comments for the discussion): https://github.com/SalesforceFoundation/Cumulus/pull/2681

Moreover, if we are to display Household Naming fields at a Lightning page to be displayed or populated, they will not be displayed since these fields are case-sensitive.

force2b:
@RadSFDeveloper and @njjc API Names by themselves are not case sensitive, however map and set keys are case sensitive which was the root cause behind the middle name error.

njjc:
@RadSFDeveloper and I had a great conversation about this. API names are case insensitive and can always be properly resolved to a single field in apex regardless of case. We agreed that it's a poor user experience to necessitate entering API names with a specific case, which is what this pull request is currently doing. She is going to further edit this pull request to no longer require any specific case to be entered by the user, and carefully handle maps and sets on the backend to avoid duplicate fields due to case differences.

RadSFDeveloper:
@njjc indicated we should not ask users to enter case-sensitive API Names. I am going to close the PR and change Household Naming processing to work without expecting users to enter case-sensitive API names.

njjc:
@davidhabib FYI don't review this yet, more changes are coming.

PR #2701, opened 2017-08-16
Automatically reschedule NPSP Scheduled Apex jobs in a Sandbox after refresh

### LINKED ISSUE #2684: On sandbox refresh, NPSP Scheduled Jobs are not functional and leave error in NPSP Settings
[Hub thread.](https://powerofus.force.com/0D58000003V7iyk)

[Known Issue](https://success.salesforce.com/issues_view?id=a1p300000008Ym6AAE) not targeted for a fix.

Error in NPSP Settings immediately upon sandbox refresh when clicking any sub tab:

image

Investigate whether we can detect these corrupted job records on refresh and delete/recreate them correctly.

force2b:
To work around a Salesforce Known Issue with corrupt scheduled jobs after a Sandbox refresh, automatically delete and reschedule all NPSP scheduled jobs when the NPSP Settings page is loaded the first time in a Sandbox org.

# Critical Changes

# Changes

# Issues Closed

Fixes #2684


src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -124,7 +124,11 @@ public without sharing class UTIL_MasterSchedulableHelper {
     public static void setScheduledJobs() {
 
         Map<String, CronTrigger> scheduledJobs = new Map<String, CronTrigger>();
-        
+
+        // always delete and reschedule all jobs in this particular scenario
+        Error_Settings__c orgErrorSettings = UTIL_CustomSettingsFacade.getOrgErrorSettings();
+        Boolean rescheduleAllJobs = (UTIL_Namespace.isSandbox() && orgErrorSettings.Sandbox_Jobs_Rescheduled__c == false);
```
jjbennett: Does it make sense to put all criteria into this new boolean so we don't need to look at 2 different locations for the condition?

src/objects/Error_Settings__c.object
```diff
@@ -70,6 +70,15 @@ To view these jobs, search for Scheduled Jobs in Salesforce Setup.</description>
         <type>TextArea</type>
     </fields>
     <fields>
+        <fullName>Sandbox_Jobs_Rescheduled__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>Used to track when NPSP apex jobs have been deleted and rescheduled in a Sandbox. It should remain unchecked in all Production orgs</description>
```
jjbennett: Does this need a string review?

src/classes/UTIL_MasterSchedulableHelper.cls
force2b: Do you mean create a method? Might be overkill since this condition will only be checked for at this one point.

src/objects/Error_Settings__c.object
force2b: I was thinking that since this field should never be visible/modifiable directly by end users the string/label wouldn't necessarily matter. What do you think @smarovich?

force2b: Also -- @judisohn FYI for the data dictionary

src/classes/UTIL_MasterSchedulableHelper.cls
jjbennett: no I just mean to grab the rest of the conditions from your if statement so you can just state if (rescheduleAllJobs) vs if (rescheduleAllJobs || this || that, so put this or that logic in your variable.

force2b: Ah. Thanks for clarifying. I suppose those could be combined into another boolean, but `rescheduleAllJobs` is referenced elsewhere by itself and the larger condition (`rescheduleAllJobs || currentJob.CreatedBy.Name == 'xx' || !currentJob.CreatedBy.IsActive)`) is only used that one time in the loop (where the values change each time).

jjbennett: Hmm yea forgot about that one, I will leave it up to you. I think it'd be cleaner even as two variables but not sure if it's worth it.

jjbennett: Or maybe rename it so it states that it's for sandbox?

force2b: How about `rescheduleAllSandboxJobs`?

src/objects/Error_Settings__c.object
smarovich: Sorry for the delay. If it's not customer facing, no need for me to review it.  And FWIW, it's clear to me anyway. :-)

davidhabib:
@force2b @jjbennett what is the status of this pull request?  if it has been reviewed and tested, I'd love to see it get merged soon.  we've got a lot of pull requests stacked up for this sprint.  thanks.

force2b:
The final decision on this is that the root issue has been addressed by Salesforce.com and as a result no actual code changes are required. The PR will be closed without merging.

PR #2702, opened 2017-08-17
Exclude household naming settings fields from queries when Automatic Household Naming is disabled

### LINKED ISSUE #2476: If Automatic Household Naming is disabled and existing naming rules have syntax error, exception is thrown

### LINKED ISSUE #2688: NPSP Settings > Household Naming Settings: Contact insert/update fails when Phone/Fax fields are included in Household Name Format
How to reproduce:
1. View NPSP Settings > People > Households > Edit > Set Household Name Format field to include the Phone or Fax in the format > Save
2. Create or select a Contact > Edit a Contact > Save > an error is generated.

RadSFDeveloper:
# Critical Changes

# Changes
We are now validating the syntax in Household Name Settings for custom name formats. To verify that your names are formatted correctly, we recommend that you go to NPSP Settings > People > Households, click **Edit**, then click **Save**. If there are any issues with your custom name formats, you will see an error. This validation occurs even when Automatic Household Naming is disabled.

# Issues Closed
Fixes #2476
Fixes #2688


src/labels/CustomLabels.labels
```diff
@@ -2752,6 +2752,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>- Household Duplicate</value>
     </labels>
     <labels>
+        <fullName>hhErrorInvalidContactFieldName</fullName>
+        <categories>Households, Household Naming</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>hhErrorInvalidContactFieldName</shortDescription>
+        <value>The field, {0}, does not match any field name on Contact.</value>
```
RadSFDeveloper: @smarovich , can you please indicate if the above error message needs any change? The error is raised if the Household Naming Settings contain a deleted or invalid Contact field. The error might be displayed when a Contact is created/updated or a new Contact added.

judisohn: @smarovich @RadSFDeveloper We should be clear that it's the API Name here, it's so confusing for folks between the Label, Developer Name and API and that's where I often see them trip up here.

RadSFDeveloper: @judisohn , good point! 
@smarovich can you please advise what should be the best error message?

davidhabib:
@RadSFDeveloper @judisohn @smarovich I'm not sure this fix warrants any extra release notes text in the Changes section.  We typically don't explain the nature and behavior changes of each fix, unless it is something important the customer should know about.  What do you think?

src/labels/CustomLabels.labels
davidhabib: This error can get displayed in places not obviously related to the settings page, such as creating a new contact, if they already have invalid formats saved.  Here's what the message is when trying to save a new contact (note I bold italicized the part we control):

Error: Invalid Data. 
Review all error messages below to correct your data.
Apex trigger TDTM_Contact caused an unexpected exception, contact your administrator: TDTM_Contact: execution of BeforeInsert caused by: HH_HouseholdNaming.HouseholdNamingException: _**The field, FirstName}, does not match any field name on Contact.**_: Class.HH_HouseholdNaming: line 259, column 1

I suggest that we add more context to the error so it says something like:

Invalid {0}: The field,  FirstName}, does not match any field name on Contact.

where we'd replace {0} with Household Name Format, Formal Greeting Format, or Informal Greeting Format.  this way the user would have more context on how to resolve the problem.

davidhabib: I just noticed that within the Settings page, you do prefix the error with "Error in Informal Greeting Format: ..." so can we also include the appropriate prefix in other scenarios?

RadSFDeveloper: Thanks @davidhabib for the feedback. The error is displayed when Household Naming Contact fields are retrieved from ` iNaming.setHouseholdNameFieldsOnContact();` method and then is used in Contact queries throughout the system whenever Household Naming should be updated. 
The method can be implemented by a client so fields do not have to belong to any Household Naming Format field. Additionally, we would need to parse each format field or tried to execute it in order to find which one is faulty (expensive operation?).
Can we make an error more generic but still indicate it is Household Naming Settings Format field? For example:  

Invalid Household Naming Settings format: The field, FirstName}, does not match any field name on Contact.

davidhabib:
@RadSFDeveloper when I have a saved invalid informal name format (which I did thru salesforce setup ui, since your code now prevents this of course!), and I bring up the Manage Household page, it comes up with two "Script thrown exception" errors on the page, but no details.  Sounds like the javascript in the lightning component needs to be more robust?

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1299,6 +1299,36 @@ private class ACCT_IndividualAccounts_TEST {
...
 
+    /*********************************************************************************************************
+    @description 
+        Test Contact insert and update when Household Naming Settings is invalid
+        and the Automatic Household Naming is turned off 
+    verify:
+        The Contact is created and updated successfully
+    **********************************************************************************************************/ 
+    private static testMethod void contactDmlShouldSucceedWhenHHNamingSettingsIsInvalidAndAutomaticHouseholdNamingIsTurnedOff() {
+        turnOffAutomaticHHNaming();
```
davidhabib: i'm not used to our code using such long and verbose method names.  we certainly don't do it in normal class methods.  we try to come up with a reasonable sized name that conveys intent, but always provide method headers to provide full details.  why should test methods be any different?

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1337,4 +1367,71 @@ private class ACCT_IndividualAccounts_TEST {
         
         return newRecTypeId;
     }
+
+    /*********************************************************************************************************
```
davidhabib: since you repeat all of these test helper methods in multiple test classes, i'd prefer seeing them in one of the HH test classes, and just made public so the other test classes can use them.  that's a paradigm we've used in other components.

src/classes/ACCT_IndividualAccounts_TEST.cls
RadSFDeveloper: These tests not only verify features but methods as well. When a test is selected (via Developer Console) it is beneficial to know what is being tested and what expected behaviour is (comments specified above tests in the Apex class are not visible). The clear name is also beneficial when test fail. Moreover, there are many tests verifying the same method on different conditions, so it is important for the name to indicate difference. 

I usually try to name test as _featureOrMethodTestedShouldProduceResultWhenConditions_. I am open to any test naming convention that is not verbose and clearly explains what is being tested and expected.

src/classes/ACCT_IndividualAccounts_TEST.cls
RadSFDeveloper: Good point, I will move them to one common test util class.

RadSFDeveloper:
@davidhabib , the "Script-thrown exception" error is displayed in this branch not in the current build which displays "Invalid field FirstName} for Contact" error. I will fix that.
The Contact cards are not displayed in current build either, so I will log a new issue for it.

src/classes/UTIL_Describe.cls
```diff
@@ -303,6 +303,109 @@ public class UTIL_Describe {
...
+
+        return null;
+    }
+
+    /*********************************************************************************************************
+    * @description Returns Sobject field results mapped by a lower case field name or a relationship name
+    * @param sobjectType Schema.SobjectType
+    * @return Map<String, Schema.DescribeFieldResult>
+    */
+    public static Map<String, Schema.DescribeFieldResult> getFieldResultsByName(Schema.SobjectType sobjectType) {
```
davidhabib: how is this different than our existing getAllFieldsDescribe(String objectName)?  it looks like you are doing work to ensure the keys of the map are all lowercase.  Is that not already the case in what Salesforce describe calls do?  If not, and we need them to, would it be better to have changed our existing getAllFieldsDescribe()?  Or if we feel like we need both flavors (mixed case keys and lowercase keys), can we use a naming convention that makes it obvious to developers when to use which one?   By now having getFieldResultsByName() and getAllFieldsDescribe(), I don't think I'd know which to use without having to look at the code for each to discern the differences.  Also, while I'm not a big fan of some of our describe code, it does try to do the work to cache describe information to avoid repetitive calls.  Your new methods are not using those caches, so I'm a bit concerned we may be adding some additional processing overhead and costly describe calls we could avoid.

RadSFDeveloper: `getFieldResultsByName()` returns fields and relationships while `getAllFieldsDescribe ()` resolves fields only. The Contact fields used in Household Naming Settings can be relationships so the method using `getAllFieldsDescribe ()` would not resolve relationship fields (Account.Name, Account.npe01__One2OneContact__r.LastName).

I did not change the existing method since the scope of the change would be more than this ticket and increase testing effort. Yes, the new Describe method do not store DescribeFieldResults into a static variable, so that is a negative side-effect if there are repetitive calls. 
Can I change the existing `getAllFieldsDescribe ()`  to return relationships as well (I would be happy to)?

src/classes/HH_HouseholdNaming.cls
```diff
@@ -218,6 +233,39 @@ public without sharing class HH_HouseholdNaming {
...
+        if (!isAutomaticNaming()) {
+            return new Set<String>();
+        }
+    
+        Set<String> fieldNames = iNaming.setHouseholdNameFieldsOnContact();
+        if (fieldNames == null || fieldNames.isEmpty()) {
+            return new Set<String>();
+        }
+
+        Map<String, Schema.DescribeFieldResult> contactFieldResultsByName = UTIL_Describe.getFieldResultsByName(Contact.SObjectType);
```
davidhabib: I'm having trouble understanding why you couldn't just loop thru the set of fields given to you by iNaming, and call UTIL_Describe.isValidField(String objectName, String fieldName).  Is it that isValidField wouldn't handle relationship fields (eg, Account.Name from Contact)?  Would it have made more sense to add the functionality there rather than creating the additional UTIL_Describe methods you created that don't do caching?  We should probably meet to discuss rather than go back and forth commenting on the pull request.  I'll schedule a time for us to meet.  thanks.

src/classes/HH_HouseholdNamingSettingValidator_TEST.cls
```diff
@@ -0,0 +1,361 @@
...
+    /*********************************************************************************************************
+    * @description Validator used to validate Household Naming Settings
+    */
+    private static HH_HouseholdNamingSettingValidator validator;
+
+
+    /*********************************************************************************************************
+    * @description Static method to initialize variables before each unit test
+    */
+    static {
```
davidhabib: i'm not familiar with this syntax.  does it really reinitialize before each test, rather than once per the class?  how is it different than using @testSetup? if the same, why use this syntax?

src/classes/HH_HouseholdNaming.cls
RadSFDeveloper: The `isValidField(String objectName, String fieldName)` does not handle relationships. All new methods added in `Util_Describe` are there to resolve any case field name (firstname, FirstName) regardless if it is a field on an SObject or a relationship field.

src/classes/HH_HouseholdNamingSettingValidator_TEST.cls
RadSFDeveloper: I have not seen using @TestSetup anywhere in the code, is there a reason why we have not used it so far?  The static initialization method resets variables before each unit test and it does what is required here. I can use @TestSetup if that is the norm.

judisohn:
@RadSFDeveloper @davidhabib @smarovich I think we should probably say something in Changes, because the text of the Issue may not be enough to explain and perhaps we should point out that the page is now validated against errors? Or edit the Issue title to match what actually was changed?

src/classes/HH_NameSpec_TEST.cls
```diff
@@ -396,6 +402,51 @@ private with sharing class HH_NameSpec_TEST {
     /*********************************************************************************************************
     @description
     operation:
+        Test Household Name, and Formal/Informal greetings when Contact API names are case insensitive
+    verify:
+        HH Name, Formal Greeting, and Informal Greeting are correct
+    **********************************************************************************************************/            
+    static testMethod void testHHNamingWhenContactApiNamesAreCaseInsensitive() {
```
davidhabib: so here's an example of a shorter name that might still work.  since this is in the HH_NameSpec_TEST class, we already know it's about HHNaming, so we don't really need to include "testHHNamingWhen".  I probably would have called this something like testNameFormatsCaseInsensitive()

src/classes/STG_PanelHouseholds_CTRL.cls
```diff
@@ -53,49 +53,66 @@ public with sharing class STG_PanelHouseholds_CTRL extends STG_Panel {
...
+    private boolean isValidHouseholdNamingSettings() {
+        try {
+            HH_HouseholdNamingSettingValidator validator = new HH_HouseholdNamingSettingValidator();
+            HH_HouseholdNamingSettingValidator.Notification notification = validator.validate(STG_Panel.stgService.stgHN);
+
+            if (notification.isSuccess()) {
+                return true;
+            }
+
+            // Display the first error only?
```
davidhabib: given you did the work to implement your notification class that provides all errors, why only display the first?

src/classes/STG_PanelHouseholds_CTRL.cls
```diff
@@ -53,49 +53,66 @@ public with sharing class STG_PanelHouseholds_CTRL extends STG_Panel {
...
+            addPageErrorMessage(notification.getErrors()[0]);
+
+        } catch (Exception ex) {
+            addPageErrorMessage(ex.getMessage());
```
davidhabib: i thought one of the reasons for using your validation service was to avoid exceptions.  why the try/catch?  if callers who want to use the validation service always still need to use a try/catch, then what's the point?

src/classes/STG_SettingsManager_TEST.cls
```diff
@@ -240,39 +240,266 @@ public with sharing class STG_SettingsManager_TEST {
...
     }
 
     /*********************************************************************************************************
-    * @description Tests that the name specs are valid in the Households panel
-    * @return void
-    */
-    public static testMethod void testNameSpecs() {
-        if (strTestOnly != '*' && strTestOnly != 'testNameSpecs') return;
-        STG_PanelHouseholds_CTRL panel = new STG_PanelHouseholds_CTRL();
+    @description
```
davidhabib: given the amount of test you've added just for the household panel, I think it would be better to have them in their own class file.  this would also alleviate the need to prefix each test with STG_Panelhouseholds_

src/classes/HH_HouseholdNamingSettingValidator_TEST.cls
davidhabib: most of our test code was written before @testsetup was introduced by Salesforce, and most of us just follow patterns we've used in the past.  I'm fine with people using testsetup, or init routines, or just initializing static vars in tests.  I'm just not familiar with creating an anonymous static block to init static vars.  Why not just put the initialization on the same line as the declaration, if you don't need a full init routine?

davidhabib:
@RadSFDeveloper so based on our discussion, I assume you will rework your solution to only do the full validation from the settings page.  I assume you will then still need to change how the sets of fields are combined, so case sensitivity isn't an issue, in the several callers who are putting together these SOQL strings.  you could create a trivial helper that lowercased the sets before adding. 

As for the validation routines in UTIL_Describe, I think it would be worth considering whether you can modify isValidField() so that when it encounters a related field, it then does the work to fill the object describe cache for the new object, and then validates against it.  My hope is that the calling code doesn't need to care about this and can just call isValidField() as always, and that UTIL_Describe can be smart enough to only add more describes to the cache on demand, when needed, and use them.  Does that sound like a reasonable strategy?

src/labels/CustomLabels.labels
RadSFDeveloper: @smarovich , please ignore request for the error message review. The message is not required any longer. The changes will be done in the new commit.

src/classes/HH_NameSpec_TEST.cls
RadSFDeveloper: Thanks David. I believe it is very important for the test name to indicate what is being tested and expected. The reasons are mentioned above (readability is the main one), but most importantly tests should be self-documenting, and developer should not guess what is being tested when a test is run from Developer Console or any other app. I will decrease the test names, since they are too long.

src/labels/CustomLabels.labels
smarovich: Hi @RadSFDeveloper, I spoke with @judisohn and we updated the release notes.

src/classes/UTIL_Describe.cls
RadSFDeveloper: New `getFieldResultsByName()` and `getApiName()` methods have been removed from `UTIL_Describe` since they are not used for the Household Naming Settings validation any longer.

src/classes/HH_HouseholdNamingSettingValidator_TEST.cls
RadSFDeveloper: The vars are initialized in the when declared. The `static{}` initialization method have been removed.

src/classes/STG_PanelHouseholds_CTRL.cls
RadSFDeveloper: The validator was created to remove responsibility from the STG_PanelHouseholds_CTRL class to validate settings. It is up to the controller to call the validator to validate and if there a validation error, then display an error, all errors or raise an exception where required. The code was changed, but without impact to user experience since one error is still displayed. However, since validator collects all errors now, I will make the change to display them all.

src/classes/STG_PanelHouseholds_CTRL.cls
RadSFDeveloper: The try/catch has been removed.

src/classes/STG_SettingsManager_TEST.cls
RadSFDeveloper: New STG_PanelHouseholds_TEST has been created.

RadSFDeveloper:
@davidhabib , the re-work of the solution has been done and the code is ready for the review. A Query builder utility has been created to create a SOQL with unique field names.

The new UTIL_Describe methods added by previous commits in this branch have been removed since they are not used any longer. If there is a need to use them, they can be retrieved from this branch commits.

judisohn:
@RadSFDeveloper FYI - I had my branch from when I tested before the recent changes and I just deployed the latest to see how it behaved now. I noticed this right away and not sure if it's an issue just in my scratch org, but thought I should bring to your attention to see if you're seeing the same thing? I hadn't changed any of the formats.

image

RadSFDeveloper:
@judisohn , is the Debug Log Level for ApexCode set to Finest? If so, it is best to change it to Info or Debug log level. Note when changing, ensure there is a red indicator on the ApexCode field before clicking on Save.

judisohn:
@RadSFDeveloper Ah, gotcha. I knew this sounded familiar. This was in a scratch org where I didn't think to check where the Debug Log was checked. Interesting to note it sets that way by default. Thanks.

src/classes/HH_Container_TEST.cls
```diff
@@ -361,9 +452,9 @@ private class HH_Container_TEST {
     verify:
         Name and Formal/Informal Greetings are not modified on the Household
     **********************************************************************************************************/  
-    static testMethod void getHHNamesGreetingsShouldNotModifyNameAndGreetingsOnHouseholdWhenHHNamingSettingsIsInvalidAndAutomaticHouseholdNamingIsTurnedOff() {
-        turnOffAutomaticHHNaming();
-        setupInvalidHHNamingSettings();
+    static testMethod void testHHNameAndGreetingsAreNotModifiedWhenAutomaticHHNamingIsOff() {
```
davidhabib: thank you for the shorter, yet still meaningful, test names!

src/classes/UTIL_Query.cls
```diff
@@ -0,0 +1,214 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Utilities
+* @description SOQL Builder
+*/
+public with sharing class UTIL_Query {
```
davidhabib: great class!  going to be very useful for all of our code!

PR #2703, opened 2017-08-17
Feature/joy error log sched jobs

### LINKED ISSUE #2348: NPSP Error Logs sent to inactive instances
Report of NPSP Error Log sent to Administrator of an organization that was last active in 2014. Customer/Partner has no way to log in and resolve the error.

joy-sfdo:
# Critical Changes

# Changes

# Issues Closed
#2348 - NPSP Error Logs sent to inactive User


src/classes/ERR_Notifier.cls
```diff
@@ -68,13 +70,13 @@ public class ERR_Notifier {
             //We will send email
             } else {
                 List<String> sendList = new List<String>();
-                if (errorNotifRecipient instanceof id && errorNotifRecipient.startsWith(NotificationOptions.user)) {
+                if (errorNotifRecipient instanceof id && errorNotifRecipient.startsWith(NotificationOptions.user) && context != CONTEXT_USER) {
```
davidhabib: why the added contact != CONTEXT_USER?  not sure I understand the scenario.  thanks.

src/classes/USER_FrozenInActiveUser_TDTM.cls
```diff
@@ -0,0 +1,158 @@
...
+* @date 2017
+* @group User
+* @description Trigger handler on User object to handle User status changes.
+*/
+
+public class USER_FrozenInActiveUser_TDTM extends TDTM_Runnable {
+	
+	public USER_FrozenInActiveUser_TDTM() {}
+
+	/*******************************************************************************************************
```
davidhabib: looks like all of your comment header 1st lines start with an extra tab or spaces?

davidhabib: also, your tabstop should be set to 4 spaces, and you need to have your editor save out spaces.  looks like maybe it's saving out as 8 spaces.

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
```diff
@@ -0,0 +1,159 @@
...
+	//================================
+	public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
+	public static final String ERROR_NOTIFICATION_RECIPIENT_ALL_SYS_ADMINS = 'All Sys Admins';
+	
+	
+	/*********************************************************************************************************
+    * @description Tests the handling of the TDTM_Runnable.Action for BeforeInsert 
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testRunOverrideBeforeInser() {
```
davidhabib: while I wouldn't have created the empty stubs in USER_FrozenInActiveUser_TDTM for all the trigger events you weren't handling, I was fine leaving them in.  But writing tests for them which can't actually assert anything seems misleading.  What benefit are these?  They seem a detriment, to me, because they imply that tests are written for the other trigger events, but in fact there is no testing logic here, and there is no valid way to write tests for functionality that doesn't yet exist or has even been defined.  I'd prefer seeing these removed.  thanks.

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
```diff
@@ -0,0 +1,159 @@
...
+        List<UserLogin> currentUserLogin = [SELECT IsFrozen, UserId FROM UserLogin WHERE UserId = :newUsers[1].Id];
+		currentUserLogin[0].IsFrozen = true;
+		Database.update(currentUserLogin, false);
+
+		Test.startTest();
+		userFrozenInActiveUserTDTM.run(newUsers, null, TDTM_Runnable.Action.AfterUpdate, null);
+		Test.stopTest();
+
+		Error_Settings__c errorSettingsSysAdmin = UTIL_CustomSettingsFacade.getErrorSettings();
+        System.assertEquals(ERROR_NOTIFICATION_RECIPIENT_ALL_SYS_ADMINS, errorSettingsSysAdmin.Error_Notifications_To__c);
```
davidhabib: I'm not clear on what this test is testing.  looks like you are just asserting that the setting is set to ALL SYS ADMINS.  but that might have been the default value for that setting before you even did your run() call.  What state did run() change that we can test against to verify it really did change it?

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+	public static Map<Id, Boolean> isSysAdmin(List<User> updatedUsers) {
+		Map<Id, Boolean> userIdToIsSysAdmin = new Map<Id, Boolean>();
+		
+		// Retrieve the SysAdmin Profile Id
+		Id sysAdminId = retrieveSysAdminProfileId();
+
+		// Determine if the User is a SysAdmin
+		if(null != updatedUsers && !updatedUsers.isEmpty()) {
+			for(User currentUser : updatedUsers) {
+				if(sysAdminId == currentUser.ProfileId) {
```
davidhabib: you could code this up more efficiently with 

userIdToIsSysAdmin.put(currentUser.Id, sysAdminId == currentUser.ProfileId);

thus avoid the if/else, and the two separate invocations of put.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+    * @param updatedUsers The users that were modified.
+    * @return A Map containing the UserId and Boolean value based on the User's IsActive status.
+    **********************************************************************************************************/
+	public static Map<Id, Boolean> isActiveUser(List<User> updatedUsers) {
+		Map<Id, Boolean> userIdToIsActive = new Map<Id, Boolean>();
+		
+		// Determine if the User isActive
+		if(null != updatedUsers && !updatedUsers.isEmpty()) {
+			for(User currentUser : updatedUsers) {
+				if(currentUser.IsActive) {
```
davidhabib: i'd just go with userIdToIsActive.put(currentUser.Id, currentUser.IsActive);

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+
+
+	/*********************************************************************************************************
+    * @description Utility method to retrieve the System Adminstrator's ProfileId.
+    * @return Id The Id of the System Administrator Profile.
+    **********************************************************************************************************/
+	private static Id retrieveSysAdminProfileId() {
+		Id sysAdminProfileId = null; 
+
+		List<Profile> sysAdminProfile = [SELECT Id FROM Profile WHERE Name = :SYSTEM_ADMINISTRATOR];
```
davidhabib: what do we do about users who use a cloned profile off of System Adminstrator?  is there any way for us to catch those as well?

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+		return sysAdminProfileId;
+	}
+	
+
+	/*********************************************************************************************************
+    * @description Retrieves the UserLogin records for Users.
+    * @param updatedUsers The users that were modified.
+    * @return A Map containing the UserId and Boolean value based on the User's IsFrozen status.
+    **********************************************************************************************************/
+	private static Map<Id, Boolean> retrieveUserLogins(List<User> updatedUsers) {
```
davidhabib: this method seems misnamed.  it sounds like it will be returning UserLogin records, but in fact it is returning the map of whether each user is frozen or not.  Might be better named retrieveUserFrozenStatus, or something like that.  the fact that the frozen status is derived from the UserLogin record is not really the concern of the caller.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+    **********************************************************************************************************/
+	public static Map<Id, Boolean> isFrozenUser(List<User> updatedUsers) {
+		Map<Id, Boolean> userIdToIsFrozen = new Map<Id, Boolean>();
+
+		// Retrieve the UserLogin records for the updatedUsers
+		if(null != updatedUsers && !updatedUsers.isEmpty()) {
+			Map<Id, Boolean> userLoginUserIdToIsFrozen = retrieveUserLogins(updatedUsers);
+			
+			// Populate the Map with UserId and IsFrozen
+			for(User currentUser : updatedUsers) {
```
davidhabib: since retrieveUserLogins() gives you the actual result map you want, what is the reason for then creating a separate map and copying values to it.  seems inefficient and unnecessary.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+		return userIdToIsFrozen;
+	}
+	
+
+	/*********************************************************************************************************
+    * @description Handles User changes to IsActive/IsFrozen status by updating the Error Notification Recipients, logging an Error__c, and notifying the org's System Administrators.
+    * @param sysAdminUserIds The User Ids that were modified.
+    * @return void
+    **********************************************************************************************************/
+	@future
```
davidhabib: why the need to use @Future?  we have a naming convention that all of our Future methods should end with the work Future, to make it more obvious to callers.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+		// Sets Error Notification recipient to All Sys Admin
+		setErrorNotificationToAllSysAdmins(sysAdminUserIds);
+
+		// Logs an Error and Emails the Sys Admins
+		handleScheduledJobsSubmittedByInActiveIsFrozenSysAdmin(sysAdminUserIds);
+	}
+
+
+	/*********************************************************************************************************
+    * @description Sets the NPSP Settings Error Notification Recipients to All Sys Admins.
```
davidhabib: does it always set the recipients to All Sys Admins, or only if one of the users made inactive/frozen is the current recipient?  I believe the latter, and your description should make this clear.  I'd prefer a method name that also didn't imply the action was always taken.  While reviewing the code in the callers, I thought it was wrong because it didn't check if the current user was the only recipient.  maybe a name like handleErrorRecipientInactiveIsFrozen() or something like that?

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+					Database.update(errorSettings, true);
+					break;
+				}
+			}
+		}
+	}
+
+
+	/*********************************************************************************************************
+    * @description Determines if a User is the owner of the Scheduled Jobs.
```
davidhabib: description should also give a hint as to what it does in this case.

src/classes/USER_FrozenInActiveUser_TDTM.cls
```diff
@@ -0,0 +1,158 @@
...
+		Map<Id, Boolean> userIdToIsActive = USER_UserService.isActiveUser(newlist);
+		Map<Id, Boolean> userIdToIsFrozen = USER_UserService.isFrozenUser(newlist);
+
+		Set<Id> sysAdminUserIds = userIdToIsSysAdmin.keySet();
+		for(Id currentUserId : sysAdminUserIds) {
+			if(userIdToIsSysAdmin.get(currentUserId)      
+				&& (!userIdToIsActive.get(currentUserId)
+				 	|| userIdToIsFrozen.get(currentUserId))) {
+
+				USER_UserService.applySysAdminIsActiveIsFrozenStatusChanges(sysAdminUserIds);
```
davidhabib: shouldn't you just be adding the frozen/inactive/ sysadmins to your Set, and then after your for loop, call your service with the resultant Set?  you are currently calling your service multiple times with an ever growing set.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+		Boolean breakOuterCircle = false;
+
+		List<CronTrigger> scheduledJobs = retrieveScheduledJobs(sysAdminUserIds);
+
+		for(Id currentSysAdminId : sysAdminUserIds) {
+			for(CronTrigger eachScheduledJob : scheduledJobs) {
+				if(currentSysAdminId == eachScheduledJob.OwnerId) {
+					notifySysAdminsAboutInActiveIsFrozenSysAdminUser();
+					breakOuterCircle = true;
+					break;
```
davidhabib: using the boolean to break out of two for loops is a bit non-standard.  if you are really trying to stop processing that method, you should just use a return.  I do wonder, however, whether you should be ensuring to verify all  of the passed in admins, not just stopping at the first?

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+	//================================
+	// Constants
+	//================================
+	public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
+	public static final String OBJECT_ID_PREFIX_USER = '005';
+	public static final String ERROR_NOTIFICATION_RECIPIENT_ALL_SYS_ADMINS = 'All Sys Admins';
+	public static final String JOB_TYPE_SCHEDULED_APEX = '7';
+	public static final String ERROR_TYPE_SCHEDULED_APEX = 'Scheduled Apex Error';
+	public static final String OBJECT_TYPE_USER = 'User';
+	public static final String ERROR_MESSAGE_SCHEDULED_APEX = 'There are scheduled jobs owned by an inactive or frozen user. Review your scheduled jobs and ensure the owner of each one is an active user.';
```
davidhabib: this should be using a custom label, so that it can be translated.  we  avoid hardcoding any user visible string.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+
+	/*********************************************************************************************************
+    * @description Emails the org's System Administrators.
+    * @return void
+    **********************************************************************************************************/
+	@testVisible
+	private static void sendInActiveIsFrozenUserEmail() {
+		ERR_Notifier.sendErrorNotifications(CONTEXT_TYPE_USER);
```
davidhabib: you should be passing in ERR_Handler_API.Context.USER, rather than defining your own string for this enum value.

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,240 @@
...
+* @group User
+* @description Tests for USER_UserService class.
+*/
+
+@isTest
+public with sharing class USER_UserService_TEST {
+	//================================
+	// Constants
+	//================================
+	public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
```
davidhabib: all of these are already defined publicly in the User Service class.  you should not be redefining them again in your test class, but just have the code refer directly to those public strings.

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,240 @@
...
+    * @description Tests if modified Users are System Administrators.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testIsSysAdmin() {
+    	List<User> newUsers = createUsers();
+        
+    	Map<Id, Boolean> userIdToIsSysAdmin = USER_UserService.isSysAdmin(newUsers);
+    	
+    	System.assertEquals(true, userIdToIsSysAdmin.get(newUsers[0].Id));
+        System.assertEquals(true, userIdToIsSysAdmin.get(newUsers[1].Id));
```
davidhabib: indented wrong. extra space?

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,240 @@
...
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testIsActiveUser() {
+        List<User> newUsers = createUsers();
+        
+    	Map<Id, Boolean> userIdToIsActive = USER_UserService.isActiveUser(newUsers);
+
+    	System.assertEquals(true, userIdToIsActive.get(newUsers[0].Id));
+        System.assertEquals(false, userIdToIsActive.get(newUsers[1].Id));
+    }
```
davidhabib: indenting wrong

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,240 @@
...
+   	/*********************************************************************************************************
+    * @description Tests if User is the owner of Scheduled Jobs and calls the methods that logs an error and sends an email.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testHandleScheduledJobsSubmittedByInActiveIsFrozenSysAdmin() {
+    	List<User> newUsers = createUsers();
+    	Set<Id> sysAdminIds = new Map<Id, User>(newUsers).keySet();
+
+    	Test.startTest();
+    	UTIL_MasterSchedulableHelper.setScheduledJobs();
```
davidhabib: won't this logic create the scheduled jobs using the current running user, rather than one of the users you have created.  thus running your service shouldn't affect the jobs.  seems like you need two tests.  one that verifies that deactivating a user who isn't the owner of the jobs has no affect on the jobs (userId better be what it originally was), and another test where you schedule the jobs as a user who you the freeze, and then you verify the jobs gets cleared?  resaved with another user?

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
davidhabib: In general, we've tested our TDTM logic by always doing the actual DML that should fire the TDTM logic.  by invoking the run method directly, you are not 100% verifying that a given DML operation has the intended effect that we expect.

joy-sfdo:
Hi @davidhabib, 

Thanks for the code review! I'll address the items and reach out if I have any questions.

Best,
Joy

src/classes/ERR_Notifier.cls
joy-sfdo: The context check was used to support the async handling of the Error Notification Settings. In the earlier version of UserService (prior to check in), it was possible that the settings were still set to the InActive/Frozen User. However, this check is currently no longer necessary so I've removed it. Thanks!

src/classes/USER_FrozenInActiveUser_TDTM.cls
joy-sfdo: Updated the editor to set tabs to 4 spaces.

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
joy-sfdo: I removed the tests for the empty stubs. 

Also, I'm open to removing the empty stubs. I included them as part of the design structure. To keep the structure we can replace the stubs with comments. Thoughts?

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
joy-sfdo: Updated the test to confirm the change of the Error Notification Recipient. 

Also, updating the User to call/test the TDTM logic.

src/classes/USER_UserService.cls
joy-sfdo: Implemented this code change. Thanks!

src/classes/USER_UserService.cls
joy-sfdo: Implemented this code change. Thanks!

src/classes/USER_UserService.cls
joy-sfdo: Presently the code looks for the System Administrator Profile when handling the Error Notification Recipient setting. We could add the functionality to handle cloned profiles. If we do, I'd recommend making the change throughout the code.

src/classes/USER_UserService.cls
joy-sfdo: Updated this method to reflect its original intention. 

Good catch!

src/classes/USER_UserService.cls
joy-sfdo: Updated this method based on changes to the retrieveUserLogin method.

src/classes/USER_UserService.cls
joy-sfdo: The @future is used to handle the MIXED_DML_OPERATION that occurs when updating the User's isActive field along with other dml operations.

src/classes/USER_UserService.cls
joy-sfdo: Updated the method's name.

src/classes/USER_UserService.cls
joy-sfdo: Updated the description. Thanks!

src/classes/USER_FrozenInActiveUser_TDTM.cls
joy-sfdo: Updated the code accordingly. Good catch!

src/classes/USER_UserService.cls
joy-sfdo: I agree about the usage of the boolean. I hesitated about using return since I was in a void method. However, I like using return in this case and have updated the code.

src/classes/USER_UserService.cls
joy-sfdo: Added the Custom Label.

src/classes/USER_UserService.cls
joy-sfdo: Using the ERR_Handler_API.Context.USER. Thanks!

src/classes/USER_UserService_TEST.cls
joy-sfdo: Using the constants from the UserService class.

src/classes/USER_UserService_TEST.cls
joy-sfdo: Updated the editor to set tabs to 4 spaces.

src/classes/USER_UserService_TEST.cls
joy-sfdo: Updated the editor to set tabs to 4 spaces.

src/classes/USER_UserService_TEST.cls
joy-sfdo: Updated the test to create the jobs as the User being modified. Thanks!

src/classes/USER_UserService.cls
davidhabib: is there even a foolproof way to know  it was cloned from Sys Admin?

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+		System.assertEquals(newUsers[0].Id, errorSettings.Error_Notifications_To__c);
+
+		// Freeze a User
+		List<UserLogin> currentUserLogin = [SELECT Id, IsFrozen, UserId FROM UserLogin WHERE UserId = :newUsers[0].Id];
+		currentUserLogin[0].IsFrozen = true;
+
+		Test.startTest();
+		Database.update(currentUserLogin, false);
+		Database.update(newUsers[0], true);
+		Test.stopTest();
```
davidhabib: I'm confused why you are updating currentUserLogin, as well as newUsers[0].  shouldn't it just be the first one?

src/classes/USER_UserService.cls
davidhabib: you should remove ERROR_MESSAGE_SCHEDULED_APEX now that you have the custom label to use.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,253 @@
...
+	public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
+	public static final String OBJECT_ID_PREFIX_USER = '005';
+	public static final String ERROR_NOTIFICATION_RECIPIENT_ALL_SYS_ADMINS = 'All Sys Admins';
+	public static final String JOB_TYPE_SCHEDULED_APEX = '7';
+	public static final String ERROR_TYPE_SCHEDULED_APEX = 'Scheduled Apex Error';
+	public static final String OBJECT_TYPE_USER = 'User';
+	public static final String ERROR_MESSAGE_SCHEDULED_APEX = 'There are scheduled jobs owned by an inactive or frozen user. Review your scheduled jobs and ensure the owner of each one is an active user.';
+	
+
+	/*********************************************************************************************************
```
davidhabib: this class file is still saving tabs, not 4 spaces.  please make sure all of the files in this pull request save 4 spaces, not tabs. thanks.

src/classes/USER_UserService.cls
davidhabib: could you put that info in the description, as it is very helpful to know why a future needed to be used.  thanks.

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,247 @@
...
+    * @description Tests logging and Error__c and sending and email to System Administrators.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testNotifySysAdminsAboutInActiveIsFrozenSysAdminUser() {
+        Test.startTest();
+        USER_UserService.notifySysAdminsAboutInActiveIsFrozenSysAdminUser();
+        Test.stopTest();
+
+        List<Error__c> retrievedErrorTypeAfterEmailSent = [SELECT Email_Sent__c FROM Error__c WHERE Object_Type__c = :USER_UserService.OBJECT_TYPE_USER];
+        System.assertNotEquals(null, retrievedErrorTypeAfterEmailSent);
```
davidhabib: this assert just says the soql returned a list, which it always will.  it doesn't assert that the list isn't empty.

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,247 @@
...
+        List<Error__c> retrievedErrorTypeAfterEmailSent = [SELECT Email_Sent__c FROM Error__c WHERE Object_Type__c = :USER_UserService.OBJECT_TYPE_USER];
+        System.assertEquals(true, retrievedErrorTypeAfterEmailSent[0].Email_Sent__c);
+    }
+
+
+    /*********************************************************************************************************
+    * @description Utility method to create Users with System Administrator profile and sets their IsActive status accordingly.
+    * @return List containing the newly created Users.
+    **********************************************************************************************************/
+    private static List<User> createUsers() {
```
davidhabib: rather than duplicate this helper method, why not make the one in USER_FrozenInActiveUser_TDTM_TEST public and use it, or vice-versa.

src/classes/USER_FrozenInActiveUser_TDTM.cls
```diff
@@ -0,0 +1,160 @@
...
+            onAfterUpdate(newlist);
+
+        } else if(TDTM_Runnable.Action.AfterDelete == triggerAction) {
+            onAfterDelete();
+
+        } else if(TDTM_Runnable.Action.AfterUndelete == triggerAction) {
+            onAfterUnDelete();
+
+        }
+
```
davidhabib: yes, I'd prefer this class to only refer to and implement the triggerAction for onAfterUpdate().  thanks.

src/classes/USER_FrozenInActiveUser_TDTM_TEST.cls
joy-sfdo: The update to currentUserLogin is to freeze the User by setting the IsFrozen flag. The update to the User is to test the User trigger.

src/classes/USER_UserService.cls
joy-sfdo: My apologies. I've checked all the files in this PR to save 4 spaces. Thanks.

src/classes/USER_UserService_TEST.cls
joy-sfdo: This assert was to minimize repetition of the checks. However, this method needs to implement the checks in the testLogInActiveIsFrozenUserError and testSendInActiveFrozenUserEmail test methods.

src/classes/USER_UserService_TEST.cls
joy-sfdo: Didn't know if we wanted to keep a clear delineation between data creation for test classes. However, making the createUsers method public is a good because we can later move it and other setup code into a class that handles test data setup like the UTIL_UnitTestData_TEST. I didn't do this initially to keep the User functionality encapsulated.

src/classes/USER_FrozenInActiveUser_TDTM.cls
joy-sfdo: Sounds good. Thanks.

src/classes/USER_UserService.cls
```diff
@@ -0,0 +1,214 @@
...
+* @description Service Class to handle actions on the User object.
+*/
+public with sharing class USER_UserService {
+    public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
+    public static final String OBJECT_ID_PREFIX_USER = '005';
+    public static final String ERROR_NOTIFICATION_RECIPIENT_ALL_SYS_ADMINS = 'All Sys Admins';
+    public static final String JOB_TYPE_SCHEDULED_APEX = '7';
+    public static final String ERROR_TYPE_SCHEDULED_APEX = 'Scheduled Apex Error';
+    public static final String OBJECT_TYPE_USER = 'User';
+    public static final Integer NUMBER_OF_SCHEDULED_JOBS = 9;
```
davidhabib: i'm uncomfortable having this constant in the code.  you are only using it in a test, and the reality is that we periodically add jobs to our scheduled jobs table, and I'd hate for this to get out of sync.  I'd suggest we remove this, and then figure out what the test can assert.  thanks.

src/classes/USER_UserService_TEST.cls
```diff
@@ -0,0 +1,223 @@
...
+        List<CronTrigger> scheduledJobs = [SELECT Id, OwnerId, CronExpression, CronJobDetail.Name, CreatedBy.Name, CreatedBy.IsActive 
+                                            FROM CronTrigger
+                                            WHERE CronJobDetail.JobType = :USER_UserService.JOB_TYPE_SCHEDULED_APEX];
+
+        Test.startTest();
+        USER_UserService.handleScheduledJobsSubmittedByInActiveSysAdmin(sysAdminIds);
+        Test.stopTest();
+
+        if (!scheduledJobs.isEmpty()) {
+            System.assertEquals(USER_UserService.NUMBER_OF_SCHEDULED_JOBS, scheduledJobs.size());
```
davidhabib: as mentioned earlier, no real value in this assert, and it introduces a dependency on keeping this constant in sync with the actual scheduled job table, that we periodically add to.

davidhabib:
@joy-sfdo Hi Joy. real close now.  let's just remove the constant about the size of our scheduled job table.  thanks.

src/classes/USER_UserService_TEST.cls
joy-sfdo: Sounds good, I'll remove this assert.

src/classes/USER_UserService.cls
joy-sfdo: @davidhabib -- Thanks for the information about scheduled jobs. I agree, if we may periodically add jobs, then this constant would be out of sync. I'll remove it.

joy-sfdo:
@davidhabib -- Hi Dave, the requested code updates are available.

Thanks again for the code review!

PR #2710, opened 2017-08-25
Link the Payment currency code to the parent Opportunity

### LINKED ISSUE #2709: Changing CurrencyIsoCode on Campaign fails when there are attached Allocation records.

force2b:
Also prevents the Payment.Currency from being updated by the User directly and fixes a bug in the related logic for Allocations.

# Critical Changes

# Changes

# Issues Closed

Fixes #2709


src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +104,61 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Opportunity 
     * @return dmlWrapper. 
     ********************************************************************************************************/
-    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist, 
-    TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+    public static DmlWrapper runForOpportunities(List<Opportunity> newlist, List<Opportunity> oldlist,
+            TDTM_Runnable.Action triggerAction) {
 
         map<Id, Opportunity> mapIdOppNew;
```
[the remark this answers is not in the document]
force2b: It's a good change to make, though I typically don't go into refactoring older code that hasn't been modified unless there's something inherently wrong or where a fix is really needed for efficiency, bulk, etc.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +104,61 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+        DmlWrapper dmlWrapper = new DmlWrapper();
+
+        // Create new Payments as necessary
+        PMT_PaymentCreator pc = new PMT_PaymentCreator(mapIdOppNew, mapIdOppOld, triggerAction, dmlWrapper);
+
+        //if currency has changed, reset payment currencies
+        list<Opportunity> listOppsForProcessing = new list<Opportunity>();
+        for (integer i=0; i<newlist.size(); i++) {
+            Opportunity opp = newlist[i];
+            if (triggerAction == TDTM_Runnable.Action.AfterUpdate && UTIL_Currency.getInstance().isMultiCurrencyOrganization()) {
```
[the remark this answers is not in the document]
force2b: Good suggestion.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +104,61 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+
         return dmlWrapper;
     }
+
+    /*******************************************************************************************************
+    * @description Updates an opportunity's payment record when currency on the Opportunity is changed.
+    * @param listOpps A list of opportunities whose currency has changed.
+    * @return void This method modifies the dmlWrapper var and does not return anything
+    ********************************************************************************************************/
+    private static void opportunityAmountCurrencyChange(list<Opportunity> listOpps, DmlWrapper dmlWrapper){
```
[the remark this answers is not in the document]
force2b: Good catch. I had copied the method from the Allocations TDTM class, but never changed the method name.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+* @author Salesforce.org
+* @date 2017
+* @group Payments
+* @description tests for the PMT_Payment_TDTM class
+*/
+@isTest
+private class PMT_Payment_TEST {
+
+    /** @description Non default currency type sObject for this org.*/
+    public static sObject nonDefaultCurrencyType;
```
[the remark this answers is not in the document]
force2b: Good catch. Copied logic from another uni test.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+
+    static testMethod void test_MultiCurrency() {
+        //only run test in multicurrency orgs
+        if (!userInfo.isMultiCurrencyOrganization()) {
+            return;
+        }
+
+        setupISOCodes();
+
+        npe01__Contacts_And_Orgs_Settings__c PaymentsSettingsForTests = UTIL_CustomSettingsFacade.getContactsSettingsForTests(
```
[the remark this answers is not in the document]
force2b: Good point. This is done in so many unit tests, but I hadn't actually noticed that the var is in ever referenced in any of those. I'll have to remember this when I'm working in other unit test classes.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+
+        // Verify that individual changes to the Payment CurrencyIsoCode are blocked
+        pmt.put('CurrencyIsoCode', (String)nonDefaultCurrencyType.get('IsoCode'));
+        try {
+            update pmt;
+            System.Assert(False, 'Updating the Payment.CurrencyIsoCode directly should have failed');
+        } catch (Exception ex) {
+            // GOOD!
```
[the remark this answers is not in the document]
force2b: Good suggestion. Worth verifying the message

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -170,9 +170,12 @@ public with sharing class STG_InstallScript_TEST {
...
         List<Trigger_Handler__c> afterScriptConfig = [select Active__c, Class__c from Trigger_Handler__c where Class__c IN :tdtmClasses];     
-        //The -1 is because there are two NPSP flags (npe01__DISABLE_OpportunityAccount_trigger__c and 
-        //npe01__DISABLE_OpportunityContactRoles_trigger__c) that point to the same TDTM class (OPP_OpportunityContactRoles_TDTM)
-        System.assertEquals(tdtmClassesWithValues.size() - 1, afterScriptConfig.size());     
+        // This used to -1 because there are two NPSP flags (npe01__DISABLE_OpportunityAccount_trigger__c and
+        // npe01__DISABLE_OpportunityContactRoles_trigger__c) that point to the same TDTM class
+        // (OPP_OpportunityContactRoles_TDTM. However, a new trigger on the Payment object was added using the
+        // same PMT_Payment_TDTM class referenced in the above list of getNpspToCumulusMap(). That nets out the
+        // -1 to 0.
+        System.assertEquals(tdtmClassesWithValues.size(), afterScriptConfig.size());
```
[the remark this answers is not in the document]
force2b: There was a comment here already that noted that the number of returned records should be `tdtmClassesWithValues.size()-1`. The new trigger handler record added in this story changed. I wanted to explain why the it looks the way it does now, without losing the historical context. It's possible that at `-1` might be needed again at some point if another trigger handler is added that changes the return count. If that happens, I thought the extra comment might help the dev know that adding `-1` would solve the issue.

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -170,9 +170,12 @@ public with sharing class STG_InstallScript_TEST {
...
         for(Trigger_Handler__c handler : afterScriptConfig) {
             UTIL_Debug.debug('****Is class ' + handler.Class__c + ' still active: ' + handler.Active__c);
```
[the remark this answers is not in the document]
force2b: Sure. The debug statement was probably added for some reason, and I agree that it's probably not needed any more. I think this goes back into how far do we refactor existing code

src/classes/PMT_Payment_TDTM.cls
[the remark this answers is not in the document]
RadSFDeveloper: @force2b , can we reuse the method from the Allocations TDTM class?

src/classes/STG_InstallScript_TEST.cls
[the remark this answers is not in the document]
RadSFDeveloper: I am wondering if using the `Set` to store unique classes will remove the need for the comment and replace the code lines 162-168 altogether?

    Set<String> uniqueTdtmClasses = new Set<String>(
        TDTM_DefaultConfig.getNpspToCumulusMap().values()
    );

src/classes/PMT_Payment_TDTM.cls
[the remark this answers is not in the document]
force2b: No. That method does more than just update the CurrencyIsoCode.

src/classes/STG_InstallScript_TEST.cls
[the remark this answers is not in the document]
force2b: Possibly, though we'd need to do the same with the list of `TriggerHandler__c` records returned since those could also include more than one record per Class Name. Would definitely make for a cleaner test.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -29,13 +29,72 @@
...
-* @date 2014 
+* @date 2017
 * @group Payments
 * @group-content ../../ApexDocContent/Payments.htm
-* @description trigger handler on Opps for supporting Payments 
+* @description trigger handler on Payment and Opportunity objects for supporting Payments
 */
 public class PMT_Payment_TDTM extends TDTM_Runnable {
 
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
```
davidhabib: please add an apexdoc header comment block.

force2b: good catch!

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -29,13 +29,72 @@
...
+
+        return dmlWrapper;
+    }
+
+    /*******************************************************************************************************
+    * @description Trigger Handler on Payment for managing multi-currency related items
+    * @param newlist the list of Payments from trigger new.
+    * @param oldlist the list of Payments from trigger old.
+    * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.).
+    * @param objResult the describe for Payment (none currently)
```
davidhabib: remove this param comment since there is no parameter for this.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +103,51 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Opportunity
```
davidhabib: remove the objResult comment since it isn't part of the method signature

force2b: Got this on the last commit.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +103,51 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+
+        Map<Id, Opportunity> mapIdOppNew = (newlist != null ? new Map<Id, Opportunity>(newlist) : new Map<Id, Opportunity>());
+        Map<Id, Opportunity> mapIdOppOld = (oldlist != null ? new Map<Id, Opportunity>(oldlist) : new Map<Id, Opportunity>());
+
+        DmlWrapper dmlWrapper = new DmlWrapper();
+
+        // Create new Payments as necessary
+        PMT_PaymentCreator pc = new PMT_PaymentCreator(mapIdOppNew, mapIdOppOld, triggerAction, dmlWrapper);
+
+        //if currency has changed, reset payment currencies
```
davidhabib: is this logic that really should be in PMT_PaymentCreator() instead?

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -44,18 +103,51 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+        // Get a list of the Opportunity Id's
+        Set<Id> oppIds = (new Map<Id, Opportunity>(listOpps)).keySet();
+
+        // Get a list of the Payment records by Opportunity and update the CurrencyIsoCode value
+        String soql = 'SELECT Id, CurrencyIsoCode, npe01__Opportunity__r.CurrencyIsoCode FROM npe01__OppPayment__c ' +
+            'WHERE npe01__Opportunity__c IN :oppIds ORDER BY npe01__Opportunity__c';
+        for (npe01__OppPayment__c pmt : database.query(soql)) {
+            Opportunity opp = (Opportunity)pmt.getSObject('npe01__Opportunity__r');
+            pmt.put('CurrencyIsoCode', opp.get('CurrencyIsoCode'));
+            dmlWrapper.objectsToUpdate.add(pmt);
```
davidhabib: are there scenarios where the payments will already have the correct isocode, and we should only update them if they need it?  just trying to keep down unnecessary dml if possible.

src/classes/PMT_Payment_TDTM.cls
force2b: It's legacy logic (and method), but in general I think so. That class handles creating and updating payment records as necessary from the Opportunity.

src/classes/PMT_Payment_TDTM.cls
force2b: That probably shouldn't happen. This method would only be called if the Opportunity CurrencyIsoCode gets changed, so if the values are kept in sync, and the Opportunity value is changed, then the Payment would not have the right currency until this runs. Basically every record _should_ be updated.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -109,6 +109,21 @@ private class PMT_Payment_TEST {
         update opp;
         pmt = getPaymentRecord(opp.Id);
         System.AssertEquals((String)opp.get('CurrencyIsoCode'), (String)pmt.get('CurrencyIsoCode'));
+
+        // Verify that a new Payment cannot be inserted with a different currencyIsoCode
```
RadSFDeveloper: In order to create clean tests, it is advisable to have one concept tested in one unit test. For this new code, can we create a new test method with name clearly indicating what is being tested and expected?

force2b: I actually prefer testing one specific set of functionality in the same test. In this case, the test is specifically validating functionality around the CurrencyIsoCode on the Payment object. I think it makes sense to keep these together in a single test.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * @description Validate that an error is thrown when the CurrencyIsoCode is modified on Payment record directly.
+     * Also validate that a change to the CurrencyIsoCode on the Opportunity record is copied down the Payment records.
+     */
+    static testMethod void test_MultiCurrency() {
```
RadSFDeveloper: Can we rename this test method to indicate what is being tested and expected? For example: `testPaymentCcyIsInSyncWithOppCcyInMultiCurrencyOrg`?

force2b: Yes. I'd much prefer to go with the more specific name than split out the test into multiple tests.

src/classes/PMT_Payment_TEST.cls
RadSFDeveloper: We don't want long test functions that go testing one thing after another. The test is verifying if a new Payment with currency different than a Opprtunity currency will generate an error. The previous code in the same test verifies the Opportunity currency change is updating Payment currency. So if we would create two different methods (preferred, right now we have the "if" control statements to separate the logic) in a class to handle this behaviour, why not create two test methods to test different scenarios?

Moreover, if the first part of the test fails, we'll not know if the second part (related to a different functionality) fails or passes, so we might have two issues but we would not be aware of it until the first one is fixed. If we would not write a long method to verify more than one thing, why would we write a test method to verify more than one thing?

force2b: Well stated. My past experience with unit tests has generally driven me towards consolidating tests for with related functionality as long as the decision branches are the same. In this case you're absolutely right. The first part of the test does test Opportunity dml while the remaining tests focus on Payment dml. Your second point is just as critical though. A failure in the first part would mask a failure in the second part - at least until the first part is fixed. I think that is acceptable in some situations (such as this one where the tests are extremely small), however there's absolutely value in seeing each individual failure to know that something is not working. If debugging those, it's also satisfying to fix one thing and see that it fixes multiple tests. I'll rework the test class. :)

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -29,33 +29,136 @@
...
+            // Prevent the Payment CurrencyCode from being modified directly
+            if (triggerAction == TDTM_Runnable.Action.BeforeUpdate && pmtCurrency != oldPmtCurrency && pmtCurrency != oppCurrency) {
+                pmt.addError(Label.pmtModifyCurrency);
+
+            } else if (triggerAction == TDTM_Runnable.Action.BeforeInsert && pmtCurrency != oppCurrency && pmtCurrency != null) {
+                // If the User is attempting to set the Payment Currency to something different than the Opp currency
+                pmt.addError(Label.pmtModifyCurrency);
+
+            } else if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
+                // Otherwise efault the Payment CurrencyIsoCode to the value from the parent Opportunity
```
RadSFDeveloper: Typo in the comment

src/classes/PMT_Payment_TEST.cls
```diff
@@ -56,44 +56,124 @@ private class PMT_Payment_TEST {
...
+
+        Opportunity opp = new Opportunity(
+                Name = 'success',
+                Amount = 10,
+                AccountID = acc.id,
+                CloseDate = system.today(),
+                StageName = UTIL_UnitTestData_TEST.getClosedWonStage());
+        insert opp;
+
+        npe01__OppPayment__c pmt = getPaymentRecord(opp.Id);
```
RadSFDeveloper: Do we need to verify the Payment currency after retrieving the Payment on this line?

force2b: No. That functionality is verified in the previous test.

RadSFDeveloper: So we do not retrieve the Payment here if we are not going to check it or use it before line 142?

force2b: Yeah, true. I left the the retrieve up there when technically it's not needed until a few lines down. I will fix.

PR #2718, opened 2017-08-29
remove dependency on sf_chttr_apps in metadata.

davidhabib:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/components/RP_GettingStarted.component-meta.xml
+++ b/src/components/RP_GettingStarted.component-meta.xml
@@ -27,9 +27,4 @@
         <minorNumber>9</minorNumber>
         <namespace>npo02</namespace>
     </packageVersions>
-    <packageVersions>
-        <majorNumber>1</majorNumber>
-        <minorNumber>14</minorNumber>
-        <namespace>sf_chttr_apps</namespace>
-    </packageVersions>
 </ApexComponent>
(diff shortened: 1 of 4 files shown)


davidhabib:
@cpurodriguez I found that the 3 visualforce components and the page component had 

    <packageVersions>
        <majorNumber>1</majorNumber>
        <minorNumber>14</minorNumber>
        <namespace>sf_chttr_apps</namespace>
    </packageVersions>

in their metadata file, and it would not deploy in one of my DE orgs.  Not sure what this package dependency is from, but I don't believe it is required for the getting started page.  So I will be removing this dependency once the build passes.

any idea what this came from?  thanks.

cpurodriguez:
@davidhabib it is not required. Not sure where it came from but I will review. Maybe a default meta because of being part of a dev org?

davidhabib:
one of our team members says it's from the connected app for chatter desktop.  I guess scratch orgs must be getting it enabled by default, which is why our build didn't catch it.

cpurodriguez:
@davidhabib good catch.

davidhabib:
@cpolcyn @jlantz removing the metadata dependency has been removed and passes build.  can one of you please merge this asap (so we can then move onto other pull requests).  thanks!

PR #2719, opened 2017-08-29
fix invalid assert in getting started page tests

davidhabib:
@cpurodriguez one of our test orgs failed this test, and it turns out the test assert is invalid.  it assumes the datetime formatting will map to Pacific time with locale US.  but if the user running the test has a different timezone or locale, it will fail.  I couldn't come up with a quick fix to do the correct assert.  I'm hoping your team can address this quickly and update this pull request with the correct assert.  if not, I'll need to remove the assert, so we don't block any builds or customers.  but we will still need a real fix later.  thanks.


szeitz:
Tagging @mdeleon0 and @Sonhia19 in case they are able to assist.

Issue #2700, opened 2017-08-15
BDI: custom Contact and Account Unique ID's don't support fields from other packages


davidhabib:
the optional Contact Custom Unique ID and Account Custom Unique ID settings allow the user to specify their own custom field for doing matching with.  These fields currently are assumed to always be in the org's non-namespaced code.  We don't support the fields living in another managed package, as well as a mixture of the Contact or Account field coming from another managed package, but the Data Import field being added by the customer without a namespace.  

Note that no customer has yet asked for this support, but it is a limitation of the current code and settings.

davidhabib:
**lurch: add

judisohn:
@davidhabib good one. I'm frankly surprised this hasn't come up already... in the meantime, should anyone review this, a workaround would be a workflow rule or process that copies the value to a custom field?

davidhabib:
that might work, but having workflows run during data import is probably not a good practice.

judisohn:
I was thinking the workflow rule would fire when the external ID was added by the app, which is not necessarily in bulk.

mrbelvedere:
Included in beta release 3.110 (Beta 7)

cpolcyn:
Included in production release rel/3.110

Agroag:
I can no longer see fields we were using for custom mapping - if there is another issue or article that could help that would be great.  I'm not sure how to get fields we were using for custom account and contact ID matching to show up again.  They were there and now they're not.  What could have changed to make them not available? Thanks!

mpusto:
https://github.com/SalesforceFoundation/Cumulus/releases/latest says "The Batch Data Import Configuration Options for Contact Custom Unique ID and Account Custom Unique ID now only display fields that are marked Unique or External ID. Additionally, those fields can now come from other managed packages."  That means the field needs to be marked as Unique in the custom field set up. This prevents errors for scenarios where two records could have the same ID (if not marked unique) and then data could be overwritten.

Agroag:
Hi - Thanks for the quick response.  We were using a formula field to match.  Do we need to create a new external ID field and a workflow to populate it with the IDs from our custom ID field? Do you have any recommendations?

mpusto:
Yes that would be the best way. It can be either an external Id field or simply marked as "unique"

PR #2721, opened 2017-08-30
BDI Enhancements

davidhabib:
# Critical Changes

## Batch Data Import Enhancements

To support the new Batch Data Import feature, you need to perform some setup tasks. In the profiles of all users who perform data imports using the NPSP Data Importer:

- Set the NPSP Data Import Batch tab setting to Default On.
- Set object permission for the NPSP Data Import Batch object to Read, Create, Edit.
- In the NPSP Data Import object settings, set field permissions for the NPSP Data Import Batch field to Read Access and Edit Access.
- Override the Clone, Edit, and New buttons on the NPSP Data Import Batch object to use the BDI_BatchOverride Visualforce page.

 
# Changes

## Batch Data Import Enhancements

- The NPSP Data Importer now includes support for processing your NPSP data imports in batches, using a specific set of configuration settings. Use the new NPSP Data Import Batch object to create a batch and specify its unique configuration settings.
- You can also now schedule NPSP Data Import Batches to be processed automatically using a scheduled job. You can find the new scheduled job, **NPSP 09 - Data Import Batch Processing**, in Setup under Scheduled Jobs. NPSP will automatically schedule the new NPSP 09 - Data Import Batch Processing job, but the default "Nonprofit Success Pack" user has limited permissions. To ensure smooth functioning of scheduled imports, the job should be scheduled by a System Administrator. A System Administrator can automatically reschedule the job by simply visiting the NPSP Settings page.
- The NPSP Data Importer now supports invoking a custom Post Process Apex class to perform additional tasks as each set of Data Import records is processed. This feature requires working with a developer to write an Apex class that implements the new BDI_IPostProcess interface and its method. You then specify the Apex class in the **Post Process Implementing Class** field in either the NPSP Data Import configuration options or in the NPSP Data Import Batch configuration options.
- For complete documentation, see the NPSP Advanced User's Guide to Importing Data (https://powerofus.force.com/articles/Resource/NPSP-Advanced-User-s-Guide-to-Importing-Data).

# Issues Closed


davidhabib:
@smarovich Can you please review and update this pull request comment to capture all of the new features in BDI that we want to hilight, as well as pointing people at your new advanced guide.  @judisohn did i forget any of the new features worth highlighting?  thanks both of you!

judisohn:
@smarovich @davidhabib This covers the main features! We may want to include similar language for the post processing as we did for the custom household naming to guide developers in the right direction in release notes. Should we also have language advising the enabling of the new object/new fields for push upgrades?

davidhabib:
@judisohn @smarovich I've updated the critical changes comments to include the items that need enabling.  please cleanup to use appropriate language as you see fit.  @njjc I've added a new night batch job.  Do you recall if our upgrade script will add new jobs, or whether we need to tell people to open the NPSP Settings page to cause the jobs to get recreated?

njjc:
We have a bit of a two-pronged approach, we do schedule it in the post install script, but it's scheduled by the `Nonprofit Success Pack` user which as limited permissions and can hit some errors. Then, the first time a SysAdmin accesses the Settings page, it will reschedule the job with that user. I don't think we need to mention it in the release notes.

judisohn:
@davidhabib @njjc The doc @smarovich talks about looking for and verifying that scheduled job. I think we can leave it at that for now. On the Hub post, we should be sure to link to the new doc.

src/classes/BDI_SettingsUI_CTRL.cls
```diff
@@ -133,45 +125,36 @@ public with sharing class BDI_SettingsUI_CTRL {
...
-    * @description saves the provided Settings.  since we want this operation controlled by the containing
-    * page, it is static to be callable by the controlling page.  it must pass the settings instance however.
-    * @param diSettings The BDI settings to save
-    * @return True if success, False if error.
+    * @description creates a list of SelectOptions of potential Unique ID fields from the specified object
+    * @param strObject the object to query fields on
+    * @return list<SelectOption>
     */
-    public static boolean saveBDISettings(Data_Import_Settings__c diSettings) {
+    private list<SelectOption> listSOForObject(String strObject) {
```
njjc: I take it you're not using dfr.isExternalID() in case they're using a field that's not marked as external ID?

I'm also wondering if it's possible that a unique ID field could be packaged at some point, are you sure showing only unpackaged fields makes sense?

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -71,7 +71,8 @@ public without sharing class UTIL_MasterSchedulableHelper {
         'NPSP 05 - GAU Allocation Rollups'          => new List<String>{'0 0 23 ? * *', 'ALLO_Rollup_SCHED'},
         'NPSP 06 - Recurring Donation Updates'      => new List<String>{'0 0 23 ? * *', 'RD_RecurringDonations_BATCH'},
         'NPSP 07 - Seasonal Address Updates'        => new List<String>{'0 0 23 ? * *', 'ADDR_Seasonal_SCHED'},
-        'NPSP 08 - Level Assignment Updates'        => new List<String>{'0 0 4 ? * *',  'LVL_LevelAssign_SCHED'}
+        'NPSP 08 - Level Assignment Updates'        => new List<String>{'0 0 4 ? * *',  'LVL_LevelAssign_SCHED'},
+        'NPSP 09 - Data Import Batch Processing'    => new List<String>{'0 0 22 ? * *', 'BDI_DataImportBatch_SCHED'}
```
njjc: I think it might be worth testing what happens when trying to run this with the `Nonprofit Success Pack` packaging user with limited access. If it borks, which seems likely, it makes sense to update the docs to suggest visiting the NPSP settings page to get it rescheduled. What do you think?

src/classes/BDI_BatchOverride_CTRL.cls
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /*********************************************************************************************************
+    * @description actionMethod to Save the current Batch.
+    */
+    public PageReference save() {
+        try {
+            BDI_DataImportService.validateSettings(diBatch);
+            upsert diBatch;
+            return new PageReference('/' + diBatch.Id);
```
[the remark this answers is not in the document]
njjc: This is definitely broken in LEX. When clicking Save, I get a blank page with this popup:
screen shot 2017-09-07 at 10 34 31 am
Both options are broken.
- If I click Cancel, it sends me back to edit the UI with everything populated, but the Batch object has been created. 
- If I click Open, it opens a new tab on the newly created Batch, but the old tab still has the Edit UI with everything I filled out populated.
In both cases, I can then click Save again and it creates a duplicate Batch.

src/layouts/DataImportBatch__c-NPSP Data Import Batch Layout.layout
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
njjc: The default button options on the page layout are weird. Any idea if we can update the defaults to show Edit, Clone, and Process Batch instead of all those weird New (Object) buttons?
screen shot 2017-09-07 at 10 38 06 am

src/objects/DataImportBatch__c.object
```diff
@@ -173,28 +190,123 @@
...
+        <availability>online</availability>
+        <description>invoke the Data Importer on all records associated with this Batch</description>
+        <displayType>button</displayType>
+        <encodingKey>UTF-8</encodingKey>
+        <height>600</height>
+        <linkType>url</linkType>
+        <masterLabel>Process Batch</masterLabel>
+        <openType>sidebar</openType>
+        <protected>false</protected>
+        <url>/apex/npsp__BDI_DataImport?batchId={!DataImportBatch__c.Id}&amp;retURL={!URLFOR(&apos;/&apos; + DataImportBatch__c.Id)}</url>
```
njjc: I'm unable to test this Process Batch button in LEX due to the namespace, we should have this tested in a beta package or during regression.

src/classes/BDI_DataImportService.cls
```diff
@@ -0,0 +1,2252 @@
...
+        list<string> listDiKey = new list<string>();
+
+        // we need any Contact SFId key to always be checked first when matching
+        if (con.Id != null) {
+            listDiKey.add(con.Id);
+        }
+
+        // we need the Contact CustomId key to always be checked second when matching
+        string strUniqueId = '';
+        if (isCustomIDInContactMatchRules && (strUniqueId = strNull(con.get(diSettings.Contact_Custom_Unique_ID__c))) != '') {
```
njjc: This is confusing to me. It looks like strUniqueId is getting a Boolean value.

njjc: I see you're doing the same thing in a bunch of places. I guess it's fine ;)

src/pages/BDI_BatchOverride.page
```diff
@@ -0,0 +1,56 @@
...
+
+            <!-- PAGE HEADER -->
+            <div class="slds-page-header">
+                <div class="slds-grid">
+                    <div class="slds-col slds-has-flexi-truncate">
+                        <div class="slds-media slds-no-space slds-grow">
+                            <div class="slds-media__figure">
+                              <span class="slds-icon_container slds-icon-custom-custom18">
+                                <svg class="slds-icon" aria-hidden="true">
+                                  <use xlink:href="{!URLFOR($Asset.SLDS, 'assets/icons/custom-sprite/svg/symbols.svg#custom17')}" ></use>
```
jjbennett: The SVGs have an issue in Edge when in Lightning and in IE11, I switched to using the .pngs instead. I also created a header component to just pass the png name and button actions for reusability but it's not merged yet, jess is currently testing it.

src/classes/BDI_BatchOverride_CTRL.cls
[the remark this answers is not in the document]
jjbennett: I would use the javascript hybrid approach, we have the UTIL_NavigateBack component that will refresh the page, not sure if that will help with the popup. If you enable popups for your org than you don't get that.

src/classes/BDI_SettingsUI_CTRL.cls
davidhabib: yes, the field does not have to be an official external ID.  As for packaged fields from another package, I entered issue #2700 several weeks ago that BDI does not correctly support them.

src/classes/UTIL_MasterSchedulableHelper.cls
davidhabib: yes, I believe sarah's doc does tell people to verify the job, and if it isn't there, to go to the settings page.

src/classes/BDI_DataImportService.cls
davidhabib: sorry it isn't clear.  it's assigning the unique Id to the variable (so it can be used again later), and then comparing to ''.

src/layouts/DataImportBatch__c-NPSP Data Import Batch Layout.layout
davidhabib: I've just updated the quick action section to put Edit, Clone, Process Batch, and Delete first.  I left the others alone since they are Salesforce defaults, and I really didn't know what was best.  good catch by the way!

davidhabib:
@njjc I think this pull request is ready again.  Thanks to @jjbennett for doing the LEX nav fixes.  I've updated the page layout.  please let me know if there is anything else that needs to be done.

As for the globals, I've reviewed them again, and I'm good with moving forward with them.  Basically, I only made global the most minimal set of properties and methods that our implementation of IMatchDonation needed to use.  This is a pretty small set, and I believe they do make sense to expose to customizers, and I don't believe they will prevent us from further evolution of the BDI_DataImportService in the future.

njjc:
Everything looks great, I agree that the globals are looking good.

My only concern, as mentioned before, is that the scheduled imports won't work with the "Nonprofit Success Pack" user that will schedule the job during the push upgrade's post install script. I think @smarovich should update the release notes to say something along the lines of "Verify that the job is scheduled by a System Administrator, not the `Nonprofit Success Pack` user."

davidhabib:
@judisohn @smarovich please see Nic's comment about the scheduling of the job.  We may need a clarification in the doc.  thanks.

smarovich:
@davidhabib Nic and I discussed his comments and added to the release notes. I'll add it to the docs today too. Thanks!

src/classes/BDI_DataImportService.cls
```diff
@@ -0,0 +1,2252 @@
...
+*
+* The Data Import Service for BDI that processes Data Import records
+*
+*/
+global with sharing class BDI_DataImportService {
+
+    /*******************************************************************************************************
+    * @description constructor
+    */
+    public BDI_DataImportService() {
```
njjc: I'm wondering if this constructor needs to be global so it can be constructed by an external caller.

src/classes/BDI_DataImportService.cls
```diff
@@ -0,0 +1,2252 @@
...
+    */
+    public map<string, Contact> mapDIKeyToContact = new map<string, Contact>();
+    public map<string, Account> mapDIKeyToAccount = new map<string, Account>();
+    private map<Id, Address__c> mapDIIdToAddr = new map<Id, Address__c>();
+    public map<string, Id> mapStrNameToCampaignId = new map<string, Id>();
+
+    /*******************************************************************************************************
+    * @description holds the list of DataImport records we are currently processing
+    */
+    global list<DataImport__c> listDI;
```
njjc: Should these global vars be moved further up so they're more immediately apparent?

src/classes/BDI_IMatchDonations.cls
```diff
@@ -47,6 +47,6 @@ public interface BDI_IMatchDonations {
     * @param mapDIIdToPmt Map to fill in of DI Id's to matched Payments
     * @return void
     */ 
-    void match(BDI_DataImport_BATCH bdi, list<DataImport__c> listDI, 
+    void match(BDI_DataImportService bdi, list<DataImport__c> listDI,
```
njjc: I think we should add global here for clarity, even if it's implicit.

src/classes/BDI_IPostProcess.cls
```diff
@@ -0,0 +1,45 @@
...
+*/
+global interface BDI_IPostProcess {
+
+    /*******************************************************************************************************
+    * @description Perform any additional processing steps after NPSP has completed processing all Data
+    * Import records in the current batch execution.
+    * @param bdi The BDI instance currently running
+    * @return void
+    */
+    void process(BDI_DataImportService bdi);
```
njjc: I think we should add global here for clarity, even if it's implicit.

njjc: Adding global isn't possible with an interface, apparently.

src/classes/BDI_IMatchDonations.cls
njjc: Adding global isn't possible with an interface, apparently.

src/classes/BDI_DataImportService.cls
njjc: The instance of this class is passed fully constructed to the caller.

PR #2726, opened 2017-08-31
Fix apex cpu timeout in TDTM_DMLgt200_TEST class

force2b:
After a lot of unit test experiments with increasing the number of inserted records to force crashes (still sporadic though), I made 3 sets of changes to hopefully avoid future apex cpu time out errors in this test class:

1. Add logic to disable some of the triggers that were not used at all in the test method. Applied this to most of the test methods.
2. Force disable UTIL_Debug logging in all of the unit tests where triggers are disabled. From experience, System.Debug() statements are expensive in terms of CPU cycles especially if they're being called in a loop somewhere.
3. In test_household_account_naming(), removed the update of the Contact address fields to avoid any need for the address related triggers/logic.

Once the above changes were completed I've been unable to get tests to fail even after doubling the number of records inserted. Also all testing was done in a scratch org since that appears to be more fragile in terms of CPU Timeouts than my persistent Dev Org. This will be difficult to "test" other than to let it go through the build process as many times as possible.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/TDTM_DMLgt200_TEST.cls
+++ b/src/classes/TDTM_DMLgt200_TEST.cls
@@ -51,9 +51,9 @@ private class TDTM_DMLgt200_TEST {
     @isTest
     static void test_contacts_with_affiliations() {
 
-        UTIL_Debug.enabled = false;     // Override debug logging for this test because it's an expensive operation
-        disableRelationshipTriggers();
-        disableUnusedContactTriggers();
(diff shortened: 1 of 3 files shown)


davidhabib:
nice job @force2b !  glad disabling some of the unused triggers and other things you did helped!  if you can now do 2x, i'm confident that our little 1.1x should no proceed fine!

PR #2734, opened 2017-09-06
Fix recurring donations unit test issue with testing open ended rds with a date of the 30th

### LINKED ISSUE #2729: Bug in Recurring Donations testOpenEndedMonthlyEndDates() Test
There is a bug in this unit test when calculating the number of generated installments using the current month.

force2b:
Rewrite the unit test method to work properly when the current month has something other than 30 days in it.

# Critical Changes

# Changes

# Issues Closed

Fixes #2729

### DIFF
--- a/src/classes/RD_RecurringDonations_TEST2.cls
+++ b/src/classes/RD_RecurringDonations_TEST2.cls
@@ -1026,7 +1026,6 @@ private with sharing class RD_RecurringDonations_TEST2 {
         }
     }
 
-    /* Commenting this out temporarily
     static testMethod void testOpenEndedMonthlyEndDates(){
         if (strTestOnly != '*' && strTestOnly != 'testOpenEndedMonthlyEndDates') return;
 
@@ -1035,18 +1034,18 @@ private with sharing class RD_RecurringDonations_TEST2 {
             npe03__Opportunity_Forecast_Months__c = 12,
             npe03__Maximum_Donations__c = 50,
             npe03__Open_Opportunity_Behavior__c = RD_RecurringDonations.RecurringDonationCloseOptions.Mark_Opportunities_Closed_Lost.name()        
-            ));             
-        
+            ));
+
         Account a = new Account();
         a.Name = 'test Individual';
         insert a;
-        
+
         Contact c = new Contact();
         c.FirstName = 'test';
         c.LastName = 'contact';
         c.AccountId = a.Id;
-        insert c;       
-        
+        insert c;
+
         Campaign camp = new Campaign(name = 'Campaign', isActive = true);
         insert camp;
         
@@ -1062,7 +1061,7 @@ private with sharing class RD_RecurringDonations_TEST2 {
         RD.npe03__Schedule_Type__c = system.label.npe03.RecurringDonationMultiplyValue;
         RD.npe03__Open_Ended_Status__c = system.label.npe03.RecurringDonationOpenStatus;
         rdlist.add(RD);
-        
+
         RD = RD.clone(false);
         RD.Name = 'testEndOfMonth';
         RD.Always_Use_Last_Day_Of_Month__c = true;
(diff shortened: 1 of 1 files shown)


src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -1084,36 +1091,48 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
         test.stopTest();
-        
+
         list<Opportunity> listOpp28th = [select Id, CloseDate from Opportunity where npe03__Recurring_Donation__r.Name = 'test28th' order by CloseDate];
-        system.assertEquals(12, listOpp28th.size());
+        system.assertEquals(12, listOpp28th.size(), 'There should be 12 Opps with a Close Date of the 28th');
         for (Opportunity opp : listOpp28th) {
-            system.assertEquals(28, opp.CloseDate.Day());
+            system.assertEquals(28, opp.CloseDate.Day(), 'The Opp.CloseDate.Day should be the 28th');
         }
```
jjbennett: @force2b I think that it might be worth splitting this into 2 separate tests. One for the 28th and the other for the 30th.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -1084,36 +1091,48 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
+            system.assertEquals(Date.DaysInMonth(opp.CloseDate.Year(), opp.CloseDate.Month()), opp.CloseDate.Day(),
+                'The Opps Close Date should be the very last day of the month: ' + opp.CloseDate);
+        }
+
+        // For each created installment (going by the 30th), compare the month to the created installments for
+        // the 28th installments. There will always be 12 of those, but there may be less installments for the
+        // ones with a close date of the 30th if the start date was in the future.
+        for (integer i = 0; i < 12-opps30thMonthOffset; i++) {
+            Integer expectedMonth = (listOpp28th[i].CloseDate.Month() + opps30thMonthOffset > 12
+                    ? (listOpp28th[i].CloseDate.Month() + opps30thMonthOffset) - 12
```
jjbennett: This section is a bit confusing, I am not sure we should have the 30th test referring to the 28th records?

force2b: Given the root issue and complexity in this particular test, I've been hesitant to completely rewrite/split into different tests. The above loop is definitely not intuitive, but my reading of it is that the original intent appears to be to compare the generated installments for the 28th and 30th to make sure that the "Month" for each one of the matches. For example, if the RD starts this month, both should create installments from Sept through Aug (next year) - assuming 12 installments are created. Of course, the new `opps30thMonthOffset` var is used to track if the logic creates less than 12 installments and adjust the loop/logic accordingly. The goes through each installment for each RD and makes sure the month matches.

If we are willing to drop this particular verification, then it would definitely make sense to split the `testOpenEndedMonthlyEndDates` method into individual methods to test the 28, 30th and end of month. I suppose that theoretically if each individual test can validate the number of created installments, and possibly the starting and ending months of those installments, then it wouldn't actually be necessary to use a loop like this.

jjbennett: I think that makes the most sense because you are really testing the end dates are the end of the month vs the month itself. It might be worth testing the end of months by just looping through your query results and ensuring that the date + 1 is equal to the next date + 1 month - 1day to assert that it's the last day of each month.

force2b: I'll give it a shot.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -1026,94 +1026,140 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
+        // This should create Installments with a date of the 28th.
+        npe03__Recurring_Donation__c rd = new npe03__Recurring_Donation__c();
+        rd.Name = 'test28th';
+        rd.npe03__Contact__c = c.Id;
+        rd.npe03__Amount__c = 100;
+        rd.npe03__Installment_Period__c = system.label.npe03.RecurringDonationInstallmentPeriodMonthly;
+        rd.npe03__Date_Established__c = system.today().toStartOfMonth().addDays(27);
+        rd.npe03__Schedule_Type__c = system.label.npe03.RecurringDonationMultiplyValue;
+        rd.npe03__Open_Ended_Status__c = system.label.npe03.RecurringDonationOpenStatus;
+
```
jjbennett: @force2b this looks great, I apologize for not mentioning this in the first review but I don't think we should rely on the default being false and should explicitly set - rd.Always_Use_Last_Day_Of_Month__c = false;... thoughts?

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -1026,94 +1026,140 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
-        // TODO: The calculated StartDate is causing the test to fail if the current month has only 30 days.
-        Date startDt = system.today().toStartOfMonth().addMonths(1).addDays(-2);
+        npe03__Recurring_Donation__c rd = new npe03__Recurring_Donation__c();
+        rd.Name = 'test30th';
+        rd.npe03__Contact__c = c.Id;
+        rd.npe03__Amount__c = 100;
+        rd.npe03__Installment_Period__c = system.label.npe03.RecurringDonationInstallmentPeriodMonthly;
+        rd.npe03__Schedule_Type__c = system.label.npe03.RecurringDonationMultiplyValue;
+        rd.npe03__Open_Ended_Status__c = system.label.npe03.RecurringDonationOpenStatus;
+
```
jjbennett: Same for this one.

src/classes/RD_RecurringDonations_TEST2.cls
force2b: We could, but then should we update all of the test methods in the class to do the same thing?

jjbennett: Yea I thought about that too, do you know if customers can change the default on a managed field?

force2b: Just checked. They can't change the default.

jjbennett: Ok sounds good, I think it's ok, I know as a best practice they recommend not relying on default values but since it is in our control, should be fine.

PR #2735, opened 2017-09-07
Rollups + Multi-Currency: Workaround roll-up summary fields when ACM is enabled

force2b:
When Advanced Currency Management is enabled in an Organization, there are 3 fields on the Opportunity that will stop functioning. Two of them are Roll-Up Summary fields from the Payment object and the third is a formula that references those two fields. In addition, the Opportunity.Amount field cannot be used with <Apex:OutputField> on the PaymentWizard VF page. This PR adds logic to manually calculate the total payments and amount outstanding to use with triggers (specifically to auto-close an Opportunity when TotalPayments==Opp.Amount), the payment wizard page controller, and unit tests. 

This merges into the feature/rollups branch only.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/TDTM_DefaultConfig.cls
+++ b/src/classes/TDTM_DefaultConfig.cls
@@ -197,7 +197,7 @@ public without sharing class TDTM_DefaultConfig {
         // Payment records - currency code validation
         handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false,
                 Class__c = 'PMT_Payment_TDTM', Load_Order__c = 0, Object__c = 'npe01__OppPayment__c',
-                Trigger_Action__c = 'BeforeUpdate;BeforeInsert'));
+                Trigger_Action__c = 'BeforeUpdate;BeforeInsert;AfterInsert;AfterUpdate'));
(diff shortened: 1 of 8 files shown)


src/classes/TDTM_DefaultConfig.cls
```diff
@@ -197,7 +197,7 @@ public without sharing class TDTM_DefaultConfig {
         // Payment records - currency code validation
         handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false,
                 Class__c = 'PMT_Payment_TDTM', Load_Order__c = 0, Object__c = 'npe01__OppPayment__c',
-                Trigger_Action__c = 'BeforeUpdate;BeforeInsert'));
+                Trigger_Action__c = 'BeforeUpdate;BeforeInsert;AfterInsert;AfterUpdate'));
```
njjc: I think this is where the ordering of the operations is important for health check; was that fixed another way?

force2b: Yes. That was fixed in PR #2737.

PR #2736, opened 2017-09-07
Rollups + Multi-Currency: Add support for Advanced Currency Management and exchange rate caching

### LINKED ISSUE #2438: upon Contact creation with specified currency, make the Account have a matching currency code
This would only be for cases where user is using automatic Account creation (using HH model) and starting with creating the Contact record first. If the currency is set on the Contact, it should copy over to the Account as well.

force2b:
This PR adds utility classes to handle caching of exchange rates (both standard and dated) using the Platform Cache or a static var if the Platform Cache is not available. A new Utility class with conversion methods is included for use with the new Roll-Ups logic to support converting any amount using the appropriate date (Opportunity.CloseDate) when ACM is enabled. Legacy roll-ups have been refactored to use the new conversion methods, but will force them to use the standard currency exchange rate even if ACM is enabled in the org.

# Critical Changes

# Changes

# Issues Closed


src/classes/UTIL_CurrencyConversion.cls
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * @description Static flag to force the conversion logic to always use the standard currency rates
+     * rather than dated exchange rates (if enabled). This is necessary to support the legacy roll-ups that
+     * are based on aggragate queries since those only use the standard (i.e., default rate) rather than
+     * the dated exchange rate.
+     */
+    public static Boolean alwaysUseStandardCurrencyRate = false;
+
+    private static final Integer RATE_DECIMAL_PLACES = 6;
+    private static final Integer FINAL_DECIMAL_PLACES = 2;
```
force2b: Move this static var to a dynamic method that retrieves the defined decimal place values for the target currency code.

src/classes/ALLO_Multicurrency_TDTM.cls
```diff
@@ -159,33 +156,14 @@ public class ALLO_Multicurrency_TDTM extends TDTM_Runnable {
...
-        }        
-        return multiCurrency;
+
+    /**
+     * @description Wrapper for the utility method to determine if multi-currency is enabled in the org
+     * The target utility class can be mocked for unit tests.
+     * @return True if multi-currency is enabled in the org
+     */
+    public static Boolean isMultiCurrency() {
+        return UTIL_Currency.getInstance().isMultiCurrencyOrganization();
```
force2b: Is this method necessary at all?

src/classes/ALLO_Multicurrency_TDTM.cls
```diff
@@ -194,12 +172,9 @@ public class ALLO_Multicurrency_TDTM extends TDTM_Runnable {
     * @param amt A decimal of the amount to convert.
     * @return decimal The converted amount.
     ********************************************************************************************************/
-        //do a conversion from amount in corporate currency to the currency specified by the ISO code
+    //do a conversion from amount in corporate currency to the currency specified by the ISO code
     public static decimal convertFromCorporate(String ISO,Decimal amt) {
```
force2b: Where else are these referenced? If not many, is the method still necessary or can we just call UTIL_ConvertCurrency. directly?

src/classes/RLLP_OppRollup_UTIL.cls
```diff
@@ -319,67 +317,54 @@ public without sharing class RLLP_OppRollup_UTIL {
...
-            return ((amt / exchangeRates.get(fromISO)) * exchangeRates.get(toISO)).setScale(2,System.RoundingMode.UP);
+
+    /*******************************************************************************************************
+    * @description Converts an amount from any given currency to any other currency.
+    * @param fromISO A string of the currency's iso code of the amount we wish to convert.
+    * @param toISO A string of the currency's iso code that we wish to convert the amount to.
+    * @param amt A decimal of the amount to convert.
+    * @return decimal The converted amount.
+    ********************************************************************************************************/
+    public static decimal convertCurrency(String fromISO, String toISO, Decimal amt) {
```
force2b: Is this needed or can we call UTIL_CurrencyConversion directly from wherever this is currently referenced?

src/classes/RLLP_OppRollup_UTIL.cls
```diff
@@ -319,67 +317,54 @@ public without sharing class RLLP_OppRollup_UTIL {
...
     /*********************************************************************************************************
-    * @description Determines if multi-currency is enabled for the org.
+    * @description Determines if multi-currency is enabled for the org, but also adds the currency iso code
+    * fields to the related
     * @return Boolean representing whether or not multi-currency is enabled.
     */
     public static boolean isMultiCurrency(){                    
+
         if (multiCurrency==null){
             //create currency field handles for later dynamic dml
```
force2b: This class can be refactored further to remove older logic, if time allows and it will help with readability/functionality.

src/objects/npo02__Households_Settings__c.object
```diff
@@ -13,4 +13,13 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Use_Dated_Conversion_Rates__c</fullName>
+        <defaultValue>false</defaultValue>
+        <externalId>false</externalId>
+        <inlineHelpText>heck this box to use Dated Conversion Rates when Advanced Currency Management is enabled in this Organization. If unchecked and ACM is enabled, all NPSP roll-ups will use the default exchange rate for by currency.</inlineHelpText>
```
force2b: Spelling Error!

force2b:
@njjc All discussed changes have been made.

src/classes/ALLO_Multicurrency_TDTM.cls
njjc: Let's remove all these different methods and consolidate them all to use UserInfo.isMultiCurrencyOrganization()

src/classes/ALLO_Rollup_SCHED.cls
```diff
@@ -197,16 +198,19 @@ public without sharing class ALLO_Rollup_SCHED extends UTIL_AbstractRollup_BATCH
     * @return void This method modifies the objMap parameter if needed, and returns nothing.
     ********************************************************************************************************/
     public override void customFinalProcessing (map<id, sobject> objMap) {
+        // TODO: Fix the currency conversion logic here; though it'll probably all get stripped out anyway
```
njjc: Isn't this already done below?

src/classes/UTIL_CurrencyCache.cls
```diff
@@ -0,0 +1,310 @@
...
+
+                UTIL_CurrencyCache.CurrencyData currData = mapOfCurrencyDataByCurrencyCode.get(currCode);
+                if (currData == null) {
+                    currData = new UTIL_CurrencyCache.CurrencyData();
+                }
+
+                currData.defaultRate = (Decimal)scr.get('ConversionRate');
+                currData.isOrgDefault = (Boolean)scr.get('IsCorporate');
+                currData.decimalPlaces = (Integer)scr.get('DecimalPlaces');
+                mapOfCurrencyDataByCurrencyCode.put(currCode, currData);
```
njjc: Couldn't this be overwriting the dated conversion rate data for the same code from the previous block? Can't orgs have both?

src/classes/UTIL_CurrencyConversion.cls
```diff
@@ -0,0 +1,131 @@
...
+    /*******************************************************************************************************
+    * @description Converts an amount from any given currency to any other currency always using the
+    *   non-dated exchange rate exchange rate even if ACM is enabled in the org.
+    * @param fromISO A string of the currency's iso code of the amount we wish to convert.
+    * @param toISO A string of the currency's iso code that we wish to convert the amount to.
+    * @param amt A decimal of the amount to convert.
+    * @return decimal The converted amount.
+    ********************************************************************************************************/
+    public static decimal convertAmountUsingStandardRates(String fromISO, String toISO, Decimal amt) {
+        alwaysUseStandardCurrencyRate = true;
```
njjc: What if the same transaction wants standard rates for one conversion, then dated for another conversion? For example, converting an amount on a non-ACM supported object, then for an Opportunity? Wouldn't this set all other conversions in that transaction to use standard rates?

src/classes/ALLO_Multicurrency_TDTM.cls
force2b: They do have slightly different uses though:
* UserInfo.isMultiCurrencyOrganization() - Absolute. Based on the Org's configuration.
* UTIL_Currency.getInstance().isMultiCurrencyOrganization() - Can be mocked in Unit Tests to return a value of True even if the org doesn't support multi-currency.

njjc: Makes sense!

src/classes/UTIL_CurrencyCache.cls
force2b: Line 255 attempts to retrieve the CurrencyData object from the map. If that returns null, it creates a new object. Line 263 puts the new/updated object back into the map.

PR #2737, opened 2017-09-07
Fix Health Check warning when trigger handler actions do not exactly match in alphabetical order

### LINKED ISSUE #2732: Health Check false positive of a Trigger Handler error on PMT_Payment_TDTM on object npe01__OppPayment__c
[Thread.](https://powerofus.force.com/0D58000003dBd0t)

Confirmed in test org with latest 3.106 release. 

image

force2b:
Update the Health Check trigger verification logic to compare the trigger actions without regards to the order or case of the actions in the lists.

# Critical Changes

# Changes

# Issues Closed

Fixes #2732

### DIFF
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -587,8 +587,8 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
                         alreadyExists = true;
     
                         // Determine if order, action, active, async has changed
-                        if (defaultHandler.Load_Order__c != existingHandler.Load_Order__c || 
-                            defaultHandler.Trigger_Action__c != existingHandler.Trigger_Action__c ||
+                        if (defaultHandler.Load_Order__c != existingHandler.Load_Order__c ||
+                            !compareTriggerHandlerActions(defaultHandler, existingHandler) ||
                             defaultHandler.Active__c != existingHandler.Active__c ||
                             defaultHandler.Asynchronous__c != existingHandler.Asynchronous__c) {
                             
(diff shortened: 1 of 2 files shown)


src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -620,6 +620,25 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+    private static Boolean compareTriggerHandlerActions(Trigger_Handler__c handlerA, Trigger_Handler__c handlerB) {
+        List<String> listA = handlerA.Trigger_Action__c.toLowerCase().split(';');
+        List<String> listB = handlerB.Trigger_Action__c.toLowerCase().split(';');
+
+        Set<String> unionBintoA = new Set<String>();
+        unionBintoA.addAll(listA);
+        unionBintoA.addAll(listB);
+
+        // The number of elements in in unionBintoA should match the number of elements in listA and listB
+        if (listA.size() != listB.size() || unionBintoA.size() != listA.size() || unionBintoA.size() != listB.size()) {
```
jjbennett: Hi @force2b, I'm finding this conditional a little hard to follow. Isn't the requirement that the values are the same so A should always be equal to B and the values should be the same not including order? Could we just create 2 sets here and attempt adding all of set B to set A and see if that returns false and then check the flip of adding all of A into B and verify that returns false as well?

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -587,8 +587,8 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
                         alreadyExists = true;
     
                         // Determine if order, action, active, async has changed
-                        if (defaultHandler.Load_Order__c != existingHandler.Load_Order__c || 
-                            defaultHandler.Trigger_Action__c != existingHandler.Trigger_Action__c ||
+                        if (defaultHandler.Load_Order__c != existingHandler.Load_Order__c ||
+                            !compareTriggerHandlerActions(defaultHandler, existingHandler) ||
```
jjbennett: Should we also have a regression test for this?

src/classes/STG_PanelHealthCheck_CTRL.cls
force2b: I didn't feel it was necessary to flip the merges. Here is I looked at the test and result:

*Test Scenario 1:*
* Set A: { A1, B2, C3 }
* Set B: { A1, c3, B2 }
* Merged (lowercase): { a1, b2, c2 }
* Pass: The size of all the is still the same.

*Test Scenario 2*
* Set A: { A1, B2, C3 }
* Set B: { A1, c3, D2 }
* Merged (lowercase): { a1, b2, c3, d2 }
* Fail: The size of the merged set is not the same as sets A and B

*Test Scenario 3:*
* Set A: { A1 }
* Set B: { A1, B2, C3 }
* Merged (lowercase): { a1, b2, c3 }
* Fail: The size of the merged set is not the same as set A

src/classes/STG_PanelHealthCheck_CTRL.cls
force2b: I was surprised there was no existing unit test for the trigger handler test specifically. One can be built though.

src/classes/STG_PanelHealthCheck_CTRL.cls
jjbennett: Yup, for whatever reason I thinking unionAintoB was another list not a set. Thanks for clarifying!

src/classes/STG_PanelHealthCheck_CTRL.cls
jjbennett: If you wouldn't mind that'd be great.

PR #2752, opened 2017-09-14
Feature/jen usr trigger bug fix

jjbennett:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/TDTM_TriggerHandler.cls
+++ b/src/classes/TDTM_TriggerHandler.cls
@@ -79,6 +79,7 @@ public class TDTM_TriggerHandler {
 
         // If there are no records, insert the defaults.
         if(!defaultRecordsInserted && dao.isEmpty()) {
+            // We will not insert the default records when Triggered by the User object to prevent MixedDML errors
             if (describeObj == Schema.SObjectType.User) {
                 return;
             }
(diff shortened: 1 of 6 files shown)


src/classes/USER_MixedDML_TEST.cls
```diff
@@ -0,0 +1,44 @@
...
+/**
+* @author Salesforce.org
+* @date 2017
+* @group User
+* @description Test class for the USER Trigger Mixed DML bug regression.
+*/
+@isTest
+public with sharing class USER_MixedDML_TEST {
+    private static testMethod void testInsertingUserWithTriggerHandlerMixedDML() {
+        List<User> newUsers = USER_UserService_TEST.createUsers();
```
force2b: Should there be an assert here to verify that the TriggerHandler object is still empty?

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -79,6 +79,9 @@ public class TDTM_TriggerHandler {
 
         // If there are no records, insert the defaults.
         if(!defaultRecordsInserted && dao.isEmpty()) {
+            if (describeObj == Schema.SObjectType.User) {
```
davidhabib: please put in a comment why we are doing this!

PR #2754, opened 2017-09-14
Update scoped slds to use native slds

force2b:
Other key changes:
* Rewrote much of Contact Merge page to use custom labels for all static text to allow for eventual localization
* Rewrote much of the PaymentWizard page to standardize structure and notification messages
* Updates to the UTIL_PageHeader and UTIL_FormField components to implement new attributes
* Remove the now deprecated SFDO203 static resource
* The SLDS static resource and UTIL_JobProgress component can be removed after the work item to complete SLDS changes on the settings pages.

# Critical Changes

# Changes

# Issues Closed


jjbennett:
@RadSFDeveloper this should be ready for review. I have all fixes in for the header issue we spoke about and Jess is doing the QA as well now. Thanks for being a 2nd pair of eyes on this!

src/labels/CustomLabels.labels
```diff
@@ -1847,71 +1847,71 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
         <value>Search Contacts</value>
     </labels>
     <labels>
         <fullName>conMergeSelectAll</fullName>
         <categories>Contact Merge</categories>
         <language>en_US</language>
-        <protected>false</protected>
+        <protected>true</protected>
         <shortDescription>conMergeSelectAll</shortDescription>
-        <value>select All</value>
```
force2b: @jjbennett I had left this lower case because the original text was lower case, though I think it's probably better this way.

force2b:
@jjbennett I found a few page controllers (Account conversion and BDI) that have `'/home/home.jsp'` hardcoded as the destination on the close/cancel operation. Those fail in LEX. I think it's better to fix these here rather than open a new work item for them. I'll commit changes for them shortly.

src/pages/OPP_SendAcknowledgmentBTN.page
```diff
@@ -2,26 +2,21 @@
...
-                            <apex:commandButton id="close" styleClass="slds-button slds-button_neutral" action="{!close}"
-                                value="{!$Label.sendAcknowledgmentReturnLinkLabel}" immediate="true" reRender="vfForm"/>
-                        </div>
-                    </div>
-                </div>
-            </div>
+            <c:UTIL_PageHeader headerLabel="{!$ObjectType.Opportunity.Label}"
+                               header="{!$Label.sendAcknowledgmentPageTitle}"
+                               icon="opportunity_120" iconCategory="standard"
+                               showSaveBtn="false" cancelLabel="{!$Label.bdiBtnClose}"
```
RadSFDeveloper: Should the Cancel label be `$Label.sendAcknowledgmentReturnLinkLabel`?

force2b: I believe that this button can be called either from a specific Opportunity or from a list of Opportunities. In the case of the latter, it would make more sense to have the button label of Close instead of "Return to Opportunity".  Is that correct @jjbennett?

jjbennett: @RadSFDeveloper @force2b Yes, that is correct so I changed it to "Close" to work in either case.

PR #2757, opened 2017-09-18
special case all 4 Audit fields that may have been enabled.

### LINKED ISSUE #1864: If Audit Fields are enabled through Permission Set, NPSP Data Import should allow edit as custom field
Steps to repeat:
- Create Permission Set enabling Audit Field creation after allowing in User Interface settings
- Add to Admin User
- Custom Date/Time field for Contact 1 Created Date in NPSP Data Import w/help text Contact1.CreatedDate
- Create test record and process. Error:    You do not have permissions to modify Contact.CreatedDate

Power of Us thread: https://powerofus.force.com/0D58000002Q46yv

davidhabib:
# Critical Changes

# Changes

# Issues Closed
Fixes #1864

### DIFF
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -676,18 +676,19 @@ global with sharing class BDI_DataImportService {
 
                     // make sure they have permissions to modify the target field
                     // we can't do this in all tests, because many profiles won't have FLS set for all our new fields!
-                    // also they may have enabled "Set Audit Fields upon Record Creation", which doesn't seem to change the
-                    // describe results!
-                    if (strField != 'CreatedDate' &&
-                        strField != 'CreatedById' &&
-                        strField != 'LastModifiedDate' &&
-                        strField != 'LastModifiedById' &&
-                        (!Test.isRunningTest() || failFieldLevelSecurity)) {
+                    if (!Test.isRunningTest() || failFieldLevelSecurity) {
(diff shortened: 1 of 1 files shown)


njjc:
@davidhabib we're checking isCreatable() for audit fields in the [conversion utility](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/CONV_Account_Conversion_BATCH.cls#L157), does your code comment mean that approach doesn't work?

davidhabib:
@njjc from the comment I put in the code back in 2016: "also they may have enabled "Set Audit Fields upon Record Creation", which doesn't seem to change the describe results!"

when I tested back then, enabling the perms didn't seem to be enough.  the describe still said false to updateable.

if we are really concerned about this, I can go back and test and do research.  I can't remember what I found out so long ago.

njjc:
I'm just concerned that the Conversion code might not work, but I'm quite positive it does. I think the difference is that audit fields set isCreatable() but not isUpdateable(), since they're only available to set on insert, not on update.

davidhabib:
ah, you might be onto something.  BDI is checking for updateable, since it frequently is updating exisiting contacts.  but it does make sense that the audit fields are only enabling creatable.  I'll look into it some more and decide how to code it best.  don't merge this yet.

davidhabib:
@njjc ok, updated code to check for creatability for audit fields, and creatability and updateability for all other fields.

PR #2767, opened 2017-09-22
Address verification fixes and enhancements

### LINKED ISSUE #2713: With SmartyStreets address verification, Verification Status field not updating correctly
Hub thread [here](https://powerofus.force.com/0D58000003btFpg).

njjc:
# Critical Changes

# Changes

- When Enable Automatic Verification (in NPSP Settings under People > Addresses > Verification API Account Settings) is selected, NPSP no longer tries to verify Addresses that already have the Verified checkbox selected.
- When verification is attempted, NPSP will now always select an Address's Verified checkbox, regardless of the outcome of the verification. You can use the Verification Status field on an Address to determine if a match was found.

# Issues Closed
- Fixed #2713

# New Metadata

# Deleted Metadata


njjc:
@smarovich could you review the changes? Thanks!

davidhabib:
@njjc @szeitz @mpusto I don't think we want to merge this into this release until we've spec'd out and implemented all the changes in this area.  This way we roll out to customers a single release with all the address changes we decide to do.

So unless someone disagrees, I'm going to hold off on reviewing this.  thanks.

njjc:
I disagree. I think running address verification on every Address object DML is a bug worth addressing immediately, since it has already costed one of our customers a lot of money. I thought we were all in agreement in the meeting we had on Thursday, which is why I took on this work in the middle of a sprint. @davidhabib

davidhabib:
this is a non-trivial change to our address behavior from the last 3+ years, so I don't view it as urgent.  And I'm just channeling what I heard from PM over the last few months where they really wanted us to hold off on features/changes that were not 100% done, so that we only rolled them out to customers once with all the functionality changes and documentation in place.  That said, I leave this up to @mpusto and @szeitz to decide what is appropriate here.

szeitz:
Given that this issue puts customers at risk of being charged when unnecessary address verifications are completed, PM would like to have this bug fixed in this release. 

FYI - @mpusto

src/classes/ADDR_Validator_TDTM.cls
```diff
@@ -68,21 +68,22 @@ public class ADDR_Validator_TDTM extends TDTM_Runnable {
...
                     if(!system.isFuture() && !system.isBatch()) {
                         address.Verification_Status__c = Label.Addr_Pending_Verification;
                     } else {
                         address.Verification_Status__c = Label.Addr_Verification_Required;
                     }
                 }
             }
-            if ((triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) &&
-               settings.enable_automatic_verification__c) {
+            if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
```
davidhabib: why do the afterInsert/Update handlers now process without checking enable_automatic_verification__c?

davidhabib:
@njjc @szeitz what is the expected behavior of the user clicking the Verify Address button on the address page layout?  with this change, it will do nothing if the address was previously verified, even if the user has changed the mailing fields.  will this issue be covered by future work items?  thanks.

njjc:
Good catch, I think we should update the button to uncheck Verified before sending the record for verification.

davidhabib:
agreed.  @njjc is that an easy enough change to put into this pull request?

njjc:
Yep, I'm on it.

njjc:
@davidhabib this should be good to go! Thank you.

src/classes/ADDR_Validator.cls
```diff
@@ -59,22 +59,29 @@ public with sharing class ADDR_Validator {
...
     @future(callout=true)
     public static void verifyRecordsFuture(set<id> recordIds) {
-        verifyRecords(recordIds);
+        verifyRecords(recordIds, false);
     }
     
     /*******************************************************************************************************
     * @description Verifies a set of address records.
     * @param recordIds A set containing the IDs of the address records to verify.
     * @return void
```
davidhabib: please add the new param to the header comment!

src/classes/ADDR_Validator_TDTM.cls
davidhabib: @njjc still wondering about this one.  shouldn't it continue to check for enable-automatic_verification__c?

njjc: It's still there in the first if statement... just refactored a bit.

davidhabib: had to look at the full file.  looking at just the diff I missed it.  thanks.

judisohn:
Thank you all for getting on this quickly.

PR #2774, opened 2017-09-28
Feature/jen id fixups

### LINKED ISSUE #2233: update Manage Allocations page UI to be consistent with Manage Soft Credits UI
Related to [Issue 2227](https://github.com/SalesforceFoundation/Cumulus/issues/2227), the presentation of error messages in the Manage Allocations UI should be evaluated and modified to provide a consistent experience. This would include modifying the error message presentation to match styling of the Manage Soft Credits UI as well as having the doc team review the messages for consistency.

### LINKED ISSUE #2781: Manage Save Button is not Disabling When Allocating Amount exceeds Parent Amount
After the page was updated to the latest version of SLDS the Save button is no longer disabling when the Allocated Amount exceeds the Parent Amount.

jjbennett:
Adding Ids where the browser tests were looking at slds styles, updated Ids for the payment wizard page in the payment fieldset to avoid any merge conflicts. Also fixed up the hyphens in the slds styles in the Engagement Plan controller. Fixed disabling of button for on page javascript logic. Updating page to use SLDS Error message.

# Critical Changes

# Changes

# Issues Closed
Fixes #2233 
Fixes #2781

# New Metadata

# Deleted Metadata


src/components/UTIL_PageHeader.component
```diff
@@ -63,17 +63,17 @@
     <apex:attribute name="saveAndNewDisabled" type="Boolean" default="false"
         description="Whether the save and new button should be disabled, defaults to false."/>
 
-    <div class="slds-page-header" role="banner">
+    <div class="slds-page-header" role="banner" id="header">
```
force2b: Should this have a more specific Id value to avoid a conflict? For example, `npsp-page-header`

src/components/UTIL_PageHeader.component
```diff
@@ -63,17 +63,17 @@
...
         <div class="slds-grid">
             <div class="slds-col slds-col_padded">
                 <apex:outputPanel layout="none" rendered="{!showBreadcrumb}">
                     <nav role="navigation" class="slds-p-bottom_x-small">
                         <p id="bread-crumb-label" class="slds-assistive-text">{!$Label.lblYouAreHere}</p>
-                        <ol class="slds-breadcrumb slds-list_horizontal slds-wrap" aria-labelledby="bread-crumb-label">
-                            <li class="slds-breadcrumb__item slds-text-title_caps">
+                        <ol class="slds-breadcrumb slds-list_horizontal slds-wrap" aria-labelledby="bread-crumb-label" id="headerBreadcrumb">
+                            <li class="slds-breadcrumb__item slds-text-title_caps" id="parentEntity">
```
force2b: Same with these Id's. It's possible that some other native Salesforce markup might use Id's with generic names such as these; or that pages that use this component might use these same Id's.

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -417,43 +418,43 @@ public with sharing class EP_ManageEPTemplate_CTRL {
...
+            'class="slds-button slds-button_brand" id="btnAddDepTask">' + Label.EPAddDependentTask + 
             '<span class="slds-assistive-text"> ' + Label.EPTo + Schema.Sobjecttype.Task.getLabel() + ' ' + wrapper.level + 
             '</span></button>',
             escape=false
         );
          
         Component.Apex.OutputText deleteTaskButton = new Component.Apex.OutputText(
             value='<button onclick="deleteTaskJS(\''+ wrapper.level +'\');return false;" '+ 
-            'class="slds-button slds-button--destructive">' + Label.EPDeleteTask +
+            'class="slds-button slds-button_destructive" id="btnDelTask">' + Label.EPDeleteTask +
```
force2b: These Id's probably exist mostly for the browser tests, so this may not be necessary, however these buttons can appear multiple times on the page. Would it make sense to use an iterative number on the button Id's to ensure uniqueness -- `btnAddDepTask0`, `btnDelTask0`, `btnAddDepTask1`, `btnDelTask1`, etc.

src/pages/ALLO_ManageAllocations.page
```diff
@@ -71,10 +71,10 @@
                 $('#totalAmount').text(reformatCurrency(unallocated));
                 if (unallocated < 0) {
                     $('#totalAmount').css({ "color": "red", "font-weight": "bold" });
-                    $('[id$="saveCloseBTN"]').attr('disabled','disabled');
```
force2b: I'm not seeing the buttons enable/disable on the page. I think it's because the Id on the button component is an Apex attribute id, creating a fully qualified Id along the lines of `j_id0:theForm:j_id6:j_id7:saveBTN`. The selector here would need to either select that fully formed Id (challenging) or use a class selector - maybe something like `$('.npsp-pageheader-savebtn')`

src/pages/CON_ContactMerge.page
```diff
@@ -153,7 +153,7 @@
                                             <td class="slds-row-select">
```
force2b: Didn't this page also need the `tabStyle="Contact_Merge__tab"` attribute?

src/pages/ALLO_ManageAllocations.page
jjbennett: @force2b I did test this out and the button should be disabling now.

src/pages/CON_ContactMerge.page
jjbennett: @force2b thanks, fixed!

src/pages/ALLO_ManageAllocations.page
force2b: I'm not seeing that behavior in my scratch org. Also, do you think we fix the error message styling on this page in this branch?

src/pages/ALLO_ManageAllocations.page
```diff
@@ -155,7 +155,7 @@
                 parentRecordAction="/{!parentId}" header="{!pageTitle}" icon="custom_120" iconCategory="standard"
                 cancelAction="{!cancel}" saveAction="{!saveClose}" saveDisabled="{!opp<>null&&parentAmount==0}"/>
             <div> 
-                <apex:pageMessages id="messages"/>
+                <c:UTIL_PageMessages id="messages"/>
```
force2b: I'm seeing messages in the new format, however if the [Manage Allocations] button is clicked on an Opportunity with a $0 or null Amount field, the page is displayed without any message. The Save button is disabled (which is good).

jjbennett: @force2b I do not see anything in the controller that would add that error message, I agree that would be good to have but wondering if it should be done with this work?

force2b: I'm pretty sure it was there earlier when we were testing on hangouts earlier. I cleared the amount field, and then clicked the Manage Allocations button. I could swear there was a message displayed on the screen. Could be wrong too, but I really thought it was there.

jjbennett: I wonder if it was because you already had allocations saved and then nulled out the amount and then went back to manage?

force2b: Just put the <apex:pageMessages> tag back in and retested. The message displayed is "You can only create allocations for opportunities with a positive amount."

jjbennett: @force2b ok, I will take another look, thanks!

jjbennett: @force2b thanks again for catching that! Chicken before the egg type of an issue, should be all set now!

force2b:
@jjbennett Should the PR description mention the fix to the allocation page for the buttons and error message styling?

jjbennett:
@smarovich - would you have time to review the error messages on this page? There is an open issue to fix the error message on this page which we have done but I didn't change the errors themselves as described in this item... #2233

src/classes/EP_ManageEPTemplate_CTRL.cls
jessingrass: That seems like a good solution and helpful if it is possible. Would putting iterative numbers mean that the numbers would change if a button was erased (say you have buttons 1-5, if you erased button 4, would button 5's number change to 4 or stay the same?)

jjbennett: @jessingrass - I'd like to leave what we have for now and revisit once we've had a chance to review what is needed to move from style class to Id's. Then we can start coming up with best practices and start working through each page together, does that work?

chrismcmahon: @force2b instead of an iterative number, ideal would be a description of the button's function, like btnDoesThis, btnDoesThat, btnDoesTheOther, so that if btnDoesThat is not on the page, we know from the label what the other buttons on the page are supposed to accomplish.

PR #2786, opened 2017-10-03
BDI support for UniqueID field being namespaced

### LINKED ISSUE #2700: BDI: custom Contact and Account Unique ID's don't support fields from other packages

davidhabib:
# Critical Changes

# Changes
- The Batch Data Import picklist settings 'Contact Custom Unique ID' and 'Account Custom Unique ID' now only display fields that are marked Unique or External ID, and those fields can be from other managed packages.

# Issues Closed
Fixes #2700 

# New Metadata

# Deleted Metadata


davidhabib:
@jjbennett not sure why gitHub isn't currently letting me assign this to you!  but please review.

davidhabib:
@jjbennett nevermind.  i really should have @njjc review this since he is more familiar with BDI.  thanks.

PR #2788, opened 2017-10-03
Rollback transaction on error

RadSFDeveloper:
Apply rollback when a save method fails applying one of Database operations in the try/catch block.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -247,7 +257,30 @@ public with sharing class EP_ManageEPTemplate_CTRL {
...
+
+                Id existingParentId = epTask.Id == null
+                    ? null
+                    : existingParentIdsByTaskId.get(epTask.Id);
+
+                epTask.Parent_Task__c = existingParentId != null
+                    ? existingParentId
+                    : wrapper.parent != null ? wrapper.parent.detail.Id : null;
+            }
+
```
davidhabib: interesting!  so even though we've done the rollback, we have to manually fixup the state of any data structures we may have modified during the save.  I don't believe we discussed that during our team discussions about transaction handling.  It makes sense when you think about it.  Could you revisit your transaction investigation document and make sure this is clearly mentioned in it.  thanks!

src/classes/HH_Container_LCTRL.cls
```diff
@@ -349,15 +349,22 @@ public with sharing class HH_Container_LCTRL {
...
             update listUpdate;
-            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.HH, false);
+
         } catch (Exception ex) {
+            Database.rollback(sp);
+
             throw new AuraHandledException(ex.getMessage());
+
+        } finally {
+            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.HH, false);
```
davidhabib: I don't mind that you moved these into a finally block.  but I just want to confirm that since we are throwing an exception back to Aura and the client, that the transaction scope would have completed, and our state flags would have gone out of scope, no?

src/classes/OPP_SendAcknowledgmentBTN_TEST.cls
```diff
@@ -35,43 +35,89 @@
...
+
+        for (Opportunity opp : [SELECT Acknowledgment_Status__c FROM Opportunity]) {
+            System.assertEquals(TO_BE_ACKNOWLEDGED_STATUS, opp.Acknowledgment_Status__c, 'The Acknowledgment Status should be unchanged');
+        }     
+    }
+
+    // Helpers
+    ////////////
+
+    private static List<Opportunity> setUpOpportunities(Integer oppCount) {
```
davidhabib: even though this is a private helper method in a test class, it would still be nice to still see it with its ApexDoc function description.

src/classes/PSC_ManageSoftCredits_CTRL.cls
```diff
@@ -201,64 +201,91 @@ public with sharing class PSC_ManageSoftCredits_CTRL {
...
         List<SoftCredit> upsertedSoftCredits = new List<SoftCredit>();
-        for (SoftCredit sc : softCredits) {
+        Map<Integer, OpportunityContactRole> rollbackContactRole = new Map<Integer, OpportunityContactRole>();
+        Map<Integer, Partial_Soft_Credit__c> rollbackPartialSoftCredit = new Map<Integer, Partial_Soft_Credit__c>();        
+
+        for (Integer i = 0, size = softCredits.size(); i < size; i++) {
+            SoftCredit sc = softCredits[i];
+
+            rollbackContactRole.put(i, sc.contactRole == null ? null : sc.contactRole.clone(true, true, false, false));
+            rollbackPartialSoftCredit.put(i, sc.partial == null ? null : sc.partial.clone(true, true, false, false));
```
davidhabib: why when cloning the psc's and ocr's, you don't want to preserve readonly fields and autonumbers?

src/classes/STG_InstallScript.cls
```diff
@@ -51,17 +55,27 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+
+            if (context.previousVersion() == null || context.previousVersion().compareTo(new Version(3, 79)) < 0) {
+                // update record type settings ONLY if upgrading from a version prior to 3.78
+                UTIL_RecordTypeSettingsUpdate.getInstance().updateSettings();
+            }
+
         } catch(Exception e) {
+            Database.rollback(sp);
+
+            if (rethrowException) {
```
davidhabib: it's not clear to me that the original intent was if the updateSettings() hit an error, that we'd want the install script to throw the exception.  I just wonder if it was accidentally added after the try/catch block, rather than being intentional.  Can you please talk to @dospromptman, who added this settings work, about a year ago, to see if he recalls whether the intent was for this to stop the install, rather than just log the problem like other install errors?  thanks.

src/classes/STG_InstallScript.cls
```diff
@@ -51,17 +55,27 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+                throw e;
+                return;
```
davidhabib: don't need a return statement after a throw.

src/classes/EP_ManageEPTemplate_CTRL.cls
RadSFDeveloper: When a user fixes a cause for the error and clicks Save again, an error might be generated due modifications applied in the first Save (insert of sObject will not clear the Id on the rollback for example). However, not every method having Database.Rollback() applied needs to "rollback" its variables.
The transaction document has been updated to include the reset of sObjects when rollback is applied. Please let me know if anything more should be added to the document.

src/classes/HH_Container_LCTRL.cls
RadSFDeveloper: Yes, that is correct. If a such method is called from a VF page, then changes applied to the static variables are reset when the method completes regardless if it is success or not. The issue is only applicable if the method is called from another one and the caller catches the exception and continues the transaction: static/instance/local (primitive data type or SObject) variables are not implicitly rolled back.
The reset of the static variable in the above method was not required even though the method is called from another one. The caller does not catch the exception and the transaction stops if the exception has been raised. I can revert back changes -- please let me know.

src/classes/STG_InstallScript.cls
dospromptman: Yes, it was intentional.  Here is a link to the comment and pull request that talks about moving it outside the try/catch: https://github.com/SalesforceFoundation/Cumulus/pull/2184#issuecomment-236970992

I can't find the original conversation we had around this-- it probably happened in HipChat.  But the reasoning was that if this conversion process from Record Type Name to Record Type Id failed, then the user could possibly lose data and regardless the behavior of NPSP in their org would not be correct (due to the code expecting the settings to be in the new format).

So, if the setting conversion process failed for any reason then we would want the push upgrade to fail so we didn't put the user's org in a state where NPSP would behave incorrectly.  The way to fail the install was to allow the exception to remain unhandled.

As I recall, there were some instances where it did actually did cause the push upgrade to fail in customer orgs.  I don't remember the exact reasons, but I think it was something about the length limit of the field and the converted settings not fitting?  I think the failed install helped identify the issue and the customer fixed it and was able to get upgraded without losing data.

src/classes/OPP_SendAcknowledgmentBTN_TEST.cls
RadSFDeveloper: This has been overlooked, it will be fixed. Thanks.

src/classes/EP_ManageEPTemplate_CTRL.cls
davidhabib: doc edits looks good.  thanks.

src/classes/HH_Container_LCTRL.cls
davidhabib: the finally is fine.

src/classes/STG_InstallScript.cls
davidhabib: Thanks for the info @dospromptman !  glad to hear it was intentional.  @RadSFDeveloper can you just add a quick comment in the code that we explicitly intend for a failure in the settings update to abort the upgrade.  thanks.

src/classes/PSC_ManageSoftCredits_CTRL.cls
RadSFDeveloper: I believe the readonly and auto-number fields are not relevant in this method since they are not used. However, now I see a potential of the issue due to method changes (not sure if this information is applicable in triggers though). Will make the change to preserve the readonly and auto-number fields.

src/classes/STG_InstallScript.cls
RadSFDeveloper: The comment about re-throwing exception to stop the package installation script has been added. Please let me know if needs to be changed more.

src/classes/STG_InstallScript.cls
RadSFDeveloper: The return statement has been removed.

PR #2789, opened 2017-10-06
Rollups: Save applied effective exchange rate on Payment and/or Allocation records

force2b:
For organizations with Advanced Currency Management enabled, add support to allow storing the effective exchange rate used to convert the Payment and/or Allocation amount to the corporate currency in a custom (user defined) field on the Payment and/or Allocation objects. This works around the limitation of Salesforce does natively supporting dated exchange rates on custom objects; allowing Customers to leverage the stored rate to more accurately convert the Payment/Allocation amount for reporting.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -45,6 +45,14 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
     /** @description Allocations settings. */
     public static Allocations_Settings__c settings = UTIL_CustomSettingsFacade.getAllocationsSettings();
 
+    /** @description  Retrieve and cache the household settings the first time this class is loaded only */
+    public static final npo02__Households_Settings__c householdSettings = UTIL_CustomSettingsFacade.getHouseholdsSettings();
```
njjc: UTIL_CustomSettingsFacade already caches, do you think this extra reference is useful?

src/objects/npo02__Households_Settings__c.object
```diff
@@ -2,6 +2,18 @@
 <CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
     <enableEnhancedLookup>false</enableEnhancedLookup>
     <fields>
+        <fullName>Field_for_Effective_Exchange_Rate__c</fullName>
```
force2b: @smarovich FYI - New custom settings field and related inline help text. These PR is being merged into the larger roll-ups branch so it won't appear in the release notes for quite a while.

src/classes/ALLO_Allocations_TDTM.cls
force2b: Isn't this how we currently retrieve settings for use within a class, or is just done as `UTIL_CustomSettingsFacade.getHouseholdsSettings().Some_Field_Value__c`?

njjc: Yep, you can reference it directly.

force2b: For consistency, should the class be updated to use that method for 'householdSettings' as well as the `settings` (AllocationSettings) from above? Also, I find it more readable to have a static var such as householdSettings referenced in the code. If CustomSettingsFacade is already caching the data, then I think this static var is a by reference version of that Sobject.

src/objects/npo02__Households_Settings__c.object
smarovich: Hi @force2b , I'd make some minor changes to the Help text:

The API name of a field, on both the OppPayment__c and Allocation__c objects, where the exchange rate used for conversion is stored. The referenced field must be a Number field with a minimum of 6 decimals. This field is only used if Advanced Currency Management is enabled.

Also, for the field name, would "Exchange Rate for Conversion" work?

force2b: Thanks for the quick review. Two things:
* The help text is limited to 255 characters. I can make that fit by shortening some of the terms (such as OppPayment__c to just Payment and Management to just mgt).
* I'm hesitant on describing it as "exchange rate for conversion". The field is for their use. It may be used for conversion, it may be used for something else (though conversion is a likely use case). It will contain the exchange rate, but how they use it is sort of up to the customer. I hope that makes sense.

smarovich: Thanks, Mike. How about this for the help text: 
The API name of a field, on both the OppPayment__c and Allocation__c objects, where the exchange rate is stored. It must be a Number field with a minimum of 6 decimals. This field is used only if Advanced Currency Management is enabled.

As for the field name, could we keep it simple and call it "Exchange Rate" or "Exchange Rate Field"?

force2b: Thanks @smarovich. Two thoughts:
* What do you think about a label of "Custom Exchange Rate Field"?
* For the inline help text, how about this small change: "... where the exchange rate is *optionally* stored. ... "

smarovich: Your suggestions sound good to me, Mike! Thanks!

force2b: Great. Thank you.

PR #2790, opened 2017-10-06
Feature/kofi healthcheck account rt

kofijohnson:
A new Health Check is added to check if the selected Record Type for the Account Model is the default Account Record Type. This check runs only for the logged in user. If the selected Record Type for the Account Model is the default Account Record Type for the logged in user, the Health Check shows a failed result.

# Critical Changes

# Changes

- There is a new Health Check that verifies the validity of the Record Type selected for your chosen Account Model (in NPSP Settings, under People | Account Model). The Health Check fails if the selected Record Type for the Account Model:
  - Is the same as the default Account Record Type for the logged in user
  - Is invalid

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -321,7 +321,7 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
                         Label.healthLabelAccountDefaultRT, 
                         statusError, 
                         String.format(Label.healthDetailsAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()}),
-                        String.format(Label.healthSolutionAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()})
+                        Label.healthSolutionAccountDefaultRTIssue
                     );
                     return;
                 }
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -163,7 +163,7 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
         verifyAccountModel(); 
         verifyAccountData();
         verifyAccountData2();
-        verifyAccountDefaultRecordType();
+        verifyAccountModelRecordType();
         verifyContactData();
         verifyNoOrphanedHouseholds();
         verifyNoOrphanedOne2OneAccounts();
@@ -291,19 +291,19 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
     }
 
     /*********************************************************************************************************
-    * @description Verifies HouseHold RecordType is not the default RecordType for the current User.
+    * @description Verifies HouseHold and One To One RecordTypes are not the default RecordType for the current User.
     */
     @TestVisible
-    private void verifyAccountDefaultRecordType() {
+    private void verifyAccountModelRecordType() {
         try {
            
-            Id npspHouseholdRecordType = UTIL_CustomSettingsFacade.getOrgContactsSettings().npe01__HH_Account_RecordTypeID__c;
-            Id npspOneToOneRecordType = UTIL_CustomSettingsFacade.getOrgContactsSettings().npe01__One_to_One_RecordTypeID__c;
-            Map<Id,Schema.RecordTypeInfo> accountRecordTypeInfosById = Account.SObjectType.getDescribe().getRecordTypeInfosById();
+            Id npspHouseholdRecordType = STG_Panel.stgService.stgCon.npe01__HH_Account_RecordTypeID__c;
+            Id npspOneToOneRecordType = STG_Panel.stgService.stgCon.npe01__One_to_One_RecordTypeID__c;
+            Map<Id,Schema.RecordTypeInfo> accountRecordTypeInfosById = UTIL_Describe.getObjectDescribe('Account').getRecordTypeInfosById();
 
-            //Check if the Household and the One To One RecordRecord Type in the Settings is valid. 
-            //This should never happen, unless the User manually updated the Account Model Record Type in the Custom Settings.
-            if(
+            //Check if the Household and the One To One Record Type in the Settings are valid. 
+            //This should never happen, unless the User manually updates the Account Model Record Type in the Custom Settings.
+            if (
                 (npspHouseholdRecordType != null && accountRecordTypeInfosById.keySet().contains(npspHouseholdRecordType) == false) ||
                 (npspOneToOneRecordType != null && accountRecordTypeInfosById.keySet().contains(npspOneToOneRecordType) == false)
             ) {
@@ -311,25 +311,23 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
                 return;
             }
 
-            //Check the Household and One To One Record Type in the Settings and create a Failed Test if it is the Default one on the Account.
-            for(Schema.RecordTypeInfo accountRecordTypeInfo : accountRecordTypeInfosById.values()) {
-                if(
+            //Check the Household and One To One Record Type in the Settings and create a failed test if it is the Default one on the Account.
+            for (Schema.RecordTypeInfo accountRecordTypeInfo : accountRecordTypeInfosById.values()) {
+                if (
                     accountRecordTypeInfo.isDefaultRecordTypeMapping() &&
-                    (
-                        accountRecordTypeInfo.getRecordTypeId() == npspHouseholdRecordType ||
-                        accountRecordTypeInfo.getRecordTypeId() == npspOneToOneRecordType
-                    )
+                    (accountRecordTypeInfo.getRecordTypeId() == npspHouseholdRecordType || accountRecordTypeInfo.getRecordTypeId() == npspOneToOneRecordType)
                 ) {
                     createDR(
                         Label.healthLabelAccountDefaultRT, 
                         statusError, 
-                        Label.healthDetailsAccountDefaultRTIssue, 
-                        String.format(Label.healthDetailsAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()})
+                        String.format(Label.healthDetailsAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()}),
+                        String.format(Label.healthSolutionAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()})
                     );
                     return;
                 }
             }
 
+            //Create a success test
             createDR(Label.healthLabelAccountDefaultRT, statusSuccess, null, null);
         }
         catch(Exception ex) {
--- a/src/classes/STG_PanelHealthCheck_CTRL.cls
+++ b/src/classes/STG_PanelHealthCheck_CTRL.cls
@@ -163,6 +163,7 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
         verifyAccountModel(); 
         verifyAccountData();
         verifyAccountData2();
+        verifyAccountDefaultRecordType();
         verifyContactData();
         verifyNoOrphanedHouseholds();
         verifyNoOrphanedOne2OneAccounts();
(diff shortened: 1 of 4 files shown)


src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,53 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
 
     /*********************************************************************************************************
+    * @description Verifies HouseHold RecordType is not the default RecordType for the current User.
+    */
+    @TestVisible
+    private void verifyAccountDefaultRecordType() {
+        try {
+           
+            Id npspHouseholdRecordType = UTIL_CustomSettingsFacade.getOrgContactsSettings().npe01__HH_Account_RecordTypeID__c;
+            Id npspOneToOneRecordType = UTIL_CustomSettingsFacade.getOrgContactsSettings().npe01__One_to_One_RecordTypeID__c;
```
davidhabib: the settings page already has the settings cached, so you should use:
STG_Panel.stgService.stgCon.npe01__HH_Account_RecordTypeID__c

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,53 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+            Map<Id,Schema.RecordTypeInfo> accountRecordTypeInfosById = Account.SObjectType.getDescribe().getRecordTypeInfosById();
+
```
davidhabib: we have a class, UTIL_Describe, that caches describe info so that within the execution context we avoid multiple describe calls.  take a look at what's in that class, and consider whether you should leverage it instead, or add a getRecordTypeInfosById method to that class.

src/classes/STG_PanelHealthCheck_TEST.cls
```diff
@@ -213,4 +213,257 @@ public with sharing class STG_PanelHealthCheck_TEST {
...
                 'There should be two warning messages generated for this\n' + hc.listDR[0]);
     }
+
+    /*******************************************************************************************************
+    * @description In this test scenario, we set a non default Household Record Type in the Contact And Orgs Settings.
+    * This should create a Success Detect Result the Test.
+    */ 
+    static testmethod void test_VerifyAccountDefaultRecordType_NonDefaultRT_In_SettingsHHRT() {
+        //Check if the Org has more than 1 Record Type for Account.
+        //In case there is only one RecordType, it will be the default, therefore we should run this test.
```
davidhabib: i think your comment meant to say "therefore we should NOT run this test"?

src/classes/STG_PanelHealthCheck_TEST.cls
```diff
@@ -213,4 +213,257 @@ public with sharing class STG_PanelHealthCheck_TEST {
...
+        Schema.RecordTypeInfo accountDefaultRecordTypeInfo;
+        if(Account.SObjectType.getDescribe().getRecordTypeInfos().size() < 2) {
+            return;
+        }
+        //Get a non default Account RecordType for the current user.
+        for(Schema.RecordTypeInfo accountRecordTypeInfo : Account.SObjectType.getDescribe().getRecordTypeInfos()) {
+            if(accountRecordTypeInfo.isDefaultRecordTypeMapping() == false) {
+                accountDefaultRecordTypeInfo = accountRecordTypeInfo;
+                break;
```
davidhabib: i'm a bit confused.  seems like this loop is trying to find a non-default account record type, but you are storing it in a variable that says it is the default?

src/classes/STG_PanelHealthCheck_TEST.cls
```diff
@@ -213,4 +213,257 @@ public with sharing class STG_PanelHealthCheck_TEST {
...
+    }
+
+    /*******************************************************************************************************
+    * @description In this test scenario, we set a non default One To One Record Type in the Contact And Orgs Settings.
+    * This should create a Success Detect Result the Test.
+    */
+    static testmethod void test_VerifyAccountDefaultRecordType_NonDefaultRT_In_SettingsOneToOneRT() {
+        //Check if the Org has more than 1 Record Type for Account.
+        //In case there is only one RecordType, it will be the default, therefore we should run this test.
+        Schema.RecordTypeInfo accountDefaultRecordTypeInfo;
```
davidhabib: NOT run the test (in the comment)

src/classes/STG_PanelHealthCheck_TEST.cls
```diff
@@ -213,4 +213,257 @@ public with sharing class STG_PanelHealthCheck_TEST {
...
+        if(Account.SObjectType.getDescribe().getRecordTypeInfos().size() < 2) {
+            return;
+        }
+        //Get the non default Account RecordType for the current user.
+        for(Schema.RecordTypeInfo accountRecordTypeInfo : Account.SObjectType.getDescribe().getRecordTypeInfos()) {
+            if(accountRecordTypeInfo.isDefaultRecordTypeMapping() == false) {
+                accountDefaultRecordTypeInfo = accountRecordTypeInfo;
+                break;
```
davidhabib: same confusing on the variable name.

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,53 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+            //Check if the Household and the One To One RecordRecord Type in the Settings is valid. 
+            //This should never happen, unless the User manually updated the Account Model Record Type in the Custom Settings.
+            if(
+                (npspHouseholdRecordType != null && accountRecordTypeInfosById.keySet().contains(npspHouseholdRecordType) == false) ||
```
davidhabib: our current coding convention, https://sites.google.com/a/salesforce.com/sfdo-products-team/artifacts-docs/technical-development-resources/cumulus-coding-conventions, has a space between if (...) and for (...), as does the majority (though not all) of our code base.

src/classes/STG_PanelHealthCheck_TEST.cls
```diff
@@ -213,4 +213,257 @@ public with sharing class STG_PanelHealthCheck_TEST {
...
+        System.assertEquals(Label.healthDetailsAccountDefaultRTInvalid, panelHealthCheckController.listDR[0].strDetails);
+        System.assertEquals(Label.healthSolutionAccountDefaultRTInvalid, panelHealthCheckController.listDR[0].strSolution);
+    }
+
+    /*******************************************************************************************************
+    * @description In this test scenario, we set an Invalid value of One To One Record Type in the Contact And Orgs Settings.
+    * In this case, we use a Contact Record Type as an invalid value for the One to One Record Type.
+    * This should create a Failed Detect Result the Test.
+    */ 
+    static testmethod void test_VerifyAccountDefaultRecordType_InvalidRT_In_SettingsOneToOneRT() {
```
davidhabib: for all of the tests you added, you have two sets (one for HH, one for 1:1), which are virtually identical except for which property you set.  might have been nice to use a shared test routine for each set, and just pass in a boolean specifying whether to set the HH rt or the 1:1 rt.

src/labels/CustomLabels.labels
```diff
@@ -1988,6 +1988,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>There are {0} One-to-One Accounts who have no Contacts.</value>
     </labels>
     <labels>
+        <fullName>healthDetailsAccountDefaultRTInvalid</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>healthDetailsAccountDefaultRTInvalid</shortDescription>
+        <value>The Record selected for the Account Model is not valid.</value>
```
davidhabib: you should get the strings reviewed by @smarovich , but this one should definitely say the Record TYPE selected...

davidhabib:
@kofijohnson @force2b I know this got assigned for review with Michael, but I was just curious to see Kofi's first pull request!  Overall, was good Kofi, with a few minor issues I found.  Michael, you should still review and be the one to handle reviewing any additional changes and the merge.  Thanks!

force2b:
@davidhabib I know you spent some time on this story with Kofi and have background in the health check pages, so I appreciate you going through this.

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,53 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
     }
 
     /*********************************************************************************************************
+    * @description Verifies HouseHold RecordType is not the default RecordType for the current User.
```
force2b: This method appears to do more than just verify that the HH_Account RT is not the default. That should be reflected in the method comment.

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,53 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+                    accountRecordTypeInfo.isDefaultRecordTypeMapping() &&
+                    (
+                        accountRecordTypeInfo.getRecordTypeId() == npspHouseholdRecordType ||
+                        accountRecordTypeInfo.getRecordTypeId() == npspOneToOneRecordType
+                    )
+                ) {
+                    createDR(
+                        Label.healthLabelAccountDefaultRT, 
+                        statusError, 
+                        Label.healthDetailsAccountDefaultRTIssue,
```
force2b: This label has a {0} merge field, but that's not being replace here. The result is an error message such as the following:
image

force2b: Also, I don't think message explains what the actual error is. Maybe something like - "The default Household Record Type has been set to "Organization", which has been set as the default Record Type for the current user". I'm not sure this is all that clear either, but that's where @smarovich can help.

src/labels/CustomLabels.labels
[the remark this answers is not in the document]
smarovich: @kofijohnson Agreed that it should say Record Type. Question though: Is it that the record type isn't valid at all or that it doesn't work with the Account Model set for this org?

davidhabib:
@kofijohnson what are you doing updating pull requests at 4am Saturday morning!!!  only @judisohn  is allowed to work crazy off hours when sane people are sleeping!

src/labels/CustomLabels.labels
```diff
@@ -1988,6 +1988,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>There are {0} One-to-One Accounts who have no Contacts.</value>
     </labels>
     <labels>
+        <fullName>healthDetailsAccountDefaultRTInvalid</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>healthDetailsAccountDefaultRTInvalid</shortDescription>
+        <value>The Record Type selected for the Account Model isn&apos;t valid.</value>
```
force2b: Should this say "The Record Type selected for the *current* Account Model ..." (adding the word "current"). Otherwise, the message - at least to me - appears to be a bit ambiguous. What do you think @smarovich? Same for the label below (`healthDetailsAccountDefaultRTIssue`)

src/labels/CustomLabels.labels
```diff
@@ -2403,6 +2419,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>Account Data</value>
     </labels>
     <labels>
+        <fullName>healthLabelAccountDefaultRT</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>healthLabelAccountDefaultRT</shortDescription>
+        <value>Account Model Record Type</value>
```
force2b: I'm not sure about this title for the check. @smarovich Would you think about "Account Model -- Selected Record Type" instead?

smarovich:
@kofijohnson  I wrote up a quick release note for the new health check. How does this sound:
There is a new Health Check that verifies the validity of the Record Type selected for your chosen Account Model (in NPSP Settings, under People | Account Model). The Health Check fails if the selected Record Type for the Account Model:

- Is the same as the default Account Record Type for the logged in user
- Is invalid

src/labels/CustomLabels.labels
smarovich: That works for me, Mike!

src/labels/CustomLabels.labels
smarovich: I suggested this title and would like to stick with it.

kofijohnson:
That looks good @smarovich! Thanks

src/classes/STG_PanelHealthCheck_CTRL.cls
```diff
@@ -290,6 +291,51 @@ public without sharing class STG_PanelHealthCheck_CTRL extends STG_Panel {
...
+            //Check the Household and One To One Record Type in the Settings and create a failed test if it is the Default one on the Account.
+            for (Schema.RecordTypeInfo accountRecordTypeInfo : accountRecordTypeInfosById.values()) {
+                if (
+                    accountRecordTypeInfo.isDefaultRecordTypeMapping() &&
+                    (accountRecordTypeInfo.getRecordTypeId() == npspHouseholdRecordType || accountRecordTypeInfo.getRecordTypeId() == npspOneToOneRecordType)
+                ) {
+                    createDR(
+                        Label.healthLabelAccountDefaultRT, 
+                        statusError, 
+                        String.format(Label.healthDetailsAccountDefaultRTIssue, new List<String>{accountRecordTypeInfo.getName()}),
```
force2b: Rather than make the same `String.format()` call twice, I suggest assigning this result to a var and then use that var in the `createDR()` call.

src/labels/CustomLabels.labels
```diff
@@ -1988,6 +1988,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <shortDescription>healthDetailsAccountDefaultRTInvalid</shortDescription>
+        <value>The Record Type selected for the Account Model isn&apos;t valid.</value>
+    </labels>
+    <labels>
+        <fullName>healthDetailsAccountDefaultRTIssue</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>healthDetailsAccountDefaultRTIssue</shortDescription>
+        <value>The Record Type, {0}, is selected for the Account Model (in NPSP Settings) and is used as your profile&apos;&apos;s default Account Record Type.</value>
```
force2b: This is the other label to add the word "current" before "Account Model".

src/labels/CustomLabels.labels
```diff
@@ -2699,6 +2723,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <shortDescription>healthSolutionAccountDefaultRTInvalid</shortDescription>
+        <value>Go to &lt;strong&gt;People | Account Model&lt;/strong&gt; and select a valid Record Type.</value>
+    </labels>
+    <labels>
+        <fullName>healthSolutionAccountDefaultRTIssue</fullName>
+        <categories>Health Check</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>healthSolutionAccountDefaultRTIssue</shortDescription>
+        <value>The Record Type selected for the current Account Model should not be used as the default Account Record Type for any Profile in your org. Using the same Record Type could cause various data issues.</value>
```
force2b: @smarovich Question for you. Do you think the the phrase `in your org` in the above label will be clear to admins who see this message, or should that be spelled out as something like `in this organization` or `in this Salesforce Organization`?

smarovich: Great point, @force2b . I use the term "org" too freely sometimes... "in this Salesforce organization" is a great alternative.

PR #2792, opened 2017-10-10
Add "System" prefix to the Label in the Controller and the Test Class

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


kofijohnson:
@davidhabib this PR is to fix the issue with the Build. I've added "System" prefix to the Labels in the Controller and Test Class.

PR #2798, opened 2017-10-16
Feature/kofi query ldv optimize err

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
- Apex Class UTIL_Profile
- Apex Class UTIL_Profile_TEST

# Deleted Metadata

### DIFF
--- a/src/classes/ERR_Notifier.cls
+++ b/src/classes/ERR_Notifier.cls
@@ -83,25 +83,26 @@ public class ERR_Notifier {
     }
 
     /*******************************************************************************************************
-    * @description Get the active System Admininstrators emails.
-    * @return List<String> list of active System Administrators emails
+    * @description Get the active System Administrators' emails.
+    * @return List<String> Emails
     */
     @TestVisible
     private static List<String> getSystemAdminEmails() {
-        Set<String> result = new Set<String>();
-        List<Id> systemAdminProfileIds = UTIL_Profile.getInstance().getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR);
-        if (systemAdminProfileIds.isEmpty() == false) {
-            for (User systemAdminRecipient : [
+        Set<String> emails = new Set<String>();
+        List<Id> sysAdminProfileIds = UTIL_Profile.getInstance().getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR);
+
+        if (sysAdminProfileIds.isEmpty() == false) {
+            for (User usr : [
                     SELECT Email 
                     FROM User 
-                    WHERE ProfileId 
-                    IN :systemAdminProfileIds 
+                    WHERE ProfileId IN :sysAdminProfileIds 
                     AND IsActive = true
             ]) {
-                result.add(systemAdminRecipient.Email);
+                emails.add(usr.Email);
             }
         }
-        return new List<String>(result);
+
+        return new List<String>(emails);
     }
 
     /*******************************************************************************************************
--- a/src/classes/ERR_Notifier.cls
+++ b/src/classes/ERR_Notifier.cls
@@ -84,7 +84,7 @@ public class ERR_Notifier {
 
     /*******************************************************************************************************
     * @description Get the active System Admininstrators emails.
-    * @return Set<String> list of active System Administrators emails
+    * @return List<String> list of active System Administrators emails
     */
     @TestVisible
     private static List<String> getSystemAdminEmails() {
--- a/src/classes/ERR_Notifier.cls
+++ b/src/classes/ERR_Notifier.cls
@@ -37,8 +37,6 @@
 public class ERR_Notifier {
 
     public static NotificationOptions notificationOptions = new NotificationOptions();
-
-    @TestVisible private static Set<String> emailNotificationRecipients;
     
     /*******************************************************************************************************
     * @description Inner class with the 3 types of notification options: user, chatter group, and all system
@@ -76,23 +74,34 @@ public class ERR_Notifier {
                     for (User u : useremaillist)
                         sendList.add(u.email);
                 } else if(errorNotifRecipient == NotificationOptions.sysAdmins) {
-                    Set<Id> systemAdminProfileIds = UTIL_Profile.getInstance().getProfileIds('System Administrator');
-                    if (systemAdminProfileIds.isEmpty() == false) {
-                        for (User systemAdminRecipient : [
-                                SELECT Email 
-                                FROM User 
-                                WHERE ProfileId 
-                                IN :systemAdminProfileIds 
-                                AND IsActive = true
-                        ]) {
-                            sendList.add(systemAdminRecipient.Email);
-                        }
-                    }
+                    sendList.addAll(getSystemAdminEmails());
                 }
-                emailNotificationRecipients = new Set<String>(sendList);
+
                 sendEmailNotifications(context, sendList);
             }
         }
+    }
+
+    /*******************************************************************************************************
+    * @description Get the active System Admininstrators emails.
+    * @return Set<String> list of active System Administrators emails
+    */
+    @TestVisible
+    private static List<String> getSystemAdminEmails() {
+        Set<String> result = new Set<String>();
+        List<Id> systemAdminProfileIds = UTIL_Profile.getInstance().getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR);
+        if (systemAdminProfileIds.isEmpty() == false) {
+            for (User systemAdminRecipient : [
+                    SELECT Email 
+                    FROM User 
+                    WHERE ProfileId 
+                    IN :systemAdminProfileIds 
+                    AND IsActive = true
+            ]) {
+                result.add(systemAdminRecipient.Email);
+            }
+        }
+        return new List<String>(result);
     }
 
     /*******************************************************************************************************
--- a/src/classes/ERR_Notifier.cls
+++ b/src/classes/ERR_Notifier.cls
@@ -38,8 +38,7 @@ public class ERR_Notifier {
 
     public static NotificationOptions notificationOptions = new NotificationOptions();
 
-    //This is created only for Testing Purpose. It contains the number of email addresses.
-    @TestVisible private static Set<String> emailNotificationSendersList;
+    @TestVisible private static Set<String> emailNotificationRecipients;
     
     /*******************************************************************************************************
     * @description Inner class with the 3 types of notification options: user, chatter group, and all system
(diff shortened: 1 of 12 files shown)


src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+@isTest
+private class UTIL_Profile_TEST {
+	private static Integer BULK_CALL = 200;
+
+	/**************************************************************************************************************************
+    @description Test multiple calls of the method GetInstance() to make sure the UTIL_Profile instance is constructed only once.
+    verify: The query (Get Profiles) should run only once.
+    ***************************************************************************************************************************/   
+    private static testMethod void testGetInstanceSingleton() {
+    	Test.startTest();
```
RadSFDeveloper: No need to enclose the whole method in start/stopTest(). It is not required here, or if it is to be used, it can be written before and after `UTIL_Profile profileUtil = UTIL_Profile.getInstance();`.

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+* @author Salesforce.org
+* @date 2017
+* @group Utilities
+* @description Unit tests for UTIL_Profile class
+*/
+@isTest
+private class UTIL_Profile_TEST {
+	private static Integer BULK_CALL = 200;
+
+	/**************************************************************************************************************************
```
RadSFDeveloper: The indentation for this and few other files is inconsistent and need to be changed to four spaces.

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+    @description Test multiple calls of the method GetInstance() to make sure the UTIL_Profile instance is constructed only once.
+    verify: The query (Get Profiles) should run only once.
+    ***************************************************************************************************************************/   
+    private static testMethod void testGetInstanceSingleton() {
```
RadSFDeveloper: Let's write test method names to indicate what is being tested. For example, here we are testing that `Profile` sObject is queried once when `UTIL_Profile.getInstance()` is called multiple times, so one of the test names can be: `testGetInstanceQueriesOnceOnMultipleCalls()` (this is just an example).

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+    	}
+    	//The GetInstance should not call the Constructor of UTIL_Profile anymore.
+    	System.assertNotEquals(null, profileUtil, 'The GetInstance should have the same instance of UTIL_Profile');
+    	System.assertEquals(1, Limits.getQueries(), 'The query count should not change');
+    	Test.stopTest();
+    }
+
+    /**************************************************************************************************************************
+    @description Test the GetProfileIds.
+    verify: The result returned by the GetProfileIds should is correct.
```
RadSFDeveloper: The expected result might need more details. For example:

> verify: GetProfileIds returns all Profile Ids for the specified Profile Name

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,110 @@
...
+    		);
+    		for (Profile currentProfile : orgProfilesByName.get(currentProfileName)) {
+    			System.assert(profileUtil.getProfileIds(currentProfileName).contains(currentProfile.Id));
+    		}
+    	}
+    }
+
+    /*********************************************************************************************************
+    * @description Get a Map of All Profiles, grouped by Name
+    * @return Map of Profiles, grouped by Name
```
RadSFDeveloper: The `@return` should contain the data type and then optional description. For example: 

` * @return Map<String, List<Profile>> Profiles grouped by Name`

Also, note the space between data types in the Map.

src/classes/UTIL_Profile.cls
```diff
@@ -0,0 +1,92 @@
...
+	/*********************************************************************************************************
+    * @description Constructor. The Constructor is private because we are using the Singleton Pattern.
+    * The UTIL_Profile object needs to be constructed only from the static method getInstance
+    */
+	private UTIL_Profile() {
+		List<Profile> profileRecords = [
+			SELECT Name
+			FROM Profile
+		];
+		profilesByName = getProfilesByName(profileRecords);
```
RadSFDeveloper: No need to separate Profile SOQL from the `getProfilesByName`, ie the query can be moved into the `getProfilesByName`.  The method can be moved just after the constructor so related methods are grouped together in the class (for better readability).
 
Moreover, DB queries or any heavy-duty work is not recommended in constructors; however, since this is a Singleton and the constructor is private, it can stay as it is.

src/classes/ERR_Notifier.cls
```diff
@@ -37,6 +37,9 @@
 public class ERR_Notifier {
 
     public static NotificationOptions notificationOptions = new NotificationOptions();
+
+    //This is created only for Testing Purpose. It contains the number of email addresses.
+    @TestVisible private static Set<String> emailNotificationSendersList;
```
RadSFDeveloper: Let's extract the code that needs to be unit-tested in a `@TestVisible` method instead of using a var just for testing purposes.

src/classes/ERR_Notifier_TEST.cls
```diff
@@ -0,0 +1,85 @@
...
+
+        List<User> allSystemAdmin = [
+            SELECT Email
+            FROM User
+            WHERE Profile.Name = 'System Administrator'
+            AND IsActive = TRUE
+        ];
+        Set<String> allSystemAdminEmails = getEmails(allSystemAdmin);
+        System.assert(ERR_Notifier.emailNotificationSendersList != null, 'The Email List should not be null');
+        //The number emails list from ERR_Notifier should be equal to the List of System Admininstrator emails
```
RadSFDeveloper: Typo in the comment. Also, since comments can easily become out of sync with the code, let's use them only when required. 

> Comments should say things that the code cannot say for itself.

src/classes/ERR_Notifier_TEST.cls
```diff
@@ -0,0 +1,85 @@
...
+        System.assertEquals(allSystemAdminEmails.size(), ERR_Notifier.emailNotificationSendersList.size());
+        for (String sysAdminEmail : ERR_Notifier.emailNotificationSendersList) {
+            System.assert(allSystemAdminEmails.contains(sysAdminEmail));
+        }
+    }
+
+    /*********************************************************************************************************
+    * @description Get a List of All emails from Users
+    * @param List of the users
```
RadSFDeveloper: `@param` is missing the var name, ie `* @param users List of the Users`

src/classes/ERR_Notifier.cls
kofijohnson: Good idea @RadSFDeveloper! Here is the logic we want to test here. 
1- Set the Error Notification Settings to "All System Admin",
2- Call the method "sendErrorNotifications", 
3- Confirm the message will be sent the correct Recipients (active System Admins emails)
As discussed, extracting and unit test the piece of code does not really test the scenario above. Your idea is good but to make it work, we will need a refactor of the whole method, which is beyond the scope of this ticket, and will need more analysis.
I think having the variable emailNotificationSendersList does not harm here, it helps to test the logic without the need to refactor the whole method. 
But it is an open discussion... @davidhabib any suggestion?

src/classes/UTIL_Profile.cls
```diff
@@ -0,0 +1,90 @@
...
+    /** @description static instance of the current class (UTIL_Profile). It is used as the instance in a Singleton context*/
+	private static UTIL_Profile utilProfile;
+	/** @description Map of the Profiles grouped by Name*/
+	private Map<String, List<Profile>> profilesByName;
+
+	/*********************************************************************************************************
+    * @description get the instance of the current class
+    * @return UTIL_Profile The instance of the current class.
+    */
+	public static UTIL_Profile getInstance() {
```
force2b: @kofijohnson What is the value of using a singleton instance of this class versus just using static vars and methods? I'm not seeing any class properties that class methods would take actions on other than the `profilesByName` map would could just be a static map referenced by static methods.

src/classes/ERR_Notifier.cls
force2b: @kofijohnson Is the only purpose for this `@TestVisible` var to reference in the Unit Test to validate that one or more Users were found to send a notification to?

src/classes/ERR_Notifier.cls
```diff
@@ -74,11 +76,20 @@ public class ERR_Notifier {
                     for (User u : useremaillist)
                         sendList.add(u.email);
                 } else if(errorNotifRecipient == NotificationOptions.sysAdmins) {
-                    List<User> sysadminlist = [SELECT email FROM User WHERE User.Profile.Name = 'System Administrator' AND isActive = true];
-                    for (User u : sysadminlist) {
-                        sendList.add(u.email);
+                    Set<Id> systemAdminProfileIds = UTIL_Profile.getInstance().getProfileIds('System Administrator');
```
RadSFDeveloper: As discussed this morning, let's extract the code in this `else if` statement into a separate method that is testable. Even though `emailNotificationRecipients` variable makes `sendErrorNotifications()` testable, it is best to remove it since it is used in tests only.

The Work Item task is to retrieve Profile Ids for a specified Profile Name in order to optimize the User query, so using `UTIL_Profile` is a proper way instead re-writing the query this time and every time Profile Ids are required.

RadSFDeveloper: Instead of hard-coding `'System Administrator'`, let's create a new constant in `UTIL_Profile` and refactor test classes and one class to use the constant. Also, see if those referencing classes can be refactored to use  `UTIL_Profile.getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR)` or`UTIL_Profile.getSysAdminProfileIds()`?

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+    	UTIL_Profile profileUtil = UTIL_Profile.getInstance();
+    	Test.stopTest();
+
+    	System.assertNotEquals(null, profileUtil, 'The GetInstance should return an instance of UTIL_Profile');
+
+    	for (String profileName : orgProfilesByName.keySet()) {
+            List<Id> profileIds = profileUtil.getProfileIds(profileName);
+            Set<Id> currentProfileIds = new Map<Id, Profile>(orgProfilesByName.get(profileName)).keySet();
+
+    		System.assertEquals(
```
RadSFDeveloper: Indentation issue

src/classes/UTIL_Profile_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+
+    // Helpers
+    /////////////
+
+    /*********************************************************************************************************
+    * @description Get a Map of all Profiles grouped by Name
+    * @return Map<String, List<Profile>> Map of Profiles grouped by Name
+    */
+	private static Map<String, List<Profile>> getProfilesByName() {
+		Map<String, List<Profile>> result = new Map<String,List<Profile>>();
```
RadSFDeveloper: Indentation issue in the method

PR #2809, opened 2017-10-20
Rollups: CMT Filter/Rollup Selector and CMT Filter Service Classes

force2b:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/objects/Rollup__mdt.object
```diff
@@ -131,6 +106,7 @@
     </fields>
     <fields>
         <fullName>Integer__c</fullName>
+        <description>Integer value to indicate how many years ago or days back to roll-up for</description>
```
force2b: @njjc I think we're missing the corresponding field on this object to indicate the "Yearly Operation Type". The writeup indicate a pick-list with values for All Time/Yearly/N Days Back (https://salesforce.quip.com/DcFZApxrEY3I).

njjc: Yep, let's add it.

src/classes/UTIL_Namespace.cls
```diff
@@ -98,17 +98,20 @@ public class UTIL_Namespace {
...
     * If you ever deploy this package as unmanaged, this routine will do nothing!
     * @param str string that contains 0 or more token names 
     * @return same string with token names, namespace prefixed, if required.
     ********************************************************************************************************/
     public static string StrAllNSPrefix(string str) {
         if (getNamespace() == '') return str;
-        Pattern pat = Pattern.Compile('[a-zA-z0-9]*__c');
-        Matcher match = pat.matcher(str);
-        return match.replaceAll(getNamespace()+'__$0');
+        Pattern pat1 = Pattern.Compile('[a-zA-z0-9]*__[cr]');
```
njjc: I believe these can be combined into a single  ```Pattern.Compile('[a-zA-z0-9]*__(?:c|r|mdt)');```

Note: I haven't tested the above regex :)

src/classes/CMT_FilterRuleEval_SVC.cls
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * @description Static map of instantiated Filter_Rule__mdt by FilterGroupId
+     */
+    private static Map<Id, List<CMT_FilterRule>> mapOfFilterRulesByGroupId = new Map<Id, List<CMT_FilterRule>>();
+
+    /**
+     * @description Static cache of filter group eval results by a concatenation of detail row Ids
+     */
+    private static Map<String, Boolean> cachedFilterEvalResults = new Map<String, Boolean>();
```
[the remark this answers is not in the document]
force2b: Yes. evaluateFilterGroup() accepts a list of SObjects, so each of those has to be evaluated against the conditions in a given group of individual rules. The cache holds a concatenation of all the FilterGroupId plus all the individual Sobject record Ids. If that concatenated string isn't in the cache, then each rule is evaluated to determine if filtergroup passes or not. That result is stored within the cache. If that same group of records is passed into this method with the same FilterGroupId, the cache will hold the result - meaning isIncluded() won't be called again for each of those records.

src/classes/CMT_FilterRuleUI_SVC.cls
```diff
@@ -0,0 +1,247 @@
...
+public class CMT_FilterRuleUI_SVC {
+
+    // ============================================================================================
+    // STATIC METHODS FOR HANDLING UI INTERACTION AND OVERALL PROCESSING
+    // ============================================================================================
+
+    /**
+     * @description Converts the enum strings into friendly strings for the pick-lists.
+     */
+    public static Map<CMT_FilterRule.FilterOperation, String> getFilterOperations() {
```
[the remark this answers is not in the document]
force2b: True. We can also separate the labels from the picklist API name so that label changes don't require modifying the underlying code and enums.

src/classes/CMT_UnitTestData_TEST.cls
```diff
@@ -0,0 +1,233 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Custom Metadata Types
+* @description Because CMT objects cannot be instantiated or inserted in a Unit Test, the methods in this
+* class build a JSON String in the format of the specified MDT Object with all necessary field values and references.
+*/
+@isTest
+public class CMT_UnitTestData_TEST {
```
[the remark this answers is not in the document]
force2b: Yes. The Id's are arbitrary in the serialized data, so I've set it up to use `getNewRecordId()` to generate a new unique record id that we can use to represent anything - filterGroupId, FilterRuleId, etc.

src/classes/UTIL_Namespace.cls
force2b: Worked perfectly! Regex is definitely my weakness!

PR #2824, opened 2017-10-31
Update Campaign Household Mailing List report button to work in both Lightning and Classic

### LINKED ISSUE #2324: Household Mailing List button on Campaign does not function correctly in Lightning Experience
The button correctly adds 'Household Duplicate' to the Campaign Member Status, however the report one lands on is filtered based on Campaign ID (which returns no results) rather than correctly selecting the Campaign. Workaround is to edit the report configuration to see results.

force2b:
# Critical Changes

# Changes

The old "NPSP Campaign Household Mailing List" report included with the NPSP Households package is being replaced with a new version of the report that renders correctly in both the Salesforce Classic and Lightning user interfaces. The old report will still exist in customer orgs, but can be deleted. The new "NPSP Campaign Household Mailing List V2" report should not be deleted. Note that If a customer organization has modified the old report, those changes will need to be applied to the new report.

# Issues Closed

Fixes #2324 

# New Metadata

Reports/Nonprofit Edition Reports/NPSP Campaign Household Mailing List V2

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -3046,7 +3046,7 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <language>en_US</language>
         <protected>false</protected>
         <shortDescription>hhCmpDedupeNoReport</shortDescription>
-        <value>Could not find report. Create a report with the Report Unique Name of  &apos;NPSP_Campaign_Household_Mailing_List&apos;. NOTE: Report type should be Campaigns with Contacts. Filter for Member Status does not contain &apos;Duplicate&apos;.</value>
```
force2b: @smarovich and @judisohn Change to the label used to display an error when the report cannot be found

src/reports/Nonprofit_Edition_Reports/NPSP_Campaign_Household_Mailing_List_V2.report
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: @judisohn The new report I created. Effectively identical to the old report that was in the Households package with a couple key changes:
* New report folder (needed because I can't install a new report from this package in a folder created by the Households package).
* Report field filter used instead Scope.

I kept the same report label because ideally the old report label will be changed by the Household package push, though I'm not 100% sure of that. I wasn't able to find this documented clearly.

judisohn: @force2b I just pushed this to a scratch org and for some reason, the report is not showing with V2 in the name. I think I'd prefer the folder be named something other than 'Nonprofit Edition' reports since we never call it 'Nonprofit Edition' anywhere. Maybe rename 'NPSP Health Check' to 'NPSP Settings Reports' and it will include health check and anything that comes out of 'system' type reports. @smarovich?

image

As a result, had to guess which one: 

image

Also, we'll need to change the text below to indicate the new report name.

force2b: Yep. That makes sense based on what I posted just before this. We can't rename either during the deploy, so let's remove both the report and folder from this package and use "V2" in the report label for the new report in the Cumulus package. What do you think?

davidhabib: please don't go renaming the NPSP Health Check folder.  I believe we have multiple messages and doc that refer to specific reports in that specific folder.

force2b: This PR shouldn't touch that folder. I'm just removing the Campaign Household Mailing List report from the "Nonprofit Edition Reports" folder, which I think is actually deployed in a different package.

judisohn: Okay @davidhabib - I would just prefer we didn't call it 'Nonprofit Edition Reports' since this would be a good time to rename that folder. But if not, fine.

force2b: @judisohn and @davidhabib We can't rename the folder that was installed with npe01, but the new report can either go into a new folder (that's what I did here) or we might be able to deploy into the already existing 'Nonprofit Edition Reports' folder that was installed with npe01. I was having difficulty getting IntelliJ to pull down the report metadata when it was in the npe01 folder, but that should work since the Households package (npo02) was deploying the report into the `npe01__Nonprofit_Edition_Reports` folder

src/labels/CustomLabels.labels
smarovich: @force2b I'd make some very minor changes: 
Could not find the report. Create a report and save it with the **Report Unique Name** of  &apos;NPSP_Campaign_Household_Mailing_List&apos;. Use the Campaigns with Contacts report type and add a filter for Member Status does not contain &apos;Duplicate&apos;.

force2b: @smarovich Good suggestion, though the report unique name as to be NPSP_Campaign_Household_Mailing_List_V2. I'm open to changing the report unique name or label, but it has to be different than the previous report's unique name.

smarovich: Ah ok. That report name is fine with me. I think we should keep it similar to the original name.  

Could not find the report. Create a report and save it with the Report Unique Name of 'NPSP_Campaign_Household_Mailing_List_V2'. Use the Campaigns with Contacts report type and add a filter for Member Status does not contain 'Duplicate'.

force2b: Thanks, though one more item. I had added a fragment for "CampaignId equals blank and" as part of the filters. This is a new requirement for the report filtering for it to work correctly (see below). Can I put that back in so it reads as follows?

`Could not find the report. Create a report and save it with the Report Unique Name of 'NPSP_Campaign_Household_Mailing_List_V2'. Use the Campaigns with Contacts report type and add filters for CampaignId equals blank and Member Status does not contain 'Duplicate'.`

smarovich: That looks great!

njjc:
I'm wondering if it's worth holding off merge on this feature until next sprint when we do the underlying package update, which will then allow deleting the old deprecated report. I hate having to tell admins "Update this new report with stuff from the old report, then wait a few weeks before you can delete the old one." Thoughts @force2b @judisohn @smarovich ?

judisohn:
@njjc Guess what? There is nothing I can see that prevents the user from deleting the report right now. I fired up a scratch org with the managed release, went to a managed report, hit Delete and it's gone. 

image

image

So I think we should do it in the underlying only to prevent it in new orgs, but it doesn't seem to matter for existing customers.

image

image

@force2b @smarovich

force2b:
Yeah. That makes sense based on this doc. Reports are subscriber deletable even if the report isn't removed from the original package. https://developer.salesforce.com/docs/atlas.en-us.packagingGuide.meta/packagingGuide/packaging_packageable_components.htm#packaging_packageable_components

njjc:
Wonderful, maybe let's add to the release notes that users can safely delete the old one once the new one has been updated, if applicable?

judisohn:
@smarovich See comments above. Maybe 'Note that If a customer organization has modified the old report, those changes will need to be applied to the new report. Once the changes are applied to the new report and NPSP Settings updated in People > Households to reference the new v2 report, the old "NPSP Campaign Household Mailing List" can be safely deleted.' What do you think?

force2b:
@judisohn and @smarovich The settings won't need to be updated by the User. I added code to the post install script to do that for them.

PR #2825, opened 2017-11-01
BDI Performance Logging and Enhancements

davidhabib:
This pull request includes all of the various optimizations made to improve our performance when running BDI in XL LDV orgs.  Optimizations include:
- caching DI records to their Contact
- deferring Household Naming until all Contacts have been updated/created.
- turning off Opportunity Rollups
- optimizing Contact soql to filter just on Name IN :setNames when both Firstname and Lastname are included in the criteria
- optimizing Opportunity & Payment soql to avoid unnecessary sorting
- optimizing all object updates to only update them if they actually change

This pull request also moves all code dealing with Contacts from BDI_DataImportService to BDI_ContactService, as well as refactoring most of the Contact code to use methods that can be used for both Contact1 as well as Contact2, thus avoiding duplication of code.

Also, this pull request introduces a new utility class, UTIL_PerfLogger, for doing performance measurements.  It detects dynamically whether PerfLog__c exists in the org, and runs with or without it being installed.  BDI also makes use of this class by extending it with its own to manage BDI specific data in the PerfLog object.

# Critical Changes

# Changes
- The NPSP Data Importer has been optimized to handle quickly processing large batches of data in organizations with millions of records.

# Issues Closed

# New Metadata
- Run_Opportunity_Rollups_while_Processing__c field added to DataImportBatch__c
- Run_Opportunity_Rollups_while_Processing__c field added to Data_Import_Settings__c

# Deleted Metadata


davidhabib:
@njjc this pull request is ready to review. please let me know if you want me to go over any of it with you.  thanks!

src/classes/BDI_ContactService.cls
```diff
@@ -0,0 +1,1095 @@
...
+    * the Contacts to the database.
+    * @param strCx specifies which contact to update and save; either Contact1 or Contact2
+    * @return Set<Id> the set of DI Id's that were updated
+    */
+    private Set<Id> updateAndSaveContacts(string strCx) {
+        Set<Id> setDIIdUpdated = new Set<Id>();
+        Integer iCx = strCx == 'Contact1' ? 1 : 2;
+
+        // handle multiple di's referring to the same existing contact
+        set<string> setDIKeyUpsert = new set<string>();
```
njjc: Would you mind capitalizing the first letter of collections and class names? It highlights them differently and makes readability easier.

src/classes/BDI_ContactService.cls
```diff
@@ -0,0 +1,1095 @@
...
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description adds a non-null Double to the provided set
+    * @param dbl the Double value
+    * @param setDbl the set to add to
+    * @return void
+    */
+    private void addDoubleToSet(Double dbl, Set<Double> setDbl) {
```
njjc: Any reason not to combine these three methods into ```void addObjectToSet(Object obj, Set<Object> setObj)```?

src/classes/UTIL_PerfLogger.cls
```diff
@@ -0,0 +1,229 @@
...
+    private Long tmStart;
+    private Long tmEnd;
+    private Map<String, PerfLog> mapPerfLogByComponentLabel;
+
+    /*******************************************************************************************************
+    * @description constructor, that starts the logger's timer
+    * @param countRecords the number of records that will be analyzed
+    */
+    public UTIL_PerfLogger(Integer countRecords) {
+        tmStart = System.now().getTime();
```
njjc: I've read [somewhere](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_platform_cache_best_practices.htm) that System.currentTimeMillis() is preferred.

src/classes/BDI_ContactService.cls
davidhabib: done

src/classes/BDI_ContactService.cls
davidhabib: I had tried that, but it complains about signatures not matching.  So I went with the explicit types, rather than casting.  This was important since a few callers are getting their data from a obj.get() call that needs to be cast to the correct underlying type.

src/classes/UTIL_PerfLogger.cls
davidhabib: not sure if that will give different numbers than System.now.getTime(), but I made the change.

PR #2829, opened 2017-11-02
Feature/joy dmloptions allowsave

### LINKED ISSUE #2383: Enhance NPSP Data Imports to pass dml.DuplicateRuleHeader.AllowSave = true;
Even if your rule says to just warn, but allow the save to go thru, any code doing inserts (apex, triggers) must explicitly pass in a flag to the Salesforce API to allow the save. (dml.DuplicateRuleHeader.AllowSave = true;) 

Unfortunately, NPSP does not pass in this flag in most scenarios. we should see if we can add in this flag when inserting contacts and accounts for data import.

hub thread: https://powerofus.force.com/0D580000035esUw

### LINKED ISSUE #2475: Investigate widely implementing DMLOptions.DuplicateRuleHeader.AllowSave
Duplicate rules set to "allow save" are still blocked in DML operations unless the DMLOptions.DuplicateRuleHeader.AllowSave is specifically set for each DML operation.

Investigate applying this header to all processDML methods, and also to manual insert/update calls to Contact and Account objects that are likely to have this type of rule enabled.

[From the HUB.](https://powerofus.force.com/0D58000003Cjyxb)

joy-sfdo:
# Critical Changes

# Changes
NPSP now respects the "Allow" value in the Action On Create or Action On Edit fields in a Duplicate Rule. To use this feature, you must enable the Respect Duplicate Rule Settings checkbox in NPSP Settings, under System Tools | Error Notifications. To find out if a duplicate rule has these fields set to "Allow," from Setup, search for Duplicate Rules. Click on a rule name and look for the Action On Create or Action On Edit fields.

# Issues Closed
Fixes #2475
Fixes #2383

# New Metadata
- UTIL_DMLService.cls (& meta-xml)
- UTIL_DMLService_TEST.cls (& meta-xml)
- Error_Settings__c.Respect_Duplicate_Rule_Settings__c (Custom Field)
- stgHelpRespectDuplicateRuleSettings (Label)

# Deleted Metadata


force2b:
@joy-sfdo The new DMLService class looks very interesting. Is the intent that all npsp dml operations should now be funneled through this service class? If now, is there a doc that describes under what circumstances it's better to use the new DMLService class versus a standard dml operation?

joy-sfdo:
Hi @force2b -- For consistency, it would be great to have all DML operations funneled through this service or another means. For now, this class is set to handle duplicates in the following areas: Account, Contact, Address, and BDI. Great idea regarding the doc, I'll write one up and we can discuss it at either a dev hangout or Arch meeting. If we decide to move in that direction, we can definitely expand the scope and usage of this service.

force2b:
Thanks. I'm thinking to what we're building for Rollups, which will update both the Account and Contact objects.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and defaults the DMLOptions optAllOrNone to false 
+    *   and DuplicateRuleHeader.AllowSave to true.
+    * @param recordToInsert The sObject to add to the org.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult insertRecord(sObject recordToInsert) {
+        Database.DMLOptions dmlOptions = new Database.DMLOptions();
+        
+        dmlOptions.optAllOrNone = ALL_OR_NONE_FALSE;
```
davidhabib: shouldn't this default to true, since that's the behavior of doing a normal insert sobject call?

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and defaults the DMLOptions optAllOrNone to false 
+    *   and DuplicateRuleHeader.AllowSave to true.
+    * @param recordsToInsert The sObjects to add to the org.
+    * @return List<Database.SaveResult>
+    **********************************************************************************************************/
+    public static List<Database.SaveResult> insertRecords(List<sObject> recordsToInsert) {
+        Database.DMLOptions dmlOptions = new Database.DMLOptions();
+        
+        dmlOptions.optAllOrNone = ALL_OR_NONE_FALSE;
```
davidhabib: shouldn't this default to true since that's the behavior of doing a normal insert listSObject call?

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordToInsert The sObject to add to the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult insertRecord(sObject recordToInsert, Database.DMLOptions dmlOptions) {
+        dmlOptions.optAllOrNone = ALL_OR_NONE_FALSE;
```
davidhabib: if the user passes in dmlOptions, I don't think we should be overriding their choice of optAllOrNone.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordsToInsert The sObjects to add to the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return List<Database.SaveResult>
+    **********************************************************************************************************/
+    public static List<Database.SaveResult> insertRecords(List<sObject> recordsToInsert, Database.DMLOptions dmlOptions) {
+        dmlOptions.optAllOrNone = ALL_OR_NONE_FALSE;
```
davidhabib: if the user passes in dmlOptions, I don't think we should be overriding their choice of optAllOrNone.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordToInsert The sObject to add to the org.
+    * @param allOrNone The optional parameter that specifies whether the operation allows partial success.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult insertRecord(sObject recordToInsert, Boolean allOrNone, Database.DMLOptions dmlOptions) {
+        dmlOptions.optAllOrNone = allOrNone;
```
davidhabib: why have both a version of insert that takes dmlOptions, and one that takes dmlOptions and a boolean that will get set into dmlOptions?  Seems unnecessary and a bit confusing.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordsToInsert The sObjects to add to the org.
+    * @param allOrNone The optional parameter that specifies whether the operation allows partial success.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return List<Database.SaveResult>
+    **********************************************************************************************************/
+    public static List<Database.SaveResult> insertRecords(List<sObject> recordsToInsert, Boolean allOrNone, Database.DMLOptions dmlOptions) {
```
davidhabib: why have both a version of insert that takes dmlOptions, and one that takes dmlOptions and a boolean that will get set into dmlOptions? Seems unnecessary and a bit confusing.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,306 @@
...
+    }
+ 
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class update method and defaults the DMLOptions optAllOrNone to false 
+    *   and DuplicateRuleHeader.AllowSave to true.
+    * @param recordToInsert The sObject to update in the org.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult updateRecord(sObject recordToInsert) {
```
davidhabib: I won't repeat my same comments that I put on the insert methods to the other dml methods, since they would be the same.  thanks.

davidhabib: all of the update methods named their arg as record(s)ToInsert.  should be record(s)ToUpdate.

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -479,7 +479,7 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
             individual.npe01__SYSTEMISINDIVIDUAL__c = true;
             //individual.Type = CAO_Constants.BUCKET_ORGANIZATION_TYPE;
             individual.npe01__SYSTEM_AccountType__c = CAO_Constants.BUCKET_ORGANIZATION_TYPE;
-            insert individual;
+            UTIL_DMLService.insertRecord(individual, true);
```
davidhabib: seems like UTIL_DMLService.insertRecord(individual) would have been the logical replacement for straight insert calls, but as I call out in DMLService, I think it should default to allOrNone to be consistent with straight insert calls.

src/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -726,9 +726,7 @@ public with sharing class CON_ContactMerge_CTRL {
             
             System.SavePoint sp = Database.setSavepoint();
             try {
-                Database.DMLOptions dml = new Database.DMLOptions();
-                dml.DuplicateRuleHeader.AllowSave = true;
-                Database.SaveResult sr = Database.update(master, dml);
+                Database.SaveResult sr = UTIL_DMLService.updateRecord(master);
```
davidhabib: @mpusto @joy-sfdo here is a scenario (contact merge) where we were previously forcing AllowSave to true.  Are we sure we want to now change this to be based off the setting?  I'm concerned that contact merges will now fail that used to succeed, since we defaulting our new setting Respect_Duplicate_Rule_Settings__c to false.

src/classes/HH_Container_LCTRL.cls
```diff
@@ -347,16 +347,13 @@ public with sharing class HH_Container_LCTRL {
             TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.HH, true);
             
             // allow potential duplicates when rules to warn are turned on.
-            Database.DMLOptions dml = new Database.DMLOptions();
-            dml.DuplicateRuleHeader.AllowSave = true;
-
-            list<Database.SaveResult> listSR = Database.insert(listInsert, dml);
+            List<Database.SaveResult> listSR = UTIL_DMLService.insertRecords(listInsert);
```
davidhabib: @mpusto @joy-sfdo another example scenario where we were explicitly allowing the save regardless of duplicate warning, in ManageHouseholdUI.  do we want this behavior changed to use the setting, which is defaulting to false?

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,699 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Utilities
+* @description Service Class to handle DML actions.
```
davidhabib: probably should say this is the test class!

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,699 @@
...
+        Contact con = new Contact(FirstName = 'Blue', LastName = 'Cloud', AccountId = acct.Id); 
+                 
+        Test.startTest();
+        UTIL_DMLService.insertRecord(con);
+        Test.stopTest();
+
+        List<Contact> retrievedContacts = [SELECT Id, FirstName, LastName, AccountId 
+                                           FROM Contact];
+
+        if (!retrievedContacts.isEmpty()) {
```
davidhabib: shouldn't this test assert that there was a retrieved contact, rather than gracefully handling it if it failed?  that's the whole point of testing your new insertRecord() method.

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,699 @@
...
+                                                        new Contact(FirstName = 'Astro', LastName = 'Cloud', AccountId = acct.Id) };
+                         
+        Test.startTest();
+        UTIL_DMLService.insertRecords(contacts);
+        Test.stopTest();
+
+        List<Contact> retrievedContacts = [SELECT Id, FirstName, LastName, AccountId 
+                                           FROM Contact];
+
+        if (!retrievedContacts.isEmpty()) {
```
davidhabib: same comment as previous test.

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,699 @@
...
+        if (strTestOnly != '*' && strTestOnly != 'testRecordsInsertWithAllOrNone') return;
+        
+        Account acct = new Account(Name = 'DMLOptionsAccount');
+        insert acct;
+
+        List<Contact> contacts = new List<Contact> { new Contact(FirstName = 'Blue', LastName = 'Cloud', AccountId = acct.Id),
+                                                        new Contact(FirstName = 'Astro', LastName = 'Cloud', AccountId = acct.Id) };
+                         
+        Test.startTest();
+        UTIL_DMLService.insertRecords(contacts, false);
```
davidhabib: if the goal is to really ensure your cover correctly handles the allOrNone flag, then I'd expect to see a test that inserts two contacts, one without a lastname (which should get an error).  if allorNone=true, then 0 contacts saved.  if allOrNone=false, then 1 contact saved.

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,699 @@
...
+        Account acct = new Account(Name='Blue Cloud', BillingCity='San Francisco');
+        UTIL_DMLService.insertRecord(acct);
+        List<Account> retrievedAccts = [SELECT Id, Name FROM Account];
+        if (!retrievedAccts.isEmpty()) {
+            System.assertEquals('Blue Cloud', retrievedAccts[0].Name);
+        }
+        
+        Account dupeAcct = new Account(Name='Blue Cloud', BillingCity='San Francisco');
+        Database.SaveResult sr = UTIL_DMLService.insertRecord(dupeAcct);
+        if (!sr.isSuccess()) {
```
davidhabib: i assume there is no way to create/turn on duplicate alerts in a test, and so the best we can do is see if it failed or not?

src/labels/CustomLabels.labels
```diff
@@ -5324,6 +5324,15 @@ This process may take some time, but you can close this page and the process wil
         <value>The custom Relationships fields you do not want synchronized between the two contacts in a relationship pair. Control or command click to select more than one field.</value>
     </labels>
     <labels>
+        <fullName>stgHelpRespectDuplicateRuleSettings</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>stgHelpRespectDuplicateRuleSettings</shortDescription>
```
davidhabib: we've been asking that all new labels actually have a useful description for the localizers.  it's very unfortunate that salesforce automatically pastes the label into the description, which led to most of our old labels not having useful text.

src/classes/UTIL_DMLService.cls
joy-sfdo: Yes, that is the normal behavior of the insert so I'll make the change as well as update the calling methods.

src/classes/UTIL_DMLService.cls
joy-sfdo: Same as above.

src/classes/UTIL_DMLService.cls
joy-sfdo: I agree. In the UTIL_DMLService doc, I'll make sure to call out that along with any dmlOptions a dev might set, they will also need to set the optAllOrNone.

src/classes/UTIL_DMLService.cls
joy-sfdo: Same as above.

src/classes/UTIL_DMLService.cls
joy-sfdo: The boolean was added along with the dmlOptions, to ensure that it would be set. However, the usage can be communicated differently.

src/classes/UTIL_DMLService.cls
joy-sfdo: Same as above.

src/classes/UTIL_DMLService.cls
joy-sfdo: Thanks for calling out these copy and paste items.

joy-sfdo: Same as above.

src/labels/CustomLabels.labels
joy-sfdo: Good to know. I'll update the label.

src/classes/CON_ContactMerge_CTRL.cls
joy-sfdo: @davidhabib -- Chatted with @mpusto, we'll consider this a special case so I'll set the AllowSave to true.

src/classes/HH_Container_LCTRL.cls
joy-sfdo: @davidhabib -- Same as Contact Merge, will set AllowSave to true.

src/classes/UTIL_DMLService_TEST.cls
joy-sfdo: Sounds good.

src/classes/UTIL_DMLService_TEST.cls
joy-sfdo: Yes, that's correct.

joy-sfdo:
@davidhabib -- Thanks for the review. Code updates completed.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,239 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordToInsert The sObject to add to the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult insertRecord(sObject recordToInsert, Database.DMLOptions dmlOptions) {
+        dmlOptions.DuplicateRuleHeader.AllowSave = retrieveRespectDuplicateRuleSettings();
```
davidhabib: I think this should be AllowSave |= retrieveRespectDuplicateRuleSettings().  in other words, if the caller explicitly set AllowSave to true in dmlOptions (such as in Contact Merge), I don't think we want the global setting to turn it off.  But if the global setting is on, then it makes sense to force it on.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,239 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class insert method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordsToInsert The sObjects to add to the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return List<Database.SaveResult>
+    **********************************************************************************************************/
+    public static List<Database.SaveResult> insertRecords(List<sObject> recordsToInsert, Database.DMLOptions dmlOptions) {
+        dmlOptions.DuplicateRuleHeader.AllowSave = retrieveRespectDuplicateRuleSettings();
```
davidhabib: don't override caller having set AllowSave to true

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,239 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class update method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordToUpdate The sObject to update in the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return Database.SaveResult
+    **********************************************************************************************************/
+    public static Database.SaveResult updateRecord(sObject recordToUpdate, Database.DMLOptions dmlOptions) {
+        dmlOptions.DuplicateRuleHeader.AllowSave = retrieveRespectDuplicateRuleSettings();
```
davidhabib: don't override caller having set AllowSave to true

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,239 @@
...
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class update method and sets the DMLOptions optAllOrNone accordingly 
+    *   and defaults DuplicateRuleHeader.AllowSave to true.
+    * @param recordsToUpdate The sObjects to update in the org.
+    * @param dmlOptions The optional parameter that specifies additional data for the transaction.
+    * @return List<Database.SaveResult>
+    **********************************************************************************************************/
+    public static List<Database.SaveResult> updateRecords(List<sObject> recordsToUpdate, Database.DMLOptions dmlOptions) {
+        dmlOptions.DuplicateRuleHeader.AllowSave = retrieveRespectDuplicateRuleSettings();
```
davidhabib: don't override caller having set AllowSave to true

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,565 @@
...
+        if (strTestOnly != '*' && strTestOnly != 'testRespectDuplicateRuleSettingsForAccounts') return;
+        
+        UTIL_CustomSettingsFacade.getErrorSettingsForTests(new Error_Settings__c(Respect_Duplicate_Rule_Settings__c = false));
+        Error_Settings__c errorSettings = UTIL_CustomSettingsFacade.getErrorSettings();
+        UTIL_DMLService.insertRecord(errorSettings);
+        
+        Account acct = new Account(Name='Blue Cloud', BillingCity='San Francisco');
+        UTIL_DMLService.insertRecord(acct);
+        List<Account> retrievedAccts = [SELECT Id, Name FROM Account];
+        if (!retrievedAccts.isEmpty()) {
```
davidhabib: when would we not expect this insert to succeed?

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,565 @@
...
+            }
+        } catch (DmlException e) {}
+
+        errorSettings.Respect_Duplicate_Rule_Settings__c = true;
+        UTIL_DMLService.updateRecord(errorSettings);
+         
+        UTIL_DMLService.insertRecord(dupeAcct);
+        retrievedAccts = [SELECT Id, Name FROM Account];
+        if (!retrievedAccts.isEmpty()) {
+            System.assertEquals(NUMBER_OF_RECORDS_TWO, retrievedAccts.size());
```
davidhabib: why wouldn't we assert this always succeeds?

src/classes/UTIL_DMLService.cls
joy-sfdo: That's a very good point.

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -424,17 +424,17 @@ public with sharing class LD_LeadConvertOverride_CTRL {
                     ID rtIdAcccount = ACCT_IndividualAccounts_TDTM.rtIdForNewAccount();
                     if (rtIdAcccount != null) 
                         myAccount.put('RecordTypeID', rtIdAcccount);
-                    myAccount.npe01__SYSTEMISINDIVIDUAL__c = true;
-                    myAccount.npe01__One2OneContact__c = con.Id;
-                    update myAccount;
+                        myAccount.npe01__SYSTEMISINDIVIDUAL__c = true;
```
davidhabib: the if on line 425 was only for 426, not 427-429.  so rather than indenting 427-429, let's add the recommended bracing on our If.  thanks!

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,256 @@
...
+        return Database.update(recordsToUpdate, dmlOptions);
+    }
+
+
+    /*********************************************************************************************************
+    * @description Checks if the Allow Save property is set and uses it accordingly.
+    * @return Boolean
+    **********************************************************************************************************/
+    private static Boolean useAllowSaveProperty(Database.DMLOptions dmlOptions) {
+        return (dmlOptions.DuplicateRuleHeader.AllowSave != null);
```
davidhabib: good solution!  respect true or false, and only overwrite with the setting if null!

src/classes/BDI_DataImportService.cls
```diff
@@ -962,7 +962,7 @@ global with sharing class BDI_DataImportService {
 
         // now update the Contacts to create their Affiliations
         if (mapConIdToConUpdate.size() > 0) {
-            update mapConIdToConUpdate.values();
+            UTIL_DMLService.updateRecords(mapConIdToConUpdate.values());
```
davidhabib: This class has lots of other places where we insert/update/upsert Accounts, Addresses, Campaigns, and CampaignMembers.  do you think some or all of those should also be going thru UTIL_DMLService to respect the new setting?  I'm guessing Accounts is the likely one to be problematic (ie, the user may have specified a rule).  But maybe worth doing for all for consistency?  what are your thoughts?  thanks.

src/classes/BDI_Donations.cls
```diff
@@ -186,7 +186,7 @@ public class BDI_Donations {
         }
         
         // insert our new Opps
-        list<Database.SaveResult> listSR = database.insert(listOppInsert, false);
+        list<Database.SaveResult> listSR = UTIL_DMLService.insertRecords(listOppInsert, false);
```
davidhabib: on line 201, we do database.update(listOppUpdate, false).  shouldn't this also go thru UTIL_DMLService?

davidhabib: also in updatePayment() we do a database.update(listPmtUpdate, false).  do we care about this one?  I doubt users will put duplicate rules on payments, but who knows!  best to go thru UTIL_DMLService for consistency?

src/classes/BDI_DataImportService.cls
davidhabib: Also the class BDI_ContactService does all of BDI's inserting and updating of Contacts.  I assume we would want all of its saving methods to go thru UTIL_DMLService.

src/classes/UTIL_DMLService.cls
joy-sfdo: Thanks!

src/classes/LD_LeadConvertOverride_CTRL.cls
joy-sfdo: Braces added and indentation fixed. Thanks!

src/classes/BDI_DataImportService.cls
joy-sfdo: Added the code for consistency. Good call.

src/classes/BDI_Donations.cls
joy-sfdo: Added the code for consistency. Good call.

src/classes/BDI_DataImportService.cls
joy-sfdo: Thanks for pointing out this class.

src/classes/UTIL_DMLService.cls
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+
+    /*********************************************************************************************************
+    * @description Wraps the Database class upsert method.
+    * @param recordToUpsert The sObject to upsert in the org.
+    * @return Database.UpsertResult
+    **********************************************************************************************************/
+    public static Database.UpsertResult upsertRecord(sObject recordToUpsert) {
+        return Database.upsert(recordToUpsert);
```
davidhabib: don't your new upsert methods need to set the allowSave property just like the insert and update covers?

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,697 @@
...
+
+
+    /*********************************************************************************************************
+    * @description Tests the NPSP Setting Respect Duplicate Rule Settings for duplicate Accounts
+    *   set to false 
+    * verify:
+    *   Duplicate Accounts detected and allowed when the Respect Duplicate Rule Setting set to true.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testRespectDuplicateRuleSettingsForAccounts() {
```
davidhabib: it would be good to also get in testing that covers the new upsert methods.  rather than copying this whole test and using upsert instead of update, maybe make a common routine that takes a parameter on whether to do upsert vs update, and then have to thin testmethod covers that use it, one doing update, one doing upsert.

src/classes/UTIL_DMLService_TEST.cls
```diff
@@ -0,0 +1,697 @@
...
+
+
+    /*********************************************************************************************************
+    * @description Tests the NPSP Setting Respect Duplicate Rule Settings for duplicate Contacts
+    *   Records To Be Saved set to true 
+    * verify:
+    *    Duplicate Contacts detected and allowed when the Respect Duplicate Rule Setting set to true
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testRespectDuplicateRuleSettingsForContacts() {
```
davidhabib: consider doing the same with this routine, so we can test both updates and upserts.

src/classes/UTIL_DMLService.cls
joy-sfdo: Database.upsert(...) does not support Database.DMLOptions. This basically wraps the upsert method for consistency.

davidhabib:
@joy-sfdo weird, but in my DE org against your branch, I get two test failures:

System.AssertException: Assertion Failed: Expected: 1, Actual: 2
Class.UTIL_DMLService_TEST.testRespectDuplicateRuleSettingsForAccounts: line 614, column 1

System.AssertException: Assertion Failed: Expected: 1, Actual: 2
Class.UTIL_DMLService_TEST.testRespectDuplicateRuleSettingsForContacts: line 673, column 1

PR #2830, opened 2017-11-02
Mark required fields on page

### LINKED ISSUE #2813: Required fields are not being marked required by default.
On pages using the new UTIL_InputField component, the required field is defaulting to false. This is overriding the native apex:inputField required tag being set automatically on Visualforce pages.

RadSFDeveloper:
Currently some required fields are not marked as required on our custom pages. If a value is not provided, an error will be raised at DB level as expected and save will fail. To fix the issue, we'll check if a field is not nillable, and if so, mark it as required. This means the required attribute on the form is still optional and the required value does not have to be passed in.

# Critical Changes

# Changes

# Issues Closed
Fixes #2813

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/components/UTIL_FormField.component
+++ b/src/components/UTIL_FormField.component
@@ -56,15 +56,19 @@
         rendered="{!($ObjectType[sObjType].fields[field].updateable
                     || $ObjectType[sObjType].fields[field].createable && sObj['Id']=='') && $ObjectType[sObjType].fields[field].type != 'boolean'}">        
         <apex:outputLabel for="inputX" styleClass="slds-form-element__label" rendered="{!showLabel == true && ftype!='boolean'}" id="lblInputField">
-            <apex:outputText rendered="{!required}" id="lblInputFieldRequired">
+            <apex:outputText rendered="{!required || NOT($ObjectType[sObjType].fields[field].nillable)}" id="lblInputFieldRequired">
                 <abbr class="slds-required" title="{!$Label.lblRequired}">*</abbr>
             </apex:outputText>
             <apex:outputText rendered="{!showLabel == true}" value="{!label}"/>
         </apex:outputLabel>
         <div class="slds-form-element__control">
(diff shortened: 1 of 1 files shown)


src/components/UTIL_FormField.component
```diff
@@ -56,15 +56,19 @@
         rendered="{!($ObjectType[sObjType].fields[field].updateable
                     || $ObjectType[sObjType].fields[field].createable && sObj['Id']=='') && $ObjectType[sObjType].fields[field].type != 'boolean'}">        
         <apex:outputLabel for="inputX" styleClass="slds-form-element__label" rendered="{!showLabel == true && ftype!='boolean'}" id="lblInputField">
-            <apex:outputText rendered="{!required}" id="lblInputFieldRequired">
+            <apex:outputText rendered="{!required || NOT($ObjectType[sObjType].fields[field].nillable)}" id="lblInputFieldRequired">
```
RadSFDeveloper: @jjbennett, the `required` attribute cannot be set to be empty because even if it is empty, null, 0, ... it is considered as `false`.  So, in order to check if the optional `required` attribute needs to be set, we need to check if the field is not `nillable`. 
If this approach does not work, the other (only) option is to change all pages using the `UTIL_FormField` to pass in required value.

jjbennett: @RadSFDeveloper Instead of setting a default on line 6, could we create an apex:variable and check if the attribute has a value and if it does use it, if it does not then use the value from nillable then use the new var instead of the required attribute?

RadSFDeveloper: @jjbennett ,  I have removed the `default` and created a new variable to check if the `required` is not set. However, since `required` is a `Boolean`, then any passed value that is not `true` is `falsey`, ie the returned value is `false`. Do you see anything wrong with the following code that does not recognize the `required` attribute is empty?

`
    <apex:variable var="isRequired"
        value="{!IF(required == null, NOT($ObjectType[sObjType].fields[field].nillable), required)}"
    />`

PR #2839, opened 2017-11-06
FIX: Custom batch size now being used for batch rollup classes

### LINKED ISSUE #2794: Scheduled Job should use Batch Size as set in NPSP Settings
[Thread.](https://powerofus.force.com/0D58000003iOM1U)

Bug?

While it appeared to me that the [code](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/RLLP_OppContactRollup_BATCH.cls#L44) [should](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/RLLP_OppRollup_UTIL.cls#L234) be using the same Batch Size as set in NPSP Settings > Donations > Donor Statistics, it appears to be using the default of 200. Running Bulk Data Processes > Rollup Donations Batch uses the correct NPSP Setting value.

Process run manually in test org with a batch size of 5:

image

Scheduled Job ran automatically a few hours later with no other change to settings:

image

Confirmed that batch size is still set to 5.

randi274:
This fix addressed a bug that was causing batch rollup classes to ignore the user's desired batch size as specified on the NPSP Settings page.

# Critical Changes

# Changes

# Issues Closed
Fixes #2794 
# New Metadata

# Deleted Metadata


src/classes/RLLP_OppContactRollup_BATCH.cls
```diff
@@ -48,7 +51,7 @@ public class RLLP_OppContactRollup_BATCH implements Database.Batchable<SObject>,
     }
     
     /*******************************************************************************************************
-    * @description Batch start method, instantiates RLLP_OppRollup_UTIL and builds query for contact rollups.
+    * @description Batch start method and builds query for contact rollups.
     ********************************************************************************************************/
     public database.Querylocator start(Database.BatchableContext bc) {
         RLLP_OppRollup_UTIL oppRollupUtil = new RLLP_OppRollup_UTIL();
```
davidhabib: since you now instantiate oppRollupUtil as a member variable on line 40 above, do you still need to do it here?

randi274: Nope, let me change that as well. I had some versioning issues with this class for some reason. Thanks!

PR #2843, opened 2017-11-09
Fixing bug with apex:variable uncovered with payments field set

### LINKED ISSUE #2842: Required Field Error when Payment Method is not the last field in the Payment Wizard Field Set
The page is requiring both the date and payment method when the payment method is not the last field in the field set on the Payment Wizard Schedule page.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
Fixes #2842 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/components/UTIL_FormField.component
+++ b/src/components/UTIL_FormField.component
@@ -22,10 +22,6 @@
     <!-- The fixup style class to apply to dependent and date fields -->
     <apex:variable var="fixup"
         value="{!IF(dependent, IF(ftype=='picklist', ' dependentPicklistFixup', ' dependentMultiPicklistFixup'),'')}"/>
-    <!-- Determine if the field will be marked as required -->
-    <apex:variable var="isRequired"
-        value="{!IF(required == true, true, NOT($ObjectType[sObjType].fields[field].nillable))}"
-    />
     <!-- The field label to be displayed -->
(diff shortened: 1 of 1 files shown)


src/components/UTIL_FormField.component
```diff
@@ -59,7 +55,7 @@
         rendered="{!($ObjectType[sObjType].fields[field].updateable
                     || $ObjectType[sObjType].fields[field].createable && sObj['Id']=='') && $ObjectType[sObjType].fields[field].type != 'boolean'}">        
         <apex:outputLabel for="inputX" styleClass="slds-form-element__label" rendered="{!showLabel == true && ftype!='boolean'}" id="lblInputField">
-            <apex:outputText rendered="{!isRequired}" id="lblInputFieldRequired">
```
force2b: @jjbennett should there be a comment in the code to let future devs know not to go back to a var for these 2 lines? Otherwise I could see someone refactoring the code to make that same change again.

jjbennett: @force2b - @RadSFDeveloper talked about this and we are thinking it might be better to start a known issues doc, there was a similar comment about not using an apex:variable but the component is so large it might easily be missed.

force2b: Great idea. We can include a note about the two Custom Labels issues as well (LEX and System.Label.).

PR #2844, opened 2017-11-09
Add option to prevent address overwrite from Cicero API address verification

### LINKED ISSUE #1686: Option to exclude an Address record from verification
Hub thread: https://powerofus.force.com/0D58000002FFDL3

### LINKED ISSUE #2538: Address Verification with Cicero: Populate Administrative Area field
[From the hub.](https://powerofus.force.com/0D58000003IsB6Y)

randi274:
# Critical Changes

# Changes

- Users can now select a checkbox on the NPSP Settings - Address page to preserve original address after verification when using the Cicero API.
# Issues Closed

- Fixes #1686
- Fixes #2538 

# New Metadata

# Deleted Metadata


src/classes/ADDR_Cicero_Validator.cls
```diff
@@ -197,13 +197,15 @@ public with sharing class ADDR_Cicero_Validator implements ADDR_IValidator {
...
                         }
-                    
-                        addr.MailingStreet__c = candidate.match_addr.split(',')[0].trim();
-                        addr.MailingCity__c = candidate.match_city;
-                        addr.MailingState__c = candidate.match_region;
-                        addr.MailingPostalCode__c = candidate.match_postal;
-                        if (!String.isBlank(lastFourPostalDigits)) {
-                            addr.MailingPostalCode__c += '-' + lastFourPostalDigits;
+
+                        if(!settings.Prevent_Address_Override__c){
```
njjc: I think we should write the API response value if the field is blank, even if the setting is checked. So in your doesOverrideAddr() test, it wouldn't update the populated fields, but it would populate the missing City from the response.

randi274: Got it, I can make changes for City and State for sure. Would there be a reasonable case where zip code or street would need to be populated if blank? There might be some random responses if that's the case. Currently taking a look at 2538 too.

njjc: Good point, probably not Street, but definitely City, State, Zip, County, and Country.

njjc:
Great job @randi274!

src/objects/Addr_Verification_Settings__c.object
```diff
@@ -60,6 +60,16 @@
         <type>Checkbox</type>
     </fields>
     <fields>
+        <fullName>Prevent_Address_Overwrite__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>Prevent Address Overwrite Checkbox</description>
+        <externalId>false</externalId>
+        <inlineHelpText>Prevents the Cicero API from overwriting Addresses during verification. If an Address field is blank, Cicero updates it.</inlineHelpText>
+        <label>Prevent Address Overwrite</label>
```
njjc: @mpusto

smarovich:
Hi @randi274, I updated the release notes to be a bit more detailed. Let me know if you would like me to make any changes. Thanks!

PR #2846, opened 2017-11-10
Feature/2826 smartystreets verified

### LINKED ISSUE #2826: SmartyStreets not properly setting Verified checkbox when verification is attempted but fails.

njjc:
# Critical Changes

# Changes

# Issues Closed
- Fixes #2826

# New Metadata

# Deleted Metadata


src/classes/ADDR_SmartyStreets_Validator.cls
```diff
@@ -143,69 +143,70 @@ public with sharing class ADDR_SmartyStreets_Validator implements ADDR_IValidato
 
         //Don't even make the address verification request to SmartyStreets if no US address is included, to save in API calls
         if(includesUSAddress) {
```
randi274: If a US Address isn't included, it doesn't set verified on the address. It could be verified and set as a foreign address.

src/classes/ADDR_SmartyStreets_Validator.cls
```diff
@@ -222,66 +223,68 @@ public with sharing class ADDR_SmartyStreets_Validator implements ADDR_IValidato
         
         //Don't even make the address verification request to SmartyStreets if no US address is included, to save in API calls
         if(includesUSAddress) {
```
randi274: Same comment re: US Address.

src/classes/ADDR_SmartyStreets_Validator.cls
```diff
@@ -222,66 +223,68 @@ public with sharing class ADDR_SmartyStreets_Validator implements ADDR_IValidato
...
-	                av.Verification_Status__c = Label.Zip_Not_Found;
-	                ERR_Handler.processError(ex, ERR_Handler_API.Context.ADDR);
-	            }
-	            ERR_Handler.processError(ex, ERR_Handler_API.Context.ADDR);
-	        }
+            try {
+                ADDR_SmartyStreets_Gateway la = new ADDR_SmartyStreets_Gateway(settings);
+                List<ADDR_SmartyStreets_Gateway.ZipResponse> responses = la.verifyZips(zrList);
+                UTIL_Debug.debug('****Gateway responses: ' + JSON.serializePretty(responses));
+                if (responses != null && !responses.isEmpty()) {
```
randi274: Why are we checking for !responses.isEmpty() here but not above? It looks like we're doing the empty check already in verifyAddresses and verifyZips.

src/classes/ADDR_SmartyStreets_Validator.cls
```diff
@@ -143,69 +143,70 @@ public with sharing class ADDR_SmartyStreets_Validator implements ADDR_IValidato
...
+                            av.Ambiguous__c = true;
+                            responseGroup.add(sr);
+                        }
+                        String response = JSON.serializePretty(responseGroup);
+                        av.API_Response__c = trimResponse(response);
+                        av.Verification_Status__c = Label.Addr_Verified;
+                    }
+                }
+                for (Address__c av : avList) {             
+                    if (av.API_Response__c == null && !String.isBlank(settings.Auth_ID__c) &&
```
randi274: Is this check on the Auth ID, Auth Token and Class a little late? Should we be doing this check at an earlier point in the code? If not, why is this required here, but not in verifyZipRecords?

src/classes/ADDR_SmartyStreets_Validator.cls
njjc: Good question. This is feels like a spec issue that we can discuss with @mpusto tomorrow.

src/classes/ADDR_SmartyStreets_Validator.cls
njjc: Agreed that this is a really weird place for this check. I'll have to dig in a bit to figure out if we should move it up or remove it altogether.

src/classes/ADDR_SmartyStreets_Validator.cls
njjc: Right, I think we can remove the check since VerifyZips does it already and returns null.

src/classes/ADDR_SmartyStreets_Validator.cls
```diff
@@ -222,66 +223,68 @@ public with sharing class ADDR_SmartyStreets_Validator implements ADDR_IValidato
...
+                List<ADDR_SmartyStreets_Gateway.ZipResponse> responses = la.verifyZips(zrList);
+                UTIL_Debug.debug('****Gateway responses: ' + JSON.serializePretty(responses));
+                if (responses != null && !responses.isEmpty()) {
+                    if(responses.size() == avList.size()) {
+                        for (integer i = 0; i < responses.size(); i++) {
+                            ADDR_SmartyStreets_Gateway.ZipResponse sr = responses[i];
+                            Address__c av = avList[i];             
+        
+                            // look for an error
+                            if (sr.status != null && sr.reason != null) {
```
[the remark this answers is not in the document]
njjc: Good catch, I'll set it.

src/classes/ADDR_SmartyStreets_Validator.cls
njjc: Yep, we can remove these checks since we'll only get to this code path if the class setting is set to this class, and we're already checking if the auth vars exist on the ADDR_SmartyStreets_Gateway lines [54](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/ADDR_SmartyStreets_Gateway.cls#L54) and [185](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/ADDR_SmartyStreets_Gateway.cls#L185). I'll remove it.

src/classes/ADDR_SmartyStreets_Validator.cls
njjc: @mpusto Randi noticed that our SmartyStreets address verification code filters out non-US addresses. It seems that when we implemented the feature years ago, they only supported US address verification, but in visiting smartystreets.com it seems like they now support international addresses. I think we could use a work item to investigate updating our code to stop filtering out international addresses with SmartyStreets, what do you think?

mpusto: Yes, I agree! It's come up in the HUB before and I answered with "well they didn't support it when we wrote the code". :)

njjc: Perfect, I'll let you add it to the backlog! Thanks.

mpusto: https://github.com/SalesforceFoundation/Cumulus/issues/2853

PR #2850, opened 2017-11-15
Fix missing field in SOQL error when User Defined Rollups are enabled

### LINKED ISSUE #2849: User Defined Rollup error in rollup batch
[Post.](https://powerofus.force.com/0D58000003mxV0w)

Replicated in test org. Create custom currency field on Opportunity, custom currency field on Contact. Create SUM UDR. Populate a record. Run batch. Error is:

First error: SObject row was retrieved via SOQL without querying the requested field: Contact.Test_Currency_Rollup__c

force2b:
Re-initialize any UDR fields before executing the query to select Accounts, Contacts, or Households for the batch rollup. This is needed because the instance of RLLP_OppRollup_UTIL created in the batch job instantiation sets a static var that is thrown away before the batch start() method is executed by Salesforce.

# Critical Changes

# Changes

# Issues Closed

Fixes #2849 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RLLP_OppAccRollup_BATCH.cls
+++ b/src/classes/RLLP_OppAccRollup_BATCH.cls
@@ -54,7 +54,8 @@ public class RLLP_OppAccRollup_BATCH implements Database.Batchable<SObject>, Sch
     * @description Batch start method and builds query for account rollups.
     ********************************************************************************************************/
     public database.Querylocator start(Database.BatchableContext bc) {
-        String acctQuery = oppRollupUtil.buildAccountQuery();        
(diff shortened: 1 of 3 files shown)


njjc:
Thanks @force2b; I take it there's no way to reproduce this issue in an apex test?

force2b:
I could not find a way to get the unit test to fail because the static var is initialized by the batch job instantiation executed within the unit test. Salesforce's internal execution sequence must be different. Christian confirms that Salesforce instantiates the class on the App Server and then stores that instance (less the static vars which are thrown away) for the eventual execution. That would match the behavior we're seeing in a live org vs. a unit test.

PR #2852, opened 2017-11-15
Revert PR #2844

cpolcyn:


src/package.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
njjc: My only concern is this standalone tag inserted by @force2b and removed by @randi274.

Any idea if it could be a problem, @cpolcyn @jlantz?

cpolcyn: Ran into this the other day in another project and I think it had no effect really. I think if build passes we let this slide.

njjc:
Make sure we don't include the release notes from the PR this is reverting in the release.

PR #2874, opened 2017-11-30
Rollups: Scheduled, Batch and Trigger rollup support

force2b:
Add scheduleable, batch, ldv-batch, queueable, and trigger rollup support. Ensure additional code coverage and functional testing in unit test classes.  Refactor the RollupBatch_SVC class into RollupProcessor_SVC and merge logic for non-ldv and ldv into a single callable method that is shared across all rollup actions. 

Functioning rollups as of this PR are Opportunity and Opportunity/Payment to the Account and Contact objects as well as Allocations to GAU. All of these support LDV/data-skew scenarios.


src/classes/CRLP_Rollup_BATCH.cls
```diff
@@ -0,0 +1,135 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Customizable Rollups Batch for non-LDV-type Rollups
+*/
+public class CRLP_Rollup_BATCH implements Database.Batchable<SObject>, Database.Stateful {
```
njjc: Really love the LDV/nonLDV approach here. Do we need Stateful for the non-LDV approach?

src/classes/CRLP_Rollup_TDTM.cls
```diff
@@ -0,0 +1,185 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Customizable Rollups for TDTM
+*
+*/
+public class CRLP_Rollup_TDTM extends TDTM_Runnable {
```
njjc: I'm not sure I grok how this trigger is handling the common insert case where an Opportunity is inserted, which triggers Payment and Allocation insert. Does this manage to enqueue the rollups only once for the parent records in that case?

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -58,33 +58,79 @@ public without sharing class UTIL_MasterSchedulableHelper {
...
-        'NPSP 03 - Opportunity Household Rollups'   => new List<String>{'0 0 23 ? * *', 'RLLP_OppHouseholdRollup_BATCH'},
-        'NPSP 04 - Opportunity Soft Credit Rollups' => new List<String>{'0 0 23 ? * *', 'RLLP_OppSoftCreditRollup_BATCH'},
-        'NPSP 05 - GAU Allocation Rollups'          => new List<String>{'0 0 23 ? * *', 'ALLO_Rollup_SCHED'},
-        'NPSP 06 - Recurring Donation Updates'      => new List<String>{'0 0 23 ? * *', 'RD_RecurringDonations_BATCH'},
-        'NPSP 07 - Seasonal Address Updates'        => new List<String>{'0 0 23 ? * *', 'ADDR_Seasonal_SCHED'},
-        'NPSP 08 - Level Assignment Updates'        => new List<String>{'0 0 4 ? * *',  'LVL_LevelAssign_SCHED'},
-        'NPSP 09 - Data Import Batch Processing'    => new List<String>{'0 0 22 ? * *', 'BDI_DataImportBatch_SCHED'}
-    };
+    @testVisible
+    public static Map<String, List<String>> defaultScheduledJobs {
```
njjc: Let's meet to discuss the approach taken here and the end goals.

src/classes/CRLP_Operation.cls
```diff
@@ -102,12 +110,18 @@ public class CRLP_Operation {
...
      */
     public static CRLP_Operation.RollupType getRollupTypeFromString(String operation) {
-        for (CRLP_Operation.RollupType rt : getRollupOperationTypes().keySet()) {
-            if (rt.name() == operation) {
-                return rt;
+        if (rollupOperationTypesByName == null) {
+            rollupOperationTypesByName = new Map<String, CRLP_Operation.RollupType>();
+            for (CRLP_Operation.RollupType rt : getRollupOperationTypes().keySet()) {
+                // Need multiple revisions of the same operation type to ensure the map.get() returns a value.
+                // Example: Best_Year != BEST_YEAR != 'Best Year' -- as a result all variations needed as the Map key
```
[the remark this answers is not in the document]
force2b: Good point. There are only 2 calls to this method, though I kind of like the idea of having the method/map handle the complicated logic of normalizing the key. This way the caller only needs to pass a value rather than have to deal with normalizing that value.

src/classes/CRLP_Query_SEL.cls
```diff
@@ -120,4 +140,52 @@
...
+
+        // Get the base query
+        String detailsQuery = CRLP_Query_SEL.getDetailsQuery(mainObject, childObjects);
+
+        // Add the appropriate filtering
+        detailsQuery += ' WHERE ' + detailKeyField + ' IN :summaryRecordIds';
+        if (objType == General_Accounting_Unit__c.SObjectType) {
+            // In the specific case of the Allocation to GAU roll-up, explicitly filter out any
+            // Allocation__c records where the Opportunity__c field is null.
+            detailsQuery += ' AND ' + SObjectType.Allocation__c.fields.Opportunity__c.getName() + ' != null';
```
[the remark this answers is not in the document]
force2b: Good question. The LDV batch uses a different method to retrieve the question. I wrote this one before the LDV batch job. However, given the current architecture, this method is never called upon to generate a query of Allocation records for a set of GeneralAccountingUnit parent records. All rollups of GAU's go directly through the CRLP_Rollup_Batch_LDV class. As a result, I've removed any references to GAU and Allocation from this method.

src/classes/CRLP_Rollup.cls
```diff
@@ -88,7 +90,7 @@ public class CRLP_Rollup {
             this.dateObject = rollupMdt.Date_Object__r.QualifiedApiName;
             this.detailObject = rollupMdt.Detail_Object__r.QualifiedApiName;
             if (rollupMdt.Detail_Field__c !=  null) {
-                DescribeFieldResult resultField = CRLP_Rollup_SVC.getSobjectFieldFromRollup(detailObject, rollupMdt.Detail_Field__r.QualifiedApiName);
+                DescribeFieldResult resultField = CRLP_Rollup_SVC.getSObjectFieldDescribe(detailObject, rollupMdt.Detail_Field__r.QualifiedApiName);
```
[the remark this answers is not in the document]
force2b: The `getSObjectFieldDescribe` method does small bit of reformatting and concatenation and then it calls `UTIL_Describe.getAllFieldsDescribe` to do the actual work

src/classes/CRLP_RollupProcessor_SVC.cls
```diff
@@ -0,0 +1,282 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Customizable Rollups service methods specific to Batch and Queueable rollup operations
```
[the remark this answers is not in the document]
force2b: Interesting. You may be right about that. I believe that all of the actions in this class are service related with the exception of the `updateChangedRecordsWithLogging` method which handles the DML and logging. Maybe this should be moved or the description needs to be improved.

src/classes/CRLP_Rollup_BATCH.cls
force2b: No. Good catch.

src/classes/CRLP_Rollup_TDTM.cls
force2b: It's not yet. I'll add a TODO to the code so it's not forgotten.

PR #2887, opened 2017-12-13
Feature/randi 2592 dml batch errors

### LINKED ISSUE #2592: DML errors in any of the batch jobs cause the batch of 200 to fail
When a dml exception occurs in the batch rollups it will cause the whole batch of 200 records to fail. Instead use the `allOrNothing` flag to allow as many record as possible to successfully update and then log the failures in the NPSP Error Log. (Should that be one log entry per failed batch with a list of Id's, or one log entry per failed record?)

randi274:
# Critical Changes

# Changes

- We've made some improvements to the way NPSP handles errors when running batches. Specifically, we now:
  - Prevent rollup batches from not updating any records when any single record update fails.
  - Record batch DML (Data Manipulation Language) errors in the Error Log in NPSP Settings. These types of errors will no longer appear on the Apex Job page, so you'll need to review the Error Log to find complete batch failure information.
  - Capture all of the individual record Id's that failed along with the error reason, so that you can better troubleshoot why records weren't processed.

# Issues Closed
#2592 

# New Metadata

# Deleted Metadata


src/classes/ERR_Handler.cls
```diff
@@ -138,62 +138,155 @@ public class ERR_Handler {
     	}
     	return errors;
     }
-        
+
```
force2b: @randi274 I could be wrong, but these look like tab indents. I think our standard is spaces for indentations, though there are some classes that for some reason have a mix of both. For any new methods in these classes, or in methods that you're updating, the indents should be changed to spaces.

src/classes/ERR_Handler.cls
```diff
@@ -138,62 +138,155 @@ public class ERR_Handler {
...
+		}
+		return errors;
+	}
+
+	/*******************************************************************************************************
+    * @description Checks if a given DML operation has succeeded depending on the type of DML result.
+    * @param dmlResults
+    * @return isSuccess Returns true if all DML operations passed successful; returns false if there was a DML exception.
+    */
+	public static boolean isDmlSuccess(Object dmlResults) {
```
force2b: Probably a bit nit-picky, but since this refers to a single instance of SaveResult, etc., should the var name here be `dmlResult` instead of `dmlResults`?

src/classes/ERR_Handler.cls
```diff
@@ -138,62 +138,155 @@ public class ERR_Handler {
...
+		return isSuccess;
+
+	}
+
+	/*******************************************************************************************************
+    * @description Retrieves errors from a DML operation depending on the type of DML result.
+    * @param dmlResults
+    * @return isSuccess Returns true if all DML operations passed successful; returns false if there was a DML exception.
+    */
+	public static List<Database.Error> getDmlResultErrors(Object dmlResults) {
```
force2b: Same as above

src/classes/ERR_Handler.cls
```diff
@@ -239,10 +332,25 @@ public class ERR_Handler {
...
+	/*******************************************************************************************************
+	* @description Creates a single error record from a Database class DML operation in a batch, future or queued job.
+	* @param message The full message of all the errors for a given job and compiled IDs.
+	* @return The error record.
+	*/
+	private static Error__c createJobError(String message) {
+		Error__c error = new Error__c();
+		error.Datetime__c = System.now();
+		error.Full_Message__c = message;
+		error.Error_Type__c = 'Job Error';
```
force2b: Do you think there's any value including parameters or method overloads to pass in values for the Context__c and/or Object__c fields?

src/classes/RLLP_OppPartialSoftCreditRollup.cls
```diff
@@ -161,7 +161,13 @@ public class RLLP_OppPartialSoftCreditRollup {
                 toUpdate.add(newC);
             }
         }
-        update toUpdate;
+        //update toUpdate;
```
force2b: The commented out code can be removed

src/classes/RLLP_OppRollup.cls
```diff
@@ -363,7 +363,15 @@ public without sharing class RLLP_OppRollup {
             }
         }         
 
-        update accountsToUpdate.values();     
+        /**Randi - Changes for Accounts - can I pass in account type?
```
force2b: Is this comment a to-do, a note, or just something that can be removed?

src/classes/RLLP_OppPartialSoftCreditRollup.cls
```diff
@@ -161,7 +161,13 @@ public class RLLP_OppPartialSoftCreditRollup {
...
+        List<Database.SaveResult> srList = Database.update(toUpdate, false);
+
+        ERR_Handler.Errors errorResult = ERR_Handler.getJobErrors(srList, toUpdate);
+        if (errorResult.errorsExist) {
+            ERR_Handler.processErrors(errorResult.errorRecords, 'Jobs');
```
force2b: It's probably better to use the `ERR_Handler_API.Context.RLLP.name()` enum instead of the hardcoded values of `'Jobs'` for all the calls to processErrors()

src/classes/ERR_Handler.cls
force2b: Also, do you think this method could be rewritten as the following? Basically, both methods take a similar action on the `dmlResult` parameter. If `getDmlResultErrors` returns 0, then I would expect that isSuccess is true (unless maybe there's a scenario where isSuccess is false and there are no errors).
`return getDmlResultErrors(dmlResult).size() == 0;`

src/classes/ERR_Handler.cls
```diff
@@ -138,62 +138,155 @@ public class ERR_Handler {
...
+    *
+    * @param dmlResults The results of a DML batch operation performed using the Database class.
+    * @param sObjects The records the DML operation was performed on.
+    * @return Errors The errors with a single list that occurred during a batch DML operation performed using the Database class.
+    */
+	public static Errors getJobErrors(List<Object> dmlResults, List<SObject> sObjects) {
+		Errors errors = new Errors();
+		List<String> errorsFound = new List<String>();
+
+		if(dmlResults != null) {
```
davidhabib: our coding standard is a space after if and for, so it should be if (...) and for (...)

src/classes/ERR_Handler.cls
randi274: I think this is one of the mixed classes, but everything above it seems to already be tabbed in (with some strange spacing too), and I've mostly just copied that. Should I reformat the file to use all tabs/spaces? Also, have we documented that somewhere, or is this something we should discuss at a team meeting?

src/classes/RLLP_OppRollup.cls
randi274: I did want to discuss if you had any suggestions on how to pass in the record type so we could include that in the error messaging, but otherwise I'll just remove this.

src/classes/ERR_Handler.cls
randi274: Cool, good to know! I'll make the necessary edits on those.

src/classes/ERR_Handler.cls
randi274: Changed var name to a singular. And agreed, I had made a note to condense those and that got list in the mix somewhere. As far as I'm aware, there shouldn't be a scenario where there's a failure without any messaging!

src/classes/ERR_Handler.cls
randi274: Check!

src/classes/ERR_Handler.cls
randi274: Do we have any docs on how the context__c field should be populated? I like the overload for the object__c, but would need to determine how best to pass that information.

src/classes/ERR_Handler.cls
force2b: I'm pretty sure that's documented, though @davidhabib also noted this in his comments.

src/classes/RLLP_OppPartialSoftCreditRollup.cls
randi274: I agree. There was another place in the code that used a different convention, and I wasn't sure which was the better way to do it so ended up passing both over!

src/classes/ERR_Handler.cls
force2b: There's a Enum with context values, though in my opinion it wouldn't hurt to refactor that at some point.

src/classes/RLLP_OppRollup.cls
force2b: Is Record Type the SObjectType? If so, you can get that from the record Id using .getSObjectType.

randi274: Right, I guess I'm just wondering if this will always be a list of the same object type? If so, I could just query the first object and call it good, otherwise I'll need to check the whole list to make sure, and maybe break if a different object type is found.

force2b: Good question. For the most part, all of our dml operations are on the same object, so I don't see any issue in just using the first Id with `getSObjectType().getDescribe().getName()`.

src/classes/ERR_Handler.cls
force2b: @randi274 Found it - https://sites.google.com/a/salesforce.com/sfdo-products-team/artifacts-docs/technical-development-resources/cumulus-coding-conventions?pli=1. The doc is a bit old. We've discussed changes to this such as always using curly-brackets for if/for blocks even if there is only a single statement within.

src/classes/ERR_Handler.cls
```diff
@@ -71,137 +71,236 @@ public class ERR_Handler {
...
+					String fullObjectErrMessage = 'There were ' + totalErrors + ' errors saving to object ' + sObjects[i].Id + '. ' + errDetailsMessage;
+
+					errorsFound.add(fullObjectErrMessage);
+				}
+			}
+			if (errorsFound.size() > 0){
+				String errorMessage = String.join(errorsFound, '\n* ');
+				errors.errorsExist = true;
+
+                String objectType = sObjects[1].getSObjectType().getDescribe().getName();
```
force2b: There are probably alternative approaches to get the ObjectType as well as match types in the `UTIL_Describe` class. My biggest concern is using getDescibe() in a for loop in the `areObjectsSameType()` method below. Plus, you could technically compare `SObjectType` instead of comparing against the object name.

src/classes/ERR_Handler.cls
force2b: Sorry, I can't tell - did you update the indentations to use spaces?

src/classes/ERR_Handler.cls
randi274: I see there's method isObjectIdThisType in UTIL_Describe, but it also requires a full getDescribe for each object. I'll trim it down to just call getSObjectType(), which should hopefully improve performance, but let me know if you think it would be better to use isObjectIdThisType. I'll also make sure I correct to using spaces :)

src/classes/RLLP_OppRollup.cls
```diff
@@ -363,7 +363,13 @@ public without sharing class RLLP_OppRollup {
             }
         }         
 
-        update accountsToUpdate.values();
```
force2b: @randi274  The 3 sections in this class and the one in RLLP_OppPartialSoftCreditRollup that do the update and error processing are all nearly identical. Is it worth creating a single method in UTIL_DMLService for `updateRecordsAndLog(List<SObject> records, ERR_Handler_API.Context context)` that each of these can just call with the list of records?

randi274: Good point, I can use that in the soft credit rollup as well!

PR #2895, opened 2017-12-28
Rollups: Soft Credits

force2b:
# Critical Changes

# Changes

* Add support for rolling up Soft Credits using the "LDV" approach of a single large query on OpportunityContactRole as the primary batch object. By sorting the data by Contact.AccountId and ContactId within that, the single batch job can roll-up both Contact and Account soft credits (and partial soft credits) in a single batch job regardless of the number of OpportunityContactRole records attached to any single Contact.
* Soft Credit rollups include changes to the Rollups selector class to support OCR as the primary object and to also handle retrieving Opportunities with child OCR and PSC records.
* Minor fixes to how Stateful Rollups are handled by the processor
* Fix an issue with how First/Last/Largest/Smallest data is aggregated by the Operations Service
* Update the Recalculate Rollups button to also recalculate Soft Credits

*Temporary changes that will be reverted before the final merge into Master*
* Commit the Rollup Definitions for all current hard-coded NPSP rollups to use for testing
* Add CRLP_Debug_UTIL class with some handy debugging methods to use in the build/testing process

# Issues Closed

# New Metadata

* CRLP_RollupSoftCredit_SVC
* CRLP_RollupSoftCredit_TEST
* CRLP_Rollup_BATCH_PSC_LDV
* CRLP_Debug_UTIL

# Deleted Metadata

* ALLO_RollupBTN_CTRL


src/classes/CRLP_RollupSoftCredit_SVC.cls
```diff
@@ -0,0 +1,93 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Rollup Services unique to the Soft Credit roll-ups TO both Contact and Household Account
```
njjc: "Rollup Services unique to the Contact Soft Credit roll-ups TO both Contact and Account"

src/classes/CRLP_RollupSoftCredit_SVC.cls
```diff
@@ -0,0 +1,93 @@
...
+* The Partial_Soft_Credit__c object is used for all Soft Credit Rollups as a Proxy for OpportunityContactRole
+* which isn't supported as a selectable Entity in the Rollup__mdt object.
+* A Single Instance of this class will be for EITHER the Account or the Contact, but NOT BOTH AT THE SAME TIME!
+*/
+public class CRLP_RollupSoftCredit_SVC extends CRLP_VRollupHandler {
+
+    public CRLP_RollupSoftCredit_SVC() {
```
njjc: May as well remove this?

src/classes/CRLP_Rollup_BATCH_PSC_LDV.cls
```diff
@@ -0,0 +1,370 @@
...
+/**
+* @author Salesforce.org
+* @date 2017
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Soft Credits rollup to the Account and Contacts object using OpportunityContactRole as the base
+*  LDV style query. Stateful vars are used to hold rollups of the Account and Contact across multiple batch
+*  iterations.
+*/
+public class CRLP_Rollup_BATCH_PSC_LDV implements Database.Batchable<SObject>, Database.Stateful {
```
njjc: I think this class name could use less acronyms ;) Maybe CRLP_RollupSoftCredits_LDV_BATCH?

src/classes/CRLP_Rollup_BATCH_PSC_LDV.cls
```diff
@@ -0,0 +1,370 @@
...
+        this.cntctStatefulRollups = new List<CRLP_Rollup>();
+        this.acctStatefulRollups = new List<CRLP_Rollup>();
+    }
+
+    /*******************************************************************************************************
+    * @description Constructs the class, sets the properties
+    */
+    public CRLP_Rollup_BATCH_PSC_LDV(List<Id> recordIds) {
+        this.recordIds = recordIds;
+        this.cntctStatefulRollups = new List<CRLP_Rollup>();
```
njjc: Super nit-pick, but I like just calling the previous constructor so that it's abundantly clear what we're doing differently. Also makes additions to the other constructor a bit easier.

src/classes/CRLP_Rollup_TDTM.cls
```diff
@@ -155,7 +157,8 @@ public class CRLP_Rollup_TDTM extends TDTM_Runnable {
                 }
             }
 
-            if (!accountIds.isEmpty()) {
+            // Don't submit the batch job when unit tests are running
+            if (!Test.isRunningTest() && !accountIds.isEmpty()) {
```
njjc: I don't see anywhere you're adding to the accountIds List?

njjc: Looks like this if will never be true.

src/classes/CRLP_Rollup_TDTM.cls
```diff
@@ -172,13 +175,32 @@ public class CRLP_Rollup_TDTM extends TDTM_Runnable {
...
-            if (!accountIds.isEmpty()) {
-                System.enqueueJob(new CRLP_RollupQueueable(accountIds));
-            }
-            // Contact Rollup
-            if (!contactIds.isEmpty()) {
-                System.enqueueJob(new CRLP_RollupQueueable(contactIds));
+            // Execute the roll-ups in real-time instead of asynchronously in the following cases:
+            // - Currently in a Unit Test with only a couple of summary objects (unit tests do not support chaining)
+            // - Called from a Batch Job, @Future or Queueable context
+            //  || System.isBatch() || System.isFuture() || System.isQueueable()
```
njjc: Why were the the extra batch/future/queueable conditions removed?

src/classes/STG_PanelOppBatch_CTRL.cls
```diff
@@ -59,9 +62,34 @@ public with sharing class STG_PanelOppBatch_CTRL extends STG_Panel {
...
+        } else {
+            Integer batchSizeAcct = CRLP_Rollup_SVC.getBatchSize(Account.SObjectType);
+            Integer batchSizeCntct = CRLP_Rollup_SVC.getBatchSize(Contact.SObjectType);
+
+            Database.executeBatch(new CRLP_Rollup_BATCH(Account.SObjectType), batchSizeAcct);
+            Database.executeBatch(new CRLP_Rollup_BATCH(Contact.SObjectType), batchSizeCntct);
+            this.numJobSubmitted = 2;
+
+            Integer maxRelatedOppsForNonLDVMode = CRLP_Rollup_SVC.getMaxRelatedOppsForNonLDVMode();
+            Boolean hasLDVAccounts = [SELECT Count() FROM Account WHERE Number_of_Opportunities__c >= :maxRelatedOppsForNonLDVMode LIMIT 1] > 0;
```
njjc: I've seen this a couple times, could be a method somewhere.

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -111,7 +110,7 @@ public without sharing class UTIL_MasterSchedulableHelper {
                     jobsToAbort.add('NPSP 02 - Opportunity Contact Rollups (New)');
                     jobsToAbort.add('NPSP 03A - Opportunity Account Rollups LDV (New)');
                     jobsToAbort.add('NPSP 03B - Opportunity Contact Rollups LDV (New)');
-                    jobsToAbort.add('NPSP 04 - Opportunity Soft Credit Rollups (New)');
+                    jobsToAbort.add('NPSP 04 - Opportunity Soft Credit Rollups LDV (New)');
```
njjc: Since this text is user-facing we shouldn't use the LDV acronym.

src/classes/CRLP_RollupSoftCredit_SVC.cls
force2b: The empty constructor is needed to instantiate this as an Interface.

src/classes/CRLP_Rollup_BATCH_PSC_LDV.cls
force2b: @njjc Good point. Though I'd like to keep 3 batch class names consistent:
* CRLP_Rollup_BATCH
* CRLP_Rollup_BATCH_LDV
* CRLP_Rollup_BATCH_PSC_LDV ==> CRLP_Rollup_BATCH_SoftCredits_LDV

Technically, this batch job is for the Contact Soft Credits - differentiating from the future Account Soft Credit functionality. The batch size name size is limited though. I think the above will work, and we could call the next one CRLP_Rollup_BATCH_AcctSoftCredits_LDV (if there is really a need for another batch class, which I'm not so sure of).

This is where the instant refactoring capability of IluminatedCloud v2 would be helpful. :)

src/classes/CRLP_Rollup_TDTM.cls
force2b: I had removed them for some testing and forgot to add them back in. Interesting that having them not in there didn't create any issues in my test rollups. Maybe I just didn't hit a scenario where these conditions were met. I put them back in.

src/classes/STG_PanelOppBatch_CTRL.cls
force2b: The condition itself is included a few times, but I was unable to find a second query like this that is against the entire Account object

src/classes/UTIL_MasterSchedulableHelper.cls
force2b: @njjc  I was wondering about this, along with the staggered job times. 
* NPSP 01 - Opportunity Account Rollups (New) -- 11:00 PM
* NPSP 02 - Opportunity Contact Rollups (New) -- 11:10 PM
* NPSP 03A - Opportunity Account Rollups LDV (New) -- 11:20 PM
* NPSP 03B - Opportunity Contact Rollups LDV (New) -- 11:30 PM
* NPSP 04 - Opportunity Soft Credit Rollups LDV (New) -- 11:50 PM
* NPSP 05 - GAU Allocation Rollups LDV (New) -- 11:00 PM

My goal was keep the same naming pattern/sequence, but since one new batch job was needed it ended up as 3A and 3B. I'm definitely open to other naming patterns. Just not sure of another good term to use to describe the LDV type job and differentiate from the other jobs listed. The main difference is which batch job class gets executed. I suppose we could go with something like "Contact Rollups Job 1" and "Contact Rollups Job 2", though it's not very explanatory. 

I'm not tied to the `(New)` text either. I added that initially just to make it easier for me know when the jobs were scheduled. At this point I think we can drop `(New)`.

src/classes/CRLP_Rollup_BATCH_PSC_LDV.cls
njjc: For some reason I want BATCH to be a suffix like it is for the rest of the classes, but I'm not going to be dogmatic about it ;)

force2b: True, that is a pattern for the rest of the classes. I could go with:
* CRLP_Rollup_BATCH
* CRLP_Rollup_HardCredit_LDV_BATCH _(noting that technically this is also used for Allocations to GAU which isn't explicitly a "Hard Credit" type rollup)_.
* CRLP_Rollup_SoftCredit_LDV_BATCH

njjc: I like that pattern better but I'd remove the HardCredit.

src/classes/UTIL_MasterSchedulableHelper.cls
njjc: I think including the word "Customizable" is a good idea, instead of (New). I'm fine with 3A and 3B so that we don't have to reconfigure everything. I don't have a strong preference on the timing; the only potential hangup I can think of is row locking errors if we're updating a contact and account at the same time, but I don't think we've hit that issue today and they're all kicked off at the same time.

Some day I'd love to come up with a better scheme so that we're not iterating over every record in the database every night... something like do the current full job weekly and the nightly jobs just iterate over Opportunity where LastModifiedDate is within the timeframe since the last batch job completed successfully... but that's a project for another time :)

force2b: How about this?
* NPSP 01 - Customizable Rollups - Account Hard Credit
* NPSP 02 - Customizable Rollups - Contact Hard Credit
* NPSP 03A - Customizable Rollups - Account Hard Credit Job 2
* NPSP 03B - Customizable Rollups - Contact Hard Credit Job 2
* NPSP 04 - Customizable Rollups - Soft Credits
* NPSP 05 - Customizable Rollups - General Allocation Units

njjc: How about "Contact/Account Hard Credit for Data Skew" or "for Large Data Volumes"? Curious what @smarovich thinks.

force2b: While we're on this topic, should the "NPSP 06 - Recurring Donation Updates" job run at 11PM as well? That job can create/modify Opportunities - which are being rolled up by the various rollup jobs at the same time. Seems like that job should either start at 10PM or 12AM. I suppose that to handle a truly LDV org that has tens of millions of records, there may be a need to stagger that job much more than an hour or so.

force2b: Here's what the jobs list looks like at the moment:

image

njjc: Good point, since it only creates open opps and we didn't used to roll those up it didn't matter. I think having the RD job first then the rollups jobs start an hour or two later sounds good.

Some LDV orgs have rollups jobs running longer than 24 hour s:(

force2b: Hopefully our new engine will improve the job running times, but I still expect 30 million Opportunities to take a while. I'll move the RD start time to 10PM.

smarovich: @force2b I'm so sorry I missed this! I think the more info we can give in the name, the better. I like your proposed text "Contact Hard Credit for Large Data Volumes." That seems the most clear to me.

PR #2910, opened 2018-01-09
Nightly Membership Rollups may cause unnecessary record updates

### LINKED ISSUE #2909: Nightly Membership Rollups were causing all Accounts and Contacts to be updated when a blank LastMembershipLevel value did not match the null value in the record.
If nightly rollup settings included a specific record type for Memberships and that caused the LastMembershipLevel or LastMembershipOrigin rollups to result in a blank ("") value, comparing that "" value to the value in the record (null) did not match; causing all of those records to be updated. At a database level, Field History did not capture the change because the blank ("") value was effectively equivalent to null - keeping this from showing up in the Field History lists.

The fix is to convert "" to null for the LastMembershipLevel or LastMembershipOrigin calculated rollup values before comparing to the existing values in the record.

force2b:
The calculated lastMemberLevel and lastMemberOrigin vars may result to an empty string ("") which isn't an exact match to the value in the database (null), causing the code to think a change was made resulting in the unnecessary DML. The fix is to convert the empty string ("") to a null prior to the comparison to ensure the comparison to the database field matches.

# Critical Changes

# Changes

# Issues Closed

Fixes #2909 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RLLP_OppRollup.cls
+++ b/src/classes/RLLP_OppRollup.cls
@@ -955,14 +955,15 @@ public without sharing class RLLP_OppRollup {
 
         // fill in summary totals
         obj.put('npo02__TotalMembershipOppAmount__c', amt);
-        obj.put('npo02__NumberOfMembershipOpps__c', cnt);               
-        obj.put('npo02__LastMembershipDate__c', (date)(r.get('LastCloseDate')));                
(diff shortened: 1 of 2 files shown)


njjc:
@force2b I take it there's no way to write an apex test that reproduces the issue?

force2b:
Good question. Not sure, but it's worth investigating.

PR #2924, opened 2018-01-19
Cascade Delete Selective Query

### LINKED ISSUE #2650: Cascade Delete Non Selective Query Errors
[Users have reported](https://powerofus.force.com/0D58000003FmTM9) non-selective query errors when bulk deleting accounts.

In testing query selectiveness for cascade delete queries that use multiple cross object lookups, (e.g. ```Recurring_Donation__r.npe03__Contact__r.AccountId```), it appears that only the last lookup can use the available index; the __r relationships use a table scan every time, which causes non-selectiveness.

The solution is redesign cascade deletion to split cross object lookups into multiple separate queries, allowing each one to use the index.

More information [here](https://salesforce.quip.com/pKYsAzptcsax), lines 225-230.

kofijohnson:
# Critical Changes

# Changes

# Issues Closed
Fixes #2650

# New Metadata

# Deleted Metadata


src/classes/CDL_CascadeDeleteLookups.cls
```diff
@@ -0,0 +1,349 @@
...
+            
+            return String.join(result, ', ');
+        }
+
+        /*******************************************************************************************************
+        * @description Get the current user Theme Context (Lightning Experience or Classic).
+        * @return Boolean.
+        ********************************************************************************************************/
+        private Boolean isLightningContext() {
+            if (LIGHTNING_CONTEXT != null) {
```
RadSFDeveloper: Since this is a variable used in tests only, can we rewrite the code as following so the variable works for both tests and production code:
```
@TestVisible
private static Boolean isLightningContext {
    get {
         return isLightningContext = isLightningContext == null
             ? USER_UserService.isUserRunningLightning() || USER_UserService.isUserRunningMobile()
             : isLightningContext
    }
    set;
}
```

The variable can be set from a test code as CDL_CascadeDeleteLookups.isLightningContext = true;//or false;

src/classes/CDL_CascadeDeleteLookups.cls
```diff
@@ -0,0 +1,349 @@
...
+            while (currentChildrenLoader != null) {
+                List<SObject> children = currentChildrenLoader.load(undeletedRecordIds);
+                if (children.isEmpty()) {
+                    break;
+                }
+
+                List<Database.UndeleteResult> undeleteResults = Database.undelete(children, false);
+                ERR_Handler.Errors errorResult = ERR_Handler.getErrors(undeleteResults, children);
+                if (errorResult.errorsExist) {
+                    ERR_Handler.processErrors(errorResult.errorRecords, 'TDTM');
```
RadSFDeveloper: Does `processErrors` throw an error and stops undelete from processing any further?
If not, does `currentChildrenLoader = currentChildrenLoader.next();` selects next for only successfully undeleted children?

src/classes/ACCT_CascadeDeleteLookups_TDTM.cls
```diff
@@ -33,51 +33,240 @@
...
+                ALL ROWS
+            ];
+        }
+
+        /*******************************************************************************************************
+        * @description Retrieves the related Relationships of the undeleted Accounts.
+        * @param List<Contact> The Contacts of the undeleted Accounts.
+        * @return List<npe4__Relationship__c> Undeleted Accounts' Relationships.
+        ********************************************************************************************************/
+        private List<npe4__Relationship__c> getRelatedRelationships(List<Contact> relatedContacts) {
```
davidhabib: this seems to be an exact duplicate of the previous getRelatedRelationships() in CascadeDeleteLoader.  why not share this method by using a private static?  I'm guessing several of your other select routines also could be shared.

src/classes/CDL_CascadeDeleteLookups.cls
davidhabib: agreed with Radmila's comment.  Also, I don't like seeing variables fully capitalized, which we should be reserving for constants.

src/classes/ACCT_CascadeDeleteLookups_TDTM.cls
kofijohnson: @davidhabib they are different functions. The first one returns relationships that are not deleted, and the second one return relationships that are already deleted.

davidhabib: missed that.  thanks @kofijohnson

src/classes/CDL_CascadeDeleteLookups.cls
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        /*******************************************************************************************************
+        * @description Validates the children records. This method uses the injected CascadeDeleteValidator,
+        * to validate the children records.
+        * @return void
+        ********************************************************************************************************/
+        public void validate() {
+            if (validator != null) {
+                validator.validate(deletedRecords.values(), children);
```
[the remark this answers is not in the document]
kofijohnson: @RadSFDeveloper, deletedRecords comes from the Trigger Contexts, so we are fine here.

src/classes/CDL_CascadeDeleteLookups.cls
kofijohnson: @RadSFDeveloper you have a good point. "Database.undelete(children, false)" is silent, so if there is an error the current logic will go to the next group. That means, the undeletion of a parent record will fail but the logic will proceed and undelete its child. The issue is, the reference from the parent to child will be broken. 
One way I think we can solve this is if there is an error "errorResult.errorsExist()", the logic should not go to the next group.
@njjc any thoughts?

njjc: Can we only pass the record IDs that undeleted successfully to the next round of deletion?

kofijohnson: Great idea @njjc. I'll implement the fix.

kofijohnson:
@RadSFDeveloper @njjc I completed the test yesterday. All looks good now. Thanks for doing a last review and if everything is good, we can merge.

src/classes/CON_CascadeDeleteLookups_TDTM.cls
```diff
@@ -30,64 +30,85 @@
 /**
  * @author Salesforce.org
  * @date 2017
- * @group Utilities
- * @description When contacts are deleted, cascade deletes related relationships, recurring donations, and 
- * opportunities. Prevents Contact deletion if any attributed opportunities are closed/won or attributed
- * to an organization.
+ * @group Cascade Delete
+ * @description Cascade deletion extension for Opportunity object.
```
RadSFDeveloper: The description associates different sObject than expected.

src/classes/CDL_CascadeDeleteLookups.cls
```diff
@@ -0,0 +1,350 @@
...
+    /** @description The single instance of the CascadeDelete. */
+    private static CascadeDelete cascadeDelete;
+
+    /*******************************************************************************************************
+    * @description Gets the current user Theme Context (Lightning Experience or Classic).
+    * @return Boolean.
+    ********************************************************************************************************/
+    public static Boolean isLightningContext {
+        get {
+            return (isLightningContext == null)
```
RadSFDeveloper: The var `isLightningContext` is not assigned, so this check will be done every time. Feel free to break it down, ie does not have to be as indicated in my comment. So, either is fine:

```
get {
    return isLightningContext = isLightningContext == null
        ? USER_UserService.isUserRunningLightning() || USER_UserService.isUserRunningMobile()
        : isLightningContext;
}

```
OR

```
get {
    if (isLightningContext == null) {
        isLightningContext = USER_UserService.isUserRunningLightning() || USER_UserService.isUserRunningMobile();
    }
    return isLightningContext;
}
```

src/classes/CDL_CascadeDeleteLookups_TEST.cls
```diff
@@ -0,0 +1,364 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2016
+* @group Cascade Delete
+* @description Utilities for the Cascade Delete Test Classes
```
RadSFDeveloper: Description need update

src/classes/CON_DeleteContactOverride_TEST.cls
```diff
@@ -65,7 +65,7 @@ private class CON_DeleteContactOverride_TEST {
         insert con;
 
         list<Contact> queryCon = getContactList(con.Id);
-        system.assertNotEquals(null, queryCon[0].AccountId, 'The contact should have an auto-created household.');
+        system.assertNotEquals(null, queryCon[0].AccountId, 'The contact should have an auto-buildd household.');
```
RadSFDeveloper: Typo

src/classes/CON_DeleteContactOverride_TEST.cls
```diff
@@ -319,7 +319,214 @@ private class CON_DeleteContactOverride_TEST {
...
+
+    /*********************************************************************************************************
+    * @description Attempts to delete contact with closed/won and org opps, verifies correct exception is thrown.
+    */
+    static testMethod void testContactDeleteExceptionWithOverride(){
+        //use a non-system account for the contact's account, so that the delete override doesn't redirect us to delete the account
+        Account actualAccount = new Account(Name = 'TEST_ACCOUNT');
+        insert actualAccount;
+
+        Contact actualContact = new Contact(LastName = 'TEST_CONTACT', AccountId = actualAccount.Id);
```
RadSFDeveloper: We can simply declare them `contact`, `account`, `case` vars in this and the following test.

njjc:
@kofijohnson Will this be able to make it into master today?

kofijohnson:
Yes @njjc the code review was done and fixes were applied. Thank you @njjc  for doing a final check and merge if there everything is good.

PR #2928, opened 2018-01-26
Configurable Recurring Donation Batch Size

### LINKED ISSUE #2610: Allow configurable batch size for RD_RecurringDonations_BATCH
We receive errors ("Too many query rows: 50001") when the daily Recurring Donations batch (RD_RecurringDonations_BATCH) process runs to create/refresh opportunities in Recurring Donations. The batch size for RD_RecurringDonations_BATCH is set to 50 and cannot be changed. We've reduced the Opportunity forecast months and it partially alleviates the problem, but a configurable batch size would be great.  

Can we allow the RD_RecurringDonations_BATCH batch size to be configurable, similar to how the Donation Rollup processes are in the NPSP Settings tab? (like RLLP_OppContactRollup_BATCH, RLLP_OppAccRollup_BATCH etc).

jesslopez-sf:
Added Custom Setting field to allow Recurring Donation Batch Size to be configurable. The nightly scheduled job and bulk data process both use this custom setting for the batch size. If no custom setting field value is present, it defaults to 50. The default for the custom setting field is also 50.

# Critical Changes

# Changes

- The batch size for the Recurring Donations nightly scheduled job and bulk data process was previously hard coded to 50. You can now configure this value using the new **Recurring Donation Batch Size** field located in **Setup | Custom Settings | Recurring Donations Settings**. If this field is left blank, it will default to 50.

# Issues Closed
Fixes #2610

# New Metadata
npe03__Recurring_Donations_Settings__c. Recurring_Donation_Batch_Size__c
# Deleted Metadata

### DIFF
--- a/src/classes/STG_PanelRDBatch_CTRL.cls
+++ b/src/classes/STG_PanelRDBatch_CTRL.cls
@@ -51,6 +51,10 @@ public with sharing class STG_PanelRDBatch_CTRL extends STG_Panel {
     * @description Whether the batch process is currently running  
     */
     public boolean isRunningBatch { get; set; }
+
+    /* @description Batch size of Recurring Donations Apex job  
+    */
+    public Integer batchSize = (UTIL_CustomSettingsFacade.getRecurringDonationsSettings().Recurring_Donation_Batch_Size__c <> null) ? UTIL_CustomSettingsFacade.getRecurringDonationsSettings().Recurring_Donation_Batch_Size__c.intValue() : 50;
     
     /*********************************************************************************************************
     * @description Action Method to execute the RD Batch process
@@ -62,7 +66,7 @@ public with sharing class STG_PanelRDBatch_CTRL extends STG_Panel {
 	    	isRunningBatch = true;
 	        
 	        //call the batch job, processing 50 at a time 
-	        Id batchInstanceId = Database.executeBatch(new RD_RecurringDonations_BATCH(), 50); 
+	        Id batchInstanceId = Database.executeBatch(new RD_RecurringDonations_BATCH(), batchSize); 
         }
         catch(Exception e) {
             Database.rollback(sp);
--- a/src/classes/STG_PanelRDBatch_CTRL.cls
+++ b/src/classes/STG_PanelRDBatch_CTRL.cls
@@ -51,10 +51,6 @@ public with sharing class STG_PanelRDBatch_CTRL extends STG_Panel {
     * @description Whether the batch process is currently running  
     */
     public boolean isRunningBatch { get; set; }
-
-    /* @description Batch size of Recurring Donations Apex job  
-    */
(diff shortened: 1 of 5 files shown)


src/classes/STG_PanelRDBatch_CTRL.cls
```diff
@@ -51,6 +51,10 @@ public with sharing class STG_PanelRDBatch_CTRL extends STG_Panel {
     * @description Whether the batch process is currently running  
     */
     public boolean isRunningBatch { get; set; }
+
+    /* @description Batch size of Recurring Donations Apex job  
+    */
+    public Integer batchSize = (UTIL_CustomSettingsFacade.getRecurringDonationsSettings().Recurring_Donation_Batch_Size__c <> null) ? UTIL_CustomSettingsFacade.getRecurringDonationsSettings().Recurring_Donation_Batch_Size__c.intValue() : 50;
```
force2b: Rather than recreate the var here (repeating the code), it's probably better to do the following in the `runBatch()` method below:

```
RD_RecurringDonations_BATCH rdb = new RD_RecurringDonations_BATCH();
Id batchInstanceId = Database.executeBatch(rdb, rdb.batchSize);
```

jesslopez-sf: Totally. Good call. I'll make that change now.

davidhabib:
@jesslopez-sf congrats on your first pull request (I think?) to NPSP!

jesslopez-sf:
Thanks, @davidhabib ! It was my first pull request. :)

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fields>
+        <fullName>Recurring_Donation_Batch_Size__c</fullName>
+        <defaultValue>50</defaultValue>
+        <description>The number of records processed at a time when running the Recurring Donations job. The default size is 50. Reduce to a smaller number if the Recurring Donation job is failing due to system limits.</description>
```
force2b: @smarovich Would you mind taking a look at the description/inline help text for this new custom settings field? It's not visible in the NPSP Setting UI yet, but it will be soon.

smarovich: @force2b  @jesslopez-sf   Looks great. I'd make some very minor changes:  The number of records to process at a time when running the Recurring Donations batch job. The default size is 50. Reduce to a smaller number if the Recurring Donations batch job is failing due to system limits.

smarovich:
@jesslopez-sf In the release notes, I want to add a little more detail on where to find this new field. Is this the right place: In Setup, search for Custom Settings, then click Manage next to Recurring Donations Settings.

src/objects/npe03__Recurring_Donations_Settings__c.object
jesslopez-sf: Thanks, @smarovich ! I updated the label with your suggestions.

smarovich:
@jesslopez-sf @force2b I modified the release notes a bit to provide more info. Let me know if I need to make any changes. Thanks!

jesslopez-sf:
Looks good. Thank you, @smarovich!

PR #2940, opened 2018-02-02
BDI payment fixes

### LINKED ISSUE #2938: BDI: new Open Opps should not have Payment Date set on unpaid payments
If you specify a new Donation in BDI, and its stage is an open stage (not closed won), the payment under the Opp is correctly marked unpaid, but both Scheduled Date and Payment Date are set to the Donation Date.  Payment Date should remain empty for unpaid Payments BDI creates.

### LINKED ISSUE #2939: BDI: matched Payments should get marked Paid regardless of other field updates
Due to bug #2938, an imported Donation that matched an existing unpaid Payment would NOT get updated with Paid, if no other BDI payment fields were specified.  BDI must ensure that it updates Paid, regardless of other fields being updated or not

davidhabib:
I found two issues that were causing the behavior during our demo that a matched payment was still not getting updated to paid.  The first issue was that we were incorrectly setting a payment's Payment Date when creating an unpaid Payment; in this case we should have left Payment Date empty.  The second issue was that Payments were not being updated to Paid, if no other fields were being updated on the payment; now we ensure we force the update if we are marking it paid.

# Critical Changes

# Changes

# Issues Closed
fixes #2938 
fixes #2939 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_ContactService.cls
+++ b/src/classes/BDI_ContactService.cls
@@ -137,11 +137,11 @@ public with sharing class BDI_ContactService {
     */
     private void verifyContactData() {
         for (DataImport__c di: bdi.listDI) {
-            if (isContactSpecified(di, 'Contact1') && di.Contact1_Lastname__c == null && di.Contact1Imported__c == null) {
+            if (isContactSpecified(di, 'Contact1') && string.isBlank(di.Contact1_Lastname__c) && di.Contact1Imported__c == null) {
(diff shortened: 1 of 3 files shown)


njjc:
@davidhabib some BDI tests are failing with this PR.

davidhabib:
Thanks @njjc .  I needed to add a field to the soql that we are always now referencing.  interesting that it only tripped up a very small set of tests, and none in the class BDI_Donations_TEST, that I ran since all the work was related to payments!

PR #2942, opened 2018-02-06
Adding translation files for Recurring Donations Settings

jesslopez-sf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
Added  language translation files for Recurring Donations Settings
# Deleted Metadata


src/package.xml
```diff
@@ -304,6 +304,8 @@
         <members>UTIL_Currency_TEST</members>
         <members>UTIL_CustomSettingsFacade</members>
         <members>UTIL_CustomSettings_API</members>
+        <members>UTIL_DMLService</members>
```
force2b: These 2 files shouldn't be in here. I think they're just out of order in the file, but they should be there already so this would duplicate them.

jesslopez-sf: Ok. I just removed them and pushed the change.

PR #2950, opened 2018-02-12
Rollups: Calculation and other updates

force2b:
# Critical Changes

# Changes
* Various changes to streamline the code and resolve a few calculation issues that were discovered during testing.  
* Abstracted out the processing options logic into its own class and started logic to support filtering job types for performance.
* Add support for filtering the detail query on Opportunity using the current set of Filters to reduce the number of records queried/processed.
* Update the Contact Hard Credit logic to exclude Organization Donations
* Update the Contact Hard Credit logic to check for detail (Opportunity) records to be rolled up. If there are none then skip the rollup of that Contact entirely. Significantly improves the job performance - at least when there are a significant number of Contacts that have no related Hard Credit Opportunities.
* Update the Contact Soft Credit logic to included Organization Donations for the Primary contact
* Fix the LDV batch jobs (actually the ProcessingOptions logic) to fully clone the map of stateful rollups when passing to the RollupProcesser_SVC.startRollupProcess() method. Otherwise the map is passed by reference so all additions the startRollupProcess() method makes to the map are held as stateful throughout the life of the batch job. That cause a heap size issue.
* Identified that using a sub-query on Payments in the batch queryLocator may create apex cpu time out errors in LDV orgs. Completely refactored how payments are queried and processed. This will actually help with allowing payments to be passed to the rollup engine from the archive bridge.
* Refactor Custom Metadata Type deployments to eliminate the nested deployment process as it appears to not be necessary given that the FilterGroup relationship is through DeveloperName not a generated Id
* Refactor the Rollup & FilterRule Operation enums to match the exact API name of the related fields. Avoid hardcoding the labels for these as well. Where necessary use getPicklistValues() to return the appropriate label based on the api name, which should handle localization properly.
* To address a potential heap size issue in the LDV batch jobs, clear out List<CRLP_Rollup> items from the stateful map after a fixed number of batch iterations (500) have passed without seeing the related parentId in the scope. It's a bit of an assumption to guess that Salesforce won't skip more than 500 batch iterations before returning a block, but it appears to be solid in terms of observed behavior.

# Issues Closed

# New Metadata
* CRLP_RollupProcessingOptions.cls

# Deleted Metadata


src/classes/CRLP_Rollup.cls
```diff
@@ -372,4 +405,32 @@ public class CRLP_Rollup {
 //        CRLP_Debug_UTIL.amendLastCurrentRollupState('==> ' + resultValue);
         return resultValue;
     }
+
+    public class Rollupmdt {
```
njjc: Header needed.

src/pages/PMT_PaymentWizard.page
```diff
@@ -260,7 +260,10 @@ Copyright (c) 2008 Evan Callahan, evanc@npowerseattle.org, NPower Seattle, 403 2
                                     <label class="slds-form-element__label" for="oppAmount4">{!$ObjectType.Opportunity.Fields.Amount.Label}</label>
                                     <div class="slds-form-element__control">
                                         <span class="slds-form-element__static">
-                                            <apex:outputField value="{!currentOpp.Amount}" id="oppAmount4"/>
+                                            <span class="slds-form-element__addon">{!currencySymbol}</span>
```
njjc: Any idea why these are showing in this diff? Some sort of problem with auto merge from master?

force2b: This (and a few others) were change I had made to the page way back when I worked on advanced currency management support in rollups. I think at some point when doing a manual merge conflict sync, I overwrote those changes requiring them to be recreated. Here's the full commit: https://github.com/SalesforceFoundation/Cumulus/pull/2950/commits/8a31168af298c4cfb78eea4e11c17b5f72d8d39f

PR #2957, opened 2018-02-20
Recurring Donations Processing Updates

### LINKED ISSUE #1386: Any change to Recurring Donation with Open-ended status of 'None' causes donations to be deleted and re-inserted
Based on NPSP documentation, I would expect that when a Recurring Donation Close Date or Amount is updated, all related donations are deleted and recreated.   However, I am now seeing this deletion and recreation happen when I make any change in the recurring donation.  In the process, any user-defined data set at the Opportunity level is being lost.    

This may be related to https://github.com/SalesforceFoundation/Cumulus/issues/1359 
The behaviour has only been reported this week, so may be related to Spring 15.

bethbrains:
# Critical Changes

# Changes

- Open Opportunities are now updated when a Recurring Donation is edited. Previously, upon update of a Recurring Donation, any open Opportunities were deleted and reinserted, potentially causing data loss.

# Issues Closed
- Fixes #1386

# New Metadata

# Deleted Metadata


src/classes/RD_RecurringDonations.cls
```diff
@@ -216,29 +219,41 @@ public class RD_RecurringDonations {
...
+                    }
                     rd.npe03__Total_Paid_Installments__c = (integer)obj.get('oppcount');
-                    updateMap.put(rdid, rd);                        
-                }                   
+                    updateMap.put(rdid, rd);
+                }
+
+                // if no open opportunities remain because forecast months is too short,
+                // use last payment date to project next payment date
+                if(rd.npe03__Last_Payment_Date__c != null) {
```
davidhabib: should we only be doing this for open ended RD's?  was there an existing bug for this scenario?  if so, we should refer to the bug number in the pull request issues section.

src/classes/RD_RecurringDonations.cls
```diff
@@ -728,9 +699,10 @@ public class RD_RecurringDonations {
         //opps for updating
         list<Opportunity> updateOppsList = new list<Opportunity>();
         
+        // TODO: refactor this loop out into its own method so i can use it elsewhere as well
```
davidhabib: still planning on this TODO?  we should avoid checking in any TODO's into master.

src/classes/RD_RecurringDonations.cls
bethbrains: It is only for open-ended, based on the original query at 186. 
There wasn't a bug for it that I'm aware of, but became necessary as a result of the refactor, since we use getStartDate, which returns the Date Established if Next Payment Date is null. Michael and I can discuss with you further if this isn't clear.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+
+            Boolean needsRecalc = needsRecalculation(r,oldRecurringDonations.get(r.id));
+
+            if (needsRecalc) {
+                if(r.npe03__Open_Ended_Status__c == closedlabel
+                        //&& oldRD.npe03__Open_Ended_Status__c != r.npe03__Open_Ended_Status__c
                     ){
-                    reevaluateopps.add(r.id);
+                    // closing an open ended donation
+                    // does it actually matter if it's newly closed? shouldn't it go to the runRecurringDonationClosedActions method regardless?
```
davidhabib: i'm guessing the logic was that if it was already closed, then there was nothing that would need to get updated.  so for example, if it already was closed, and the user changes the schedule from monthly to yearly, should we do anything?  i'm guessing the intent was no, once it is closed, leave its opps alone.  If you feel we have a case for still updating the RD, lets discuss with PM.

src/classes/RD_RecurringDonations.cls
bethbrains: removed. no longer planned. thanks for the catch.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+                    , CampaignId
+                    , AccountID
+                    , CloseDate
+                    , StageName
+                    , isWon
+                    , Recurring_Donation_Installment_Number__c
+                    , npe03__Recurring_Donation__c
+            FROM Opportunity
+            WHERE npe03__Recurring_Donation__c IN :fixedLengthRDsToRecalculate
+            ORDER BY CloseDate ASC];
```
davidhabib: is it intentional that open ended recalc only looks at existing open opps, but fixed length is looking at open and closed opps?

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+            handleFixedLengthScheduleRecalculation(fixedLengthRDsToRecalculate, opps, dmlWrapper);
+        }
+    }
+
+    private static void handleOpenEndedScheduleRecalculation(List<npe03__Recurring_Donation__c> recurringDonations
```
davidhabib: please put ApexDoc headers all all methods, even private ones.  thanks.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+
+            Date nextDate = getStartDate(r);
+
+            integer rdcounter = 1;
+            if (r.npe03__Total_Paid_Installments__c > 0) {
+                rdcounter = (integer)r.npe03__Total_Paid_Installments__c + 1;
+            }
+
+            // loop the existing opportunities
+            //if(RDOppMap.get(r.Id) != null && !RDOppMap.get(r.Id).isEmpty()) {
```
davidhabib: i'm not a big fan of keeping commented out code that is no longer relevant.  since you put in empty lists for RD's with no open opps, i'm guessing this comment isn't needed?

src/classes/RD_RecurringDonations.cls
bethbrains: Yes, since Fixed Length RDs need to see how many opps are already closed in order to know how many should still be open.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+                for (Opportunity o : RDOppMap.get(r.Id)) {
+                    if (nextDate <= dtForecast) {
+                        // fix close dates on existing opportunities
+                        o.CloseDate = nextDate;
+
+                        // TODO: are these really all the fields necessary to sync to oppties on an update?
+                        // This is borrowed from roughly line 700
+                        o.Amount = r.npe03__Amount__c;
```
davidhabib: let's resolve our todo's before going into master

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +767,391 @@ public class RD_RecurringDonations {
...
+            Date nextDate = getStartDate(r);
+
+            integer rdcounter = 1;
+            if (r.npe03__Total_Paid_Installments__c > 0) {
+                rdcounter = (integer)r.npe03__Total_Paid_Installments__c + 1;
+            }
+
+            Integer numInstallments = r.npe03__Installments__c.intValue();
+
+            Integer j=0;
```
davidhabib: don't see j being used anywhere.

src/classes/RD_RecurringDonations.cls
```diff
@@ -884,29 +1170,69 @@ public class RD_RecurringDonations {
...
                 listRDNonOpen.add(rd);
             } else {
                 listRDOpen.add(rd);
             }
         }
 
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
         if (listRDNonOpen.size() > 0) {
+            // TODO: update this to use the new methods
```
davidhabib: todo done?

src/classes/RD_RecurringDonations.cls
```diff
@@ -1214,4 +1543,92 @@ public class RD_RecurringDonations {
...
+    /*******************************************************************************************************
+    * @description Returns true if the RD changes should trigger recalculation of the child opportunities
+    * @param new The updated Recurring Donation
+    * @param old The original Recurring Donation before updates
+    * @return boolean
+    */
+    @testVisible
+    private static Boolean needsRecalculation(npe03__Recurring_Donation__c newRD, npe03__Recurring_Donation__c oldRD) {
+
+        if(newRD.npe03__Organization__c != oldRD.npe03__Organization__c
```
davidhabib: not sure if this is because you copied existing code, but I our standard is space between if (...)

src/classes/RD_RecurringDonations.cls
```diff
@@ -1214,4 +1543,92 @@ public class RD_RecurringDonations {
...
+
+
+        if(isMultiCurrency){
+            //set the donation currency field to equal the recurring donation currency
+            o.put(OppCurrencyField,r.get(RDCurrencyField));
+        }
+
+        o.Amount = r.npe03__Amount__c;
+
+        String oName = '';
```
davidhabib: is this method only used for Open RD's or Fixed Length too?  If also for Fixed Length, then the naming strategy needs to change.

bethbrains: fixed in latest commit. good catch.

src/classes/RD_RecurringDonations.cls
force2b: This query will also include records with a npe03__Open_Ended_Status__c value of "None" as well as "Open" (since the query is for `!= 'Closed'`). Would be worth understanding the impact

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
         }
-        //needs a total reinsert, not an update
-        if (!reinsertopps.isEmpty()){
-            //clear existing open opps and recreate
-            delete [select id from Opportunity where npe03__Recurring_Donation__c IN :reinsertopps and isClosed = false];
-            if (reinsertopps.size() == 1 || system.isFuture() || system.isBatch()){
-                insertOppsOnRecurringDonationInsert(reinsertopps);                  
+
+        if (!fixedLengthRDsToRecalculate.isEmpty()) {
+            List<Opportunity> opps = [SELECT Id
```
force2b: I'm pretty sure this code gets called in a trigger context, so ideally it should limit the number of SOQL queries. These 2 queries are nearly identical except for the filter. It might be better to combine into a single query and then split the data in a for loop into two collections for passing to the appropriate method for handling.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
+    * @return void
+    */
+    private static void handleOpenEndedScheduleRecalculation(List<npe03__Recurring_Donation__c> recurringDonations
+            , List<Opportunity> openOpportunities
+            , TDTM_Runnable.DmlWrapper dmlWrapper) {
+
+        npe03__Recurring_Donations_Settings__c rds = UTIL_CustomSettingsFacade.getRecurringDonationsSettings();
+
+        //build a map of recurring donation ID and opps
+        Map<Id, List<Opportunity>> RDOppMap = new Map<Id, List<Opportunity>>();
```
force2b: For new code, all vars should start with a lowercase. I know this class is pretty inconsistent with that, but anything new we write or fix should use camelcase vars.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
+            if (!RDOppMap.containsKey(rd.Id)) {
+                RDOppMap.put(rd.Id, new List<Opportunity>());
+            }
+        }
+
+        List<Opportunity> updateOppsList = new List<Opportunity>();
+        List<Opportunity> closeOppsList = new List<Opportunity>();
+        List<Opportunity> insertOppsList = new List<Opportunity>();
+
+        integer forecast_months = (integer)rds.npe03__Opportunity_Forecast_Months__c;
```
force2b: I believe we've settled on upper case for type declaration ("Integer" instead of "integer"). Plus, the two lines here are not consistent for case - "integer" and "Date"

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
+                    if (r.npe03__Recurring_Donation_Campaign__c != null && rds.npe03__Add_Campaign_to_All_Opportunites__c) {
+                        o.CampaignId = r.npe03__Recurring_Donation_Campaign__c;
+                    }
+
+                    if (r.npe03__Organization__c != null) {
+                        o.AccountID = r.npe03__Organization__c;
+                    }
+
+                    //modify the name
+                    String oName = '';
```
force2b: I've noticed there are a number of blocks of code in this class that concatenate some values into an Opportunity name. If feasible, it would be helpful to refactor these into a single method that build the Opportunity.Name value using parameters passed to the method (such as RecurringDonation__c, Opportunity, CloseDate, etc.)

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
+
+        if (!insertOppsList.isEmpty()) {
+            Database.SaveResult[] lsr;
+            list<Database.SaveResult> badSRs = new list<Database.SaveResult>();
+            list<Opportunity> badOpps = new list<Opportunity>();
+
+            lsr = Database.insert(insertOppsList, false);
+
+            set<id> updateRecurringDonationSet = new set<id>();
+            for (Opportunity o : insertOppsList) {
```
force2b: Not sure I'm clear on what happens if there are insert errors here because the code to check for errors is below this loop.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,77 +766,391 @@ public class RD_RecurringDonations {
...
+        }
+
+        if (!insertOppsList.isEmpty()) {
+            Database.SaveResult[] lsr;
+            list<Database.SaveResult> badSRs = new list<Database.SaveResult>();
+            list<Opportunity> badOpps = new list<Opportunity>();
+
+            lsr = Database.insert(insertOppsList, false);
+
+            set<id> updateRecurringDonationSet = new set<id>();
```
force2b: Same here -- error checking logic is below this loop that gets the Opp.RecurringDonation id.

src/classes/RD_RecurringDonations.cls
```diff
@@ -884,29 +1169,65 @@ public class RD_RecurringDonations {
...
         }
 
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
         if (listRDNonOpen.size() > 0) {
-            //clear existing open opps and recreate
-            delete [select id from Opportunity where npe03__Recurring_Donation__c IN :listRDNonOpen and isClosed = false];
-            insertOppsOnRecurringDonationInsert(listRDNonOpen);
+
+            List<Opportunity> opps = [SELECT Id
```
force2b: Same comment here about combining these two queries into a single one and then split the data into 2 collections to pass into their respective handlers.

src/classes/RD_RecurringDonations.cls
```diff
@@ -1214,4 +1538,94 @@ public class RD_RecurringDonations {
...
+        }
+        oName += ') ';
+        oName += o.CloseDate.format();
+        o.Name = oName;
+
+        o.Recurring_Donation_Installment_Number__c = rdcounter;
+
+        if (rds.npe03__Record_Type__c != null){
+            o.put('RecordTypeId', rds.npe03__Record_Type__c);
+            //o.RecordTypeID = rds.npe03__Record_Type__c;
```
force2b: Why not use `o.RecordTypeID = rds.npe03__Record_Type__c` for this?

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
...
+        originalOpps[0].StageName = closedStage;
+        update originalOpps[0];
+        r1.npe03__Installments__c = 4;
+        update r1;
+        Test.stopTest();
+
+        errors = [SELECT Id, Full_Message__c FROM Error__c];
+        System.assertEquals(errors.size(),0,errors);
+
+        List<Opportunity> postUpdateOpps = [SELECT Id
```
force2b: I see a lot of this specific query throughout the test class. It can probably be refactored as a method that accepts an Id or list of Ids as a parameter.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
...
+                        // npe03__Opportunity_Forecast_Months__c = 12,
+                        npe03__Maximum_Donations__c = 50,
+                        npe03__Open_Opportunity_Behavior__c = RD_RecurringDonations.RecurringDonationCloseOptions.Mark_Opportunities_Closed_Lost.name()
+                ));
+
+        Account a = new Account();
+        a.Name = 'test Individual';
+        insert a;
+
+        Contact c = new Contact();
```
force2b: I recommend using `Contact c = UTIL_UnitTestData_TEST.getContact();` for all code that creates a test contact rather than using the generic 'test' name assignment. Uniqueness of these becomes important as we attempt to eventually move to parallel unit tests.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
...
+        c.FirstName = 'test';
+        c.LastName = 'contact';
+        c.AccountId = a.Id;
+        insert c;
+
+        npe03__Recurring_Donation__c r1 = new npe03__Recurring_Donation__c();
+        r1.Name = 'test';
```
force2b: For recurring donation names, I suggest using the name of the method rather than 'test' to ensure uniqueness

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
...
+
+        UTIL_CustomSettingsFacade.getRecurringDonationsSettingsForTest(
+                new npe03__Recurring_Donations_Settings__c(
+                        // npe03__Opportunity_Forecast_Months__c = 12,
+                        npe03__Maximum_Donations__c = 50,
+                        npe03__Open_Opportunity_Behavior__c = RD_RecurringDonations.RecurringDonationCloseOptions.Mark_Opportunities_Closed_Lost.name()
+                ));
+
+        Account a = new Account();
+        a.Name = 'test Individual';
```
force2b: For Account names, I suggest using the name of the method rather than 'test' to ensure uniqueness.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
         integer cOpp =  [select count() from Opportunity where npe03__Recurring_Donation__r.Name = 'testweekly'];
         system.assert(cOpp == 52 || cOpp == 53);        
     }
-    
+
+    static testMethod void updateFixedLengthRecurringDonationIncrease(){
+        if (strTestOnly != '*' && strTestOnly != 'updateFixedLengthRecurringDonationIncrease') return;
```
force2b: Now that we can run individual test methods directly through the IDE and DevConsole, this old pattern of `if (strTestOnly != '*' && strTestOnly != ....` is no longer necessary.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -469,26 +469,354 @@ public class RD_RecurringDonations_TEST {
...
+        System.assertEquals(2, originalOpps.size(), 'Two opportunities should be created initially.');
+        System.assertEquals(100, originalOpps[0].Amount);
+        Test.startTest();
+        originalOpps[0].StageName = closedStage;
+        update originalOpps[0];
+        r1.npe03__Installments__c = 4;
+        update r1;
+        Test.stopTest();
+
+        errors = [SELECT Id, Full_Message__c FROM Error__c];
```
force2b: I see a number of these queries throughout the class as well. This could be refactored into a method call
`errors = getErrorRecords();` or even something like `assertNoErrors();` where that method does the query and asserts the result is zero records.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -501,19 +829,57 @@ public class RD_RecurringDonations_TEST {
...
+                                            WHERE npe03__Recurring_Donation__c = :r1.id];
+
+        System.assertEquals(postUpdateOpps.size(), originalOpps.size(), 'No opportunities should have been deleted.');
+
+        for(Integer i; i<originalOpps.size(); i++) {
+            System.assertEquals(originalOpps[i].Id, postUpdateOpps[i].Id, 'The original opportunity records should persist.');
+        }
+
+        // these two integers, 12 and 13, are likely fragile
+        // TODO: refactor with hard-coded start date so it's stable
```
force2b: What does the TODO refer to here since the code above does use a fixed date for `Date_Established__c`?

src/classes/RD_RecurringDonations.cls
```diff
@@ -91,7 +91,8 @@ public class RD_RecurringDonations {
 
     /*******************************************************************************************************
     * @description holds the current custom field mappings used when creating opps for the RD.
-    */ 
+    */
+    @testVisible
```
force2b: Is `@testVisible` needed if this is already defined a `public`; or should this actually be `private`?

src/classes/RD_RecurringDonations.cls
davidhabib: I'm not sure we can assume recordTypeId exists on opp.  I think if no recordtypes have been defined, then the field doesn't exist, thus why this code uses put().

force2b: I'm almost positive that most of the codebase assumes that there are Opportunity RecordTypes defined; or at least that the RecordTypeId field exists. I did a search for "RecordTypeId" and there are lots and lots of results where that field is accessed on the Opportunity object.

davidhabib: if you are seeing other npsp code using opp.RecordTypeId, then I stand corrected!  thanks for checking.  I only looked in the RD file.

src/classes/RD_RecurringDonations.cls
```diff
@@ -91,7 +91,8 @@ public class RD_RecurringDonations {
 
     /*******************************************************************************************************
     * @description holds the current custom field mappings used when creating opps for the RD.
-    */
```
force2b: Does the comment block at the top of this class that explains the various behaviors need to be updated?

bethbrains: Yes absolutely.

src/classes/RD_RecurringDonations.cls
bethbrains: Didn't even look at this one as it wasn't part of this refactor.

src/classes/RD_RecurringDonations_TEST.cls
bethbrains: Thinking through this more, it is stable. It does use a dynamic npe03__Next_Payment_Date__c, but months are stable, whereas weeks are not, hence the || in line 844. Removed the comments.

src/classes/RD_RecurringDonations.cls
bethbrains: Updated ApexDocs header to clarify that this *does* apply to both
    * non-closed (still open) Open-Ended Recurring Donations (status = "Open")
    * and Fixed Length Recurring Donations (status = None or null).

src/classes/RD_RecurringDonations.cls
bethbrains: Refactored into utility.

src/classes/RD_RecurringDonations.cls
bethbrains: Fixed to use o.RecordTypeID = rds.npe03__Record_Type__c; since we can now assume as have RTs configured.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,277 +721,463 @@ public class RD_RecurringDonations {
...
     */ 
     public static void handleRecurringDonationUpdate(list<npe03__Recurring_Donation__c> recurringDonations, map<id, npe03__Recurring_Donation__c> oldRecurringDonations, TDTM_Runnable.DmlWrapper dmlWrapper){     
-        //opening a closed recurring donation
-        set<id> reevaluateOpps = new set<id>();
-        list<npe03__Recurring_Donation__c> reinsertOpps = new list<npe03__Recurring_Donation__c>(); 
-        // closed an open recurring donation
-        list<npe03__Recurring_Donation__c> newlycloseddonations = new list<npe03__Recurring_Donation__c>();
-         
+
+        System.debug('+++++Hitting handleRecurringDonationUpdate');
```
force2b: All `System.debug` statements should be removed. If they're really needed for some level of future debugging, they should be replaced with UTIL_Debug.debug().

src/classes/RD_RecurringDonations.cls
force2b: Since this is new code, can you update the var/type first letter case on these?

src/classes/RD_RecurringDonations.cls
```diff
@@ -1214,4 +1326,100 @@ public class RD_RecurringDonations {
...
+        //evaluate any custom field mappings
+        for (String s : customFieldMappingSettings.keySet()){
+            npe03__Custom_Field_Mapping__c cfm = customFieldMappingSettings.get(s);
+            o.put(cfm.npe03__Opportunity_Field__c, r.get(cfm.npe03__Recurring_Donation_Field__c));
+        }
+
+        return o;
+    }
+
+    private static String getOpportunityName(Opportunity o, npe03__Recurring_Donation__c r) {
```
force2b: This method needs an apex doc header.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -467,31 +441,403 @@ public class RD_RecurringDonations_TEST {
         system.assertEquals(4, [select count() from Opportunity where npe03__Recurring_Donation__r.Name = 'testquarterly']);
         system.assertEquals(1, [select count() from Opportunity where npe03__Recurring_Donation__r.Name = 'testyearly']);
         integer cOpp =  [select count() from Opportunity where npe03__Recurring_Donation__r.Name = 'testweekly'];
-        system.assert(cOpp == 52 || cOpp == 53);        
+        //depending on the day of week of the start date, there could be 52 or 53 in a given calendar year
+        system.assert(cOpp == 52 || cOpp == 53);
     }
-    
+
+    static testMethod void updateFixedLengthRecurringDonationIncrease(){
```
force2b: This method should have an apex doc header describing what's being tested.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -467,31 +441,403 @@ public class RD_RecurringDonations_TEST {
...
+            System.assertEquals(originalOpps[i].Id, postUpdateOpps[i].Id, 'The original opportunity records should persist.');
+        }
+
+        for(Integer i=0; i<postUpdateOpps.size(); i++) {
+            System.assertEquals(system.label.npe03.RecurringDonationStageName,postUpdateOpps[i].StageName, 'All opportunities should still be open.');
+        }
+
+    }
+
+    static testMethod void updateFixedLengthRecurringDonationIncreaseWithClosedOpps() {
```
force2b: This method should have an apex doc header describing what's being tested.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -467,31 +441,403 @@ public class RD_RecurringDonations_TEST {
...
+        for(Integer i; i<originalOpps.size(); i++) {
+            System.assertEquals(originalOpps[i].Id, postUpdateOpps[i].Id, 'The original opportunity records should persist.');
+        }
+        System.assertEquals(closedStage, postUpdateOpps[0].StageName, 'First Opp should still be Closed Won.');
+        for(Integer i=1; i<postUpdateOpps.size(); i++) {
+            System.assertEquals(system.label.npe03.RecurringDonationStageName,postUpdateOpps[i].StageName, 'Opportunities 2-4 should still be open.');
+        }
+    }
+
+    static testMethod void updateFixedLengthRecurringDonationDecrease() {
```
force2b: This method should have an apex doc header describing what's being tested.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -467,31 +441,403 @@ public class RD_RecurringDonations_TEST {
...
+        System.assertEquals(system.label.npe03.RecurringDonationStageName,postUpdateOpps[1].StageName, 'Opportunity 2 should still be open.');
+
+        System.assertEquals(25, postUpdateOpps[2].Amount, 'Opportunity 3 should still be $25.');
+        System.assertEquals(system.label.npe03.RecurringDonationClosedLostOpportunityStage, postUpdateOpps[2].StageName, 'Opportunity 3 should be Closed Lost.');
+
+        System.assertEquals(25, postUpdateOpps[3].Amount, 'Opportunity 4 should still be $25.');
+        System.assertEquals(system.label.npe03.RecurringDonationClosedLostOpportunityStage, postUpdateOpps[3].StageName, 'Opportunity 4 should be Closed Lost.');
+    }
+
+    static testMethod void updateFixedLengthRecurringDonationDecreaseWithTooManyClosedOpps() {
```
force2b: This method should have an apex doc header describing what's being tested.

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -1100,6 +1514,103 @@ public class RD_RecurringDonations_TEST {
         system.assertEquals(system.today(), rd.npe03__Last_Payment_Date__c);
     }
 
+    static testMethod void testNeedsRecalculation() {
```
force2b: This method should have an apex doc header describing what's being tested.

src/classes/RD_RecurringDonations.cls
```diff
@@ -1354,41 +1375,52 @@ public class RD_RecurringDonations {
...
-        o.Amount = r.npe03__Amount__c;
-
-        o.Recurring_Donation_Installment_Number__c = rdcounter;
-
-        o.Name = getOpportunityName(o, r);
+        if ( r.npe03__Organization__c != null ) {
+            opp.AccountId = r.npe03__Organization__c;
+        } else if ( con != null ) {
+            opp.AccountId = con.AccountId;
+            if (ConIDForRole != null){
```
force2b: Is there a reason that the `npe01__Contact_Id_For_Role__c` would not exist on the Opportunity object?

bethbrains:
This is how it was in legacy, so I kept it. Not sure if there was a good
reason.

src/classes/RD_RecurringDonations.cls
```diff
@@ -795,277 +725,467 @@ public class RD_RecurringDonations {
...
+                , AccountID
+                , CloseDate
+                , StageName
+                , isClosed
+                , isWon
+                , Recurring_Donation_Installment_Number__c
+                , npe03__Recurring_Donation__c
+        FROM Opportunity
+        WHERE npe03__Recurring_Donation__c IN :openEndedRDsToRecalculate.keySet()
+        OR npe03__Recurring_Donation__c IN :fixedLengthRDsToRecalculate.keySet()
```
force2b: You could rewrite the filter as follows so that the query only retrieves the records needed versus retrieving all the records for those RD's. I'm thinking about RD's that might have more than 50K attached Opportunities here, so even though adding the filters on isClosed/isWon are on non-indexed fields, I suspect reducing the number of queried records will help with overall performance.

`WHERE (npe03__Recurring_Donation__c IN :openEndedRDsToRecalculate.keySet() AND IsClosed = False)
OR npe03__Recurring_Donation__c IN :fixedLengthRDsToRecalculate.keySet()`

If this should included ClosedLost opps for openEndedRDsToRecalculate then the query would be:
`WHERE (npe03__Recurring_Donation__c IN :openEndedRDsToRecalculate.keySet() AND (IsClosed = False OR (IsClosed = True AND IsWon=False))
OR npe03__Recurring_Donation__c IN :fixedLengthRDsToRecalculate.keySet()`

src/classes/RD_RecurringDonations.cls
force2b: @bethbrains Any thoughts on this one?

bethbrains: Ah, my email reply didn't work, sorry.
This is how it was in the legacy code, so I kept it. Not sure if there was a good reason.

force2b: Hard to tell what's new and what's old code for some of this. There's no reason for it to be this way, so if you're going to do another commit to fix the unit teste I'd change this and delete the code that creates the `ConIDForRole` var.

src/classes/RD_RecurringDonations.cls
```diff
@@ -1214,4 +1279,137 @@ public class RD_RecurringDonations {
...
+        if(isMultiCurrency){
+            //set the donation currency field to equal the recurring donation currency
+            opp.put(OppCurrencyField,r.get(RDCurrencyField));
+        }
+
+        if ( r.npe03__Organization__c != null ) {
+            opp.AccountId = r.npe03__Organization__c;
+        } else if ( con != null ) {
+            opp.AccountId = con.AccountId;
+            opp.npe01__Contact_Id_For_Role__c = con.id;
```
njjc: Could you add Primary_Contact__c here as well?

bethbrains: done

sam-knox:
@bethbrains - can you take a minute to flesh out the release notes for this PR? I can edit and finalize, but wanted to make sure all the aspects of this PR are documented.

PR #2958, opened 2018-02-21
Feature/jen cci qa fields

jjbennett:
Creating qa install flows that will add fields and relationship data used for testing purposes only.
 
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -184,6 +184,7 @@ flows:
                 task: deploy_dev_config
 
     qa_org:
+        description: Deploys the unmanaged package metadata with additional fields for qa and all dependencies to the target org
             0.5:
                 task: unschedule_apex
             1:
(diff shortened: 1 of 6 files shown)


orgs/beta_wave.json
```diff
@@ -0,0 +1,16 @@
+{
+  "orgName": "NPSP - Beta Org - Analytics",
```
cdcarter: this org needs a mapping in cumulusci.yml, nah?

cumulusci.yml
```diff
@@ -177,6 +183,76 @@ flows:
             9:
                 task: deploy_dev_config
 
+    qa_org:
```
cdcarter: please add a description to this flow as well

jjbennett: Thanks @cdcarter, I updated the descriptions on the qa flows and added the wave org mapping to the yml file.

cdcarter: i think this is a syntax error failing the build

PR #2971, opened 2018-02-26
Fix Invalid Record Type Settings 

### LINKED ISSUE #2862: Health Check needs to gracefully handle invalid ID's
hub post: https://powerofus.force.com/0D58000003nLqrT

invalid grant id is causing Health Check to throw an exception, rather than detected and reported as a specific Health Check error.

### LINKED ISSUE #2890: When an existing customer upgrades from NPSP 2 and has Record Type values by name in Custom Settings, they're no longer getting converted to IDs
There have been multiple reports of this in the Hub over the past few weeks. This is causing the inability to load NPSP Settings, errors in Health Check, etc.

Example [here](https://powerofus.force.com/0D58000003s0jTX) and [here](https://powerofus.force.com/0D58000003rjlhu) and [here](https://powerofus.force.com/0D58000003nNGD8) and [here](https://powerofus.force.com/0D58000003nLqrT). 

resolution:

Click Manage to the left of Contacts & Orgs Settings. Make sure 'Excluded Opp Record Types' 'Household Account Record Type' and 'One-to-One Record Type' do not have the word 'Membership.'

If that's good, go to Households Settings, click Manage to the left of that and check for the word 'Membership' in 'Household Creation Excluded Record Types' 'Membership Record Type Names' (this is where I think the problem may be) 'Opp RecTypes Excluded for Account Rollup' 'Opp RecTypes Excluded for Contact Rollup' or 'Household OCR Excluded Record Types'

Basically, NPSP now requires an ID in those fields rather than a name. I suspect that for whatever reason, you had the name and it's not mapping correctly to the record type so it didn't update to the correct name.

RadSFDeveloper:
# Critical Changes
We are fixing an unexpected exception(s) that is generated when at least one Custom Setting has a Record Type field that contains names instead of Ids. 

A System Administrator can now clean up the invalid Record Type Custom Settings by visiting the NPSP Settings page. You should also run Health Check to identify any other Record Type errors and get advice on how to fix them.
# Changes

# Issues Closed
Fixes #2890
Fixes #2862

# New Metadata

# Deleted Metadata


src/classes/ERR_Handler.cls
```diff
@@ -261,6 +262,16 @@ public class ERR_Handler {
             processErrors(new Error__c[]{error}, context);
         }
     }
+    
+    
+    public static void processErrors(List<Exception> exceptions, ERR_Handler_API.Context context) {
```
davidhabib: please add a header comment for this new method.

src/classes/ERR_Handler.cls
```diff
@@ -276,7 +287,6 @@ public class ERR_Handler {
             }
 
             if (UTIL_CustomSettingsFacade.getErrorSettings().Store_Errors_On__c) {
-               UTIL_Debug.debug('**** Storing errorRecords: ' + errorRecords);
                Database.insert(errorRecords, false);
```
davidhabib: could you put this UTIL_Debug.debug() back.  we always remove system.debug calls.  but UTIL_Debug calls are going thru our cover that has an NPSP Setting to turn on/off, so that customers can enable it to get more info in debug logs to help figure out problems.  This one specifically is an important one to keep enabled, since it allows us to get the actual error record into the log.

src/classes/ERR_Handler.cls
```diff
@@ -294,17 +304,19 @@ public class ERR_Handler {
         error.Datetime__c = System.now();
         error.Error_Type__c = e.getTypeName();
         error.Full_Message__c = e.getMessage();
+
         //Putting it into a local string to see if that helps storing the stack trace when we
         //are in a managed package.
         String stackTrace = e.getStackTraceString();
-        UTIL_Debug.debug('****Stack trace: ' + stackTrace);
+
```
davidhabib: this is another one to put back in, since it can help us debug customer issues.

davidhabib:
@RadSFDeveloper for the settings that are the list of recordtypeId's, when I manually test, if I put in one valid name and one invalid name (grant;foo or foo;grant), then the entire settings gets cleared out.  Wasn't the fixup code trying to keep valid names?

davidhabib:
I did verify that if the recordtype settings contain valid and invalid id's, only the invalid id's are dropped.  so maybe for the case of the initial upgrade to 3.0, if any recordtype name is invalid, we clear them all.  @judisohn what do you think?  worth extra coding to try to preserve the ones that are still valid, or better to have them empty so the admin knows they need to reset them?

davidhabib:
@RadSFDeveloper other than my question for Judi, which we can probably live with, the pull request looks great!  once we decide on that, I can merge if appropriate.  @njjc no need for you to review as well I think.  I did the full code review, ran tests, and did manual testing.

judisohn:
@davidhabib I think there's enough time that an upgrader needs to spend in NPSP Settings immediately after they upgrade, that if the entire field is 'Error' and they have to update them all, it's fine. They may not realize that previous values are missing otherwise.

RadSFDeveloper:
@davidhabib , this is the current *unchanged* behaviour for the Record Type Settings clean up:

1. Invalid names are deleted from the field.
2. Valid Record Type names are converted into Ids
3. Valid Record Type Ids are kept as they are.
4. *Nonexisting but still valid* Ids are kept as they are.
5. Invalid Ids are deleted from the field.
6. Record Type names are case sensitive

If it is only one Record Type name expected, then records are not split by ";" and the whole name as it is tried to be matched.

For the case above, when one invalid and one valid Record Type was provided:
- Is the valid Record Type case sensitive correct?
- Does this Settings field expect one or more Record Type names/Ids?

davidhabib:
yes, the settings I tested that had a mixture of good and bad, were expecting multiple record types.  But I'm ok that we leave this as is (ie, any bad record type name clears the whole field), since Judi is fine with it, and I do believe it will make it more obvious to the admin that they need to edit the setting.  we also send them the email, so it should not go unnoticed.  i'm ok merging this as is now.

davidhabib:
@RadSFDeveloper what user strings are you asking @smarovich to review?  you didn't add any new ones that I recall.

RadSFDeveloper:
@davidhabib , we will only send them an email on the exception. We will not send the email containing which Record Type Settings fields were changed and how (unless I am missing something?).

davidhabib:
@RadSFDeveloper oops.  you are right.  we only do the email if errors occur during the install script.  the invalid id's are silently fixed or removed.  thanks for clarifying.  so @judisohn just so we are all on the same page, when the fixup occurs during install or going to npsp settings, and there are invalid names, then the whole setting will be cleared to empty.  so they won't see an error in NPSP settings; they will hopefully notice that it is empty and they should respecify their recordtypes.  ok?

RadSFDeveloper:
@davidhabib , I added @smarovich as reviewer since I thought that is required when a new PR is required. I have not created any comment for her though, only Changes section is relevant.
Moreover, should we revise the email we send on the installation steps failure (the email was changed to send all exceptions instead of just one), or the code is good as it is?

davidhabib:
I think the email is fine as you now have it.  no need to call out that change.  Got it about adding Sarah.  it was just to review the release notes, not for any content within the pull request.  thanks.

judisohn:
@davidhabib Yes, I understand and I think it's fine to remove all values if even one is invalid.

RadSFDeveloper:
@davidhabib , which Record Type Setting field ended up having blank value when one value is correct and the other incorrect? What were the values?
In my testing, if the name is split by ";" and then each name validated and the valid ones are saved. So, for example if the Household Settings "Membership Record Type Names" field has "Membership;InvalidName", then when the System Admin loads the NPSP Settings page, the field has the Id of the Membership Record Type Id.

davidhabib:
@RadSFDeveloper I tested the following settings under Household Settings:
Membership Record Type Names
Opp RecTypes Excluded for Account Rollup
Opp RecTypes Excluded for Contact Rollup

I set these manually under Setup, Custom Settings to:
membership;foo
grant;foo
foo;grant

and I have Opp record types Membership, Grant, NPSP Default.

when going to the NPSP Settings page, all three of the custom settings get cleared out to null.
just tested again spelling Membership and Grant with the correct case, and they are preserved.  So obviously our name matching isn't case insensitive and is treating them as invalid.

seems worth fixing this trivial fix to do case insensitive matches?

davidhabib:
@RadSFDeveloper can't seem to find this comment now so readding it (ignore me if dupe).  the namespaced tests are failing, probably due to some of the new code not using our namespace fixup utilities.

STG_PanelHealthCheck_TEST.testVerifyRecordTypes - Fail
2018-02-26 10:42:20: 	Message: npsp.UTIL_Describe.SchemaDescribeException: Invalid object name 'Allocations_Settings__c'
2018-02-26 10:42:20: 	StackTrace: Class.npsp.UTIL_Describe.fillMapsForObject: line 65, column 1

RadSFDeveloper:
@davidhabib , I will work on making the Record Type Settings case insensitive.
I will also check into the namespace issue, thanks for letting me know (the comment was created to a branch Kofi is/was working on, but I read it so it is fine).

RadSFDeveloper:
Btw, @davidhabib , I demoed the case sensitive (unchanged) Record Type Settings update behaviour, and @judisohn mentioned it is fine since clients needed to specify the correct case sensitive information before. Just to mention that in the case we need to keep the behaviour as it is.

RadSFDeveloper:
@davidhabib , the change has been done to validate Report Type Settings as case insensitive values.
Also, the namespace consideration was added when record type values are validated. The namespaced build will confirm this shortly.

RadSFDeveloper:
@judisohn , should we notify clients and specify details in the Critical Changes so their System Admin loads the NPSP Settings page in order to automatically fix the Record Type Settings containing names and then run the Health Check to ensure there is no Record Type Settings that need the manual check?
@smarovich , @davidhabib

smarovich:
@RadSFDeveloper @judisohn @davidhabib I just updated the release notes and moved the changes into the Critical Changes section, but let me know if I should move them back.

src/classes/UTIL_RecordTypeSettingsUpdate.cls
```diff
@@ -261,42 +263,46 @@ global virtual class UTIL_RecordTypeSettingsUpdate {
...
     * @description Determines the Record Type Id based on the name
      * @param name A Record Type name
-     * @param rtNamesById A mapping of Record Type names to Record Type Ids
+     * @param nameMap A mapping of Record Type names to Record Type Ids
      * @return String An Id if the name is a valid Record Type name, otherwise null.
     */
-    private String getIdFromName(String name, Map<String, Id> rtNamesById) {
-        if (String.isBlank(name)) {
+    private String getIdFromName(String name, Map<String, Id> nameMap) {
+        if (name == null) {
```
davidhabib: why not continue to use isBlank()?  did the code previously handle ignoring spaces between semicolons that now it won't?

RadSFDeveloper:
@davidhabib , with regards to check for the null in the `UTIL_RecordTypeSettingsUpdate` , there is a test that specifically test the field being returned as null if it is passed in as the null: https://github.com/SalesforceFoundation/Cumulus/pull/2971/files#diff-478888b6314e4db421930620fe08aac6R60

judisohn:
@RadSFDeveloper I think it's fine if it's in the release notes and we don't necessarily have to reach out separately if that's what you're asking. Hopefully we're getting folks in the habit of running to Health Check when they get an error as the first troubleshooting step, and that's certainly what support would point them towards.

PR #2977, opened 2018-02-28
Feature/rollups  user interface sprint 2a

randi274:
# Critical Changes

# Changes
- Added Filter Group detail page so users can view a specific filter group and its associated filter rules.
- Revised the New Rollup flow to walk users through the process of creating a rollup__mdt record in a step-by-step manner.

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@randi274 Can you add some details to the top of the PR comment with what new functionality this PR brings to the UI?

force2b:
@randi274 I noticed that most of the equality and inequality operators were fixed, but there are still remaining items and some other minors issues. I suggest using the linter to address whatever is left over.
`sfdx force:lightning:lint src/aura | grep -v "console"`

The `grep` part removes the warnings about console.log usage. Normally, there's a linter config file that can be used to strip that out, but Salesforce considers those warnings critical so they don't allow those to be ignored. These will have to be stripped out eventually, but there's no rush on that.

src/aura/CRLP_Rollup/CRLP_RollupHelper.js
```diff
@@ -63,278 +82,643 @@
...
+                cmp.set("v.isReadOnly", false);
+            } else if (mode === "create") {
+                cmp.set("v.isIncomplete", true);
+                cmp.set("v.isReadOnly", false);
+                this.hideAllFields(cmp);
+            }
+        }
+    },
+
+    filterFieldsByType: function(cmp, typeList, allFields) {
```
force2b: This applies to all the javascript, but I'm dropping the comment here. I'd like to have all of the JS methods documented the same way our apex methods are documented. That will make it much easier to support the code long term. For example:

```
/*
* @description: {describe what this function does}
* @param typeList {...}
* @param allFields {...}
*/
filterFieldsByType: function(cmp, typeList, allFields) {
```

src/aura/CRLP_Rollup/CRLP_RollupHelper.js
```diff
@@ -63,278 +82,643 @@
...
+
+        var type;
+        for (var i = 0; i < summaryFields.length; i++) {
+            if (summaryFields[i].name === field) {
+                type = summaryFields[i].type;
                 break;
             }
         }
-        return label;
+        console.log(type);
```
force2b: Testing this in a scratch org and found that for the "GAU: Count of Allocation Last Year" rollup, the Operation field is blank on the View and Edit pages. This appears to be because the `type` var is undefined, so the code below isn't able to include the various operations for the numeric field. I've noticed that Operation is blank on other 'View' pages as well, but shows up when edit is clicked.

randi274: Thanks for the reminder on this! It did get fixed on the next branch, but it required a few changes that I'd like to keep separate since they were part of a larger change.

PR #2981, opened 2018-03-07
Change Allocations from percentage-based to amount-based

### LINKED ISSUE #2921: Percentage-based Opportunity GAU Allocations can't be updated to Amount-based
[Thread.](https://powerofus.force.com/0D58000003vfSSf)

Confirmed in scratch org.

1. Create 2 GAU records
2. Create Campaign and split 50% between each GAU
3. Create Opportunity (closed/won, but stage shouldn't matter) for Amount = $100, Primary Campaign source set to above
4. 'Manage Allocation' and delete % and change first to 90, second to 10. Save.

image

5. Note that the GAU Allocation related list retains the original 50% split

image

If the percentage is changed (for example 80/20) then the change is saved.

njjc:
# Critical Changes

# Changes
- GAU Allocations now allow percentages with up to 6 decimal places, up from 2.

# Issues Closed
- Fixed #2921

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/objects/Allocation__c.object
+++ b/src/objects/Allocation__c.object
@@ -126,9 +126,9 @@
         <externalId>false</externalId>
         <inlineHelpText>Percent of opportunity amount to allocate to this general accounting unit. Modifying an opportunity amount of a percent based allocation will modify the allocation amount.</inlineHelpText>
         <label>Percent</label>
-        <precision>5</precision>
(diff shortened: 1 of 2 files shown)


njjc:
@jjbennett fixed this bug in a better way in #2980

PR #2984, opened 2018-03-09
Rollups: NPSP Settings UI to enable Customizable Rollups

bethbrains:
# Critical Changes

# Changes

- Customizable Rollups panel in NPSP Settings allows admin to enable new CRLP Engine
- Updated notifications when disabling System Administrators whose user accounts are tied to critical system functions to also include custom System Administrator profiles

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -1,19 +1,158 @@
...
 
-public with sharing class STG_PanelCustomizableRollup_CTRL {
+public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
 
-    public static void enableCRLP(){
-        //method to enable CRLP
+    /*******************************************************************************************************
+    * @description is the panel currently Polling for deployment results
+    */
+    public Boolean isPolling {
```
force2b: @bethbrains You don't have to change this if you don't want to, but there are a couple of other ways to approach this. First, if `isPolling` is supposed to be read only on the VF page, then technically set should be private: `private set;`. Alternatively, you could write this method as `public Boolean getisPolling() { return ...; }`. That would implement only the getter for `isPolling`.

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -1,19 +1,158 @@
...
+    */
+    public Boolean isAdmin {
+        get { return STG_SettingsManager_CTRL.runningUserIsAdmin(); }
+        set;
+    }
+
+    /*******************************************************************************************************
+    * @description the jobId of the deployment
+    */
+    public String jobId {public get; set;}
```
force2b: Technically `public` isn't needed for `get` because the `jobId` var itself if public. However, you could add `private` to the `set` to prevent the var from being modified from outside the class.

src/classes/STG_SettingsManager_CTRL.cls
```diff
@@ -102,7 +107,7 @@ public with sharing class STG_SettingsManager_CTRL extends STG_Panel {
     * @description Returns whether the running user is a System Administrator.
     * @return void
     */
-    private boolean runningUserIsAdmin() {
+    public static boolean runningUserIsAdmin() {
```
force2b: Since this method is now being changed to `public`, there is a similar method in `USER_UserService.isSysAdmin`. The logic implemented by these two methods are quite different however. @njjc Wondering if the method name on one or both of these should be changed to further clarify how the methods differ; and possibly if this method should be moved to `USER_UserService`. What do you think?

src/labels/CustomLabels.labels
```diff
@@ -4628,6 +4628,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <language>en_US</language>
+        <protected>false</protected>
+        <shortDescription>stgCRLPHouseholdAccountError</shortDescription>
+        <value>You must be using the Household Account Model in order to use Customizable Rollups.</value>
+    </labels>
+    <labels>
+        <fullName>stgCRLPNonAdminError</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>false</protected>
```
force2b: I believe all custom labels should be set with protected = true. @jjbennett Can you confirm that?

src/pages/STG_SettingsManager.page
```diff
@@ -71,6 +81,25 @@
         if (idPanelCurrent != '{!idPanelCurrent}')
             setPanelCurrent(idPanelCurrent);
     }
+
+    function reRenderTree(isCRLPEnabled) {
+        if(isCRLPEnabled === undefined) {
+            isCRLPEnabled = document.querySelector('[id$="cbxCRLPE"]').checked;
+        }
+        console.log('isCRLPEnabled: '+isCRLPEnabled);
```
force2b: Don't forget to remove `console` statements before this is merged into Master.

src/labels/CustomLabels.labels
jjbennett: Confirmed - essentially it makes the class public to our namespace but private to others and setting it to "false" would be similar to making it global.

src/classes/STG_SettingsManager_CTRL.cls
njjc: Better names and centralizing in the service sounds like a good idea to me!

src/classes/CRLP_DefaultConfigBuilder_SVC.cls
```diff
@@ -45,6 +45,15 @@ public class CRLP_DefaultConfigBuilder_SVC {
             WHERE SobjectType = 'Opportunity' AND IsActive = True]);
 
 
+    /*************************************************************************************************************
+     * @description Call each of the methods in this class to build the default Filter Groups and each of the four
+     * Rollup Types (Account Hard Credit/Membership, Contact Hard Credit/Membership, Contact Soft Credit, and GAU)
+     */
+    @future
+    public static void convertLegacyRollupsIntoCustomizableRollupsFuture() {
```
njjc: This isn't needed anymore as we're updating the custom settings in the callback, right?

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -1,19 +1,165 @@
...
+        }
+        private set;
     }
 
-    public PageReference navigate(){
+    /*******************************************************************************************************
+    * @description is the org on the HH Account model
+    */
+    public Boolean isHHAccount {
+        get { return (STG_Panel.stgService.stgCon.npe01__Account_Processor__c == CAO_Constants.HH_ACCOUNT_PROCESSOR); }
```
njjc: I think stylistically we want curly brackets to always open a new indented line, even when there's only a single line inside the brackets.

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -1,19 +1,165 @@
...
+        }
+        return null;
+    }
+
+    /*******************************************************************************************************
+    * @description Action Method to navigate to separate VF page that contains custom CMDT UI in Lightning Components
+    * @return void
+    */
+    public PageReference navigate(){
         //TODO: check this will work with upcoming URL changes
```
njjc: Can we check whether this works yet?

njjc:
@bethbrains could you solve the merge conflicts and let me know about the inline comments? Thank!

src/classes/STG_SettingsManager_CTRL.cls
bethbrains: done and pushed.

src/labels/CustomLabels.labels
```diff
@@ -4782,6 +4798,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>stgDocumentation</shortDescription>
         <value>Documentation</value>
     </labels>
+    <labels>
+        <fullName>stgDeploymentInProgress</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>false</protected>
```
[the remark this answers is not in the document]
bethbrains: good catch, MS. I'll make the change in my RD rollups branch.

PR #2994, opened 2018-03-15
Feature/automated soft credits  organizational gifts

joy-sfdo:
Automated Soft Credits: Admin defined setting handling -- W-023032. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata
- src/labels/CustomLabels.labels > stgHelpOrgOCRDefaultRole
- src/objects/npe01__Contacts_And_Orgs_Settings__c.object > Default_Organizational_Contact_Role__c

# Deleted Metadata


src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -473,13 +495,33 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
     * @param fieldName The name of the opportunity field.
     * @return OpportunityContactRole The newly created OCR.
     ********************************************************************************************************/
-    private OpportunityContactRole getOCR(Opportunity opp, string fieldName) {
+    private OpportunityContactRole getOCR(Opportunity opp, String fieldName) {
+        OpportunityContactRole ocr = new OpportunityContactRole();
```
davidhabib: rather than duplicating this whole method, why not just have getOCR(opp, fieldname) call 
getOCR(opp, fieldname, null), so there is only one implementation.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -495,21 +537,50 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+            return accountDetails;
+        }
+
+        for (Opportunity eachOppty : currentOpportunities) {
+            if (eachOppty.AccountId != null) {
+                accountIds.add(eachOppty.AccountId);
+            }
+        }
+
+        accountDetails = new Map<Id, Account>([SELECT Id, Name, RecordTypeId, npe01__SYSTEMIsIndividual__c
```
davidhabib: as mentioned earlier, let's see if this method and extra query is really necessary.

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -92,9 +94,13 @@ private class OPP_OpportunityContactRoles_TEST {
         );
         insert opp2;
 
-        result = [select OpportunityId, ContactId, Role from OpportunityContactRole where OpportunityId = :opp2.Id];
+        result = [SELECT OpportunityId, ContactId, Role FROM OpportunityContactRole WHERE OpportunityId = :opp2.Id];
         system.assertEquals(con.Id, result[0].ContactId);
-        system.assertEquals(donorRoleforTest, result[0].Role);
+        if (OPP_AutomatedSoftCreditsService.isOrganizationalAccount(acc.npe01__SYSTEMIsIndividual__c)) {
```
davidhabib: since you are explicitly setting acc.npe01__SYSTEMIsIndividual__c = false at the beginning of this test method, this if should always be true, and the else clause will never get tested.  Shouldn't we have two separate tests; one for household account donations, and one for organizational donations.  they each tests could assert that the correct role is created.

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -144,7 +150,7 @@ private class OPP_OpportunityContactRoles_TEST {
         Contact con = UTIL_UnitTestData_TEST.getContact();
         insert con;
         
-        Contact[] createdContacts = [select AccountId from Contact where Id = :con.id];
+        Contact[] createdContacts = [SELECT AccountId, Account.npe01__SYSTEMIsIndividual__c FROM Contact WHERE Id = :con.id];
```
davidhabib: why the need to add acc.npe01__SYSTEMIsIndividual__c to the query?  you aren't testing it later.  did you mean to add an assert?

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -177,7 +183,7 @@ private class OPP_OpportunityContactRoles_TEST {
...
```
davidhabib: I don't see any assert using acc.npe01__SYSTEMIsIndividual__c

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -283,7 +283,7 @@ public class RD_RecurringDonations_TEST {
             npe01__Opportunity_Contact_Role_Default_role__c = system.label.npe03.RecurringDonationContactRole
         ));
             
-        CAO_Constants.setIndividualAccountForTests(CAO_Constants.INDIVIDUAL_ACCOUNT_NAME_FOR_TESTS);
+        Id accountId = CAO_Constants.setIndividualAccountForTests(CAO_Constants.INDIVIDUAL_ACCOUNT_NAME_FOR_TESTS);
```
davidhabib: accountId doesn't seem to be used anywhere

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -309,7 +309,9 @@ public class RD_RecurringDonations_TEST {
             WHERE Opportunity.npe03__Recurring_Donation__c = :r1.id
         ];
         system.assertEquals(2,installmentCRs.size());
-        system.assertEquals(system.label.npe03.RecurringDonationContactRole,installmentCRs[0].Role);
+        if (!OPP_AutomatedSoftCreditsService.isOrganizationalAccount(c.Account.npe01__SYSTEMIsIndividual__c)) {
```
davidhabib: this test gets called three times, once with 1:1, once with HH Account, and once with Individual Bucket accounts.  So in all three cases, these are not Organizational accounts.  If we think this routine might get called with an org account, then I'd like to see an else assert checking for the other value.  If that isn't going to happen, then I'd rather just see an assert that these aren't org accounts, rather than have an if with no else.  make sense?

src/pages/STG_PanelContactRoles.page
```diff
@@ -23,6 +24,18 @@
...
+                <apex:outputLabel value="{!$ObjectType.npe01__Contacts_And_Orgs_Settings__c.Fields.Default_Organizational_Contact_Role__c.Label}" for="slOrgOCRDR" styleClass="slds-form-element__label" />
+                <div class="slds-form-element__control">
+                    <apex:outputField value="{!stgService.stgCon.Default_Organizational_Contact_Role__c}" rendered="{!isReadOnlyMode}" styleClass="slds-form-element__static" />
+                    <apex:selectList value="{!stgService.stgCon.Default_Organizational_Contact_Role__c}" multiselect="false" size="1" rendered="{!isEditMode}" id="slOrgOCRDR" html-aria-describedby="{!$Component.slOrgOCRDRHelp}" styleClass="slds-select" >
+                        <apex:selectOptions value="{!listSOContactRoles}"/>
+                    </apex:selectList>
+                    <apex:outputPanel id="slOrgOCRDRHelp" layout="block">
+                        <apex:outputText styleClass="slds-form-element__help" value="{!$Label.stgHelpOrgOCRDefaultRole}" />
+                    </apex:outputPanel>
+                </div>
```
davidhabib: I don't know why, but in the Settings page, the new org setting is pushed up against the help text of the preceding setting.  This doesn't happen if you look at the bottom two settings for Honoree and Notification.  Markup looks the same, but I may have missed something.  here's a pic to show what I mean:

image

davidhabib:
@joy-sfdo also gitHub is complaining there is a merge conflict in RD_Recurring_Donations_TEST.cls that needs to get resolved.

joy-sfdo:
Hi @davidhabib, thanks for the code review. I'll review your comments and update the code accordingly.

riordan:
I'm really looking forward to this!

Is there a place to look for the general spec of this feature as it's not linked to a prior github issue?

src/classes/OPP_AutomatedSoftCreditsService_TEST.cls
```diff
@@ -0,0 +1,94 @@
...
+*/
+
+@isTest
+private class OPP_AutomatedSoftCreditsService_TEST {
+
+	/*********************************************************************************************************
+    * @description If you only want to run a single test in this class, fill in its name here.
+    * If you want to run all tests, then use '*'
+    **********************************************************************************************************/
+    private static string strTestOnly = '*';
```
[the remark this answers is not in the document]
joy-sfdo: Good to know, thanks.

src/classes/OPP_AutomatedSoftCreditsService_TEST.cls
```diff
@@ -0,0 +1,94 @@
...
+
+
+    /*********************************************************************************************************
+    * @description Tests the Opportunity's Account has an Organization record type.
+    * verify:
+    *   Account's RecordType is Organization.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testOpportunityAccountRecordTypeIsOrganization() {
```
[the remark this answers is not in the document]
joy-sfdo: Good catch. Using RecordType was the original game plan.

src/classes/OPP_AutomatedSoftCreditsService_TEST.cls
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+
+    /*********************************************************************************************************
+    * @description Tests the Opportunity's Account does not have an Organization record type.
+    * verify:
+    *   Account's RecordType is not Organization.
+    * @return void
+    **********************************************************************************************************/
+    private static testMethod void testOpportunityAccountRecordTypeNotOrganization() {
```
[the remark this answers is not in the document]
joy-sfdo: Same as above. Thanks.

judisohn:
@riordan Nothing publicly available at this time. We'll be updating the pull request and related documentation as the feature nears release. Stay tuned. :-)

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -36,6 +36,11 @@
 */
 public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runnable {
 
+    private static final String FIELD_NAME_PRIMARY_CONTACT = 'Primary_Contact__c';
+    private static final String FIELD_NAME_HONOREE_CONTACT = 'Honoree_Contact__c';
+    private static final String FIELD_NAME_NOTIFICATION_RECIPENT_CONTACT = 'Notification_Recipient_Contact__c';
```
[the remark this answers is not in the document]
joy-sfdo: Good catch.

davidhabib:
@joy-sfdo I just noticed that this pull request is trying to go into Master, rather than a parent branch that will hold all of the reviewed Automated Soft Credit work, so once all the pieces are in place and we are ready to publish to customers, we just merge the one reviewed branch.

joy-sfdo:
@davidhabib -- That's interesting. I'll look into the branching.

davidhabib:
@joy-sfdo Sorry if I hadn't been clear enough last week when I mentioned this.  We want to follow the same mechanism that rollups team is using.  Basically you would have branches like this:

/feature/AutomatedSoftCredits
/feature/AutomatedSoftCredits/workitem1
/feature/AutomatedSoftCredits/workitem2
/feature/AutomatedSoftCredits/workitem3

when any workitem is ready for a pull request, it is to /featureAutomatedSoftCredits, not Master.  this allows us to formally review and test it and merge it (in order to close the work item), without it going to customers.  The build system is automated to then push any change to /feature/AutomatedSoftCredits into the child branches (/workitem1, /workitem2, /workitem3) just like when a change to Master updates all our feature branches. 

Once all the workitems are done, and we are ready to release to customers, all the code in /feature/AutomatedSoftCredits has already been reviewed, and we really just need to do some final testing before merging to master.

make sense?  I think Jason Lantz wrote up a doc on this, that you can probably find in quip.

joy-sfdo:
@davidhabib -- That's the process that I used to set up the branches. As I mentioned, when we spoke I had created the branches in that fashion. I spoke to Chris Polcyn about the process of setting up Parent and Child branches and followed the steps the in Quip document (https://salesforce.quip.com/8F5aAdZBaK1r). I'll follow up on the branches.

joy-sfdo:
@davidhabib -- Also, we can merge a branch into another branch as well.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -224,13 +231,15 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
             //cloned opportunity is necessary in order to modify the Primary Contact lookup and update the record in the trigger set
             Opportunity clonedOpp = opp.clone(true,true,false,true);
 
+            Boolean isIndividualAccount = (oppAccountDetails != null && oppAccountDetails.containsKey(opp.AccountId)) ? oppAccountDetails.get(opp.AccountId).npe01__SYSTEMIsIndividual__c : null;
```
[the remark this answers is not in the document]
joy-sfdo: Removed the null check as an empty Map will be returned from the populateAccountDetails method. 

I decided to leave the code that references the system field since it is the argument that is passed into the getRole and getOCR methods. Its usage makes the code more readable because it clearly shows how the isIndividualAccount variable is populated. I'm open to further discussion.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -325,33 +338,35 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
             for (integer i=0; i<listOpps.size(); i++) {
                 Opportunity newOpp = listOpps[i];
                 Opportunity oldOpp = listOldOpps[i];
+
+                Boolean isIndividualAccount = (oppAccountDetails != null  && oppAccountDetails.containsKey(newOpp.AccountId)) ? oppAccountDetails.get(newOpp.AccountId).npe01__SYSTEMIsIndividual__c : null;
```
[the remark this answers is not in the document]
joy-sfdo: Same as above.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -432,7 +446,7 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
             //contact lookup changed from one contact to another
             if (idOldCon != null) {
                 //an OCR for the old contact exists, and has the role for this field
-                if (ocrOldCon != null && ocrOldCon.Role == getRole(fieldName)) {
+                if (ocrOldCon != null && ocrOldCon.Role == getRole(fieldName, opp.Account.npe01__SYSTEMIsIndividual__c)) {
```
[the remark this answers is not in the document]
joy-sfdo: This call to the method did not get updated. Thanks for catching this.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
joy-sfdo: Glad to see this comment. I've updated the code accordingly.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
joy-sfdo: The query is necessary.

src/classes/OPP_OpportunityContactRoles_TEST.cls
joy-sfdo: I added a test since we now have different handling for Individual and Organizational Accounts.

src/classes/OPP_OpportunityContactRoles_TEST.cls
joy-sfdo: It's not necessary -- it was on my radar. I have removed it from the query.

src/classes/OPP_OpportunityContactRoles_TEST.cls
joy-sfdo: Same as above.

src/classes/RD_RecurringDonations_TEST.cls
joy-sfdo: True, accountId is not used. I placed it there to show the setIndividualAccountForTests method returns a value. However, I removed it.

src/classes/RD_RecurringDonations_TEST.cls
joy-sfdo: The issue that was present earlier development is no longer surfacing. I have removed the if check.

src/pages/STG_PanelContactRoles.page
joy-sfdo: Good catch. In order to align the spacing, I have added <br /> tags to the page.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -0,0 +1,51 @@
...
+/**
+* @author Salesforce.org
+* @date 2018
+* @group Opportunity
+* @description Handles Automated Soft Credits for Individual and Organizational gifts.
+*/
+
+public class OPP_AutomatedSoftCreditsService {
+
+	/*********************************************************************************************************
```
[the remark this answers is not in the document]
joy-sfdo: Changed tabs to spaces.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -0,0 +1,51 @@
...
+    * @description Determines if the Account is an Organization based on its account classification, 
+    *				e.g., Individual.
+    * @param accountClassification The classification of an Opportunity's associated Account.
+    * @return Boolean
+    **********************************************************************************************************/
+    public static Boolean isOrganizationalAccount(Boolean accountClassification) {
```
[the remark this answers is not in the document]
joy-sfdo: I see where you're going. However, since the code is being called internally, as opposed to from an outside method, l I think the system field is enough rather than pass the entire Account object when only that attribute is necessary. Also, I think this method is better than looking directly at the system field because it improves readability. When someone is looking at the code, they can read isOrganizationalAccount and understand what's happening quicker than if they just saw the system field.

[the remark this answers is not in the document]
davidhabib: I agree your creating the isOrgAccount() is a good thing!  i hate that we have a bunch of code looking at these system flags (that I always forget which ones to look at).  I just want the abstraction to be real, rather than forcing the callers to still pass the correct system flag.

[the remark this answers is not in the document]
joy-sfdo: Yes, I get that. Let's put some cycles on that one.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -325,33 +340,35 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
             for (integer i=0; i<listOpps.size(); i++) {
                 Opportunity newOpp = listOpps[i];
                 Opportunity oldOpp = listOldOpps[i];
+
+                Boolean isIndividualAccount = (oppAccountDetails != null  && oppAccountDetails.containsKey(newOpp.AccountId)) ? oppAccountDetails.get(newOpp.AccountId).npe01__SYSTEMIsIndividual__c : null;
```
davidhabib: seems like you addressed the code at line 234, but didn't change line 344.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -0,0 +1,51 @@
...
+
+public class OPP_AutomatedSoftCreditsService {
+
+    /*********************************************************************************************************
+    * @description Determines if the Account is an Organization based on its account classification, 
+    *               e.g., Individual.
+    * @param accountClassification The classification of an Opportunity's associated Account.
+    * @return Boolean
+    **********************************************************************************************************/
+    public static Boolean isOrganizationalAccount(Boolean accountClassification) {
```
davidhabib: I know we disagree on this, and I don't like forcing people to a design that isn't there's, but this still makes no sense to me.  we've created an entire service class and testing class, for a single method that take an isIndividualAccount boolean, and returns whether it is false.  I still fail to see the benefit of this service class or method.  It isn't adding any abstraction, since all callers have to know to pass the correct boolean field, and that field has even been documented in your apex methods that use it that it specifically says whether the class is an Individual account or not.  Organizational account is just the opposite of Individual account.  there is no third choice in NPSP.   Would you be open to having Jenny and Radmila taking a look and sharing their thoughts, and we go with the majority preference?

src/classes/OPP_OpportunityContactRoles_TEST.cls
davidhabib: not a big deal, but I still see it in the pull request.  did you forget to push your change?

src/classes/OPP_OpportunityContactRoles_TDTM.cls
joy-sfdo: Good catch. I'll fix that one.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: @davidhabib -- I'm totally open to having a group discussion about the design. Team discussion is great! Also, please note the code for the Relationships and Affiliations handling will be added to the new service class.

src/classes/OPP_OpportunityContactRoles_TEST.cls
joy-sfdo: Good catch. I have removed the fields and pushed the code. Keep me posted. Thanks.

src/classes/OPP_AutomatedSoftCreditsService.cls
davidhabib: ah, glad to hear this class is intended for more than just this one method.  given that, let's go ahead and move forward with what you have.  thanks.

davidhabib:
@joy-sfdo looks like there is a compilation problem in the test orgs?  weird, since it deployed ok to my org.

joy-sfdo:
@davidhabib -- Thanks for the update. I'll look into the compilation problems.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: Sounds good.

PR #3001, opened 2018-03-20
Rollups: Add support for "Always Rollup to Primary Contact"

force2b:
# Critical Changes

# Changes

* Support scenarios where there is no filter group on a rollup definition
* Update the CreateDefaultRollupConfiguration logic to respect the legacy "Enable Soft Credit Rollups" field when creating default rollup metadata

# Issues Closed

# New Metadata

# Deleted Metadata

* Account.Number_of_Opportunities__c


src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -306,6 +316,13 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
                 this.parentIdsNotCommittedWithBatch.put(lastParentIdProcessed, this.batchIteration);
             }
 
+            try {
```
njjc: Do we want this in the final product?

force2b: No. Though I took that out. Must have accidentally committed instead of removing.

njjc: Cool, I'll leave this PR open until you remove it.

force2b:
@njjc FYI - Removed system.debug statements.

PR #3008, opened 2018-03-21
Update orgs_qa.txt

jessingrass:
adding an org to receive updates so we can add and test workflow rules.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


njjc:
@cdcarter ping

PR #3011, opened 2018-03-22
Recurring Donation rollups handled by CRLPs

bethbrains:
# Critical Changes

# Changes

Upon initial CRLP enablement, 4 Recurring Donation rollups are created as CRLP custom metadata records.
When CRLPs are enabled, calculation of these 4 Recurring Donation rollups is handled by the CRLP engine, not the hard-coded logic in the RecurringDonations class.

# Issues Closed

# New Metadata

- CRLP_RDSkew_BATCH.cls
- CRLP_RD_BATCH.cls
- CRLP_RollupRD_SVC.cls
- CRLP_RollupRecurringDonation_TEST.cls

# Deleted Metadata


src/classes/CRLP_RollupRD_SVC.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>42.0</apiVersion>
```
force2b: API Version

src/classes/CRLP_RollupRD_SVC.cls
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @description Primary action method. Called for each record of summaryObject with a list of the detail objects
+     * being rolled-up to that summary object. The instance of this class was instantiated with the Summary object.
+     * The constructor created a list of Rollup instances for each rollup that will be calculated by this method.
+     * This method loops through each Rollup, testing the detail records against the appropriate FilterGroup. If the
+     * filter passes, then the detail record(s) are passed to the Rollup.doOperation() method to handle tha actual
+     * rollup operations. The state of the rollup (across all detail records) is kept in the instace of the Rollup.
+     * @param details List of detail records being rolled up to the Summary object defined in the constructor.
+     * @param childRecordsByDetailId Payments for each Opportunity. Split out into a Map by Detail (Opp) Id to
```
force2b: We're not supporting payment rollups for RD's so this can just say that childRecordsByDetailId is not handled in this implementation

src/classes/CRLP_RollupRecurringDonation_TEST.cls
```diff
@@ -0,0 +1,258 @@
...
+        ));
+
+        UTIL_CustomSettingsFacade.getContactsSettingsForTests(new npe01__Contacts_And_Orgs_Settings__c(
+                npe01__Opportunity_Contact_Role_Default_role__c = 'Donor',
+                npe01__Account_Processor__c = CAO_Constants.HH_ACCOUNT_PROCESSOR
+        ));
+
+        UTIL_CustomSettingsFacade.getRecurringDonationsSettingsForTest(
+                new npe03__Recurring_Donations_Settings__c(
+                        npe03__Maximum_Donations__c = 50,
```
force2b: Do you need to set the RD rollup batch size?

src/classes/RD_RecurringDonations.cls
```diff
@@ -151,103 +151,115 @@ public class RD_RecurringDonations {
...
-                    updateMap.put(rdid, rd);
-                }
-                //if this is the closed set of opps    
-                else if (isClosed == true && isWon == true) {
-                    rd.npe03__Last_Payment_Date__c = (date)obj.get('MaxDate');
 
-                    //convert rollup total to RD's currency
-                    if (RLLP_OppRollup_UTIL.isMultiCurrency()) {
-                        rd.npe03__Paid_Amount__c = UTIL_CurrencyConversion.convertFromCorporateUsingStandardRates((String) rd.get(RDCurrencyField), (Decimal) obj.get('Total'));
+        Boolean CRLPEnabled = UTIL_CustomSettingsFacade.getCustomizableRollupSettings().Customizable_Rollups_Enabled__c;
```
force2b: It recommend using the existing implementation for this:
`CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled`.

src/classes/CRLP_Query_SEL.cls
```diff
@@ -64,8 +64,13 @@
             allFieldsToQuery.add(SObjectType.Opportunity.fields.Amount.getSobjectField().getDescribe());
             allFieldsToQuery.add(SObjectType.Opportunity.fields.CloseDate.getSobjectField().getDescribe());
             allFieldsToQuery.add(SObjectType.Opportunity.fields.npe03__Recurring_Donation__c.getSobjectField().getDescribe());
-        } else  if (resultObject == npe01__OppPayment__c.SObjectType) {
+        } else if (resultObject == npe01__OppPayment__c.SObjectType) {
             allFieldsToQuery.add(SObjectType.npe01__OppPayment__c.fields.npe01__Opportunity__c.getSobjectField().getDescribe());
+        } else if (resultObject == npe03__Recurring_Donation__c.SObjectType) {
+            allFieldsToQuery.add(SObjectType.npe03__Recurring_Donation__c.fields.npe03__Next_Payment_Date__c.getSobjectField().getDescribe());
```
force2b: There shouldn't be a need to hardcode the target recurring donation fields here. This should just pick up fields that are not referenced by the CMT records for the given rollup type.

src/classes/CRLP_RollupRecurringDonation_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>42.0</apiVersion>
```
force2b: Api version here needs to be 37

src/classes/RD_RecurringDonations.cls
```diff
@@ -151,103 +151,115 @@ public class RD_RecurringDonations {
...
-                else if (isClosed == true && isWon == true) {
-                    rd.npe03__Last_Payment_Date__c = (date)obj.get('MaxDate');
 
-                    //convert rollup total to RD's currency
-                    if (RLLP_OppRollup_UTIL.isMultiCurrency()) {
-                        rd.npe03__Paid_Amount__c = UTIL_CurrencyConversion.convertFromCorporateUsingStandardRates((String) rd.get(RDCurrencyField), (Decimal) obj.get('Total'));
+        Boolean CRLPEnabled = CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled;
+
+        // only use this hard-coded RD rollup code if Customizable Rollups aren't enabled
+        if (!CRLPEnabled) {
```
force2b: @bethbrains After running through a test on this in a scratch org and comparing the new functionality to the old functionality, the only difference I found is that the "Next Payment Date" is not automatically populated after creating a new RD when Customizable Rollups are enabled. I suspect this is because the new rollups are not running after the installment Opps are created. I think this is because there is no associated `else` with this condition. Basically, it's calculating the values if CRLP is disabled, but nothing happens when CRLP is enabled.

bethbrains: The trigger does pick it up as designed, it's just LEX caching. I've added this to the tests to verify, and tested in the UI by turning off the persistent caching setting and refreshing after RD creation. Arrrrrgh LEX caching...

PR #3013, opened 2018-03-23
Remove Duplicate Primary OCRs Batch

### LINKED ISSUE #2885: In Lightning Experience, standard New button on Opportunity related list produces duplicate Primary Opportunity Contact Role records
While our best practice recommendation is to use an Action on Contact to create an Opportunity from a Contact record, Winter '18 introduced a change in behavior that allows a user to create an Opportunity directly from the standard New button on the Opportunity related list on Contact. Previously, this button in Lightning Experience created a new Contact Role record only for an already-existing Opportunity.

With this change to the 'New' button, Salesforce automatically creates an Opportunity Contact Role record in Lightning Experience. NPSP is also creating an OCR so one ends up with 2 records incorrectly:

image

In Classic, the standard New button does not create a duplicate Contact Role with the exact same fields populated with the same values (Opp Name, Amount, Close Date, Stage):

image

As it is likely the creation of the OCR is the desired behavior by Salesforce for general purposes, our code should not create a duplicate and instead correctly populate the role on the OCR if it exists.

kofijohnson:
# Critical Changes

# Changes
- We created a new batch that checks for and removes duplicate Primary Opportunity Contact Roles. You can find this new batch, Remove Duplicate Primary OCRs, in NPSP Settings under Bulk Data Processes. You can find out if there are duplicate Primary Opportunity Contact Roles in your org by running Health Check.

# Issues Closed
- Fixes #2885

# New Metadata
ApexClass
- OPP_PrimaryContactRoleMerge.cls
- OPP_PrimaryContactRoleMerge_BATCH.cls
- OPP_PrimaryContactRoleMerge_TEST.cls
- STG_PanelPrimaryContactRoleMerge_CTRL.cls

ApexPage
- STG_PanelPrimaryContactRoleMerge.page

CustomLabels
- OCRRemovePrimaryDuplicateBatchErrorText
- stgHelpPrimaryContactRoleMergeBatch
- stgLabelPrimaryContactRoleMergeBatch

# Deleted Metadata


smarovich:
@kofijohnson @judisohn I added some release notes. Let me know if they look ok or if there is anything I should add. Thanks!

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,258 @@
...
+        * @param OpportunityContactRole the OCR to set as primary.
+        * @return void.
+        ********************************************************************************************************/
+        public void makePrimary(OpportunityContactRole inputOCR) {
+            OpportunityContactRole ocrToUpdate = ocrRecords.get(inputOCR.Id);
+            ocrToUpdate.IsPrimary = true;
+            ocrRecordsToUpdate.put(ocrToUpdate.Id, ocrToUpdate);
+
+            Opportunity opportunityToUpdate = opportunityRecords.get(inputOCR.OpportunityId);
+            opportunityToUpdate.Primary_Contact__c = inputOCR.ContactId;
```
jjbennett: Do we need to always update the contact or if it's already that contact can we skip?

src/classes/OPP_PrimaryContactRoleMerge_BATCH.cls
```diff
@@ -0,0 +1,127 @@
...
+    ********************************************************************************************************/
+    public void execute(Database.BatchableContext BC, List<Opportunity> scope) {
+        List<AggregateResult> primaryOCRsAggregateResult = getPrimaryOCRsAggregateResult(scope);
+        if (primaryOCRsAggregateResult.isEmpty()) {
+            return;
+        }
+
+        OPP_PrimaryContactRoleMerge.DuplicatePrimaryOCRs duplicatePrimaryOCRs  = new OPP_PrimaryContactRoleMerge.DuplicatePrimaryOCRs(primaryOCRsAggregateResult);
+        String settingsRole = UTIL_CustomSettingsFacade.getContactsSettings().npe01__Opportunity_Contact_Role_Default_role__c;
+        validateOpportunityOCRs(duplicatePrimaryOCRs, settingsRole);
```
jjbennett: Just commenting about our discussion here, is it possible to make this both readable and efficient so that we are not looping through the same records multiple times?

src/classes/STG_PanelPrimaryContactRoleMerge_CTRL.cls
```diff
@@ -0,0 +1,64 @@
...
+    */
+    public boolean isRunningBatch { get; set; }
+
+    /*********************************************************************************************************
+    * @description Action Method to run the Remove Duplicate Primary OCRs Batch.
+    * @return null 
+    */
+    public PageReference runBatch() {
+        isRunningBatch = true;
+        Database.executeBatch(new OPP_PrimaryContactRoleMerge_BATCH(), 2000);
```
force2b: @kofijohnson Does it make sense to put the batch size into a Custom Settings field with a default of 2000? There will be a new UI page as part of the Customizable Rollups project to allow admins to define batch sizes for a bunch of different batch jobs. We can add this new field to that list so that admins will eventually be able to override the default size of 2000. @jjbennett

src/classes/OPP_PrimaryContactRoleMerge.cls
kofijohnson: Good point! We don't need to update if that contact is already there. That way, we can save DML.

src/classes/STG_PanelPrimaryContactRoleMerge_CTRL.cls
kofijohnson: Good point @force2b. However, in this case, the admin will not need to define the batch size. We want the size to be high as possible (2000) so the batch can go through the Opportunities as quick as possible.

kofijohnson:
Thank you @smarovich for the release note. It looks good!

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,256 @@
...
+        * @return void.
+        ********************************************************************************************************/
+        public void makePrimary(OpportunityContactRole inputOCR) {
+            OpportunityContactRole ocrToUpdate = ocrRecords.get(inputOCR.Id);
+            ocrToUpdate.IsPrimary = true;
+            ocrRecordsToUpdate.put(ocrToUpdate.Id, ocrToUpdate);
+
+            Opportunity opportunityToUpdate = opportunityRecords.get(inputOCR.OpportunityId);
+            if (opportunityToUpdate.Primary_Contact__c != inputOCR.ContactId) {
+                opportunityToUpdate.Primary_Contact__c = inputOCR.ContactId;
```
jjbennett: Nic is verifying the logic on this whether we should be updating the opp or if the contact specified on the opp should win.

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,256 @@
...
+            }
+            return result;
+        }
+    }
+
+
+    /*******************************************************************************************************
+    * @description Error Log. This class contains the logic that creates the error logs.
+    ********************************************************************************************************/
+    public class Errors {
```
jjbennett: Could we just collect a list of errors and process all of the dml at the same time?

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        /*******************************************************************************************************
+        * @description Commits the DMLs (delete/Update OCRs, update Opportunities)
+        * @return void.
+        ********************************************************************************************************/
+        public void processDML() {
+            delete ocrRecordsToDelete.values();
+            update ocrRecordsToUpdate.values();
+            update opportunityRecordsToUpdate.values();
```
jjbennett: If we decide to move forward with updating the Opps, do we need exception handling here?

src/classes/OPP_PrimaryContactRoleMerge_BATCH.cls
jjbennett: What about placing the validation directly into the merge method or moving the loop higher up so you aren't looping through the same records multiple times doing the same check?

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,256 @@
...
+
+        /*******************************************************************************************************
+        * @description Gets the OCRs with Role equals to the Role in the parameter.
+        * @param String OCR Role.
+        * @return List<OpportunityContactRole> The OCRs with Role equals to the Role in the parameter.
+        ********************************************************************************************************/
+        public List<OpportunityContactRole> getRoleEquals(String ocrRole) {
+            List<OpportunityContactRole> result = new List<OpportunityContactRole>();
+            for (OpportunityContactRole ocr : record.OpportunityContactRoles) {
+                if (ocr.Role == ocrRole) {
```
jjbennett: Case sensitivity could be an issue here.

src/classes/OPP_PrimaryContactRoleMerge.cls
```diff
@@ -0,0 +1,256 @@
...
+* @date 2018
+* @group Opportunity Contact Role
+* @description Domain Classes for the deletion of duplicate primary OCRs. 
+*/
+
+public with sharing class OPP_PrimaryContactRoleMerge {
+
+    @TestVisible private static final String ERROR_MESSAGE = Label.OCRRemovePrimaryDuplicateBatchErrorText;
+    @TestVisible private static final String ERROR_CONTEXT = 'Remove Duplicate Primary OCRs';
+    @TestVisible private static final String ERROR_SOBJECT = 'Opportunity';
```
jjbennett: Instead of hard coding Opportunity, could we use the String.valueOf Opportunity?

src/classes/OPP_PrimaryContactRoleMerge_BATCH.cls
```diff
@@ -0,0 +1,126 @@
...
+    public Database.QueryLocator start(Database.BatchableContext info) {
+        return Database.getQueryLocator('SELECT Id FROM Opportunity');
+    }
+
+    /*******************************************************************************************************
+    * @description Batch execute method. It gets the Opportunities with duplicate primary OCRs and runs 
+    * the logic that removes the Primary OCR duplicates.
+    * @return void.
+    ********************************************************************************************************/
+    public void execute(Database.BatchableContext BC, List<Opportunity> scope) {
```
jjbennett: What do you think about moving the logic outside of the batch execute method maybe into your other class?

src/classes/OPP_PrimaryContactRoleMerge.cls
njjc: Yes, if we update opps we need exception handling with partial success. I really don't think we should be updating opps though.

src/classes/OPP_PrimaryContactRoleMerge.cls
njjc: ocrRole.equalsIgnoreCase(ocr.Role)

src/classes/OPP_PrimaryContactRoleMerge.cls
njjc: UTIL_Describe.getObjectLabel('Opportunity');

src/classes/OPP_PrimaryContactRoleMerge_BATCH.cls
njjc: It would be nice to have a single method to call that simply runs the dedup logic on a set of opps instead of doing some of the logic in the execute method.

kofijohnson: Good point @njjc, I moved the logic in the Batch execute to the OPP_PrimaryContactRoleMerge class.

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
```diff
@@ -0,0 +1,426 @@
...
+    * @description Given an Opportunity with blank Primary Contact AND the Opportunity has duplicate primary 
+    * OCRs with Role != NPSP Settings Role, when the batch runs, then an error log created, containing a link 
+    * to the Opportunity.
+    */
+    @isTest static void test_LogErrorIfNoOCRsRoleDefinedInNPSPSettings() {
+        String nonNPSPRole = 'NON_NPSP_ROLE';
+
+        deactivateTriggers();
+
+        Account account = new Account(Name = 'TEST_ACCOUNT');
```
jjbennett: What do you think about moving some of the setup data to the test setup class? You could leverage class level variables to hold the account and contact records created from test setup and also some of the constant strings that repeat throughout could be added as class constants.

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
```diff
@@ -0,0 +1,426 @@
...
+        };
+        insert primaryOCRs;
+
+        activateTriggers();
+        
+        Test.startTest();
+        Database.executeBatch(new OPP_PrimaryContactRoleMerge_BATCH());
+        Test.stopTest();
+
+        List<Opportunity> expectedOpportunities = [
```
jjbennett: Could this be extracted into a separate method?

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
```diff
@@ -0,0 +1,426 @@
...
+
+@isTest
+private class OPP_PrimaryContactRoleMerge_TEST {
+
+    /*********************************************************************************************************
+    * @description Given an Opportunity with blank Primary Contact AND the Opportunity has duplicate primary 
+    * OCRs with Role != NPSP Settings Role, when the batch runs, then an error log created, containing a link 
+    * to the Opportunity.
+    */
+    @isTest static void test_LogErrorIfNoOCRsRoleDefinedInNPSPSettings() {
```
jjbennett: One other question on the name of this test, it sounds like npsp default contact role is not set in settings but the test appears to actually test when the role doesn't match what is set as the default in settings?

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
kofijohnson: I think since it is only 1 line "Database.executeBatch(new OPP_PrimaryContactRoleMerge_BATCH());" we don't need to create another method.

jjbennett: I was referring to the query on the expectedOpportunities, I noticed that was repeated throughout your test methods.

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
kofijohnson: The NPSP Settings Role here is NULL, which is different than the role of the OCRs (NON_NPSP_ROLE). But I get your point, I can create NPSP Settings Role to be more explicit.

src/classes/OPP_PrimaryContactRoleMerge_TEST.cls
jjbennett: Instead of querying for the results, one option would be to create the reused variables as static class level variables and then assign the results of the TestSetup there.

PR #3026, opened 2018-04-06
Rollups: Implement Package Features to track CRLP Usage

force2b:
# Critical Changes

# Changes

* From the NPSP Settings Page, submit a simple Boolean back to the LMO when Customizable Rollups are enabled or disabled.
* For this to work, the Feature Management App (FMA) needs to be installed in the Salesforce.org License Management Organization (LMO). For now it's effectively a placeholder until the feature is defined.
* Updated FeatureParameter values are automatically sync'd with the LMO asynchronously each evening by core Salesforce functionality.

# Issues Closed

# New Metadata

* Boolean Feature Parameters: `isEnabled_CustomizableRollups`
* Apex Class: UTIL_OrgTelemetry_SVC.cls

# Deleted Metadata


src/classes/STG_SettingsService.cls
```diff
@@ -233,4 +233,46 @@ public with sharing class STG_SettingsService {
...
+    }
+
+    /*********************************************************************************************************
+    * @description FeatureManagement operations are subject to MIXED_DML errors and must be conducted asynchronously
+    */
+    @future
+    private static void submitFeatureTelemetryToLMO_ASync() {
+        STG_SettingsService stgSvc = new STG_SettingsService();
+
+        System.FeatureManagement.setPackageBooleanValue('IsEnabled_CustomizableRollups', (stgSvc.stgCRLP.Customizable_Rollups_Enabled__c == true));
```
cdcarter: Hey @force2b, given our conversation in standup today and @njjc's very valid point about the XI message being able to fail... do we want to consider a date featureParam of something like `DateCustomizableRollupsEnabled` which we could then re-send for the next few evenings in a scheduled job?

force2b: It's a great question. I'm hesitant to write code to work around potential short term/sporadic/possibly rare outages with the cross-org sync process used by feature management; at least right now. 

Given the nightmare stories I've heard about our LMO, I propose the following rather conservative approach:

*Phase 1:*
* Simple submit of the featureParameter when the user enables or disables CRLP from Custom Settings. 
* Build out a report and dashboard that can be used to show CRLP enablement statistics (for example: # enabled in Sandbox, # enabled in Production, % enabled overall).
* Get buy-in from EntSys to expand what is captured and the frequency

*Phase 2:*
* Add 3 or 4 additional feature parameters to what is sync'd back to the LMO
* Enable the submit process in the install script (which BC has noted should work going forward)
* This can be implemented in a few sprints, once we have additional buy-in from EntSys to expand telemetry capture
* Enhance the reports/dashboards from phase 1 to include these new statistics

*Phase 3:*
* Evaluate the feature parameter results. We should be able to get some data on the number of orgs that have submitted feature parameter data the day after a sandbox or production push to determine the effectiveness. The worse case scenario is that we move from a post-install script approach to building something into one of our nightly batch jobs. The downside of the batch jobs is that not every org has our batch jobs scheduled.
* Add the remainder of the feature parameters we've identified
* Enhance the reports/dashboards from phase 1 to include these new statistics

cdcarter: To clarify, we don't need buy in from Entsys around the volume, we are holding onto one parameter by request of packaging. We don't expect this to hurt our LMA at all, packaging wants to monitor all those XI calls at first tho.

We should not take on any work for building any reports/dashboards on this information in the LMA as a part of this ticket.

Did BC give you a work item for the Install Script bug? I believe he said he was hoping to work on it, but I didn't see any actual work. I do not think we should expect this to work. 

We can absolutely wait and see what happens with the data/evaluate the results, but capturing the enablement date instead of a boolean means its harder for us to *lose* data even if its delayed.

cdcarter: So...in summary...I think it should still be a date, even if we only send it on enablement/disablement events.

cdcarter: we're also getting very close to needing to split up our batch job story so that application services always run, even if RDs don't, for example.

force2b: Post Install Script Bug: https://gus.my.salesforce.com/a07B0000004r7Df

force2b: Thank you for the clarifications. 

A date might work, but we'd need a way to know that it's been disabled for accurate reporting. Such as two Date featureParameters - one for "Date CRLP Enabled" and one for "Date CLRP Disabled". The plus I see for going with two date parameters is that we get a little more detail on data coming in, but if an org both enables and disables on the same day we really have no idea what their current status is because both fields have the same value. 

That noted, I could see going with a Boolean and a Date. The Boolean gives us the actual current state while the Date tells us when they last enabled it or disabled it. There's really no way to know whether the sync did not happened though. If the default value for the date is "1/1/1970" and the org never enables CRLP the date value never changes. If the org enables CRLP and the sync doesn't happen - the date never changes. We can't know why the date never changed.

The only real way to know a sync isn't happening is if we do it during the post install script. Then we could correlate the time stamps on the featureParameter records with the last push to the org.

cdcarter: Did we confirm if the timestamp on the featureParameter is the time we set it in the subscriber org, or the time the nightly sync happens. Can't recall..

I was very naively thinking we could just 1/1/1970 the date on disablement. We'd treat the date like a boolean, it just has replayability (we can keep sending it nightly, if we choose).

cdcarter: I think that there really is no way to know that the sync didn't happen. We may have to live with that.

cdcarter: It's loathesome that this isn't built on platform events, but here we are.

cdcarter: Ultimately I am pretty willing for this to just be a boolean, but I thought it worth the virtue of bringing it up, and i think it's causing good conversation.

Bottom line, I agree we shouldn't design around a possible XI issue, so we can probably merge this, but we do need to think about (and perhaps chat more specifically with packaging) about how frequently we can/should do sends.

force2b: Yes. I was using the LastModifiedDate as the way to know that the records updated and can confirm that they were updated when the org sync'd.

I think for now we have to live with the fact that core's process may not be perfect. Once we're able to move to doing this in the install script though, we should be able to rely a lot more on the data. The data isn't necessarily mission critical, so there's probably an acceptable tolerance on a small percentage of orgs not sync'ing. Once we move to the install script method I think it will be possible to generate some solid stats on what percent of orgs sync or do not sync their featureParameter data.

force2b: Agreed on talking more with packaging. It seemed to me that BC felt we're on the cutting edge of using this functionality so they may not have any guidance on best practices for sync frequency at this point.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -0,0 +1,53 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2018
+* @group Utilities
+* @description Return some basic org and NPSP telemetry data to the License Management Application to assist with
+* better understanding of how our customers use our application.
+*/
+public without sharing class UTIL_OrgTelemetry_SVC {
```
njjc: @force2b any justification for `without sharing`?

force2b: This particular class isn't returning much, but the next version will attempt to return stats data, such as record counts, so I wanted the queries to be sure to include all records not just the records that the current user can see. I suppose that there's an assumption that the installing user/job user will be an admin and have visibility to all records, but just in case.

PR #3029, opened 2018-04-11
Feature/rollups  ui filter rule picklist values

randi274:
# Critical Changes

# Changes
- Added filter rule picklist values for filter rules with values that can be chosen from a picklist or multi-picklist.
- Upgraded logic to prevent duplicate and require all filter rule fields.
- Made general improvements to the existing filter rule modal.
- Included toast and spinner components from the rollup component. 

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -125,50 +140,58 @@
...
+        var operator = event.getSource().get("v.value");
+        helper.rerenderValue(cmp, operator);
     },
 
-    /* @description: saves a new filter group and associated filter rules
-    */
+    /**
+     * @description: adds constant field to selected list
+     */
+    /*onChangeFilterRuleConstantPicklist: function(cmp, event, helper){
```
force2b: Is this commented out for future use or should it be removed?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -69,24 +71,199 @@
...
+        var action = cmp.get("c.getFilterRuleConstantPicklistOptions");
+        action.setParams({objectName: cmp.get("v.activeFilterRule.objectName"), selectedField: field});
+        console.log(cmp.get("v.activeFilterRule.objectName"), field);
+        action.setCallback(this, function (response) {
+            var state = response.getState();
+            if (state === "SUCCESS") {
+                cmp.set("v.filterRuleConstantPicklist", response.getReturnValue());
+            }
+            else if (state === "ERROR") {
+                var errors = response.getError();
```
force2b: Should errors be displayed via toast?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -69,24 +71,199 @@
...
+                cmp.set("v.filterRuleFieldType", 'text');
+            } else {
+                if (operator === 'Equals' || operator === 'Not_Equals') {
+                    cmp.set("v.filterRuleFieldType", 'picklist');
+                } else if (operator === 'In_List' || operator === 'Not_In_List') {
+                    cmp.set("v.filterRuleFieldType", 'multipicklist');
+                }
+            }
+        } else if (type === 'time') {
+            cmp.set("v.filterRuleFieldType", 'time');
```
force2b: The filter rule logic will not support the new Time type. This code is probably ok, but we need to ensure that fields of type "Time" cannot be selected as a filter field.

Sorry, I completely forgot about this one.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -109,35 +286,107 @@
...
+        var canSave = cmp.find("filterRuleField").reduce(function (validSoFar, filterRuleCmp) {
+            filterRuleCmp.showHelpMessageIfInvalid();
+            return validSoFar && filterRuleCmp.get("v.validity").valid;
+        }, true);
+
+        //check for duplicates
+        if (cmp.get("v.filterRuleMode") === 'create') {
+            //check for specific API names instead of the full row to avoid issues with the ID not matching
+            for (var i = 0; i < filterRuleList.length; i++) {
+                if (filterRuleList[i].objectName === filterRule.objectName
```
force2b: I think these two blocks can be combined into a single for loop. Within the loop, it can check for either `cmp.get("v.filterRuleMode") === 'create'` or `i !== filterRule.index` to determine if the comparisons should be made.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
randi274: Will DateTime be accepted? Or just not Time?

force2b: Yes. DateTime is technically OK (though I'm not sure I can find a use case), but Time is a new data type and I couldn't really come up with a use case that made it worth the time to build support for it.

force2b: Time and Blob type (Large/Rich Text Area) are the only ones I can think of that would not be supported for filters. The reason I want to exclude Blob fields is that they have a negative impact on query performance, especially in a batch job. I can't really think of a good reason to support filtering on a 32K field anyway. Technically IsDeleted should be excluded as well because we never query deleted records, so including it could be confusing to users.

randi274: Cool, I can include Blob no problem, and refactor a bit for IsDeleted. Thanks!

PR #3030, opened 2018-04-11
Escape user input text and/or output text

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/components/UTIL_PageMessages.component
+++ b/src/components/UTIL_PageMessages.component
@@ -15,7 +15,7 @@
         />
 
     <apex:attribute name="escMsg"
-        description="If true, the page message will be escaped, by default it will not."
+        description="If false, the page message will be not be escaped, by default the message will be escaped."
         type="Boolean"
         default="true"
         required="false"
--- a/src/components/UTIL_PageMessages.component
+++ b/src/components/UTIL_PageMessages.component
@@ -14,6 +14,13 @@
         required="false"
         />
 
+    <apex:attribute name="escMsg"
+        description="If true, the page message will be escaped, by default it will not."
+        type="Boolean"
+        default="true"
+        required="false"
+        />
+
     <apex:outputText rendered="{!NOT(useToast)}">
         <style type="text/css">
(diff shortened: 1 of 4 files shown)


src/components/UTIL_PageMessages.component
```diff
@@ -14,6 +14,13 @@
         required="false"
         />
 
+    <apex:attribute name="escMsg"
+        description="If true, the page message will be escaped, by default it will not."
+        type="Boolean"
+        default="true"
+        required="false"
```
davidhabib: your description says by default it will not escape, but your default value = true. shouldn't we make it default=true, since that is the most secure behavior.  those that explicitly want to override escape need to say so (escape=false).  this would be consistent with other visualforce components.

jjbennett: Yes, sorry I was going to default to false worried about the other consumers but after checking through the pages that use links they appear to use them directly vs in the error message. I was  unable to find anything that appeared to use links in the errors.

jjbennett: Good catch, I was looking at the hrefs we were using in labels and tested the contact delete override but not with child opps.

jjbennett:
@davidhabib - I updated the contact delete html text so that the Name field is now being escaped on the contact delete page. I have searched through the code and tested most consumers and I do not think anyone else is displaying HTML. I do remember this page specifically since it also had issues displaying the HTML in LEx, we removed the links from LEx but I still encoded the name since during testing it would sometimes display a broken image depending on the Name I stored in the Opp.

davidhabib:
@jjbennett Are all other users of the component specifying escape=true (or not specifying it at all)?  thanks.

jjbennett:
All other consumers are using the default value of true and not specifying.

jjbennett:
@davidhabib I hope you don't mind but I grabbed another xss P1 item that I am going to push a change for to this PR.

davidhabib:
thanks for picking up the additional security bug.  please update the pull request header with the bug numbers.  I'll do some testing this morning and get it merged in.  thanks!

PR #3035, opened 2018-04-12
Feature/rollups  enablement3

bethbrains:
# Critical Changes

# Changes

Revised NPSP settings panel to expose new Rollup batch size options.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_Currency.cls
+++ b/src/classes/UTIL_Currency.cls
@@ -85,15 +85,9 @@ public with sharing class UTIL_Currency implements UTIL_Currency.Interface_x {
             isACMEnabled = false;
             if (isMultiCurrencyOrganization()) {
                 // Using a string in Database.query because this query won't compile if multicurrency isn't enabled
-                // Using try/catch in case platform bug gets fixed whereby
-                // DatedConverstionRate will (accurately) not exist in MultiCurrency org without ACM enabled
-                try {
-                    Integer DCRCount = Database.query('SELECT Id FROM DatedConversionRate LIMIT 1').size();
-                    if (DCRCount == 1) {
-                        isACMEnabled = true;
-                    }
-                } catch (Exception e) {
-                    // Nothing to do, ACM = false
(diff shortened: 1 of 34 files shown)


src/classes/UTIL_Currency.cls
```diff
@@ -77,15 +77,18 @@ public with sharing class UTIL_Currency implements UTIL_Currency.Interface_x {
...
     private Boolean isACMEnabled = null;
     public Boolean isAdvancedCurrencyManagementEnabled() {
         if (isACMEnabled == null) {
+            isACMEnabled = false;
             if (isMultiCurrencyOrganization()) {
-                isACMEnabled = Schema.getGlobalDescribe().containsKey('DatedConversionRate');
-            } else {
-                isACMEnabled = false;
+                // Using a string in Database.query because this query won't compile if multicurrency isn't enabled
+                Integer DCRCount = Database.query('SELECT Id FROM DatedConversionRate LIMIT 1').size();
```
force2b: Since we're technically working around a core bug, I recommend wrapping this query in a try/catch in case they eventually fix the problem of this SObject existing when ACM hasn't actually been enabled.

src/objects/Customizable_Rollup_Settings__c.object
```diff
@@ -135,7 +135,7 @@
         <description>Only used by the new Customizable Rollups engine</description>
         <externalId>false</externalId>
         <inlineHelpText>The maximum number of related Opportunities on an Account or Contact allowed in the non-Skew Mode Rollup Batch Jobs. An Account or Contact with more than this number of related Opportunities will always be rolled up using Skew Mode.</inlineHelpText>
-        <label>Rollups: Max Related Record in Skew Mode</label>
+        <label>Skew Mode Threshold</label>
```
force2b: spelling error

bethbrains: "Threshold" is correct. Is there something else I'm not seeing?

force2b: Sorry. Need to not review code at the end of a long Friday.

PR #3036, opened 2018-04-16
Feature/rollups  crlp naming

bethbrains:
# Critical Changes

# Changes

- Improved apex error handling on CRLP enablement page (polling stops)
- Alignment between Rollup__mdt records and target field names, substituting # for Number to avoid truncation

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_DefaultConfigBuilder.cls
+++ b/src/classes/CRLP_DefaultConfigBuilder.cls
@@ -306,49 +306,49 @@ public class CRLP_DefaultConfigBuilder {
             'Account.npo02__FirstCloseDate__c' =>
                     new List<String>{ 'Account: First Gift Date', 'Date of the first gift from the Household' },
             'Account.npo02__LargestAmount__c' =>
-                    new List<String>{ 'Account: Largest Gift Amount', 'Largest gift amount from the Household' },
+                    new List<String>{ 'Account: Largest Gift', 'Largest gift amount from the Household' },
             'Account.npo02__LastCloseDate__c' =>
                     new List<String>{ 'Account: Last Gift Date', 'Date of the last gift from the Household' },
             'Account.npo02__LastOppAmount__c' =>
                     new List<String>{ 'Account: Last Gift Amount', 'Amount of the last gift from the Household' },
             'Account.npo02__NumberOfClosedOpps__c' =>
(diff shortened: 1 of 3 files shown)


src/classes/CRLP_DefaultConfigBuilder.cls
```diff
@@ -306,49 +306,49 @@ public class CRLP_DefaultConfigBuilder {
...
             'Account.npo02__LastOppAmount__c' =>
                     new List<String>{ 'Account: Last Gift Amount', 'Amount of the last gift from the Household' },
             'Account.npo02__NumberOfClosedOpps__c' =>
-                    new List<String>{ 'Account: Count All Gifts', 'Total number of all gifts from the Household' },
+                    new List<String>{ 'Account: Total Number of Gifts', 'Total number of all gifts from the Household' },
             'Account.npo02__OppAmount2YearsAgo__c' =>
                     new List<String>{ 'Account: Total Gifts Two Years Ago', 'Total of all gifts from 2 Calendar Years ago' },
             'Account.npo02__OppAmountLastNDays__c' =>
-                    new List<String>{ 'Account: Total Gifts Last ' + LastNDaysString + ' Days', 'Total of all gifts for the last ' + LastNDaysString + ' days' },
+                    new List<String>{ 'Account: Total Gifts Last N Days', 'Total of all gifts for the last N days. N is defined in Household Settings.' },
```
force2b: @bethbrains  Out of curiosity, why remove the logic that replaces "N" with the actual number of days that was defined in the settings? I thought that having the actual number of days in there (ex: "Account: Total Gifts Last 365 Days") was better for Users than "Account: Total Gifts Last N Days", especially given that they can create other Rollups with different numbers of days. If they're limited to only one set of "N" day rollups, than I could see forcing the "N" value, but that limitation no longer exists for Customizable Rollups.

bethbrains: From Dina in the Work Item:
The Contact Object > Details > Donation Totals section has a field called "Total Gifts Last N Days." In Customizable Rollups, this is called "Contact: Total Gifts Last 365 Days." If a user edits the # of days to reflect the Total Gifts Last 90 days, the Name of the Customizable Rollup is still "Contact: Total Gifts Last 365 Days." Therefore, this name doesn't make sense in this use case

PR #3037, opened 2018-04-16
Feature/rollups  ui requery rollups

randi274:
# Critical Changes

# Changes
- Rollup list is required if the associated filter group name is changed. Otherwise, saved filter group is only updated into the filter groups grid.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
+++ b/src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
@@ -141,11 +141,10 @@
                 var errors = response.getError();
                 if (errors) {
                     if (errors[0] && errors[0].message) {
-                        console.log("Error message: " +
-                            errors[0].message);
+                        this.showToast(cmp, "error", cmp.get("v.labels.error"), errors[0].message);
                     }
                 } else {
-                    console.log("Unknown error");
+                    this.showToast(cmp, "error", cmp.get("v.labels.error"), "Unknown Error");
                 }
             }
         });
@@ -336,12 +335,13 @@
         cmp.find("nameInput").showHelpMessageIfInvalid();
         cmp.find("descriptionInput").showHelpMessageIfInvalid();
 
-        if (!description) {
+        if (!description || !name) {
             canSave = false;
-        } else if (!name) {
+        }
+
+        if (filterRuleList.length === 0) {
             canSave = false;
-        } else if (filterRuleList.length === 0) {
-            canSave = false;
+            cmp.find("noFilterRulesMessage").set("v.severity", "error");
         }
 
         return canSave;
@@ -358,6 +358,18 @@
             filterRuleCmp.showHelpMessageIfInvalid();
             return validSoFar && filterRuleCmp.get("v.validity").valid;
         }, true);
+
+        //check for specific API name duplicates instead of the full row to avoid issues with the ID not matching
+        for (var i = 0; i < filterRuleList.length; i++) {
+            if (i !== filterRule.index
+                && filterRuleList[i].objectName === filterRule.objectName
+                && filterRuleList[i].fieldName === filterRule.fieldName
+                && filterRuleList[i].operatorName === filterRule.operatorName
+                && filterRuleList[i].constantName === filterRule.constantName) {
+                cmp.set("v.filterRuleError", cmp.get("v.labels.filterRuleDuplicate"));
+                return canSave = false;
+            }
+        }
 
         //check for duplicates
         if (cmp.get("v.filterRuleMode") === 'create') {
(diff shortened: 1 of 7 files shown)


src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -287,42 +287,44 @@ public with sharing class CRLP_RollupUI_SVC {
     */
 
     @AuraEnabled
-    public static FilterGroupModel setupFilterGroupDetail(Id id, List<String> objectNames) {
-
+    public static FilterGroupModel setupFilterGroupDetail(String filterGroupId, List<String> objectNames) {
```
force2b: @randi274 It's going to be a major challenge to merge this because nearly all of the code in this method has been re-written to not use `FilterRulesTableItem`. In fact, I've removed `FilterRulesTableItem` entirely from the class because for the save logic to work the objects passed back to the component are the `FilterGroup` and `FilterRule` inner classes from `CRLP_RollupCMT`. It's the same logic used for saving Rollups. I hate to scrap this PR, but I'm not sure how we're going to reconcile the changes. Let's talk about this tomorrow.

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
```diff
@@ -40,9 +40,42 @@
         }
     },
 
+    /**
+     *  @description: merges and saves the updated row item into the existing list of rows
+     *  @param list - list of items to merge into
+     *  @param item - item to be merged into the list
+     *  @param context - context running the merge
+     */
+    mergeRowItem: function(cmp, list, item, context){
```
force2b: Nice!

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -536,6 +538,18 @@
...
 
+        //check for specific API name duplicates instead of the full row to avoid issues with the ID not matching
+        for (var i = 0; i < filterRuleList.length; i++) {
+            if (i !== filterRule.index
+                && filterRuleList[i].objectName === filterRule.objectName
+                && filterRuleList[i].fieldName === filterRule.fieldName
+                && filterRuleList[i].operatorName === filterRule.operatorName
+                && filterRuleList[i].constantName === filterRule.constantName) {
+                cmp.set("v.filterRuleError", cmp.get("v.labels.filterRuleDuplicate"));
+                return canSave = false;
```
force2b: Is there a need to use `return canSave = false;` versus just `return false;` in this function? Seems like `canSave` is a local var only.

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
```diff
@@ -40,9 +40,42 @@
...
+        var newItem = true;
+        for (var i = 0; i < list.length; i++) {
+            //todo: refactor to lowercase Id after CMT save information is added
```
force2b: Why refactor the Id to lowercase?

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
```diff
@@ -1,7 +1,7 @@
 ({
     /* @description: resets the view assignments, clears detail information, and displays rollup grid
```
force2b: Is it worth fixing all the apexDoc comments in the JS to move the `@description` line down below `/**`?

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
```diff
@@ -63,13 +96,45 @@
...
+                var rollupList = JSON.parse(JSON.stringify(response.getReturnValue()));
+
+                cmp.set("v.rollupList", rollupList);
+                cmp.set("v.cachedRollupList", rollupList);
+            }
+            else if (state === "ERROR") {
+                var errors = response.getError();
+                if (errors) {
+                    if (errors[0] && errors[0].message) {
+                        console.log("Error message: " +
```
force2b: As we're getting closer to a production release, does it make sense to start adding toasts for unexpected errors in apex callbacks so the user has some understanding of why the page may not be working as expected?

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
randi274: Ah, this comment needs to be removed actually. There was a weird issue with the key of Id being uppercase in only some cases, but recordId looks to have resolved the problem I was seeing.

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
randi274: I think so! I've been moving them down gradually, but this is something we can do when we remove the console.logs (if not before).

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
randi274: Yes, I agree with that. I added one where you had asked for it, but I think adding them in for these is a good idea (and will help us/support with debugging too).

PR #3038, opened 2018-04-17
Rollups: Save Filter Groups & Filter Rule CMT Data

force2b:
# Critical Changes

* Change the FilterGroupModel inner class to use `CRLP_RollupCMT.FilterGroup`
* Refactor all use of the FilterGroupRuleItem to use `CRLP_RollupCMT.FilterRule`

# Changes

# Issues Closed

# New Metadata

*Custom Labels:*
* CRLP_SaveError
* CRLP_SaveProgress
* CRLP_SaveSuccess
* CRLP_SaveTimeout

# Deleted Metadata

### DIFF
--- a/src/aura/CRLP_FilterGroup/CRLP_FilterGroup.cmp
+++ b/src/aura/CRLP_FilterGroup/CRLP_FilterGroup.cmp
@@ -119,7 +119,7 @@
                                              name="{!v.labels.name}"
                                              value="{!v.activeFilterGroup.label}"
                                              required="{! !v.isReadOnly}"
-                                             readonly="{!v.isReadOnly}"/>
+                                             readonly="{! v.isReadOnly}"/>
                         </lightning:layoutItem>
                     </lightning:layout>
 
@@ -260,31 +260,31 @@
                                 <!--START CONSTANT INPUT OPTIONS-->
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'text')}">
                                     <!--todo: see if text, date, datetime-local, time and number can be combined into one-->
-                                    <lightning:textarea aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.constantName}"
+                                    <lightning:textarea aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.value}"
                                                         required="true" label="{!v.labels.constant}" maxlength="300" />
                                 </aura:if>
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'date')}">
-                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.constantName}"
+                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.value}"
                                                      required="true" label="{!v.labels.constant}" type="date"/>
                                 </aura:if>
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'datetime-local')}">
-                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.constantName}"
+                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.value}"
                                                      required="true" label="{!v.labels.constant}" type="datetime-local"/>
                                 </aura:if>
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'time')}">
-                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.constantName}"
+                                    <lightning:input aura:id="filterRuleField" class="slds-p-around_small" value="{!v.activeFilterRule.value}"
                                                      required="true" label="{!v.labels.constant}" type="time"/>
                                 </aura:if>
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'number')}">
-                                    <lightning:input class="slds-p-around_small" value="{!v.activeFilterRule.constantName}"
+                                    <lightning:input class="slds-p-around_small" value="{!v.activeFilterRule.value}"
                                                      required="true" label="{!v.labels.constant}" type="number" step="any"/>
                                 </aura:if>
                                 <aura:if isTrue="{!equals(v.filterRuleFieldType, 'picklist')}">
                                     <lightning:select aura:id="filterRuleField" class="slds-p-around_small" label="{!v.labels.constant}"
-                                                      required="true" value="{!v.activeFilterRule.constantName}">
+                                                      required="true" value="{!v.activeFilterRule.value}">
                                         <option value="" text=""/>
                                         <aura:iteration items="{!v.filterRuleConstantPicklist}" var="option">
-                                            <option value="{!option.value}" text="{!option.label}" selected="{!option.value == v.activeFilterRule.constantName}"/>
+                                            <option value="{!option.value}" text="{!option.label}" selected="{!option.value == v.activeFilterRule.value}"/>
                                         </aura:iteration>
                                     </lightning:select>
                                 </aura:if>
@@ -298,7 +298,7 @@
                         <aura:if isTrue="{!and((v.filterRuleMode != 'delete'), equals(v.filterRuleFieldType, 'multipicklist'))}">
                             <lightning:dualListbox aura:id="filterRuleField" label="" class="slds-p-around_small"
                                                    sourceLabel="{!'Available ' + v.labels.constant}" selectedLabel="{!'Selected ' + v.labels.constant}"
-                                                   options="{!v.filterRuleConstantPicklist}" value="{!v.activeFilterRule.constantName}"
+                                                   options="{!v.filterRuleConstantPicklist}" value="{!v.activeFilterRule.value}"
                                                    min="1"/>
                         </aura:if>
                     </lightning:layout>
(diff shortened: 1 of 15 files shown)


src/aura/CRLP_FilterGroup/CRLP_FilterGroup.cmp
```diff
@@ -58,6 +58,8 @@
     <aura:attribute name="filterRuleFieldType" type="String" description="Determines type of field" default="text"/>
 
     <!-- private attributes -->
+    <aura:attribute name="deletedRuleList" type="List" description="Filter Rules that were deleted" access="private"/>
+    <aura:attribute name="isPageDirty" type="Boolean" default="false" access="private" description="True when there are changes to be saved"/>
```
randi274: Does this attribute need to be used for the filter group name/description too? If not, why is it only necessary for the filterRules? Is there a process to reset it at any point, like when the rules are saved and mode changes to view, or if a rule is removed/deleted?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -192,20 +198,23 @@
...
         //add check for description, name and a filter rule
         var activeFilterGroup = cmp.get("v.activeFilterGroup");
-        var canSave = helper.validateFilterGroupFields(cmp, activeFilterGroup);
+        var filterRuleList = cmp.get("v.filterRuleList");
+        var deletedRuleList = cmp.get("v.deletedRuleList");
+        var canSave = helper.validateFilterGroupFields(cmp);
         if(canSave){
             cmp.set("v.mode", 'view');
 
             //todo: add save code here
```
randi274: This todo can be removed now that the code for it is in 👍

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -228,13 +238,13 @@
...
                 filterRule.objectLabel = helper.retrieveFieldLabel(filterRule.objectName, cmp.get("v.detailObjects"));
                 filterRule.fieldLabel = helper.retrieveFieldLabel(filterRule.fieldName, cmp.get("v.filteredFields"));
-                filterRule.operatorLabel = helper.retrieveFieldLabel(filterRule.operatorName, cmp.get("v.filteredOperators"));
+                filterRule.operationLabel = helper.retrieveFieldLabel(filterRule.operation, cmp.get("v.filteredOperators"));
 
                 //special reformatting for multipicklist and semi-colon delimited lists
-                if (filterRule.operatorName === 'In_List' || filterRule.operatorName === 'Not_In_List') {
-                    filterRule.constantLabel = helper.reformatConstantLabel(cmp, filterRule.constantName, filterRule.operatorName);
+                if (filterRule.operation === 'In_List' || filterRule.operation === 'Not_In_List') {
+                    filterRule.valueLabel = helper.reformatvalueLabel(cmp, filterRule.value, filterRule.operation);
```
randi274: Should use camel case for reformatvalueLabel function

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -153,26 +163,129 @@
...
+                            // Save the inserted/updated record id
+                            cmp.set("v.activeFilterGroup", model.filterGroup);
+                            cmp.set("v.activeFilterGroupId", model.filterGroup.recordId);
+
+                            // for a new record, copy the activeFilterGroup map to the cachedFilterGroup map
+                            if (cmp.get("v.cachedFilterGroup") && cmp.get("v.cachedFilterGroup.recordName")) {
+                                var activeFilterGroup = cmp.get("v.activeFilterGroup");
+                                var cachedFilterGroup = cmp.get("v.cachedFilterGroup");
+                                for (var key in activeFilterGroup) {
+                                    if (activeFilterGroup.hasOwnProperty(key)) {
```
randi274: I noticed this in rollups too, but is there value in checking for all keys instead of straight replacing the cachedFilterGroup with activeFilterGroup? My questions is whether there may be leftover properties from the cachedFilterGroup, and why we would want/need them using this method.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -153,26 +163,129 @@
...
+                            }
+
+                            var filterRuleList = helper.restructureResponse(model.filterRuleList);
+                            var filterRuleListCached = helper.restructureResponse(model.filterRuleList);
+
+                            cmp.set("v.filterRuleList", filterRuleList);
+                            cmp.set("v.cachedFilterRuleList", filterRuleListCached);
+                            var rollupItems = cmp.get("v.rollupItems");
+
+                            // need to sent back an object with the following properties
```
randi274: typo in sent -> send

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -153,26 +163,129 @@
...
+                            var filterGroupTableItem = {};
+                            filterGroupTableItem.label = model.filterGroup.label;
+                            filterGroupTableItem.description = model.filterGroup.description;
+                            filterGroupTableItem.name = model.filterGroup.recordName;
+                            filterGroupTableItem.recordId = model.filterGroup.recordId;
+                            if (filterRuleList) {
```
randi274: Couldn't you send back the deployResult.filterGroupItem like you did in rollups, just with a few extra properties attached to it? Or is it preferable to make a new object here?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -231,7 +344,7 @@
      * @description: opens a modal popup so user can add or edit a filter rule
      */
     resetActiveFilterRule: function(cmp) {
-        var defaultFilterRule = {objectName: '', fieldName: '', operatorName: '', constant: ''};
+        var defaultFilterRule = {objectName: '', fieldName: '', operation: '', constant: ''};
```
randi274: We should rename constant to value or valueName here.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -286,6 +399,76 @@
...
+                // Response value will be in the format of "JobId-RecordDeveloperName"
+                var responseText = response.getReturnValue();
+                var jobId = responseText.split("-")[0];
+                var recordName = responseText.split("-")[1];
+
+                console.log('Response = ' + response.getReturnValue());
+                console.log('Returned jobId = ' + jobId);
+                console.log('Returned RecordName = ' + recordName);
+
+                /*cmp.set("v.activeRollup.recordName", recordName);
```
randi274: Can this be removed?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -360,13 +543,15 @@
         }, true);
 
         //check for duplicates
-        if (cmp.get("v.filterRuleMode") === 'create') {
+        if (!filterRuleList) {
+            // nothong to compare
```
randi274: typo :)

src/classes/CRLP_RollupUI_TEST.cls
```diff
@@ -147,32 +149,34 @@ public class CRLP_RollupUI_TEST {
...
     * @description Tests getDeploymentStatus()
     */
-    public static testMethod void testGetDeploymentStatus() {
+    public static testMethod void testGetDeploymentStatus_Rollup() {
 
         // Create the test CMDT data
         mockRollupCMTValues();
 
-        //test a failure
+        // Rest a failure
```
randi274: rest should be test?

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -535,42 +518,60 @@ public with sharing class CRLP_RollupUI_SVC {
             // CMT data will generate a unique RecordName to use for the new record.
             CRLP_RollupCMT.Rollup rollupCMTObj = (CRLP_RollupCMT.Rollup) JSON.deserialize(rollupCMT, CRLP_RollupCMT.Rollup.class);
 
-            // TODO Remove this code after the UI properly sets the fields. For need it to test inserting new rollups
```
randi274: After removing this, I'm getting errors for the amountObject missing. We can put this back, set the amount object directly in the UI, or set it in Apex, but it's not setting correctly just yet.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
force2b: Great catch. I didn't mean to rename the function from `reformatConstantLabel` to `reformatValueLabel` so this was actually failing. It's fixed now.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
force2b: Good question. Does replacing the object directly create a reference clone of the original object? Trying to remember why I'm replacing each property and that's the only thing that comes to mind.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
force2b: The issue here was that the values needed in the RollupsContainer were not quite the same as the straight filterGroupItem wrapper.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
force2b: Yes. Good catch!

PR #3051, opened 2018-04-20
Rollups: Add recalculate button to the Household & Contact page layouts

force2b:
# Critical Changes

# Changes

* The buttons have been there for a while, but where never added to these 2 layouts in the repository.
* The button was already added to the GAU layout in the repo
* The release notes and docs will need to note that existing customers must manually add the button to their Account, Contact and GAU layouts because layout changes are never pushed in a package upgrade.

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
Good question. I only added it to the Household Layout. What do you think @judisohn?

judisohn:
Yes, for sure @force2b - Thanks!

PR #3055, opened 2018-04-21
Add Soft Credit Rollup logic to the Recalculate Rollups button

kofijohnson:
When the user clicks the button "Recalculate Rollups" on the Contact page layout, it calculates the Contact Soft Credits as well.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/CRLP_RecalculateBTN_TEST.cls
```diff
@@ -172,4 +172,367 @@ private class CRLP_RecalculateBTN_TEST {
...
+    */
+    @isTest static void test_CalculateContactSoftCredit_AlwaysRollupToPrimaryContactIsFalse() {
+        String role = 'Soft Credit';
+        UTIL_CustomSettingsFacade.getHouseholdsSettingsForTests(new npo02__Households_Settings__c (
+            npo02__Always_Rollup_to_Primary_Contact__c = false,
+            npo02__Soft_Credit_Roles__c = role
+        ));
+        deactivateTriggers();
+
+        Account household = new Account(
```
force2b: The unit test data set (and most of the logic) looks very very similar for these 4 new test methods. Is it possible to combine these into a single shared method that combines logic where needed. See the `CRLP_RollupContact_TEST` class as an example of using a single test method for 6 different tests.

src/classes/CRLP_RecalculateBTN_CTRL.cls
```diff
@@ -130,4 +130,17 @@ public with sharing class CRLP_RecalculateBTN_CTRL {
         }
         return null;
     }
+
+    /*********************************************************************************************************
+    * @description Calculates the Soft Credit for the Contact.
+    * @return void.
+    */
+    private void calculateContactSoftCredit() {
```
force2b: Do you think it's worth doing this as an asynchronous process? Right now, the logic will complete the Soft Credit rollup synchronously first and then kick off an async job to do the Hard Credits. 

I think these should be consistent - either both synchronous or both asynchronous. Open to discussing this though.

src/classes/CRLP_RecalculateBTN_TEST.cls
force2b: Also, this test class should be focused on testing that the button controller works. Tests that validate the actual rollup logic should be in the unit test classes that focus on rollups. That might be `CRLP_RollupContact_TEST` for Customizable Rollups or `RLLP_OppPartialSoftCreditRollup_TEST` for legacy rollups. I also think the `CRLP_RollupContact_TEST` class is already testing the 'AlwaysRollupToPrimary` logic.

src/classes/CRLP_RecalculateBTN_TEST.cls
```diff
@@ -172,4 +172,367 @@ private class CRLP_RecalculateBTN_TEST {
...
+            WHERE Id = :contact.Id
+        ];
+        System.assertEquals(contactPartialSoftCredit.Amount__c, updatedContact[0].npo02__Soft_Credit_Total__c);
+    }
+
+    /*******************************************************************************************************
+    * @description Deactivates all the Triggers.
+    * @return void.
+    ********************************************************************************************************/
+    private static void deactivateTriggers() {
```
force2b: I'm curious why the tests need to deactivate and then reactivate all the triggers? Also there are methods in `UTIL_UnitTestData_TEST` that will disable specific triggers by type.

src/classes/CRLP_RecalculateBTN_CTRL.cls
kofijohnson: Good point! I don't see the need to make the Hard Credit asynchronous since it is not a heavy process and it runs only on 1 Contact. If you don't see any downside, I can make the Hard Credit synchronous.

src/classes/CRLP_RecalculateBTN_TEST.cls
kofijohnson: The Triggers are deactivated during the creation of the test data to ensure that the Triggers don't alter the test data. After the test data are created, the Triggers are activated.

src/classes/CRLP_RecalculateBTN_TEST.cls
kofijohnson: For this scenario, I think combining the test scenarios into a single method can make it complicated to read. But you have a good point, I can move the creation of the Test data into separate methods that can be called by the test methods.

src/classes/CRLP_RecalculateBTN_CTRL.cls
force2b: I think you need to go the other direction - both asynchronous. Otherwise this will fail on any contact that has a decent number of attached hard credits + soft credits.

src/classes/CRLP_RecalculateBTN_TEST.cls
force2b: Do you need to deactivate all of the triggers, or just specific ones?

src/classes/CRLP_RecalculateBTN_TEST.cls
force2b: Thanks. The tests related to the AlwaysRollupToPrimary option should be moved as well since they're not related to the button.

PR #3064, opened 2018-04-26
Delete a Rollup from the Rollup View in the custom UI

bethbrains:
# Critical Changes

# Changes

- Admin can delete a Rollup from the Rollup View in the custom CRLP UI

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/CRLP_Rollup/CRLP_RollupController.js
```diff
@@ -110,11 +113,16 @@
      */
     onSave: function(cmp, event, helper){
         var activeRollup = cmp.get("v.activeRollup");
-        var canSave = helper.validateFields(cmp);
-        if(canSave){
+        if (cmp.get("v.mode") == 'delete') {
```
force2b: I may have put a comment here already, but if not: `==` in Javascript should be `===`.

src/aura/CRLP_Rollup/CRLP_RollupController.js
```diff
@@ -110,11 +113,16 @@
...
+            helper.toggleModal(cmp);
             helper.saveRollup(cmp, activeRollup);
-            cmp.set("v.mode", 'view');
-            helper.updateRollupName(cmp);
+        } else {
```
force2b: Does it make sense to combine these 3 lines into one?
`} else if (helper.validateFields(cmp) === true) {`

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -603,12 +603,17 @@ public with sharing class CRLP_RollupUI_SVC {
                         if (objectType == 'Rollup') {
                             Id rollupId = CRLP_Rollup_SEL.getRollupIdByDeveloperName(recordName);
                             response.completed = true;
-                            response.rollupItem = createRollupItem(CRLP_Rollup_SEL.getRollupById(rollupId));
+                            if (mode != 'delete') {
```
force2b: Inequality operations should be `!==`

bethbrains: as per quip convo, this one is apex so it's fine.

PR #3066, opened 2018-04-26
Delete Filter Group in custom CRLP UI

bethbrains:
# Critical Changes

# Changes

- Filter Group can be deleted from Filter Group View page. Rules are left untouched.

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -91,7 +91,10 @@
     /**
      * @description: cancels the pop up for filter rule and clears the active filter rule
```
randi274: small add here: need to update the comment with new functionality. You also won't need to run resetActiveFilterRule or set the filterRuleMode if mode == delete since those only apply to the filter rule modal settings.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -148,6 +157,7 @@
             //json shenanigans to avoid shared reference
             cmp.set("v.activeFilterGroup", helper.restructureResponse(cachedFilterGroup));
             cmp.set("v.filterRuleList", cmp.get("v.cachedFilterRuleList"));
+            cmp.set("v.deletedRuleList", []);
```
randi274: Was this added for an existing issue?

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -31,9 +32,18 @@
                 cmp.set("v.isReadOnly", false);
             } else if (mode === "create") {
                 cmp.set("v.isReadOnly", false);
+                cmp.set("v.rollupItems",[]);
```
randi274: also here, were rollupItems carrying over when new FGs were created? I'm wondering if this was an issue with clone too.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -437,13 +437,15 @@ public with sharing class CRLP_RollupUI_SVC {
                 'edit' => Label.stgBtnEdit,
                 'error' => Label.PageMessagesError,
                 'field' => Schema.Filter_Rule__mdt.Field__c.getDescribe().getLabel(),
+                'filterGroupDeleteConfirm' => String.format(Label.CMT_DeleteConfirm, new List<String>{Schema.SObjectType.Filter_Group__mdt.getLabel()}),
```
randi274: I like the reusability of CMT_DeleteConfirm, but we may need to use more specific error messages about the consequences of deleting each of these items. @smarovich made a suggestion for Delete Filter Rule warning in Rollups UI Text, and may be able to suggest text for the filter group and rollup warnings.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
randi274: also found a bug likely from here: hit the cancel button when adding or editing a filter rule and edit mode is being removed.

force2b: Found that one as well

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
```diff
@@ -130,6 +134,11 @@
             //cautions user about deleting filter rule
             cmp.set("v.filterRuleMode", 'delete');
             helper.toggleFilterRuleModal(cmp);
+            if(cmp.get("v.rollupItems").length == 0) {
```
force2b: Javascript equality should be `===`.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
```diff
@@ -31,6 +32,18 @@
...
+            } else if (mode === "delete") {
+                //verify no rollups use the filter group before deleting
+                var countRollups = cmp.get("v.rollupItems").length;
+
+                if (countRollups == 0) {
+                    this.toggleFilterRuleModal(cmp);
```
force2b: Looks like you could put `this.toggleFilterRuleModal(cmp);` outside the if block here since it's used on both sides of the `else`

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -254,6 +255,7 @@ public with sharing class CRLP_RollupUI_SVC {
             item.countFilterRules = mdtList[i].Filter_Rules__r.size();
             item.countRollups = (rollupsByFilterGroup.containsKey(mdtList[i].Id) ? rollupsByFilterGroup.get(mdtList[i].Id).size() : 0);
             item.recordId = mdtList[i].Id;
+            item.isDeleted = mdtList[i].Is_Deleted__c;
```
force2b: Is there ever a scenario where Deleted Filter Groups are being queried and the `item.isDeleted` would have a value other than false?

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -71,6 +71,7 @@ public with sharing class CRLP_RollupUI_SVC {
         @AuraEnabled public Integer countRollups;
         @AuraEnabled public String name;
         @AuraEnabled public String recordId;
+        @AuraEnabled public boolean isDeleted;
```
force2b: `Boolean` should be capitalized

src/classes/CRLP_RollupUI_SVC.cls
bethbrains: good point. changed to assign directly to false, because the value/key is needed on the JS side so we can set it to true upon deletion.

src/aura/CRLP_FilterGroup/CRLP_FilterGroupController.js
bethbrains: comment added for clarity

src/aura/CRLP_FilterGroup/CRLP_FilterGroupHelper.js
bethbrains: great point. moved to a default instead.

src/classes/CRLP_RollupUI_SVC.cls
bethbrains: discussed. leaving as-is for now.

PR #3067, opened 2018-04-27
Fixed Detail Object/Field behavior on Soft Credits and for GAUs

bethbrains:
# Critical Changes

# Changes

- Allow for multiple potential detail objects upon selection of a soft-credit rollup type

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/CRLP_Rollup/CRLP_Rollup.cmp
```diff
@@ -336,10 +340,9 @@
                             <c:CRLP_SelectField isReadOnly="{!v.isReadOnly}"
                                                 auraId="amountField"
                                                 fieldLabel="{!v.labels.amountField}"
-                                                entityLabel="{!v.activeRollup.detailObjectLabel+': '+v.activeRollup.amountFieldLabel}"
+                                                entityLabel="{!v.activeRollup.amountFieldLabel}"
```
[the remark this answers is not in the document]
bethbrains: OK, fixed now @force2b

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -301,8 +301,14 @@ public with sharing class CRLP_RollupUI_SVC {
         Set<String> excludedTypes = new Set<String>{'TIME', 'TEXTAREA'};
 
         for(String obj : objectNames){
-            List<Map<String, String>> detailNameToFieldTypeObj = getFilterRuleFieldsExcludingTypes(obj, excludedTypes);
-            objNameToFieldTypeMap.put(obj, detailNameToFieldTypeObj);
+            if (obj == 'Partial_Soft_Credit__c') {
```
force2b: This should be `UTIL_Namespace.StrAllNSPrefix('Partial_Soft_Credit__c')`

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -301,8 +301,14 @@ public with sharing class CRLP_RollupUI_SVC {
...
+                List<String> fieldsToReturn = new List<String>{'Role_Name__c','Amount__c'};
```
force2b: These field names won't work this way either in the final package. Use the `UTIL_Namespace.StrAllNSPrefix()` method to apply the namespace as needed.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -385,8 +391,14 @@ public with sharing class CRLP_RollupUI_SVC {
         Map<String,List<Map<String,String>>> objNameToFieldTypeMap = new Map<String,List<Map<String,String>>>();
 
         for(String detailObj : detailObjectNames){
-            List<Map<String, String>> detailNameToFieldTypeObj = getAllFieldsWithType(detailObj);
-            objNameToFieldTypeMap.put(detailObj, detailNameToFieldTypeObj);
+            if (detailObj == 'Partial_Soft_Credit__c') {
+                List<String> fieldsToReturn = new List<String>{'Role_Name__c','Amount__c'};
```
force2b: Same here

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -303,8 +303,11 @@ public with sharing class CRLP_RollupUI_SVC {
         Set<String> excludedTypes = new Set<String>{'TIME', 'TEXTAREA'};
 
         for(String obj : objectNames){
-            if (obj == 'Partial_Soft_Credit__c') {
-                List<String> fieldsToReturn = new List<String>{'Role_Name__c','Amount__c'};
+            if (obj == UTIL_Namespace.StrAllNSPrefix('Partial_Soft_Credit__c')) {
```
force2b: @bethbrains We don't need to fix this here given the limited usage, but in general it would be better to assign UTIL_Namespace.StrAllNSPrefix('Partial_Soft_Credit__c') to a var outside the loop rather than call the utility method for each loop iteration.

PR #3080, opened 2018-04-30
Feature/bge nima  bugs s34

mdeleon0:
Fixed/changed the following

- Total of Records incorrect in grid.

- Enter shouldn't go into Edit mode.

- Edit mode displays text in a different place than view mode. The text should not shift between the two.

- Typing is in a different location than final location after enter. It appears to show up in the top left corner when typing and then after enter it will center its location. We want the text location not to change between view and edit modes.

- Remove “Commit and Save” and “Commit and Exit” Buttons, replaced with single Commit button, which redirects to the process batch page.

- Right Click Context Menu: Remove "Read Only" and "Alignment".


src/pages/BGE_HandsOnGrid.page
```diff
@@ -88,16 +96,22 @@
...
 
-		//Map to save cells with invalid format values
+                // Flag that is set true if the selected cell have just moved horizontally.
+                movedSideWays = false;
+
+                // Flag needed to fix a bug, if true saving when the user changes row is delayed for a second.
+                wait = false;
+
+        		$scope.totalPages = '{!totalOfPages}';
+
```
ricardonimacloud: @mdeleon0 please review identation

PR #3101, opened 2018-05-02
Fix Duplicate Primary OCRs Bulk Job

### LINKED ISSUE #3063: Remove Duplicate Primary OCR bulk job: If two Primary OCRs have Roles that do not match default in NPSP Setting, one is removed when it should log as an error
NPSP 3.126

STR:

1. Use LEX standard New button on related list to create Opportunity from Contact. 
2. Manually edit the duplicate OCR roles to select different roles (such as 'Decision Maker' and 'Influencer)
3. Run the new Bulk Process batch to resolve duplicates

Should report an error since it can't resolve which OCR is the correct one to retain. Instead, one OCR is deleted and unrecoverable.

kofijohnson:
# Critical Changes

# Changes

# Issues Closed
- Fixes #3063

# New Metadata

# Deleted Metadata


kofijohnson:
@smarovich, do we need to add more detail in the release note?

PR #3113, opened 2018-05-04
Fix the package check for Adv

weizsforg:
Update the Adv package check with the describe call. So, the code will work in both manage and unmanaged environment. 
Note for testing: This PR does not change any NPSP logic. So, the normal regression test of NPSP should be sufficient.
# Critical Changes

# Changes

# Issues Closed

# New Metadata
npe01__Contacts_And_Orgs_Settings__c.Advancement_Namespace__c (Custom Field)
# Deleted Metadata


weizsforg:
Hi @njjc @ceiroa @mpusto . I just opened this PR in order to fix a issue that related to Gift Processing. No NPSP logic change in here.

weizsforg:
Hi @juliesnow, I added a custom field to store the Gift Processing Namespace in NPSP custom setting. This field is mainly for internal use although it is visible for Sys Admin. Can you review the description and help text of the field? 

Description: The namespace of the Gift Processing package.
inlineHelpText: The namespace of the Gift Processing package.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
```diff
@@ -1,5 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fields>
+        <fullName>Advancement_Namespace__c</fullName>
+        <defaultValue>&quot;adv1&quot;</defaultValue>
+        <description>The namespace of the Gift Processing package.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The namespace of the Gift Processing package.</inlineHelpText>
```
njjc: Is the name "Gift Processing" final?

ceiroa: Define "final" ;-).  But yeah, that's what it's being called now.

juliesnow:
Hi @weizsforg - Same comment as last: "Your org is not configured for Gift Processing. Contact your admin for help." (I'd change "admin" to "administrator" unless the shorter version is standard practice.) 

Or, perhaps the other one I proposed: "You are trying to access Gift Processing without the required license or configuration information. Contact your administrator for help."

weizsforg:
Hi @juliesnow , actually the error msg is for HEDA branch. For NPSP, we only have the new field. Thanks!

src/classes/ADV_PackageInfo_SVC.cls
```diff
@@ -61,43 +61,69 @@ public with sharing class ADV_PackageInfo_SVC {
...
+
+        if (String.isBlank(orgSetting.Advancement_Namespace__c)) {
+            orgSetting.Advancement_Namespace__c = String.valueOf(npe01__Contacts_And_Orgs_Settings__c.Advancement_Namespace__c.getDescribe().getDefaultValueFormula()).remove('\"');
+        }
+
+        //Check if Advancement is installed
         try {
-            //If the Advancement is not installed, then it will throw TypeException. Otherwise, it will return True/False
-            isUserLicensed = UserInfo.isCurrentUserLicensed('adv1');
+            Schema.DescribeSObjectResult[] descResult = Schema.describeSObjects(new String[]{orgSetting.Advancement_Namespace__c + '__Advancement_Setting__mdt'});
```
force2b: @weizsforg  I'm curious why this is implementing a new custom settings field for package the namespace value. Is there a chance that the package namespace can be different in different orgs?

weizsforg: @force2b In general, namespace will be the same. However, we want to make the namespace configurable for the future testing and development. In same cases, we want to do some investigations and experiment for Adv pacakge. And it is not always good to cut a real Bata package (with adv1 namespace) for that. So, we will use a developer org with different namespace to create the package. In NPSP, if we hardcode the namespace, then it will be lock to a specific namespace.

weizsforg: @force2b That is a great point!!
Actually, We had a discussion with @njjc before about similar stuff in this class. And we are trying to do less query in this class if possible. And that is why we think custom setting might be the simplest way to store the namespace for advancement. 
Any ideas @njjc @ceiroa ?

force2b: Good point. Querying CMT's will use up one query and 1 query row. Like Custom Settings, querying the CMT should be fast, and you can cache the setting in a static var. But, as you point out it's an extra query.

force2b: @weizsforg Something else occurred to me earlier. Custom Settings records are not copied to Developer and DeveloperPro sandboxes, so if you have Advancement installed in a Production org with a record in the `npe01__Contacts_And_Orgs_Settings__c` object and a value in the `Advancement_Namespace__c` field; and the Admin creates a new or refreshes an existing Developer or Developer Pro sandbox, this value will not persist into the Sandbox. That could still work if there is other logic to automatically populate the field value such as how the `Trigger_Handler__c` object is auto-populated by the `TDTM_Trigger_Handler` class if there are no records in the object. One of the benefits of using Custom Metadata Types is that as metadata, these _records_ are copied into all Sandboxes because they're treated as _metadata_ as opposed to _data_. 

CMT's have their complexities as well, such as the fact that inserting/updating them is done through the asynchronous apex metadata api as opposed to dml. It's definitely something that can be accomplished in apex, and more than likely in a package install script. Not as quick and dirty as DML into a Custom Settings object though.

weizsforg: @force2b That is true that the setting value will not get copy to the sandbox. Actually, we do not expect Admin to do put anything in the setting (not sure why Admin wants to put something into this field if they know they are not using it). Our code here will get the value from its default value if the field is empty. Also, I think this field is mainly use by a developer to dynamically refer to a different namespace for Adv package.

force2b: @weizsforg So the field is basically only for burner namespaces during the development/test cycle so that managed betas or full packages can be installed using a namespace other than "adv1". Makes sense. I was trying to find a way to avoid packaging a custom setting field because some admins do actually play around in Custom Settings.

weizsforg: @force2b Yes, I totally agree with you! 
I was thinking to use a custom label to store it since we can't hide the setting from admin. However, it is a little bit strange to do it since usually custom label is from multi-language. Please let me know if you have other thoughts that we can do an enhancement later.

force2b: Actually I think a custom label is good idea. It's a lot more hidden than a Custom Setting and I would not expect anyone to attempt to translate the label into a different language. In fact, the label description can say "Do Not Translate". I'm not even sure most admins would even be able to find the custom label given that we have so many others. I believe it's also a lot easier to remove from the package than a custom settings field.

weizsforg: Yes, I can talk with @ceiroa  about this. Since in HEDA, we have similar class. And we can use label is HEDA if needed. Maybe you can un approve this PR for now?

weizsforg: oh, actually if we use Custom Label, then we can't change it after we deploy the package in the org. So, we lose the ability to change the value like Custom Setting. Sorry about the confusion.

force2b: Managed custom labels can changed using the [New Local Translation/Override] button on the label.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
njjc: Hi @smarovich, this custom setting field is being added to our product to enable internal developers. Customers never need to understand what it is, and they can't use it for anything. I'm thinking:

"System use only: namespace for cross package functionality"

Thoughts?

smarovich: Looks good, @njjc. Some minor edits: "For system use only: Namespace for cross package functionality."

PR #3115, opened 2018-05-04
Rollup enablement page fix

bethbrains:
# Critical Changes

# Changes

- Hides GAU Rollup settings when CRLPs are enabled.
- Does NOT re-deploy CRLP defaults upon edit/save if value stays true.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/STG_PanelCustomizableRollup_CTRL.cls
+++ b/src/classes/STG_PanelCustomizableRollup_CTRL.cls
@@ -89,24 +89,18 @@ public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
     public virtual override PageReference saveSettings() {
 
         STG_SettingsManager_CTRL.idPanelCurrent = idPanel();
-        Customizable_Rollup_Settings__c crlpSettings = UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
         Savepoint sp = Database.setSavepoint();
         isEditMode = false;
         try {
             if (STG_Panel.stgService.stgCRLP.Customizable_Rollups_Enabled__c) {
-                String oldStatus = STG_Panel.stgService.stgCRLP.CMT_API_Status__c;
-                // Only start polling if it's their first time or the last deployment wasn't a success
-                if(String.isEmpty(oldStatus) || !oldStatus.contains('Succeeded')) {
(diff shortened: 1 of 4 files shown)


force2b:
@bethbrains  Since this is on enablement I'll note this issue that I found. When Customizable Rollups is enabled, the "GAU ALLOCATIONS ROLLUP SETTINGS" section on the "GAU Allocations" page in NPSP Settings should be hidden.

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -81,6 +81,23 @@ public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
     */
     public String jobId { get; private set; }
 
+    /*******************************************************************************************************
+    * @description the existing value of the CRLP enablement setting, cached upon entering edit mode
+    */
+    private Boolean cachedSetting;
```
force2b: This should have a more descriptive name such as `isCRLPEnabledCached`

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -89,33 +106,40 @@ public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
...
-                // This method includes committing Customizable_Rollups_Enabled__c as true if the deployment succeeds
-                jobId = CRLP_DefaultConfigBuilder_SVC.convertLegacyRollupsIntoCustomizableRollups();
-                if (Test.isRunningTest()) {
-                    jobId = '123';
-                }
-            } else {
-                // The deployment method takes care of setting the Enabled flag to true if it succeeds.
-                // In here, we only set it to false.
-                stgService.saveAll();
+                UTIL_MasterSchedulableHelper.setScheduledJobs();
```
force2b: If the Custom Setting isn't being set to true the deployment completes, then the call to `UTIL_MasterSchedulableHelper.setScheduledJobs();` probably needs to move because that has method has code to check for the current setting to determine which jobs to schedule (legacy or crlp rollups). Also, what if the user enables CRLP and the job fails. Are the new jobs scheduled anyway?

PR #3118, opened 2018-05-06
Rollups: Extend CRLP debugging tool for customer package deployment

force2b:
# Critical Changes

# Changes

* Enable the `CRLP_Debug_UTIL` class to be included in the managed package
* Debugging is disabled until the admin adds a custom field for `Debug_Level__c` to the `Customizable_Rollups_Settings` custom settings object. This is done to avoid packaging a custom field for debugging.
* The initial intent is to remove the class shortly after the initial go live; leaving it there only to troubleshoot any unexpected rollup issues post go-live. However, it is possible this will be left in long term, in which case the debugging logic might be expanded.
* It is important to note that enabling debugging during the rollup has a pretty significant impact on performance.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Debug_UTIL.cls
+++ b/src/classes/CRLP_Debug_UTIL.cls
@@ -1,29 +1,92 @@
+/*
+    Copyright (c) 2018, Salesforce.org
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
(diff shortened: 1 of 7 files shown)


src/classes/CRLP_Debug_UTIL.cls
```diff
@@ -1,37 +1,100 @@
...
+*/
 /**
-* ====================================
-* TODO REMOVE THIS CLASS EVENTUALLY
-* ====================================
-* TEMPORARY STATIC VALUE AND METHODS TO HOLD CURRENT STATE TO USE FOR ROLLUP DEBUGGING
+* @author Salesforce.org
+* @date 2018
+* @group Customizable Rollups Debug Logger
+* @description TEMPORARY USE ONLY
```
njjc: Let's update this description to reflect the purpose of the class, we may end up leaving it in permanently!

src/classes/CRLP_Debug_UTIL.cls
```diff
@@ -1,37 +1,100 @@
...
+     * Level 0 = OFF
+     * Level 1 = MINIMAL
+     * Level 2 = DETAILED
+     */
+    public static Integer debuggingLevel {
+        get {
+            if (debuggingLevel == null) {
+                Customizable_Rollup_Settings__c crlpSettings = UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
+                try {
+                    debuggingLevel = ((Double) crlpSettings.get('Debug_Level__c')).intValue();
```
njjc: Should we package Debug_Level__c? Will we ever need it to be an integer, or can we just make it a checkbox? @force2b  (cc @bethbrains)

bethbrains: I like it as a (default null) integer so someday we could actually have different levels based on the integer value. (1=light logging, 2=medium logging, 3=omgwhat'sgoingonlogallthethings)

force2b: Actually it already supports 2 levels:
* 0/null = disabled
* 1=Level 1 debug (only write a summary at the end of the rollups)
* 2=Level 2 debug (collect more information to write at the end, but will have an impact on performance).

kofijohnson: I would suggest using Text instead of Integer for the Debug_Level__c field. That way, it will be more readable to developer/admin. So the reader won't need to look for what is the meaning of the numbers. Also, in the code, we can use Enums instead of Integer. I'm thinking about something like:

public enum DebuggingLevel {OFF, MINIMAL, DETAILED}

To be more explicit, should we remove the DebuggingLevel OFF, and have a field Active__c (checkbox) instead?

force2b: I don't foresee this being something we'll ever want to give directly to customers to use. It's been built specifically for our own debugging use only. Thus the reason to not package the field. The class captures debug data, but what it generates will not be helpful at all to customer admins. I'd much prefer to keep this both extremely simple and somewhat obfuscated from customer admins.

src/classes/CRLP_Debug_UTIL.cls
```diff
@@ -91,14 +157,14 @@ public class CRLP_Debug_UTIL {
      * all records no matter how many records there are in the system.
      * @param summObjectType Summary Object Type
      */
-    public static void clearAllRollupFields(SObjectType summObjectType) {
+    /*public static void clearAllRollupFields(SObjectType summObjectType) {
```
kofijohnson: @force2b, because this is commented, could we remove it?

src/classes/CRLP_Debug_UTIL.cls
```diff
@@ -91,14 +157,14 @@ public class CRLP_Debug_UTIL {
...
         System.enqueueJob(new clearRollupFieldsASync(summObjectType, 0));
-    }
+    }*/
 
     /**
      * @description Queueable Implementation called by clearAllRollupFields()
      */
-    private class clearRollupFieldsASync implements System.Queueable {
+    /*private class clearRollupFieldsASync implements System.Queueable {
```
kofijohnson: @force2b same here too.

PR #3125, opened 2018-05-07
Feature/rollups  randi507bugs

randi274:
# Critical Changes

# Changes
- W-024497: Days Back Help Text Fix
- W-024496: Field to Roll Up calculating in 'Edit' mode from row action menu
- W-024518: Days Back null value handling (field is required if displayed)
- Fixed demo issue: filter groups can be successfully deleted again
- Resolved issue where Field Labels and Schema Describes were causing alternative labels on view and edit modes for rollups and filter rules.
# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/aura/CRLP_Rollup/CRLP_Rollup.cmp
+++ b/src/aura/CRLP_Rollup/CRLP_Rollup.cmp
@@ -208,16 +208,17 @@
                             <lightning:layoutItem padding="around-medium" smallDeviceSize="6" mediumDeviceSize="3"
                                                   size="12">
                                 <aura:if isTrue="{!not(v.isReadOnly)}">
-                                    <c:CRLP_Tooltip helpText="{!v.labels.createTimeBoundHelp}" fieldLabel="{!v.selectedTimeBoundOperationLabel}"/>
+                                    <c:CRLP_Tooltip helpText="{!v.labels.createDaysHelp}" fieldLabel="{!v.selectedTimeBoundOperationLabel}" isRequired="true"/>
                                 </aura:if>
                                 <lightning:input aura:id="integerInput"
                                                  type="text"
                                                  label="{!v.selectedTimeBoundOperationLabel}"
                                                  value="{!v.activeRollup.intValue}"
                                                  readonly="{!v.isReadOnly}"
-                                                 pattern="^[0-9]*$"
+                                                 pattern="^\d{1,4}|10000$"
                                                  messageWhenPatternMismatch="{!v.labels.daysBackError}"
-                                                 variant="{!v.isReadOnly ? '' : 'label-hidden'}"/>
+                                                 variant="{!v.isReadOnly ? '' : 'label-hidden'}"
+                                                 required="true"/>
                             </lightning:layoutItem>
                         </aura:if>
                         <aura:if isTrue="{!v.renderMap.integerYears}">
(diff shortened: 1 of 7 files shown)


src/aura/CRLP_Rollup/CRLP_RollupHelper.js
```diff
@@ -160,9 +160,8 @@
             cmp.set("v.detailFields", newFields);
         }
 
-        //todo: is this necessary? pulled in my commit
         var currentMode = cmp.get("v.mode");
-        if (currentMode !== 'view') {
+        if (currentMode === 'create') {
```
force2b: This resets the field to null for a new record, but does it prevent the field from being reset to null when cloning?

src/classes/CRLP_RollupCMT.cls
```diff
@@ -302,7 +302,7 @@ public class CRLP_RollupCMT {
                 this.summaryObjectLabel = rlp.Summary_Object__r.Label;
                 this.summaryFieldLabel = rlp.Summary_Field__r.Label;
                 this.detailObjectLabel = rlp.Detail_Object__r.Label;
-                this.detailFieldLabel = rlp.Detail_Field__r.Label;
```
force2b: Good question as to why one of these used `.Label` and the other used `.MasterLabel`. I thought I had copied these from `CRLP_RollupUI` class, but maybe I did a little of both. That noted, I still see a mix of `MasterLabel` and `Label` used for getting entity reference labels in this class.

src/aura/CRLP_Rollup/CRLP_Rollup.cmp
```diff
@@ -208,16 +208,17 @@
...
                                 <aura:if isTrue="{!not(v.isReadOnly)}">
-                                    <c:CRLP_Tooltip helpText="{!v.labels.createTimeBoundHelp}" fieldLabel="{!v.selectedTimeBoundOperationLabel}"/>
+                                    <c:CRLP_Tooltip helpText="{!v.labels.createDaysHelp}" fieldLabel="{!v.selectedTimeBoundOperationLabel}" isRequired="true"/>
                                 </aura:if>
                                 <lightning:input aura:id="integerInput"
                                                  type="text"
                                                  label="{!v.selectedTimeBoundOperationLabel}"
                                                  value="{!v.activeRollup.intValue}"
                                                  readonly="{!v.isReadOnly}"
-                                                 pattern="^[0-9]*$"
```
force2b: I noticed that putting in a value of "-1" into the "Days Back" field will generate a validation message, but it will still allow the rollup to be saved.
image

src/aura/CRLP_Rollup/CRLP_RollupHelper.js
randi274: Yep! 'clone' is its own separate mode, so this fixes that problem too.

src/aura/CRLP_Rollup/CRLP_Rollup.cmp
randi274: Good catch, I'll fix this one too!

randi274: @force2b this has been fixed together with the 0 value issue in the latest commit.

PR #3130, opened 2018-05-09
fixed so we don't send a null value to the getFieldLabel util method,…

bethbrains:
… which errors

Fixes https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x8000000AWZTtAAP/view

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupCMT.cls
+++ b/src/classes/CRLP_RollupCMT.cls
@@ -299,7 +299,7 @@ public class CRLP_RollupCMT {
                 this.summaryObjectLabel = rlp.Summary_Object__r.Label;
                 this.summaryFieldLabel = rlp.Summary_Field__r.Label;
                 this.detailObjectLabel = rlp.Detail_Object__r.Label;
-                this.detailFieldLabel = rlp.Detail_Field__r.QualifiedApiName == null ? null : UTIL_Describe.getFieldLabel(rlp.Detail_Object__r.QualifiedApiName, rlp.Detail_Field__r.QualifiedApiName);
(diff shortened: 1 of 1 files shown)


src/classes/CRLP_RollupCMT.cls
```diff
@@ -299,7 +299,7 @@ public class CRLP_RollupCMT {
                 this.summaryObjectLabel = rlp.Summary_Object__r.Label;
                 this.summaryFieldLabel = rlp.Summary_Field__r.Label;
                 this.detailObjectLabel = rlp.Detail_Object__r.Label;
-                this.detailFieldLabel = UTIL_Describe.getFieldLabel(rlp.Detail_Object__r.QualifiedApiName, rlp.Detail_Field__r.QualifiedApiName);
+                this.detailFieldLabel = rlp.Detail_Field__r.QualifiedApiName == null ? null : UTIL_Describe.getFieldLabel(rlp.Detail_Object__r.QualifiedApiName, rlp.Detail_Field__r.QualifiedApiName);
```
randi274: Does this change also need to be made in the filterGroup code?

force2b: Also, I recommend changing this to read: `rlp.Detail_Field__c == null` so that you're checking the root field value versus the relationship of QualifiedFieldName against a potentially null `Detail_Field__c` value.

bethbrains: @randi274 no because it'll never be null there, right? all fields are required.

PR #3140, opened 2018-05-10
Prevent Duplicate OCR's from rolling up Soft Credits to a single Contact

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -130,7 +129,8 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
                         parentRelationshipPrefix = Opportunity.fields.Primary_Contact__c.getDescribe().getName().replace('__c', '__r') + '.';
                     } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
                         parentRelationshipPrefix = 'Contact.';
-                    } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
+                    } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit || 
+                            this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
                         parentRelationshipPrefix = 'Contact.Account.';
```
force2b: This isn't one of those if blocks that needed changing. If you look at line 130 above this it's already checking for `(this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit)` so the change made in line 132 will never find match the `||` condition. Plus the `parentRelationshipPrefix` for a Contact rollup should not be `'Contact.Account.'` because that points to the AccountId as opposed to the ContactId.

You can remove this change.

src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -165,7 +165,8 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
                             this.statefulCacheOfRollupsToBeProcessed);
                 } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
                     addlFilter = CRLP_Query_SEL.buildCommonQueryFilterOnOpportunity(Contact.SObjectType, Partial_Soft_Credit__c.SObjectType);
-                } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
+                } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
```
force2b: This also needs to be removed because the code in the if block is extremely specific Accounts, not Contacts. Line 170 below (the `addFilter=` line) shows that it's specific to Accounts).

src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -216,7 +217,8 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
             // parentKeyField needed for the rollup logic.
             if (this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
                 parentKeyField = 'ContactId';
-            } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
+            } else if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit || 
+                    this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
```
force2b: Remove this too.

src/classes/CRLP_RollupSoftCredit_SVC.cls
```diff
@@ -78,9 +78,9 @@ public class CRLP_RollupSoftCredit_SVC extends CRLP_VRollupHandler {
             Partial_Soft_Credit__c psc = pscs[n];
             Opportunity opp = psc.Opportunity__r;
 
-            // Account Soft Credit Rollups are unique in that a single Opportunity cannot be rolled up to
-            // the same parent Account more than one time.
-            if (summaryObjectType == Account.SObjectType && this.processedOppIds.contains(opp.Id)) {
+            // Account and Contact Soft Credit Rollups are unique in that a single Opportunity cannot be rolled up to
+            // the same parent Account or Contact more than one time.
+            if ((summaryObjectType == Account.SObjectType || summaryObjectType == Contact.SObjectType) && this.processedOppIds.contains(opp.Id)) {
```
force2b: Currently there is only Account and Contact soft credits, so this if block could just be changed to `if (this.processedOppIds.contains(opp.Id)) {`, skipping the check for the Summary Object Type. I don't foresee us ever rolling up soft credits to a different object, so that single check would be fine.

src/classes/CRLP_RollupSoftCredit_TEST.cls
```diff
@@ -288,14 +288,16 @@ private class CRLP_RollupSoftCredit_TEST {
         insert opps;
 
         // Change the OCR Role on the org donation so that it will roll up as a soft credit.
-        OpportunityContactRole orgOCR = [SELECT Id, Role FROM OpportunityContactRole
+        OpportunityContactRole orgOCR = [SELECT Id, OpportunityId, ContactId, Role FROM OpportunityContactRole
                     WHERE Opportunity.AccountId = :orgAcct.Id AND IsPrimary = True LIMIT 1];
         orgOCR.Role = label.npo02.Household_Member_Contact_Role;
-        update orgOCR;
+        OpportunityContactRole duplicateSoftCreditOCR = orgOCR.clone();
```
force2b: Soft Credit rollups will only roll up entries with `IsPrimary`=false, so cloning the queried row shouldn't have an impact here. It would be better to create a new OCR with the same Role, but making sure that `IsPrimary` is false. 

To really test this out across a bunch of records, I suggest doubling every non-primary OCR record using something along the lines of:

```
List<OpportunityContactRole> dupeOCRs = new List<OpportunityContactRole>();
for (OpportunityContactRole ocr : [SELECT OpportunityId, ContactId, Role FROM OpportunityContactRole WHERE IsPrimary=False AND Opportunity.IsWon=True]) {
   dupeOCRs.add(new OpportunityContactRole(
      OpportunityId = ocr.OpportunityId,
      ContactId = ocr.ContactId,
      Role = ocr.Role,
      IsPrimary = false));
}
insert dupeOCRs;
```

src/classes/CRLP_RollupProcessingOptions.cls
```diff
@@ -155,7 +155,7 @@ public class CRLP_RollupProcessingOptions {
          * Used to expose this value to the LDV Batch Job to retrieve the current setting.
          */
         public Map<Id,Set<Id>> getUpdatedMapOfProcessedOppsByAccountId() {
```
force2b: This method name can also be refactored since it's no longer retrieving the map "ByAccountId". Sorry if this was missed in the WI.

src/classes/CRLP_RollupSoftCredit_TEST.cls
kofijohnson: @force2b, when I made the change by cloning the Primary OCR, the Soft Credit rollup amount double (Test failed). So I think the Rollup count Primary OCR as well. I will double check. But I get your point, I will create non-primary OCRs.

src/classes/CRLP_Batch_Base_Skew.cls
kofijohnson: @force2b that makes sense. I will update.

src/classes/CRLP_RollupSoftCredit_TEST.cls
force2b: Very interesting. Let's test a double IsPrimary OCR outisde of a unit test to see what happens.

src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -216,7 +215,7 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
             // parentKeyField needed for the rollup logic.
             if (this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
                 parentKeyField = 'ContactId';
-            } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
```
force2b: Minor, but the `this.` part should still be there.

PR #3141, opened 2018-05-10
Rollups: Filter Rules accurately save Record Type Developer Names

bethbrains:
# Critical Changes

# Changes

-Fixed FR behavior so the value saved is the Developer Name and the value displayed in the UI is the Name/label.

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/CRLP_RollupCMT.cls
```diff
@@ -168,10 +168,33 @@ public class CRLP_RollupCMT {
...
+                this.valueLabel = getValueLabel(this.value, this.objectName, this.fieldName);
                 this.isDeleted = filterRule.Is_Deleted__c;
             } catch (Exception ex) {}
         }
+        
+        private String getValueLabel (String value, String objectName, String fieldName) {
+            String valueLabel;
+            List<String> rtLabels = new List<String>();
+            if (fieldName == 'RecordTypeId') {
+                List<RecordType> rts = [SELECT Id, Name, DeveloperName FROM RecordType WHERE SobjectType=:objectName];
```
force2b: This query should be limited to Active=True

PR #3146, opened 2018-05-11
Feature/rollups  randi508bugs

randi274:
# Critical Changes

# Changes
- Updated widths on filter group page
- Refactored delete code to remove duplicate code
- Set focus instead of showHelpMessageIfInvalid message on filter group page to avoid stateful errors
- Improve save action to check for more relevant fields before saving without all fields
- Refactored delete logic in container to minimize duplicate code
- Fix for W-024385: Ensure Cancel Button Works The First Time It's Clicked
- Fix for W-024432: how to handle target and detail fields when no options are available
- Fix duplicate bug issue for detail field picklist
# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupUI_SVC.cls
+++ b/src/classes/CRLP_RollupUI_SVC.cls
@@ -69,7 +69,7 @@ public with sharing class CRLP_RollupUI_SVC {
         @AuraEnabled public String description;
         @AuraEnabled public Integer countFilterRules;
         @AuraEnabled public Integer countRollups;
-        @AuraEnabled public String name;
+        @AuraEnabled public String recordName;
         @AuraEnabled public String recordId;
         @AuraEnabled public Boolean isDeleted;
     }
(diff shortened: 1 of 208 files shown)


src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -857,6 +859,13 @@ public with sharing class CRLP_RollupUI_SVC {
                 fieldObjectList.add(fieldObject);
             }
         }
+        /*FOR TESTING*/
```
randi274: this is in here for @dgamboni to complete testing on the warning toast when no fields are available. We'll need to create a break out bug for when the mode isn't create, but the rest is ready for testing.

force2b:
When creating a new rollup and there are no valid fields available, a nice toast is displayed. However, Save still works and the user gets an error during the save.
image

bethbrains:
image
Freshly re-deployed this branch, and I'm showing the save button as deactivated in that case...can you re-deploy and check again?

As for PSC Role Name, that is a text field and always has been, so I don't think this is an issue.

bethbrains:
image
This text needs adjusting though, to not explicitly reference a picklist. @smarovich

force2b:
@bethbrains and @randi274  I agree that the issue with the enabled save button cannot be reproduced. Strange.

I think we did discuss allowing the Soft Credit Role to go out as a plain text field instead of as a picklist for MVP. I remember the conversation, but I didn't see a work item so I created one that we should pull in after this sprint (W-024721). In the mean time, we'll need to have the documentation explain how to use the field in this format.

PR #3155, opened 2018-05-14
Feature/rollups  randi514bugs

randi274:
# Critical Changes

# Changes
- W-024722: Removed geolocation compound fields
- (Modified) W-024723: Removed "Best Year" and "Years Donated" as Operations for Picklist/Multipicklist (moved the logic into the JS) --> Removed this, and made other modifications.
- When rollup type is changed, detail field is now cleared. Prevents errors when changing rollup type and getting a save error if the detail field isn't selected.
- Date and Amount field labels being set accurately when the rollup type is changed.
- When rollup is deleted or the filter group reference is changed, the filter group list is updated to reflect the accurate count.
- (Added) When filter group is edited, the count of rollups on the filter group grid is now correct for that filter group.
- (Added) Fixed component bug where integerInput couldn't be found. To test: save rollup with and without a days back value.
- (Added) Added try/catch blocks to the CRLP_RollupUI_SVC
- (Added) When a filter group has no rollups, then a rollup is modified to use that rollup, prevent the FG from being deleted on the server side with an exception. 
# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
Found this issue after adding a new filter group:
image

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
```diff
@@ -131,11 +131,11 @@
 
         if (context === 'rollup') {
             this.resetRollupDataGrid(cmp, list);
-            //todo: MAYBE change this to only be called when filter group is changed
-            this.requeryFilterGroups(cmp);
         } else if (context === 'filterGroup') {
             cmp.set("v.filterGroupList", list);
         }
+        //todo: MAYBE change this to only be called when filter group is changed
+        this.requeryFilterGroups(cmp);
```
randi274: @bethbrains why did this one need to be called in both contexts?

force2b:
@bethbrains  Thanks for reverting, but the options are still too limited. Though "Years Donated" is a good fit for a multi-select picklist it's not meant to be *only* for a multi-select picklist. This operation type should work for text and picklist fields as well. Generally I thought that we were treating all 3 of these types the same (Text == Picklist == Multi-Select). I was expecting that this should not be any different for these operation types. Is there a specific use-case in mind that you feel requires restrictions on these operations?

randi274:
@force2b that was the way it was written before to allow picklists/multipicklists to rollup into strings, but @njjc recommended we make the change with W-024723. We can pull this part out if you feel it's too limited and Nic agrees!

src/aura/CRLP_RollupsContainer/CRLP_RollupsContainerHelper.js
randi274: Ah jk I see why now. Thanks!

force2b:
Found this somewhat minor issue.

PR #3160, opened 2018-05-15
Rollups: Spring cleaning, realphabetizing package.xml

bethbrains:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/package.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
njjc: I keep seeing this stupid standalone appearing and disappearing. We don't care about it, right @cdcarter ?

src/package.xml
```diff
@@ -960,6 +952,26 @@
...
+        <members>Schedulable__c.Last_Time_Run__c</members>
+        <members>Trigger_Handler__c.Active__c</members>
+        <members>Trigger_Handler__c.Asynchronous__c</members>
+        <members>Trigger_Handler__c.Class__c</members>
+        <members>Trigger_Handler__c.Load_Order__c</members>
+        <members>Trigger_Handler__c.Object__c</members>
+        <members>Trigger_Handler__c.Trigger_Action__c</members>
+        <members>Trigger_Handler__c.User_Managed__c</members>
+        <members>Trigger_Handler__c.Usernames_to_Exclude__c</members>
+        <members>Filter_Group__mdt.Description__c</members>
```
force2b: @bethbrains This is definitely not in alphabetical order (F does not come after T).

src/package.xml
```diff
@@ -1957,9 +1955,11 @@
         <members>Opportunity_Naming_Settings__c</members>
         <members>Partial_Soft_Credit__c</members>
         <members>Relationship_Sync_Excluded_Fields__c</members>
-        <members>Rollup__mdt</members>
         <members>Schedulable__c</members>
         <members>Trigger_Handler__c</members>
+        <members>Filter_Group__mdt</members>
```
force2b: Same here. The CMT objects should not come after "T".

bethbrains:
@force2b correction....this is cleaned up to how `cci task run update_package_xml` does things. Interesting discrepancies... @cdcarter ?

force2b:
@cdcarter Can you clarify what the sequence of MDT objects/fields should be in the package.xml? Historically we've always alphabetized the members by the full name (i.e., `Filter_Rule__mdt` comes before `Trigger_Handler__c`). That no longer appears to be the case. I'd like to be sure that the correct sort is to put all the CMT objects/fields at the end rather than sort alphabetically before we merge this.
@bethbrains @njjc

PR #3162, opened 2018-05-15
Feature/rollups  dupe target fix

bethbrains:
# Critical Changes

# Changes

- When a deactivated rollup is reactivated from the custom UI, its target field is checked against existing active rollups before save. If it matches one, save is blocked and helpful message is exposed to the user.

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@bethbrains The code looks like there are more changes than just the one item listed above (such as field types), or are all those changes related to the dupe check?

bethbrains:
@force2b I accidentally rebased off Randi's branch. If that merge gets propogated down correctly, this one should functionally be only one commit.

PR #3168, opened 2018-05-16
Membership Rollups inactive if no Membership RT setting

bethbrains:
# Critical Changes

# Changes

- Membership rollups are inactive on enablement if no membership record types are defined in legacy settings

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@bethbrains The change appears to work fine, but there is a unit test failure: https://mrbelvedereci.herokuapp.com/builds/25551/tests

PR #3180, opened 2018-05-18
Rollups: Enablement Fix for RecordTypes 

bethbrains:
# Critical Changes

# Changes

* Fix the enablement process to not exclude Inactive RecordTypes during the conversion from legacy RecordType.Id to RecordType.DeveloperName when building CRLP filters
* Remove the default RecordType for building the CRLP filters. All values should always come directly from the custom settings object, even if the field in the custom settings is null
* In Skew Mode jobs, significantly increase the `MAX_BATCHES_TO_HOLD_BEFORE_COMMIT` value to avoid unnecessary commits to the parent object.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Batch_Base_Skew.cls
+++ b/src/classes/CRLP_Batch_Base_Skew.cls
@@ -44,7 +44,7 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
      * this number of batch iterations. After that number, the record will be committed to the database.
      * Salesforce can easily put data significantly out of sequence -- hundreds or even thousands of batch iterations.
      */
-    private Integer MAX_BATCHES_TO_HOLD_BEFORE_COMMIT = 200;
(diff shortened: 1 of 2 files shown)


force2b:
@bethbrains Though this technically fixes the problem in question, it appears the case was slightly more nuanced. The Id value left in the string was actually a valid `RecordType.Id`, but it belonged to a RecordType that was inactive. So, I believe the additional fix we need to make is to ensure the query of RecordType that populates the `recordTypesById`  map does not exclude inactive record types. Between the two changes that completely resolve the problem.

njjc:
@force2b I'm curious why the 'Membership' record type dev name is hard coded in a bunch of places in that class.

force2b:
@njjc  Looks like when the code was first written I defined `Membership` as the default record type to exclude if the custom settings field was null. That's definitely not correct though. It should use whatever value is in the custom settings field. I'll remove.

PR #3183, opened 2018-05-21
Rollups: Final rollup changes

force2b:
* Create the Customizable_Rollup_Settings record during package install. Otherwise found that the next time this was referenced was by the scheduled batch jobs at night, all of which attempted to insert the default record at the same time causing an error.
* Fix the enablement process to not exclude Inactive RecordTypes during the conversion from legacy RecordType.Id to RecordType.DeveloperName when building CRLP filters
* Remove the default RecordType for building the CRLP filters. All values should always come directly from the custom settings object, even if the field in the custom settings is null
* In Skew Mode jobs, significantly increase the MAX_BATCHES_TO_HOLD_BEFORE_COMMIT value to avoid unnecessary commits to the parent object.
* Create separate Filter Groups for HardCredit and SoftCredit rather than use the same group for both types of rollups (even though that is technically supported)

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


njjc:
woo

PR #3209, opened 2018-05-25
Rollups: Create default crlp settings during install to avoid an insert error when first batch job runs after upgrade

force2b:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
This has to be tested in a beta org, so merging because it's not testable right now.

PR #3211, opened 2018-05-25
Skip any invalid UDR on Rollup conversion.

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_DefaultConfigBuilder_SVC.cls
+++ b/src/classes/CRLP_DefaultConfigBuilder_SVC.cls
@@ -735,6 +735,8 @@ public class CRLP_DefaultConfigBuilder_SVC {
             try {
                 if(udr.npo02__Object_Name__c != 'npo02__Household__c') {
                     String udrSourceFieldType = UTIL_Describe.getFieldType('Opportunity', udr.npo02__Source_Field__c.toLowerCase());
+                    String udrTargetFieldType = UTIL_Describe.getFieldType(udr.npo02__Object_Name__c, udr.npo02__Target_Field__c.toLowerCase());
+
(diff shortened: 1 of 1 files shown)


src/classes/CRLP_DefaultConfigBuilder_SVC.cls
```diff
@@ -732,40 +732,45 @@ public class CRLP_DefaultConfigBuilder_SVC {
...
-                rollup.dateField = (sourceFieldIsDate ? udr.npo02__Source_Field__c : 'CloseDate');
-                rollup.operation = operation;
-                rollup.useFiscalYear = CRLP_DefaultConfigBuilder.legacySettings.npo02__Use_Fiscal_Year_for_Rollups__c;
-                rollup.isActive = true;
-                rollup.isDeleted = false;
-                rollups.add(rollup);
+            try {
+                if(udr.npo02__Object_Name__c != 'npo02__Household__c') {
+                    String udrSourceFieldType = UTIL_Describe.getFieldType('Opportunity', udr.npo02__Source_Field__c.toLowerCase());
+                    Boolean sourceFieldIsAmount = (udrSourceFieldType == 'CURRENCY'
```
force2b: The above only checks the field type on the source field, and it will throw an exception if the field is bad. Unfortunately I missed that this only checks the source field. The target field is left unchecked. To fix this we should insert one line here to get the field type for the target field so that an exception is thrown if the field is invalid.
`String udTargetFieldType = UTIL_Describe.getFieldType(udr.npo02__Object_Name__c, udr.npo02__Target_Field__c);`

PR #3214, opened 2018-05-25
Remove or change Account and Contact filters and sorting

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CAO_Constants.cls
+++ b/src/classes/CAO_Constants.cls
@@ -145,11 +145,6 @@ public with sharing class CAO_Constants {
     public static Id individualAccountId;
 
     /*******************************************************************************************************
-    * @description The name of the Individual Bucket Account
-    */ 
-    private static String individualAccountName = CAO_Constants.BUCKET_ACCOUNT_NAME;
-
-    /*******************************************************************************************************
     * @description returns the Salesforce Id of the Individual Bucket Account
     * @return  Id
     */ 
@@ -186,7 +181,6 @@ public with sharing class CAO_Constants {
         insert individualAccountForTests;
 
         CAO_Constants.BUCKET_ACCOUNT_NAME = accountName;
-        individualAccountName = accountName;
 
         return individualAccountForTests.Id;
     }
--- a/src/classes/CAO_Constants.cls
+++ b/src/classes/CAO_Constants.cls
@@ -142,32 +142,32 @@ public with sharing class CAO_Constants {
     /*******************************************************************************************************
     * @description The Salesforce Id of the Individual Bucket Account
     */ 
-    public static Id IndividualAccountId;
+    public static Id individualAccountId;
 
     /*******************************************************************************************************
     * @description The name of the Individual Bucket Account
     */ 
-    private static String IndividualAccountName = CAO_Constants.BUCKET_ACCOUNT_NAME;
+    private static String individualAccountName = CAO_Constants.BUCKET_ACCOUNT_NAME;
 
     /*******************************************************************************************************
     * @description returns the Salesforce Id of the Individual Bucket Account
     * @return  Id
     */ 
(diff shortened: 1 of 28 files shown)


src/classes/UTIL_Typeahead_TEST.cls
```diff
@@ -49,11 +49,11 @@ public with sharing class UTIL_Typeahead_TEST {
 	      Test.setFixedSearchResults(fixedSearchResults);
 	
 	    Test.startTest();
-	    list<sObject> accts = UTIL_Typeahead.searchRecords( 'test', 'Account', null, null, null, null, null );
+	    list<sObject> accts = UTIL_Typeahead.searchRecords( 'test', 'Account', null, null, null, 'Id', null );
```
force2b: Is there a change needed in the `UTIL_Typeahead.component` file as well? I noticed that this method is called from that component passing in a value for the sort parameter, which could be null (or `Name` - the default value in the component).

src/classes/CAO_Constants.cls
```diff
@@ -142,32 +142,32 @@ public with sharing class CAO_Constants {
...
-            if(acct.size()>0){
-                if(acct[0].npe01__SYSTEMISINDIVIDUAL__c==false || acct[0].npe01__SYSTEM_AccountType__c<>CAO_Constants.BUCKET_ORGANIZATION_TYPE){
-                    acct[0].npe01__SYSTEMISINDIVIDUAL__c = true;
-                    acct[0].npe01__SYSTEM_AccountType__c = CAO_Constants.BUCKET_ORGANIZATION_TYPE;
-                    update acct[0];
-                }
-                IndividualAccountId = acct[0].id;
+        if (individualAccountId == null) {
+            List<Account> acct = [
+                SELECT Id, npe01__SYSTEMISINDIVIDUAL__c, npe01__SYSTEM_AccountType__c
```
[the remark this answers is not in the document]
njjc: I don't believe there are any cases where orgs have Individual accounts accessible by Name only, I'm totally fine with this solution. (The code that does this check is over 7 years old)[https://github.com/SalesforceFoundation/Contacts_and_Organizations/blame/master/src/classes/Constants.cls#L165], and its reason for existing is lost in the ether... my guess is that we introduced those SYSTEM fields and used that method to populate them.

src/classes/CON_ContactMerge.cls
```diff
@@ -126,15 +126,22 @@ public class CON_ContactMerge {
         
         //get loser contact Accounts that should be merged out.  This includes
         // One2One accounts and Solo Household Accounts
-        loserAccountsToBeMerged = [Select Id,npe01__SYSTEMIsIndividual__c,npe01__SYSTEM_AccountType__c,Name from Account a
-            where Id IN :loserAccountIds and 
-            (npe01__SYSTEM_AccountType__c=:CAO_Constants.ONE_TO_ONE_ORGANIZATION_TYPE or
-                (npe01__SYSTEM_AccountType__c=:CAO_Constants.HH_ACCOUNT_TYPE and Number_of_Household_Members__c = 1))
-            and Name<>:CAO_Constants.BUCKET_ACCOUNT_NAME];
```
RadSFDeveloper: The check for `and Name<>:CAO_Constants.BUCKET_ACCOUNT_NAME` has been removed here. Can there be an Account with `ONE_TO_ONE_ORGANIZATION_TYPE` or `HH_ACCOUNT_TYPE` but with `CAO_Constants.BUCKET_ACCOUNT_NAME` Name?
@njjc , @force2b

jjbennett: The CAO constants attach to bucket account will only attach to an account that meets the system account type criteria so even if we have an Account with the same name it's not going to be used as a bucket account in our code.

PR #3220, opened 2018-05-29
Account Soft Credits: Object and Rollups integration

bethbrains:
# Critical Changes

## Account Soft Credits

***IMPORTANT***: To get the most value out of Account Soft Credits, you'll want to enable Customizable Rollups so you can roll up the soft credit values to the Account records. At this time, you can't roll up Account soft credits with legacy NPSP rollups or User Defined Rollups.

Before you can start using Account soft credits, you need to give users access to the Account Soft Credit object and fields, and add the Account Soft Credit related list to your page layout. You can also optionally add more Account soft credit roles. For instructions, see the Account Soft Credit Setup section of [Soft Credit and Matching Gift Setup](https://powerofus.force.com/articles/Resource/NPSP-Soft-Credit-and-Matching-Gift-Setup).

# Changes
## Account Soft Credits
Account soft credits are credits for donations that Accounts don't actually make, but may have somehow influenced. They're very useful for situations where an Account, such as a Donor Advised Fund, uses another organization, such as a bank, to manage their money. When a donation comes from the Donor Advised Fund, the check is actually generated by the bank, so the bank gets the hard credit for the donation. Using Account soft credits, you can ensure that the Donor Advised Fund gets a soft credit for their donation.

For details on Account Soft Credits, see the [Soft Credits Overview](https://powerofus.force.com/articles/Resource/NPSP-Soft-Credits-Overview). For details on creating Account Soft Credits, see [Create Soft Credits](https://powerofus.force.com/articles/Resource/NPSP-Create-Soft-Credits).

# Issues Closed

# New Metadata

- Account_Soft_Credit__c object
- CRLP_AccountSkew_AccSoftCredit_BATCH
- CRLP_Account_AccSoftCredit_BATCH
- CRLP_RollupAccSoftCredit_SVC
- TDTM_AccountSoftCredit
- CRLP_RollupAccSoftCredit_TEST

# Deleted Metadata


src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -191,29 +196,41 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
                 if (!whereClauses.isEmpty()) {
                     query += ' WHERE ' + String.join(whereClauses, ' AND ');
                 }
 
                 // Note that the ORDER BY here doesn't gaurantee that the records will actually be in this order
-                query += ' ORDER BY ' + this.keyField + ', Opportunity.CreatedDate ASC ';
+                query += ' ORDER BY ' + this.keyField;
 
-                UTIL_Debug.debug(LoggingLevel.WARN, '>> OCR Query =' + query);
+                if (jobType != CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
```
force2b: Why not sort by Opp.CreatedDate when rolling up AccountSoftCredits?

src/classes/CRLP_Query_SEL.cls
```diff
@@ -56,6 +56,11 @@
             allFieldsToQuery.add(SObjectType.Partial_Soft_Credit__c.fields.Contact__c.getSobjectField().getDescribe());
             allFieldsToQuery.add(SObjectType.Partial_Soft_Credit__c.fields.Amount__c.getSobjectField().getDescribe());
             allFieldsToQuery.add(SObjectType.Partial_Soft_Credit__c.fields.Role_Name__c.getSobjectField().getDescribe());
+        } else if (resultObject == Account_Soft_Credit__c.SObjectType) {
+            allFieldsToQuery.add(SObjectType.Account_Soft_Credit__c.fields.Opportunity__c.getSobjectField().getDescribe());
+            allFieldsToQuery.add(SObjectType.Account_Soft_Credit__c.fields.Account__c.getSobjectField().getDescribe());
+            allFieldsToQuery.add(SObjectType.Account_Soft_Credit__c.fields.Amount__c.getSobjectField().getDescribe());
```
force2b: Is there a need to force a query of *all* the fields on the object? Generally the ones above are for fields that might be used for special conditions within the code, but where the fields themselves might not be referenced by a rollup or filter rule. In the cases of these 4 fields, I would expect that fields like Amount and Role would always be referenced by a filter rule or rollup. Opportunity and Account though would be needed since those will likely not be referenced by a rule or rollup.

src/classes/CRLP_RollupQueueable.cls
```diff
@@ -92,20 +92,31 @@ public class CRLP_RollupQueueable implements System.Queueable {
...
+            }
+            if (CRLP_Rollup_SVC.hasActivePartialSoftCreditRollupsForSObject(objType)) {
+                jobTypesToRun.add(CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit);
+            }
+            //options.rollupJobType = CRLP_RollupProcessingOptions.RollupType.AccountHardCredit;
         } else if (objType == Contact.SObjectType) {
             options.parentKeyField = SObjectType.Opportunity.fields.Primary_Contact__c.getName();
-            options.rollupJobType = CRLP_RollupProcessingOptions.RollupType.ContactHardCredit;
+            jobTypesToRun.add(CRLP_RollupProcessingOptions.RollupType.ContactHardCredit);
+            if (CRLP_Rollup_SVC.hasActivePartialSoftCreditRollupsForSObject(objType)) {
```
force2b: We previously agreed not to roll up soft credits on the Contact via a trigger. I did some experiments with this early on and discovered that you cannot run the hard credit and soft credit rollups at the same time - which is what this code is trying to do.  We pulled out the original code I wrote to do this once I put this into action in one of the larger orgs. You'll need to pull out this code from the AccountSoftCredit logic as well since it'll have exactly the same conflict with AccountHardCredit.

src/classes/CRLP_Rollup_TDTM.cls
```diff
@@ -161,9 +161,23 @@ public class CRLP_Rollup_TDTM extends TDTM_Runnable {
             }
             opps = [SELECT Id, AccountId, Primary_Contact__c, npe03__Recurring_Donation__c FROM Opportunity WHERE Id IN :oppIds];
 
+        } else if (detailObjType == Account_Soft_Credit__c.SObjectType) {
```
force2b: As I noted in the Queueable class, we should not attempt to rollup soft credits via a trigger because it will not work with a simultaneous hard credit rollup. The only alternative would be to find a way to chain the rollups, but NPSP has always stated that Soft Credit rollups are only calculated nightly. I don't think it's worth trying to crack that nut in this story. It's a much larger project of its own, if it's even possible.

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -333,6 +333,11 @@ public without sharing class TDTM_DefaultConfig {
                 Class__c = 'CRLP_Rollup_TDTM', Load_Order__c = 2, Object__c = 'npe01__OppPayment__c',
                 Trigger_Action__c = 'AfterInsert;AfterUpdate;AfterDelete;AfterUndelete'));
 
+        // Customizable Rollups on Account Soft Credit
```
force2b: I was going to say that the trigger needs to be removed, however that's not 100% correct. The trigger should *not* attempt to roll up Account Soft Credit data as previously noted, but what we do want to do is have logic to ensure that the CurrencyIsoCode on the AccountSoftCredit object is inherited from the parent Opportunity object. You can find some similar logic on the Payment, Allocation and hopefully PartialSoftCredit objects.

src/objects/Account_Soft_Credit__c.object
```diff
@@ -0,0 +1,126 @@
...
+        <fullName>Account__c</fullName>
+        <externalId>false</externalId>
+        <label>Account</label>
+        <referenceTo>Account</referenceTo>
+        <relationshipLabel>Account Soft Credits</relationshipLabel>
+        <relationshipName>Account_Soft_Credits</relationshipName>
+        <relationshipOrder>1</relationshipOrder>
+        <reparentableMasterDetail>false</reparentableMasterDetail>
+        <trackTrending>false</trackTrending>
+        <type>MasterDetail</type>
```
force2b: Should this be an MD field? The issue we have with Partial Soft Credit is the Contact lookup is an MD field, and that has limited what we can do with the object. (@njjc  what do you think?)

src/objects/Account_Soft_Credit__c.object
```diff
@@ -0,0 +1,126 @@
...
+        <label>Role</label>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Picklist</type>
+        <valueSet>
+            <restricted>true</restricted>
+            <valueSetDefinition>
+                <sorted>false</sorted>
+                <value>
+                    <fullName>Influence</fullName>
```
force2b: Influence or Influencer?

src/objects/Account_Soft_Credit__c.object
```diff
@@ -0,0 +1,126 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: I believe we'll need to add all the translation files for this object as well

src/classes/TDTM_DefaultConfig.cls
bethbrains: It's relatively trivial to add this to the existing PSC_Opportunity_TDTM class. Is having it buried in there to share code acceptable, or should this have its own namespace and class?

src/classes/CRLP_Query_SEL.cls
bethbrains: I was following the pattern from PSC. Is there a reason it needs Amount and Role always queried that doesn't apply to AccSC? Or should we remove Amount and Role from both objects?

src/classes/CRLP_Batch_Base_NonSkew.cls
bethbrains: Because the Oppty relationship is different so it was breaking the query. Fixed to make the string added conditional and use `Opportunity.` or `Opportunity__r.` as needed.

src/objects/Account_Soft_Credit__c.object
force2b: @bethbrains Just checking back on these couple of comments from May.

src/classes/CRLP_Query_SEL.cls
force2b: @bethbrains  Sorry. I never responded to your question here. Amount and Role on the PSC object are manually added because I believe they can be referenced in one of the query builder utility methods or loops even if the fields are technically included in an active rollups. Basically these two fields really should be there for any soft credit rollup. 

I can see your point of force adding them to the AccountSoftCredit query. It's roughly the same concept, so let's leave them as it. The logic won't hurt anything even if those fields are included dynamically as well.

src/classes/CRLP_Batch_Base_NonSkew.cls
force2b: @bethbrains I see your note about fixing to use a different sort key, but the code above showing using `this.keyField` as the sort. Will that variable contain either `Opportunity.CreatedDate` or `Opportunity__r.CreatedDate`? Also, it's missing the `ASC' sort sequence.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -48,7 +48,10 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
     protected List<String> fieldsToCheckForNonZero {
         get {
             if (this.jobType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit ||
-                    this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit) {
+                    this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit ||
+                    this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
```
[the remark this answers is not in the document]
force2b: @bethbrains I think this comment still stands. It should not do this check for the SoftCreditRollup types.

src/classes/CRLP_RollupQueueable.cls
force2b: @bethbrains  The trigger rollup logic still needs to be pulled out.

force2b:
Minor bug:

Create a new Account Soft Credit Rollup, selecting a Filter Group:
image

Save it. The view version shows no Filter Group:
image

src/classes/CRLP_AccountSkew_AccSoftCredit_BATCH.cls
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public CRLP_AccountSkew_AccSoftCredit_BATCH(List<Id> recordIds, Id lastIdProcessedForChunking) {
+        super(CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit, recordIds, lastIdProcessedForChunking);
+    }
+
+    /**
+     * @description Scheduler execute method to support scheduling this job directly from the UI
+     */
+    public void execute(SchedulableContext context) {
+        Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
```
force2b: The two execute methods in the BATCH classes need the same logic recently added to the other BATCH classes.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -185,34 +188,55 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+                    detailRecords = Database.query(query);
+                }
                 UTIL_Debug.debug(LoggingLevel.WARN, '>> PSC Detail Record Size =' + detailRecords.size());
 
             }
 
-            String parentObjectField, parentKeyField = options.parentKeyField;
+            String parentObjectField;
+            String parentKeyField = options.parentKeyField;
+            System.debug('parentkeyfield: '+parentKeyField);
```
force2b: Can you remove the system.debug() statement?

src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -249,8 +256,21 @@ private class CRLP_RollupAccount_TEST {
 
         insert opps;
 
+        // create 2 $10 Account Soft Credits
+        Integer scAmount = 10;
+        Integer ascCount = 2;
+        List<Account_Soft_Credit__c> accSCs = new List<Account_Soft_Credit__c>();
+        for (Integer i=0; i<ascCount; i++) {
+            Account_Soft_Credit__c accSC = new Account_Soft_Credit__c (
```
force2b: Is it worth setting a Role value and validating the filter rule against the Role?

src/classes/CRLP_Rollup_SEL.cls
```diff
@@ -306,6 +308,10 @@ public class CRLP_Rollup_SEL {
                         rollupType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit) {
                     includeRollupDefinition = ((r.Detail_Object__r.QualifiedApiName == oppObjectName || r.Detail_Object__r.QualifiedApiName == pmtObjectName)
                             && r.Amount_Object__r.QualifiedApiName != pscObjectName);
+                } else if (rollupType == CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
+                    includeRollupDefinition = ((r.Detail_Object__r.QualifiedApiName == oppObjectName || r.Detail_Object__r.QualifiedApiName == accSCObjectName)
+                            && r.Amount_Object__r.QualifiedApiName == accSCObjectName);
+                    System.debug('wanna hit here! '+includeRollupDefinition);
```
force2b: Another system.debug()

src/classes/CRLP_VRollupHandler.cls
```diff
@@ -53,6 +53,8 @@ public virtual class CRLP_VRollupHandler implements CRLP_Rollup_SVC.IRollupHandl
     protected final String allocObjectName = UTIL_Describe.getObjectDescribe(UTIL_Namespace.StrAllNSPrefix('Allocation__c')).getName();
     /* @description 'Name' of the object referenced in the rollup. Visible to classes that extend this virtual class. */
     protected final String rdObjectName = UTIL_Describe.getObjectDescribe('npe03__Recurring_Donation__c').getName();
+    /* @description 'Name' of the object referenced in the rollup. Visible to classes that extend this virtual class. */
+    protected final String ascObjectName = UTIL_Describe.getObjectDescribe(UTIL_Namespace.StrAllNSPrefix('Account_Soft_Credit__c')).getName();
```
force2b: Not sure this is used anywhere

src/classes/PSC_Opportunity_TDTM.cls
```diff
@@ -63,10 +65,10 @@ public with sharing class PSC_Opportunity_TDTM extends TDTM_Runnable {
     * @description For any given list of opportunities, makes child PSCs match the opportunity's currency.
     * @param listNew The list of opportunities from trigger new. 
     * @param listOld The list of opportunities from trigger old. 
-    * @return list<Partial_Soft_Credit__c> Partial soft credits with modified currency that need updating via DML.
+    * @return list<sObject> Partial Soft Credits or Account Soft Credits with modified currency that need updating via DML.
     ********************************************************************************************************/
-    public static list<Partial_Soft_Credit__c> updatePartialSoftCreditCurrencyFromParent(list<sObject> newlist, list<sObject> oldlist) {
-        list<Partial_Soft_Credit__c> results = new list<Partial_Soft_Credit__c>();
+    public static list<sObject> updateSoftCreditCurrencyFromParent(list<sObject> newlist, list<sObject> oldlist, SObjectType objType) {
+        list<sObject> results = new list<sObject>();
```
force2b: Should be `SObject`

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -185,34 +188,55 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+                String createdDateClause;
+                if (jobType != CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
+                    createdDateClause = ', Opportunity.CreatedDate ASC ';
+                } else {
+                    createdDateClause = ', Opportunity__r.CreatedDate ASC ';
+                }
+                query += createdDateClause;
 
+                UTIL_Debug.debug(LoggingLevel.WARN, '>> OCR Query =' + query);
+                System.debug('PROBLEM QUERY: '+query);
```
force2b: Another system.debug()

scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
```diff
@@ -304,17 +330,17 @@ private static Boolean ALLOC_INCLUDES_MEMBERSHIP = true;
      */
     private static void assertValue(Decimal expected, Decimal actual, String message) {
         if (expected != (actual == null ? 0 : actual)) {
-            errorsList.add('VALIDATION: ' + message + ' ' + expected + ', but acual is ' + actual);
+            errorsList.add('VALIDATION: ' + message + ' ' + expected + ', but actual is ' + actual);
```
force2b: omg. can't believe I never noticed that!

scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
```diff
@@ -519,6 +557,33 @@ private static Boolean ALLOC_INCLUDES_MEMBERSHIP = true;
...
+        for (Integer i=0; i<opps.size(); i++) {
+            Opportunity opp = opps[i];
+            if (opp.AccountId != orgAcct.Id
+                    && opp.RecordTypeId != membershipRTId
+                    && opp.StageName == closedWonStage) {
+                Account_Soft_Credit__c accSC = new Account_Soft_Credit__c(
+                        Account__c = accForSoftCredits.Id,
+                        Opportunity__c = opp.Id,
+                        Amount__c = ACCOUNT_SOFT_CREDIT_AMOUNT,
+                        Role__c = 'Influencer'
```
force2b: @bethbrains  Might be best to put this in a constant at the top of the class. Also, is there a Filter Rule that includes or excludes this specific Role?

src/classes/PSC_Opportunity_TDTM.cls
```diff
@@ -63,29 +65,39 @@ public with sharing class PSC_Opportunity_TDTM extends TDTM_Runnable {
...
+                results.add(psc);
+            }
+        } else if (objType == Account_Soft_Credit__c.SObjectType) {
+            //query all AccSCs associated with these opps, update their ISO code
+            for (Account_Soft_Credit__c accountSC : database.query('SELECT Id, Opportunity__c, CurrencyIsoCode FROM Account_Soft_Credit__c WHERE Opportunity__c IN :oppIds')) {
+                string oppCurrency = mapOppIdToISO.get(accountSC.Opportunity__c);
 
-            psc.put('CurrencyIsoCode', oppCurrency);
-            results.add(psc);
+                accountSC.put('CurrencyIsoCode', oppCurrency);
```
force2b: Seems that the code that updates the AccountSoftCredit and PartialSoftCredit Currency code is not actually working correctly.

In the example below, I started with a $151 USD Opportunity that had one AccountSoftCredit for $151, one PartialSoftCredit for $1 and one Payment for $151. I then changed the Opportunity.Currency to CAD (where the CAD exchange rate is 1.33). When saved, the Opportunity changes to CAD$151 (USD $113.53). The payment update does the same thing (i.e., the two amounts match). That is not the case for the AccountSoftCredit and PartialSoftCredit objects though. In these cases, the amounts changed to CAD$200.83 and CAD$1.33 respectively. See the `PMT_Payment_TDTM.setPaymentCurrencyFromOpportunity` method for an example of how this is handled to ensure the amount is correct. Basically, you have to update both the CurrencyIsoCode *and* the Amount fields. 

image

scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
bethbrains: Done, and there is now!

src/classes/OPP_AccountSoftCredit_TDTM.cls
```diff
@@ -0,0 +1,113 @@
...
+    * @return void.
+    ********************************************************************************************************/
+    private void blockManualCurrencyChange(list<Account_Soft_Credit__c> listAccSCNew, list<Account_Soft_Credit__c> listAccSCOld) {
+
+        for (integer i = 0; i < listAccSCNew.size(); i++) {
+            Account_Soft_Credit__c accSCNew = listAccSCNew[i];
+            Account_Soft_Credit__c accSCOld = listAccSCOld[i];
+
+            // don't allow manually changing the AccSC currency: it must be changed through the parent opportunity
+            if (userInfo.isMultiCurrencyOrganization() && accSCNew.get('CurrencyIsoCode') != accSCOld.get('CurrencyIsoCode') && blockCurrencyChange) {
```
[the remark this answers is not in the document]
bethbrains: I'd like to understand the rationale here a bit better at some point. Thanks.

src/classes/CRLP_AccountSkew_AccSoftCredit_BATCH.cls
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public CRLP_AccountSkew_AccSoftCredit_BATCH(List<Id> recordIds, Id lastIdProcessedForChunking) {
+        super(CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit, recordIds, lastIdProcessedForChunking);
+    }
+
+    /**
+     * @description Scheduler execute method to support scheduling this job directly from the UI
+     */
+    public void execute(SchedulableContext context) {
+        String className = CRLP_AccountSkew_AccSoftCredit_BATCH.class.getName();
```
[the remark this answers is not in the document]
force2b: @bethbrains Just making sure you saw this above comment.

src/classes/CRLP_RollupAccSoftCredit_TEST.cls
```diff
@@ -0,0 +1,7 @@
+/**
+ * Created by bbreisnes on 2018-07-31.
+ */
+
+public with sharing class CRLP_RollupAccSoftCredit_TEST {
```
force2b: This should be `private' without a sharing declaration to match the format of all the other CRLP unit tests. Generally there isn't a reason for unit tests to be public, and "with sharing" doesn't necessarily specify anything here because the methods cannot be called by other classes that might use "without sharing".

src/classes/PSC_Opportunity_TDTM.cls
```diff
@@ -84,18 +84,20 @@ public with sharing class PSC_Opportunity_TDTM extends TDTM_Runnable {
 
         if (objType == Partial_Soft_Credit__c.SObjectType) {
             //query all PSCs associated with these opps, update their ISO code
-            for (Partial_Soft_Credit__c psc : database.query('SELECT Id, Opportunity__c, CurrencyIsoCode FROM Partial_Soft_Credit__c WHERE Opportunity__c IN :oppIds')) {
-                string oppCurrency = mapOppIdToISO.get(psc.Opportunity__c);
+            for (Partial_Soft_Credit__c psc : Database.query('SELECT Id, Amount__c, Opportunity__c, CurrencyIsoCode FROM Partial_Soft_Credit__c WHERE Opportunity__c IN :oppIds')) {
+                String oppCurrency = mapOppIdToISO.get(psc.Opportunity__c);
 
                 psc.put('CurrencyIsoCode', oppCurrency);
+                psc.put('Amount__c', psc.Amount__c);
```
force2b: I'm always a bit wary when it comes to using non-namespaced field names in a string. It **may** work in a namespaced org, but it may fail. Given that we know the `Amount__c` field exists here (and below), it might be safer to just use `psc.Amount__c = psc.Amount__c;` (though we'll want to test that it still works as expected). The alternative is to call the UTIL_Namespace class to apply the namespace to the field.

src/labels/CustomLabels.labels
```diff
@@ -4634,7 +4634,7 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <language>en_US</language>
         <protected>false</protected>
         <shortDescription>pscManageSoftCreditsCantChangeCurrency</shortDescription>
-        <value>You can't modify Partial Soft Credit currencies directly. Update the parent Opportunity's currency, and NPSP will automatically update the curency of all related Partial Soft Credit records.</value>
+        <value>You can't modify soft credit currencies directly. Update the parent Opportunity's currency, and NPSP will automatically update the currency of all related soft credits.</value>
```
force2b: The single apostrophes should be escaped.

src/classes/CRLP_VRollupHandler.cls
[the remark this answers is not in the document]
force2b: @bethbrains  Good point. I added them early on thinking they'd be used. Feel free to remove any of them ones that are not referenced in this class or the classes that implement this class.

src/classes/CRLP_RollupAccSoftCredit_TEST.cls
bethbrains: yep sorry, this was just a placeholder!! it'll go in correctly on tuesday.

src/classes/CRLP_AccountSkew_AccSoftCredit_BATCH.cls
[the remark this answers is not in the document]
bethbrains: done

bethbrains:
@force2b all addressed and ready for final (!) review.

src/classes/CRLP_AccountSkew_AccSoftCredit_BATCH.cls
```diff
@@ -0,0 +1,73 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2018
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
```
force2b: Can you update the Rollups2.htm file with the new classes?

src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -217,6 +221,8 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
                 parentKeyField = 'ContactId';
             } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
                 parentKeyField = 'Contact.AccountId';
+            } else if (this.jobType == CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
+                parentKeyField = 'Account__c';
```
force2b: Not sure this will work in a namespaced context. The rest of the `parentKeyField` values are standard fields. This is the only one referencing a custom field. I feel like it should have the namespace applied, though I'm always a bit unsure of when the namespace is absolutely required for these kind of references. Probably worth testing in a namespaced scratch org to be sure it's working as expected.

PR #3222, opened 2018-05-30
Rollups: Prevent duplicate batches running

bethbrains:
# Critical Changes

# Changes

- Before a batch kicks off, it checks to see if there are other instances of itself still in process, and only proceeds if it finds none.
- If a job is skipped because the previous day's job is still running, an Error is logged with an explanatory message.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_AccountSkew_BATCH.cls
+++ b/src/classes/CRLP_AccountSkew_BATCH.cls
@@ -62,7 +62,9 @@ public class CRLP_AccountSkew_BATCH extends CRLP_Batch_Base_Skew implements Data
      */
     public void execute(SchedulableContext context) {
         if (CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled) {
-            Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
+            if (!UTIL_MasterSchedulableHelper.hasRunningJob('CRLP_AccountSkew_BATCH')) {
+                Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
+            }
         }
     }
 }
--- a/src/classes/CRLP_AccountSkew_BATCH.cls
+++ b/src/classes/CRLP_AccountSkew_BATCH.cls
@@ -66,9 +66,7 @@ public class CRLP_AccountSkew_BATCH extends CRLP_Batch_Base_Skew implements Data
             if (!UTIL_MasterSchedulableHelper.hasRunningJob(className)) {
                 Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
             } else {
-                BatchException ex = new BatchException();
-                ex.setMessage(String.format(Label.CRLP_BatchSkipped, new List<String>{className}));
-                ERR_Handler.processError(ex, ERR_Handler_API.Context.CRLP);
+                logDuplicateBatchError(className);
             }
         }
     }
--- a/src/classes/CRLP_AccountSkew_BATCH.cls
+++ b/src/classes/CRLP_AccountSkew_BATCH.cls
@@ -61,9 +61,14 @@ public class CRLP_AccountSkew_BATCH extends CRLP_Batch_Base_Skew implements Data
      * @description Scheduler execute method to support scheduling this job directly from the UI
      */
     public void execute(SchedulableContext context) {
+        String className = 'CRLP_AccountSkew_BATCH';
         if (CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled) {
-            if (!UTIL_MasterSchedulableHelper.hasRunningJob('CRLP_AccountSkew_BATCH')) {
+            if (!UTIL_MasterSchedulableHelper.hasRunningJob(className)) {
                 Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
+            } else {
+                Exception ex;
+                ex.setMessage(String.format(Label.CRLP_BatchSkipped, new List<String>{className}));
+                ERR_Handler.processError(ex, ERR_Handler_API.Context.CRLP);
             }
         }
     }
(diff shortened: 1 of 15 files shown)


src/classes/CRLP_AccountSkew_BATCH.cls
```diff
@@ -61,6 +61,15 @@ public class CRLP_AccountSkew_BATCH extends CRLP_Batch_Base_Skew implements Data
      * @description Scheduler execute method to support scheduling this job directly from the UI
      */
     public void execute(SchedulableContext context) {
-        Database.executeBatch(new CRLP_SkewDispatcher_BATCH(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
+        String className = 'CRLP_AccountSkew_BATCH';
```
force2b: Classes are also namespaced, so this would check will fail in a namespaced org. The way to work around this is to use the following syntax to get the class name. This will return the full class name, including the namespace if there is one.
`String className = CRLP_AccountSkew_BATCH.class.name();`

src/classes/CRLP_GAU_BATCH.cls
```diff
@@ -61,7 +61,11 @@ public class CRLP_GAU_BATCH extends CRLP_Batch_Base_Skew implements Database.Bat
      * @description Scheduler execute method to support scheduling this job directly from the UI
      */
     public void execute(SchedulableContext context) {
-        CRLP_RollupBatch_SVC.executeBatchRollupJob(CRLP_RollupProcessingOptions.RollupType.GAU,
-                CRLP_RollupProcessingOptions.BatchJobMode.SkewMode);
+        if (CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled) {
```
force2b: Is there a reason this class doesn't follow the same patter as the other classes - where the classname var is set and an exception is logged if the job is already running?

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -185,4 +185,22 @@ public without sharing class UTIL_MasterSchedulableHelper {
...
         }
     }
+
+    /*******************************************************************************************************
+    * @description Checks for presently running jobs of a particular method
+    * @param methodName The name of the method to check for running jobs on
+    * @return Boolean
+    */
+    public static Boolean hasRunningJob(String className) {
+        List<AsyncApexJob> runningJobs = [SELECT Id, ApexClass.Name
```
force2b: Could this just be something like:
```Integer cnt = [SELECT Count() FROM AsyncApexJob WHERE ApexClass.Name = :className AND Status NOT IN ('Queued','Aborted','Completed','Failed'];```

File: src/labels/CustomLabels.labels [hunk 4]
```diff
@@ -723,6 +723,14 @@
         <shortDescription>Context for when to add a date, amount, or detail field to a Rollup__mdt record.</shortDescription>
         <value>These fields determine what information is being summarized and provide the normal defaults for Donations. You can edit them as needed.</value>
     </labels>
+    <labels>
+        <fullName>CRLP_BatchSkipped</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error logged when a scheduled batch is skipped b/c yesterday's still running.</shortDescription>
```
force2b: I think the the single quote needs to be escaped. If you refresh the label metadata file from your org it should pull down the properly formatted text for this label.

src/classes/CRLP_ContactSkew_BATCH.cls
```diff
@@ -61,9 +61,16 @@ public class CRLP_ContactSkew_BATCH extends CRLP_Batch_Base_Skew implements Data
      * @description Scheduler execute method to support scheduling this job directly from the UI
      */
     public void execute(SchedulableContext context) {
+        String className = 'CRLP_ContactSkew_BATCH';
```
force2b: Getting the class name should be consistent throughout all the classes that use this logic. There are some other classes that are not using the `.class.getName()` syntax.

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -185,4 +185,22 @@ public without sharing class UTIL_MasterSchedulableHelper {
...
+
+    /*******************************************************************************************************
+    * @description Checks for presently running jobs of a particular method
+    * @param methodName The name of the method to check for running jobs on
+    * @return Boolean
+    */
+    public static Boolean hasRunningJob(String className) {
+        Integer jobCount = [SELECT COUNT()
+            FROM AsyncApexJob
+            WHERE ApexClass.Name = :className
```
force2b: I was thinking about this the other day. The `ApexClass` object has a separate field for `NamespacePrefix`, which means I wasn't quite right about the query failing in a namespaced org. The `Name` field is the name of the class without the namespace. I still prefer the syntax in the classes to get the full name of the class (ex: `CRLP_RDSkew_BATCH.class.getName()`) rather than hardcoding the classname as a string, but the method here will need a slight change to remove the namespace from the classname before the query. I suggest something like the following:
```className = (className.contains('__') ? className.split('__')[1] : className);```

Also, wouldn't hurt to test this out in a namespaced scratch org to verify that the logic works as expected with or without the namespace.

File: src/classes/CRLP_RD_BATCH.cls [hunk 7]
```diff
@@ -60,7 +60,11 @@ public class CRLP_RD_BATCH extends CRLP_Batch_Base_NonSkew implements Database.B
      * @description Scheduler execute method to support scheduling this job directly from the UI
      */
     public void execute(SchedulableContext context) {
-        CRLP_RollupBatch_SVC.executeBatchRollupJob(CRLP_RollupProcessingOptions.RollupType.RecurringDonations,
-                CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode);
+        if (CRLP_Rollup_SVC.isCustomizableRollupEngineEnabled) {
+            if (!UTIL_MasterSchedulableHelper.hasRunningJob('CRLP_RD_BATCH')) {
```
force2b: The class name is still hardcode in this one.

src/classes/UTIL_MasterSchedulableHelper.cls
```diff
@@ -185,4 +185,23 @@ public without sharing class UTIL_MasterSchedulableHelper {
...
+    * @description Checks for presently running jobs of a particular method
+    * @param methodName The name of the method to check for running jobs on
+    * @return Boolean
+    */
+    public static Boolean hasRunningJob(String className) {
+        className = (className.contains('__') ? className.split('__')[1] : className);
+        Integer jobCount = [SELECT COUNT()
+            FROM AsyncApexJob
+            WHERE ApexClass.Name = :className
+            AND Status NOT IN ('Queued','Aborted','Completed','Failed')
```
force2b: I thought about this more last night. I like this query, however there is one potential problem I see. If for some reason, a User clicks the button to manually recalculate rollups on a single Account, Contact, GAU or Recurring Donation the related batch jobs will run. And, if that button is clicked at around the same time the scheduled jobs are running, it'll block the scheduled jobs. It's a slim use case, but I don't think it's totally impossible. It could also happen in an integration if an Opp is inserted on an Account that has more than 250 attached Opps (because the Skew Mode job will be kicked off).

I think the fix here is to add a simple condition to this query for `CreatedDate < :tenMinutesAgo` with a simple calc for `DateTime tenMinutesAgo = DateTime.Now().addMinutes(-10);`. The "10" minutes part is arbitrary and I'm open to a different time frame if you think more or less time is warranted.

force2b:
@bethbrains Did you mean to merge this?

bethbrains:
No. Crap. Wanted to get your final approval on the last commit.

force2b:
Might be a bit challenging to revert, but I'm also not fully comfortable releasing this without more testing and approval. If it's possible to revert the merge and re-open this that would be the preferred solution.

bethbrains:
Reverted. I'll open a new PR.

PR #3229, opened 2018-05-31
Feature/rollups  fg ui tweaks

randi274:
# Critical Changes

# Changes
- Alphabetized detail objects
- Fixed spacing on rollup labels and corrected spacing to display error on the filter rule modal.
- Change 'Value' icon on the FG popup to only appear for value "In List" or "Not In List" when a multipicklist is not available
- Changed filter rule modal to dynamically render. Test by saving a FG modal with all incorrect values, generating errors, close, then open again to ensure there are no residual errors.

# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
@force2b this seems like a bigger list of changes than it is. The dynamic rendering fix required wrapping everything for the FG modal in an aura:if which changed a lot of indentation. I've also requested Dina to retest for the filter rule dynamic rendering since it's a larger change (W-024781).

This PR is also tied to the alphabetizing FG fields one and should follow the merge of that.

PR #3235, opened 2018-05-31
Rollups: restrict RD target fields

bethbrains:
# Critical Changes

# Changes
- Restrict target fields available from Recurring Donations to a user's own custom fields.
- Ignore Rollup__mdt records with banned target fields from the main rollup selector query

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupUI_SVC.cls
+++ b/src/classes/CRLP_RollupUI_SVC.cls
@@ -844,6 +844,7 @@ public with sharing class CRLP_RollupUI_SVC {
         List<Map<String, String>> fieldObjectList = new List<Map<String, String>>();
 
         Map<String, Schema.DescribeFieldResult> objectFields = UTIL_Describe.getAllFieldsDescribe(objectName);
+        // TODO: consider putting this into a map instead of a set
         Set<String> excludedFields = new Set<String>{'ConnectionReceivedId', 'ConnectionSentId', 'IsDeleted', 'IsSplit', 'Fiscal', 'FiscalQuarter', 'FiscalYear', 'SyncedQuoteID'};
 
         for (Schema.DescribeFieldResult dfr : objectFields.values()) {
@@ -883,12 +884,12 @@ public with sharing class CRLP_RollupUI_SVC {
             String fieldType = UTIL_Describe.getFieldType(objectName, dfr.name.toLowerCase());
             Boolean isValidType = validTypeSet.contains(fieldType);
 
-            //check that summary field isn't being used on another rollup
+            //check that target field isn't being used on another rollup
             Boolean isRolledUp = referencedFieldSet.contains(dfr);
 
-            if ((dfr.isUpdateable() && dfr.isCustom() && isValidType && !isRolledUp && !dfr.name.toLowerCase().contains('__system')
-                    && !dfr.name.contains('Latitude__s') && !dfr.name.contains('Longitude__s')
-                    && !fieldIsBanned(objectName,dfr.name))
+            if (( isValidType
+                    && !isRolledUp
+                     && isTargetFieldAllowed(objectName,dfr))
                     || selectedField.equals(dfr.name)) {
                 Map<String, String> fieldObject = new Map<String, String>();
                 fieldObject.put('name', dfr.name);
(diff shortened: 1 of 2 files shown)


src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -906,6 +907,38 @@ public with sharing class CRLP_RollupUI_SVC {
         return fieldObjectList;
     }
 
+    /*******************************************************************************************************
+    * @description Finds writable, custom, unused target fields of specific types to provide field lists for Rollup definitions.
+    *              Includes an exception for a set summary field on the rollup detail.
+    * @param objectName the name of the object to look up
+    * @param selectedField the name of a summary field already in use
+    * @return List<Map<String, String>> an object with field api names, labels and data type.
+    */
```
randi274: Documentation will need updating for fieldIsBanned

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -906,6 +907,38 @@ public with sharing class CRLP_RollupUI_SVC {
...
+    private static Boolean fieldIsBanned(String objectName, String fieldName) {
```
randi274: Is there a compelling reason for making a new method to do this, instead of doing it inline like we have with getFilterRuleFieldsExcludingTypes (eg: Set<String> excludedFields = new Set<String>{'ConnectionReceivedId', 'ConnectionSentId', 'IsDeleted', 'IsSplit', 'Fiscal', 'FiscalQuarter', 'FiscalYear', 'SyncedQuoteID'};)? You could create the map there too if the perf gains seem worth it. I wouldn't be opposed to similar mapping in getFilterRuleFieldsExcludingTypes too.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -887,7 +887,8 @@ public with sharing class CRLP_RollupUI_SVC {
             Boolean isRolledUp = referencedFieldSet.contains(dfr);
 
             if ((dfr.isUpdateable() && dfr.isCustom() && isValidType && !isRolledUp && !dfr.name.toLowerCase().contains('__system')
-                    && !dfr.name.contains('Latitude__s') && !dfr.name.contains('Longitude__s'))
+                    && !dfr.name.contains('Latitude__s') && !dfr.name.contains('Longitude__s')
+                    && !fieldIsBanned(objectName,dfr.name))
```
randi274: If it is preferable here to create a new custom method, maybe we should rename to something like 'targetFieldIsAllowed' and avoid the extra work of inversing it since it's a single-use method. 

...While we're at it, why don't we add the check for contains system, latitude and longitude into this method and get rid of a lot of these extra checks in this massive if block?? :) Otherwise let's rename the method to something more specific since technically all of those are 'banned' as well.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -844,6 +844,7 @@ public with sharing class CRLP_RollupUI_SVC {
         List<Map<String, String>> fieldObjectList = new List<Map<String, String>>();
 
         Map<String, Schema.DescribeFieldResult> objectFields = UTIL_Describe.getAllFieldsDescribe(objectName);
+        // TODO: consider putting this into a map instead of a set
```
force2b: Out of curiosity, why change to a map?

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -906,6 +908,49 @@ public with sharing class CRLP_RollupUI_SVC {
...
+                ,'npe03__Date_Established__c'
+                ,'npe03__Donor_Name__c'
+                ,'npe03__Installment_Amount__c'
+                ,'npe03__Installment_Period__c'
+                ,'npe03__Installments__c'
+                ,'npe03__Open_Ended_Status__c'
+                ,'npe03__Organization__c'
+                ,'npe03__Schedule_Type__c'
+                ,'npe03__Total__c'
+                ,'Always_Use_Last_Day_Of_Month__c'});
```
force2b: This last field is in the Cumulus repo, so it will get the `npsp__` namespace when delivered to customers. It'll need a call to the UTIL_Namespace class here to apply the namespace if required.

src/classes/CRLP_Rollup_SEL.cls
```diff
@@ -62,6 +63,16 @@ public class CRLP_Rollup_SEL {
                         WHERE Is_Deleted__c = false
                         ORDER BY Summary_Object__c, MasterLabel
                 ];
+
+                // Never return Rollups with banned Target Fields
+                // You can't create them in the CRLP UI, but they may have snuck in via native CMDT UI
+                for (Rollup__mdt rollup : stagedRollups) {
```
force2b: I understand what you're going for here, but I'm wondering if this is overkill. I feel like we're trying to prevent users from harming themselves, which to me goes outside of the scope of this project and our role overall. We're giving them a tool to use. If they purposely choose to use Salesforce Settings or a metadata deployment tool to insert CMT's that our UI wouldn't normally support that is their choice. I don't feel like we have the right to tell them they can't do it. @njjc What do you think?

src/classes/CRLP_RollupUI_SVC.cls
randi274: This was a change that Beth had made above to search specific objects for specific fields. IsDeleted I guess would need to be checked every time, but for the other fields do they match more than one object?

force2b: I see. The only ones unique to Opportunity are the last 5 fields. The first 3 can appear in all objects.

src/classes/CRLP_Rollup_SEL.cls
njjc: I agree with Mike..."we don't recommend you shoot yourself in the foot, but we're not going to provide you with bullet-proof shoes to help avoid it."

bethbrains: OK, I'll take it out. Thanks!

PR #3249, opened 2018-06-02
Fix Cross-Site Scripting

kofijohnson:
Fixed the RecordType and Reciprocal Relationship Cross-Site Scripting.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


njjc:
@kofijohnson will you be adding more xss fixes to this PR?

kofijohnson:
No @njjc! I've checked the methods in the main controller STG_PanelHealthCheck_CTRL and it does not have any XSS fix. However the related controller methods like STG_PanelADDRVerification_CTRL.verifyAddressVerificationService, STG_PanelPaymentMapping_CTRL.verifyPaymentMappings, etc. may have some XSS issue, but it will need more investigations. 
So my plan is we fix the current ticket issues first, then create a new one to do the investigation/fixes in other controllers.

PR #3253, opened 2018-06-04
CCI: Auto Create all NPSP OpportunityContactRole values in new Scratch Orgs

force2b:
Update CCI to automatically create **all** the NPSP related Opportunity Contact Role values (*as active*) on the OCR object.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cdcarter:
Unpackaged/pre is run for mpinstaller installs as well. Is that intended or should this only be for dev/test orgs?

force2b:
What does the `mpinstaller` do? Generally, I think this should only be for new scratch orgs that we create manually.

force2b:
@cdcarter To make this change only apply to scratch orgs, would the new changes need to be in the `dev_config` folder?

unpackaged/pre/opportunity_record_types/package.xml
```diff
@@ -4,5 +4,5 @@
     <members>*</members>
     <name>CustomObject</name>
   </types>
-  <version>29.0</version>
+  <version>42.0</version>
```
cdcarter: would you revert these back to 29, since we aren't changing them?

force2b: Will do

PR #3261, opened 2018-06-05
Feature/automated soft credits  ordered

joy-sfdo:
- Addressed items from the previous PR review.
- Resolved test failures.
- Updated Help Text per Work Item #W-025199.

# Critical Changes

# Changes
**Automated Soft Credits**

NPSP now supports several new automated soft credits. Previously, NPSP could automatically create soft credits for household members of a donor, for Honoree or Notification Contacts, and for donors in a Matching Gift scenario. 

In addition, NPSP now creates automated soft credits for the:

- Primary Contact on an Organizational Opportunity
- Related Contact on a Relationship record with an Individual donor
- Contact on an Affiliation record with an Organizational donor

Many organizations get by just fine with only Household soft credits. But some organizations have a donor base that requires a little more attention. Depending on the kind of organization you work for, you might have to accommodate donor advised funds, memorial gifts, planned giving, major donor portfolios, giving circles, crowdfunding, peer-to-peer fundraising, and much more. In many of these cases, you need to assign soft credits to one or more people, such as a board member, members of a family foundation, or simply relatives who don’t live in the same household. 

Getting all of this right is important! Donors with more complex needs will expect that you are representing their family, businesses, associates, and financial institutions correctly, and that your organization is crediting the right people for the donations they've helped to bring in.

NPSP now lets you configure soft credit contact roles in multiple ways to help accommodate these different scenarios. And since NPSP creates the soft credits automatically, you can scale your fundraising efforts while continuing to honor the wishes of your donors. 

**Resources**:

- [Automated Soft Credits](https://powerofus.force.com/articles/Resource/Automated-Soft-Credits)
- [NPSP Logic for Creating Contact Roles](https://powerofus.force.com/articles/Resource/Logic-for-Creating-OCRs)
- [Configure Soft Credits](https://powerofus.force.com/articles/Resource/NPSP-Configure-Soft-Credit-and-Matching-Gift-Rollup-Settings)

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/OPP_AutomatedSoftCreditsService.cls
+++ b/src/classes/OPP_AutomatedSoftCreditsService.cls
@@ -38,6 +38,10 @@
 public class OPP_AutomatedSoftCreditsService {
 
     private static final String RELATED_OPPORTUNITY_CONTACT_ROLE_NONE = '--None--';
+    private static final String OBJECT_TYPE_RELATIONSHIP = 'npe4__Relationship__c';
+    private static final String OBJECT_TYPE_AFFILIATION = 'npe5__Affiliation__c';
+    private static final String FILED_NAME_RELATED_CONTACT = 'npe4__RelatedContact__c';
+    private static final String FILED_NAME_CONTACT = 'npe5__Contact__c';
 
 
     /*********************************************************************************************************
@@ -99,7 +103,7 @@ public class OPP_AutomatedSoftCreditsService {
             }
         }
 
-        Map<Id, List<npe4__Relationship__c>> uniquePrimaryContactToRelationships = removeDuplicateRelationships(primaryContactToRelationships);
+        Map<Id, List<npe4__Relationship__c>> uniquePrimaryContactToRelationships = removeDuplicateAssociatedRecords(primaryContactToRelationships, OBJECT_TYPE_RELATIONSHIP, FILED_NAME_RELATED_CONTACT);
 
         return uniquePrimaryContactToRelationships;
     }
@@ -192,7 +196,7 @@ public class OPP_AutomatedSoftCreditsService {
             }
         }
 
-        Map<Id, List<npe5__Affiliation__c>> uniqueAccountToAffiliations = removeDuplicateAffiliations(accountToAffiliations);
+        Map<Id, List<npe5__Affiliation__c>> uniqueAccountToAffiliations = removeDuplicateAssociatedRecords(accountToAffiliations, OBJECT_TYPE_AFFILIATION, FILED_NAME_CONTACT);
 
         return uniqueAccountToAffiliations;
     }
@@ -238,56 +242,33 @@ public class OPP_AutomatedSoftCreditsService {
 
 
     /*********************************************************************************************************
-    * @description Removes duplicate Relationship records based on the Contact Id.
-    * @param primaryContactToRelationships Map of Primary Contact to its Relationship records.
-    * @return Map<Id, List<npe4__Relationship__c>> Map of unique Relationship records based on Contact Id.
+    * @description Removes duplicate Relationship/Affiliation records based on the Contact Id.
+    * @param sourceIdToAssociatedRecords Map of Primary Contact/Account to its related records.
+    * @return Map<Id, List<sObject>> Map of unique Relationship records based on Contact Id.
     **********************************************************************************************************/
     @testVisible
-    private static Map<Id, List<npe4__Relationship__c>> removeDuplicateRelationships(Map<Id, List<npe4__Relationship__c>> primaryContactToRelationships) {
-        Map<Id, List<npe4__Relationship__c>> primaryContactToUniqueRelationships = new Map<Id, List<npe4__Relationship__c>>();
+    private static Map<Id, List<sObject>> removeDuplicateAssociatedRecords(Map<Id, List<sObject>> sourceIdToAssociatedRecords, String objectType, String fieldName) {
+        Map<Id, List<sObject>> sourceIdToUniqueAssociatedRecords = new Map<Id, List<sObject>>();
 
-        for (Id primaryContactId: primaryContactToRelationships.keySet()) {
-            Map<Id, npe4__Relationship__c> cleanedRelationships = new Map<Id, npe4__Relationship__c>();
-            List<npe4__Relationship__c> uniqueRelationships = new List<npe4__Relationship__c>();
-            List<npe4__Relationship__c> relationships = primaryContactToRelationships.get(primaryContactId);
-            for (npe4__Relationship__c eachRelationship : relationships) {
-                if (!cleanedRelationships.containsKey(eachRelationship.npe4__RelatedContact__c)) {
-                    cleanedRelationships.put(eachRelationship.npe4__RelatedContact__c, eachRelationship);
-                    uniqueRelationships.add(eachRelationship);
+        for (Id sourceId : sourceIdToAssociatedRecords.keySet()) {
+            Map<Id, sObject> cleanedRelatedRecords = new Map<Id, sObject>();
+            List<sObject> uniqueAssociatedRecords = new List<sObject>();
+            List<sObject> relatedRecords = sourceIdToAssociatedRecords.get(sourceId);
+            for (sObject eachAssociatedRecord : relatedRecords) {
+                if (!cleanedRelatedRecords.containsKey((Id)eachAssociatedRecord.get(fieldName))) {
+                    cleanedRelatedRecords.put((Id)eachAssociatedRecord.get(fieldName), eachAssociatedRecord);
+                    if (OBJECT_TYPE_RELATIONSHIP.equals(objectType)) {
+                        uniqueAssociatedRecords.add((npe4__Relationship__c)eachAssociatedRecord);
+                    } else if (OBJECT_TYPE_AFFILIATION.equals(objectType)) {
+                        uniqueAssociatedRecords.add((npe5__Affiliation__c)eachAssociatedRecord);
+                    }
                 }
             }
 
-            primaryContactToUniqueRelationships.put(primaryContactId, uniqueRelationships);
+            sourceIdToUniqueAssociatedRecords.put(sourceId, uniqueAssociatedRecords);
         }
 
-        return primaryContactToUniqueRelationships;
-    }
-
-
-    /*********************************************************************************************************
-    * @description Removes duplicate Affiliation records based on the Contact Id.
-    * @param accountToAffiliations Map of Account to Affiliation records.
-    * @return Map<Id, List<npe5__Affiliation__c>> Map of unique Affiliation records based on Contact Id.
-    **********************************************************************************************************/
-    @testVisible
-    private static Map<Id, List<npe5__Affiliation__c>> removeDuplicateAffiliations(Map<Id, List<npe5__Affiliation__c>> accountToAffiliations) {
-        Map<Id, List<npe5__Affiliation__c>> accountToUniqueAffiliations = new Map<Id, List<npe5__Affiliation__c>>();
-
-        for (Id accountId: accountToAffiliations.keySet()) {
-            Map<Id, npe5__Affiliation__c> cleanedAffiliations = new Map<Id, npe5__Affiliation__c>();
-            List<npe5__Affiliation__c> uniqueAffiliations = new List<npe5__Affiliation__c>();
-            List<npe5__Affiliation__c> affiliations = accountToAffiliations.get(accountId);
-            for (npe5__Affiliation__c eachAffiliation : affiliations) {
-                if (!cleanedAffiliations.containsKey(eachAffiliation.npe5__Contact__c)) {
-                    cleanedAffiliations.put(eachAffiliation.npe5__Contact__c, eachAffiliation);
-                    uniqueAffiliations.add(eachAffiliation);
-                }
-            }
-
-            accountToUniqueAffiliations.put(accountId, uniqueAffiliations);
-        }
-
-        return accountToUniqueAffiliations;
+        return sourceIdToUniqueAssociatedRecords;
     }
 
 
--- a/src/classes/OPP_AutomatedSoftCreditsService.cls
+++ b/src/classes/OPP_AutomatedSoftCreditsService.cls
@@ -38,10 +38,10 @@
 public class OPP_AutomatedSoftCreditsService {
 
     private static final String RELATED_OPPORTUNITY_CONTACT_ROLE_NONE = '--None--';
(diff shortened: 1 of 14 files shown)


src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -38,6 +38,10 @@
 public class OPP_AutomatedSoftCreditsService {
 
     private static final String RELATED_OPPORTUNITY_CONTACT_ROLE_NONE = '--None--';
+    private static final String OBJECT_TYPE_RELATIONSHIP = 'npe4__Relationship__c';
```
force2b: I'm fine with using the strings, but another approach (and one that we used with Customizable Rollups) is to work entirely with SObjectTypes & Field Tokens:
```
private static final SObjectType OBJECT_TYPE_RELATIONSHIP = npe4__Relationship__c.SObjectType;
private static final SObjectType OBJECT_TYPE_AFFILIATION = npe5__Affiliation__c.SObjectType;
private static final SObjectField FIELD_NAME_CONTACT = npe5__Affiliation__c.npe5__Contact__c;
private static final SObjectField FILED_NAME_RELATED_CONTACT = npe4__Relationship__c. npe4__RelatedContact__c;
```

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -38,6 +38,10 @@
...
+    private static final String OBJECT_TYPE_AFFILIATION = 'npe5__Affiliation__c';
+    private static final String FILED_NAME_RELATED_CONTACT = 'npe4__RelatedContact__c';
```
force2b: **Field** is misspelled as `FILED` in these constants.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -238,56 +242,33 @@ public class OPP_AutomatedSoftCreditsService {
...
+
+        for (Id sourceId : sourceIdToAssociatedRecords.keySet()) {
+            Map<Id, sObject> cleanedRelatedRecords = new Map<Id, sObject>();
+            List<sObject> uniqueAssociatedRecords = new List<sObject>();
+            List<sObject> relatedRecords = sourceIdToAssociatedRecords.get(sourceId);
+            for (sObject eachAssociatedRecord : relatedRecords) {
+                if (!cleanedRelatedRecords.containsKey((Id)eachAssociatedRecord.get(fieldName))) {
+                    cleanedRelatedRecords.put((Id)eachAssociatedRecord.get(fieldName), eachAssociatedRecord);
+                    if (OBJECT_TYPE_RELATIONSHIP.equals(objectType)) {
+                        uniqueAssociatedRecords.add((npe4__Relationship__c)eachAssociatedRecord);
```
force2b: Since your list of `uniqueAssociatedRecords` is on `SObject`, there's actually no need to split these in in the `if` and cast as the explicit object type. It would be more concise to just add them to the collection as an Sobject: `uniqueAssociatedRecords.add(eachAssociatedRecord);`

src/labels/CustomLabels.labels
```diff
@@ -5705,7 +5705,7 @@ IMPORTANT: Both of these fields must have the same data type.</value>
         <language>en_US</language>
         <protected>false</protected>
         <shortDescription>Help text for Contact Role for Organizational Opps</shortDescription>
-        <value>The default Contact Role automatically assigned to the Opportunity's Primary Contact when the Opportunity's Account is an Organization Account.</value>
+        <value>The default Contact Role automatically assigned to the Opportunity's Primary Contact when the Opportunity's Account is NOT a Household, One-to-One, or Individual Account.</value>
```
force2b: Single quotes should be encoded in custom labels. You should be able to get the encoded version by refreshing the 'custom.label' file from within your ide.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
```diff
@@ -49,9 +49,9 @@
     </fields>
     <fields>
         <fullName>Contact_Role_for_Organizational_Opps__c</fullName>
-        <description>The default Contact Role automatically assigned to the Opportunity's Primary Contact when the Opportunity's Account is an Organization Account.</description>
+        <description>The default Contact Role automatically assigned to the Opportunity's Primary Contact when the Opportunity's Account is NOT a Household, One-to-One, or Individual Account.</description>
```
force2b: Single quotes should be encoded in this metadata file as well. You should be able to get the encoded version by refreshing the object file from within your ide.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -527,7 +530,7 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
 
                     //our new contact already has an OCR, delete the old one
                     } else {
-                        dmlWrapper.objectsToDelete.add(ocrOldCon);
+                        delete ocrOldCon;
```
force2b: This method (`manageOtherOCR`) is called by other methods from inside a for loop, so issuing individual dml operations within that for loop could be problematic. Is there a reason you moved away from adding the object to the `dmlWrapper.objectsToDelete` collection?

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -38,6 +38,10 @@
 public class OPP_AutomatedSoftCreditsService {
 
     private static final String RELATED_OPPORTUNITY_CONTACT_ROLE_NONE = '--None--';
+    private static final String OBJECT_TYPE_RELATIONSHIP = 'npe4__Relationship__c';
+    private static final String OBJECT_TYPE_AFFILIATION = 'npe5__Affiliation__c';
+    private static final String FILED_NAME_RELATED_CONTACT = 'npe4__RelatedContact__c';
+    private static final String FILED_NAME_CONTACT = 'npe5__Contact__c';
```
force2b: To be consistent with the above constant, should this one be `FIELD_NAME_AFFILIATED_CONTACT`?

force2b:
@joy-sfdo  Just got the following error saving an Opportunity.

- One Relationship between the Primary Contact and another Contact (Carl Harvey in this example)
- Edit the Opp and remove the Primary Contact and then assign "Carl Harvey" to the Honoree and Tribute Contact fields.
- Click save.

image

```
TDTM_Opportunity: execution of AfterUpdate caused by: System.DmlException: Update failed. First exception on row 0 with id 00K19000001chCSEAY; first error: INSUFFICIENT_ACCESS_ON_CROSS_REFERENCE_ENTITY, insufficient access rights on cross-reference id: [] Class.UTIL_DMLService.updateRecords: line 202, column 1 Class.TDTM_TriggerHandler.processDMLAllOrNothing: line 241, column 1 Class.TDTM_TriggerHandler.processDML: line 221, column 1 Class.OPP_OpportunityContactRoles_TDTM.run: line 81, column 1 Class.TDTM_TriggerHandler.runClass: line 183, column 1 Class.TDTM_TriggerHandler.run: line 105, column 1 Trigger.TDTM_Opportunity: line 34, column 1
```

force2b:
Then edited that same Opportunity add "Debra Knight" back as the Primary Contact. That saved and created a Primary OCR Role for Debra and a Solicitor OCR for Carl, but it deleted the Honoree and Tribute OCR roles that were created previously.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: Cool, I'll look into that.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: Good catch.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: I'll look into that.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
joy-sfdo: Yes, this is a timing issue with updating OCRs. There was a previous GitHub issue that had a similar timing issue when updating the Primary Contact on an Opportunity when that Contact already had an OCR for that Opportunity.

src/classes/OPP_AutomatedSoftCreditsService.cls
joy-sfdo: It is consistent based on the field names. The field for Relationships is RelatedContact__c whereas the field name for Affiliations is Contact__c. However, for readability, I can change the name.

joy-sfdo:
Thanks for the code review, Mike. I'll address the items listed and look into the error.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
force2b: Is there an alternative way to handle this so that it's not executing DML inside a for loop?

force2b:
@joy-sfdo Found another issue this morning. Created a new Org Donation, specifying a Primary Contact. The "Contact Role for Organizational Opps" field in NPSP Settings is blank. The result was an OCR record marked as IsPrimary, but with a blank role:
image

joy-sfdo:
@force2b -- That adds up. If the Role was not specified then the value would be blank. What is the expected behavior?

@mpusto -- What are your thoughts on the above issue - "blank OCR Role value when a value is not specified in NPSP Settings" @force2b found?

force2b:
@joy-sfdo Found another issue where duplicate OCR's are created. In this case, I have defined "Donor" as the "Contact Role for Organizational Opps" in NPSP Setting as well as set up an Affiliation for Debra Knight on the Edgewire Account with the "Related Opportunity Contact Role" field set to "Soft Credit".

**Opp showing dupe OCR's**
image

**The Affiliation:**
image

mpusto:
I looked at how this is handled currently for Individual Opps and there is validation in NPSP Settings to prevent this scenario from happening. It seems like we should have the same validation for consistency. @force2b @joy-sfdo Do you agree?

joy-sfdo:
@mpusto -- Sounds good. Please show me where the validation takes place and I'll update the code. Thanks.

Good catch, @force2b!

force2b:
@mpusto and @joy-sfdo  NPSP Settings will already require a value for this new field, however the field will still be blank when this new feature is deployed into all existing customer orgs. I think we should auto-populate the new field with a value of "Donor" when the updated package is installed (via the Install Script).

joy-sfdo:
@force2b -- If we auto-populate the field, I'd recommend the value of "Soft Credit" to distinguish it from Individual gifts that default to "Donor". Thoughts?

mpusto:
Looks like this is where the validation occurs: https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/STG_PanelContactRoles_CTRL.cls

For a default I think @joy-sfdo's suggestion of Soft Credit is good.

force2b:
Sounds great. Thank you.

@joy-sfdo There are two places that will need changing to support this:

1. UTIL_CustomSettingsFacade.configHouseholdSettings() - set the field to "Soft Credit"
2. STG_InstallScript.createNewDefaultCustomSettingRecords() - check to see if the field is blank and if so, update the value to "Soft Credit"

joy-sfdo:
@mpusto -- Thanks for the link to the code. I was asking about the action/steps a User would take to generate the validation error.

force2b:
Just leave the value blank in NPSP Settings.
image

joy-sfdo:
@force2b  -- I just recreated the error. Thanks.

mpusto:
My bad! Essentially a user would go to NPSP Settings-->Donations-->Contact Roles, click Edit, then attempt to clear the value from the Organization Contact Role field (change to --None--) and click Save.

joy-sfdo:
@force2b and @mpusto -- I see the error for Organizational Contact Roles being blank in my scratch Org.

@mpusto -- thanks for the steps.

joy-sfdo:
screen shot 2018-06-07 at 8 15 18 am

joy-sfdo:
@force2b -- I see that's the screen capture you shared earlier.

joy-sfdo:
@force2b -- 
In UTIL_CustomSettingsFacade.cls the Individual value is set in configContactsSettings and the Organizational value is set there too. When spinning up a new scratch org sometimes the value of "Soft Credit" is not set in the Opportunity Contact Roles. Therefore, when a person navigates to NPSP Settings the Soft Credit value displays, but when they click edit the field becomes blank. I know you're resolving this issue.

I'll checkout -- 
STG_InstallScript.createNewDefaultCustomSettingRecords() - check to see if the field is blank and if so, update the value to "Soft Credit"

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -238,56 +242,29 @@ public class OPP_AutomatedSoftCreditsService {
...
-
-        for (Id primaryContactId: primaryContactToRelationships.keySet()) {
-            Map<Id, npe4__Relationship__c> cleanedRelationships = new Map<Id, npe4__Relationship__c>();
-            List<npe4__Relationship__c> uniqueRelationships = new List<npe4__Relationship__c>();
-            List<npe4__Relationship__c> relationships = primaryContactToRelationships.get(primaryContactId);
-            for (npe4__Relationship__c eachRelationship : relationships) {
-                if (!cleanedRelationships.containsKey(eachRelationship.npe4__RelatedContact__c)) {
-                    cleanedRelationships.put(eachRelationship.npe4__RelatedContact__c, eachRelationship);
-                    uniqueRelationships.add(eachRelationship);
+    private static Map<Id, List<sObject>> removeDuplicateAssociatedRecords(Map<Id, List<sObject>> sourceIdToAssociatedRecords, sObjectType objectType, sObjectField fieldName) {
```
force2b: @joy-sfdo Doesn't look like the `objectType` var is used here any more. Even the two constants for `OBJECT_TYPE_AFFILIATION` and `OBJECT_TYPE_RELATIONSHIP` aren't used either.

judisohn:
@joy-sfdo On the Affiliation object, the Related Opportunity Contact Role field has 'Soft Credit' 'Solicitor' as the out of the box options, but also 'Matching Gift' 

Why 'Matching Gift'? This seems out of place and would need to be created as a defined OCR value if used, no?

joy-sfdo:
@judisohn -- Thanks for following up. The value 'Matching Gift' is one of the field values specified in the ACs for the following Work Item -- 
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x800000001k20AAA/view.

@mpusto -- Please see Judi's questions regarding the 'Matching Gift' field value.

PR #3283, opened 2018-06-12
Rollups: allow a datetime for the date field

bethbrains:
# Critical Changes

# Changes
- Datetime is an acceptable data type for date field, allowing first/last operations by createddate, for instance.

# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
@force2b this looks good from UI, but can you verify the back-end change is ok too so we can merge this?

force2b:
@randi274 Is this a critical one to get into the current release?

force2b:
@bethbrains and @randi274 I'm not seeing where this works. In the example below I'm trying to populate a Date field on the Account with the CreatedDate for the Opportunity, however CreatedDate is not an available field.
image

bethbrains:
@force2b the change was to allow datetime in the Date field to allow first/last sorting by a datetime like CreatedDate, not the Field To Roll Up. That would be a different story (that we can let the community ask for, as they did for this one).

PR #3286, opened 2018-06-12
Account Creation Uses Contact Owner

### LINKED ISSUE #2577: Account Converter should create accounts with Contact's owner
Current the Account converter does not explicitly set the owner of any Household Accounts that get created, so the running user defaults to the owner.  Customers have requested to use the Contact owner.

force2b:
Ensure that if the Contact.OwnerId is set manually during a Contact insert (Apex, DataLoader, etc.) as well as during Household-Account conversion, so that the Account inserted is assigned to the Contact Owner.

# Critical Changes

# Changes

When creating a new Contact through Apex or the Data Loader and the OwnerId has been set, the new Household Account created (if required) will be assigned the same owner as the new Contact.

# Issues Closed

Fixes #2577 

# New Metadata

# Deleted Metadata


src/classes/CONV_Account_Conversion_BATCH_TEST.cls
```diff
@@ -49,19 +49,36 @@ public with sharing class CONV_Account_Conversion_BATCH_TEST {
         if(ADV_PackageInfo_SVC.useAdv()) return;
 
         if (strTestOnly != '*' && strTestOnly != 'testAccountConversion1to1') return;
+
+        // Create a single new User
+        Id sysAdminProfileId = UTIL_Profile.getInstance().getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR)[0];
```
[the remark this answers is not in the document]
force2b: Good idea! The method could actually create a more random username/alias which goes towards our eventual need for a better test data generator.

src/classes/CONV_Account_Conversion_BATCH.cls
```diff
@@ -219,6 +223,7 @@ global class CONV_Account_Conversion_BATCH implements Database.Batchable<SObject
                 c.AccountId = a.Id;
                 if (!conGiving.containsKey(a.Id) || (conGiving.containsKey(a.Id) && c.npo02__TotalOppAmount__c > conGiving.get(a.Id))){
                     a.npe01__One2OneContact__c = c.Id;
+                    a.OwnerId = c.OwnerId;
```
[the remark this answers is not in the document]
jjbennett: Awesome, should we have a test case around this as well?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -559,16 +560,18 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
                 a.Name = CAO_Constants.isHHAccountModel() ? strNameHHAccountForContact(c) : strNameOne2OneAccountForContact(c);
                 
-                //connect the Account to the Contact for 1:1 and HHAccount
+                // connect the Account to the Contact for 1:1 and HHAccount
                 if (c.Id != null) {
                     a.npe01__One2OneContact__c = c.Id;
                 }
                 a.Phone = c.Phone;
                 a.Fax = c.Fax;
+                a.OwnerId = c.OwnerId;
```
jjbennett: @force2b do we need to be concerned with the owner being active here as well?

force2b: @jjbennett  Good question. No because normally it would not be possible to insert a Contact with an Inactive Owner. However if for some reason a setting is enabled to allow an Contact insert with an inactive owner, then we'd want the Account owner to be the same.

jjbennett: Ok, great thanks!

jjbennett: It looks like a test is failing regarding an inactive owner.

force2b: Working on that still. Almost there.

jjbennett:
@sam-knox @mpusto - @force2b is wondering if we might want to call this change out in the changes section as well since it goes beyond the fix? We will now default to Contact owner whenever it is set, via conversion, data loader, apex, etc.

mpusto:
I agree, it's a change we should call out.

PR #3291, opened 2018-06-12
Rollups: Exclude skew dispatcher jobs from panel progress component

bethbrains:
# Critical Changes

# Changes

- When Customizable Rollups are enabled, Skew dispatcher jobs now display correctly in the panel progress component within NPSP Settings

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/UTIL_JobProgress_CTRL.cls
```diff
@@ -87,10 +87,14 @@ public with sharing class UTIL_JobProgress_CTRL {
         }
 
         Integer completedCount = 0;
-        //Query the Batch apex jobs
+        //Query the Batch apex jobs, excluding skew dispatcher jobs
+        String skewDispatcherClassName =  CRLP_SkewDispatcher_BATCH.class.getName();
```
force2b: Like with the prevent dupes code, this will need to strip out the namespace from the class name before using in the query.

src/classes/UTIL_JobProgress_CTRL.cls
```diff
@@ -151,6 +155,7 @@ public with sharing class UTIL_JobProgress_CTRL {
         List<AsyncApexJob> jobs = new List<AsyncApexJob>();
         List<BatchJobStatus> jobsStatus = new List<BatchJobStatus>();
 
+        String skewDispatcherClassName =  CRLP_SkewDispatcher_BATCH.class.getName();
```
force2b: same here

src/classes/STG_PanelOppBatch_CTRL.cls
```diff
@@ -95,7 +95,7 @@ public with sharing class STG_PanelOppBatch_CTRL extends STG_Panel {
                 Boolean executeSkewMode = (database.countQuery(query) != 0);
                 if (executeSkewMode) {
                     Database.executeBatch(new CRLP_SkewDispatcher_BATCH(jobType), CRLP_SkewDispatcher_BATCH.BATCH_SIZE);
-                    this.numJobSubmitted++;
+                    // do not increment numJobSubmitted because we do not display skew dispatcher jobs
```
force2b: What concerns me about this line is that the SkewDispatcher class itself will kick off the other Skew batch jobs (such as CRLP_AccountSkew_BATCH). Not incrementing the counter here could hide that job from the list.

force2b:
** lurch: attach W-024894

force2b:
@bethbrains and @kofijohnson This is ready for review. I've been testing the page in the LDV-XL org since that has enough data for the batch jobs to run. All of this code is in that org already.

src/classes/UTIL_JobProgress_CTRL.cls
```diff
@@ -31,139 +31,163 @@
...
+        if (!batchJobs.isEmpty()) {
+            this.bj = batchJobs[0];
+        }
+
         return batchJobs;
     }
 
     //This is the wrapper class the includes the job itself and a value for the percent complete
-    public Class BatchJob{
+    public class BatchJob {
```
kofijohnson: @force2b this wrapper class BatchJob and the wrapper class BatchJobStatus look almost the same. Should we merge them?

PR #3292, opened 2018-06-12
Feature/rollups  fls soql bug

### LINKED ISSUE #3280: Customizable Rollups Error Related to Field Level Security
If a non-System Administrator creates a Won Opportunity or sets an existing Opportunity to Closed Won (triggering rollups) and they don't have permissions on fields referenced in customizable rollups, an error is captured in the Error object. These should not be dependent on FLS of the current user.

**Error Type** = System.SObjectException
**Full Message:** SObject row was retrieved via SOQL without querying the requested field
{Name of Field}

The majority of reports involve the Opportunity's Primary Contact field, but we've also replicated with a custom field where the creating user did not have permissions. When a System Administrator uses the Recalculate Rollups button, no errors occur. Likewise, if the running user of the scheduled job has permissions, there are no errors.

primary contact
custom field

bethbrains:
# Critical Changes

# Changes

- Removed Field Accessibility check from soql on parent records

# Issues Closed
Fixes #3280

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Query_SEL.cls
+++ b/src/classes/CRLP_Query_SEL.cls
@@ -230,18 +230,16 @@
         }
 
         for (Schema.DescribeFieldResult fld : allFieldsToQuery) {
-            if (fld.isAccessible()) {
-                addlQuery += ',' + parentKeyField + '.' + fld.getName();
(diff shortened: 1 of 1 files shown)


bethbrains:
@force2b removed the other isaccessible check

PR #3306, opened 2018-06-14
Rollups: payment rollup type bug when using percent summary field

bethbrains:
# Critical Changes

# Changes
- Amount object is always defaulted. Amount field is only defaulted when Summary Field is not percent.

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/CRLP_Rollup/CRLP_RollupHelper.js
```diff
@@ -912,27 +912,37 @@
...
+            if(setAmountField) {
                 amountFieldName = labels.objectOpportunity + ' Amount';
-                cmp.set("v.activeRollup.detailObject", labels.objectOpportunity);
             }
-            cmp.set("v.activeRollup.amountField", amountFieldName);
-            cmp.set("v.activeRollup.amountFieldLabel", this.retrieveFieldLabel(amountFieldName, amountFields));
+            cmp.set("v.activeRollup.detailObject", labels.objectOpportunity);
         }
+        cmp.set("v.activeRollup.amountField", amountFieldName);
+        cmp.set("v.activeRollup.amountFieldLabel", this.retrieveFieldLabel(amountFieldName, amountFields));
```
randi274: Is this a conscious choice to have amountFieldName set as undefined for percents? Also, is setting the amountFieldLabel necessary for percents too?

PR #3307, opened 2018-06-14
Rollups: Prevent duplicate batches running

bethbrains:
# Critical Changes

# Changes
- Before a batch kicks off, it checks to see if there are other instances of itself still in process, and only proceeds if it finds none.
- If a job is skipped because the previous day's job is still running, an Error is logged with an explanatory message.

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
Let's try this one again, @force2b !!! :) Sorry about that.

PR #3308, opened 2018-06-14
Automated Soft Credits

### LINKED ISSUE #961: As a user, I'd like to have OCRs automatically created for Contacts based on relationship defined through the realtionship object so I can manage board referrals, etc.
"Is there a way for contact relationships to automatically translate into an opportunity contact role?
Our Board has a minimum give/get requirement that we have to track. Our Board members provide us with their solicitation lists, which we have uploaded into relationships in Salesforce with the Board member as "referrer" and their prospect as "referee." For every single time the prospect gives, we would like this relationship to automatically translate into a "referrer" contact role in the donation opportunity, so that we can then attribute soft credit. This would make it significantly easier to track their give/get requirement."
https://powerofus.force.com/0D58000001OrXDY

### LINKED ISSUE #1893: As a user, I would like an Affiliated Contact to receive automatic soft credit on Organization Opportunity
Thread here: https://powerofus.force.com/0D58000002S0LVl

### LINKED ISSUE #2708: Update Soft Credit rollups to count Organizational gifts with a soft credit Opportunity Contact Role marked Primary
For a Household opportunity, the Primary Contact field on the opportunity automatically creates an Opportunity Contact Role (OCR) record, set to the donor role and checked as primary. This is right and gives that contact full soft credit and the gift rolls up to the Household. Good.

However, we do not have distinct behavior for Business gifts and this is challenging. Organizations want to track who the grant officer/primary contact is on those gifts from organizations, so it's logical to use the Primary Contact field on the Opportunity. And they want that contact to receive soft credit. When they populate the Primary Contact field, that automatic OCR is created and checked as Primary. NPSP ignores that OCR since it won't roll up as hard credit and with the primary box checked (and using the donor role) it won't roll up as soft credit either. Requires manual editing to correct.

Suggest we have settings allowing the admin to specify the specific behavior for Individual gifts vs the behavior for Organization gifts when it comes to the Primary Contact.

Thread [here](https://powerofus.force.com/0D58000003avBWO), [here](https://powerofus.force.com/0D58000003WgGqB) and [here](https://powerofus.force.com/0D58000003HShda) - and I'm sure there are quite a few more.

### LINKED ISSUE #2894: Updating Opportunity Primary Contact to a Contact already receiving a Contact Role updates Contact Role to correct Contact and Role Name, but does not check 'Primary'
With fresh NPSP 3.117 and default NPSP Contact Role settings:

1. Create an Opportunity from a Contact ("John Smith") using NPSP New Contact Donation button.  The new Opp receives a Contact Role record where John Smith is Contact, Primary = TRUE, Role = Donor, as expected.

2. Click New button on Contact Role to add a new CR record.  Choose a different Contact ("Jane Doe"), Role = "Influencer", Primary = FALSE.  Now there are 2 CR records - John is primary, Jane is not.

3. Update the value of the Primary Contact field on Opportunity to "Jane Doe".

Issue: John's CR record is deleted and Jane's CR record is not set to Primary=TRUE, leaving the Opp with no Primary CR records.  I believe that setting Primary Contact on Opportunity should never leave the Opp with no primary CR records.

Note: It doesn't matter whether Jane is in the same or a different HH Account than John.

joy-sfdo:
# Critical Changes

# Changes
## Automated Soft Credits

NPSP now supports several new automated soft credits. Previously, NPSP could automatically create soft credits for Household members of a donor, for Honoree or Notification Contacts, and for donors in a Matching Gift scenario.

NPSP now also creates automated soft credits for the:

- Primary Contact on an Organizational Opportunity
- Related Contact on a Relationship record with an Individual donor
- Contact on an Affiliation record with an Organizational donor

Many organizations get by just fine with only Household soft credits. But some organizations have a donor base that requires a little more attention. Depending on the kind of organization you work for, you might have to accommodate donor advised funds, memorial gifts, planned giving, major donor portfolios, giving circles, crowdfunding, peer-to-peer fundraising, and much more. In many of these cases, you need to assign soft credits to one or more people, such as a board member, members of a family foundation, or simply relatives who don’t live in the same household.

Getting all of this right is important! Donors with more complex needs will expect that you are representing their family, businesses, associates, and financial institutions correctly, and that your organization is crediting the right people for the donations they've helped to bring in.

NPSP now lets you configure soft credit contact roles in multiple ways to help accommodate these different scenarios. And since NPSP creates the soft credits automatically, you can scale your fundraising efforts while continuing to honor the wishes of your donors.

**Resources:**

[Automated Soft Credits](https://powerofus.force.com/articles/Resource/Automated-Soft-Credits)
[NPSP Logic for Creating Contact Roles](https://powerofus.force.com/articles/Resource/Logic-for-Creating-OCRs)
[Configure Soft Credits](https://powerofus.force.com/articles/Resource/NPSP-Configure-Soft-Credit-and-Matching-Gift-Rollup-Settings)

# Issues Closed
Fixes #961
Fixes #1893
Fixes #2894
Fixes #2708

Ideas:
Idea [1040 -- Automatically create soft credit Contact Roles based on Relationship](https://foundation.lightning.force.com/lightning/r/Community_Idea__c/a1W80000004DUb2EAG/view)
Idea [1133 -- Soft Credit Contact Role on Organization donations based on Primary Contact](https://foundation.lightning.force.com/lightning/r/Community_Idea__c/a1W80000006bkSJEAY/view)
Idea [1044 -- Automatically create soft credit Contact Roles based on Affiliation](https://foundation.lightning.force.com/lightning/r/Community_Idea__c/a1W80000004DUllEAG/view)

# New Metadata

* OPP_AutomatedSoftCreditsService.cls
* OPP_AutomatedSoftCreditsService_TEST.cls

# Deleted Metadata


src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -158,7 +166,7 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
     ********************************************************************************************************/
     private static void updateHonoreeNotificationFields(list<Opportunity> listOpps, list<Opportunity> listOldOpps) {
         map<id,Contact> mapConIdContact = new map<id,Contact>();
-        list<String> listFields = new list<string>{'Honoree_Contact__c','Notification_Recipient_Contact__c'};
+        list<String> listFields = new list<string>{FIELD_NAME_HONOREE_CONTACT, FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT};
```
force2b: @joy-sfdo I'm not sure this will work in a namespaced org. These _constants_ contain a non-namespace field name only, so once the `npsp` namespace is applied I'm pretty sure the code below (not displayed here - line 175+) will fail when it attempts to do a `get()` on the field name. Given how these constants are used (there's a 3rd one up there too), you may need to change everything to an SObject Field Token rather than a string unless you're confident that the namespace can be applied to the string without breaking any string comparisons. I'll leave the final implementation up to you.

src/classes/STG_PanelContactRoles_CTRL.cls
```diff
@@ -87,13 +87,20 @@ public with sharing class STG_PanelContactRoles_CTRL extends STG_Panel {
...
+        //add error for blank Default Individual Contact Role value
         if (String.isBlank(STG_Panel.stgService.stgCon.npe01__Opportunity_Contact_Role_Default_role__c)) {
             String defaultRoleLabel = UTIL_Describe.getFieldLabel('npe01__Contacts_And_Orgs_Settings__c', 'npe01__Opportunity_Contact_Role_Default_role__c');
             ApexPages.addMessage(new ApexPages.message(ApexPages.Severity.ERROR, defaultRoleLabel + ' ' + Label.isARequiredField));
             isValid = false;
         }
 
+        // Add error for blank Default Oranizational Contact Role value
+        if (String.isBlank(STG_Panel.stgService.stgCon.Contact_Role_for_Organizational_Opps__c)) {
+            String defaultOrganizationalRoleLabel = UTIL_Describe.getFieldLabel('npe01__Contacts_And_Orgs_Settings__c', 'Contact_Role_for_Organizational_Opps__c');
```
force2b: I think this is another spot that won't work in a namespaced org because the `"Contact_Role_for_Organizational_Opps__c"` fieldname is hardcoded without a namespace. You can call `UTIL_Namespace.StrTokenNSPrefix()` with the fieldname to add the namespace.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
joy-sfdo: @force2b  -- I tested in a namespaced scratch org and the code ran as expected. Given that Strings were used originally, I'd like to leave this code as is to minimize changes. However, if something is uncovered during testing, I'm open to making the necessary changes.

src/classes/STG_PanelContactRoles_CTRL.cls
joy-sfdo: @force2b -- Good catch! Thanks for identifying this issue. I've updated the code per your suggestion and it ran successfully in both a namespaced and a regular scratch org.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
force2b: @joy-sfdo I see what you mean. It's really really strange. There's absolutely no valid reason this should work. The field string is not namespaced, yet I absolutely agree that it is not crashing. I tested this out heavily with debugging on and added some debug statements to see what's happening in the code. Here's an example showing two gets, one with a namespace and one without. They **both** work!
```
ERROR|>>>> NewOpp.npsp__Honoree_Contact__c == "003f000000z0SftAAE"
ERROR|>>>> NewOpp.Notification_Recipient_Contact__c == "003f000000z0SfvAAE"
```

However, I'm concerned that this is simply a core bug that allows us to get away with field get. I can see where this can easily fail if the customer has a custom field with the same field name. As a result, I think we must fix it in the code even if it's currently working.

joy-sfdo: @force2b -- Adding the UTIL_Namespace method to the methods in OPP_OpportunityContactRoles_TDTM caused the code to stop working in a namespaced scratch org. I see your concern about it possibly being a core bug and impacting customers down the line. However, given that this code is already in production and customers are using it (Strings, the Constants are new), the risk of continuing with the code as is is lower than introducing a new element. Also, since adding the UTIL_Namespace code to the various methods using field get caused the code to stop working in a namespaced org, it will take time to investigate, debug, and rewrite the code which would mean stopping testing and possibly delaying the release of this feature. Therefore, we need to have a larger group discussion about modifying this code.

@mpusto, @alex-peralta -- Thoughts?

mpusto: I can't speak to it from a technical perspective, but is it possible to release and then change this next sprint to protect against the core issue? If it is working in both there will be no customer impact.

force2b: It does technically work in our namespaced scratch org, but that is not representative of a package install in a customer org. There is still an open known issue that relates to this as well -- https://success.salesforce.com/issues_view?id=a1p30000000sY2FAAU. For example, a customer who happens to have their own `Honoree_Contact__c` field on the Opportunity object would almost certainly either get an error creating/changing Opportunities or the triggers would not create/change the proper OCR's.

mpusto: Is there a way to test against a representative org?

force2b: @joy-sfdo pointed out that the actual base code was there before. I've personally bumped into this issue, so I'm setting up a full beta org to test against now. Given that the field names were hardcoded without namespaces previously (just in a different manner) I think that best case we leave it as is. Worse case, we can probably leave it as is and create a new GH issue for it as a bug to be fixed in a future release.

force2b: Ok. Worked fine in a beta org even with 3 custom fields with identical names to the packaged ones. I'm baffled. The known issue is still there, and I was personally subject to this issue a few years back. Maybe core changed the functionality to always assume that the packaged fields are the ones to retrieve when using string field names? I guess it would be an issue for code that needed to actually retrieve the customer's field as opposed to the managed one, but that's not an issue for us to worry about. 

Let's leave the code here as is. I apologize for the fire drill.

joy-sfdo: @force2b -- Sounds good. Thanks for calling out this item, the link to the Known Issue, and testing out in the beta org. 

I'll leave the code here as is.

PR #3324, opened 2018-06-19
Change Individual Account query to filter by Indexed Name field

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


jjbennett:
@force2b @RadSFDeveloper - this is the fix we spoke about this morning by not supported the individual account model in an encrypted org.

PR #3328, opened 2018-06-19
Rollups: changed labels to not reference active

bethbrains:
# Critical Changes

# Changes
- Label on list of Rollups referenced by a Filter Group no longer inaccurately specifies "Active"

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
If time permits, I may explore adding an indication in the list that a Rollup is inactive. For now, this is now accurate.

PR #3354, opened 2018-06-26
Fix NPE exception when for a blank Account on an Opp in One-To-One Model

### LINKED ISSUE #3353: OCR TDTM throws Exception for blank AccountId on Opportunity
This was introduced in the 3.132 release. If the AccountId is cleared from an Opportunity and our OPP_OpportunityContactRoles_TDTM Trigger Handler fires, a null pointer Exception is thrown.

`TDTM_Opportunity: execution of AfterUpdate caused by: System.NullPointerException: Attempt to de-reference a null object Class.OPP_OpportunityContactRoles_TDTM.upsertContactRolesAfterUpdate: line 382, column 1 Class.OPP_OpportunityContactRoles_TDTM.run: line 75, column 1 Class.TDTM_TriggerHandler.runClass: line 183, column 1 Class.TDTM_TriggerHandler.run: line 105, column 1 Trigger.TDTM_Opportunity: line 34, column 1`

This was originally surfaced by a customer using a custom trigger in their sandbox which seemed to clear the Account Id from the Opportunity, which on update, triggered our OCR TDTM to run.

I reproduced in a scratch org.

Steps to Repeat:
1. Create a new Account Record Type called One-to-One.
2. On the page layout you chose for the new Record Type, edit the Account Name properties so that it is not required.
3. Go to NPSP Settings-->People --> Account Model and set the Account Model to "One-to-One" and the One-To-One Record Type to "One-to-One".
4. Create a Contact. This will create an Account with the same name.
5. Create an Opportunity for that Contact. Do not fill in the Account Name. (But do fill in the Contact).
6. Save the record and you will see the error. [Screenshot attached]
ocr tdtm

force2b:
To prevent an NPE exception, check for a null `AccountId` value before using that in a get() against the `oppAccountDetails` map. Primarily updated the first loop to match the second loop in its structure and pattern, which made it easier to apply the check for a null AccountId.

# Critical Changes

# Changes

# Issues Closed

Fixes #3353 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/OPP_OpportunityContactRoles_TDTM.cls
+++ b/src/classes/OPP_OpportunityContactRoles_TDTM.cls
@@ -375,22 +375,29 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
         // Retrieve the Account Ids for the Organizational Opportunities
         Map<Id, List<Id>> accountsToOpportunities = populateAccountToOpportunities(listOpps, oppAccountDetails);
 
-        //find changed opportunities and instantiate a map entry to hold OCRs
-        for (integer i=0; i<listOpps.size(); i++) {
-            if (listOpps[i].Primary_Contact__c != listOldOpps[i].Primary_Contact__c ||
-                ((listOpps[i].AccountId != listOldOpps[i].AccountId)
-                    && (!oppAccountDetails.get(listOpps[i].AccountId).npe01__SYSTEMIsIndividual__c || !oppAccountDetails.containsKey(listOldOpps[i].Id))) ||
-                needsManageOCR(listOpps[i], listOldOpps[i], FIELD_NAME_HONOREE_CONTACT) ||
-                needsManageOCR(listOpps[i], listOldOpps[i], FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT)) {
+        // find changed opportunities and instantiate a map entry to hold OCRs
+        for (Integer i=0; i<listOpps.size(); i++) {
+            Opportunity newOpp = listOpps[i];
+            Opportunity oldOpp = listOldOpps[i];
 
-                mapOppIdMapConIdOCR.put(listOpps[i].Id, new Map<Id, OpportunityContactRole>());
+            Boolean isIndividualAccount = (newOpp.AccountId != null && oppAccountDetails.containsKey(newOpp.AccountId))
+                    ? oppAccountDetails.get(newOpp.AccountId).npe01__SYSTEMIsIndividual__c
+                    : null;
+
+            if (newOpp.Primary_Contact__c != oldOpp.Primary_Contact__c ||
+                    ((newOpp.AccountId != oldOpp.AccountId) && (isIndividualAccount == false ||
+                        !oppAccountDetails.containsKey(oldOpp.Id))) ||
+                        needsManageOCR(newOpp, oldOpp, FIELD_NAME_HONOREE_CONTACT) ||
+                        needsManageOCR(newOpp, oldOpp, FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT)) {
+
+                mapOppIdMapConIdOCR.put(newOpp.Id, new Map<Id, OpportunityContactRole>());
             }
 
-            if (listOpps[i].Primary_Contact__c != listOldOpps[i].Primary_Contact__c ||
-                ((listOpps[i].AccountId != listOldOpps[i].AccountId)
-                    && (!oppAccountDetails.get(listOpps[i].AccountId).npe01__SYSTEMIsIndividual__c || !oppAccountDetails.containsKey(listOldOpps[i].Id)))) {
+            if (newOpp.Primary_Contact__c != oldOpp.Primary_Contact__c ||
+                    ((newOpp.AccountId != oldOpp.AccountId) && (isIndividualAccount == false ||
+                        !oppAccountDetails.containsKey(oldOpp.Id)))) {
 
-                primaryContactAccountChangeOpptys.add(listOpps[i]);
+                primaryContactAccountChangeOpptys.add(newOpp);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -375,22 +375,29 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
-                ((listOpps[i].AccountId != listOldOpps[i].AccountId)
-                    && (!oppAccountDetails.get(listOpps[i].AccountId).npe01__SYSTEMIsIndividual__c || !oppAccountDetails.containsKey(listOldOpps[i].Id))) ||
-                needsManageOCR(listOpps[i], listOldOpps[i], FIELD_NAME_HONOREE_CONTACT) ||
-                needsManageOCR(listOpps[i], listOldOpps[i], FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT)) {
+        // find changed opportunities and instantiate a map entry to hold OCRs
+        for (Integer i=0; i<listOpps.size(); i++) {
+            Opportunity newOpp = listOpps[i];
+            Opportunity oldOpp = listOldOpps[i];
+
+            Boolean isIndividualAccount = (newOpp.AccountId != null && oppAccountDetails.containsKey(newOpp.AccountId))
```
jjbennett: What do you think about pulling this into it's own method? We have the same variable declared 3x, the [first one](https://github.com/SalesforceFoundation/Cumulus/blob/4f58258775c723310459ad808f10488eb9631726/src/classes/OPP_OpportunityContactRoles_TDTM.cls#L251) doesn't have the accountid null check but not sure if it's really needed either?

force2b: Good suggestion.

joy-sfdo: I also agree with @jjbennett suggestion for putting the code for determining the value of the isIndividualAccount in its own method now that the code is being called twice in a method. However, I understand minimizing changes. 

I'd also recommend adding the null check to the isIndividualAccount code in the `upsertContactRolesAfterInsert` method.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -375,22 +373,27 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
-                mapOppIdMapConIdOCR.put(listOpps[i].Id, new Map<Id, OpportunityContactRole>());
+                mapOppIdMapConIdOCR.put(newOpp.Id, new Map<Id, OpportunityContactRole>());
             }
 
-            if (listOpps[i].Primary_Contact__c != listOldOpps[i].Primary_Contact__c ||
-                ((listOpps[i].AccountId != listOldOpps[i].AccountId)
-                    && (!oppAccountDetails.get(listOpps[i].AccountId).npe01__SYSTEMIsIndividual__c || !oppAccountDetails.containsKey(listOldOpps[i].Id)))) {
+            if (newOpp.Primary_Contact__c != oldOpp.Primary_Contact__c ||
+                    ((newOpp.AccountId != oldOpp.AccountId) && (isIndividualAccount == false ||
+                        !oppAccountDetails.containsKey(oldOpp.Id)))) {
```
jjbennett: Won't this always return true since oppAccountDetails has AccountIds as keys?

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -375,22 +373,27 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+        // find changed opportunities and instantiate a map entry to hold OCRs
+        for (Integer i=0; i<listOpps.size(); i++) {
+            Opportunity newOpp = listOpps[i];
+            Opportunity oldOpp = listOldOpps[i];
+
+            Boolean isIndividualAccount = isIndividualAccount(oppAccountDetails, newOpp.AccountId);
+
+            if (newOpp.Primary_Contact__c != oldOpp.Primary_Contact__c ||
+                    ((newOpp.AccountId != oldOpp.AccountId) && (isIndividualAccount == false ||
+                        !oppAccountDetails.containsKey(oldOpp.Id))) ||
```
jjbennett: Same question here on using the oppId.

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -879,5 +879,67 @@ private class OPP_OpportunityContactRoles_TEST {
...
+
+        Test.startTest();
+        opp1.AccountId = null;
+        update opp1;
+        OpportunityContactRole[] result = [SELECT OpportunityId, Opportunity.AccountId, ContactId, isPrimary, Role FROM OpportunityContactRole WHERE OpportunityId = :opp1.Id];
+        //should be a contact role
+        System.assertEquals(1, result.size());
+        System.assertEquals(OPPORTUNITY_CONTACT_ROLE_DONOR, result[0].Role);
+        System.assertEquals(true, result[0].isPrimary);
+
```
joy-sfdo: @force2b -- A few items:
1. Add `System.assertEquals(null, result[0].Opportunity.AccountId);` after line 933
2. Formatting suggestion -- Change from `OpportunityContactRole[]` to `List<OpportunityContactRole>`
3. Change the variable name from `result` to `results` since the data type is a List/Array.

None of the above items are showstoppers.

force2b: @joy-sfdo Very good points. I copied an existing test method and changed only a couple of items. I'm also not a fan of the pattern of using `[]` versus `List<>` or querying a single record into a list and then references as `[0]`.  As much as I really want to fix all those items, I think we're in a time crunch given how long builds take so if you don't mind I think we should merge as is.

joy-sfdo: @force2b -- Sounds good, given the time crunch.

PR #3359, opened 2018-06-27
Feature/bge nima  sal 579

ricardonimacloud:
# Critical Changes

# Changes
Added custom editors for differents cell types for change the standard center alignment for left and right

# Issues Closed
SAL-579
# New Metadata

# Deleted Metadata


StaticResourceSources/BGEResources/js/grid/app.js
```diff
@@ -783,27 +823,33 @@
...
                     col.className = "htRight htMiddle slds-truncate";
                     col.title = '<div class="amount-style">' + templateField.label.toUpperCase() + '</div>';
                     col.colWidths = 80;
+                    col.editor = NumberEditorCustom;
                 }
-                else if (templateField.type === "EMAIL") {
+                else if (templateField.type === "EMAIL" || templateField.type === "STRING" ||
+                         templateField.type === "TEXTAREA" || templateField.type === "URL") {
 
+                    col.editor = TextEditorCustom;
```
mdeleon0: You can put this line above before starting with the IF clauses therefore making TextEditorCustom the default option and omit it everywhere else, so the editor would be set only on the columns where the editor is not TextEditorCustom.

PR #3369, opened 2018-06-29
Adding a global run method to TDTM_Config_API

### LINKED ISSUE #2569: TDTM framework on custom object
When attempting to create a trigger on a custom object and utilize the TDTM framework
```
trigger TDTM_Account on My_Custom_Object__c (after delete, after insert, after undelete, 
    after update, before delete, before insert, before update) {

    npsp.TDTM_TriggerHandler handler = new npsp.TDTM_TriggerHandler();  
    handler.run(Trigger.isBefore, Trigger.isAfter, Trigger.isInsert, Trigger.isUpdate, Trigger.isDelete, 
        Trigger.isUnDelete, Trigger.new, Trigger.old, Schema.Sobjecttype.My_Custom_Object__c, 
        new npsp.TDTM_ObjectDataGateway());
}
```
I get an error `Type is not visible: tdtm_triggerhandler`.

This appears to be because the class is set to public instead of global, essentially saying only the managed package can use said class.

In doing research, it looks like HEDAP uses the same TDTM_TriggerHandler code but they have provided a mechanism for other objects to utilize the framework. https://github.com/SalesforceFoundation/HEDAP/issues/267

Can this be added to NPSP as well?

lparrott:
# Critical Changes

# Changes

- New global `run` method within the TDTM_Config_API class.
- The `run` method lets users create triggers for their custom objects or for standard objects for which NPSP doesn't have a trigger.
- Existing NPSP triggers have been updated to use this method. There is no action required by Admins.

# Issues Closed

Fixes #2569 

# New Metadata

# Deleted Metadata

### DIFF
--- a/ApexDocContent/TDTM.htm
+++ b/ApexDocContent/TDTM.htm
@@ -15,6 +15,7 @@
         <li>TDTM_TriggerActionHelper - class that contains helper methods for the TDTM Trigger Handler logic.</li>
         <li>TDTM_TriggerHandler - class called from each of the triggers (only one per object) that follow the TDTM design. 
         It's in charge of figuring out which of the classes that do the actual work need to be called, and calling them.</li>
+        <li>TDTM_Config_API - global API class that exposes the TDTM configuration, including the run method.</li>
     </ul>
     </body>
 </html>
--- a/src/classes/TDTM_Config_API.cls
+++ b/src/classes/TDTM_Config_API.cls
@@ -66,6 +66,26 @@ global class TDTM_Config_API {
         return TDTM_ObjectDataGateway.listTH;
     }
 
+    /*******************************************************************************************************
+    * @description Method to be called from each of the triggers (only one per object) that follow the TDTM 
+    * design. Figures out which classes need to be called, and calls them.
+    * @param isBefore Are we in a before trigger context.
+    * @param isAfter Are we in an after trigger context.
+    * @param isInsert Did the trigger run in response to an insert operation.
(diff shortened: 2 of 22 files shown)


sam-knox:
I just saw this, and it's great. I will update the TDTM docs for NPSP to reflect this change. Right now they say that you can't create triggers in the TDTM architecture but with this global run method, admins can.

sam-knox:
@lparrott - is this documented in the NPSP Codebase doc already?

ceiroa:
That should be all, @lparrott. With that and the annotations the ApexDocs should get automatically updated once we cut the release.

lparrott:
@juliesnow Would you mind reviewing the Changes listed in this PR? Carlos suggested combining it into a single bullet point.

annejackson:
FYI @lparrott  and @juliesnow . The second bullet point covers this but doesn't mention admins (just users), so not sure if we want any change to read how @sam-knox  is doing the TDTM doc changes: He mentioned this:
TDTM docs for NPSP to reflect this change. Right now they say that you can't create triggers in the TDTM architecture but with this global run method, admins can.

juliesnow:
@lparrott @ceiroa @annejackson - Sorry, was out end of last week. Pls lmk if updates to "Changes" look good.

ceiroa:
Thanks, @juliesnow! The only suggestion I'd make is combining the "No action is required as a result of these changes." line with the previous one. When the master build runs these bullets will be combined with those from all other pull requests, so it might not be clear then to which changes this is referring to. What do you think?

PR #3399, opened 2018-07-03
add initialize anon apex

cdcarter:
Add default batch templates to dev_config.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cdcarter:
@force2b - I got this anon apex block from Nic on W-024975. This is to make sure the pilot orgs have a template to use. Please approve and merge when you get a chance.

PR #3404, opened 2018-07-05
Incremental Rollup Mode Settings

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@kofijohnson  This PR as a grandchild branch of `rollups__IncrementalBatch` can work, but only if it's the only grandchild branch and as long as there are no merges into `feature/rollups` before this branch is merged. This is because MetaCI does not support more than a single level down. So in this case, any other merge into `feature/rollups` will be sync'd directly into `feature/rollups__IncrementalBatch__UI` rather than first sync'ing into `feature/rollups__IncrementalBatch` and then up into `feature/rollups__IncrementalBatch__UI`.

kofijohnson:
We are closing this PR because we have created a new one https://github.com/SalesforceFoundation/Cumulus/pull/3606

PR #3411, opened 2018-07-05
Fixing duplicate default allocation issue

pedrodalcol:
# Critical Changes

# Changes
- Fixing issue that causes default GAU allocation amount to be incorrectly calculated when there are more than one allocation to the default GAU on the same Opportunity.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/ALLO_Allocations_TDTM.cls
+++ b/src/classes/ALLO_Allocations_TDTM.cls
@@ -62,6 +62,7 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
         decimal totalPercent = 0;                                            //holds total percentage allocations
         list<Allocation__c> triggerList = new list<Allocation__c>();         //all allocations for this opportunity in the current trigger set
         list<Allocation__c> listAllo = new list<Allocation__c>();            //all non-default allocations for this parent object
+        Map<Id, Allocation__c> mapDefaultDupes = new Map<Id, Allocation__c>();//duplicate default allocations, if any
         Allocation__c defaultAllo = null;                                    //the default allocation for this parent object
         boolean defaultInTrigger = false;                                    //is the default allocation in the current trigger set?
     }
@@ -249,7 +250,7 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
 
             //verify the total allocations don't exceed the parent amount
             if (wrap.totalAmount > wrap.parentAmount) {
-                allo.addError(Label.alloTotalExceedsOppAmt);
+                allo.addError(Label.alloTotalExceedsOppAmt + ' ' + wrap.totalAmount);
             }
 
         }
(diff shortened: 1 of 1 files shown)


judisohn:
Should we have an issue for this as it's more of a fix than a change (thinking of the release notes)?

pedrodalcol:
@judisohn I created issue #3434 to track this.

pedrodalcol:
### Testing Notes
1. Enable default allocations and set a default GAU.
2. Create a new Opportunity and set amount to $1000 (It will be automatically allocated to default GAU)
3. Click manage allocations and add an allocation for $500 to the the default GAU (from step 1).
4. Click save and return to the list of allocations. There should only be one allocation for $1000.
5. Edit the allocation using the standard UI, update the amount to 400 and save. It should get reset to $1000.

annejackson:
Don't forget once changes are approved and it's ready for testing to put me in the assignees field. Thanks @pedrodalcol  and @lparrott

annejackson:
@judisohn - what would you like us to do for the #3434 Issue item?

annejackson:
@pedrodalcol  before this fix what would have happened? At step 3 or 5 would it have created duplicate records? And this is at the opportunity level not payment level? I could not duplicate this on this feature branch in a scratch org- (feature/pdalcol/Payment-allocations)
**Testing Notes**
1. Enable default allocations and set a default GAU.
2. Create a new Opportunity and set amount to $1000 (It will be automatically allocated to default GAU)
3. Click manage allocations and add an allocation for $500 to the the default GAU (from step 1).
4. Click save and return to the list of allocations. There should only be one allocation for $1000.
5. Edit the allocation using the standard UI, update the amount to 400 and save. It should get reset to $1000.

annejackson:
Tried this scenario and got an error:

1. Create an oppt for $1000.
2. Go to manage allocations (this didn't happen on the GAU allocations list view page)->and add many duplicates, not just one duplicate GAU item. So... I did one line of default GAU $300, another line of default GAU $300, and another line of default GAU $300, then pressed save. 
Before the fix: The page will try to add a fourth default automatically for $100. So you will end up with 4 default allocations without the fix.

With the fix: The page errors: 
Error: TDTM_Allocation: execution of AfterInsert caused by: System.QueryException: List has more than 1 row for assignment to SObject External entry point Class.ALLO_Allocations_TDTM.run: line 97, column 1 Class.TDTM_TriggerHandler.runClass: line 183, column 1 Class.TDTM_TriggerHandler.run: line 105, column 1 Trigger.TDTM_Allocation: line 34, column 1

see screenshot below. 

error

annejackson:
How do the changes sound for release notes @juliesnow  @judisohn?

judisohn:
@annejackson I haven't had a chance to test this in a scratch org yet myself. Is it just the default GAU or doesn't this apply to any GAU assigned more than once as allocations for the same Opportunity?

annejackson:
@judisohn  and @pedrodalcol  I did try breaking a 1000 opportunity into multiple allocations (that were not the non-default GAU) and it showed 3 separate allocation records. Should it? See screenshot:

3 seperate alllcation recrods

judisohn:
@annejackson yes that's the expected behavior.

annejackson:
This looks good. I am not longer seeing duplicates or the other error I found that @pedrodalcol  fixed.
@juliesnow  @judisohn (and perhaps @sam-knox  or @smarovich - could you take a look at the changes section for release notes before I merge? 
This fix/change only applies to default GAU's. And it was causing duplicate GAU's to be created and listed (which would effect GAU total amounts/rollups). 

So far we have: 
Fixing issue that causes default GAU allocation amount to be incorrectly calculated when there is more than one allocation to the default GAU on the same Opportunity.

juliesnow:
@annejackson @sam-knox @smarovich, sending comments in mail.

pedrodalcol:
@juliesnow it looks good to me!

PR #3412, opened 2018-07-06
Feature/rollups  invalid field date bug

### LINKED ISSUE #3395: Customizable Rollups: Invalid Field Error for Soft Credit Rollups Using Custom Date Field
When Soft Credit Rollups are modified so that their Date field is not Opportunity: Close Date, the following error is received (in this case, for a custom field on Opportunity called Commitment Date):

`Invalid field Commitment_Date__c for Partial_Soft_Credit__c` 

**Stack Trace:**
`Stacktrace available at the time of saving the record - Class.CRLP_Operation_SVC.doRollupOperation: line 81, column 1 Class.CRLP_RollupSoftCredit_SVC.processRollups: line 94, column 1 Class.CRLP_RollupProcessor_SVC.completeRollupForParent: line 295, column 1 Class.CRLP_RollupProcessor_SVC.startRollupProcess: line 114, column 1 Class.CRLP_Batch_Base_NonSkew.execute: line 263, column 1
 `

**Steps to Reproduce**
1. Create new Date field on Opportunity called Commitment Date
2. Enable Customizable Rollups in NPSP Settings. Edit the Contact: First Soft Credit Date Rollup and change the Date Field to Opportunity: Commitment Date.
3. Create a Contact named Jane Smith and add another Contact to the Household (John Smith).
4. Create an Opportunity for Jane Smith. (Amount = 10, Stage = Closed Won, Commitment Date = today, Close Date = today). Verify that a soft credit OCR was created for John Smith.
5. Go to John Smith's Contact record and click the Recalculate Rollups button.
6. The Invalid Field error is captured in the Error Log in NPSP Settings.

randi274:
# Critical Changes

# Changes

# Issues Closed
- Fix #3395 
# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Operation_SVC.cls
+++ b/src/classes/CRLP_Operation_SVC.cls
@@ -74,10 +74,10 @@ public class CRLP_Operation_SVC {
         // Get the Date value from the appropriate object. There should always be a date!
         // It's a LOT faster the use the actual field name to retrieve the date than to use .get(fieldName).
         // Since the Date is almost always the CloseDate field, hardcoding this performance tweak.
-        SObject dateSObject = detailSObject;
         if (rollup.dateObject == oppObjectName && dateField == 'CloseDate') {
             dateValue = oppSObject.CloseDate;
         }
+        SObject dateSObject = rollup.dateObject == 'Opportunity' ? oppSObject : detailSObject;
         if (dateValue == null && dateField != null && dateSObject.get(dateField) != null) {
             if (rollup.dateFieldType == SoapType.DATE) {
                 dateValue = (Date) dateSObject.get(dateField);


src/classes/CRLP_Operation_SVC.cls
```diff
@@ -74,10 +74,10 @@ public class CRLP_Operation_SVC {
         // Get the Date value from the appropriate object. There should always be a date!
         // It's a LOT faster the use the actual field name to retrieve the date than to use .get(fieldName).
         // Since the Date is almost always the CloseDate field, hardcoding this performance tweak.
-        SObject dateSObject = detailSObject;
         if (rollup.dateObject == oppObjectName && dateField == 'CloseDate') {
```
kofijohnson: @randi274, Do we need this if block here. We can remove the lines 77-80 and make the if block on line 80 simpler:
if (dateField != null && dateSObject.get(dateField) != null)

src/classes/CRLP_Operation_SVC.cls
```diff
@@ -74,10 +74,10 @@ public class CRLP_Operation_SVC {
...
             dateValue = oppSObject.CloseDate;
         }
+        SObject dateSObject = rollup.dateObject == 'Opportunity' ? oppSObject : detailSObject;
```
kofijohnson: Can we use the parameter variable "detailSObject" instead of creating a new one "dateSObject".

src/classes/CRLP_Operation_SVC.cls
randi274: I think this if block is here for performance reasons, as @force2b mentions in the comment above, so I'd be wary of changing it without his ok.

src/classes/CRLP_Operation_SVC.cls
randi274: So just reuse it like:
detailSObject = rollup.dateObject == 'Opportunity' ? oppSObject : detailSObject;

I don't think we can do that since detailSObject is reused later, so resetting it with a ternary may have adverse effects later on.

kofijohnson: I see your point. I missed we are assigning "oppSObject" as well. That is good then, thank you!

src/classes/CRLP_Operation_SVC.cls
kofijohnson: Yes, I see the comment as well. I wonder if the performance gain is significant. If @force2b confirms it is not significant, we can remove it to make the code simpler?

force2b: I'm pretty sure that I tested this in a for loop of a few thousand iterations to confirm that it is actually faster to directly reference Opp.CloseDate versus (Date)Opp.get('CloseDate'). The issue is that a single Opportunity may get passed through this logic 10+ times, and when there are 1000 opportunities being processed in a batch iteration, even micro-seconds start to add up.

dgamboni: I tested this and it seems like it's OK, but I'm unclear if it's been reviewed already and is ready to merge. Please let me know. Thanks!

PR #3456, opened 2018-07-11
Feature/bge nima  staging  sal 592

mdeleon0:
# Critical Changes
NO
# Changes
Modified function getHotColumns in file app.js so each column would have the corresponding width based on its display type.
# Issues Closed
Column widths should match expected content (e.g. Amount or ZIP are short fields; Address is a long field).
# New Metadata
NO
# Deleted Metadata
NO


mdeleon0:
This was already merged in PR https://github.com/SalesforceFoundation/Cumulus/pull/3453

PR #3479, opened 2018-07-13
Feature/bge  sal 594

ricardonimacloud:
# Critical Changes

# Changes
Fix error: when user, after delete rows, makes drag & drop or copy & paste for fill new rows with data, the external ids initially generated when table is initialized are removed, This causes errors when new data imports are being created.

# Issues Closed
SAL-594
# New Metadata

# Deleted Metadata


StaticResourceSources/BGEResources/js/grid/app.js
```diff
@@ -1007,6 +1016,7 @@
                 window.addEventListener('resize', updateHotTable, true);
             }
 
+            console.log('daleeeeeeeeee');
```
mdeleon0: Remove debug logs.

PR #3497, opened 2018-07-18
Rollups: show all OCR roles in Manage Soft Credits page when CRLPs is enabled

bethbrains:
# Critical Changes

# Changes
- When CRLPs is enabled, all OCR roles appear in the dropdown in Manage Soft Credits page, not just those in the Soft Credit Roles custom setting. (The custom setting is effectively deprecated when CRLPs are enabled)

# Issues Closed

# New Metadata

# Deleted Metadata


scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
```diff
@@ -0,0 +1,696 @@
+/**
```
force2b: @bethbrains Out of curiosity, why include this class in this PR as opposed to a new PR for it with the class and any related metadata files we want to related?

bethbrains: It wasn't in this PR.... I committed it directly to feature/rollups, or at least I thought I did. Investigating now....

bethbrains: Ahhh, I did merge to feature/rollups, but this branch got that because it's a child branch, and now we're merging this child branch into master directly. :/ Is that ok?

randi274: Once we work this out, we should remove feature/rollups to avoid this confusion. We've been discussing this in our standup meetings to avoid losing all of our child branches getting merged to master. @force2b any objections?

PR #3513, opened 2018-07-23
Feature/pdalcol/fixing duplicate default issue

### LINKED ISSUE #3434: Default GAU allocation amount is incorrectly calculated if there are multiple allocations to the default GAU in one Opportunity
### Steps to reproduce:

1. Enable default allocations and set a default GAU. 
2. Create a new Opportunity and set amount to $1000 (It will be automatically allocated to default GAU)
3. Click manage allocations and add an allocation for $500 (or any other amount) to the the default GAU (from step 1).
4. Click save and return to the list of allocations. There will be two allocations for $1000 each even thought the Opportunity total amount is $1000.
5. If you try editing the records using the standard UI, only one can be modified, the other one keeps getting reset to $1000. 

### Suggested solution: 

* Allow only 1 allocation to the default GAU on the same Opportunity.
* Detect and delete duplicate default GAU allocations, and combine all amounts into one allocation.

pedrodalcol:
# Critical Changes

# Changes
- NPSP now correctly calculates GAU Allocations if you assign more than one Allocation to the default GAU for an Opportunity.

# Issues Closed
Fixes #3434

# New Metadata

# Deleted Metadata


annejackson:
@lparrott  can you pass review on this again since we changes the branch name?

annejackson:
Merging.

PR #3532, opened 2018-07-28
Add option to specify day of month for Recurring Donations

### LINKED ISSUE #3455: RD's Next Pledged Opportunity has Different Close Date if Next Donation Date Blank and Last Payment Day within 28-31 Range
This issue is related to Issue #3126, but this provides a much tighter scenario/use case for it occurring. The issue is that future pledged Opportunities are being created based on the Date Established day of the month as opposed to the most recent Opportunity's Close Date day of the month when that day falls within the 28-31 of the month. It works as expected if the day falls within the 1-27 range of the month. Specifically, this occurs when the Next Donation Date is blank. This should only ever be the case if no future pledged Opportunities exist (this is dependent on the Opportunity Forecast Months setting in NPSP Settings). 

For example, if the Opportunity Forecast Months setting in NPSP Settings is 3 and I create a quarterly Open-Ended Recurring Donation with a Date Established of today, one pledged Opportunity will be created with a Close Date of today. If I then set that Opportunity's stage to Closed Won, the Next Donation Date field remains blank until the nightly RD script creates the next pledged Opportunity. 

Things get a bit more complicated when the day of the Date Established is different from the day of the Next  Donation Date (definitely a realistic scenario). 

* Scenario 1 (the control): the Date Established is 7/1/2018 and the Next Donation Date is 7/15/2018. When that Opportunity is set to Closed Won, the Next Donation Date will remain blank until the script creates the next one. In this case, the Close Date of the pledged Opportunity uses the date of the most recent Opportunity's Close Date: *10/15/2018*.

* Scenario 2 (the bug): the Date Established is 7/1/2018 and the Next Donation Date is 7/28/2018. When that Opportunity is set to Closed Won, the Next Donation Date will remain blank until the script creates the next one. In this case, the Close Date of the pledged Opportunity uses the date of the Date Established vs. the day of the most recent Opportunity's Close Date: *10/1/2018*.

**Steps to Repeat:**
We need to delete an Opportunity for each scenario to replicate exactly.

1. Set Opportunity Forecast Months to 3 in NPSP Settings → Recurring Donations → Recurring Donations.

2. Create the Scenario 1 (Control) data.
    1. Create a Contact.
    2. Create a new Recurring Donation (Contact = your Contact, Amount = 10, Open Ended Status = Open, Date Established = 6/1/2018, Next Donation Date = 6/1/2018, Installment Period = Quarterly).
    3. Two pledged Opportunities are created with Close Dates of 6/1/2018 and 9/1/2018.
    4. Delete the 9/1/2018 Opportunity. (so we can automatically add it later).
    5. Set the 6/1/2018 Stage to Closed Won.
    6. The Recurring Donation's Next Donation Date will be blank because there are no future pledged Opportunities.
    7. Click the Refresh Opportunities button on the Recurring Donation to simulate the nightly script running.
**_The newly created Opportunity will have a Close Date of 9/1/2018. [May need to refresh a few times if you're using Lightning...] This is created as expected._**

3. Create the Scenario 2 (Bug) Scenario:
    1. Create a new Recurring Donation (Contact = your Contact, Amount = 20, Open Ended Status = Open, Date Established = 6/1/2018, Next Donation Date = 6/28/2018, Installment Period = Quarterly).
    2. Two pledged Opportunities are created with Close Dates of 6/28/2018 and 9/28/2018.
    3. Delete the 9/28/2018 Opportunity. (so we can automatically add it later).
    4. Set the 6/28/2018 Stage to Closed Won.
    5. The Recurring Donation's Next Donation Date will be blank because there are no future pledged Opportunities.
    6. Click the Refresh Opportunities button on the Recurring Donation to simulate the nightly script running.
**_The newly created Opportunity will have a Close Date of 9/1/2018. [May need to refresh a few times if you're using Lightning...] We would expect a Close Date of 9/28/18._**

jjbennett:
# Critical Changes

# Changes
## Recurring Donations Enhancement
 - Users now have the option to specify which day of the month NPSP uses for installments on Recurring Donations. Add the Day Of Month field to your Recurring Donations page layout and enable field visibility for your user profiles. Day of Month sets the Close Date for installment Opportunities going forward when you populate the field.

# Issues Closed
- Fixes #3455 

# New Metadata
Custom Fields:
 - npe03__Recurring_Donation__c.Day_Of_Month__c

# Deleted Metadata


jjbennett:
@sam-knox Can you please review the changes in the changes section. I'm guessing we'll probably need to walk them through adding the field to the layout and updating field access or point them to docs?

jjbennett:
**lurch: attach W-026252

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYBrqAAH" target="_blank">W-026252</a>

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -11,12 +11,183 @@
         <fullName>Always_Use_Last_Day_Of_Month__c</fullName>
         <defaultValue>false</defaultValue>
         <externalId>false</externalId>
-        <inlineHelpText>When checked, all future monthly or quarterly donation installments will be set to the last day of the month, regardless of the actual date in the &quot;Next Donation Date&quot; field. This is ignored when the installment period is not set to Monthly or Quarterly.</inlineHelpText>
+        <inlineHelpText>Sets future monthly, custom monthly, or quarterly donation installments to the last day of the month, regardless of Next Donation Date. Installment period must be Monthly or Quarterly, or a custom installment period using months.</inlineHelpText>
```
force2b: @jjbennett Assuming this checkbox overrides the DayOfMonth field value (going by the above unit test), should that be noted in the InlineHelpText (if there's room)?

jjbennett: Yes Always always wins, @sam-knox thoughts on whether or not to include something in the help text and if so, what?

force2b: There's already some validation code in the `RD_RecurringDonations_TDTM` class (line 97) around the AlwaysUseLastDayOfMonth field. Do you think it's worth adding logic there to advise the User to not use the DayInMonth field when the AlwaysUseLastDayOfMonth field is checked?

jjbennett: @force2b - It doesn't hurt anything to have then both set, I think if we explain the hierarchy in the docs we shouldn't need to prevent the user from saving with a value in both fields. @mpusto thoughts?

force2b:
@rponti-sforg Is this fully tested? If so, can you mark as Approved?

PR #3553, opened 2018-08-01
Day of Month help text suggestion

watersbeck:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/objects/npe03__Recurring_Donation__c.object
+++ b/src/objects/npe03__Recurring_Donation__c.object
@@ -20,7 +20,7 @@
     <fields>
         <fullName>Day_of_Month__c</fullName>
         <externalId>false</externalId>
-        <inlineHelpText>Sets future monthly, custom monthly, or quarterly donation installments to the selected day of the month, regardless of Next Donation Date. Installment period must be Monthly or Quarterly, or a custom installment period using months.</inlineHelpText>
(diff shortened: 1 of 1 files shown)


watersbeck:
@sam-knox  I think the first sentence in the help text adequately describes the kind of installment periods that can be used with Day of Month. Seems helpful to include info about how Day of Month interacts with Last Day of Month.

sam-knox:
I'm almost at the character limit (max is 255) for the help text already. The text currently is:

"Sets future monthly, custom monthly, or quarterly donation installments to the selected day of the month, regardless of Next Donation Date. Installment period must be Monthly or Quarterly, or a custom installment period using months."

I could replace the second sentence with:

"The Always use last day of month field takes precedence over Day of Month."

I'm trying to decide if explaining last day of month is more important than making installment period settings clear. What do you think?

watersbeck:
Right, I was suggesting swapping out your last sentence for the one about
how the two fields interact. Your first sentence already includes the
information about what kind of installment periods the field applies to
("Sets future *monthly, custom monthly, or quarterly* ..")

sam-knox:
Ok, sorry I hadn't realized that you already had a suggested change in the file. Hence the PR of course! So yes, I think what you have is fine.

PR #3557, opened 2018-08-02
Feature/bge  active fields sorting

randi274:
# Critical Changes

# Changes
- Added active field sorting, storage, and loading.
# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
**lurch: attach W-025938

src/aura/BGE_TemplateDetails/BGE_TemplateDetailsController.js
```diff
@@ -43,6 +43,27 @@
...
+        model.getTemplateFields().moveFieldsDown(selectedRows);
+    },
+
+    /* ***************************************************************
+     * @Description. Updates the selected active fields in the
+     * Template Fields Model to available.
+     *****************************************************************/
+    moveFieldsUp: function (component, event, helper) {
+        var model = component.get('v.model');
+        var selectedRows = component.find("activeTreeGrid").getSelectedRows();
```
kofijohnson: @randi274 here, we don't need to pass the Selected Rows, because the model has them already. We can do something similar to:
`moveFieldsUp: function (component, event, helper) {
        var model = component.get('v.model');
        model.getTemplateFields().moveSelectedUp();
    }`

We did a similar logic in the controller method updateToActive:
`updateToActive: function (component, event, helper) {
        var model = component.get('v.model');
        model.getTemplateFields().updateSelectedToActive();
    }`

src/aura/BGE_TemplateDetails/BGE_TemplateDetailsController.js
```diff
@@ -43,6 +43,27 @@
 
     /******************************** Template Fields Controller Functions *****************************/
 
+
+    /* ***************************************************************
+     * @Description. Updates the selected active fields in the
+     * Template Fields Model to available.
+     *****************************************************************/
+    moveFieldsDown: function (component, event, helper) {
```
kofijohnson: Same comment as below.

src/aura/BGE_TemplateDetails/BGE_TemplateDetailsController.js
randi274: True! I think we were hoping to make use of attributes the tree grid adds, like posInSet, and setSize, such as on line 610 in the helper. Is there a good way around using that logic, or is that something we might need to add to the model?

src/aura/BGE_TemplateDetails/BGE_TemplateDetailsHelper.js
```diff
@@ -580,13 +580,28 @@
...
+                });
+
+                console.log(activeFieldCountBySObject);
+
                 _allFields.forEach(function(currentField) {
                     if (!currentField.isActive && currentField.selected) {
                         currentField.isActive = true;
                         currentField.selected = false;
-                        // TODO: this seems expensive. is there a better way?
-                        var activeFieldsBySObject = getActivesBySObject();
```
kofijohnson: @randi274 can we revert this back? I think the new performance improvement code is causing an issue.

src/aura/BGE_TemplateDetails/BGE_TemplateDetailsController.js
```diff
@@ -43,6 +43,25 @@
 
     /******************************** Template Fields Controller Functions *****************************/
 
+
+    /* ***************************************************************
```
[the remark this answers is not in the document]
randi274: @kofijohnson can you elaborate on how the description is incorrect here?

PR #3566, opened 2018-08-07
CCI: New task for Deploy_Dev_Nima to reset overrides on the BGE_Template object

force2b:
Create a new CCI task (`deploy_dev_nima`) that applies the New and Edit overrides on the BGE_Template custom object to point to the BGE_TemplateCreation component.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -47,6 +47,12 @@ tasks:
         options:
             path: dev_config/src
 
+    deploy_dev_nima:
+        description: Post install config for NimaCloud to revert BGE Template Overrides (Added Aug-2018)
+        class_path: cumulusci.tasks.salesforce.Deploy
+        options:
+            path: dev_config/src_nima_bge
+
     deploy_dev_config_managed:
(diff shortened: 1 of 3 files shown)


randi274:
@force2b is this build error something we need to be concerned about?

force2b:
No. I renamed the branch to make it a child of bge. Looks like that build failure is because it was looking for the old branch name.

force2b:
The namespaced build on the new branch name passed. The other build is almost done.

davisagli:
@force2b note that the code which finds child branches is case sensitive, so this branch would need to be named `feature/BGE__cci-new-nimacloud-task-for-overrides` instead of `feature/bge__cci-new-nimacloud-task-for-overrides` in order to be recognized as a child of `feature/BGE`.

force2b:
@davisagli Thanks. I can fix that, though ideally we'll merge into feature/bge before we have to worry about a sync into the child from bge.

PR #3605, opened 2018-08-16
Add a flow to deploy custom fields and rollups for testing account soft credits

davisagli:
This adds a set of unpackaged metadata with custom fields and rollups for testing account soft credits (retrieved from @dgamboni's scratch org) and a new flow which can be used to deploy them when desired.

To use it, first deploy NPSP using a standard flow like `dev_org`. Then run `cci flow run config_acctsc_rollups --org orgname`. This will enable customizable rollups and deploy the fields and rollups. But it won't take care of adding the fields to the Account page layout.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cdcarter:
Is it worth it to add the Account page layout to the config for this flow also?

davisagli:
I thought about including the Account page layout. Was worried that including it would override core NPSP page layouts leaving out any fields that get added after now.

cdcarter:
We'd have to keep this (and any other emergent configs) up to date. Likely requires some thought/strategy but I think the benefit of not having to drag like 12 fields onto the layout is good enough here.

dev_config/src_acctsc_rollups/package.xml
```diff
@@ -0,0 +1,38 @@
...
+        <members>Rollup.AcctSCNumSCLastYear7d2ae551</members>
+        <members>Rollup.AcctSCSCBestYear52a61eb6</members>
+        <members>Rollup.AcctSCSCBestYearTotal33464ad2</members>
+        <members>Rollup.AcctSCSmallestSCAmt526efbf3</members>
+        <members>Rollup.AcctSCTotalAmtSCs2YrsAgo56adfbf4</members>
+        <members>Rollup.AcctSCTotalAmtSCsd2d11dff</members>
+        <members>Rollup.ConFirstSoftCreditAmounta60c284d</members>
+        <name>CustomMetadata</name>
+    </types>
+    <version>39.0</version>
```
force2b: I suggest using API v43 to match the current api version. I have a separate PR bringing the rest of the code base up to this version.

cumulusci.yml
```diff
@@ -194,6 +200,18 @@ flows:
             7:
                 task: test_data_relationships
 
+    config_acctsc_rollups:
```
force2b: Since the Account Soft Credit feature has been merged into master, is there still a need for a flow specific to testing it? 

Though, I really like the idea of a task that creates the AccountSoftCredit fields and rollups because I think we can expand on this to create Payment and AccountContactSoftCredit rollups and target fields for use with the CRLP Validation Script.

davisagli: You may be right that it's no longer needed, but a) it'll come in handy for testing any adjustments/bug fixes to account soft credits in the future, and b) it serves as an example of how to create a flow for deploying unpackaged metadata specific to testing one feature.

dev_config/src_acctsc_rollups/package.xml
davisagli: Sure; I updated it.

davisagli:
@force2b automating those adjustments to buttons sounds like a good idea, but probably belongs in the standard dev org config rather than in a feature-specific metadata package, right? I'd prefer to tackle that as a separate work item.

jjbennett:
@davisagli @force2b I would suggest not putting the button fixes in the standard config but instead in the qa config. I'd want it to be separate to only use when needed, otherwise I'd worry about retrieving them when working on an item.

davisagli:
Thanks @jjbennett, that makes sense.

PR #3606, opened 2018-08-16
Incremental Rollup Settings

kofijohnson:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@kofijohnson Shouldn't this PR merge into `feature/RollupsIncrementalBatch` rather than `master`?

kofijohnson:
@force2b my goal is we first merge feature/RollupsIncrementalBatch first into master (and mrbelvedere will merge back to feature/RollupsIncrementalBatch__UI). Then we merge feature/RollupsIncrementalBatch__UI (expose the settings to the users) probably the next sprint. 

The goal of merging first the feature/RollupsIncrementalBatch is to be able to do the test with big org like the NRDC before we make the settings available to all our customer.

force2b:
That makes sense, however this PR should still point to its parent `feature/RollupsIncrementalBatch` branch. Otherwise, if you look at the list of changed files it's impossible to determine which changed items relate to the UI. In the end, this is still a child of `feature/RollupsIncrementalBatch` even if you're going to merge it after a release of the main functionality. 

Also, keep in mind that the next push to production is not until 9/18. Is there a plan to test the main incremental rollups functionality as a beta package inside an NRDC Sandbox earlier?

njjc:
@kofijohnson @force2b are we going to try to get this in?

force2b:
@njjc @kofijohnson I don't think we can squeeze this in at the moment. If I remember this correctly there are still some outstanding issues that needed to be resolved for this to work. If anything, I would say we should write up document detailing what is necessary to bring this across the finish line and create the necessary work items to do so. 

Also, there are two open PR's on this which, which I think need to be merged into a single branch/PR.

njjc:
Sounds good. @kofijohnson could you merge the two into a single branch and document any outstanding items? Thanks!

PR #3623, opened 2018-08-21
SAL-727

ricardonimacloud:
# Critical Changes

# Changes
- Added Dropdown editor customization for :
	-- When dropdown editor is opened, the textarea is disabled for avoid manual editing
	-- When dropdown editor is closed, the textarea is enabled for allow autofill or copy/paste editing

- Removed old code used before this solution
# Issues Closed
SAL-727
# New Metadata

# Deleted Metadata


mdeleon0:
Old PR

PR #3632, opened 2018-09-08
NPSP Feature Telemetry

force2b:
This pull request implements the collection of feature and usage metrics from our customers into the Salesforce.org License Management Org. This information provides valuable insight into how the NPSP product is used by our customers. No personally identifiable information is collected through Feature Telemetry. Only three data types are collected: Boolean (True/False), Integer, and Date. At this point, only Boolean and Integer data types are being collected. 

Feature Telemetry data is collected as part of the normal push-upgrade process that happens automatically every two weeks. The only exception is that the "Is Customizable Rollups Enabled" data is collected when Customizable Rollups is either enabled or disabled.

**The following data points are collected per organization.**
* Is the Household Account Model Enabled
* Is Customizable Rollups Enabled
* Is Automatic Payments Enabled
* Is Default GAU Enabled
* Is Address Verification Enabled
* Is Advanced Current Management In Use
* Count of Non-English Languages In Use
* Count of all Recurring Donation related Opportunities
* Count of all Open-Ended Recurring Donation related Opportunities
* Count of NPSP Error Log Records
* Maximum Number of Opportunities associated with a single Account (Data-Skew)
* Are Custom TDTM Handlers in Use

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
** lurch: attach W-028312

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+     **/
+    @TestSetup
+    static void testDataSetup() {
+        Contact c = UTIL_UnitTestData_TEST.getContact();
+        insert c;
+
+        c = [SELECT Id, FirstName, LastName, AccountId FROM Contact LIMIT 1];
+        Id acctId = c.AccountId;
+
+        String closedStage = UTIL_UnitTestData_TEST.getClosedWonStage();
```
RadSFDeveloper: Do we need a variable for this since the closedStage is used only once.

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+
+        Opportunity opp = new Opportunity (
+                Name = 'Test Opp ' + c.FirstName + ' ' + c.LastName,
```
RadSFDeveloper: Indentation.

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * @description Test that the NPSP_Settings telemetry data is retrieved.
+     * It's not possible to validate that the call to System.FeatureManagement() call succeeded.
+     */
+    static testMethod void test_telemetry_NPSP_Settings() {
+        Test.startTest();
+        UTIL_OrgTelemetry_SVC.processNPSPSettingTelemetry_ASync();
+        Integer currQueryCount = Limits.getQueries();
+        UTIL_OrgTelemetry_SVC telemetrySvc = new UTIL_OrgTelemetry_SVC();
```
RadSFDeveloper: Let's use vertical separation so it is easier to read the related code grouped together.

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+                Primary_Contact__c = c.Id
+        );
+        insert opp;
+    }
+
+    /**
+     * @description Test that the NPSP_Settings telemetry data is retrieved.
+     * It's not possible to validate that the call to System.FeatureManagement() call succeeded.
+     */
+    static testMethod void test_telemetry_NPSP_Settings() {
```
RadSFDeveloper: Let's use `@isTest` and hopefully remove `test_` and underscores from the test method names?

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+        Test.startTest();
+        UTIL_OrgTelemetry_SVC.processNPSPSettingTelemetry_ASync();
+        Integer currQueryCount = Limits.getQueries();
+        UTIL_OrgTelemetry_SVC telemetrySvc = new UTIL_OrgTelemetry_SVC();
+        telemetrySvc.processTelemetryType(UTIL_OrgTelemetry_SVC.TelemetryTypeEnum.NPSP_Settings);
+        System.Assert(Limits.getQueries() == currQueryCount, 'There should no new queries executed because these only reference static custom setting');
```
RadSFDeveloper: Let's use `System.assertEquals(Limits.getQueries(), currQueryCount, '...');` so if there is an error, we know what difference it was.

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * @description Test that the MaxNumRelatedOpps settings telemetry data is retrieved.
+     * that the call to System.FeatureManagement() call succeeded.
+     */
+    static testMethod void test_telemetry_Data_MaxNumRelatedOpps() {
+        Test.startTest();
+        Integer currQueryCount = Limits.getQueries();
+        UTIL_OrgTelemetry_SVC telemetrySvc = new UTIL_OrgTelemetry_SVC();
+        telemetrySvc.processTelemetryType(UTIL_OrgTelemetry_SVC.TelemetryTypeEnum.Data_MaxNumRelatedOpps);
+        System.AssertEquals(currQueryCount+1, Limits.getQueries(), 'There should be one new query executed');
```
RadSFDeveloper: Typo `System.AssertEquals` should be `System.assertEquals`.

src/classes/UTIL_OrgTelemetry_TEST.cls
RadSFDeveloper: Suggestion: It can just called `service`.

src/package.xml
```diff
@@ -2324,9 +2327,23 @@
...
+        <members>IsEnabled_AddressVerification</members>
+        <members>IsEnabled_AutomaticPayments</members>
         <members>IsEnabled_CustomizableRollups</members>
+        <members>IsEnabled_DefaultGAU</members>
+        <members>IsEnabled_HouseholdAcctModel</members>
+        <members>Org_IsEnabled_AdvancedMultiCurrency</members>
         <name>FeatureParameterBoolean</name>
     </types>
+    <types>
+        <members>Data_Count_of_ErrorLog</members>
```
RadSFDeveloper: What is the naming convention to use `Data_` or `Org_` for the parameter?
Suggestion, parameters can be read the same or better if not using underscores. For example:
```
<members>ErrorLogCount</members>
<members>AllRecurringDonationOppCount</members>
<members>OpenEndedRecurringDonationOppCount</members>
<members>MaxNumRelatedOpps</members>
<members>Org_NonEnglishLanguageCount</members>
```

src/package.xml
```diff
@@ -2324,9 +2327,23 @@
         <name>EmailTemplate</name>
     </types>
     <types>
+        <members>HasUserManagedTDTM</members>
+        <members>IsEnabled_AddressVerification</members>
+        <members>IsEnabled_AutomaticPayments</members>
         <members>IsEnabled_CustomizableRollups</members>
+        <members>IsEnabled_DefaultGAU</members>
+        <members>IsEnabled_HouseholdAcctModel</members>
+        <members>Org_IsEnabled_AdvancedMultiCurrency</members>
```
RadSFDeveloper: What is naming convention for the `FeatureParameterBoolean`?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,218 @@
...
-    * This should called in two places: 1) When the Settings page is loaded; 2) When a tracked area is saved.
-    * FeatureManagement operations are subject to MIXED_DML errors and must be conducted asynchronously
-    */
+     * @description A complete list of the Telemetry Types (at a high level) that are handled by this service class.
+     * These are used by the batch job to process feature parameters in asynchronous groups to avoid governor
+     * limit issues. The concept is that by separating each of these into smaller groupings that are processed
+     * as separate child batch iterations, the entire process can complete even if for some reason one of the queries
+     * against a table (such as a really huge Opportunity object) were to fail.
+     */
+    public Enum TelemetryTypeEnum {
```
RadSFDeveloper: No need for the Enum in the variable name.

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -0,0 +1,82 @@
...
+            UTIL_OrgTelemetry_SVC.TelemetryTypeEnum.Data_CountOfErrorLog
+        };
+
+    /**
+     * @description Return the list of Telemetry Type Enum values to be processed by this batch job
+     * @param BC BatchableContext (not used)
+     * @return telemetryTypesList list from above
+     */
+    public Iterable<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum> start(Database.BatchableContext BC) {
+        if (!Test.isRunningTest()) {
```
RadSFDeveloper: Suggestion:
```

return !Test.isRunningTest()
    ? UTIL_OrgTelemetry_BATCH.telemetryTypesList
    : new List<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum>{UTIL_OrgTelemetry_SVC.TelemetryTypeEnum.Data_CountOfErrorLog};
```

src/classes/UTIL_OrgTelemetry_Batch_TEST.cls
```diff
@@ -0,0 +1,51 @@
...
+* @author Salesforce.org
+* @date 2018
+* @group Utilities
+* @description Unit Tests related to the org telemetry class
+*/
+@isTest(IsParallel=false)
+private class UTIL_OrgTelemetry_Batch_TEST {
+
+    /**
+     * @description Submits the telemetry batch job to execute. This will only execute a single child iteration
```
RadSFDeveloper: Great comments in the description of all methods!

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,218 @@
...
+        Data_CountOfErrorLog
+    }
+
+    /**
+     * @description Execute the batch job that will process each telemetry featureParameter type one at a time
+     * as individual batch iterations. This process has to be asynchronous to handle governor limits against
+     * huge queries as well as to avoid MixedDML errors in case a calling process were to modify a record.
+     */
+    public static void dispatchTelemetryBatchJob() {
+        database.executeBatch(new UTIL_OrgTelemetry_BATCH(), 1);
```
RadSFDeveloper: Typo, it should be `Database`.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -213,7 +213,8 @@ public without sharing class UTIL_OrgTelemetry_SVC {
      */
     private Integer getCountOfRecurringDonationOpps() {
         try {
-            return [SELECT Count() FROM Opportunity WHERE npe03__Recurring_Donation__c != null];
+            return [SELECT Count() FROM Opportunity
```
RadSFDeveloper: Can we have a comment indicating the query was written in this specific way to work on LDV orgs?

src/classes/CMT_MetadataAPI.cls
```diff
@@ -163,7 +163,7 @@ public class CMT_MetadataAPI {
                 upsert crlpSettings;
             }
             if (isSuccess) {
-                UTIL_OrgTelemetry_SVC.submitFeatureTelemetryToLMO();
+                UTIL_OrgTelemetry_SVC.processNPSPSettingTelemetry_ASync();
```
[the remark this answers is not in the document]
force2b: Would you prefer that as `processNPSPSettingTelemetry_Async()`?

src/classes/STG_InstallScript.cls
```diff
@@ -60,9 +60,14 @@ global without sharing class STG_InstallScript implements InstallHandler {
             executeAction(InstallAction.RUN_NEW_ORG_SCRIPT, logger);
 
         } else if (context.isUpgrade() || context.isPush()) {
+            // Ensure the any new Custom Settings objects are populated
```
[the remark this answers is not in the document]
force2b: Yeah, but I really do like commented code. I agree that code should speak for itself, and I think it probably does, but at the same time the comment that says "ensure any *new* custom settings ..." is important to understand why we're doing this here. I don't know if moving it up to the method apexdoc comment adds a lot of value, though I suppose it's possible. 

Part of me says there can never be too many comments in code.

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -0,0 +1,82 @@
...
+* one query were to fail, which is theoretically possible in a huge organization where there may be other jobs
+* executing or other factors at play, the rest of this job will continue. That ensures that our LMO collects as much
+* data as possible. In addition, it can help us identify which specific telemetry data processes are failing.
+*/
+public without sharing class UTIL_OrgTelemetry_BATCH implements Database.Batchable<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum> {
+
+    /**
+     * @description List of High Level Telemetry Data Types to be processed asynchronously by this batch job.
+     */
+    private static final List<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum> telemetryTypesList =
```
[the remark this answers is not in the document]
force2b: Very good suggestions.

src/classes/UTIL_OrgTelemetry_BATCH.cls
force2b: I'm a big fan of ternary `if` as well, but do you think that's getting too long to be readable? Though, does making the change add anything or address a specific issue in the current `if` block?

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @description Process the specific set of FeatureParameter submissions to the LMO based on the Telemetry
+     * Type Enum value.
+     * @param info BatchableContext (not used)
+     * @param types List<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum> -- Batch Size of ONE is required
+     */
+    public void execute(Database.BatchableContext info, List<UTIL_OrgTelemetry_SVC.TelemetryTypeEnum> types) {
+        UTIL_OrgTelemetry_SVC telemetrySvc = new UTIL_OrgTelemetry_SVC();
+        telemetrySvc.processTelemetryType(types[0]);
```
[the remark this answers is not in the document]
force2b: It's technically possible to check for a batch size > 1 in the execute method, however even as a public method it can only be called by managed code. I'm not sure it's worth adding validation logic to this given that that we're only protecting against our own usage and that should be covered in the class comments.

src/classes/UTIL_OrgTelemetry_Batch_TEST.cls
```diff
@@ -0,0 +1,51 @@
...
+     * and not the entire set of data. It's not possible to determine if the job actually completed.
+     * It's not possible to validate that the call to System.FeatureManagement() call succeeded.
+     * This specific test needs to be in a separate test class to avoid an internal error that is thrown when
+     * testing along side other methods that call the same featureParameter value used by this test.
+     */
+    static testMethod void test_telemetryBatchJob() {
```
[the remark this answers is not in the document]
force2b: I suppose so, though I almost always name all of my unit test methods to start with `test_`. I don't think this is something we've discussed in the past, but I can see your point. It's a test, so why name it to start with `test_`. It's very logical.

I think we should discuss adding that to the styles doc with the rest of the team. I need to schedule a follow-up meeting on that topic anyway.

src/classes/UTIL_OrgTelemetry_Batch_TEST.cls
```diff
@@ -0,0 +1,51 @@
...
+        Test.startTest();
+        UTIL_OrgTelemetry_SVC.dispatchTelemetryBatchJob();
```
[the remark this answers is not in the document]
force2b: It doesn't.  It can run a single batch iteration that makes no dml changes. I don't believe there is anything to assert against.

src/classes/UTIL_OrgTelemetry_SVC.cls
force2b: Agreed

src/package.xml
force2b: There isn't necessarily a naming convention in place. I've tried to stick to something somewhat logical based on the type of data (IsEnabled, Org_IsEnabled, Data, etc.), but to be honest it's not really visible anywhere other than our code.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,218 @@
...
+     * @param telemetryType
+     */
+    public void processTelemetryType(TelemetryTypeEnum telemetryType) {
+
+        switch on (telemetryType) {
+
+            // Simple Boolean FeatureParameters based strictly on NPSP Settings that can be processed in a single
+            // operation.
+            when NPSP_Settings {
+                STG_SettingsService stgSvc = new STG_SettingsService();
```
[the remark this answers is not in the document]
force2b: Good point. I had already split out the logic to query the data, so those same methods can handle the call to System.SetFeatureParameter as well.

src/classes/CMT_MetadataAPI.cls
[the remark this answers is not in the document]
RadSFDeveloper: That one or renaming it just as `asyncProcessNpspSettings()`. We already know it is telemetry since it is used in the `UTIL_OrgTelemetry_SVC`, but using `asyncProcessNpspSettingsTelemetry()` is fine too.

src/classes/STG_InstallScript.cls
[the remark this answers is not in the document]
RadSFDeveloper: I think execution actions are clear on their own and the comments do not specify any exception/workaround handling. If a developer needs to know more "why", they can read the method for the specific parameter, or view the change history on the file to track the PR to the work item and read more details if needed. My two cents, decision to leave comments or remove them is yours. :)

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -0,0 +1,82 @@
...
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_MaxNumRelatedOpps,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_Count_of_RD_Opps_All,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_Count_of_RD_Opps_OpenEnded,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_CountOfErrorLog
+        };
+
+    /**
+     * @description Return the list of Telemetry Type Enum values to be processed by this batch job
+     * @param BC BatchableContext (not used)
+     * @return telemetryTypesList list from above
```
RadSFDeveloper: The argument needs to be updated to `Iterable<UTIL_OrgTelemetry_SVC.TelemetryType>`

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,219 @@
...
+     * @description Count the number of DatedConversionRate records with an effective date in the last 180 days.
+     * If greater than 0, then consider AdvancedMultiCurrency to be enabled within an organization. This accounts for
+     * the fact that this object is always visible, but may not actually have any valid/recent data in it; and because
+     * there is no plan for core to ever fix this bug.
+     */
+    private void handleIsAdvancedMulticurrencyEnabled() {
+        Boolean isEnabled=false;
+        if (UserInfo.isMultiCurrencyOrganization() == true) {
+            try {
+                Integer DCRCount = Database.query('SELECT Id FROM DatedConversionRate WHERE StartDate = LAST_N_DAYS:180 LIMIT 1').size();
```
RadSFDeveloper: `DCRCount` can be just `count` (lowerCamelCase for the variable names). The method is small and it is clear what the count represents. If more info is needed in the variable name, it can be `rateCount`?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,219 @@
...
+            System.FeatureManagement.setPackageIntegerValue('Data_Count_of_ErrorLog', count);
+        } catch (Exception ex) {}
+    }
+
+    /**
+     * @description Total number of Opportunities associated with a Recurring Donation
+     */
+    private void handleCountOfRecurringDonationOpps() {
+        try {
+            Integer count = [SELECT Count() FROM Opportunity
```
RadSFDeveloper: Is it possible to re-write SOQLs in this class by the latest coding style?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,219 @@
...
+    }
+
+    /**
+     * @description Get the max number of attached opportunities (using by 'Total Number of Donations') for
+     * all Accounts. Note that this will not return useful information if the organization is not using
+     * NPSP rollups (either type)
+     */
+    private void handleMaxNumRelatedOpps() {
+        try {
+            Account acct = [SELECT npo02__NumberOfClosedOpps__c
```
RadSFDeveloper: Let's decrease the indentation:
```
Account acct = [
    SELECT npo02__NumberOfClosedOpps__c
    FROM Account
    WHERE npo02__NumberOfClosedOpps__c > 0
    ORDER BY npo02__NumberOfClosedOpps__c DESC
    LIMIT 1];
```

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,143 @@
...
+        );
+        insert opp;
+    }
+
+    /**
+     * @description Test that the NPSP_Settings telemetry data is retrieved.
+     * It's not possible to validate that the call to System.FeatureManagement() call succeeded.
+     */
+    @IsTest
+    static void test_telemetry_NPSP_Settings() {
```
RadSFDeveloper: If not for this PR, let's determine the way we name test methods and hopefully remove underscores.

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,143 @@
...
+            Primary_Contact__c = c.Id
+        );
+        insert opp;
+    }
+
+    /**
+     * @description Test that the NPSP_Settings telemetry data is retrieved.
+     * It's not possible to validate that the call to System.FeatureManagement() call succeeded.
+     */
+    @IsTest
```
RadSFDeveloper: Can we use the vertical distance for readability? For example:
```
@IsTest
    static void test_telemetry_NPSP_Settings() {
       Test.startTest();
        UTIL_OrgTelemetry_SVC.processNPSPSettingTelemetry_ASync();

        Integer expectedCount = Limits.getQueries();
        UTIL_OrgTelemetry_SVC service = new UTIL_OrgTelemetry_SVC();

        service.processTelemetryType(UTIL_OrgTelemetry_SVC.TelemetryType.NPSP_Settings);
        System.assertEquals(Limits.getQueries(), expectedCount, 'There should no new queries executed because these only reference static custom setting');
        Test.stopTest();
}
```

src/package.xml
RadSFDeveloper: Can we remove underscores from the names?

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -0,0 +1,143 @@
...
+    static void test_telemetry_NPSP_Settings() {
+        Test.startTest();
+        UTIL_OrgTelemetry_SVC.processNPSPSettingTelemetry_ASync();
+        Integer currQueryCount = Limits.getQueries();
+        UTIL_OrgTelemetry_SVC telemetrySvc = new UTIL_OrgTelemetry_SVC();
+        telemetrySvc.processTelemetryType(UTIL_OrgTelemetry_SVC.TelemetryType.NPSP_Settings);
+        System.assertEquals(Limits.getQueries(), currQueryCount, 'There should no new queries executed because these only reference static custom setting');
```
RadSFDeveloper: The `System.assertEquals()` should take the expected value as the first argument.

RadSFDeveloper: Should the `System.assertEquals()` be after the `Test.stopTest()` because of the async processing?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -33,21 +33,219 @@
...
+    /**
+     * @description Called by the batch job to process an individual telemetry type
+     * Submits specific feature teletry statistics back to the License Management Org to assist
+     * with understanding how our customers are using NPSP functionality.
+     * Each feature below must be matched to a SubscriberToLmo feature of the same name.
+     * @param telemetryType
+     */
+    public void processTelemetryType(TelemetryType telemetryType) {
+
+        switch on (telemetryType) {
```
RadSFDeveloper: Great use of the `switch`, the method is readable after using separate methods for each case.

src/classes/STG_InstallScript.cls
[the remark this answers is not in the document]
force2b: Excellent points. Some comments can definitely be self-serving; and good code along with apexdoc comments in the method should cover most needs. That said, I do like to see code comments to explain usage. In this case, I had added the comment to reflect that fact that a new action was needed to ensure that any newly created custom settings were populated. I didn't think that was specifically addressed by the method call. So, if you're OK with it, I'd like to leave this comment in.

src/classes/CMT_MetadataAPI.cls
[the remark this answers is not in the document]
force2b: Agreed. As a habit, I've been using the `_Async` suffix on my asynchronous methods to differentiate them from non-async methods of the same name. That works nicely in cases where I had to have two different entry points for an action. However, that's definitely not the case here, so I'll change the method name.

src/package.xml
force2b: We're already committed to `IsEnabled_CustomizableRollups` as a FeatureParameter name, so at this point I think the underscores are simply part of the naming pattern. I'd prefer not to have one with an underscore and the rest without.

src/classes/UTIL_OrgTelemetry_SVC.cls
force2b: Yes, and thanks for holding me to that. It's easy to forget that we have some preferred styles, though at the same time I don't really think there's anything wrong with having the FROM on the same line as the SELECT if there's only a single field.

src/classes/UTIL_OrgTelemetry_TEST.cls
force2b: I was thinking about this more last night. I think what bothers me is a test method with the name of `telemetry_NPSP_Settings`. It strikes me as missing something. I can see your point about "test_" though. What do you think about `validate_Telemetry_NPSP_Settings`?

RadSFDeveloper: `validateNpspSettingsTelemetry` or `validateNpspSettingsTelemetryIsRetrieved`. The latter has expected result in the name. Both names are not too long but they are still descriptive.

src/classes/UTIL_OrgTelemetry_TEST.cls
force2b: Thanks. I still get that backwards.

force2b: The `stopTest` resets the query count, so there really isn't a way to validate the queries after the stopTest for the async transaction. The assert before the stopTest validates the call above to `processTelemetryType`.

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @description List of High Level Telemetry Data Types to be processed asynchronously by this batch job.
+     */
+    private final List<UTIL_OrgTelemetry_SVC.TelemetryType> telemetryTypes =
+        new List<UTIL_OrgTelemetry_SVC.TelemetryType>{
+            UTIL_OrgTelemetry_SVC.TelemetryType.NPSP_Settings,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Org_Environment,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_MaxNumRelatedOpps,
+            UTIL_OrgTelemetry_SVC.TelemetryType.Data_Count_of_RD_Opps_All,
```
RadSFDeveloper: To follow the format of the variables before and after these, can we name them to remove underscores and to remove "of"?
```
    UTIL_OrgTelemetry_SVC.TelemetryType.Data_CountRdOppsAll,
    UTIL_OrgTelemetry_SVC.TelemetryType.Data_CountRdOppsOpenEnded,
    UTIL_OrgTelemetry_SVC.TelemetryType.Data_CountErrorLog
```

force2b: Yes. That works.

src/package.xml
```diff
@@ -2325,9 +2328,23 @@
...
         <name>FeatureParameterBoolean</name>
     </types>
+    <types>
+        <members>Data_Count_of_ErrorLog</members>
+        <members>Data_Count_of_RD_Opps_All</members>
+        <members>Data_Count_of_RD_Opps_OpenEnded</members>
+        <members>Data_MaxNumRelatedOpps</members>
+        <members>Org_Count_of_NonEnglishLanguages</members>
```
RadSFDeveloper: ` <members>Org_CountNonEnglishLanguages</members>`?

force2b:
** lurch: attach W-028377

PR #3633, opened 2018-09-08
Deactivate rollup triggers in BDI when rollup settings is not checked

### LINKED ISSUE #3630: NPSP Data Import Still Calculating Customizable Rollups When Calculate Donation Rollups with Batch is Unchecked in Managed NPSP
When the “Calculate Donation Rollups With Batch” field is unchecked (the default) on the NPSP Data Import configuration screen, rollups still fire when Customizable Rollups are enabled. 

**_Note: this works as expected in unmanaged NPSP, so when testing, please install the managed version of NPSP._**

**Steps to Repeat:**

1. Go to NPSP Settings → Donations → Customizable Rollups and enable Customizable Rollups.

2. If using an org with unmanaged NPSP, modify the Start Data Import button (Setup →Object Manager → NPSP Data Import → Buttons, Links, and Actions → Edit Start Data Import button. Copy/paste the following URL instead.
`/apex/BDI_DataImport?retURL={!URLFOR($Action.DataImport__c.Tab, $ObjectType.DataImport__c)}
`
3. Create a record in the NPSP Data Imports object (sample file attached to Work Item). The key is to have Donation Information which will trigger the rollups to fire.

4. Go to the NPSP Data Imports Tab and switch your View to “To Be Imported”

5. Click the Start Data Import button

6. Verify that “Calculate Donation Rollups With Batch” is unchecked (the default).

7. Start the import.

8. When the import completes, go to the newly created Contact and you'll see that rollup fields have been calculated. 

In contrast, legacy rollups do respect this setting. If testing in a scratch org with unmanaged NPSP, this setting is also respected.

RadSFDeveloper:
Rollups are not calculated upon data import when Data Import Settings > "Calculate Donation Rollups with Batch" field is unchecked.

# Critical Changes

# Changes

# Issues Closed
Fixes #3630

# New Metadata
Apex Class:
- BDI_DataImportService_TEST

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -763,23 +763,7 @@ global with sharing class BDI_DataImportService {
             return;
         }
 
-        List<Trigger_Handler__c> handlers = TDTM_Config_API.getCachedRecords();
-        for (Trigger_Handler__c th : handlers) {
-            if (isRollupTriggerHandler(th)) {
-                th.Active__c = false;
-            }
-        }
-    }
-
-    /*******************************************************************************************************
-    * @description Returns true if the trigger handler is used to calculate rollups
-    * @param handler The trigger handler
-    * @return Boolean
-    */
-    private Boolean isRollupTriggerHandler(Trigger_Handler__c handler) {
-        return (handler.Object__c == 'Opportunity' && handler.Class__c == 'RLLP_OppRollup_TDTM') ||
-            (handler.Object__c == 'Opportunity' && handler.Class__c == 'CRLP_Rollup_TDTM') ||
-            (handler.Object__c == 'npe01__OppPayment__c' && handler.Class__c == 'CRLP_Rollup_TDTM'); 
+        TDTM_Config_API.disableAllRollupTriggers();
     }
 
     /*******************************************************************************************************
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -94,7 +94,7 @@ global with sharing class BDI_DataImportService {
     * @return Data_Import_Settings__c
     */
     public static Data_Import_Settings__c loadSettings(Id batchId) {
-
+        
         // load up the default or specific settings
         if (batchId == null) {
             return UTIL_CustomSettingsFacade.getDataImportSettings();
@@ -285,7 +285,7 @@ global with sharing class BDI_DataImportService {
     * @description holds the data import settings to use for this batch.  these may be either from
     * custom settings, or if a specific BatchId was specified, the settings specified for that batch.
     */
-    global Data_Import_Settings__c diSettings { get; private set; }
+    global Data_Import_Settings__c diSettings { get; set; }
 
     /*******************************************************************************************************
     * @description whether the processing should be done in Dry Run mode (vs. normal mode).
@@ -622,7 +622,6 @@ global with sharing class BDI_DataImportService {
     * @return void
     */
     public void LogBDIError(DataImport__c di, String strError, String strStatusField) {
-        //system.debug('****DJH: LogBDIError di: ' + di + ' strError: ' + strError + ' field: ' + strStatusField);
         di.Status__c = BDI_DataImport_API.bdiFailed;
         di.FailureInformation__c = strError;
         if (strStatusField != null)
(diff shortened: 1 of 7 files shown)


RadSFDeveloper:
**lurch: attach W-028042

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fNGIAA2" target="_blank">W-028042</a>

src/classes/BDI_DataImportService_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>37.0</apiVersion>
```
RadSFDeveloper: @force2b , the new class has the version as the BDI_DataImport2_TEST from which a unit test has been moved into. Shall I change it to version 42.0?

src/classes/BDI_DataImportService.cls
```diff
@@ -748,17 +758,30 @@ global with sharing class BDI_DataImportService {
...
-            for (Trigger_Handler__c th : handlers) {
-                if (th.Object__c == 'Opportunity' && (th.Class__c == 'RLLP_OppRollup_TDTM' || th.Class__c == 'CRLP_Rollup_TDTM')) {
-                    th.Active__c = false;
-                }
+    @TestVisible private void disableAllOppRollups() {
+        if (diSettings.Run_Opportunity_Rollups_while_Processing__c) {
+            return;
+        }
+
+        List<Trigger_Handler__c> handlers = TDTM_Config_API.getCachedRecords();
```
force2b: I think this should just call `TDTM_Config_API.disableAllRollupTriggers()` here this way we have a single point to handle disabling all rollup related triggers. Though, that method can remove the line that disables the Allocation__c trigger for CRLP_Rollup_TDTM since that doesn't exist any more.

src/classes/BDI_DataImportService_TEST.cls-meta.xml
force2b: This should go to v43 since it's a new class.

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -0,0 +1,142 @@
...
+
+        System.assertEquals(expectedHandlers, TDTM_Config_API.getCachedRecords());
+    }
+
+    /*******************************************************************************************************************
+    * @description Verifies rollup trigger handlers are deactivated if the data import settings to process rollups
+    * is not checked.
+    */
+    @isTest
+    private static void rollupTriggersAreDeactivatedWhenRunRollupSettingsIsNotChecked() {
```
force2b: Nice test method

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -0,0 +1,142 @@
...
+        diSettings.Donation_Matching_Behavior__c = BDI_DataImport_API.ExactMatchOrCreate;
+
+        return diSettings;
+    }
+
+
+    /*******************************************************************************************************************
+    * @description Returns true if the trigger handler calculates rollups
+    */
+    private static Boolean isRollupTriggerHandler(Trigger_Handler__c handler) {
```
force2b: If the `TDTM_Config_API.disableAllRollupTriggers()` method is used to disable all the rollup triggers, this method will likely need to be modified to match the same logic.

src/classes/BDI_DataImportService.cls
```diff
@@ -337,6 +337,17 @@ global with sharing class BDI_DataImportService {
...
     }
 
+    /*******************************************************************************************************
+    * @description Sets diSettings based on the input settings. The method is called from tests only 
+    * since the variable is global with private set().
+    * @param diSettings Data Import Settings
+    * @return void
+    */
+    @TestVisible
+    private void injectDataImportSettings(Data_Import_Settings__c settings) {
```
force2b: Rather than implement this new pattern for settings in tests, does it make sense to create a new method for `configDataImportSettingsForTests()` in the `UTIL_CustomSettingsFacade` class? That would match how all other custom settings are handled in unit tests.

src/classes/BDI_DataImportService_TEST.cls
RadSFDeveloper: The helper test method does not need to be changed since it specifically verifies which trigger handlers should be deactivated. If the `TDTM_Config_API.disableAllRollupTriggers()` is changed to remove one of those triggers or add new, the `BDI_DataImportService_TEST` will fail, so developer will know additional changes and unit/QA tests need to be done.

src/classes/BDI_DataImportService.cls
RadSFDeveloper: `diSettings` variable is only set in the `process()` method, which means test class requires an integration test just to verify that rollup triggers are disabled. Because we do not retrieve the `diSettings` from DB in the BDI class, the `UTIL_CustomSettingsFacade` usage would not work. If we inject the `diSettings`, we need *unit tests* to prove requirements by changing what settings are. The integration tests are required, but they can prove the "happy path" and rollback on error works fine. 
On the topic of `UTIL_CustomSettingsFacade`, it seems the class contains various configurations for settings. It is long, so to improve readability and extension, we might consider splitting it into specific domain custom settings (if and where required). But that is a subject for another day (or a work item :)).

RadSFDeveloper:
@jesslopez-sf , @force2b : Changes have been done to use `TDTM_Config_API.disableAllRollupTriggers()` in order to disable rollup triggers.  Please can you review again? Thank you.

src/classes/BDI_DataImportService.cls
force2b: Fair enough. Settings is definitely used in a different way in this class. 
You're right about the size and complexity of the `UTIL_CustomSettingsFacade` class, but that's a story for a different story. :)

RadSFDeveloper:
@rponti-sforg , the PR is ready for your testing using the branch feature/radmila-prevent-rollup-calc-unchecked-settings-managed-npsp

PR #3639, opened 2018-09-12
Style fixes on BDI_DataImportService.cls

lmeerkatz:
# Critical Changes

# Changes
Updates BDI_DataImportService.cls to follow Apex style guidelines. Adds braces, corrects capitilization, etc. Does not attempt to fix semantic issues.

# Issues Closed

# New Metadata

# Deleted Metadata


lmeerkatz:
@force2b I made this to give myself experience with the branching/PR process (and because fixing style issues makes me happy). No rush on reviewing. :)

src/classes/BDI_DataImportService.cls
```diff
@@ -77,13 +77,13 @@ global with sharing class BDI_DataImportService {
     * @param batchId id of batch to load
     * @return DataImportBatch__c
     */
-    private static DataImportBatch__c loadBatch(ID batchId) {
+    private static DataImportBatch__c loadBatch(Id batchId) {
 
-        List<DataImportBatch__c> listDIBatch = BDI_DataImportBatch_SEL.selectByIds(new Set<Id>{batchId});
+        List<DataImportBatch__c> listDIBatch = BDI_DataImportBatch_SEL.selectByIds(new Set<Id>{ batchId });
```
force2b: Without knowing how extensive the `listDIBatch` var is used, it's probably better named as `dataImportBatches`.

src/classes/BDI_DataImportService.cls
```diff
@@ -77,13 +77,13 @@ global with sharing class BDI_DataImportService {
...
         if (listDIBatch.size() == 1) {
             return listDIBatch[0];
         } else {
-            throw(new BDIException(String.format(Label.bdiInvalidBatchId, new String[]{batchId})));
+            throw(new BDIException(String.format(Label.bdiInvalidBatchId, new String[]{ batchId })));
```
force2b: This and other labels should be `System.Label. bdiInvalidBatchId`

force2b: Also, there's an inconsistency in how the `String.format()` second parm is instantiated throughout this class. Sometimes it's `String[]` and sometimes it's `new List<String>`. I'm more of a fan of the latter, but in the end consistency is better.

src/classes/BDI_DataImportService.cls
```diff
@@ -305,31 +313,37 @@ global with sharing class BDI_DataImportService {
     /*******************************************************************************************************
     * @description whether the Account Custom Unique ID field is in our account matching rules
     */
-    private boolean isCustomIdInAccountMatchRules {
+    private Boolean isCustomIdInAccountMatchRules {
         get {
             String str = diSettings.Account_Custom_Unique_ID__c;
-            if (str == null || str == '')
+            if (str == null || str == '') {
```
force2b: A better var name for this would be `customId`. Also, this could be simplified to `String.isEmpty(customId)`.

src/classes/BDI_DataImportService.cls
```diff
@@ -363,10 +377,11 @@ global with sharing class BDI_DataImportService {
     * @return String either the String value or an empty String ''
     */
     public String strNull(String str) {
```
force2b: Two questions:
- Does this need to be public?
- What would you think of a method name of `replaceNullWithEmptyString(String str)`?

src/classes/BDI_DataImportService.cls
```diff
@@ -441,7 +456,9 @@ global with sharing class BDI_DataImportService {
     * @param isDryRun whether to run in Dry Run mode or commit records.
     * @return void
     */
-    public static void execute(Id apexJobId, Data_Import_Settings__c diSettings, List<DataImport__c> listDI, Boolean isDryRun) {
+    public static void execute(Id apexJobId, Data_Import_Settings__c diSettings, List<DataImport__c> listDI,
```
force2b: A better name for `listDI` might be `dataImportRecords`

src/classes/BDI_DataImportService.cls
force2b: I suspect there are a ton of vars that use the old Hungarian notation (`str*`, `list*`, `set*`, `map*`, etc.). I recommend renaming anything that's local in methods to remove the type prefix. Instance or static vars might be more complex, but I'll leave that up to you.

src/classes/BDI_DataImportService.cls
```diff
@@ -1435,7 +1527,8 @@ global with sharing class BDI_DataImportService {
     * @return String the key for the Address
     */
     private static String strKeyOfAddr(Address__c addr) {
```
force2b: What do you think of renaming method this to `convertAddressToUniqueKey'?

lmeerkatz:
@force2b I know. I decided to go with lowest risk changes only, which is why I didn't touch any variable or method names. That said, those could definitely use work. Do you think it's worth doing here, or should we save those for the next time we make functional changes to this class (which I suspect will happen soon anyway)?

force2b:
It's actually quite surprising how many things need tweaking in this class. You're definitely in BDI for this project, so I feel like this is the time to do it (not to say it has to be done for MVP/DF demo). 

That noted, I think you can get away with var name changes within method pretty safely - especially if you're using Illuminated Cloud since it does such a great job with renaming of vars and methods. Things that are global or static might get a little more complicated, so those could be left for now. I do think the consistency of the `String.format` parm should be fixed though.

src/classes/BDI_DataImportService.cls
RadSFDeveloper: Naming is hard, thus chiming in with more suggestions :) Perhaps `defaultToEmptyIfNull()`, or just `defaultToEmpty()`?

src/classes/BDI_DataImportService.cls
```diff
@@ -595,7 +617,10 @@ global with sharing class BDI_DataImportService {
     * @return void
     */
     private void customPostProcess() {
-        BDI_IPostProcess iPostProcess = (BDI_IPostProcess)instantiateClassForInterface('BDI_IPostProcess', diSettings.Post_Process_Implementing_Class__c);
+        BDI_IPostProcess iPostProcess =
+            (BDI_IPostProcess)instantiateClassForInterface(
```
RadSFDeveloper: Do we add a space after casting records, ie `(BDI_IPostProcess) instantiateClassForInterface()`?

src/classes/BDI_DataImportService.cls
```diff
@@ -733,7 +764,8 @@ global with sharing class BDI_DataImportService {
     * @param mapDIFieldToDstField The field map from a DI field to the destination field
     * @return boolean True if copied, False is not copied.
     */
-    public static boolean isCopiedDIField(DataImport__c di, String strDIField, SObject objDst, Map<String, String> mapDIFieldToDstField) {
+    public static Boolean isCopiedDIField(DataImport__c di, String strDIField, SObject objDst,
```
RadSFDeveloper: What is the guideline for the *long* argument list: all of them in a new line, or one line per argument, or split when the line is longer than 120?

src/classes/BDI_DataImportService.cls
force2b: Nice! Either of those is much better than what I came up with.

src/classes/BDI_DataImportService.cls
lmeerkatz: I tend to split lines more by feel than by guideline. I'm open to suggestions. :)

src/classes/BDI_DataImportService.cls
lmeerkatz: I typically don't, but I don't have a strong opinion either way.

src/classes/BDI_DataImportService.cls
lmeerkatz: This is used in another class, so I'm not planning to update it now. If I open up BDI_ContactService that will be another can of worms.

lmeerkatz:
@force2b I got through many changes but not all of them. Would you be willing to call this close enough for now so Nima can start working with something that's at least somewhat improved?

PR #3647, opened 2018-09-18
BGE: Use picklist for Payment Method in form

bethbrains:
# Critical Changes

# Changes
--Payment Method now renders as a picklist and draws its options from the active picklist values on the Payment Method field on Payment.

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -157,10 +157,18 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
                 col.editable = true;
                 col.type = dfr.type.name().toLowerCase();
+                // Lightning Components interpret "date" as datetime.
+                // Reassign "date" to "datelocal" to render as date only.
                 if (col.type == 'date') {
                     col.type = 'date-local';
                 }
 
+                // special case for payment type to fetch picklist values
+                // TODO: consider abstracting this out to dynamically check for data type of mapped field?
```
lmeerkatz: I don't want a hard-coded solution for this. Let's hold off on implementing this work for now. I'd close this PR without merging, but link to it from the story so we have this to start from when we're ready to pick this back up.

bethbrains:
Closing to come back to this work later.

PR #3652, opened 2018-09-19
Bug Fix - Best Gift Year Rollups

### LINKED ISSUE #3640: Customizable Rollups: Error due to 'Best Gift Year' number not removing '.' separator for non-English locales
[Thread.](https://powerofus.force.com/0D58000004SYnPZ)

Even though the year is a number, Customizable Rollups knows to remove the comma so 2018 is 2018 and not 2,018 for English/USA and other locales that use "," However, Spanish and Italian (and probably others) separate with a "." and this is causing an error.

`STRING_TOO_LONG- Best Gift Year: data value too large: 2.018 (max length=4)`

Steps to repeat:

1. With English as Locale/Language, enable Customizable Rollups
2. Create an Opportunity and verify no errors in rollups
3. Change Language to Italian and Locale to Italian (Italy) for your user
4. Create an Opportunity of any amount - closed/won
5. Observe error in rollup:

image

dvenor:
# Critical Changes

# Changes

# Issues Closed
fixes #3640 
# New Metadata

# Deleted Metadata


dvenor:
**lurch: attach W-028500

davisagli:
@dvenor the "Issues Closed" section of the pull request should link to the github issue (https://github.com/SalesforceFoundation/Cumulus/issues/3640) rather than our internal work item -- our automated tooling uses this when generating release notes and will also add a comment on the issue once the fix is included in a release.

dvenor:
Ah.  Got it.  Thanks @davisagli.  Didn't notice the github issue.

force2b:
@dvenor and @davisagli I'm pretty sure it needs to be "fixes #" followed by the issue number. More info on the autoclose keywords is here: https://help.github.com/articles/closing-issues-using-keywords/

force2b:
@dvenor Also, in this particular case I don't think it's necessary to specify the details of what was fixed in the Changes section since the Release Notes will include a link to the fixed issue.

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+public with sharing class UTIL_String {
+    
+    /*********************************************************************************************************
+     * @description Removes from a string all values that are not an english language alpha character
+     * @param target The the target string for the operation
+     * @return String
+     */
+    public static String englishAlphabetOnly(String target) {
+
+            if (target == null) {
```
force2b: It looks like the indentation is off on the methods in this class.

src/classes/UTIL_String.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="urn:metadata.tooling.soap.sforce.com" fqn="UTIL_String">
```
force2b: I've never noticed an `<apexClass> tag with this format namespace string and and `fqn`. None of our other classes have that. It appears to be referencing the tooling api instead of the soap api. Honestly not sure what to make of this. @davisagli and @cdcarter What do you think about this?

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2018
+* @group Utilities
+* @description Methods used to manipulate string objects
+*
+*/
+public with sharing class UTIL_String {
```
lmeerkatz: Should this be `without sharing`? It shouldn't affect behavior either way, it just seems odd to enforce sharing on a utility class.

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+    
+    /*********************************************************************************************************
+     * @description Removes from a string all values that are not an english language alpha character
+     * @param target The the target string for the operation
```
lmeerkatz: extra 'the' :)

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+     * @return String
+     */
+    public static String englishAlphabetOnly(String target) {
```
lmeerkatz: A name with a verb might be clearer, like removeNonEnglishCharacters(String target). (Applies to multiple method names)

lmeerkatz: Though I suppose that specific option could imply that you're changing the target vs. returning a stripped down value.

src/classes/UTIL_String_TEST.cls
```diff
@@ -0,0 +1,78 @@
...
+    @isTest private static void nonNumericCharactersShouldBeRemoved() {
+        String expectedValue = '0123456789';
+        String actualValue = UTIL_String.numericCharactersOnly(mixedCharacterString);
+        System.assertEquals(expectedValue, actualValue); 
+    }
+
+    /*******************************************************************************************************************
+    * @description Verifies that only english language characters are returned after processing the string
+    */
+    @isTest private static void nonEnglishAlphabetCharactersShouldBeRemoved() {
```
lmeerkatz: It would be good to include a test for characters with accents.

src/classes/UTIL_String.cls
force2b: True. It's really a candidate for `inherited sharing`, but that's not available in 43. How about no sharing declaration at all so it implicitly inherits?

src/classes/CRLP_FiscalYears_TEST.cls
```diff
@@ -33,14 +33,60 @@
...
 * @description Unit Test for the Fiscal Years class
 */
-@isTest(IsParallel=true)
+@isTest
 private class CRLP_FiscalYears_TEST {
 
+    /*********************************************************************************************************
+    * @description Test to ensure fiscal year is correctly formatted as a string with no punctuation
+    */
+    @isTest static void yearStringShouldNotIncludePunctuation() {
```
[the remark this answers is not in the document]
dvenor: For some reason I though we had it in the style guide to have the `@isTest` on the same line.  Looked and clearly that is not the case. Will fix.

src/classes/UTIL_String_TEST.cls
force2b: good one!

src/classes/UTIL_String.cls
force2b: You have much better eyes than I do! :)

src/classes/UTIL_String.cls-meta.xml
dvenor: I initially deployed this using VS Code and SFDX: Create Apex Class function.  Seems like it used the tooling API.  Not sure why but I'll check on it.

src/classes/UTIL_String.cls
dvenor: Good point.  I'd think a utility class would not enforce sharing.

src/classes/UTIL_String.cls
dvenor: I originally had a verb like name but changed it.  It seemed awkward to me for some reason but I now agree it would be clearer.

src/classes/UTIL_String.cls-meta.xml
davisagli: https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/reference_objects_overview.htm defines the `metadata.tooling.soap.sforce.com` namespace as "Objects and types that occur in both the Metadata API WSDL and the Tooling API WSDL." Given that, I think the platform must be treating them similarly. CumulusCI's code to process -meta.xml files is careful to not assume a particular xml namespace and this appears to have deployed successfully for CI via the metadata API, so I don't think it matters too much which one is used.

I assume `fqn` stands for fully-qualified name but I can't find any documentation of what it is used for here!

force2b: Thanks @davisagli. I guess we can leave it. I suppose in the end, it's how the file exists in the dev/packaging org that matters, not how the tooling or metadata/soap api retrieves and formats the meta-xml file. Since the file deploys correctly and to the best of my knowledge there is nothing in the Salesforce UI that allows that part of the meta-xml file to be viewed or modified, I'm cool with it.

@dvenor Maybe you're just the first dev to create & retrieve a new class using VSCode.

dvenor:
Modified the UTIL_String method names to start with a verb and describe what they do.  Also fixed some formatting issues and grammar issues in the descriptions.  Ready for re-review.

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+        if (target == null) {
+            return null;
+        }    
+
+        return target.replaceAll('[^A-Za-z]', '');
+
+    }
+
+    /*********************************************************************************************************
+     * @description Removes from a string all values that are not a numberic character
```
force2b: spelling error in `numeric`

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+*
+*/
+public class UTIL_String {
+    
+    /*********************************************************************************************************
+     * @description Removes from a string all values that are not an english language alpha character
+     * @param target The target string for the operation
+     * @return String
+     */
+    public static String removeNonEnglishCharacters(String target) {
```
force2b: Does "A-Z" technically refer to "English" characters? I'm wondering if this is better worded to as `removeNonAsciiLetters` and maybe `removeNonNumericAndNonAsciiLetters` below. Or possibly `removeNonAlphabeticCharacters`, etc.

src/classes/UTIL_String.cls
```diff
@@ -0,0 +1,83 @@
...
+        return target.replaceAll('[^A-Za-z]', '');
+
+    }
+
+    /*********************************************************************************************************
+     * @description Removes from a string all values that are not a numberic character
+     * @param target The target string for the operation
+     * @return String
+     */
+    public static String removeNonNumeric(String target) {
```
force2b: Since the other methods in this class end with `Characters`, for consistency I would rename this to `removeNonNumericCharacters`.

src/classes/CRLP_FiscalYears.cls
```diff
@@ -63,6 +63,16 @@ public class CRLP_FiscalYears {
         }
     }
 
+    /*********************************************************************************************************
+    * @description Returns the current year, fiscal or calendar depending on the rollup setting as a formatted string.
+    All punctuation will have been removed
+    * @param targetDate The date for which to look up the year.
+    * @return String The calendar or fiscal year for the given Date as a string.
+    */
+    public String getYearFormattedString(Date targetDate) {
```
force2b: Since this class already has a static method for `getFiscalYear(Date targetDate)`, you could create a new static method for `getYearAsFormattedString(Date theDate)`. This way all the places that call `UTIL_String.removeNonNumeric(closeDate.addMonths(n).year().format())` could just be `CRLP_FiscalYears.getYearAsFormattedString(closeDate)`. The instance method can be kept (I believe that's allowed), though it should probably be renamed to `getYearAsFormattedString`.

dvenor: Do you think it's worth making this change at this time?  It's slightly more invasive and requires a bit more code rework, and I'd need to be I understand the business rules around the dates being passed into the function.  Just want to be sure I don't break something.

dvenor: The `year().format()` logic is acting upon other dates, not just the current fiscal year.  Right now we don't really have a use for new static method. Even the current static method is only used within the `CRLP_FiscalYears` class.  I can see how a method that accepts a Date and return the year as a string would be useful, but I don't think that method belongs the in the `CRLP_FiscalYears` class.  Maybe it would make sense to have it in the `UTIL_String` class?

force2b: Fair enough. Let's leave as is.

davisagli:
@force2b @dvenor the yearStringShouldNotIncludePunctuation test failed in the offset fiscal year beta org: https://mrbelvedereci.herokuapp.com/builds/37419/tests

davisagli:
@force2b @dvenor the offset fiscal org has a fiscal year starting in December. So maybe the quickest fix is to use a date in December for the test instead of Jan 1.

If you want to try with the same offset fiscal year config in a scratch org, run the `config_offsetfiscal` CCI flow to set it up.

PR #3666, opened 2018-09-26
Fix an error in the featureParameter name

force2b:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davisagli:
The build failed with a CPU timeout only so I'm willing to force-merge it once @jjbennett approves.

PR #3669, opened 2018-09-27
Adding a CAO_Constants_API to expose CAO string constants

lparrott:
# Critical Changes

# Changes

- We've added a CAO_Constants_API class, which makes it possible for customers and partners to access CAO (Contacts and Organizations) string constants for use in Apex methods.

# Issues Closed

# New Metadata

- CAO_Constants_API
- CAO_Constants_API_TEST

# Deleted Metadata

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TEST.cls
+++ b/src/classes/ACCT_IndividualAccounts_TEST.cls
@@ -354,10 +354,10 @@ private class ACCT_IndividualAccounts_TEST {
         //one contact should have been created
         system.assertEquals(1,insertedContacts.size());
         //the account should have the right type
-        if (strProcessor == CAO_Constants.ONE_TO_ONE_PROCESSOR)        
-            system.assertEquals(CAO_Constants.ONE_TO_ONE_ORGANIZATION_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
-        else if (strProcessor == CAO_Constants.HH_ACCOUNT_PROCESSOR)    
-            system.assertEquals(CAO_Constants.HH_ACCOUNT_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
+        if (strProcessor == CAO_Constants_API.ONE_TO_ONE_PROCESSOR)        
+            system.assertEquals(CAO_Constants_API.ONE_TO_ONE_ORGANIZATION_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
+        else if (strProcessor == CAO_Constants_API.HH_ACCOUNT_PROCESSOR)    
+            system.assertEquals(CAO_Constants_API.HH_ACCOUNT_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
         //the account should be flagged as an individual
         system.assertEquals(true,insertedcontacts[0].Account.npe01__SYSTEMIsIndividual__c );
         //the account should have been named correctly
@@ -568,14 +568,14 @@ private class ACCT_IndividualAccounts_TEST {
         
         Contact[] insertedContacts = [Select Account.Name,AccountId,npe01__Organization_Type__c from Contact where id=:con.id];
         //the contact should have the right type
-        if (strProcessor == CAO_Constants.BUCKET_PROCESSOR)
-            system.assertEquals(CAO_Constants.BUCKET_ORGANIZATION_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
-        else if (strProcessor == CAO_Constants.HH_ACCOUNT_PROCESSOR)
-            system.assertEquals(CAO_Constants.HH_ACCOUNT_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
+        if (strProcessor == CAO_Constants_API.BUCKET_PROCESSOR)
+            system.assertEquals(CAO_Constants_API.BUCKET_ORGANIZATION_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
+        else if (strProcessor == CAO_Constants_API.HH_ACCOUNT_PROCESSOR)
+            system.assertEquals(CAO_Constants_API.HH_ACCOUNT_TYPE,insertedcontacts[0].npe01__Organization_Type__c);
         //the account should have been named correctly
-        if (strProcessor == CAO_Constants.BUCKET_PROCESSOR)
-            system.assertEquals(CAO_Constants.BUCKET_ACCOUNT_NAME,insertedcontacts[0].Account.Name);
-        else if (strProcessor == CAO_Constants.HH_ACCOUNT_PROCESSOR)
+        if (strProcessor == CAO_Constants_API.BUCKET_PROCESSOR)
+            system.assertEquals(CAO_Constants_API.BUCKET_ACCOUNT_NAME,insertedcontacts[0].Account.Name);
+        else if (strProcessor == CAO_Constants_API.HH_ACCOUNT_PROCESSOR)
             system.assertEquals(con.LastName + ' Household',insertedcontacts[0].Account.Name);        
     }
     
--- a/src/classes/CAO_Constants_API.cls
+++ b/src/classes/CAO_Constants_API.cls
@@ -1,0 +1,96 @@
+/*
+    Copyright (c) 2018, Salesforce.org
+    All rights reserved.
+    
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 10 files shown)


force2b:
@weizsforg and @lparrott Thanks for the heads up on this. One question I have is would it be better to expose these npsp-specific CAO vars as global through a more generic (non-ADV specific) class such as `API_Constants` or `CAO_Constants_API`?

My other concern (and I have no knowledge to confirm this right now) is that by exposing the CAO Constants as global, the vars in the `CAO_Constants` class might become locked because they're referenced directly by global methods in the `ADV_Adapter` class. I'm not sure if this is truly an issue and I'm not necessarily sure how to test, but I at least want to mention it in case you have additional info to provide. I suppose that we can create a global class in the CCILmaTest package to experiment with this.

Also, out of curiosity, does the ADV package have a dependency on NPSP or is it completely dynamic? I suppose I'm just wondering if there is a need to expose a method to update the `npe01__Contacts_And_Orgs_Settings__c` object or if it could just be updated directly from the ADV package? There's no business specific logic in that method and it really does an `update` call, which could just as easily be done from outside the `npsp` namespace.

cc @njjc @lmeerkatz @dvenor

weizsforg:
Hi @force2b Thanks for the feedback! 

Regarding to the CAO_Constants, currently only Adv needs to refer to the Strings. Is there any use case for the regular NPSP user to use it? I am thinking we might need to limit the scope for only Adv user by putting the Adv package check. 

@carlos, any idea about the variable gets locked when they're referenced directly by global methods? 

Currently, Adv does have a dependency on NPSP. I agree that we might just update the setting in Adv package. Any ideas @ceiroa ?

lparrott:
Hi @force2b, have you had a chance to think on where you'd like the global CAO_Constants variables to live? Just let us know if you'd prefer to add them to a new, or existing, global class. Thanks!

dvenor:
I like idea of creating the `API_Constants` class, keeping it more generic.  NPSP should expose what it wants to expose through the global interface, without too much concern about who is consuming the constants.  A more generic name would reflect that idea.

ceiroa:
Since the other API-related classes have the API suffix, would it make
sense to call it Constants_API?

Carlos Ramírez Martínez-Eiroa | Lead Member of Technical Staff, Engineering
| Salesforce.org | 630-636-0013

njjc:
I believe the CAO prefix was for the old "Contacts and Organizations" package. I think `CAO_Constants_API` is the most consistent with our current naming, so that has my vote.

force2b:
Sounds good. `CAO_Constants_API` it is. Thinking of it in a different way, as we expose API functionality out for each distinct set of functionality in the application, using the `_API` suffix would be more consistent.

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -354,10 +354,10 @@ private class ACCT_IndividualAccounts_TEST {
         //one contact should have been created
         system.assertEquals(1,insertedContacts.size());
         //the account should have the right type
-        if (strProcessor == CAO_Constants.ONE_TO_ONE_PROCESSOR)
```
force2b: @lparrott Technically these can still use the `CAO_Constants` class because this test class is in the package, but either way works.

src/classes/CAO_Constants_API.cls
```diff
@@ -0,0 +1,96 @@
...
+* @group-content ../../ApexDocContent/API.htm
+* @description API class that exposes CAO (Contacts and Organizations) Constants and helper methods.
+*/
+global class CAO_Constants_API {
+
+    /*******************************************************************************************************
+    * @description One-to-One value for system processor field
+    */ 
+    global static final String ONE_TO_ONE_PROCESSOR {
+        get {return CAO_Constants.ONE_TO_ONE_PROCESSOR;}
```
force2b: @lparrott I ran into an apex recently bug where not having a `set;` would cause a compile error when changing api versions. The recommended fix was to include `private set;` anywhere where a set was implied but not specified.

ceiroa: But since we are not changing API versions (the whole class is new) maybe this will work? I'd just hate to add cruft unnecessarily...

force2b: It's possible that changing from 43 to a newer version will introduce this bug again. Plus, in order to complete the v43 changes I had to fix every single method that had a getter without a setter, so it would be better if anything new stays consistent with that pattern.

src/classes/ACCT_IndividualAccounts_TEST.cls
lparrott: True, but to get code coverage for the new CAO_Constants_API I had to call it in a few places.

src/classes/CAO_Constants_API.cls
lparrott: That's an interesting bug! I've added the private set methods just in case.

lparrott:
@weizsforg @force2b I've made the requested changes, let me know if there's anything else, otherwise I'll move this along to QA.

force2b:
Thanks @lparrott. I'll review a little later this afternoon.

src/classes/ACCT_IndividualAccounts_TEST.cls
force2b: Very good point, though if you don't mind rather than just support "code coverage", what do you think of creating a new unit test class for `CAO_Constants_API_TEST` that has a single test method to assert each of the getter methods. Something like
```
System.AssertEquals(CAO_Constants.ONE_TO_ONE_PROCESSOR, CAO_Constants_API.ONE_TO_ONE_PROCESSOR);
```

src/classes/CAO_Constants_API.cls
force2b: Thanks for making that change.

src/classes/CAO_Constants_API.cls
```diff
@@ -0,0 +1,104 @@
...
+        private set;
+    }
+
+    /*******************************************************************************************************
+    * @description One-to-One org type value
+    */ 
+    global static final String ONE_TO_ONE_ORGANIZATION_TYPE {
```
force2b: Just noticing the `final` keyword. If these are `private set;`, is there a reason to use `final`? I'm asking because as global methods, we won't be able to ever change these once they're included in a package.

lparrott: Fair point, I'll remove final.

src/classes/ACCT_IndividualAccounts_TEST.cls
ceiroa: If I can give my $0.02, I like this better because it's an instance of us eating our own dog food. It's the same we do in the triggers, where we could call the method directly but instead go through the API one. For example in https://github.com/SalesforceFoundation/Cumulus/blob/master/src/triggers/TDTM_Account.trigger#L33.

force2b: Agreed. I'm fine with calling `if (strProcessor == CAO_Constants_API.ONE_TO_ONE_PROCESSOR)`, though I think it would be better to have a unit test for the API class rather than relying on another test to cover that code implicitly.

force2b: You've raised a really good point. What that comes down to though is us writing code differently to work around issues with our legacy unit test code base. Long term we want to fix that, so to me I feel like we anything new should be built with an eye towards better test classes, as opposed to following the older patterns. 

I think we can still accomplish this without negatively affecting the unit test performance by using `@IsTest(IsParallel=True)` in the Unit Test Class definition. That's something we've been doing a lot as new test classes are built, especially ones that follow newer design patterns and can run in parallel.

force2b: @lparrott Carlos and I discussed this. We've agreed to leave your code as is (where the method references the `CAO_Constants_API` class), however we've also agreed to create a new unit test class of  `CAO_Constants_API_TEST' with `@IsTest(IsParallel=True)` that has a single method and a series of `System.AssertEquals()` to validate each of the exposed global methods in the  `CAO_Constants_API` class.

lparrott:
@force2b Not a problem, I'm glad we came up with a solution that works for everyone (and I learned a thing or two about global classes and managed packages, so that's a bonus).

annejackson:
@njjc, @force2b and @lparrott - I'll be testing this one soon. Please review the text under the Changes section to make sure it's clear and good to go for documentation/release notes. (Feel free to have others review as well). Thanks!

annejackson:
Testing is complete. Merging PR.

davisagli:
@annejackson I'll take care of it this time, but please remember to click the "Delete branch" button after you merge a pull request. If you don't, mrbelvedere will keep updating the branch, which triggers unnecessary builds. Thanks!

PR #3677, opened 2018-10-03
Add an NPSP-specific CumulusCI task for generating fake test data.

davisagli:
This is what I used to generate data for the XXL org.
It inserts rows in an empty postgres database.
From there they can be loaded into Salesforce using the standard CumulusCI LoadData task
(well, at least that will be possible once I complete a PR with some refactoring there)
with the mapping_generated.yml included here.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


testdata/generate.py
```diff
@@ -0,0 +1,526 @@
...
+
+    def _run_task(self):
+        database_url = self.options["database_url"]
+        count = self.options["count"]
+
+        # patch faker for performance
+        self._install_faker_patch()
+
+        self.logger.info("Connecting to {}".format(database_url))
+        conn = psycopg2.connect(database_url)
```
jlantz: Is it possible that we could just use sqlalchemy to connect to the database here instead of calling psycopg2 directly?  I'd like to test this with generating data to a sqlite database if possible which seems like a common use case.

testdata/generate.py
```diff
@@ -0,0 +1,526 @@
...
+            num_opps = opp_gen.choose_count()
+        for x in range(num_opps):
+            opp_gen.make_one(contacts, account_id=id)
+
+        self.writer.writerow(attrs)
+        return attrs
+
+    def make_many(self, count, **kw):
+        i = 0
+        while i < count:
```
jstvz: Why the `while` loop?

testdata/generate.py
```diff
@@ -0,0 +1,526 @@
...
+
+    def make_one(self, n):
+        id = self.make_id()
+        attrs = {"id": id, "record_type": "HH_Account"}
+
+        # contacts
+        contact_gen = self.generators["contacts"]
+        num_contacts = contact_gen.choose_count()
+        contacts = []
+        for x in range(num_contacts):
```
jstvz: Why not use a list comprehension here? I thought they were faster than appending?

testdata/generate.py
```diff
@@ -0,0 +1,526 @@
...
+                role="Donor" if first else "Influencer",
+                is_primary=first,
+            )
+            first = False
+
+        # payments
+        payment_gen = self.generators["payments"]
+        payment_count = payment_gen.choose_count()
+        payment_amount = round(amount / payment_count, 2)
+        for n in range(payment_count):
```
jstvz: Does it matter that `range()` has different behavior in python 2/3? Probably not, right?

testdata/generate.py
davisagli: Inserting from CSV using COPY FROM is only supported by postgres, but I may take a stab at supporting other dbs the way I did for the bulkdata tasks. It'd be nice to have that flexibility.

testdata/generate.py
```diff
@@ -0,0 +1,526 @@
+# Requirements:
```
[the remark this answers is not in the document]
davisagli: I'd say give the project its own requirements.txt and recommend using a separate virtualenv for it

testdata/generate.py
davisagli: It's a small list, so I doubt it has much impact.

testdata/generate.py
davisagli: I tend to avoid list comprehensions for readability reasons, but I think it's a good call here.

testdata/generate.py
davisagli: Huh yeah, I can't remember and will change it.

PR #3678, opened 2018-10-04
Enforce minimum CCI version

davisagli:
The npsp_default_settings task doesn't work in earlier versions because I moved the anon apex task.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davisagli:
Oh but, earlier versions don't have the check for minimum_cumulusci_version either. Never mind.

PR #3679, opened 2018-10-04
New Batch Wizard

randi274:
# Critical Changes

# Changes
- Added Batch Wizard modal to the DataImportBatch record home, reusing the Batch Template component
# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/BGE_Wizard/BGE_WizardHelper.js
```diff
@@ -764,13 +834,61 @@
...
             }
 
+            /**
+             * @description sets the page header based on the step in the wizard
+             * @return void.
+             */
+            function setPageHeader() {
+                switch (parseInt(this.progressIndicatorStep)) {
+                    case 1:
+                        this.pageHeader = $A.get('$Label.c.bgeBatchTemplateOverviewWizard');
```
lmeerkatz: Could we move step configuration into an array so we can maintain simple increment/decrement logic?

```
If it's just headers ... 
var headers = [ 
   $A.get('$Label.c.bgeBatchTemplateOverviewWizard'), 
    'Select Template',
    ...
]
```

or if there's other data, move it into a step object?

src/aura/modalFooter/modalFooter.cmp
```diff
@@ -0,0 +1,82 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+-->
+<!--
+  @author Salesforce.org
+  @date 2018
+  @group Batch Gift Entry
```
lmeerkatz: If this is just for BGE, the name of the component should be specific to BGE

src/aura/modalFooter/modalFooter.cmp
```diff
@@ -0,0 +1,82 @@
...
+  @group-content ../../ApexDocContent/BatchDataImport.htm
+  @description Component used to view, edit, or create a Batch Template record.
```
lmeerkatz: Incorrect comment?

src/aura/modalFooter/modalFooter.cmp
```diff
@@ -0,0 +1,82 @@
...
+-->
+<aura:component>
+
+    <!--Public Attributes-->
+    <aura:attribute name="currentStep" default="1" type="String" description="Step in progress indicator" />
+    <aura:attribute name="dataTableChanged" default="false" type="Boolean" description="Controls whether Save is disabled because Data Table has uncommitted changes" />
+    <aura:attribute name="progressSteps" default="[]" type="List"
```
lmeerkatz: Is this the same list we'd need for the step headers?

src/aura/modalFooter/modalFooter.cmp
lmeerkatz: (applies to multiple components/classes)

src/aura/modalFooter/modalFooterController.js
```diff
@@ -0,0 +1,47 @@
...
+        } else if (channel === 'dataTableChanged') {
+            component.set('v.dataTableChanged', message);
+        }
+    },
+
+    /**
+     * @description: sends back event to modal
+     */
+    back: function(component, event, helper) {
+        helper.sendMessage(component, 'back', null);
```
lmeerkatz: Can you drop the third param if it's null?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2018
+* @group BGE
+* @group-content 
+* @description BGE Wizard Component Controller.
+*/
+public with sharing class BGE_Wizard_CTRL {
```
lmeerkatz: We probably want a more specific name here as well. BGE_ConfigurationWizard?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+                WHERE Id = :recordId];
+
+                model.name = recordDetails.Name;
+                model.description = recordDetails.Batch_Description__c;
+                model.enableTotalEntry = recordDetails.EnableTotalEntry__c;
+                model.id = recordDetails.Id;
+                model.requireTotalMatch = recordDetails.RequireTotalMatch__c;
+                model.activeFields = recordDetails.Active_Fields__c;
+            }
+        }
```
lmeerkatz: What if the record type is neither of the checked types?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+                'Account2ImportStatus__c'
+        };
+
+        Set<String> allowedObjects = new Set<String>{
+                'Opportunity',
+                'Payment'
+        };
+
+        // these fields are required only if at least 1 field from its object group is selected
+        Set<String> systemRequiredFields = new Set<String>{
```
lmeerkatz: We're not actually requiring account name/contact last name when we're using lookups. Is this still valid?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+                'Account1_Name__c',
+                'Account2_Name__c',
+                'Donation_Amount__c',
+                'Contact1_LastName__c',
+                'Contact2_LastName__c'
+        };
+
+        // looping through the fields on DataImport__c
+        for (Schema.DescribeFieldResult dfr : fieldMap.values()) {
```
lmeerkatz: This could use a more descriptive variable name. dfr -> fieldResult?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+            String dataImportFieldApiName = UTIL_Namespace.StrTokenRemoveNSPrefix(dfr.name);
+            //Special case for Donation_Donor because it's derived
+
+            //todo: rework these 2 lines
+            String dataImportFieldMappedObject = dfr.inlineHelpText != null ? dfr.inlineHelpText.split('\\.')[0] : null;
+            Boolean fieldAllowed = allowedObjects.contains(dataImportFieldMappedObject) && !bannedFields.contains(dataImportFieldApiName) && dfr.inlineHelpText != null;
+
+            if (fieldAllowed) {
+
+                AvailableField fld = new AvailableField();
```
lmeerkatz: fld -> field

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+
+                // TODO: make these localized with schema describes
+                fld.sObjectName = dataImportFieldMappedObject;
+
+                // TODO: pull these labels from target fields' schema based on help text mapping
+                fld.label = (dfr.getLabel().contains(' ') && dfr.getLabel().split(' ',2)[0] == fld.sObjectName) ?
+                        dfr.getLabel().split(' ',2)[1] :
```
lmeerkatz: Why the double indentation?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+
+            if (fieldAllowed) {
+
+                AvailableField fld = new AvailableField();
+
+                // TODO: make these localized with schema describes
+                fld.sObjectName = dataImportFieldMappedObject;
+
+                // TODO: pull these labels from target fields' schema based on help text mapping
+                fld.label = (dfr.getLabel().contains(' ') && dfr.getLabel().split(' ',2)[0] == fld.sObjectName) ?
```
lmeerkatz: I don't immediately follow what this is doing. Is it worth splitting this into its own method for clarity?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+    }
+    /*******************************************************************************************************
+    * @description Retrieves the object and field labels for the component.
+    * @param sObjectName the object being used by the component (already stripped of the namespace prefix)
+    * @return Map of object and field name keys to their labels.
+    */
+    public static Map<String, String> getLabels(String sObjectName) {
+
+        if (sObjectName == 'Batch_Template__c') {
+            Map<String, String> labels = new Map<String, String> {
```
lmeerkatz: Split contents of if statements into two separate methods: getBatchTemplateLabels(), getBatchLabels()

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+                    'missingNameDescriptionError' => Label.exceptionRequiredField + ' ' + Schema.Batch_Template__c.Name.getDescribe().getLabel() + ', ' + Schema.Batch_Template__c.Description__c.getDescribe().getLabel(),
+                    'description' => Schema.Batch_Template__c.Description__c.getDescribe().getLabel(),
+                    'totalEntryHelp' => Schema.Batch_Template__c.EnableTotalEntry__c.getDescribe().getInlineHelpText(),
+                    'totalEntryLabel' => Schema.Batch_Template__c.EnableTotalEntry__c.getDescribe().getLabel(),
+                    'totalMatchHelp' => Schema.Batch_Template__c.RequireTotalMatch__c.getDescribe().getInlineHelpText(),
+                    'totalMatchLabel' => Schema.Batch_Template__c.RequireTotalMatch__c.getDescribe().getLabel()
+            };
+            return labels;
+        } else {
+            Map<String, String> labels = new Map<String, String> {
```
lmeerkatz: What if the objectName provided doesn't match either template or batch?

src/classes/BGE_Wizard_CTRL.cls
```diff
@@ -0,0 +1,307 @@
...
+    * @param activeFields JSON string of the active fields to be saved to the record
+    * @return string of the record information with the Id
+    */
+    @AuraEnabled
+    public static string saveRecord(String sObjectName, String recordInfo, String activeFields) {
+        try {
+            BGEDetailsModel recordDetails = (BGEDetailsModel) JSON.deserialize(recordInfo, BGEDetailsModel.class);
+
+            if (sObjectName == 'Batch_Template__c') {
+                Batch_Template__c batchTemplate = new Batch_Template__c();
```
lmeerkatz: Split contents of if blocks into separate methods

src/labels/CustomLabels.labels
```diff
@@ -2452,6 +2452,13 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Header for Batch Template Info in Wizard</shortDescription>
         <value>Enter Template Info</value>
     </labels>
+    <labels>
+        <fullName>bgeBatchOverviewWizard</fullName>
```
lmeerkatz: The name, description, and value don't seem quite in sync to me.

src/objects/DataImportBatch__c.object
```diff
@@ -55,6 +55,16 @@
     <enableSearch>true</enableSearch>
     <enableSharing>true</enableSharing>
     <enableStreamingApi>true</enableStreamingApi>
+    <fields>
+        <fullName>Active_Fields__c</fullName>
+        <description>Represents the data model of active fields for a NPSP_DataImportBatch__c record. Not intended for direct customer use and should not be edited directly.</description>
+        <externalId>false</externalId>
+        <label>Active Fields</label>
+        <length>30000</length>
```
lmeerkatz: Is there a reason to cap this below 32,768?

src/objects/DataImportBatch__c.object
```diff
@@ -224,6 +234,16 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
```
lmeerkatz: I believe we removed this?

randi274:
**lurch: attach W-028584

src/aura/modalFooter/modalFooter.cmp
randi274: Good point! I'd like for this to be reusable, once we can figure out how to make the progress indicator dynamic, so I'll relabel appropriately. That might be fixed with a facet we can pass in, if not attributes.

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fNyFAAU" target="_blank">W-028584</a>

src/aura/modalFooter/modalFooter.cmp
randi274: Yes, in addition to needing the value. progressSteps is intended to be a List<Map<String,String>>

src/classes/BGE_Wizard_CTRL.cls
randi274: I'd like to leave this logic in, since we do plan to need it when we come back to not using lookups any longer.

src/classes/BGE_Wizard_CTRL.cls
randi274: I think Beth put that in, but I think to see the if/else matching lines separately. I refactored the ternary to be more readable with a boolean.

src/classes/BGE_Wizard_CTRL.cls
randi274: I changed the code to the following. Let me know if this is still unclear and I can make it another method:

Boolean fieldHasSpacesAndObjName = fieldResult.getLabel().contains(' ') && fieldResult.getLabel().split(' ',2)[0] == field.sObjectName;
field.label = (fieldHasSpacesAndObjName) ? fieldResult.getLabel().split(' ',2)[1] : fieldResult.getLabel();

src/objects/DataImportBatch__c.object
randi274: Nope! Where does the standard of 32,768 come from?

src/classes/BGE_Wizard_CTRL.cls
lmeerkatz: I don't mind using a ternary, but I think the intent of the logic could use explanation. What is this trying to do?

```
(dfr.getLabel().contains(' ') && dfr.getLabel().split(' ',2)[0] == fld.sObjectName) ?
                        dfr.getLabel().split(' ',2)[1] : 
```

src/objects/DataImportBatch__c.object
randi274: Yep! Not sure why it wasn't carried over.

src/objects/DataImportBatch__c.object
lmeerkatz: That's the default limit on a long text area, though it looks like the maximum is actually higher: https://help.salesforce.com/articleView?id=000267856&language=en_US&type=1

I was just wondering why you picked something other than the default. We should also consider whether we'll want to go directly for the maximum as field length may be something we can't alter after packaging.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
```diff
@@ -0,0 +1,79 @@
...
+                <div class="slds-col slds-no-flex slds-grid slds-align-top slds-p-bottom_xx-small headerButtonActions">
+                    <lightning:buttonGroup>
+                        <lightning:button aura:id="newBGEBatch" label="{!$Label.c.bgeNewBatch}" onclick="{!c.openNewBatchWizard}"/>
+                    </lightning:buttonGroup>
+                </div>
+            </div>
+        </div>
+    </header>
+
+    <!--listView does not rerender dynamically, so need aura:if to have it wait before loading-->
```
lmeerkatz: What happens when the list view reference is invalid?

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
```diff
@@ -0,0 +1,65 @@
...
+     * @description: instantiates component. Only called when component is first loaded.
+     */
+    doInit: function (component, event) {
+        var action = component.get('c.isOrgNamespaced');
+        action.setCallback(this, function (response) {
+            var state = response.getState();
+            if (state === 'SUCCESS') {
+                var isOrgNamespaced = JSON.parse(response.getReturnValue());
+                if (isOrgNamespaced) {
+                    var listViewNamespaced = {'objectApiName': 'npsp__DataImportBatch__c', 'listName': 'npsp__Gift_Batches', 'isLoaded': true};
```
lmeerkatz: Line length is getting a bit long. Consider breaking into multiple lines for readability.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
```diff
@@ -0,0 +1,65 @@
...
+                    modalBody = components[0];
+                    modalHeader = components[1];
+                    modalFooter = components[2];
+                    component.find('overlayLib').showCustomModal({
+                        body: modalBody,
+                        header: modalHeader,
+                        footer: modalFooter,
+                        showCloseButton: true,
+                        cssClass: 'slds-modal_large',
+                        closeCallback: function() {
```
lmeerkatz: Should this be included if its body is blank?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -0,0 +1,254 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+-->
+<!--
+  @author Salesforce.org
+  @date 2018
+  @group Batch Gift Entry
+  @group-content ../../ApexDocContent/BatchDataImport.htm
+  @description Component used to view, edit, or create a Batch Template record.
```
lmeerkatz: " ... create a Batch Template record" -> " ... create a Batch or Batch Template record"

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -0,0 +1,254 @@
...
+                                    </span>
+                                </h3>
+                            </div>
+                        </aura:set>
+                    </aura:if>
+
+                    <!--Progress Indicator-->
+                    <aura:if isTrue="{!not(v.isReadOnly)}">
+                        <div class="slds-p-bottom_large">
+                            <lightning:progressIndicator currentStep="{!v.templateMetadata.progressIndicatorStep}" type="base"
```
lmeerkatz: Are there cases where hasError should be true?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -0,0 +1,254 @@
...
+                    <lightning:layout horizontalAlign="center">
+                        <lightning:layoutItem size="10">
+                            <lightning:input type="text" name="name" label="{!$Label.c.stgLabelName}"
+                                             value="{!v.templateInfo.name}" readonly="{!v.isReadOnly}" required="true"/>
+                            <lightning:textarea name="description" label="{!v.templateMetadata.labels.description}"
+                                                value="{!v.templateInfo.description}" readonly="{!v.isReadOnly}"
+                                                required="true" class="slds-p-top_medium"/>
+                            <div class="slds-p-top_large">
+                                <aura:if isTrue="{!v.isReadOnly}">
+                                    <ui:outputCheckbox value="{!v.templateInfo.requireTotalMatch}" class="slds-form-element__control"/>
```
lmeerkatz: Add inputs for Total Expected Gifts and Total Expected Amount above the checkbox

lmeerkatz: Oh wait, this belongs with the read/write version on line 144.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -0,0 +1,254 @@
...
+
+                <!-- Step 2: Template Selection (Only for Batch Gift Entry) -->
+                <aura:if isTrue="{!v.templateMetadata.progressIndicatorStep == '2'}">
+                </aura:if>
+                <!-- End Step 2 -->
+
+                <!-- Step 3: Select Fields -->
+                <aura:if isTrue="{!v.templateMetadata.progressIndicatorStep == '3'}">
+                    <div class="slds-m-around_medium">
+                        <lightning:accordion>
```
lmeerkatz: We should consider replacing this with some basic headers unless we have solid knowledge that this will be fixed in Winter 18.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardController.js
```diff
@@ -95,43 +63,47 @@
         model.getTemplateMetadata().setDataTableChanged(false);
     },
 
-    /**
-    * @description Moves to next wizard step
-    */
-    next: function(component, event, helper) {
+    handleButtonClick: function(component, event, helper) {
```
lmeerkatz: Consider breaking the conditions into smaller methods. There's a lot going on here!

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -46,14 +45,18 @@
             model.getTemplateMetadata().onMetadataUpdated.subscribe(function() {
                 var templateMetadataView = component.get('v.templateMetadata');
                 var templateMetadata = model.getTemplateMetadata();
+                var headerChanged = Boolean(templateMetadataView.pageHeader !== templateMetadata.pageHeader);
```
lmeerkatz: Do you need 'Boolean' here?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -314,6 +336,11 @@
                                 requireTotalMatch: response.requireTotalMatch
                             }
                         );
+                        var navEvt = $A.get('e.force:navigateToSObject');
```
lmeerkatz: I'd break this into another method

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntry.cmp
```diff
@@ -70,12 +71,41 @@
...
+                        </div>
+                        <div class="slds-media__body">
+                            <p class="slds-breadcrumb__item slds-line-height_reset">Batch Gift Entry</p>
+                            <h1 class="slds-page-header__title slds-m-right_small slds-truncate">
+                                {!v.record.Name}
+                            </h1>
+                        </div>
+                    </div>
+                </div>
+                <div class="slds-col slds-no-flex slds-grid slds-align-top slds-p-bottom_xx-small headerButtonActions">
```
lmeerkatz: Comment out the entire block if we're not using it?

src/objects/DataImportBatch__c.object
lmeerkatz: I think I deleted the field from template but maybe you added it for batch?

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
randi274: Ideally I'd like to retrieve the list view name from Apex to make sure we have a match. As it is now, if the reference is invalid then this list view won't display.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: Good catch! I'll look at adding the error in today if it's a small add, otherwise will make a minor story to wire up the error here and in the footer.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: Done!

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: It looks like slds-hide doesn't cause the error in this case, so we may lean on that until we can figure out what's going wrong. The trade-off is that it doesn't open automatically, but we still get the rest of the functionality. We're also still working to reproduce what's going wrong, because a simple repro of it works just fine in other winter orgs.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: Was able to add this in, and it looks pretty nice!

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: Actually since we only have 2 items, we'll just change these both to be headers that display all available fields and remove the accordions for now.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
bethbrains: This seems like the kind of thing we would store in a custom setting so an Admin has visibility and control. Do you think we should do that at this stage, or iterate toward that later?

lmeerkatz: I was thinking design attribute on the component. I think it's okay but not ideal for the list to simply not appear, but an error message might be better. We can move this into another story.

PR #3690, opened 2018-10-10
Resolve issue with recurring donation query timeout when building org telemetry data

### LINKED ISSUE #3676: Org Telemetry can throw timeout error for count of Recurring Donation records in larger orgs
[Thread](https://powerofus.force.com/0D58000004Xpmpl) (Partner group).

Customer is getting error notification:

```
Message: First error: [REQUEST_RUNNING_TOO_LONG] Your request was running for too long, 
and has been stopped.
Context: npsp__UTIL_OrgTelemetry_BATCH
```

This is related to counting Recurring Donation records.

As this batch only runs at the time of the install/upgrade of NPSP, the error can be ignored until it's fixed.

force2b:
Implement a new batch job to handle counting RecurringDonation Opportunities to avoid a time out when counting millions of records with a straight count() query. Also instead of using strings for each FeatureParameter name, convert each to an Enum to act as a reusable constant throughout the code. Finally default all integer featureParamters to -1 to aid in identifying where orgs may be running into errors querying the data.

# Critical Changes

# Changes

# Issues Closed

Fixes #3676 

# New Metadata

* UTIL_OrgTelemetry_SObject_BATCH
* UTIL_OrgTelemetry_SObject_BATCH_TEST

# Deleted Metadata


force2b:
**lurch: attach W-028851

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fOKVAA2" target="_blank">W-028851</a>

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @description For each batch iteration, loop through the records and count as needed into
+     * stateful vars
+     */
+    public void execute(Database.BatchableContext context, List<SObject> scope) {
+        for (SObject record : scope) {
+            switch on (jobMode) {
+                when RECURRING_DONATIONS {
+                    getRecurringDonationsCounts((Opportunity)record);
```
RadSFDeveloper: Missing space, ie `((Opportunity) record);`

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * @description Process RecurringDonation counters for the specific Opportunity
+     */
+    private void getRecurringDonationsCounts(Opportunity opp) {
```
RadSFDeveloper: - To decrease the indentation and increase readability, return from the method if the Opp does not need to be processed.
- Increase readability by increasing vertical openness.

```
if (opp.npe03__Recurring_Donation__c == null) {
    return;
}

countAllRDOpps++;
            
Object rdStatus = opp.npe03__Recurring_Donation__r.npe03__Open_Ended_Status__c;
if (rdStatus != null && (String) rdStatus == recurringDonationOpenStatusValue) {
    countOpenRDOpps++;
}
```

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /** @description The job mode this particular batch iteration is executing in */
+    private UTIL_OrgTelemetry_SObject_BATCH.TelemetryBatchJobMode jobMode;
+
+    /** @description Stateful counter vars used for the different counting jobs */
+    private Integer countAllRDOpps = 0;
+    private Integer countOpenRDOpps = 0;
+
+    private static final String recurringDonationOpenStatusValue = System.Label.npe03.RecurringDonationOpenStatus;
```
RadSFDeveloper: The counter vars have `RD` in the name, should this var be named `rdOpenStatus` instead (no need for `Value`)?

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -0,0 +1,119 @@
...
+    public Database.QueryLocator start(Database.BatchableContext context) {
+        switch on (jobMode) {
+            when RECURRING_DONATIONS {
+                return Database.getQueryLocator(
+                        [SELECT Id, npe03__Recurring_Donation__r.npe03__Open_Ended_Status__c
+                        FROM Opportunity
+                        WHERE npe03__Recurring_Donation__c != null]);
+            }
+        }
+        return null;
```
RadSFDeveloper: The `start()` method cannot return `null` for the QueryLocator. For example, when the batch is executed with the `jobMode` being null
```
Database.executeBatch(new UTIL_OrgTelemetry_SObject_BATCH(null), 2000);
```
then the following error is issued:
> First error: Start did not return a valid iterable object.

I guess this is fine since the batch is not global, but wanted to notify you.

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -0,0 +1,65 @@
...
+* @group Utilities
+* @description Unit Tests related to the org telemetry class
+*/
+@IsTest(IsParallel=false)
+private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
+
+    @TestSetup
+    static void testDataSetup() {
+
+        Account a = new Account(Name = 'UTIL_OrgTelemetry_SObject_BATCH_TEST', Type='New');
```
RadSFDeveloper: Missing space, ie `Type = 'New'`.

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -0,0 +1,65 @@
...
+        insert a;
+
+        npe03__Recurring_Donation__c rd = new npe03__Recurring_Donation__c(
```
RadSFDeveloper: Suggestion: can call `insert new npe03__Recurring_Donation__c(` instead of declaring the variable since it is not used anywhere else in the code.

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -0,0 +1,65 @@
...
+                npe03__Installment_Period__c = System.Label.npe03.RecurringDonationInstallmentPeriodYearly,
+                npe03__Date_Established__c = Date.newInstance(2018,6,15),
+                npe03__Schedule_Type__c = System.Label.npe03.RecurringDonationMultiplyValue,
+                npe03__Open_Ended_Status__c = System.Label.npe03.RecurringDonationOpenStatus
+        );
+        insert rd;
+    }
+
+    @IsTest
+    static void testSObjectBatchJob() {
```
RadSFDeveloper: Missing description for the test method.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -130,15 +145,15 @@ public without sharing class UTIL_OrgTelemetry_SVC {
         STG_SettingsService stgSvc = new STG_SettingsService();
         Addr_Verification_Settings__c addrVerifSettings = UTIL_CustomSettingsFacade.getAddressVerificationSettings();
 
-        System.FeatureManagement.setPackageBooleanValue('IsEnabled_HouseholdAcctModel',
+        System.FeatureManagement.setPackageBooleanValue(TelemetryParameterName.IsEnabled_HouseholdAcctModel.name(),
```
RadSFDeveloper: Nice use of the enum!

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -194,42 +209,19 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
-     */
-    private void handleCountOfRecurringDonationOpps_OpenEnded() {
-        String openStatusValue = System.Label.npe03.RecurringDonationOpenStatus;
-        try {
-            Integer count = [SELECT Count()
-                FROM Opportunity
-                WHERE npe03__Recurring_Donation__r.npe03__Open_Ended_Status__c = :openStatusValue];
-            System.FeatureManagement.setPackageIntegerValue('Data_CountRdOppsOpenEnded', count);
-        } catch (Exception ex) {}
+            Integer count = Database.countQuery('SELECT Count() FROM Error__c');
```
RadSFDeveloper: The `Database.countQuery()` counts up to 50,000 records. If there are more, the following error is generated:
> Too many query rows: 50001 from with count()

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
force2b: I didn't think that a space is required when casting? I treat this something like `:recordId` because the character directly manipulates the var. Different than curly brackets because those are structure control characters.

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
force2b: Nice

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
force2b: True. I was just trying to be consistent in the code.

src/classes/UTIL_OrgTelemetry_SVC.cls
force2b: I just test this in the XXL org against the Opportunity object.

This query timed out: `SELECT Count() FROM Opportunity;`
While this worked fine in ExecAnon: `System.debug(Database.countQuery('SELECT Count() FROM Opportunity'));`

RadSFDeveloper: I take this back, the `countQuery()` did have the limit as the SOQL query on retrieval of 50K records, however by testing total record count in a table resulted in ~133K and ~3M records, no error was raised on LDV XL org.
However, the following query executed in the ExecAnon, but in the code it is used in the batch `start()`, resulted in the > Internal Salesforce Error:
`SELECT Count() FROM Opportunity WHERE npe03__Recurring_Donation__c != null`
The previous file version had `handleCountOfRecurringDonationOpps()` that was working on the LDV orgs, should the same query be used in the batch `start()` method?

force2b: The query handling in a queryLocator in a batch job is a bit different than the complete than the query that existed in `handleCountOfRecurringDonationOpps`. Both technically have the two minute time limit, but batch job queries are meant to be able to handle longer running queries mainly because they don't need return a full data set for the batch job to start. You can see this if you watch the ApexJobs page on a batch job with a lot of batch iterations. The number of batch iterations actually increases as the job runs because it's still compiling the data set while the job is executing. Also, I've tested the new batch job a couple of times in the XXL org and it ran perfectly both times.

PR #3696, opened 2018-10-11
Fix custom object support for Levels and Engagement Plans

### LINKED ISSUE #3687: Engagement Plan on Level not supported on custom object due to invalid name error
[Thread.](https://powerofus.force.com/0D58000004WvNLY)

_Don't troubleshoot this in a scratch org with unmanaged NPSP - must replicate in an org with managed package only._

**Issue:**

An Engagement Plan attached to a Level will cause the Level batch to fail with the error: `Invalid field Custom_Object__c__c for npsp__Engagement_Plan__c`

**Steps to Repeat:**

1. Create custom object, be sure to check to allow for Activities
2. Create custom fields on object: Level (lookup to Level object), and a Number field (to increment the Level)
3. Create an Engagement Plan Template. Add some tasks, it doesn't matter how many or how they're constructed.
4. Edit Level Target field picklist value on Level to include the correct full API name of the custom object, including the __c
5. Create a Level record with the object as the Target. It doesn't matter the number for Min/Max, use the custom number field as the Source field and the Level field as the Level. Select the Engagement Plan template for the Engagement Plan on the Level.
6. Create a record in the custom object, put a value in the number field that is within the Min/Max of the Level.
7. Run the Level batch

**Expected Result:**

The Engagement Plan should fire and populate the tasks.

**Actual Result**

The Level batch fails with the above error.

If you change the Target picklist to not include the __c, such as 'Custom_Object' instead of 'Custom_Object__c' then you get an immediate error upon trying to create the Level.

njjc:
# Critical Changes

# Changes

# Issues Closed
- Fixed #3687

# New Metadata

# Deleted Metadata


njjc:
@dvenor unfortunately, I couldn't write any tests for this bug reproduction requires adding a custom field to the Engagement Plan object and editing the picklist values for the Level object's Target field.

njjc:
@dvenor Thanks for the great feedback, let me know how it looks now.
@judisohn Would you mind doing QA on this one, since you first confirmed the issue?

force2b:
@njjc This PR included a change to the `testdata/1k/test_data.db` file. Was that intended?

njjc:
Nope, my mistake. Is that a problem @davisagli? I accidentally committed changes to the 1k test DB that got merged into master.

davisagli:
@njjc @force2b When you load test data it writes the ids of the inserted records to the db file (to use for lookups). There's no reason that needs to get committed but it's not a problem either; they'll get overwritten on the next load.

PR #3699, opened 2018-10-12
CCI: DevConfig task -- Custom buttons to override underlying packaged versions

force2b:
Includes custom buttons that a dev can simply add to their page layouts to work around the issue with buttons from the underlying managed packages have `npsp__` hardcoded in the URL. All the new buttons start with [NPSP] so they stand out.

Move all custom configuration metadata from the `config` folder into the `unpackage/config` folder.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


config/dev/objects/Campaign.object
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <webLinks>
+        <fullName>DO_NOT_PACKAGE_Household_Mailing_List</fullName>
```
davisagli: Nice :)

config/dev/objects/Campaign.object
```diff
@@ -0,0 +1,16 @@
...
+        <displayType>button</displayType>
+        <encodingKey>UTF-8</encodingKey>
+        <height>600</height>
+        <linkType>url</linkType>
+        <masterLabel>[NPSP] Household Mailing List</masterLabel>
+        <openType>sidebar</openType>
+        <protected>false</protected>
+        <url>/apex/HH_CampaignDedupeBTN?id={!Campaign.Id}</url>
+    </webLinks>
+</CustomObject>
```
davisagli: Would it be worth including a comment indicating the package where the original button is defined so it's easier to track them down to check for updates later?

scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
```diff
@@ -25,8 +25,10 @@ public class CRLP_TEST_VALIDATE_ROLLUPS {
 
     // TODO -- Add support for verifying that dupe OCR's for Soft Credit Contact Roles are not duped in results
 
+    private static Boolean ROLLUPS_INCLUDES_MEMBERSHIP = true;
+
     // When testing in a multicurrency org, set up CAD with this rate: 1.2311
-    private static final Double CURR_EXCHANGE_RATE = 1.2311;
+    private static final Double CURR_EXCHANGE_RATE = 2.0;
```
davisagli: @force2b looks like you've got some unrelated changes here and below...

force2b: I included this just because I meant to check this in earlier, but forgot. It's a script file, so I thought this was close enough. Happy to do as a separate PR too.

jjbennett:
Curious as to why these are being added to the config folder, shouldn't they live in dev_config/src_qa vs the config folder?

scripts/CRLP_TEST_VALIDATE_ROLLUPS.cls
davisagli: Oh yes, I missed that it's in scripts. That's fine.

davisagli:
@jjbennett we renamed `dev_config/src` to `config/dev` with the idea that it can hold additional sets of unpackaged config metadata that aren't necessarily meant for a dev org (there's already QA config, some config that can be optionally installed with another task, etc...). Did we miss something else that the `config` folder was already used for though? We saw it in .gitignore but weren't sure what that was from.

jjbennett:
I'm thinking it's from sublime or mavensmate maybe?
screen shot 2018-10-12 at 5 01 17 pm

davisagli:
@jjbennett oh shoot, yeah it's mavensmate. That's not very nice of it to claim that name instead of, say, a `mavensmate` folder. Okay, I guess we'll have to rename it back to something else.

davisagli:
@jjbennett Jason says that Mavensmate is no longer maintained. Do you think it's okay if we leave the move from `dev_config` to `config` in place?

jjbennett:
I am no longer using MavensMate but I believe @RadSFDeveloper does, will this cause issues if we continue to use MM or do we just need to add specific ignores or?

davisagli:
Hmm, I would guess it might cause issues if someone is still using Mavensmate. @RadSFDeveloper could you do me a favor and make a backup copy of your config directory, then update Cumulus to latest master and see what it does to your config directory?

force2b:
The MavensMate plugin is not even supported by the developer at this point. Seems to me that we probably need to just drop support for it entirely. 

That noted though, `config` is a pretty generic folder that could be used by some other ide down the road. I would prefer to go back to `dev_config` or something that will never stand a chance of conflicting with a valid local folder.

davisagli:
@force2b I'm not too worried about that, but I want to avoid putting "dev" in the name because there are use cases for unpackaged metadata besides just configuring orgs for developers. Maybe we should move these sets of metadata into the `unpackaged` folder (just not in `pre` or `post` which are looked for specially by the installer).

jjbennett:
Hmm how about unpackaged/config?

force2b:
`unpackaged/config` works for me.

.gitignore
```diff
@@ -30,5 +30,4 @@ Referenced Packages
 *.sublime-workspace
 **/.sfdx/
 **/.vscode/
-**/.idea/
-/config/
+**/.idea/
```
force2b: @davisagli @jjbennett Do you think we need to add the ignore back for /config/?

jjbennett: @force2b - if you don't mind. I know it is no longer supported but it still works :)

unpackaged/config/dev/objects/Campaign.object
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jjbennett: Do we want these buttons added to dev or to qa? I am thinking I'd only want them if/when doing qa but dev happens automagically with flow_dev so we'd have to always remember to not commit these changes right?

davisagli: Good point. I was thinking that since they are new components they would not be retrieved automatically, but I was forgetting that buttons are included in the object files rather than separate.

davisagli: @force2b The dev_org flow deploys config/dev. The qa_org flow deploys config/dev and then also config/qa. The latter is probably the right place for metadata that we want to avoid accidentally retrieving and checking in, but that comes with the tradeoff that it's not deployed for dev orgs.

jjbennett: I ran a quick test and it didn't pull down the new buttons but it also doesn't appear to be adding them to the layouts either. I also noticed are retrieve is super dirty in Cumulus. Should we do a PR just to clean up the retrieve, I usually like to do that after a new release anyway since there might be new or changed metadata? Maybe after the api v43 change goes in?

davisagli: Yeah, it's as expected that they don't get added to layouts (would be nice, but it'll take some work). Cleaning up the diff after retrieving sounds like a good separate PR after the api change goes in.

force2b: Not sure what the status on this is. Does this need to get moved to the QA process or is it fine where it is? I did notice that the QA process does not include the OCR and Opportunity.StageName picklist changes that are included in the dev process.

davisagli: @force2b the QA flow deploys both the dev config and the qa config, so it should include that. Did you find that that is not actually working?

force2b: No. I was just looking in the folders, but I see your point the `config_qa` flow runs the same tasks as `config_dev` plus the `deploy_qa_config` and some others.  

When I create a scratch org I run `cci flow run dev_org'. I assume that only runs the `config_dev` tasks. Is that correct; or should I be using a different flow after I create a scratch org?

davisagli: That's correct. It's up to you whether to run the `qa_org` flow instead of `dev_org`: if you do run the `qa_org` flow, you will also get the result of running `deploy_qa_config`, but then it's up to you to make sure you don't accidentally retrieve and commit those changes.

The metadata in `config/dev` should only include changes that won't be retrieved because they aren't included in the NPSP package.

force2b: Thinking about this a little more, as it stands right now the `deploy_dev_config` task is run by both the `config_dev` and `config_qa` flows. Given that, I think we should leave the metadata that does the default npsp config and creates the button overrides in the dev folder so these features get installed into both types of orgs. The new metadata that shouldn't be in the package are well named (at least to me) and we really should be encouraging devs to pay attention to what is being checked into the repo. 

Curious what others think -- @jjbennett @dvenor @njjc @lmeerkatz

jjbennett: My thoughts:
I prefer to have a clean retrieve, if it is something we do not want pulled down then I think it should be in the qa config so we intentionally run the task when we are reviewing a pr vs dev'ing a pr.
I also think it's easier on a newer devs not to have to worry about doing cherry picking and checkouts when stuff is being retrieved that shouldn't be committed.

force2b:
@jjbennett Ok. I moved the new buttons into the `deploy_qa_config` task folder.

jjbennett:
@force2b - this is saying the branch is out of date with master, can you check into it?

davisagli:
@jjbennett probably because https://github.com/SalesforceFoundation/Cumulus/commit/4f8908aae8588bccf148393480b6d6995b5f0dd4 was merged but the build hasn't gotten to the point of automerging back to feature branches yet.

PR #3700, opened 2018-10-13
Add Donations BTN Apex page doesn't respect user permission for opportunities

dvenor:
# Critical Changes
Fixes issue where Read-Only user can add a new donation (opportunity).  Users without "Create" permissions on the opportunity object will now see a message stating that they do not have the correct permissions, and the donation record is not created.
# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -1272,6 +1272,22 @@ If you’re not sure how to resolve these errors, post a message in the Nonprofi
         <shortDescription>Warning</shortDescription>
         <value>Warning</value>
     </labels>
+    <labels>
+        <fullName>RD_ErrorAddDonationPermissionDenied</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>You do not have permission to create a donation record</shortDescription>
+        <value>You do not have permission to create a donation record. Please contact your system administrator for more information.</value>
```
dvenor: @sam-knox - If you get a chance take a look at the wording of this "permission denied" message.  Pretty straightforward but maybe there is something better.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -835,11 +835,24 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
+        test.startTest();
+        
         ApexPages.StandardController scc = new ApexPages.StandardController(r1);
         RD_AddDonationsBTN_CTRL rdadb = new RD_AddDonationsBTN_CTRL(scc);
         system.assertEquals(r1.id, rdadb.sc.getId());
-        rdadb.buttonClick();    
-    }    
+        rdadb.buttonClick();
+       
+        //Read Only user should be denied permission to add donation
```
dvenor: Added an extra bit to an existing test rather than write a new test.   Didn't think this use case was worth the overhead of another test (since our builds are already slow).  Not great BDD test technique but easy enough to break out into it's own test if we think that is clearer.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -835,11 +835,24 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
         
+        test.startTest();
+        
         ApexPages.StandardController scc = new ApexPages.StandardController(r1);
         RD_AddDonationsBTN_CTRL rdadb = new RD_AddDonationsBTN_CTRL(scc);
         system.assertEquals(r1.id, rdadb.sc.getId());
-        rdadb.buttonClick();    
-    }    
+        rdadb.buttonClick();
+
```
dvenor: Added an extra bit to an existing test rather than write a new test.   Didn't think this use case was worth the overhead of another test (since our builds are already slow).  Not great BDD test technique but easy enough to break out into it's own test if we think that is clearer.

dvenor:
**lurch: attach W-024059

src/classes/RD_AddDonationsBTN_CTRL.cls
```diff
@@ -53,18 +53,22 @@ public with sharing class RD_AddDonationsBTN_CTRL {
     public PageReference buttonClick(){
         Savepoint sp = Database.setSavepoint();
         try {
+
```
jjbennett: @mpusto - I think as we work through the security changes, we should try to be consistent in the appearance of the messages. I like the softer touch of the info style warning on this PR but the last payment warning we issued an error toast. Attaching a screen shot of both:
screen shot 2018-10-15 at 11 38 57 am
screen shot 2018-10-15 at 11 34 52 am

@dvenor - Either way, I think we should stick with the page header and eliminate the need for creating a separate button to return to the recurring donation record.

src/classes/RD_AddDonationsBTN_CTRL.cls
```diff
@@ -74,6 +78,16 @@ public with sharing class RD_AddDonationsBTN_CTRL {
         }
     }
     
+    /*******************************************************************************************************
+    * @description Action method to redirect to the Recurring Donation calling page
+    * @return PageReference to the calling page
+    */ 
+    public PageReference redirectToRecurringDonationsPage() {
+        PageReference pageRef = sc.view();
+	    pageRef.setRedirect(true);
```
jjbennett: Looks like a tab snuck in here.

src/classes/RD_RecurringDonations_TEST2.cls
jjbennett: Can we create a separate Test for the CTRL vs having one method in another test to test the controller and follow the new standards from there, we should be able to enable parallel testing on the new test class.

src/classes/RD_AddDonationsBTN_CTRL.cls
```diff
@@ -53,18 +53,22 @@ public with sharing class RD_AddDonationsBTN_CTRL {
...
+            RD_RecurringDonations.evaluateRecurringDonationsForNewOppInsert(new list<npe03__Recurring_Donation__c>{recurringDonation});
             //This recalculates Rollups only if Customizable Rollup is enabled.
-	        recalculateRollups(sc.getId());
-
-	        PageReference pageRef = sc.view();
-	        pageRef.setRedirect(true);
-	        return pageRef;
+            recalculateRollups(sc.getId());
+            
+	        return redirectToRecurringDonationsPage();
```
jjbennett: Another tab.

src/classes/RD_AddDonationsBTN_CTRL.cls
```diff
@@ -53,18 +53,22 @@ public with sharing class RD_AddDonationsBTN_CTRL {
     public PageReference buttonClick(){
         Savepoint sp = Database.setSavepoint();
         try {
+
+            if (!Schema.SObjectType.Opportunity.isCreateable()) { 
+                ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.INFO, System.Label.RD_ErrorAddDonationPermissionDenied));
+                return null;
+            }
+
             system.assert(sc.getId() != null);
```
jjbennett: What do you think about cleaning up this method a bit while you are in here? Since we are adding an error message  with proper display maybe we can handle this scenario better as well?
Can we also fix some of our other best practices as far as capitalization / formatting, I believe we are sticking with only cleaning up the method and only what will not change the behavior but I'm still fuzzy on what we are allowed to refactor when working on the legacy code. Here is how this error looks now if we navigate to the page without an id. 
screen shot 2018-10-15 at 11 48 37 am

force2b: Where recurring donations are involved we should tread very carefully.

src/classes/RD_AddDonationsBTN_CTRL.cls
shannonsans: We should probably use the illustration pattern in Lightning Design System for access errors: https://lightningdesignsystem.com/components/illustration/#No-Access

Though it would be better if they don't see the button to create the record in the first place (recognizing the platform doesn't necessarily allow us to set perms on a custom button based on profile etc.)

src/classes/RD_RecurringDonations_TEST2.cls
dvenor: Created a couple of new tests for the controller.  One tests when the user does not have permissions to the opportunity.  Another test if the VF page standard controller doesn't have a recurring donation ID.

src/classes/RD_AddDonationsBTN_CTRL.cls
dvenor: I went with the same look as Jenny's payment example, except displayed the message as info instead of an error.

src/classes/RD_AddDonationsBTN_CTRL.cls
dvenor: Mostly left the method alone but did make a change so that the error is handled more elegantly when navigating without an id.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -835,11 +835,91 @@ private with sharing class RD_RecurringDonations_TEST2 {
         npe03__Recurring_Donation_Field__c = 'Name'
         );
         
+        test.startTest();
```
seanpat09: `test` should have a capitalized t (i.e. `Test.startTest()`) since it's a class name

src/labels/CustomLabels.labels
```diff
@@ -1272,6 +1272,30 @@ If you’re not sure how to resolve these errors, post a message in the Nonprofi
...
+        <shortDescription>Refresh Opportunities</shortDescription>
+        <value>Refresh Opportunities</value>
+    </labels>
+        <labels>
+        <fullName>RD_ErrorAddDonationMissingId</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Recurring Donation ID missing</shortDescription>
+        <value>Recurring Donation ID missing.  Please select a valid recurring donation record.</value>
```
seanpat09: I think there's two spaces after the first period here

dvenor:
@seanpat09 - Thanks for the review.  I made the changes on the items you found.  Good eye.  If you think it looks good and approve, we can send it off to Rick for testing and hopefully get it in before the merge freeze.

seanpat09:
@rponti-sforg This is ready for QA!

dvenor:
@sam-knox - The message appears when, on a Recurring Donation record, a read-only user clicks "Refresh Opportunities".  They lack the create opportunity permission and see the message.  I went with the Info style message instead of an Error style message since this wasn't actually an error condition.

sam-knox:
@dvenor - Ok, thanks. So we're checking the Create permission on Opportunities, anything else? Does the user also need View or Edit? I want to change the release note so it's clear exactly what the perms are.

sam-knox:
@dvenor - Sorry, just want to clarify the error message. I see Line 1297 displayed when a read-only user tries to use Refresh Opportunities but what about Line 1289? Does that appear in the UI at all?

Partly I'm asking here because if the issue is that the user needs to select a Recurring Donation, I'd rather say it like that instead of telling them that there's an ID missing.

dvenor:
@sam-knox - Ah, yes.  So, that error message handles a completely different, and rare, scenario where the user somehow navigates to the 'Refresh Opportunities' page without a RD id in the URL.  In that scenario the page was showing a cryptic, non-standard, error message.  I changed it to show the label text you are seeing on line 1289, with the standard formatting.  There really is no reason a user would do this, however, it is technically possible if the user mucks with the URL.  This was an edge case scenario fixed since we were already in the code.

dvenor:
@sam-knox - Users must have Create permissions on the opportunity or else they will see this permission denied message.  I'll update the critical changes section to be specific about this.  If they had no object level permissions they would see the same message.

PR #3701, opened 2018-10-15
Feature/bge lc  refactor bdi field mappings as custom metadata types with references

mdeleon0:
# Critical Changes
Modified class BDI_DataImportService
# Changes
Created Custom Setting "BDI Settings"
Created Custom Metadata "BDI Field Mapping"
# Issues Closed
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x80000007fOUpAAM/view
# New Metadata
BDI Settings
BDI Field Mapping
# Deleted Metadata
No


src/classes/BDI_DataImportService.cls
```diff
@@ -704,62 +704,119 @@ global with sharing class BDI_DataImportService {
...
+        String dataImportField = parametersByName.get('dataImportField');
+        if (targetFieldName != null && targetFieldName != '') {
+            List<String> parts = targetFieldName.split('\\.', 2);
+            if (parts.size() == 2 && parts[0] == dataImportObjectName && parts[1] != '') {
+                String fieldName = parts[1];
+
+                //if we aren't in the npsp namespace, remove those tokens from help text
+                //also removes tokens from actual detected namespace, if we happen to be
+                //in a non-npsp namespace
+                if (UTIL_Namespace.getNamespace() != 'npsp') {
```
lmeerkatz: I think this would only be true if we're in a non-namespaced org, in which case if `String.isBlank(UTIL_Namespace.getNamespace())` would be clearer. It's a bit confusing because of the way UTIL_Namespace.getNamespace() works, unless we include that class in every package we develop? @force2b Do you have any insight here?

src/classes/BDI_DataImportService.cls
```diff
@@ -704,62 +704,119 @@ global with sharing class BDI_DataImportService {
...
 
-                    dataImportFieldToTargetField.put(dataImportField, fieldName);
+                // make sure they have permissions to modify the target field
+                // we can't do this in all tests, because many profiles won't have FLS set for all our new fields!
+                if (!Test.isRunningTest() || failFieldLevelSecurity) {
+                    DescribeFieldResult fieldDescribe = UTIL_Describe.getFieldDescribe(targetObjectName, fieldName);
+                    Set<String> setAuditFields = new Set<String>{
+                        'createddate', 'createdbyid', 'lastmodifieddate', 'lastmodifiedbyid'
+                    };
+                    Boolean isAuditField = setAuditFields.contains(fieldName.toLowerCase());
```
lmeerkatz: This logic is fairly complex. I'd consider moving it to its own method.

src/classes/BDI_DataImportService.cls
```diff
@@ -704,62 +704,119 @@ global with sharing class BDI_DataImportService {
...
+                    if (failFieldLevelSecurity ||
+                        fieldDescribe == null ||
```
lmeerkatz: This logic is fairly complex. I'd consider moving it to its own method.

src/classes/BDI_DataImportService.cls
mdeleon0: I like the idea, will do this change and test.

mdeleon0:
moving this branch to do-not-merge

PR #3702, opened 2018-10-15
Fixing null string exception.

njjc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_ContactService.cls
+++ b/src/classes/BDI_ContactService.cls
@@ -367,7 +367,11 @@ public with sharing class BDI_ContactService {
         strSoql += ' WHERE AccountId IN :setHHId';
         List<Contact> listHHContacts = database.query(strSoql);
         for (Contact con : listHHContacts) {
-            mapDIKeyToContact.put(con.Firstname.toLowerCase() + '|' + con.AccountId, con);
+            String firstName = con.FirstName;
+            if (!String.isBlank(firstName)) {
+                firstName = firstName.toLowerCase();
+            }
+            mapDIKeyToContact.put(firstName + '|' + con.AccountId, con);
         }
     }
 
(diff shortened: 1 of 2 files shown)


lmeerkatz:
@njjc Can you attach this to its work item? (**lurch:attach W-0000000)

lmeerkatz:
**lurch:attach W-029124

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fOhUAAU" target="_blank">W-029124</a>

lmeerkatz:
@njjc It looks like there's a test failure. It looks like a row lock problem to me. Do you agree? If so, do you resolve those with manual rebuilds or do you update tests to stabilize them? https://mrbelvedereci.herokuapp.com/builds/39595/tests

njjc:
@lmeerkatz that's strange, I've never seen a row lock in apex tests before. Generally when there are one time intermittent failures we will just rerun, and if they happen often enough we add a backlog item to fix it.

force2b:
@njjc I have seen that on occasion. It's part of the data rollback process, and effectively the main reason we have issues with moving to parallel unit tests. What _should_ happen is that each test method has its own set of data isolated from all other unit tests, but what _actually_ happens is that the data is isolated, but the object index is shared. 

I've been told that when one test is trying to rollback data and the index is being updated, it can cause a lock error in another unit test that may be updating the index with the same key value. That's the primary reason why I've ben pushing to use unique Name values in all of our unit tests. Of course, even with unique values I've still seen some occasional lock errors against objects such as OpportunityContactRole that shouldn't be subject to this problem.

davisagli:
@njjc @lmeerkatz heads-up that since this was merged, there is a failing test in the beta builds: BDI_DataImport_TEST.OneDIWithoutFirstName

Stack Trace:
Class.npsp.BDI_DataImport_TEST.OneDIWithoutFirstName: line 160, column 1

Message:
System.AssertException: Assertion Failed: Expected: 1, Actual: 2

njjc:
Thanks @davisagli I'll check it out.

PR #3703, opened 2018-10-15
Origin/feature/robot

skristem:
# Critical Changes

# Changes
removed unnecessary sleep and used APIs to create test data for automation scripts wherever possible
# Issues Closed

# New Metadata

# Deleted Metadata


tests/browser/affiliations/change_secondary_as_primary_affiliation.robot
```diff
@@ -14,15 +14,16 @@ Create Secondary Affiliation for Contact
...
     Click Id    &{account}[Name]
-    Sleep    5
+    #Sleep    5
     Scroll Page To Location    0    400
     Click Edit Button    Edit Primary
-    Sleep    5
+    #Sleep    5
     Select Checkbox    //div[contains(@class, "uiInput")]/label/following-sibling::input[@type='checkbox']
     Click Record Button    Save
     Sleep    5
```
jessingrass: Not sure if you wanted to comment this one out too, or if this sleep is needed.

davisagli: I don't think the one on line 25 is needed, but I think we do need something to wait on line 22.

The difference is that clicking the save button triggers an XMLHTTPRequest, so the built-in wait-for-requests-to-complete-after-click logic should take care of waiting for that one. On the other hand, clicking on the edit button does not trigger a request, but does take some time to render the edit version of the checkbox. It would be better to have a line that waits for the checkbox locator to appear instead of an arbitrary sleep though.

skristem: The one on Line 25 is included on purpose. Once I hit save the change is not reflecting immediately when I refresh page to verify under details tab. I am having to wait for 5 seconds before I verify. I have run the test case multiple times and I do not see any issue with removing the sleep after clicking edit button. @davisagli if you think it would be good to add delay there I have a keyword to wait for locator and I can include that in the code.

davisagli: @skristem okay we can leave it as is for now. I have some more testing to do and will let you know if I still have trouble with it.

tests/locator_w19.py
```diff
@@ -94,12 +96,20 @@
...
         'no_payments':'//tbody/tr/td[3]',
         'pay_amount':'//tbody/tr[{}]/td[3]/span/span[text()="{}"]',
         'check_occurance':'//h2/a/span[@title="{}"]/following-sibling::span',
+        'text':'//*[@id="j_id0:vfForm:j_id76:util_formfield:inputx:util_inputfield:inputX"]',
+        },
+    'npsp_settings':{
+        'panal_link':'//div[@id="idPanelRecDonationsNav"]//a[text()="{}"]',
+        'panal_sub_link':'//ul/li/a[text()="{}"]',
+        'field_value':"//div[@class='slds-form-element'][./label[contains(text(),'{}')]]/div/span",
+        'side_panal':"//ul/div[contains(@id,'RecDonations')]/button[1]",
```
davisagli: These should be spelled "panel"

skristem: thanks, will update in next checkin

skristem: is this PR ok to merge?

davisagli: That's fine with me.

PR #3708, opened 2018-10-17
Origin/feature/robot

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/tests/NPSP.robot
+++ b/tests/NPSP.robot
@@ -35,16 +35,32 @@ API Create Opportunity
     &{opportunity} =     Salesforce Get  Opportunity  ${opp_id} 
     [return]         &{opportunity}  
  
-API Create Household Account
+API Create Organization Account
     [Arguments]      &{fields}
     ${name} =        Generate Random String
-    ${rt_id} =       Get Record Type Id  Account  HH_Account
+    ${rt_id} =       Get Record Type Id  Account  Organization
     ${account_id} =  Salesforce Insert  Account
-    ...                  Name=${name} HouseHold
+    ...                  Name=${name}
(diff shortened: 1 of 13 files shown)


tests/NPSP.robot
```diff
@@ -141,33 +157,32 @@ Create HouseHold
...
+    # &{contact} =  API Create Contact    Email=skristem@robot.com
+    Go To Record Home  ${con_id}
+    Wait For Locator    record.related.title    Volunteer Hours 
     Scroll Page To Location    50    400
     Click Related List Button   Organization Affiliations    New
-    Populate Lookup Field    Organization    &{account}[Name]
+    Populate Lookup Field    Organization    ${acc_name}
     Click Modal Button    Save
-    [Return]         ${account_id}    ${contact_id}
+    #[Return]         &{account}[Id]    &{contact}[Id]
```
davisagli: Are these keywords still used or is everything using the API version now?

skristem: these keywords are still being used in the test cases that actually tests create contact, create org, create primary affliation etc. In the test cases where these are pre requisites for creating the actual test case, I am using API version.

davisagli: That makes sense, and explains why we no longer need the Return. Thanks.

My preference would be to remove code entirely rather than leaving commented code, since we can always find older versions of the code in github. However I can understand if you'd rather keep bits of commented code around while this work is still in progress, so I'll leave it up to you.

skristem: I am not that good with git so left those as comments for now since we are still making changes. Once we finalize all the code, I will remove the comments.

PR #3711, opened 2018-10-18
Api V43: Fix payments page issue

force2b:
From @RadSFDeveloper: Keep payment fields in the columns so they do not shift left or right when an *input* field is clicked on or hovered

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


jjbennett:
When clicking Create Scheduled: 
I noticed that the interval fields are now stacked and the remaining balance is shifted when not in full screen, even when first loaded.
screen shot 2018-10-18 at 5 09 02 pm
screen shot 2018-10-18 at 5 08 51 pm

Running additional tests.

PR #3712, opened 2018-10-18
Default Batch and Template Settings

randi274:
# Critical Changes

# Changes
- Add default processing settings and suggested active fields to both Batch and Batch Template creation from the shared wizard.
# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
**lurch - attach W-028588

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fNyZAAU" target="_blank">W-028588</a>

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -164,6 +178,117 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
...
+        return model;
+    }
+
+    /*******************************************************************************************************
+    * @description Sets default values on BGEDetails Model to avoid returning any null values
+    */
+    private static void setDefaultValuesOnBGEDetailsModel(BGEDetailsModel model) {
+
+        if (model.expectedCount == null) {
+            model.expectedCount = 0;
```
lmeerkatz: Should these values come from the default model? Something like:

```
BGEDetailsModel defaultModel = getDefaultBGEDetailsModel();
if (model.expectedCount == null) {
    model.expectedCount = defaultModel.expectedCount;
}
```

You could probably even iterate through fields on either model instead of referencing each one individually.

src/classes/BGE_ConfigurationWizard_CTRL_TEST.cls
```diff
@@ -112,17 +112,21 @@ private class BGE_ConfigurationWizard_CTRL_TEST {
...
-        String activeFieldsJSON = '[{' +
-                '"label":"Name",'+
-                '"name":"npsp__Account1_Name__c",'+
-                '"sObjectName":"Account1",'+
-                '"defaultValue":null,'+
-                '"required":false,'+
-                '"hide":false,'+
-                '"sortOrder":0'+
-                '}]';
+        String activeFieldsJSON = '[{"label":"Donation Amount","name":"Donation_Amount__c","sObjectName":"Opportunity","defaultValue":null,"required":true,"hide":false,"sortOrder":0},' +
```
lmeerkatz: Add line breaks for readability

src/classes/BGE_ConfigurationWizard_CTRL.cls
bethbrains: that is....a much more elegant implementation. on it.

PR #3717, opened 2018-10-19
Fix flaky beta build test

njjc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_DataImport_TEST.cls
+++ b/src/classes/BDI_DataImport_TEST.cls
@@ -139,6 +139,12 @@ public with sharing class BDI_DataImport_TEST {
     *        contact matched
     **********************************************************************************************************/
     static testMethod void OneDIWithoutFirstName() {
+        // this test won't pass if middle name or suffix is enabled due to a
+        // salesforce won't fix issue: https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM
+        if (UTIL_Describe.isMiddleNameEnabled() || UTIL_Describe.isNameSuffixEnabled()) {
+            return;
+        }
+
         Data_Import_Settings__c diSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
         diSettings.Contact_Matching_Rule__c = 'Firstname,Lastname,Email';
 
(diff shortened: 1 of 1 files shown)


njjc:
@lmeerkatz @davisagli this is good to go and _should_ fix the beta build.

src/classes/BDI_DataImport_TEST.cls
```diff
@@ -139,20 +139,28 @@ public with sharing class BDI_DataImport_TEST {
     *        contact matched
     **********************************************************************************************************/
     static testMethod void OneDIWithoutFirstName() {
-        insert new DataImport__c(
-            Contact1_Lastname__c = 'foo',
-            Contact1_Personal_Email__c = 'foo@bar.com'
-        );
+        // this test won't pass if middle name or suffix is enabled due to a
+        // salesforce won't fix issue: https://success.salesforce.com/issues_view?id=a1p30000000eQRxAAM
```
lmeerkatz: Delightful! But also, thanks for the heads up.

lmeerkatz:
So to do QA on this, we need to set Contact_Matching_Rule__c to 'Lastname,Email', then try to process a DI with only LastName and Personal_Email__c populated in an org where middle name is enabled, correct? What org setup process would we use for that? @njjc @davisagli

njjc:
Really the purpose of this test is to make sure a DI can be processed with a null first name, as there was a null pointer exception and this is just a regression for that. The middle name and suffix stuff was just extra precautionary measures after spending 4 hours down a rabbit hole of the test failing in a managed beta org. I would just merge and see if the managed beta build passes; if so we're good to go.

lmeerkatz:
@njjc Agreed that what you've got fixes the test issue. I'm just wondering if there is some weird edge case with middle name that would block processing on a contact with a missing first name. I'm fine with merging this in and testing after the fact to see if there are additional bugs to address.

njjc:
Right, any such issues would be outstanding and not newly introduced with the case sensitivity changes. I'm really just trying to avoid getting lost down any more rabbit holes!

lmeerkatz:
What's the best way I could set up an org to identify a potential existing problem when middle name is enabled?

njjc:
I believe it has to be done through [a support case](
http://docs.releasenotes.salesforce.com/en-us/spring15/release-notes/rn_general_middle_name_suffix_enable.htm
).

davisagli:
@njjc @lmeerkatz you can get a scratch org with middle name and suffix fields using the "IsMiddleNameEnabled" org preference. There is an existing beta build which has this turned on.

lmeerkatz:
Perfect. Thanks @davisagli!

davisagli:
I'm gonna hit merge so we can see what happens on the betas.

PR #3718, opened 2018-10-19
[DO NOT MERGE] Multiple Tributes 

### LINKED ISSUE #3787: Multiple Tribute - Remove current notification recipient fields, change relationship to Opportunity to Lookup (Instead of Master Detail)
Based on feedback from @njjc, we want to make the following changes to Multiple Tributes before incorporating them into NPSP:

- Remove Notification Recipient fields
- Change Opportunity Master Detail Relationship to Lookup
- Update code appropriately based on changes.  Note: additional query will be enforcing rules concerning the different tribute modes.  This was formerly relying on a rollup summary field, which will also be removed.

weizsforg:
# Critical Changes

# Changes

# Issues Closed
Fixes #3787

# New Metadata
Tribute object

Custom Settings: npe01__Contacts_And_Orgs_Settings__c: Multiple Tributes - Used to enable Multiple Tributes support, indicate how it is supported

# Deleted Metadata

# Testing Notes
Scenario Based Testing relating to Tributes Enablement via Multiple Tributes Setting:
- Single - will not allow related tributes to be added to any Opportunity.  Batch conversion to "Multiple Tributes" is disabled on settings screen
- Mixed - will allow Opportunities to have Tribute information on either the Opportunity, or related Tributes, but not both.  Batch conversion to "Multiple Tributes" is enabled on settings screen, but will display an error indicating that it cannot be run in "Mixed Mode"
- Multiple - will not allow Opportunities to have Tribute information on Opportunity.  Batch conversion to "Multiple Tributes" is enabled on the setting screen and can be run.   New Opportunities with tribute info on the Opportunity will automatically move the information to a related Tribute and clear the Opportunity fields

Testing relating to Opportunity Contact Roles:
- Hierarchy of Primary Contact, then Honoree, then Notification Recipient Opportunity Contact Role (OCR), meaning, if the same Contact is associated with multiple lookups on the Opportunity or related Tributes, the highest precedence wins
- Roles only matter if set in Settings.  Any role that is not set is skipped evaluation
- For any given contact, there should only be one OCR record, even if there are 5 tribute records, and a Notification Recipient all pointing at the same Contact.
- If an honoree is removed, or the contact is changed, and there are no other honorees referencing that same contact, the honoree role is removed-- unless they were/are the Notification Recipient.  in this case, the role changes to Notification Recipient.
- Similarly, if the Primary Contact is changed, and that Contact exists in one of the lower precedence fields, the role will change to that value, and a new OCR will be added for the new Primary Contact (possibly removing another OCR if the new Primary Contact existed in a lower precedence field).


src/triggers/TDTM_Tribute.trigger-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexTrigger xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>41.0</apiVersion>
```
force2b: This needs to be API v43.

src/classes/OPP_CreateTribute_BATCH.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>41.0</apiVersion>
```
force2b: This needs to be API v43.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
```diff
@@ -95,6 +95,18 @@
...
+    <fields>
+        <fullName>Multiple_Tributes__c</fullName>
+        <description>Allows multiple Tributes per Opportunity.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>Valid values: 1) &quot;Allow only one tribute per Opportunity record&quot; 2) “Allow only one tribute per Opportunity, or multiple tributes on Tribute records” 3) “Allow only multiple Tribute records and migrate Tribute data entered on Opportunities to Tributes&quot;</inlineHelpText>
+        <label>Multiple Tributes</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
```
force2b: Should this field be a simple Integer with either 1, 2, or 3? My concern about storing the entire picklist string value is that it might break if the User translates the picklist values (which will eventually be done as part of the normal package). It's worth testing this to make sure everything works with the labels are translated and you have Users in multiple languages.

src/objects/Tribute__c.object
```diff
@@ -0,0 +1,219 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: The new object will require translation files for all the supported languages. Without them in the branch, the package will not build. @jjbennett can give you more details on the process to follow to ensure the translation files for this new object are added.

force2b:
@weizsforg I know this is still in draft and I haven't started a formal review of this, but I added a whole bunch of comments for you on items to change or consider. Let me know if you have any questions on these.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
jbrandolinisf: @juliesnow - What is your take on this one?  Should be change this?

src/objects/Tribute__c.object
jbrandolinisf: @jjbennett - Can you explain to me what I need to do for the new objects?  I assume if I added a new setting, and additional custom labels, those need I8N as well?  @force2b suggested I reach out to you!  Thanks for any assistance you can provide!!!

jbrandolinisf:
@weizsforg - I cannot add you as a reviewer, since you are the originator, and I am still taking care of minor text updates, and the items @force2b mentioned.  That said, if you could look over the code changes I made, and indicate if you have concerns, I would appreciate it.

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -912,4 +912,217 @@ private class OPP_OpportunityContactRoles_TEST {
         System.assertNotEquals(null, result[0].Opportunity.AccountId);
     }
 
+    /******************************************************************************************************
```
force2b: These are a lot of new test methods to add to an already large test class. Could these belong in a different and related (smaller) test class, or maybe they belong in a new class entirely to group the testing logic together.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
force2b: We have found bugs with storing picklist and record type labels in custom settings when it comes to translations. Sticking with a simple integer will avoid that and allow the product to work even if someone changes the label content (which anyone can do in any language). I'm **very** highly recommending this change.

force2b: Also, this field will default to `null` for all existing customers. I didn't notice a _default_ value/interpretation for `null` in the code, though I might have missed it.

jbrandolinisf: null to default is great callout.  Am wrapping that and returning the Single value now.  I will speak with the team about the 1,2,3 versus Strings-- totally get what you are saying...and will likely incorporate.  Will respond back if there is significant pushback with why.

jbrandolinisf: Yeah, the big place this breaks down is in multilingual orgs, correct?  This is because the stored value will only be of one language, but the labels will rendered in the user's language, if I am understanding this correctly?  If that is the case, I definitely agree this has to be changed.

sam-knox: @force2b  - In the inline help text, there's a combination of both encoded quotes and keyed quotes (&quot; and "). Is that an issue at all? What's our convention for special characters anyway?

force2b: Great catch @sam-knox. These should all be encoded for consistency, but it's not earth-shattering if they're missed. It seems that the build and packaging process is able to work with these without failing. Still we should follow best practices on encoding.

jbrandolinisf: @sam-knox, @force2b - Can you indicate which file/field this is?  I could not find the mixed mode stuff, only the encoded ones.

jbrandolinisf: @force2b - Please re-review.  With the exception of breaking out the tests from the OPP_OpportunityContactRoles_TEST class, the internationalization of the new object, and the encoding issue, the other requested changes are in.

src/objects/Tribute__c.object
jbrandolinisf: @force2b - Translation files are now present.  I also brought over the translation stubs for the newly added Multiple Tribute related labels.

force2b: Thanks!

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
sam-knox: @jbrandolinisf - The place I saw the encoding issue was on line 102 in npe01__Contacts_And_Orgs_Settings__c.object.

jbrandolinisf: Confirmed the encoding issue was from a prior version, and is no longer present with @sam-knox.

jbrandolinisf:
@force2b - Do you have any further feedback for me on this?  I know we cannot merge in until after the freeze is over, but I want to confirm whether I need to do any further work on this before QE looks at it.  Thanks!

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
```diff
@@ -95,6 +95,18 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Multiple_Tributes__c</fullName>
+        <description>Allows multiple Tributes per Opportunity.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>Specify that your org can have one tribute per Opportunity (single), one per Opportunity and multiple tributes in separate Tributes (mixed), or multiple tributes in separate Tribute records. Choosing multiple creates Tributes only for new Opportunities.</inlineHelpText>
+        <label>Multiple Tributes</label>
+        <length>255</length>
```
force2b: Was this settled on as a single string word - and one that is not defined in a custom label. If so, I think the field length should be around 10. Also, how is null handled by the code and should that be referenced in the description (not inline help text since that would be visible to a user)?

src/package.xml
```diff
@@ -2447,6 +2486,7 @@
         <members>General_Accounting_Unit__c.All</members>
         <members>Level__c.All</members>
         <members>Opportunity.To_Be_Acknowledged</members>
+        <members>Tribute__c.All</members>
```
force2b: I don't think there should be a list view included for the Tribute object. There is no packaged tab for this.

src/classes/OPP_CreateTribute_BATCH.cls
```diff
@@ -0,0 +1,114 @@
...
+* @date 2018
+* @group Tribute
+* @description Batch Class for creating Tribute from Opportunity.
+*/
+public class OPP_CreateTribute_BATCH implements Database.Batchable<sObject> {
+
+    /*******************************************************************************************************
+    * @description Opportunity Query used to find Opportunities to Convert
+    */ 
+    private static final String OPP_QUERY = 'SELECT Id, Tribute_Type__c, Honoree_Contact__c, Honoree_Name__c ' +
```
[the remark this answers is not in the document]
jbrandolinisf: Ok, makes sense.   I will move the null check to the batch, and skip ones that do not match.  I am leaning toward requerying for the data too, for each batch set-- if this is running over millions of records it is possible it was updated between when the query ran and the batch is processing.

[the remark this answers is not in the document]
jbrandolinisf: @force2b - If I do that, it is constraint in size to the set of records I am processing.  I can do an if in a loop, and build a new collection (after retrieving them, or I can just filter out the ones that are null (since it is no longer unbounded).  What is your preference?

src/classes/OPP_CreateTribute_BATCH.cls
```diff
@@ -0,0 +1,114 @@
...
+            newTributes.add(newTribute);
+
+            //clear the Opportunity fields
+            opp.Tribute_Type__c = null;
+            opp.Honoree_Contact__c = null;
+            opp.Honoree_Name__c = null;
+
+        }
+        
+        //We do not want to run through opportunity trigger when clear Tribute Fields
```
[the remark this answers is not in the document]
jbrandolinisf: We are recommending turning off all Opportunity and Tribute Triggers, so none.

If we have to scale that back, we need the OCR one off at a minimum

src/classes/OPP_CreateTribute_TEST.cls
```diff
@@ -0,0 +1,152 @@
...
+        //Make sure there is no Tribute record in the org
+        List<Tribute__c> tributeList = [Select Id from Tribute__c];
+        System.assertEquals(0, tributeList.size(), 'No Tributes Expected');
+        
+        //Create and Insert Opportunity
+        Account acc = new Account(Name = 'TheAccount');
+        insert acc;
+        
+        List<Opportunity> oppsToInsert = new List<Opportunity>();
+        for(Integer i = 0; i < 10; i++) {
```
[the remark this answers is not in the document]
jbrandolinisf: @force2b - IMO, we should look at creating rules in github to auto flag things like this so reviewers don't have to...

src/classes/OPP_CreateTribute_TEST.cls
```diff
@@ -0,0 +1,152 @@
...
+        /******************************************************************************************************
+    * @description Test the batch process for Tribute creation
+    ********************************************************************************************************/
+    static testMethod void testCreateTributeBatchWithAdditionalWhereClause() {
+        //Make sure there is no Tribute record in the org
+        List<Tribute__c> tributeList = [Select Id from Tribute__c];
+        System.assertEquals(0, tributeList.size(), 'No Tributes Expected');
+        
+        //Create and Insert Opportunity
+        Account acc = new Account(Name = 'TheAccount');
```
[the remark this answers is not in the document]
jbrandolinisf: Ok, I can do that, but considering this is a test class with no data by default, it should be ok.

src/classes/TBU_OpportunityContactRoles_TEST.cls
```diff
@@ -0,0 +1,310 @@
...
+        Tribute__c newTri = new Tribute__c(Opportunity__c = opp.Id, Honoree_Contact__c = cons[1].Id);
+
+        Test.startTest();
+        insert newTri;
+
+
+        cons = [SELECT Id, Name FROM Contact WHERE Id IN :cons];
+
+        newTri = [SELECT Id, Honoree_Name__c FROM Tribute__c WHERE Id = :newTri.id];
+        system.assertEquals(cons[1].Name, newTri.Honoree_Name__c);
```
force2b: Still a number of these `system` statements in lower case.

src/classes/OPP_CreateTribute_BATCH.cls
```diff
@@ -0,0 +1,129 @@
...
+
+    /*******************************************************************************************************
+    * @description Checks to see if there is Tribute Information on the Opportunity
+    * It returns true if one or more of the following fields null/blank: Honoree_Contact__c (null only),
+    * Honoree_Name__c (null/blank), Tribute_Type__c (null/blank).
+    * @param Opportunity opp - The opportunity to check
+    * @return Boolean true if Tribute info is present, false otherwise
+    ********************************************************************************************************/
+    @testVisible private static Boolean hasTributeInfo(Opportunity opp) {
+        return ((opp.Honoree_Contact__c != null) ||
```
force2b: I think this might be easier to read using isNotBlank(): `String.IsNotBlank(opp.Honoree_Contact__c)`
https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_methods_system_string.htm

src/classes/OPP_CreateTribute_BATCH.cls
```diff
@@ -0,0 +1,129 @@
...
+                opp.Honoree_Contact__c = null;
+                opp.Honoree_Name__c = null;
+            }
+        }
+        
+        //We do not want to run through Opportunity or Tribute triggers when clearing Tribute Fields
+        //on the Opportunity
+        TDTM_TriggerHandler.disableTDTM = true;
+        insert newTributes;
+        update scope;
```
force2b: This will still attempt to update every single Opportunity, even if it was not modified. I recommend creating a new List<Opportunity> for changed Opps that is updated at the end (if the collection isn't empty).

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -381,53 +459,100 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+            if (tributesByOpp != null && tributesByOpp.values().size()>0) {
+                for (Integer j = 0; j < tributesByOpp.get(newOpp.Id).size(); j++) {
+                    Tribute__c newTri = tributesByOpp.get(newOpp.Id)[j];
+                    Tribute__c oldTri = oldTributesByOpp.get(newOpp.Id)[j];
+                    if (needsManageOCR(newTri, oldTri, FIELD_NAME_HONOREE_CONTACT, newOpp.Primary_Contact__c)) {
+                        mapOppIdMapConIdOCR.put(newOpp.Id, new Map<Id, OpportunityContactRole>());
+                    }
+                }
+            } else {
+                if (needsManageOCR(newOpp, oldOpp, FIELD_NAME_HONOREE_CONTACT, newOpp.Primary_Contact__c) ||
```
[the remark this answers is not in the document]
jbrandolinisf: Yes

src/classes/STG_PanelMultipleTribute_CTRL.cls
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public String getMultipleTributesSettingForDisplay() {
+        return TBU_MultipleTributesSettingsFacade.getMultipleTributesSettingLocalized();
+    }
+
+    public List<SelectOption> getMultipleTributesOptions() {
+        List<SelectOption> mtSettingsOptions = new List<SelectOption>();
+
+        mtSettingsOptions.add(new SelectOption(TBU_MultipleTributesSettingsFacade.TRIBUTE_SINGLE_VAL, Label.stgMultipleTributesSingleOpp));
```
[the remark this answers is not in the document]
jbrandolinisf: I would need to add a new method, as this only returns the one for the current value, not for each of the potential values to select from.

src/classes/STG_PanelMultipleTribute_TEST.cls
```diff
@@ -0,0 +1,190 @@
...
+* @date 2018
+* @group Settings
+* @description Unit tests for STG_PanelMultipleTribute_CTRL class
+*/
+@isTest
+private class STG_PanelMultipleTribute_TEST {
+    /*******************************************************************************************************
+    * @description Role values to use in settings.
+    */
+    private static final String OPPORTUNITY_CONTACT_ROLE_DONOR = 'Donor';
```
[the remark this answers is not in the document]
jbrandolinisf: I did this, but had to make the class public to implement it, or it would not be able to see the member variables even with "@TestVisible"

jbrandolinisf:
@force2b - Completed modifications from review, and pending bugs.   Ready for re-review.

src/classes/TBU_MultipleTributesSettingsFacade.cls
```diff
@@ -0,0 +1,124 @@
...
+    /*******************************************************************************************************
+    * @description Returns the MultipleTributesSetting, or a default of Single Tribute, if setting value is null;
+    * @return Value of Multiple Tributes Setting, or Single if null
+    ********************************************************************************************************/
+    public static String getMultipleTributesSetting() {
+        npe01__Contacts_And_Orgs_Settings__c mtSettings = getSettings();
+        
+        String val = (mtSettings != null) ? mtSettings.Multiple_Tributes__c : null;
+
+        return ((val != null) && (val.trim().length() > 0)) ? val : TRIBUTE_SINGLE_VAL;
```
force2b: Could this just be `return (String.IsBlank(val) ? TRIBUTE_SINGLE_VAL : val);`?

force2b:
@jbrandolinisf UGH! Github dropped 35 of my comments. I'm going to have to start code review over again.

force2b:
@jbrandolinisf nevermind. I found them. The small list of comments on the `conversation` page appears to be random. There are a bunch more under files, but Cmd-F doesn't find them if the diff hasn't been loaded. Phew.

src/classes/OPP_OpportunityContactRolesTribute_TEST.cls
```diff
@@ -0,0 +1,737 @@
...
+    }
+
+    /******************************************************************************************************
+    * @description Test Create Opporunity with SingleTribute Only mode
+    */
+    @isTest
+    static void createOppWithTributeSingle() {
+        //skip the test if Advancement is installed
+        if (ADV_PackageInfo_SVC.useAdv()) {
+                return;
```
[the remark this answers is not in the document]
jbrandolinisf: I just went with whatever VSCode was doing.  I will see if I can get it to always use 4 spaces...sigh...

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -314,16 +326,144 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
                 = OPP_AutomatedSoftCreditsService.createAffiliationOCRs(accountsToOpportunities, transactionOCRs);
             dmlWrapper.objectsToInsert.addAll(affiliationOCRs);
         }
+        
+        //For multiple tribute, we need to process the DML here since Tribute trigger will not call run method
+        if(processDML) {
+            TDTM_TriggerHandler.disableTDTM = true;
```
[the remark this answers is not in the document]
jbrandolinisf: @weizsforg - Do you remember why you did this?  I am not sure of the exact purpose in this instance.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -477,93 +865,147 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+        if (mapConIdOCR == null) // Doing just so repeated null checks are not necessary...
+            mapConIdOCR = new Map<Id, OpportunityContactRole>();
+
         OpportunityContactRole ocrNewCon = mapConIdOCR.get(idNewCon);
         OpportunityContactRole ocrOldCon = mapConIdOCR.get(idOldCon);
 
         if (idNewCon == null) {
             //if the lookup field was blanked out, and an OCR matches the previous contact with the correct role, delete it
-            if (idOldCon!=null && ocrOldCon!=null) {
+            if (idOldCon != null && ocrOldCon != null) {
```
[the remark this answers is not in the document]
jbrandolinisf: If you look closer, you will see that the second and third levels have else logic if that condition is not true.  I could restructure it, but it is not as simple as collapsing, and would rather not (especially since the plan is/should be to replace this code line with a much simpler one).

src/classes/OPP_OpportunityContactRoles_TDTM.cls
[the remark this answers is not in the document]
weizsforg: Oh, because we call the upsertContactRolesAfterInsert from TBU_OpportunityContactRoles_TDTM class. Since TBU_OpportunityContactRoles_TDTM does not call the "Run" method, we need to process the DML right away in this method. If the logic run in
TBU_OpportunityContactRoles_TDTM, we don't want it to trigger other event like Opp TDTM to Create/Update OCR again.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -365,69 +518,150 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+                        }
+                    }
                 }
                 //finally, manage the notification, if it has changed and it's not the same as either contact already dealt with
-                if (needsManageOCR(newOpp, oldOpp, FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT)) {
-                    manageOtherOCR(newOpp, oldOpp, mapOppIdMapConIdOCR.get(newOpp.id), FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT, transactionOCRs, ocrsToDeleteImmediately);
+                if (needsManageOCR(newOpp, oldOpp, honoreeContactIdSet, true, FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT, newOpp.Primary_Contact__c)) {
+                    manageOtherOCR(newOpp, oldOpp, mapOppIdMapConIdOCR.get(newOpp.Id), FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT, uniqueOCRKeys, transactionOCRs, ocrsToDeleteImmediately);
+                } else {
+                    if ((keyFieldsChanged) && (newOpp.Primary_Contact__c == newOpp.Notification_Recipient_Contact__c)) {
```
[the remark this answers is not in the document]
jbrandolinisf: Yes, but I would rather leave this one alone.  The nested ifs are really subcategories of the "not what the if was"...this makes it seem like all four conditions are equal...which is not true.

If you insist, I will do it, but I would prefer not to....

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -34,298 +33,426 @@
...
-            CloseDate = date.today(), StageName = UTIL_UnitTestData_TEST.getClosedWonStage()
+            CloseDate = Date.today(),
+            StageName = UTIL_UnitTestData_TEST.getClosedWonStage()
         );
         insert opp1;
-        
+
         //Make sure that we haven't created a contact role yet.
-        OpportunityContactRole[] result = [SELECT OpportunityId, ContactId, Role FROM OpportunityContactRole WHERE OpportunityId = :opp1.Id];
+        OpportunityContactRole[] result = [SELECT   OpportunityId,
```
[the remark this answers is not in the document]
jbrandolinisf: Predates the work Wei and I did...I am pushing back on this one...

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -34,298 +33,426 @@
...
         npe01__Contacts_and_Orgs_Settings__c testSettings = UTIL_CustomSettingsFacade.getContactsSettingsForTests(
             new npe01__Contacts_and_Orgs_Settings__c (npe01__Account_Processor__c = CAO_Constants.ONE_TO_ONE_PROCESSOR, 
                 npe01__Opportunity_Contact_Role_Default_role__c = OPPORTUNITY_CONTACT_ROLE_DONOR));
 
         Contact con = UTIL_UnitTestData_TEST.getContact();
         insert con;
-        
-        Trigger_Handler__c triggerHandler = [SELECT Active__c FROM Trigger_Handler__c WHERE Class__c = 'OPP_OpportunityContactRoles_TDTM'];
+
+        Trigger_Handler__c triggerHandler = [SELECT     Active__c
```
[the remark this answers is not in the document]
jbrandolinisf: Predates the work Wei and I did...I am pushing back on this one...

src/classes/OPP_OpportunityContactRoles_TEST.cls
```diff
@@ -34,298 +33,426 @@
...
+                                                    IsPrimary,
+                                                    Role 
+                                           FROM OpportunityContactRole 
+                                           WHERE    OpportunityId = :opp1.Id];
         //should be a contact role
         System.assertEquals(0, result.size());
-        
-        Opportunity[] oppResult = [SELECT AccountId FROM Opportunity WHERE Id = :opp1.Id];
+
+        Opportunity[] oppResult = [SELECT   AccountId
```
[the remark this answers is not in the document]
jbrandolinisf: Sounds good.  Although I did reformat it, when doing some of the larger queries, it is legacy code.  As such, I am pushing back on this one.

src/classes/TBU_OpportunityContactRoles_TDTM.cls
```diff
@@ -0,0 +1,272 @@
...
+    * @param objResult the describe for Tributes
+    * @return dmlWrapper Always null. DML can't be deferred in this case as too many of our other opp related
+    * triggers depend on Opportunity Contact Roles existing.
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newList, List<SObject> oldList,
+            TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+
+        if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
+            //Prevent insert tribute if the setting does not indicate that the system allow Multiple Tribute
+            if (UTIL_CustomSettingsFacade.getContactsSettings().Multiple_Tributes__c != Label.stgMultipleTributesFlexible &&
```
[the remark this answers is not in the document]
jbrandolinisf: We need to talk through the SelectOptions list one.

jbrandolinisf:
@force2b - I have updated the pull request processing the majority of your items.  There are a few we need to talk further about, and even fewer I pushed back on.  I marked the ones I took care of as resolved.  Please respond to any you have further comments about or disagree with my comments on.  If you agree nothing further is required now, just resolve them.  When done, tag me back and I will do another pass, or we can meet for 15 minutes to 1/2 hour.  Thanks for turning this around during what I am aware is probably a busy week for you.

src/classes/OPP_CreateTribute_BATCH.cls
```diff
@@ -0,0 +1,132 @@
...
+                
+                updOpportunities.add(opp);
+            }
+        }
+        
+        if (updOpportunities.size() > 0) {
+            //We do not want to run through Opportunity or Tribute triggers when clearing Tribute Fields
+            //on the Opportunity
+            TDTM_TriggerHandler.disableTDTM = true;
+            insert newTributes;
```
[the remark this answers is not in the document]
force2b: Sort of. If there is an error in the batch, the entire batch fails (200 records). The error displayed in the progress component is the error that ApexJobs shows, which is generic saying something along the lines of Record 0 failed. There's no way at all for a User to determine which record failed out of that group of 200; and they'll never really be able to determine which 200 records were not processed. 

There's a pretty strong possibility that orgs with have validation or other rules set on the Opportunity object that could affect older Opportunities being updated by this batch job.  This is a one-time job though (correct?), so I would be OK with leaving the logic for now. I do think this will come back to bite us eventually and we'll need to modify the logic to trap the error(s) and log which Opportunities failed. Until then though, I'm OK with leaving it as is.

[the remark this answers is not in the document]
jbrandolinisf: @force2b - Maybe we create a WI for this one?

src/classes/OPP_OpportunityContactRolesTribute_TEST.cls
```diff
@@ -60,7 +60,7 @@ public class OPP_OpportunityContactRolesTribute_TEST {
     static void createOppWithTributeSingle() {
         //skip the test if Advancement is installed
         if (ADV_PackageInfo_SVC.useAdv()) {
-                return;
+            return;
```
force2b: Thank for fixing these indents. I didn't mean to imply you needed to remove all the 8 space indents on the rest of the code. It was just the `return` statements within the specific blocks that checked for `ADV_PackageInfo_SVC.useAdv()`. This was clearly an error in indentation, where as the rest were actually as per our style guide. No need to change it back, but going forward you don't need to change how other statements are indented. However, if you're in any of that code again, feel free to put the indentations back the the way they should be.

jbrandolinisf:
I think I am going to go beat my head against a stone column now...

So the SObject ones were correctly indented to 8 characters?

Sigh...
--
John Brandolini
Phone: 781-460-3479

src/classes/OPP_OpportunityContactRolesTribute_TEST.cls
```diff
@@ -465,96 +431,89 @@ public class OPP_OpportunityContactRolesTribute_TEST {
...
+            Primary_Contact__c = cons[2].Id
         );
 
         Set<Id> honoreeIdSet = new Set<Id>();
 
         Test.startTest();
 
-        // getRole will return a Null, which will cause it to decide this field/test is not in use, same as if role value was not set
-        System.assertEquals (false, OPP_OpportunityContactRoles_TDTM.needsManageOCR(null, null, honoreeIdSet, false, 'NO_FIELD', null));
+        System.assertEquals (false, OPP_OpportunityContactRoles_TDTM.needsManageOCR(null, null, honoreeIdSet, false, 'NO_FIELD', null), 'getRole will return a Null, which will cause it to decide this field/test is not in use, same as if role value was not set');
```
force2b: I think you just moved the comment into the 3rd parm of the assert. It's not exactly what that parameter should represent, but it's not worth changing again. This parameter is the message generated if the assert fails, which makes it easier to debug. In this case, maybe something like "Expected getRole to return a null for some reason" (honestly I don't understand the text in there currently since getRole() is not the method being called, so I'm having a hard time coming up with what it could be). Maybe it does need changing. Not really sure anymore.

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -180,32 +181,33 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+
         Map<Id, Contact> mapConIdContact = new Map<Id, Contact>();
-        
-        List<String> listFields = null;
 
-        if (sObjType == Schema.Opportunity.getSObjectType()) { // Opportunity Type
-            listFields = new List<String>{FIELD_NAME_HONOREE_CONTACT, FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT};
-        } else {
-            listFields = new List<String>{ FIELD_NAME_HONOREE_CONTACT };
+        List<String> listFields = new List<String>{FIELD_NAME_HONOREE_CONTACT };
```
force2b: Minor, but there should be space before "FIELD".

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -180,32 +181,33 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+            listFields.add (FIELD_NAME_NOTIFICATION_RECIPIENT_CONTACT);
         }
 
         for (Integer i=0; i<listSObjs.size(); i++) {
             Sobject newSObj = listSObjs[i];
             Sobject oldSObj = listOldSObjs!=null && !listOldSObjs.IsEmpty() ? listOldSObjs[i] :  sObjType.newSObject();
             for (String field : listFields) {
                 if (newSObj.get(field) != oldSObj.get(field)) {
-                    if (newSObj.get(field)!=null)
+                    if (newSObj.get(field)!=null) {
```
force2b: Also minor, but there should be spaces before and after the `!=`

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -326,8 +329,9 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
                 = OPP_AutomatedSoftCreditsService.createAffiliationOCRs(accountsToOpportunities, transactionOCRs);
             dmlWrapper.objectsToInsert.addAll(affiliationOCRs);
         }
-        
-        //For multiple tribute, we need to process the DML here since Tribute trigger will not call run method
+
```
force2b: Good explanation!

src/classes/OPP_OpportunityContactRoles_TDTM.cls
```diff
@@ -576,26 +1018,81 @@ public without sharing class OPP_OpportunityContactRoles_TDTM extends TDTM_Runna
...
+    * @param obj The opportunity in question (could be from Trigger.old)
+    * @param mapConIdOCR A map of contact id to Opportunity Contact Role records for this opportunity.
+    * @param fieldName The name of the Opportunity or tribute field
+    * @param uniqueOCRKeys A Set of OpportunityId-ContactId-Role values to be checked and augmented to help prevent duplicate roles from being added
+    * @param ocrsToDeleteImmediately Holds Opportunity Contact Roles that need to be deleted immediately.
+    * @return void This method adds records to a list of records to be deleted immediately, but returns nothing
+    ********************************************************************************************************/
+    private void cleanupOCR(SObject obj, Map<Id, OpportunityContactRole> mapConIdOCR, String fieldName, Set<String> uniqueOCRKeys,
+                                List<OpportunityContactRole> ocrsToDeleteImmediately) {
+        cleanupOCR(obj, mapConIdOCR, fieldName, null, uniqueOCRKeys, ocrsToDeleteImmediately);
```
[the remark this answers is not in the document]
force2b: Ah. Thanks for the details. I'm not clear on the origin of that either, and I agree that we should leave as is.

jbrandolinisf:
I looked at it further, and it looks like we would need to pass true if we
were to change it from null, because the very nested comparison is

val == false, and null == false would be false, so if we actually passed
false, we would change the behavior to true which is undesirable.
--
John Brandolini
Phone: 781-460-3479

src/classes/OPP_CreateTribute_BATCH.cls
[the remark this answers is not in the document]
force2b: Good idea. I don't think it needs to be for the initial release. The A/C's on the WI would likely be something like:

- The NPSP Error Log can be used to identify which specific Opportunities failed to update by Id showing the error message associated with the Id.
- If there are any failed Opportunities in the child batch of 200 records, the entire batch of 200 records will not be updated.
- If there are any failed Opportunities in the batch, the insert of tributes is rolled back
- If there are no failures, the Opportunities and Tributes are updated/inserted as appropriate

src/classes/TBU_OpportunityContactRoles_TEST.cls
```diff
@@ -0,0 +1,323 @@
...
+* @description The Test for TBU_OpportunityContactRole_TDTM
+*/
+@isTest
+private class TBU_OpportunityContactRoles_TEST {
+
+    /******************************************************************************************************
+    * @description Test update Honoree/Notification Name when the look up changes
+    ********************************************************************************************************/
+    static testMethod void testPopulateHonoreeNotificationNameField(){
+        npe01__Contacts_and_Orgs_Settings__c testSettings = OPP_OpportunityContactRolesTribute_TEST.createContactAndOrgSettingsForOCRTributeTests(TBU_MultipleTributesSettingsFacade.TRIBUTE_MULTIPLE_VAL);
```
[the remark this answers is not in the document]
force2b: Let's talk through this one so I can get a better understanding of the method and what's downstream from it.

src/classes/OPP_OpportunityContactRolesTribute_TEST.cls
jbrandolinisf: I did one more pass on these...you can look when I push the last set of changes after we talk and I confirm any other updates I need to bang out.

src/classes/STG_PanelMultipleTribute_CTRL.cls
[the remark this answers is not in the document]
force2b: This is no longer an issue

src/classes/TBU_OpportunityContactRoles_TDTM.cls
```diff
@@ -0,0 +1,318 @@
...
+        }
+        if (oppsToUpdateMap != null && oppsToUpdateMap.values().size() > 0) {
+            TDTM_TriggerHandler.disableTDTM = true;
+            update oppsToUpdateMap.values();
+            TDTM_TriggerHandler.disableTDTM = false;
+        }
+    }
+
+    /******************************************************************************************************
+    * @description Get All fields from the Opportunity object
```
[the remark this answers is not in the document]
force2b: We decided to just eliminate large/rich text area fields from the list

jbrandolinisf:
@force2b - All set for you to do your part.  For the last item we discussed, I dropped any fields that are not "filterable" (meaning they cannot be included in the where clause).  That seemed more than sufficient for what we needed, as it would drop long textareas and blobs...

force2b:
Thanks @jbrandolinisf. The only long-term concern I have with excluding non-filterable fields is that it would also exclude encrypted fields. Though we only support a limited set of encrypted fields on the Account and Contact objects currently so this should be acceptable,

force2b:
@jbrandolinisf Would you mind attaching this PR to the related WI?

force2b:
@jbrandolinisf In testing this, the one rather significant issue I see is that because the Tribute object is only a lookup relationship, deleting an Opportunity record will leave orphaned Tribute records behind. Assuming this isn't by design (it's not in the docs), I think a new WI is needed to create a cascade delete trigger for the Tribute object. Luckily that can be done in a child branch of this branch to make code review easier. CC @njjc

jbrandolinisf:
Not sure what you mean by this.  Can you explain further?

--
John Brandolini
Phone: 781-460-3479

jbrandolinisf:
**lurch: attach W-029636

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPNfQAM" target="_blank">W-029636</a>

njjc:
Basically, if the Tribute object was a master-detail (as it was in the original design) it would be deleted automatically when the parent Opportunity is deleted. Since that doesn't happen with a lookup, we need to do the cleanup ourselves. Here's an example where Luke did the same thing for Payment Allocations: https://github.com/SalesforceFoundation/Cumulus/pull/3950

njjc:
Hey @jbrandolinisf, I think it's possible to modify the lookup field definition to do cascade delete now, instead of doing it in code. Should be a much smaller lift: https://help.salesforce.com/articleView?id=000199350&type=1

force2b:
@njjc Is that supported in a managed package?

njjc:
I guess not, needs a per-org permission ;/

PR #3719, opened 2018-10-19
clear payment fields after inline edit

bethbrains:
# Critical Changes

# Changes
- When a row with a matched payment is edited so the payment no longer matches, the matched record field clears.

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
**lurch: attach W-029111

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fOgRAAU" target="_blank">W-029111</a>

PR #3732, opened 2018-10-23
Add layouts for qa with new buttons

jjbennett:
// Added new buttons to qa only layouts.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


jjbennett:
Thanks @force2b - I was missing the namespace in the layout name which was why it created the second RD layout during my initial test. This PR is now also including the RD, I did not include the HH stuff since it's rarely used but I can if you think it's worth it?

davisagli:
I suspect that at some point we'll want a task that retrieves existing layouts, adds items to them and redeploys, so that it's easier to maintain these over time as fields are added to the managed packages. But this seems like a reasonable stopgap. Thanks @jjbennett

PR #3766, opened 2018-10-31
Data validation for a new row

randi274:
# Critical Changes

# Changes
- Verify required fields are filled out and data is formatted correctly.
- Require amount field.
# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
**lurch: attach W-028257

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fNXsAAM" target="_blank">W-028257</a>

lmeerkatz:
@kofijohnson Can you take a look at this? I'm fairly booked today and I don't want to block progress on this.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -173,6 +174,11 @@
                             currentFieldGroup.values.push(currentField.id);
                         });
                     }
+                    //special case so Amount object is always visible
+                    if (sObjectName === 'Opportunity') {
+                        currentFieldGroup.requiredOptions.push('Opportunity.npsp__Donation_Amount__c');
```
lmeerkatz: Is there a way to be clever about namespace here so we don't have to repeat these?

src/aura/BGE_EntryForm/BGE_EntryForm.cmp
```diff
@@ -32,23 +32,28 @@
...
+<aura:component>
 
+    <!--Public Attributes-->
     <aura:attribute name="labels" type="Map" description="Labels retrieved in Apex related to Schema information." access="public"/>
     <aura:attribute name="donorType" type="String" default="Contact1" description="Used to circumvent unhelpful labeling of Account1/Contact1" access="public"/>
     <aura:attribute name="recordId" type="Id" description="Populated by force:hasRecordId" access="public"/>
-    <aura:attribute name="dataImportFields" type="Object[]" description="" access="public"/>
+    <aura:attribute name="record" type="DataImport__c" description="Object" access="public"/>
+    <aura:attribute name="dataImportFields" type="Object[]" description="List of dynamic field objects to use in the field iteration" access="public"/>
+    <aura:attribute name="requiredFields" type="List" description="Easily accessible list of required fields" access="private"/>
```
lmeerkatz: I'd probably drop "easily accessible" :)

src/aura/BGE_EntryForm/BGE_EntryForm.cmp
```diff
@@ -62,27 +67,38 @@
             </lightning:layoutItem>
 
             <lightning:layoutItem size="12" smallDeviceSize="6" mediumDeviceSize="4" largeDeviceSize="2" class="slds-p-horizontal_small">
-                <aura:if isTrue="{!v.donorType == 'Contact1'}">
-                    <label for="contactLookup" class="slds-form-element__label">{!v.labels.contactObject}</label>
+                <!--don't use aura:if to hide, because we need to get reference to the inputfield for data validation-->
+                <div class="{!if(v.donorType == 'Account1', 'slds-hide', '')}">
```
lmeerkatz: If we're just hiding, it is possible for a user to select Contact as Donor Type, select a contact record, change Donor Type to Account, select an account, and have both populated? Or is that handled somewhere?

src/aura/BGE_EntryForm/BGE_EntryForm.cmp
randi274: Good catch! I'd actually wanted to delete this altogether, so will fix that too 👍

src/aura/BGE_EntryForm/BGE_EntryForm.cmp
randi274: Good catch here too! I knew we'd been using aura:ifs for reasons... will put something in place for this.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
randi274: Will have to give some additional thought to this, as it's not immediately clear with the pattern the best way to do this.

kofijohnson: @randi274 can we move this logic to the Apex Controller (BGE_ConfigurationWizard_CTRL). So we can get this required options from the Controller. That way, we can reuse the Namespace Utility classes to get the namespace?

randi274: @kofijohnson and @lmeerkatz I changed this up a little bit, since it's such a small change. Is this ok, or would a larger refactor still be better?

CarmeNima:
QA has Passed. Merging.

PR #3767, opened 2018-10-31
Origin/feature/robot

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


tests/NPSP.py
```diff
@@ -706,5 +706,12 @@ def verify_opportunities(self, len_value):
                 self.selenium.page_should_contain_element(loctor_contains)
                 
     
-            
+    def page_scroll_to_locator(self, path, *args, **kwargs):
+        loc=self.get_npsp_locator(path, *args, **kwargs)
+        self.selenium.execute_javascript("window.document.evaluate('{}', document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.scrollIntoView(true)".format(loc))
```
davisagli: In cumulusci 2.1.1 we upgraded robotframework-seleniumlibrary to a version that includes a "Scroll Element Into View" keyword. So instead of doing this with javascript, we should do `self.selenium.scroll_element_into_view(loc)`

tests/NPSP.robot
```diff
@@ -61,7 +61,21 @@ API Create Secondary Affiliation
...
+    # ...               Name=${plan_name}
+    # ...               npsp__Description__c=This plan is created via Automation 
+    # ...               &{fields}
+   
+API Create GAU
+    ${name} =   Generate Random String
+    ${gau_id} =  Salesforce Insert  npsp__General_Accounting_Unit__c
+    ...               Name=${name}
+    &{gau} =     Salesforce Get  npsp__General_Accounting_Unit__c  ${gau_id}
+    [return]         &{gau}
```
davisagli: We need to use a variable for the namespace prefix (`npsp__`) because the object does not have that prefix when NPSP is installed as an unmanaged package. But I need to do some work to figure out how we should set that variable correctly based on how NPSP was installed, so it's okay if that waits until later.

tests/browser/donations/create_contact_donation.robot
```diff
@@ -7,19 +7,16 @@ Suite Teardown  Delete Records and Close Browser
...
+    &{contact} =  API Create Contact    Email=skristem@robot.com
+    Go To Record Home  &{contact}[Id]
+    Page Scroll to Locator    record.related.title    Volunteer Hours
     Click Special Related List Button   Opportunities    New Contact Donation
     Choose Frame    New Opportunity
     Click Element    p3
     Select Option    Donation    
-    Click Element    //input[@title='Continue']
-    Sleep    5  
+    Click Button With Value    Continue
```
davisagli: I think just "Click Button" will work. `value` is one of the default locator types for this keyword.

tests/browser/gaus/assign_gau_to_opportunity.robot
```diff
@@ -1,27 +1,28 @@
...
     Add GAU Allocation    percentage    0    50
     Click Task Button    Add Row    
-    Select Search    1    ${gau_name2} 
-    sleep    2   
+    Select Search    1    &{gau2}[Name] 
+    #sleep    2   
     Add GAU Allocation    amount    1    20
     Click Save    GAU
-    
+    Verify Occurance    GAU Allocations    2
```
davisagli: Can you fix the spelling of this keyword? It should be "occurrence"

PR #3768, opened 2018-11-01
Fix CRLP FilterRule Evaluation Test Failure

force2b:
Remove filter rule tests that use a Time component because time components are not supported by the UI anyway. It appears that time zone conversions make doing an exact time comparison in the rule evaluation unreliable. Better to stick with pure date only comparisons in the unit test since that is all that is supported for Users.

Also refactored the test class to split up the tests into different groups to aid in debugging any future failures.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CMT_FilterRuleEvaluation_TEST.cls
+++ b/src/classes/CMT_FilterRuleEvaluation_TEST.cls
@@ -36,7 +36,7 @@
 @isTest(IsParallel=true)
 private class CMT_FilterRuleEvaluation_TEST {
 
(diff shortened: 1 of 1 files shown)


force2b:
@jjbennett Would you mind code reviewing this and merging if all looks good. It's only unit test fix to address a build issue from last night, so full QA not needed again. The idea is that this will resolve the build issue, but we really won't know for sure until after 8PM ET tonight. I think there are some other classes that need comments/code removal because for the most part we're not supporting filter rules with a time component, but for the moment I want to get us over the hurdle of pm build failures.

PR #3772, opened 2018-11-02
CRLP Timezone Date filter fix (attempt #3)

force2b:
Turns out that my fix the other night to the CRLP Filter Rule unit test to address the timezone issue (i.e., what caused all the builds run after 8PM ET to fail) did not actually work. When I retested again last night, the builds (and my local test still failed). After a bunch of research and testing I figured out what the issue is.
* The root cause is that the FilterRule is evaluating a DateTime field (CreatedDate) against a date-only string. Basically something like CreatedDate > 2017-06-01. This should be fine, except that Salesforce does not do a good job documenting what timezone values are returned in.
* The FilterRule evaluation class uses the compareDates method to do this work (vs. compareDateTimes) because one side of the equation is a date-only.
* The method gets the date from the field using valueOf: Date fieldValue = Date.valueOf(record('CreatedDate')). I would expect that to return the Date of the field in "my" (i.e., local) timezone.
* That is not the case. Instead, Salesforce returns the date in GMT (note that it should be date, not dateTime). So, if the record was created on "11/1/2018 9PM ET" you would expect the date to be 11/1/18. However, Salesforce actually returns 11/2/18 because 9PM ET is actually 1AM GMT the next day.
* To fix this, I added code to the method to first get the DateTime value of the field, then use the .date() method to get the date portion in the local timezone. ((DateTime)record.get('CreatedDate')).date().
* What was a little more interesting is that I originally had code similar to this in the method (which was the root cause of the bug I was trying to fix). The issue there was that the code to re-cast the Date as a DateTime field ran on every single compare, whether it was a Date or DateTime field. Apparently, Salesforce is not quite consistent on what it returns based on the type of field.
* To get this working in all cases, the key was only doing that recast to a DateTime if the field is actually a DateTime type. Basically, I put this code within an if (rule.fieldType == SoapType.DATETIME) block so it only casts the field to a DateTime if it's actually a DateTime field.

This PR also changes how the compareDateTimes method converts the DateTime string that the CRLP Settings UI creates into a DateTime. The UI adds "T" and "Z" to the dateTime string, which does not work when valueOf() is used to convert the string. It appears that this was never actually tested manually because the conversion would fail. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CMT_FilterRule.cls
+++ b/src/classes/CMT_FilterRule.cls
@@ -296,14 +296,20 @@ public class CMT_FilterRule {
     }
 
     /**
-     * @description Compare implementation for a DateTime data type
+     * @description Compare implementation for a DateTime data type. The UI will format a DateTime entered value as
+     * '2018-11-02T22:00:00.000Z' (i.e., GMT), however DateTime.valueOfGMT() requires that the string not include the
+     * "T" and "Z" letters.
      */
     public class CompareDateTimes implements ICompare {
         public Boolean isIncluded(SObject record, CMT_FilterRule rule) {
             DateTime fieldValue = (record.get(rule.field) != null ? (DateTime)record.get(rule.field) : null);
             String dateConstant = (String.isNotEmpty(rule.constant) && CMT_FilterRule.validDateConstants.contains(rule.constant) ?
                     rule.constant : null);
-            DateTime compareValue = (String.isNotEmpty(rule.constant) && dateConstant == null ? DateTime.valueOf(rule.constant) : null);
+            DateTime compareValue;
+
+            if (String.isNotEmpty(rule.constant) && dateConstant == null) {
+                compareValue = DateTime.valueOf(rule.constant.replace('T',' ').replace('Z', ''));
+            }
 
             if (fieldValue == null && compareValue == null) {
                 // A null to null comparison is either Equals (true) or false for any other operation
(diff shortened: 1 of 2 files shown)


src/classes/CMT_FilterRule.cls
```diff
@@ -120,8 +121,9 @@ public class CMT_FilterRule {
         this.field = UTIL_Describe.getFieldDescribe(filterRule.Object__r.QualifiedApiName, filterRule.Field__r.QualifiedApiName).getSObjectField();
         this.constant = filterRule.Constant__c;
         this.operation = CMT_FilterRuleUI_SVC.getFilterOperationFromString(filterRule.Operator__c);
+        this.fieldType = field.getDescribe().getSoapType();
 
-        if (field.getDescribe().getSOAPType() == SoapType.DATETIME && constant != null &&
+        if (fieldType == SoapType.DATETIME && constant != null &&
```
jjbennett: Should we update the comment below to reflect that the UI will not allow for datetime and will only accept dates?

src/classes/CMT_FilterRule.cls
```diff
@@ -287,14 +296,20 @@ public class CMT_FilterRule {
     }
 
     /**
-     * @description Compare implementation for a DateTime data type
+     * @description Compare implementation for a DateTime data type. The UI will format a DateTime entered value as
+     * '2018-11-02T22:00:00.000Z' (i.e., GMT), however DateTime.valueOfGMT() requires that the string not include the
+     * "T" and "Z" letters.
```
jjbennett: Is the `CompareDateTimes` class used anymore?

force2b:
@jjbennett The tests pass on this branch now, but until it's merged all other builds > 8PM ET are still failing. I think from our conversation earlier this was good, but I didn't want to merge until you've had a chance to give your final approval.

PR #3773, opened 2018-11-02
Turn on the robot

davisagli:
This adds all of the robot tests -- they are still a work in progress but we have 25 passing tests. The tests which are stable will be run by the "Feature Test: Robot" plan in MetaCI for all feature branches, and the tests which are not yet stable and will be run by the "Feature Test: Robot Unstable" plan _only_ for branches starting with feature/robot.

This also adds Level and Previous Level fields to Contact and Account in the QA config, which are needed by the robot tests for Levels.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cdcarter:
bleep bloop

PR #3781, opened 2018-11-03
CRLP Rollup field values not cleared on Opportunity Reparenting

### LINKED ISSUE #3752: Customizable Rollups: Rollup data not cleared when Opportunity is reparented
[Thread](https://powerofus.force.com/0D51E00004aI1V5)

When Customizable Rollups are enabled, if an Opportunity is moved from one Account to another, the rollup values are not cleared from the original Account.

Steps to repeat:

1. Enable CRLP (leave default Filter Groups and Rules)
2. Create 2 Accounts: Organization A, Organization B
3. Create Opportunity (any record type), Closed/Won, $80, Account = Organization A
4. Note that rollups are correct for $80/Today on Organization A
5. Edit Opportunity and change Account from Organization A to Organization B
6. Note that rollups are correct for $80/Today on Organization B

Return to Organization A and rollups are still same as when Opportunity was attached.

Run NPSP Settings > Bulk Data Processes > Rollup Donation Batch. Note that Organization A still has same rollup values, but no connected Opportunity.

Run Recalculate Rollups button on Organization A and then and only then are the values cleared as expected.

Also note that rollups are cleared correctly if the Opportunity is deleted, not if it's reparented.

force2b:
# Critical Changes

# Changes

# Issues Closed

Fixes #3752

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupAccount_TEST.cls
+++ b/src/classes/CRLP_RollupAccount_TEST.cls
@@ -36,9 +36,9 @@
 @IsTest
 private class CRLP_RollupAccount_TEST {
 
-    /** @description The 5 testing types supported by the test engine */
+    /** @description The 6 testing types supported by the test engine */
     private Enum TestType {
-        TestTrigger, TestQueueuable, TestBatch, TestSkewBatch, TestMultiCurrBatch
+        TestTrigger, TestQueueuable, TestBatch, TestSkewBatch, TestMultiCurrBatch, TestReparentOpp
     }
 
     /**
@@ -169,6 +169,10 @@ private class CRLP_RollupAccount_TEST {
     static testMethod void test_Rollups_Trigger() {
         testRollupsServices(TestType.TestTrigger);
     }
+    /** @description Test the Trigger method of rolling up when reparenting an Opp to a different Account */
+    static testMethod void test_Rollups_ReparentOppAccount() {
+        testRollupsServices(TestType.TestReparentOpp);
+    }
 
     /* @description Default currency for the current user (if MC is enabled in the Org) */
     private static String defaultCurrCode = UserInfo.getDefaultCurrency();
@@ -218,7 +222,6 @@ private class CRLP_RollupAccount_TEST {
 
         // Execute the rollup service test
         testRollupsServices(TestType.TestMultiCurrBatch);
-        
     }
 
     /**
@@ -234,7 +237,7 @@ private class CRLP_RollupAccount_TEST {
         // Start by enabling Customizable Rollups (which disables all legacy rollup operations)
         UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
                 Customizable_Rollups_Enabled__c = true,
-                Rollups_Limit_on_Attached_Opps_for_Skew__c = (tt == TestType.testSkewBatch ? 50 : 200),
+                Rollups_Limit_on_Attached_Opps_for_Skew__c = (tt == TestType.TestSkewBatch ? 50 : 200),
                 Rollups_Account_SkewMode_Batch_Size__c = 200,
                 Rollups_Contact_SkewMode_Batch_Size__c = 200,
                 Rollups_Contact_Soft_Credit_Batch_Size__c = 200,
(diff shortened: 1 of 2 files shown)


force2b:
**lurch: attach W-029419

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fP5vQAE" target="_blank">W-029419</a>

src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -36,9 +36,9 @@
 @IsTest
 private class CRLP_RollupAccount_TEST {
 
-    /** @description The 5 testing types supported by the test engine */
+    /** @description The 6 testing types supported by the test engine */
     private Enum TestType {
-        TestTrigger, TestQueueuable, TestBatch, TestSkewBatch, TestMultiCurrBatch
+        TestTrigger, TestQueueuable, TestBatch, TestSkewBatch, TestMultiCurrBatch, TestReparentOpp
```
jjbennett: Formatting: should be lower case enum and UPPER_CASE values. I realize it is not part of this bug fix, but thought it would be worth mentioning in case you want to adjust it while you are in here. Thanks!

src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -169,6 +169,10 @@ private class CRLP_RollupAccount_TEST {
     static testMethod void test_Rollups_Trigger() {
         testRollupsServices(TestType.TestTrigger);
     }
+    /** @description Test the Trigger method of rolling up when reparenting an Opp to a different Account */
+    static testMethod void test_Rollups_ReparentOppAccount() {
```
jjbennett: Formatting: Similar to other comment, mentioning that this should be @isTest annotation.

src/classes/CRLP_Rollup_TDTM.cls
```diff
@@ -62,15 +62,17 @@ public class CRLP_Rollup_TDTM extends TDTM_Runnable {
...
         List<SObject> recordsToRecalculate = new List<SObject>();
         for (Integer i=0; i < ((newlist == null ? oldlist.size() : newlist.size())); i++) {
+            SObject record = (newlist == null ? oldlist[i] : newlist[i]);
             if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
                 // Have any of the "key" fields on the record changed?
-                if (isRollupRecalcNeeded(newlist[i], oldlist[i])) {
-                    recordsToRecalculate.add(newlist[i]);
+                if (isRollupRecalcNeeded(record, oldlist[i])) {
+                    recordsToRecalculate.add(record);
+                    recordsToRecalculate.add(oldlist[i]);
```
jjbennett: It looks like the fix is to always include the original record whenever a recalc is required. What are your thoughts about only including the original record when a target object id has been changed?

src/classes/CRLP_RollupAccount_TEST.cls
force2b: I thought about this, but there are 4 test classes that all follow this same pattern. I'd rather tackle them all at once, but I can modify this one if you prefer.

src/classes/CRLP_RollupAccount_TEST.cls
force2b: True. Same issue as above though.

src/classes/CRLP_Rollup_TDTM.cls
force2b: I started going down that route originally, but I realized that it was unnecessary. The way the logic works is to loop through the `recordsToRecalculate` collection to get the Account or PrimaryContact Id's into a Set. Then it just calls the recalculate for that set of Id's. With that in mind, it really doesn't matter whether the AccountId or PrimaryContactId was changed or not. If the Id wasn't changed, the end result is the same -- the Id only appears once in the set anyway. If it was changed, now the Set contains both the Old and New Id's. Building a new method to check for the changed value didn't actually add any value to the process.

force2b:
Thank you both for reviewing and testing do quickly. This is probably ready for merge.

PR #3782, opened 2018-11-03
Org Telemetry Fixes

force2b:
Fix two issues:
* Integer Telemetry data was returning -1 for orgs that had no records to return. The intent is for -1 to represent an error querying data as opposed no data to query. The root issue here was that queries that returned 0 records would not return 0 for the telemetry, instead leaving the default value of -1. 
* Very large organization were still getting Query Timeout errors in the `UTIL_OrgTelemetry_SObject_BATCH` job after push upgrades. This PR changes the query to remove the non-selective filter against the Opportunity object. Significant query tests were done in to determine what query can run against a very very large organization without timing out. The initial query tests used two different WHERE clauses against the npe03__Recurring_Donation__c field. Both of those consistently failed with a query timeout error. The only query that did run consistently is one without any filter at all. Technically we could use a filter with against an indexed field, however in this case the goal was to use " != null" which would not use the index. As a result this job will query and loop through all Opportunity records, but just count the recurring donation related ones. Tests show that even with 16M Opportunities, this batch job will complete in under two hours (and is read-only, without any DML on any objects).

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_OrgTelemetry_BATCH.cls
+++ b/src/classes/UTIL_OrgTelemetry_BATCH.cls
@@ -87,6 +87,9 @@ public without sharing class UTIL_OrgTelemetry_BATCH implements Database.Batchab
         if (cnt == 1) {
             Database.executeBatch(new UTIL_OrgTelemetry_SObject_BATCH(
                     UTIL_OrgTelemetry_SObject_BATCH.TelemetryBatchJobMode.RECURRING_DONATIONS), 2000);
+        } else {
+            System.FeatureManagement.setPackageIntegerValue(UTIL_OrgTelemetry_SVC.TelemetryParameterName.Data_CountRdOppsAll.name(), 0);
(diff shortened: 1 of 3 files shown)


force2b:
**lurch: attach W-029397
**lurch: attach W-029509

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fP3uQAE" target="_blank">W-029397</a>

force2b:
**lurch: attach W-029509

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPDQQA2" target="_blank">W-029509</a>

force2b:
Thank you @RadSFDeveloper!

force2b:
Free to merge after the builds are done.

PR #3786, opened 2018-11-06
Allow non-household accounts as donors during data import

lparrott:
- Skipping automatic Account naming during Data Import if HEDA + Advancement is installed, to instead rely on HEDA TDTM triggers.

# Critical Changes

# Changes

- All Account models are now valid during Data Imports, previously only Household was allowed.

# Issues Closed

# New Metadata

# Deleted Metadata


lmeerkatz:
**lurch:attach W-029603

lmeerkatz:
@njjc Do you have any background that would help us uncover implications of allowing non-household accounts in BDI?

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPL0QAM" target="_blank">W-029603</a>

njjc:
Is there any background you can share on this work item @lparrott? 

Accounts (organizations) were already allowed as the donor in the current version of BDI, you just can't also import Contacts on the same row.

This change basically just removes the validation check for the Household Account model. BDI doesn't support the 1-1 model or the Individual/bucket model, and it would take more than removing the validation to make it compatible with those models.

lmeerkatz:
@njjc There's background in a quip document - NPSP Data Import issues when using the Admin Account model in HEDA + Advancement

njjc:
I think this change would require a large amount of testing in NPSP with the 1-1 and bucket models. The whole import process is expecting the HH Account data model. I'm willing to bet that it won't "just work" with those other NPSP models, even if this change makes HEDA function correctly.

lparrott:
@njjc @lmeerkatz I've updated these changes to only affect HEDA + Advancement users, I'll run it by my team as well, but wanted to make sure you were aware!

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -56,11 +56,13 @@ public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
...
 
     /*******************************************************************************************************
-    * @description is the org on the HH Account model
+    * @description Is the org on the HH Account model. Always returns true if HEDA + Advancement is installed
+    * @return Boolean
     */
     public Boolean isHHAccount {
         get {
-            return (STG_Panel.stgService.stgCon.npe01__Account_Processor__c == CAO_Constants.HH_ACCOUNT_PROCESSOR);
+            return (STG_Panel.stgService.stgCon.npe01__Account_Processor__c == CAO_Constants.HH_ACCOUNT_PROCESSOR)
```
weizsforg: Just a small suggestion here. I think isHHAccount is used on the page to control the rendering stuff. If we put the Adv check here, it is a little bit confusing. Because it means if Adv is installed, then it is not a HHAccount model anymore. But what we actually mean is that even isHHAccount = false, if Adv is  installed, then we still want to show the customizable rollup page. So, if we want to make it more clear, then we might need to use another variable for isAdvInstalled. And check isHHAcount || isAdvInstalled.

lparrott: I totally agree, I might just rename that variable "showCRLP" or something like that, since it's only being used for that purpose right now.

weizsforg: That will be great! Thanks!

lparrott: I just updated the variable name and description, keep me posted if you see anything else I should look at!

weizsforg:
@lparrott Looks great!

lparrott:
@lmeerkatz This change is approved on our end, would you mind taking a look? We're planning on doing a quick QA smoke test as well.

src/classes/BDI_ContactService.cls
```diff
@@ -317,7 +317,9 @@ public with sharing class BDI_ContactService {
             List<String> listDiKey = listDIKeyCx(di, strCx);
             Contact con = ContactFromDiKeys(listDiKey);
             if (con != null) {
-                if (con.Id != null && con.AccountId != null && con.Account.npe01__SYSTEM_AccountType__c != CAO_Constants.HH_ACCOUNT_TYPE) {
+                Boolean contactAccountIsNotHousehold = con.Id != null && con.AccountId != null
```
lmeerkatz: I would consider expanding this to include administrative accounts. If you had contacts associated with an organization in ADV you would still want to throw an error.

lparrott: I could be wrong, but I feel like in HEDA you could potentially have a Donor Contact with an Academic Program as their Account (Professor donates back to the school?). @kevinswiggum might know if this is a potential use case?

lmeerkatz: I believe BDI would handle that scenario incorrectly - see comments in NPSP Data Import issues when using the Admin Account model in HEDA + Advancement. You should include tests for expected behavior for advancement in addition to bypassing the NGO tests since you're supporting new functionality.

kevinswiggum: @lparrott  Regarding the faculty member scenario...we'll probably need to be a bit more prescriptive in how they link contacts to accounts when Advancement is installed. Otherwise you'll start having personal donations rolling up to Department accounts. I think it's fair to steer customers in a direction where faculty either have Administrative or Household accounts and associate them to their Department account with Affiliations.

lparrott: @kevinswiggum Great feedback, thanks for the quick response! We discussed this in Seattle this week and agree that only Contacts with Household or Administrative Accounts should be able to give donations.

lparrott:
@lmeerkatz I just pushed updates to only allow HEDA Administrative Accounts to use Data Import if Advancement is installed. I also updated the (previously) failing test methods to check for expected behavior instead of just skipping those cases, good call there.

lparrott:
@lmeerkatz Did you get a chance to take a look at these changes? I'm also fine with setting up a quick call to discuss, just let me know!

lparrott:
@njjc It looks like everyone is busy this week, but @lmeerkatz and I chatted last week, and she approves the overall changes here, but wanted you to take a look at the code before moving this through to QA for regression testing.

njjc:
@lparrott looks good to me!

PR #3793, opened 2018-11-08
Prevent XSS in Contact Name

### LINKED ISSUE #3794: Prevent XSS on Contact Delete Override
Remove potential for end user to inject XSS on Contact Delete Override page.

jjbennett:
# Critical Changes

# Changes

# Issues Closed
- Fixes #3794 

# New Metadata

# Deleted Metadata


LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYBdFAAX" target="_blank">W-026073</a>

src/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -89,20 +89,34 @@ public with sharing class CON_DeleteContactOverride_CTRL {
...
 
             //This contact is alone in a system account, delete the system account and allow the cascading
             //delete to remove the contact
             if (contactsInHousehold.size() == 1 && contactsInHousehold[0].get('ct') == 1) {
                 shouldDeleteContactAlone = false;
 
-                String helpDeleteAccountInstead = String.format(Label.ConfirmDeleteAccount, new List<String>{queryContact.Name, queryContact.Account.Name});
+                String helpDeleteAccountInstead = String.format(
+                    Label.ConfirmDeleteAccount,
+                    new List<String>{ queryContact.Name.escapeHtml4(), queryContact.Account.Name.escapeHtml4() }
```
dvenor: I did a quick check of the Salesforce secure coding guide [https://resources.docs.salesforce.com/214/latest/en-us/sfdc/pdf/secure_coding.pdf](https://resources.docs.salesforce.com/214/latest/en-us/sfdc/pdf/secure_coding.pdf) and noticed this:

**Do not use the built in Apex String Encoding functions: String.escapeEcmaScript(),
String.escapeHtml3(), and String.escapeHtml4(). These functions are based on
Apache's StringEscapeUtils package which was not designed for security encoding and should not be
used**

It recommended escaping the html in the view rather than the controller.  We could do this in the UTIL_PageMessages.vfc, adding this on line 71

`<apex:outputText value="{!HTMLENCODE(m.summary)}" escape="{!escMsg}"/>`

This approach might solve numerous xss issues in one fell swoop.

dvenor: As I thought more about the situation, I guessed we probably can't escape the tags in the component as we probably are including Html in some messages. All good then.

PR #3808, opened 2018-11-13
BGE Grid: Expected Totals Validation

kofijohnson:
# Critical Changes

# Changes
- Prevent user from processing batch if totals don't match expected.
# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -83,6 +83,21 @@
      * @description: invoke the Data Importer on all records associated with this Batch
      */
     processBatch: function(component) {
+        var totals = component.get('v.totals');
+        var record = component.get('v.record');
+        if (record.RequireTotalMatch__c) {
```
randi274: This should match against the label/name for RequireTotalMatch, otherwise this doesn't work in namespaced orgs.

src/labels/CustomLabels.labels
```diff
@@ -2525,6 +2525,13 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Label for Gift Entry Grid</shortDescription>
         <value>All Gifts</value>
     </labels>
+    <labels>
+        <fullName>bgeGridErrorRequiredTotalsExpected</fullName>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Message preventing user from processing batch if totals don&apos;t match expected</shortDescription>
+        <value>Either the count of gifts or the total gifts doesn&apos;t match the expected values for this batch. These value must match before you can process the batch.</value>
```
randi274: Small typo: "These values must match"

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -83,6 +83,21 @@
...
+            if (totals.countGifts == 0 && totals.totalGiftAmount == 0) {
```
randi274: I think this should be checking against Expected_Total_Batch_Amount__c and Expected_Count_of_Gifts__c, right? Because we'll want to show the error if require total match is true and the expected values are 0.

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -83,6 +83,21 @@
...
+                this.showToast(component, $A.get('$Label.c.PageMessagesError'), $A.get('$Label.c.bgeGridWarningRequiredTotalsExpected'), 'warning');
+                return;
+            }
+            
+            if ((record.Expected_Total_Batch_Amount__c != 0 && totals.totalGiftAmount != record.Expected_Total_Batch_Amount__c) ||
```
randi274: Also will need to check against the field label/name here to make sure this will work for namespacing too.

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -83,6 +83,22 @@
      * @description: invoke the Data Importer on all records associated with this Batch
      */
     processBatch: function(component) {
+        var totals = component.get('v.totals');
+        var record = component.get('v.record');
+        var labels = component.get('v.labels');
+        if (record[labels.requireTotalMatch]) {
+            if (record[labels.expectedCountField] == 0 && record[labels.expectedTotalField] == 0) {
```
randi274: For 2 null values, this drops them into the second if. Should we actually throw this error message instead?

kofijohnson: I thought about this too. However I did not handle the "null values" use cases because these fields should not technically have null values. The Configuration Wizard does not save null values and the default values on the fields is 0.

randi274: Fair enough! In that case, this is good for me, and I'll approve it.

lmeerkatz:
**lurch:attach W-029510

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPDVQA2" target="_blank">W-029510</a>

cpose:
QA Approved

PR #3814, opened 2018-11-15
Customizable Rollups: Allow rolling up between Text and Text Area fields

njjc:
# Critical Changes

# Changes
- Customizable Rollups now supports rolling up Text Area fields to Text fields, and vice versa.
- Customizable Rollups will now truncate text results that exceed the maximum length of the target field, adding an ellipsis when that happens.

# Issues Closed

# New Metadata

# Deleted Metadata


njjc:
Interesting, maybe I should update the engine to abbreviate the string to
the appropriate length before we merge this. Thoughts?

randi274:
Agreed @njjc or at least have it not throw an error if it's too long to fit. Any other thoughts on the string length error vs just shortening the string @force2b?

njjc:
I prefer abbreviating. If we use `String.abbreviate(Integer maxWidth)` it will append an ellipsis to show that not all the data could fit.

force2b:
I like that idea @njjc

njjc:
@randi274 @force2b I've updated the engine to truncate/abbreviate the rollup value when the target field is too short to fit it.

jesslopez-sf:
I ran a bunch of tests. They all rolled up data from the Opportunity to the Account (assuming all changes apply to all types of rollups). I tested both the Queueable that runs on a Closed/Won Opportunity, as well as the Recalculate Rollups button, and Bulk Data Process. I rolled up from a text area to a text area, a long text area to a long text area, a rich text area to a rich text area. Then tested the abbreviate function by rolling up each of these types to a shorter text field. Everything worked as expected.

randi274:
Awesome, this is a great improvement @njjc. Thanks for all your testing too @jesslopez-sf!

PR #3815, opened 2018-11-15
NPSP Unit Tests fail intermittently with Apex CPU timeouts

RadSFDeveloper:
Modify Apex tests resulting in intermittent "Apex CPU Timeout" errors.
- `ADDR_Addresses_TEST.insertManyAddresses` is renamed to `ADDR_Addresses_TEST.addressShouldBePropagatedToAccountsAndContactsOnMultipleTriggerCalls`
- `TDTM_Runnable_TEST.nochunking`

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -1946,33 +1831,106 @@ public with sharing class ADDR_Addresses_TEST {
...
+        Insert more than 200 addresses to require multiple trigger calls, and to ensure our static counters 
+        are handled appropriately.
     verify:
-        address propagated to all contacts and accounts
+        Addresses are propagated to all contacts and accounts
     **********************************************************************************************************/
-    static testMethod void insertManyAddresses() {
-        if (strTestOnly != '*' && strTestOnly != 'insertManyAddresses') return;
+    @isTest
+    private static void addressShouldBePropagatedToAccountsAndContactsOnMultipleTriggerCalls() {
```
RadSFDeveloper: @force2b , the [TDTM_DMLgt200_TEST.test_addresses()](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/TDTM_DMLgt200_TEST.cls#L325) could not be reused instead of this test since this test is creating Accounts, Contacts and then Addresses ensuring when the Address is assigned to an Account, it is propagated to the Account and its Contacts. The `TDTM_DMLgt200_TEST.test_addresses()` is testing Account and Address creation when Contacts are inserted.
Let me know if the above is not correct.

src/classes/TDTM_Runnable.cls
```diff
@@ -141,33 +141,35 @@ global abstract class TDTM_Runnable {
...
+        * "Cannot have more than 10 chunks in a single operation. Please rearrange the data to reduce chunking".
+        * Also, remove duplicate records to avoid the "Duplicate id in list" error.
+        * @param sObjects The records to group by type.
         * @return List<SObject> A list with the records passed as parameters ordered by type. Duplicates removed.
         */
-        private List<SObject> groupSObjects(List<SObject> sobjects) {
-                        
-            map<string,list<SObject>> mapNameToListSObject = new map<string,list<SObject>>();
-            list<SObject> sorted = new list<SObject>();
+        private List<SObject> groupSObjects(List<SObject> sObjects) {
```
RadSFDeveloper: `DmlWrapper` code has been updated to follow the latest coding standard.

RadSFDeveloper:
**lurch: attach W-022793

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000009gIFwAAM" target="_blank">W-022793</a>

src/classes/TDTM_Runnable_TEST.cls
```diff
@@ -36,65 +36,132 @@
...
+            new Contact(Id = contactIds[2], LastName = 'GroupByClark', AccountId = accountIds[1])
+        };
+
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+        dmlWrapper.objectsToInsert.addAll(objectsToInsert);
+        dmlWrapper.objectsToUpdate.addAll(objectsToUpdate);
+
+        dmlWrapper.groupByType();
+
+        System.assertEquals(3, dmlWrapper.objectsToInsert.size(),
```
force2b: I'm not following why the `objectsToInsert` collection should return 3 items. Aren't these all new Contacts, so there shouldn't be any duplicates. I'd expect 4 here.

RadSFDeveloper: That is the question I wanted to ask, ie should we remove SObjects on insert that are the same. The code has been implemented a long time. Is this expected behaviour?

src/classes/TDTM_Runnable_TEST.cls
```diff
@@ -36,65 +36,132 @@
...
+
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+        dmlWrapper.objectsToInsert.addAll(objectsToInsert);
+
+        dmlWrapper.groupByType();
+
+        System.assertEquals(objectsToInsert.size(), dmlWrapper.objectsToInsert.size(),
+            'Insert list size should be unchanged: ' + dmlWrapper.objectsToInsert);
+
+        insert dmlWrapper.objectsToInsert;
```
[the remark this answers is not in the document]
RadSFDeveloper: Yes, the "duplicate id in the list" testing in this test, and "cannot have more than 10 chunks in a single operation" in the next test is verified by actual insert, ie update. If the `dmlWrapper.groupByType()` is commented out, the insert/update will throw an exception and the unit test would fail.

The [previous unit test version](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/TDTM_Runnable_TEST.cls#L74) actually did not throw an exception when [grouping sObjects by type](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/TDTM_Runnable.cls#L157) is commented out, meaning we do not have to insert 210 Contacts to verify `TDTM_Runnable` methods.

src/classes/ADDR_Addresses_TEST.cls
RadSFDeveloper: Difference is styling in the class, and in this method, usage of `UTIL_UnitTestData_TEST.disable*Triggers()` and `UTIL_Debug.disableDebugLogging()`. No other functional changes are applied, however, since the CPU Timeout error is intermittent, we do not know if actually this work when builds are being created.
Creating a separate class using `testSetup` is good idea. Shall I go ahead with its implementation?

src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -1946,33 +1831,106 @@ public with sharing class ADDR_Addresses_TEST {
...
+    * @description
+    * utility to set up common test data for the Address tests.
+    *    - this will create & insert N household Accounts, each with M contacts
+    *    - also will create a Default Addresses for each HH Account
+    *    - will also setup the default Account Processor to be HH Accounts
+    * @param cHHT the number of Household Accounts to create
+    * @param cConT the number of Contacts to create per Household
+    * @return  void
+    **********************************************************************************************************/
+    public static void createHHTestData(Integer cHHT, Integer cConT) {
```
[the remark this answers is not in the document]
RadSFDeveloper: It should stay `public` since the method is being called from other test classes.

src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -1946,33 +1831,106 @@ public with sharing class ADDR_Addresses_TEST {
...
+        // restore async behavior as appropriate
+        TDTM_DefaultConfig.stopTestDataInit();
+    }
+    
+    /*********************************************************************************************************
+    * @description Initializes (but not inserts) N new Address objects.
+    * @param cAddr the number of Address objects to create
+    * @return  void
+    **********************************************************************************************************/
+    public static void initTestAddr(Integer cAddr) {
```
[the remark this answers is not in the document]
RadSFDeveloper: It should stays `public` since it is called from `ADDR_Addresses_TEST`.

src/classes/TDTM_Runnable_TEST.cls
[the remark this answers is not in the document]
force2b: Makes sense. Nice solution to this.

src/classes/ADDR_Addresses_TEST.cls
force2b: Great point about disabling the triggers. Might add a slight improvement. I'm fine with this as is for now, but I like knowing that we have an alternative path to try if still get errors. 

What I've been doing to test this is use the `/rebuild` option to force a rebuild on this branch. I'm hoping that that if we do that a whole bunch of times it won't fail at all. It just takes an hour between each retry. If it does eventually fail with a time out, then we can try a different class. I just did a second time (for a total of 3 builds). When it's done, just try it again (and again, etc.). to see what happens.

RadSFDeveloper: @force2b , the rebuilds have been run (5 builds in [43157](https://mrbelvedereci.herokuapp.com/builds/43157/rebuilds) (ignore the Queued) and 5 builds in [43531](https://mrbelvedereci.herokuapp.com/builds/43531/rebuilds)). Tests passed. 
Do you think more rebuilds should be run? If not, is there anything else outstanding for the PR that I needs to be fixed?

force2b:
@RadSFDeveloper  Thanks for testing this multiple times and really great work reconstructing these two tests to be more efficient.

To me, it's incredibly hard to know how well this actually helped - other than clearly improving what these two tests actually tested and how the tests were performed. 0 out of 10 failures assumes that CPU timeouts are happening around 10% of the time. That's _probably_ too high, but it's not realistic to run the build 100 times. What's also particularly annoying is the wide variation in build times across all of these tests - anywhere from 36 minutes to a full hour. This is all a symptom of the multi-tenant nature of our platform, but it does nothing to help us understand true performance and expected behavior. 

So, basically I'm very much hoping this helps with CPU timeout errors at least with these two unit tests. It's impossible to know if it will help with performance at all (seems extremely unlikely). I definitely don't think this is the end of our unit test CPU timeout errors though. I'm positive that some other tests with sporadically fail. Eventually we may or may not decide to take those other tests on if they appear with some level of regularity (which also seems unlikely). 

cc @davisagli

RadSFDeveloper:
@force2b , feel free to merge the PR since QE is not required for it due to unit tests changes only.

PR #3818, opened 2018-11-17
Origin/feature/robot

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/tests/NPSP.py
+++ b/tests/NPSP.py
@@ -44,6 +44,10 @@ class NPSP(object):
     def cumulusci(self):
         return self.builtin.get_library_instance('cumulusci.robotframework.CumulusCI')
 
+    @property
+    def salesforce(self):
+        return self.builtin.get_library_instance('cumulusci.robotframework.Salesforce')
+
     def get_namespace_prefix(self, name):
         parts = name.split('__')
         if parts[-1] == 'c':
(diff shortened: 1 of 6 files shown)


tests/NPSP.py
```diff
@@ -409,17 +409,14 @@ def select_eng_plan_checkbox(self,title):
...
-                self.selenium.get_webelement(locator).send_keys(value)     
-            elif name == "Description":
-                id = "idDesc"
-                locator = npsp_lex_locators['id'].format(id)
-                self.selenium.get_webelement(locator).send_keys(value)  
+        locator = npsp_lex_locators['id'].format(name) 
+        self.selenium.set_focus_to_element(locator)
+        field=self.selenium.get_webelement(locator)
+        field.clear()
+        field.send_keys(value)
```
davisagli: If you add a property to get the Salesforce library at the top of the NPSP class:

    @property
    def salesforce(self):
        return self.builtin.get_library_instance('cumulusci.robotframework.Salesforce')

Then you can reuse the code for populating a field here:

    self.salesforce._populate_field(locator, value)

That's probably better in case we change the best practice for filling in fields. For example I'm in the middle of updating how we clear the existing field value so that it works more reliably.

skristem: @davisagli thank you. Made these changes

PR #3819, opened 2018-11-19
Feature/npsp stored xss stg panel err

mdeleon0:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lmeerkatz:
**lurch:attach W-026035

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYBZmAAP" target="_blank">W-026035</a>

dvenor:
I know I'm late to the party here as we've already merged, but I could not get this to work properly for me.  I could not get the `decode` function to return the "decoded" string.  Looked like the RexEx `/&amp;|&quot;|&#39;|'&lt;|&gt;/` wasn't matching.  The expression `/[^&amp;]|[^&quot;]|[^&#39;]|[^&lt;]|[^&gt;]/` seemed to work better.  This is a really minor change in the user experience, but we may want to revisit this.

lmeerkatz:
@davisagli Can we revert this without causing too much heartache, or would it be better to create a new PR for an incremental improvement?

davisagli:
@lmeerkatz It should be okay to either open a PR which makes the inverse change to this one or a PR with an incremental improvement. I would avoid using the "git revert" command to undo the change because that will make it harder to merge it in the future.

dvenor:
I'll open a new PR for this change.  That is probably the simplest approach

PR #3825, opened 2018-11-20
Only find Affiliations with matching role

### LINKED ISSUE #3748: Limit Affiliations Query for Automated Soft Credits
If an org has more than 50,000 Affiliations for an Account and an Opportunity is associated for that Account, Salesforce + NPSP return the "too many query rows: 50001" error. This is as expected since this is a Salesforce limit. The recommendation would be to distribute these for data skew purposes, however, I don't think we should be preventing Opportunities from being inserted. **_This is affecting integrations._** Can we add a 50000 Limit to the query on line 181 of OPP_AutomatedSoftCredits.cls?

**Current query:**
```
List<npe5__Affiliation__c> allAffiliations
            = new List<npe5__Affiliation__c>([SELECT Id, Name, npe5__Organization__c, npe5__Contact__c, Related_Opportunity_Contact_Role__c
                                                FROM npe5__Affiliation__c
                                                WHERE npe5__Organization__c IN :accountIds
                                                ORDER BY CreatedDate DESC]);
```

**Suggested Query:**
```
List<npe5__Affiliation__c> allAffiliations
            = new List<npe5__Affiliation__c>([SELECT Id, Name, npe5__Organization__c, npe5__Contact__c, Related_Opportunity_Contact_Role__c
                                                FROM npe5__Affiliation__c
                                                WHERE npe5__Organization__c IN :accountIds
                                                ORDER BY CreatedDate DESC LIMIT 50000]);
```

jjbennett:
# Critical Changes

# Changes

# Issues Closed
- Fixes #3748 

# New Metadata

# Deleted Metadata


jjbennett:
@force2b - I did some testing and it looks like deactivated picklist values are not returned by describe. Just wanted to make sure we consider the following scenario and are ok with the new behavior:

1. Customer has 'Solicitor' as a picklist value
2. Customer assigns the value in 1 to an Affiliation
3. Customer deactivates the role in 1
4. Customer creates the Opportunity for the Affiliated account
5. The resulting Opportunity will not have a Contact Role generated

Prior to this change, the customer would get a Contact Role generated for the inactive picklist value.  This seems like a rare occurrence but it is a change in behavior.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -177,22 +177,27 @@ public class OPP_AutomatedSoftCreditsService {
             return accountToAffiliations;
         }
 
+        List<String> relatedContactRoles = new List<String>();
+        Schema.DescribeFieldResult fieldResult = npe5__Affiliation__c.Related_Opportunity_Contact_Role__c.getDescribe();
+        List<Schema.PicklistEntry> picklistValues = fieldResult.getPicklistValues();
+        for (Schema.PicklistEntry picklistValue : picklistValues) {
+            relatedContactRoles.add(picklistValue.getLabel());
```
force2b: This should use `getValue()` instead of `getLabel()`? Picklist values are what gets stored in the db, versus the label. They're normally identical, but they don't have to be.

src/classes/OPP_AutomatedSoftCreditsService.cls
```diff
@@ -177,22 +177,27 @@ public class OPP_AutomatedSoftCreditsService {
...
             = new List<npe5__Affiliation__c>([SELECT Id, Name, npe5__Organization__c, npe5__Contact__c, Related_Opportunity_Contact_Role__c
                                                 FROM npe5__Affiliation__c
                                                 WHERE npe5__Organization__c IN :accountIds
+                                                    AND Related_Opportunity_Contact_Role__c IN :relatedContactRoles
                                                 ORDER BY CreatedDate DESC]);
 
         // Process the records that have a value in the Related_Opportunity_Contact_Role__c field
         for (npe5__Affiliation__c eachAffiliation : allAffiliations) {
-            if (eachAffiliation.Related_Opportunity_Contact_Role__c != null
-                && !eachAffiliation.Related_Opportunity_Contact_Role__c.containsIgnoreCase(RELATED_OPPORTUNITY_CONTACT_ROLE_NONE)) {
```
force2b: This is interesting code. It's checking for the role not equal to "--none--". I would never expect that to be the value of the field. It's certainly not a picklist value of the field that we package. I can't really come up with any good reason this code would have been here. What do you think?

jjbennett: Yea that was why I removed it, --none-- is a label in select list only and good catch on the other comment, I will change that to Value.

force2b:
@jjbennett Just saw your other comment on inactive picklist values. I surprised that deactivated picklist values are not returned given that there's a property to isActive(), but I do see that there are various posts on this out there - just no related Known Issue (gus lists as WAD).

If we continue to retrieve all records and then filter for picklist values, we run the risk of a governor limit exception (the root issue). Filtering for `Related_Opportunity_Contact_Role__c != null` runs a similar risk. I think that only leaves this as a possible solution. Still, we may need to document this as a Known Issue for Automated Soft Credits. FYI @sam-knox

jjbennett:
@force2b - Yep it's pretty unlikely to be hit but we might want to even mention it in the changes section of the PR?

force2b:
Yes, definitely. The PR comments as well as in the official Automated Soft Credits docs.

jjbennett:
@force2b - I made the requested change, if you approve then @rponti-sforg this is ready for testing.

sam-knox:
So the bottom line is that inactive picklist values in the related contact role field will no longer result in an OCR?

jjbennett:
@sam-knox - correct, only active values will be considered for the OCRs.

PR #3836, opened 2018-11-22
Feature/bdi donation creation when no contact first name opportunity name is created with null

mdeleon0:
# Critical Changes

# Changes
- When creating Opportunities, Batch Data Import will now exclude blank Contact First Names from Opportunity names. Previously, blank Contact first names were displaying in the Opportunity name as "null."

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_Donations.cls
```diff
@@ -433,25 +437,29 @@ public class BDI_Donations {
         }
 
         // now copy over all the data to the Opp
-        for (string strDIField : mapDIFieldToOppField.keySet()) {
+        for (String dataImportField : mapDIFieldToOppField.keySet()) {
+
+            String donationRecordTypeName = UTIL_Namespace.StrTokenNSPrefix('Donation_Record_Type_Name__c').toLowercase();
```
lmeerkatz: We should be able to move these variables outside of the for loop, right?

mdeleon0: you are right.

lmeerkatz:
**lurch:attach W-029130

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fOhyAAE" target="_blank">W-029130</a>

cpose:
QA Approval.

PR #3840, opened 2018-11-27
Stop Batch from Processing if it has Errors

randi274:
# Critical Changes

# Changes
- Prevent user from processing batch if there are any row-level errors in the table from dry run.
# Issues Closed

# New Metadata

# Deleted Metadata


cpose:
** lurch: attach W-029944

cpose:
QA Approved.

Issue #3747, opened 2018-10-26
Customizable Rollups Null Pointer Exception When RecordTypeId Referenced in Filter Rule with Null Value


jesslopez-sf:
This surfaced in an org without a Membership Record Type. The Memberships: Won Filter Group contains a Filter Rule for Opportunity Record Type ID In List, where the value is null:

![null rec type id](https://user-images.githubusercontent.com/29731456/47580671-542afc80-d904-11e8-8266-28a5ee98da34.jpg)

When a Closed Won Opportunity is inserted, there is a Null Pointer Exception.
`
Stacktrace available at the time of saving the record - Class.CMT_FilterRule.<init>: line 131, column 1 Class.CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup: line 96, column 1 Class.CRLP_RollupContact_SVC.processRollups: line 83, column 1 Class.CRLP_RollupProcessor_SVC.completeRollupForParent: line 303, column 1 Class.CRLP_RollupProcessor_SVC.startRollupProcess: line 115, column 1 Class.CRLP_RollupQueueable.runRollupsForIds: line 149, column 1 Class.CRLP_RollupQueueable.execute: line 73, column 1`

The line throwing the error is in CMT_FilterRule (Line 131):
    `    } else if (filterRule.Field__r.QualifiedApiName == 'RecordTypeId' && !filterRule.Constant__c.contains(recordTypeIdPrefix)) {
`

**_Temporary Workaround if you don't have a Membership Record Type:_**
Disable all of the Membership Rollups that are associated with the Memberships: Won Filter Group.

**Steps To Repeat:**
1. Delete the Membership Opportunity Record Type.
2. Enable Customizable Rollups in NPSP Settings --> Donations --> Customizable Rollups.
3. Create a Contact (First Name = Jess; Last Name = Lopez).
4. Create an Opportunity for that Contact (Name = Test; Amount = 100; Close Date = today; Record Type = Donation, Stage = Closed Won)
5. The Opportunity saves, but the error is captured in the NPSP Errors object.

jesslopez-sf:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fP1KQAU" target="_blank">W-029366</a>

seanpat09:
Fixed by #3846

mrbelvedere:
Included in beta release 3.145 (Beta 49)

[...]

PR #3846, opened 2018-11-29
Prevent CRLP rule and rollup creation if associated memberships record types do not exist

seanpat09:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


seanpat09:
**lurch: attach W-029366

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fP1KQAU" target="_blank">W-029366</a>

seanpat09:
@force2b this is the other half of the work item that will actually prevent the creation of the rule and it's associated rollups if the record types don't exist. It changes some behavior so I wanted to put this in a PR separate from the null pointer fix

rponti-sforg:
> @force2b this is the other half of the work item that will actually prevent the creation of the rule and it's associated rollups if the record types don't exist. It changes some behavior so I wanted to put this in a PR separate from the null pointer fix

@seanpat09 So the other PR (PR#3844) Mike needs to review.  Once that is complete, I can start testing, but will want to wait to merge until this PR is ready to be tested, merged, and branch deleted.  Correct?  Want to make sure I have the sequence correct here.

seanpat09:
@force2b do you have an opinion on whether we should test #3844 (which fixes the initial bug) and this PR (which prevents the creation of the rule at all) together or should we do it separately? My initial instinct is to review/test/deploy #3844 separately from this PR so that the bug fix could be deployed separately (and thus more quickly) in case there were more questions for this PR

seanpat09:
@force2b I added another check to make sure that the excluded record type filter rule won't be created if it doesn't find the record type

force2b:
Thanks @seanpat09. I just did another test and was able to reproduce the same issue by setting the only record type to exclude for Contact rollups to Membership and then deleting the Membership record type before enabling CRLP.

seanpat09:
@force2b This is ready for review again, I added more changes to check for null Constant__c values in `CRLP_Query_SEL.cls`

src/classes/CRLP_DefaultConfigBuilder_SVC.cls
```diff
@@ -802,16 +808,19 @@ public class CRLP_DefaultConfigBuilder_SVC {
...
-            } else if (CRLP_DefaultConfigBuilder.legacySettings.npo02__Membership_Record_Types__c == null &&
-                    rollupDetails[1] == CRLP_DefaultConfigBuilder.FilterGroup_Membership) {
-                // If the legacy Membership Record Type setting is null,
-                // set all the membership rollup definitions to inactive.
-                rollup.isActive = false;
+
+            } else if (rollupDetails[1] == CRLP_DefaultConfigBuilder.FilterGroup_Membership
+                && String.isEmpty(membershipRecordTypeDeveloperNames)) {
+                // If the legacy Membership Record Type setting does not match existing record types,
+                // skip building rollup
```
force2b: Why not create the rollup definition as inactive rather than not create at all?

seanpat09: When creating the Filter Group `Memberships: Won`, I put a check in to see if there are membership record types related to what's in `npo02__Households_Settings__c.npo02__Membership_Record_Types__c`. If there are no existing record types, then I don't create the Filter Group at all, so here I have to make sure not to create the Rollups that lookup to the Filter Group or an error will get thrown. (See [ln 342](https://github.com/SalesforceFoundation/Cumulus/pull/3846/files/f4f37413ec8afe91f7dded3c430cfdce26bbf341#diff-9f2e945687d30091f67cb34ab2557e90R342) for where change is made).

Alternatively we can always create the Filter Group `Memberships: Won` and just not create the Filter Rule associated to the record types when the record types doesn't exist. Originally this is how it was set up:

Filter Group: `Memberships: Won`
Filter Rule: `Membership: Won Only`
Filter Rule: `Membership: Amount Not Null`
Filter Rule: `Membership: for Record Types` -> Only created if `CRLP_DefaultConfigBuilder.legacySettings.npo02__Membership_Record_Types__c != null`

If I do that instead, then the rollups can be created and set to Inactive. It just seemed weird to create a filter group that referred to Membership when there were no membership record types, but maybe that is the intended behavior

force2b: Thanks for the explanation @seanpat09. I see why it would make sense to not create them, but if they're missing it leaves admins in an odd situation if they ever want to enable Membership rollups of some kind. I think we should do the latter - create the Filter Group with the IsWon/NotNull rules along with the associated (inactive) rollups. That allows an Admin to easily implement/activate membership rollups at any point in the future.

force2b:
Thanks for fixing @seanpat09. Changes work great.

src/classes/CRLP_DefaultConfigBuilder_SVC.cls
rponti-sforg: @seanpat09 I want to be sure I am taking everything into account when testing this PR.  How does Membership: Won Filter Group come into play here?  I deleted the Membership record type from the Opportunity object.  Should there still be a Membership: Won Filter Group after that is done?  The Membership: Won Filter Group that I had in my scratch org did not have a Filter Rule = Opportunity - Record Type ID - In List - null .  Did not exist.  Did I need to add that first before deleting the Membership record type and enabling CRLP's?  When I add that Filter Rule to the Memberships: Won Filter Group, the Value field is required; I can't leave it blank.  Is this a result of the changes you made?  Do you know how it was not being populated previously?
w-029366_crlp_filter groups_membership

seanpat09: @rponti-sforg 
1) The filter group `Membership: Won` should always be created, regardless of whether or not the  Membership record type exists.
2) If after creating a scratch org you delete the Membership profile, but before enabling CRLP, then the Filter Rule that references Record Type ID should not be created at all. Before what was happening is that the Filter Rule would get created, but the value would be Null because the record type doesn't exist

rponti-sforg: > @rponti-sforg
> 
> 1. The filter group `Membership: Won` should always be created, regardless of whether or not the  Membership record type exists.
> 2. If after creating a scratch org you delete the Membership profile, but before enabling CRLP, then the Filter Rule that references Record Type ID should not be created at all. Before what was happening is that the Filter Rule would get created, but the value would be Null because the record type doesn't exist

@seanpat09 Thank you Sean.  These are my observations:

1. Create scratch org;
2. Delete Membership record type from Opportunity object;
3. Check Membership: Won Filter Group.  Only two Opportunity rules listed; one for Won and one for Amount;
4. Enable CRLPs;
5. Create Opportunity off Contact;
6. Check Error Log.  No Errors observed;
7.  Check Membership: Won Filter Group again.  Same two rules listed as stated in #3 above.  No additional Opportunity rules created with value = null.

Is this the main test scenario and result we are trying to verify?

seanpat09: @rponti-sforg yes, that is correct. @force2b are there any other scenarios you think we should look out for?

force2b: Thank you both for the thorough work and testing on this. There are some additional items I'd consider:

- Don't delete any record types until after going into NPSP Settings and updating the legacy rollup record type exclusions so that one or more of them exclude the Membership Record Type. The idea is to have that record type Id referenced by the settings before deleting the record types and enabling CRLP.
- This issue is not unique to the Membership Record Type. I suggest excluding and deleting a different record type, such as "In Kind" (as an additional test) to confirm that the behavior is still as expected.
- There are 3 types of record type exclusions in NPSP Settings: Account, Contact, and GAU. Any or all of these 3 could be updated in NPSP Settings to exclude the record type(s) to be deleted to test that the rule creation logic used during CRLP Enablement ignores the record type id if it doesn't exist.

rponti-sforg: > Thank you both for the thorough work and testing on this. There are some additional items I'd consider:
> 
> * Don't delete any record types until after going into NPSP Settings and updating the legacy rollup record type exclusions so that one or more of them exclude the Membership Record Type. The idea is to have that record type Id referenced by the settings before deleting the record types and enabling CRLP.

**Thank you Mike.  Re-testing in new scratch org based on your feedback above.**  

> * This issue is not unique to the Membership Record Type. I suggest excluding and deleting a different record type, such as "In Kind" (as an additional test) to confirm that the behavior is still as expected.

**Perfect.  Testing this scenario as well in new scratch org.**

> * There are 3 types of record type exclusions in NPSP Settings: Account, Contact, and GAU. Any or all of these 3 could be updated in NPSP Settings to exclude the record type(s) to be deleted to test that the rule creation logic used during CRLP Enablement ignores the record type id if it doesn't exist.

**This last scenario, can that be tested in PR#3844 ...after PR#3846 is merged into PR#3844?  Is the last scenario the same as what we were testing with PR#3844?**

seanpat09: @rponti-sforg The last scenario is the same as what we were testing in #3844, but just with other use cases. Since this is a child branch of #3844, everything in #3844 is in #3846, so you should be able to do those tests in this branch as well.

rponti-sforg: > > Thank you both for the thorough work and testing on this. There are some additional items I'd consider:
> > 
> > * Don't delete any record types until after going into NPSP Settings and updating the legacy rollup record type exclusions so that one or more of them exclude the Membership Record Type. The idea is to have that record type Id referenced by the settings before deleting the record types and enabling CRLP.
> 
> **Thank you Mike. Re-testing in new scratch org based on your feedback above.**
> 
> > * This issue is not unique to the Membership Record Type. I suggest excluding and deleting a different record type, such as "In Kind" (as an additional test) to confirm that the behavior is still as expected.
> 
> **Perfect. Testing this scenario as well in new scratch org.**
> 
> > * There are 3 types of record type exclusions in NPSP Settings: Account, Contact, and GAU. Any or all of these 3 could be updated in NPSP Settings to exclude the record type(s) to be deleted to test that the rule creation logic used during CRLP Enablement ignores the record type id if it doesn't exist.
> 
> **This last scenario, can that be tested in PR#3844 ...after PR#3846 is merged into PR#3844? Is the last scenario the same as what we were testing with PR#3844?**

@force2b @seanpat09 While my testing did not produce any **null pointer exception errors** or **Membership filter rules** containing **Record Type ID** with **Value = null**, I wanted to confirm exactly what we should be seeing within **CLRP > Filter Groups**.  Specifically, rules that contain **Record Type ID** and a **Value = null**.  

My first question is, are we just concerned with the **Memberships: Won** Filter Group?
default filter groups
  
Second, for the **Opps: Won (HC) Excl RecTypes** Filter Group, I am seeing a rule that includes **Record Type ID** and **Value = null**.  Is this something we need to be concerned about?  I also see this in the **Opps: Won (SC) Excl RecTypes, Pmts: Paid Excl Rec Types, Pmts: WrittenOff Excl Rec Types** .  

Third, it looks like a User can exclude opp record types for **Payments**.  I tested Payments the same way I tested Accounts, Contacts, and GAUs record type exclusions (minus the Recalculate Rollups function).  Do we need to be concerned with this object and the exclusion of record types on it?
payments_exclude opp record types

force2b: @rponti-sforg Thanks for testing and for the great details/questions

1. I confirm that we should **not** be seeing a Filter Rule with the following. What should happen is that this rule (just this one rule within the group) shouldn't exist. As with before this isn't specific to the Membership Record Type. A rule like the one below shouldn't be created at all, regardless of the name of the Rule or the original record type it was intended to reference.
> Record Type ID _[Not in List]_ null
2. Good point on Payments. I'm pretty sure those follow the same logic as Opportunities when creating the Filter Groups & Rules, but it's worth validating.

There's a lot of code in the method that creates these, and it's grown since originally written so I can definitely see where this is a game of whack-a-mole. The good thing is that the parent branch for this PR fixes the bug that caused an exception when attempting to evaluate this rule. We're just trying to avoid creating the offending rule in the first place. The overall intent of this PR is that rule(s) such as the one below will never be created for any Filter Group relating to any Rollup or Record Type.
> Record Type ID _[Not in List]_ null
> Record Type ID _[In List]_ null

Just a point of clarification, technically the following rule is perfectly valid:
> Record Type ID _[Equals]_ null

rponti-sforg: > @rponti-sforg Thanks for testing and for the great details/questions
> 
> 1. I confirm that we should **not** be seeing a Filter Rule with the following. What should happen is that this rule (just this one rule within the group) shouldn't exist. As with before this isn't specific to the Membership Record Type. A rule like the one below shouldn't be created at all, regardless of the name of the Rule or the original record type it was intended to reference.
> 
> > Record Type ID _[Not in List]_ null
> 
> 1. Good point on Payments. I'm pretty sure those follow the same logic as Opportunities when creating the Filter Groups & Rules, but it's worth validating.
> 
> There's a lot of code in the method that creates these, and it's grown since originally written so I can definitely see where this is a game of whack-a-mole. The good thing is that the parent branch for this PR fixes the bug that caused an exception when attempting to evaluate this rule. We're just trying to avoid creating the offending rule in the first place. The overall intent of this PR is that rule(s) such as the one below will never be created for any Filter Group relating to any Rollup or Record Type.
> 
> > Record Type ID _[Not in List]_ null
> > Record Type ID _[In List]_ null
> 
> Just a point of clarification, technically the following rule is perfectly valid:
> 
> > Record Type ID _[Equals]_ null

@force2b @seanpat09 Thanks Mike for the input.  Helps me to put the pieces together and see the bigger picture.  How do we want to proceed?

1. My testing did not expose or surface any null pointer exception errors for when specific record types are excluded from Opp record types and subsequently deleted from the Opp object itself;

2. Record Type ID filter rules where Value = null are still being created within other Filter Groups.  As you state, this should not occur regardless of type of Filter Group.  However, it is not being created in Membership: Won Filter Group (this was the original issue reported).  Do we keep PR3844 and 46 open and the subsequent WI to address the Record Type ID filter rules still being created with Value = null?

seanpat09: @rponti-sforg Let me look into the other record type filter groups where value is still being created as null, I thought I had accounted for this in all the filter group types

force2b:
@seanpat09 and @rponti-sforg If PR #3844 has passed testing, it can be merged even if this one isn't complete. After merging the parent 3844, we'd change the destination on this PR from the parent branch to master. Of course, since we're not releasing an NPSP update until early January, there isn't necessarily a rush to merge these either.

seanpat09:
@rponti-sforg @force2b since both this PR and #3844 both fix Issue #3747, it might make sense to get this merged into #3844 so they both close the issue together, especially since there's not a rush to get this into master, tho let's talk about this a little more during sprint planning in case we want to break this out more.

@rponti-sforg I'm having trouble reproducing the issue where you are getting Record Type ID filter rules where Value = null for the other filter groups, can you show me the steps to cause the issue?

rponti-sforg:
QE testing has been completed for this PR.  Tests have passed.

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fP1KQAU/view

PR #3850, opened 2018-12-03
Removing Payment Allocations settings from Allocations settings page

njjc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


njjc:
I just reverted to the commit before the payment allocations PR was merged. It looks the same after this PR as it does in managed NPSP today. 

A small concern is related to packaging: do we need to add references to the labels that are removed from this page for packaging reasons? @davisagli

annejackson:
**lurch:attach W-030157

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQ0CQAU" target="_blank">W-030157</a>

force2b:
Something is not right with this page. The lookup icon on the Default Allocation field is missing again. 
image

njjc:
Nice catch @force2b !

davisagli:
@njjc I'm not quite sure how the packaging org will behave with regard to the removed labels but I suspect that they will still be included in the package due to being referenced explicitly in CustomLabels.labels. Does that sound right @cdcarter @force2b?

annejackson:
All looks good- merging PR. @force2b @dvenor @davisagli  fyi.

PR #3852, opened 2018-12-03
Fix the RegEx on the decodeHtml function

dvenor:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/components/UTIL_Typeahead.component
+++ b/src/components/UTIL_Typeahead.component
@@ -194,9 +194,9 @@
                 }
                 */
             }).on('typeahead:selected', function (object, datum) {
-                boxData['{!searchBoxId}'].setSelectedValue( this, datum.id, decode(datum.value) );
+                boxData['{!searchBoxId}'].setSelectedValue( this, datum.id, htmlDecode(datum.value) );
             }).on('typeahead:autocompleted', function (object, datum) {
-                boxData['{!searchBoxId}'].setSelectedValue( this, datum.id, decode(datum.value) );
+                boxData['{!searchBoxId}'].setSelectedValue( this, datum.id, htmlDecode(datum.value) );
             //}).on('typeahead:cursorchanged', function (object, datum) {
                 //boxData['{!searchBoxId}'].setSelectedValue( this, datum.id, datum.value );
             //}).on('typeahead:closed', function (object, datum) {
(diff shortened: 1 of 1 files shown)


dvenor:
This is an addendum PR for WI W-026035.  Small change here but you should notice when you select the group name containing HTML characters in the settings page, it now displays the decoded version.  Very small, subtle change.

dvenor:
@ptewson-sfdo , yeah, I see that too.  I would think that is acceptable since this is such an edge case and doesn't really affect any functionality.

dvenor:
**lurch: attach W-026035

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYBZmAAP" target="_blank">W-026035</a>

src/components/UTIL_Typeahead.component
```diff
@@ -216,8 +216,9 @@
             boxData['{!searchBoxId}'].fieldList.push('{!secondaryField}');
 
         // Decodes a String already encoded by Salesforce.
-        function decode(input) {
-            if (/&amp;|&quot;|&#39;|'&lt;|&gt;/.test(input)) {
+        function htmlDecode(input) {
+            var encodedHtml = /[^&amp;]|[^&quot;]|[^&#39;]|[^&lt;]|[^&gt;]/;
```
davisagli: @dvenor this still doesn't look right to me. Square brackets are character classes and match a single character, not a string. So I think this is equivalent to matching a string that does not contain the individual characters `&`, `a`, `m`, `p`, or `;`, or that does not contain the other character classes.

I think the only thing wrong with the previous regex was the stray ' before `&lt;`

Is there a reason to use this condition rather than always using the DOMParser? If the input doesn't have any special HTML characters then using the parser should be a no-op, and it seems safer to rely on the browser's parsing rather than try to duplicate it.

davisagli: @dvenor I guess the reason for the regex condition is that it's trying to only extract the text from the html if the input is already encoded with entities. I'm missing the bigger picture of whether that's the right thing to do here though.

dvenor: @davisagli I still cannot get the original, cleaned-up RegEx to work correctly `/&amp;|&quot;|&#39;|&lt;|&gt;/`.   Not sure what I'm missing.  But I think your idea was correct, that we really don't necessary need to have the conditional.  The data being returned through the remoting request is configured to escape the html entities.  I can't really think of a reason to not pass the selected result through the  `DOMParser.parseFromString()` function, unless we thought it might degrade performance, which seems unlikely.

lmeerkatz:
**lurch:attach W-026035

PR #3856, opened 2018-12-04
Fix Opportunity Workflow Rule to not copy null Fair Market Value

### LINKED ISSUE #3761: Opportunity FMV Workflow Rule does not account for when Amount is $0 and FMV has not been populated
[Thread](https://powerofus.force.com/0D58000004Q1GM1)

Currently, the criteria on the workflow rule is:

`((Opportunity: Amount EQUALS 0) OR (Opportunity: AmountEQUALS null)) AND (Opportunity: Fair Market Value NOT EQUAL TO 0)`

The way it currently is, if an Opportunity of any record type is created with a $0 Amount, it is being blanked out by the workflow rule.

The criteria needs to be changed to:

`((Opportunity: Amount EQUALS 0) OR (Opportunity: AmountEQUALS null)) AND (Opportunity: Fair Market Value GREATER THAN 0)`

jjbennett:
# Critical Changes

# Changes

# Issues Closed
- Fixes #3761 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/workflows/Opportunity.workflow
+++ b/src/workflows/Opportunity.workflow
@@ -48,7 +48,7 @@
             <type>FieldUpdate</type>
         </actions>
         <active>false</active>
-        <booleanFilter>(1 OR 2) AND 3</booleanFilter>
+        <booleanFilter>(1 OR 2) AND (3 AND 4)</booleanFilter>
         <criteriaItems>
             <field>Opportunity.Amount</field>
             <operation>equals</operation>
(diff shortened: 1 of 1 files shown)


src/workflows/Opportunity.workflow
```diff
@@ -63,6 +63,10 @@
             <operation>notEqual</operation>
             <value>0</value>
         </criteriaItems>
+        <criteriaItems>
+            <field>Opportunity.Fair_Market_Value__c</field>
```
seanpat09: Can Fair_Market_Value__c ever be negative? If so, do we still want it to copy over to Amount?

seanpat09: I guess that would have happened before this update, tho Judi suggested using `(Opportunity: Fair Market Value GREATER THAN 0)` in the issue

jjbennett:
@rponti-sforg - This is ready when you are. Thanks!

PR #3857, opened 2018-12-05
BGE: Grid FLS check

bethbrains:
# Critical Changes

# Changes
- If a user does not have FLS permissions (view, create, edit) on all Data Import and target fields in a BGE Batch, all batch information is hidden and an error message appears.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
+++ b/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
@@ -24,12 +24,20 @@
                     component.set('v.listView', listViewUnnamespaced);
                 }
             } else {
-                this.showToast(component, $A.get('$Label.c.PageMessagesError'), response.getReturnValue(), 'error');
+                const errors = response.getError();
+                let message;
+                if (errors) {
+                    if (errors[0] && errors[0].message) {
+                        message = errors[0].message;
+                    }
+                } else {
+                    message = 'Unknown error';
+                }
                 component.find('notifLib').showToast({
                     'variant': 'error',
                     'mode': 'sticky',
                     'title': $A.get('$Label.c.PageMessagesError'),
-                    'message': response.getReturnValue()
+                    'message': message
                 });
             }
         });
--- a/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
+++ b/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
@@ -26,10 +26,8 @@
             } else {
                 const errors = response.getError();
                 let message;
-                if (errors) {
-                    if (errors[0] && errors[0].message) {
-                        message = errors[0].message;
-                    }
+                if (errors && errors[0] && errors[0].message) {
+                    message = errors[0].message;
                 } else {
                     message = 'Unknown error';
                 }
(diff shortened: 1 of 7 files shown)


bethbrains:
**lurch: attach W-030162

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQ0bQAE" target="_blank">W-030162</a>

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -208,6 +209,16 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
                     col.options = BGE_ConfigurationWizard_CTRL.getPicklistOptions(col.fieldName);
 
                     columns.add(col);
+
+                    String targetObject = BDI_DataImportService.getTargetObject(col.fieldName);
+                    if (targetObject == 'Payment') {
```
ptewson-sfdo: not blocking approval on this thought, but maybe worth moving the mapping of David-object-names to API-names up to a class variable at least? So that it's more visible and more readily managed ...

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -43,13 +43,30 @@
...
         $A.enqueueAction(action);
     },
 
+    /**
+     * @description: handles the display of errors from an apex callout
+     * @param errors: list of potential errors passed back from apex
+     */
+    handleApexErrors: function(component, errors) {
+        let message;
+        if (errors) {
```
lmeerkatz: Any reason not to consolidate this with the next line? `if (errors && errors[0] && errors[0].message)`?

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -307,6 +318,17 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
         return fields;
     }
 
+    /*******************************************************************************************************
+    * @description checks for read, create, and edit FLS for a given field
+    * @param dfr DescribeFieldResult of the field to check
+    * @return void, throws AuraHandledException to display UI error
+    */
+    private static void checkFLS(DescribeFieldResult dfr) {
+        if (!dfr.isAccessible() || !dfr.isCreateable() || !dfr.isUpdateable()) {
```
lmeerkatz: Is there ever a case where isUpdateable() is true and either isAccessible() or isCreateable() is false? I would think those checks would be redundant, but let me know if that's not the case.

bethbrains: They could have Edit but not Create, so we need to check for both of those, but I think you're right that isAccessible (View) is redundant. I'll take it out.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabController.js
```diff
@@ -24,12 +24,20 @@
                     component.set('v.listView', listViewUnnamespaced);
                 }
             } else {
-                this.showToast(component, $A.get('$Label.c.PageMessagesError'), response.getReturnValue(), 'error');
+                const errors = response.getError();
+                let message;
+                if (errors) {
```
lmeerkatz: consolidate lines

PR #3858, opened 2018-12-05
Prevent XSS on NPSP Settings

jjbennett:
# Critical Changes

# Changes
 - Security Issues addressed on NPSP Settings Pages

# Issues Closed

# New Metadata

# Deleted Metadata


src/components/UTIL_HtmlOutput.component
```diff
@@ -0,0 +1,12 @@
+<apex:component layout="none" id="util_htmloutput">
+    <!-- This component will strip out the whitelisted html tags then encode the rest of the html and finally substitue back in the original tags -->
+    <apex:attribute name="html" type="String" description="The html string to output." required="true" />
+
+    <apex:variable var="temp" value="{!SUBSTITUTE(SUBSTITUTE(SUBSTITUTE(SUBSTITUTE(SUBSTITUTE(html,
```
jjbennett: @force2b @davisagli - Super elementary approach to sanitizing the html but we don't use a lot of tags so I wanted to run it past you to see what you think.

jjbennett: Eeek I had bigger issues than that in there. I didn't update my variable references either. My links are not properly working after I fixed it. Thanks.

src/classes/UTIL_HtmlOutput_CTRL.cls
```diff
@@ -0,0 +1,55 @@
+public with sharing class UTIL_HtmlOutput_CTRL {
```
jjbennett: @force2b @davisagli - Here is another idea, still fairly simple approach with the exception of the url's that we are changing to placeholders. It would take a bit of cleanup on the translation files and custom labels to swap everything for the placeholders.
A couple things I like about this approach, it's easier on translators to read the content and it provides one location for all of our page labels. I have only fixed the URL's on the home panel at this point. Just wanted to get early feedback before I take this too far. Thanks!

jjbennett: @sam-knox - thoughts on this one? We link to docs in the custom labels which are translatable but have historically never been modified by translators. How do we deal with translating the docs themselves?

force2b: I got side tracked and didn't commit my earlier comment and its now no longer relevant, so ...

I also like this in the apex controller, but I'm not really sure I like hardcoding target URL's into apex and relying on a Dev/Doc/Translator to know how to reference a link. Plus, adding a new link would require modifying the apex rather than just updating the label. I guess the question to be asked here is whether this is a workable solution to the root XSS issue we're trying to solve. If this class logic covers every possible vulnerability that could be implemented on a custom label, and we're confident that it could be documented clearly for everyone to know how to use it, then I think it could work. One thing I might suggest rather than using {0} and {1} - since it's possible that there might be some other code that wants to use these - could it instead use something like `{URL-START}` and `{URL-END}`? I think that might make the custom label easier to understand as well.

Also, I've seen `<p>`, `<ul>`, and `<li>` used in our tags as well. I wouldn't be surprised if there are others, such as `<hr>`, possibly in our underlying packages.

src/components/UTIL_HtmlOutput.component
```diff
@@ -0,0 +1,6 @@
+<apex:component layout="none" id="util_htmloutput" controller="UTIL_HtmlOutput_CTRL">
```
force2b: What would you think about `UTIL_HtmlLabelOutput` for a filename?

jjbennett: Thanks team, I have a new idea I will put together and bug you again in a bit.

src/classes/UTIL_HtmlOutput_CTRL.cls
jjbennett: Ok, third times the charm? I still have testing to do but wanted to get it back out here to get your feedback. So far this is my favorite approach, we whitelist the allowed URLs which allows them to change the page to the docs if needed. I know there are other places we are using different html tags but I thought we could add them as we need them since all of the settings labels only had bold and breaks so far.

davisagli: This approach seems okay to me. I haven't reviewed the PR closely because the VF autoformatting makes it hard to spot the consequential changes.

jjbennett: I try to make the work commits separate from the auto formatting commits to help with that a bit, in case your curious about what actually changed. This [commit](https://github.com/SalesforceFoundation/Cumulus/pull/3858/commits/5183cff4f3155e7fb6832bca2ed5c34c676a6639) will show the new component in use.

jjbennett: @force2b - How are you feeling about this approach? If you like the approach this PR is ready for Code Review. Thanks!

force2b: I like where you're heading with this, but I do find it's a bit fragile. For example, the following two URL's should be effectively identical, but the result of passing through this controller breaks the one with the space:
* `<a href="https://powerofus.force.com/NPSP_Documentation">`
* `<a href="https://powerofus.force.com/NPSP_Documentation" >`

Do you think it's possible to strip spaces in some places, such as converting ` >` to just `>`?

src/classes/UTIL_HtmlOutput_CTRL.cls
```diff
@@ -0,0 +1,135 @@
...
+        '<br>' => '|badBreak|',
+        '<br/>' => '|break|',
+        '<b>' => '|startBold|',
+        '</b>' => '|endBold|',
+        '<ol>' => '|orderList|',
+        '</ol>' => '|endOrderList|',
+        '<ul>' => '|unorderList|',
+        '</ul>' => '|endunOrderList|',
+        '<li>' => '|lineItem|',
+        '</li>' => '|endLineItem|'
```
force2b: I think this should include heading elements (`<h1>`, etc.) and paragraph (`<p>`, `</p>`)

src/classes/UTIL_HtmlOutput_CTRL.cls
```diff
@@ -0,0 +1,135 @@
...
+    };
+    /** @description The map of allowed urls and their temporary substitution values */
+    private static final Map<String, String> SUBSTITUTION_BY_ALLOWED_URL = new Map<String, String> {
```
force2b: As we talked about, this should include whitelists for `https://github.com/SalesforceFoundation` and maybe some other common/safe destinations.

jjbennett: @force2b - is it ok if we wait until we get to that particular work item? I don't think we want to whitelist here and then find out we can't actually go that route?

src/classes/UTIL_HtmlOutput_CTRL.cls
jjbennett: I can add them, we aren't using them anywhere so I thought we should stick to the bare minimum. Since it's an added iteration for each one we add. Let me know if you would still prefer to add all tags we want to allow. I can see both sides.

force2b: If you're building out this class, I think it's more efficient to build it out once with support for everything (or at least as much as we know is coming) than to modify this class each time we get to a new work item.

src/classes/UTIL_HtmlOutput_CTRL.cls
force2b: Fair enough.

src/classes/UTIL_HtmlOutput_CTRL.cls
jjbennett: I changed this around a bit and allowed closing tags in general and whitelisted specific opening tags. This should prevent any issues with spaces on the back end of the url. I also added spacing and additional tags based on your feedback. Thanks!

src/pages/STG_PanelAddrVerification.page
```diff
@@ -5,17 +5,17 @@
...
             var selection = j$('[id$=iClass__c] :selected').text();
 
             //existing value, if in View mode
-            if(!selection) {
+            if (!selection) {
                 selection = j$('[id$=iServiceName]').text();
             }
 
-            if(selection=='SmartyStreets') {
+            if (selection == 'SmartyStreets') {
```
force2b: If you're adding spaces to these, should these also be changed to `===`?

jjbennett: I can totally disable that feature in VSCode, I actually like the way it cleans up the spacing but if you would prefer to turn it off I can do that and resubmit without formatting. Or maybe we turn on the auto formatting and just have a cleanup PR? Please let me know if it's ok for this PR or if you would like it changed.

force2b: It's just that it makes it look like huge blocks of code were changed in the file, when it may just be a few areas. I do like the idea of a separate cleanup PR, but we'd have to make sure that it wouldn't get overridden by other IDE's and update our docs to state that all VF files should be formatted this way, as well as document how to do this in IntelliJ as well as VSCode.

jjbennett:
@rponti-sforg - This is ready for testing. @force2b - Removed the auto formatting and added a styleClass attribute to the component since you last approved.

jjbennett:
**lurch: attach W-026534

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYCDiAAP" target="_blank">W-026534</a>

rponti-sforg:

For further testing information, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x8000000AYCDiAAP/view

sam-knox:
@smarovich - Does this need a release note at all since it's a bug fix?

smarovich:
Yep. We agreed a while back that we would put in a generic statement about security issues. Looks like Jenny already included the statement! I just added punctuation.

PR #3859, opened 2018-12-06
Prevent users without permissions to view Address Verification Settings

RadSFDeveloper:
# Critical Changes

# Changes
* In order to view Address Settings in NPSP Settings, and to run the Health Check on address verification, users must have the Customize Application system permission and Read/Create/Delete permissions on the Address object.

# Issues Closed

# New Metadata
ApexComponent:
- UTIL_PageMessageInfo

CustomLabel:
- stgAddrPermissionDenied

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -4980,6 +4980,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>stgAddrMustChooseClass</shortDescription>
         <value>You must choose a verification service in order to enable automatic verification.</value>
     </labels>
+    <labels>
+        <fullName>stgAddrPermissionDenied</fullName>
+        <categories>Settings, address</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Permission Denied on Address Verification Panel</shortDescription>
+        <value>You do not have permission to verify Address. Please contact your system administrator for more information.</value>
```
RadSFDeveloper: @sam-knox , can you please advise on the message displayed when user has no permission to view Address Settings on NPSP Settings? The same message is displayed when user has no permission to verify addresses when Health Check is executed.

screen shot 2018-12-06 at 10 30 06 am

screen shot 2018-12-06 at 10 31 00 am

RadSFDeveloper:
@sam-knox , can you please verify the release notes in Changes?

RadSFDeveloper:
**lurch: attach W-027089

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fMTzAAM" target="_blank">W-027089</a>

src/classes/ADDR_Validator_TEST.cls
```diff
@@ -475,9 +419,487 @@ public with sharing class ADDR_Validator_TEST {
...
+        assertAuthIdAndTokenAreNull(ctrl);
+    }
+
+    /*******************************************************************************************************
+    * @description Verify permission denied message is displayed on health check panel
+    * when read only user verifies address
+    */
+    @isTest
+    private static void healthCheckShouldDisplayPermissionDeniedMessageWhenUserHasNoAccess() { 
+        Addr_Verification_Settings__c settings = createDefaultSettings();
```
force2b: This var is never used. There is quite a bit of code in this class that assigns the result of `createDefaultSettings();` to a never used var as well.

src/classes/ADDR_Validator_TEST.cls
```diff
@@ -35,39 +35,26 @@
 */
 @isTest
 public with sharing class ADDR_Validator_TEST {
```
force2b: Given what this test class does, I think the class could be marked with `IsParallel=true`. That would aid in performance. The only caveat might be just making sure that Contacts and Accounts are created with the same Name in multiple test methods. I think you've moved most of the data creation into a helper method so that could easily just generate a random number to use in the name. Given that the class size is effectively doubling with all of the new tests, I think this would be a welcome relatively small refactor with a larger postiive result.

src/classes/ADDR_Validator_TEST.cls
```diff
@@ -475,9 +419,487 @@ public with sharing class ADDR_Validator_TEST {
...
+
+        public override Boolean hasAccess() {
+            return hasAccess;
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description Address Verifier Mock
+    */
+    private class MockAddressVerifier implements STG_PanelAddrVerification_CTRL.IAddressVerifier {
```
force2b: Fantastic use of an interface to mock the verification class for testing!

src/classes/STG_PanelADDRVerification_CTRL.cls
```diff
@@ -35,10 +35,47 @@
 */
 public with sharing class STG_PanelADDRVerification_CTRL extends STG_Panel {
 
+    /*******************************************************************************************************
+    * @description Interface for the address verifier
```
force2b: Might be worth expanding the comment to explain how the interface is used.

src/classes/STG_PanelADDRVerification_CTRL.cls
```diff
@@ -35,10 +35,47 @@
...
+    */
+    public Interface IAddressVerifier {
+        String verifyService(Addr_Verification_Settings__c settings, String zipCode);
+        Boolean hasAccess();
+        void encodeToken(Addr_Verification_Settings__c settings);
+    }
+
+    private static final String SMARTY_STREETS_VALIDATOR_CLASS = 'ADDR_SmartyStreets_Validator';
```
force2b: Should this use `ADDR_SmartyStreets_Validator.class.name()` rather than hardcoding?

src/classes/USER_UserService.cls
```diff
@@ -68,6 +68,19 @@ public with sharing class USER_UserService {
         return userIdToIsSysAdmin;
     }
 
+    /*********************************************************************************************************
+    * @description Returns user Customize Application permission
+    * @param profileId User profile Id
+    * @return Boolean True if the user has Customize Application permission
+    **********************************************************************************************************/
+    public static Boolean hasCustomizeApplicationPermission(Id profileId) {
```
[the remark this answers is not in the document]
RadSFDeveloper: Good point about the permission set. The change will be applied.

src/pages/STG_PanelAddrVerification.page
```diff
@@ -5,17 +5,17 @@
...
             var selection = j$('[id$=iClass__c] :selected').text();
 
             //existing value, if in View mode
-            if(!selection) {
+            if (!selection) {
                 selection = j$('[id$=iServiceName]').text();
             }
 
-            if(selection=='SmartyStreets') {
+            if (selection == 'SmartyStreets') {
```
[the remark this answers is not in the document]
RadSFDeveloper: Fixed.

src/classes/ADDR_Validator_TEST.cls
RadSFDeveloper: The var assignment has been removed.

src/classes/ADDR_Validator_TEST.cls
RadSFDeveloper: Great point to make gradual changes to unit tests especially since sometimes code needs to be refactored in order to create a testable code. However, this class cannot be set to have parallel testing due to creating a user in a test method:
> System.DmlException: Insert failed. First exception on row 0; first error: APEX_DATA_ACCESS_RESTRICTION, In parallel Apex tests, DML operations aren't permitted on this type of object: User: []

I have just upvoted the following [idea](https://success.salesforce.com/ideaView?id=0873A000000lGvtQAE) to enable user creation in parallel test.

src/classes/STG_PanelADDRVerification_CTRL.cls
RadSFDeveloper: Good point, it has been done.

src/classes/STG_PanelADDRVerification_CTRL.cls
RadSFDeveloper: Done in the next commit, please let me know if/how the wording should be changed.

RadSFDeveloper:
@force2b , @sam-knox : the changes based on the code review have been applied. Let me know if there is anything else outstanding.

src/classes/STG_PanelADDRVerification_CTRL.cls
RadSFDeveloper: The `class.getName()` is removed since the variable should not include the namespace, so I changed it back to be hardcoded value as before. Let me know if there are issues with the approach.

src/classes/USER_UserService.cls
```diff
@@ -71,15 +71,33 @@ public with sharing class USER_UserService {
...
+        Id profileId = UserInfo.getProfileId();
+        Id userId = UserInfo.getUserId();
+
         Profile userProfile = [
             SELECT Id, PermissionsCustomizeApplication
             FROM Profile
             WHERE Id = :profileId];
 
-        return userProfile != null ? userProfile.PermissionsCustomizeApplication : false;
+        if (userProfile.PermissionsCustomizeApplication == true) {
```
force2b: @RadSFDeveloper The above query against Profile is not necessary because the query below against PersmissionSetAssignment actually includes both Profiles and Permission Sets.

src/classes/STG_PanelADDRVerification_CTRL.cls
RadSFDeveloper: The change has been done, I left the hardcoded name in the test class, since we would like to verify the value as expected is used in the controller.

force2b:
@RadSFDeveloper I'm noticing that a User who has "Customize Application", but who does not have full permissions to the Address object will see permissions error on multiple pages in NPSP Settings, not just the Addresses page. For example:
image

RadSFDeveloper:
> @RadSFDeveloper I'm noticing that a User who has "Customize Application", but who does not have full permissions to the Address object will see permissions error on multiple pages in NPSP Settings, not just the Addresses page. 

That is expected, the user has to have both Customize Application and Address sObject permissions. Acceptance Criteria:
- Before allowing the page to render, do the following:
   - Check that the user has the "customize application" permission
   - Check that the user has CRUD access to the Address object

force2b:
@RadSFDeveloper I don't believe this is correct in this instance. The user **does** have Customize Application, but they do not have access to the Address object. As a result, the User should only see this message on pages that relate to Addresses. It's a different error than the more global missing "Customize Application" permission error.

src/classes/UTIL_Namespace.cls
```diff
@@ -57,6 +57,15 @@ public class UTIL_Namespace {
         }
         return plainNamespace;   
     }
+    
+    /*******************************************************************************************************
+    * @description Returns class name without the namespace (if any)
+    */
+    public static String removeNamespace(String className) {
```
force2b: @RadSFDeveloper Since this is limited to removing the namespace for dot notation formats only (effectively just for Apex Classes), should this be renamed to `removeClassNamespace` or `removeDotNotationNamespace`? Alternatively, could the existing (and poorly named) `StrTokenRemoveNSPrefix` method support removing the namespace prefix for both `__` and `.` notation formats?

RadSFDeveloper: I think separation of removing namespace `__` (sObjects) and `.` (classes) is better since it affects less code so testing of this PR is restricted on exactly the change required by the PR.
I can rename the method to be more descriptive (even though the argument indicates it is the class name) and reuse the `getNamespace()`.

RadSFDeveloper: Actually, using `getNamespace()` does not support testable code (existing remove class namespace unit test failed), so I have just renamed the method to the `removeClassNamespace()`.

RadSFDeveloper:
@force2b , the PR is ready for the review. The info message displayed when user cannot manage Address settings is displayed on Addresses panel only.
screen shot 2018-12-12 at 2 49 32 pm

force2b:
@RadSFDeveloper Thank you for changing. I confirm that the message only appears on the address tab. I did notice that a user without Address (or Customize Application) permissions is able to run Health Check though. I'm not 100% sure that we blocked Health Check without CustomizeApplication before, but if it was blocked without that permission it should still be blocked. Blocking that page for users without the Address object permissions isn't explicitly required for this, and nothing is exposed during the process.

RadSFDeveloper:
@force2b , executing the Health Check is not blocked if a user has no Access on Address sObject, sorry for the confusion. The validation will be done only if the Address Verification Settings is populated for *SmartyStreets*. If the custom settings is not provided for SmartyStreets, then there is nothing to be validated, thus no error.

If the user has no access on the Address sObject, and the custom settings is provided, then the error is displayed:
screen shot 2018-12-17 at 11 34 07 am

RadSFDeveloper:
@rponti-sforg , the address verification on NPSP Settings is ready to be tested.

PR #3864, opened 2018-12-10
BGE: disable save button to prevent duplicate batches

bethbrains:
# Critical Changes

# Changes
- Disable Batch save button after it's clicked

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
** lurch: attach W-029655

bethbrains:
**lurch: attach W-029655

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPPCQA2" target="_blank">W-029655</a>

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -887,6 +892,15 @@
                 this.onMetadataUpdated.notify();
             }
 
+            /**
+             * @description sets the pendingsave flag to disable Save button so duplicates can't be created
+             * @return void.
+             */
+            function togglePendingSave() {
+                this.pendingSave = !this.pendingSave;
```
lmeerkatz: Would it be safer to explicitly set pendingSave to true or false? Is there a condition like clicking twice that could cause the state to toggle back to false incorrectly?

ptewson-sfdo:
Not blocking this feature, but noted that errors from the first page (blank values for expected total for instance) don't surface to user at the save state - so save fails without a visible explanation. However, you can navigate back and fix the issue if you figure out what's going on, then return to the end and save.

bethbrains:
that scenario is fixed in another branch :)

PR #3870, opened 2018-12-11
Use CRLP on Contact Merge if enabled

seanpat09:
# Critical Changes

# Changes

# Issues Closed
W-029712

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CON_ContactMergeTDTM_TEST2.cls
+++ b/src/classes/CON_ContactMergeTDTM_TEST2.cls
@@ -36,6 +36,7 @@
 
 @isTest
 public with sharing class CON_ContactMergeTDTM_TEST2 {
+    private final static Double OPP_AMOUNT = 100;
 
     /*******************************************************************************************************
     * @description merge three Contacts from different Household Accounts with many Relationships.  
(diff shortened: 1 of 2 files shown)


force2b:
@seanpat09 Can update the PR description to reference a GH issue if there is one? If not, maybe just add some text explaining the issue to the changes section.

seanpat09:
**lurch: attach W-029712

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPTsQAM" target="_blank">W-029712</a>

src/classes/CON_ContactMergeTDTM_TEST2.cls
```diff
@@ -290,4 +289,107 @@ public with sharing class CON_ContactMergeTDTM_TEST2 {
...
+
+        List<Contact> contacts = [SELECT npo02__NumberOfClosedOpps__c FROM Contact];
+        System.assertEquals(2, contacts[0].npo02__NumberOfClosedOpps__c, 'The rollup should have been calculated');
+    }
+
+    /*******************************************************************************************************************
+    * @description Builds contacts with opportunities
+    */
+    private static List<Contact> buildContactsWithOpps() {
+        UTIL_UnitTestData_TEST.createAccountContactTestData(CAO_Constants.HH_ACCOUNT_PROCESSOR, 1, 2, 1);
```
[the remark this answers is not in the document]
seanpat09: I wasn't sure which one to use - before I put these tests in this class I was mistakenly working in `CON_ContactMerge_TEST`, which is where I found `HH_ACCOUNT_PROCESSOR` being used. I can switch to `HH_ACCOUNT_TYPE` if that makes more sense in this context.

I'll change the method name to `buildContactsWithOppsForAccount()` so it's clear that there is only one contact being used. Also good catch with the unique param, I'll add a random integer

src/classes/CON_ContactMergeTDTM_TEST2.cls
```diff
@@ -290,4 +312,113 @@ public with sharing class CON_ContactMergeTDTM_TEST2 {
             system.assertEquals(2, accWinner.npo02__NumberOfClosedOpps__c);            
         }
     }
+
+    /*******************************************************************************************************
+    * @description Verifies rollup is calculted on merge
+    */
+    @isTest
+    private static void contactMergeShouldRecalculteRollup() {
```
jjbennett: Typo on the name: Recalculate

src/classes/CON_ContactMergeTDTM_TEST2.cls
```diff
@@ -290,4 +312,113 @@ public with sharing class CON_ContactMergeTDTM_TEST2 {
...
+        Test.stopTest();
+
+        System.assert(mock.rollupContactsCalled, 'Legacy rollup should have been called');
+    }
+
+    /*******************************************************************************************************
+    * @description Verifies legacy rollup is NOT used when Customizable Rollups is activated
+    */
+    @isTest
+    private static void contactMergeShouldUseCRLPWhenEnabled() {
```
jjbennett: Since we can't verify that the CRLP is being used, should we rename this to contactMergeShouldNotUseLegacyRollupWhenCRLPEnabled?

force2b:
@seanpat09 FYI I modified the PR description to remove the Work Item and change the keyword on the issue to support auto-closing (https://help.github.com/articles/closing-issues-using-keywords/)

seanpat09:
@rponti-sforg 

This is ready for QA

We want to confirm that when Contacts are merged CRLP is used when it is enabled. A good way to test this is to merge two contacts with opportunities before enabling CRLP, then activate CRLP, edit a rollup so that it has a different behavior than the legacy roll up, merge another two Contacts and compare the results. For example:

## Scenario 1: Legacy rollup
* Create Contact Jon Klaus (which creates the Klaus Household)
* Create a Donation Opportunity for Jon Klaus that is Closed Won with an amount of $1000
* Create Contact Ellen Cho (which creates the Cho Household)
* Create a Donation Opportunity for Ellen Cho that is Closed Won with an amount of $500
* Merge the two contacts, doesn't matter which is the winner. Also, you may need to add the Merge Contacts button to the Contact list view
* After merging, the Total Gifts field on the contact should be $1500

## Scenario 2: CRLP Is Enabled
* Enable CRLP
* Edit the `Contact: Total Gifts` rollup so that the Operation is `Average` instead of sum
screen shot 2018-12-14 at 1 35 09 pm
* Create Contact Raymond Stevens (which creates the Stevens Household)
* Create a Donation Opportunity for Raymond Stevens that is Closed Won with an amount of $1000
* Create Contact Amy Rodriguez (which creates the Rodriguez Household)
* Create a Donataion Opportunity for Amy Rodriguez that is Closed Won with an amount of $500
* Merge the two contacts, doesn't matter which is the winner. Also, you may need to add the Merge Contacts button to the Contact list view
* After merging, the Total Gifts field on the contact should be $750 since it is now using CRLP (the average of 500 and 1000)

PR #3875, opened 2018-12-12
BGE: Fix issue with review donations link focus on firefox [W-030307].

ptewson-sfdo:
Replace the aura:if in the BGE_EntryForm component with a slds-hide/show approach to correct a bug with firefox where a component error was thrown when the element was not yet rendered in the DOM.  

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


ptewson-sfdo:
**lurch: attach W-030307

src/aura/BGE_EntryForm/BGE_EntryForm.cmp
```diff
@@ -68,26 +68,24 @@
...
-                                            aria-hidden="true"/>
-                        </div>
-                        <div class="slds-media__body">
-                            <p>{!$Label.c.bgeMatchingAlert}
-                                <a class="slds-m-left_x-small" onclick="{!c.openMatchModal}" id="selectOpenDonation">
-                                    {!$Label.c.bgeMatchingSelect}
-                                </a>
-                            </p>
-                        </div>
+            <lightning:layoutItem size="12" smallDeviceSize="12" class="{!'slds-p-horizontal_small'+if(or(not(empty(v.openOpportunities)), not(empty(v.unpaidPayments))),'',' slds-hide')}">
```
lmeerkatz: Is there a known issue about this somewhere. Would love to know why hide works but aura:if doesn't ...

ptewson-sfdo: css leaves the markup in the dom but controls the visibility to the page, whereas aura:if actually prevents the markup from being included in the page until the condition is true. https://salesforce.stackexchange.com/questions/240348/auraif-vs-styling-to-show-hide-content

lmeerkatz: I get that. Is this a problem everywhere we use aura:ifs, or is there something about this case that's different?

ptewson-sfdo: I believe the combination of dynamically revealing that feature and trying to set focus on an element inside of it is relatively rare, and this particular issue is constrained to that combination. There's nothing inherently wrong with using aura:if outside of this particular set of circumstances.

lmeerkatz: Cool. Thanks for clarifying! Given that, my only suggestion is maybe moving the hide onto a div outside the original component so it acts like an aura:if? That would make the intent of the styling clearer.

```
<div class="{!'if(or(not(empty(v.openOpportunities)), not(empty(v.unpaidPayments))),'',' slds-hide')}">
     <lightning:layoutItem size="12" smallDeviceSize="12" class="slds-p-horizontal_small">
```

Does that work?

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQCcQAM" target="_blank">W-030307</a>

PR #3877, opened 2018-12-13
CRLP: Fix an issue with rollup filtering when a Filter Group is null (all records)

### LINKED ISSUE #3854: Customizable Rollups: Unexpected Filtering Results
It seems that the **_Account: Total Number of Gifts_** rollup has some bearing on other rollups. If I hook that rollup up to a Filter Group where anything with an Opportunity Type of “New Customer” is filtered out (Not in List or Not Equals), the rollup works as expected (it only counts Opportunities that meet the criteria). If I modify the **_Account: Total Gifts_** rollup so that there is no Filter Group, it is not rolling up the Opportunity. It almost seems like it's taking the other rollup's Filter Group into consideration. There are no errors in the Error Log.

**Screenshot of Opps Won:HC Filter Group:**
filter group

**_Steps to Repeat:_**

1. Create a Contact (First Name = Debbie; Last Name = Donor)

2. Enable Customizable Rollups in NPSP Settings → Donations --> Customizable Rollups

3. Configure Customizable Rollups and click on Filter Groups.

4. Modify the **Opps:Won (HC)** Filter Group so you have an additional Filter Rule:
    A. Object = Opportunity
    B. Field = Opportunity Type
    C. Operator: Not in List
    D. Value = New Customer

4. Go back to Customizable Rollups and make the following changes
    A. Account: Total Gifts — Change Filter Group to “No Filter Group (Include All Records)”
    B. Account: Total Number of Gifts - Verify that Filter Group = Opps:Won (HC)

5. Create an Opportunity (Name = Issue 1, Account Name = Donor Household, Amount = 1000, Close Date = today, Opportunity Type = New Customer, Stage = Closed Won)
Go to the Donor Household Account record. The Total Gifts field should be $1000, but it's showing $0. The Total Number of Gifts field is correct at 0, since the Filter Group should have filtered this Opportunity out. Hitting the Recalculate Rollups button does not resolve the issue.

6. Create a second Opportunity without an Opportunity Type (Name = Issue 2, Account Name = Donor Household, Amount = 500, Close Date = today,  Stage = Closed Won)
Go to the Donor Household Account record. The Total Gifts field should be $1500, but it's showing $500. The Total Number of Gifts field is correct at 1, since the Filter Group should have filtered the first Opportunity out. Hitting the Recalculate Rollups button does not resolve the issue.

force2b:
# Critical Changes

# Changes

# Issues Closed

* Fixes #3854 

# New Metadata

# Deleted Metadata


force2b:
**lurch: attach W-030179

davisagli:
@force2b looks like this may have broken a test with multicurrency: https://mrbelvedereci.herokuapp.com/builds/47556/tests

PR #3897, opened 2018-12-17
Prevent multiple acknowledgements from being sent for a single Opportunity

patrick-yan-sf:
# Critical Changes

# Changes
NPSP no longer sends gift acknowledgements when the Opportunity has status of "Acknowledged". Previously, it was possible to send multiple acknowledgements for the same gift.

# Issues Closed

# New Metadata
CustomLabel :
- sendAcknowledgmentNumberAlreadySent

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -4932,6 +4932,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>sendAcknowledgmentNumberSent</shortDescription>
         <value>{0} acknowledgment(s) sent.</value>
     </labels>
+    <labels>
+        <fullName>sendAcknowledgmentNumberAlreadySent</fullName>
+        <categories>Send Acknowledgment Button</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>sendAcknowledgmentNumberAlreadySent</shortDescription>
+        <value>{0} acknowledgment(s) already sent.</value>
```
patrick-yan-sf: @sam-knox Can you please verify the message displayed when the user attempts to send multiple acknowledgements for the same Opportunity?

Can you also review the changes in the PR?

patrick-yan-sf:
**lurch: attach W-028690

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fO7CAAU" target="_blank">W-028690</a>

src/labels/CustomLabels.labels
force2b: @patrick-yan-sf The new label needs to be included in the `package.xml`.

src/classes/OPP_SendAcknowledgmentBTN_CTRL.cls
```diff
@@ -51,60 +56,75 @@ public with sharing class OPP_SendAcknowledgmentBTN_CTRL {
...
     /*******************************************************************************************************
     * @description the count of Opportunities that were acknowledged
     */
-    public integer cSent { get; set; }
+    public Integer cSent { get; set; }
+
+    /*******************************************************************************************************
+    * @description the count of Opportunities where acknowledgements were already sent
+    */
+    public Integer cAlreadySent { get; set; }
```
force2b: I realize that the other vars use the "c" prefix, though this is not common practice for our controllers. I'd guess it stands for "count". I think it would be better to clean up the public vars to remove the "c" prefix, so these become something like `sentCount`, `alreadySentCount` and `notSentCount`. I'm open to another naming pattern as well.

src/classes/OPP_SendAcknowledgmentBTN_CTRL.cls
```diff
@@ -51,60 +56,75 @@ public with sharing class OPP_SendAcknowledgmentBTN_CTRL {
...
     */
-    public list<Opportunity> listOppNotSent { 
-        get {
-            if (listOppNotSent == null)
-                listOppNotSent = new list<Opportunity>(); 
-            return listOppNotSent;
-        }
-        set; 
-    }
+    public List<Opportunity> listOppNotSent { get; set; }
```
force2b: This one follows a Hungarian notation pattern, where the data type prefixes the var name. I suggest renaming to `oppsNotSent`.

src/classes/OPP_SendAcknowledgmentBTN_TEST.cls
```diff
@@ -76,53 +79,184 @@ public with sharing class OPP_SendAcknowledgmentBTN_TEST {
...
+                )
+            );
         }
-        insert opps;  
+        insert opps;
         
         List<OpportunityContactRole> ocrs = new List<OpportunityContactRole>();
         for (Opportunity opp : opps) {
-            ocrs.add(new OpportunityContactRole(OpportunityId = opp.Id, ContactId = contact.Id, IsPrimary = true));
+            ocrs.add(
```
force2b: This isn't handled automatically by the triggers?

src/classes/OPP_SendAcknowledgmentBTN_CTRL.cls
patrick-yan-sf: @force2b I've renamed it.  FYI - @RadSFDeveloper  reminded me that the name changes might require some existing dev orgs to blank out the visual force page (because of the dependency on the variable) and then redeploy the page.

src/classes/OPP_SendAcknowledgmentBTN_TEST.cls
patrick-yan-sf: That would make sense.  I have removed it as it is not necessary.

src/classes/OPP_SendAcknowledgmentBTN_CTRL.cls
force2b: Thanks @patrick-yan-sf, though I'm not sure about the rest. Only a class var was renamed, not an actual Visualforce Page or Apex Class. No changes should be required by customers at all, and we definitely wouldn't want to require any. Am I missing something on this @RadSFDeveloper?

patrick-yan-sf: @force2b 
I think @RadSFDeveloper meant this change would only impact our internal DEV orgs.  Customers would not be impacted.

force2b: Interesting. I've never had to do that for any var name changes.

RadSFDeveloper: @force2b , I thought internal Dev orgs are impacted if VF page controller's public vars/methods are renamed. If that is not the case, please ignore the comment. No custom org is affected.

patrick-yan-sf:
@rponti-sforg, this should be ready for you to test now

PR #3898, opened 2018-12-17
Reverting change to ignore Household naming when ADV is installed

lparrott:
- To fix NPSP Unit Tests when ADV is installed, we are reverting the change to ignore Household Naming when ADV is installed

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lparrott:
@lmeerkatz As it turns out, this change makes a lot of NPSP unit tests fail when ADV is installed (I should have seen that coming...) so we're going to rely on the Household Naming custom setting isntead.

davisagli:
@annejackson please remember to delete the branch after merging!

PR #3899, opened 2018-12-17
Fix for contact primary affiliation status 

### LINKED ISSUE #2817: Affiliation auto creation is setting Status to Translated Value
[Thread.
](https://powerofus.force.com/0D58000003k4u0f)
Steps to replicate:
Set Spanish as User's language
Set Primary Affiliation on a Contact Record
View Status Field on Affiliation Object and note new value added "Actual"

### LINKED ISSUE #3860: Creating a new primary Affiliation is not changing prior Affiliation Status to Former or adding End Date value
[Case](https://org62.lightning.force.com/one/one.app#/sObject/5000M00000msIeGQAU/view)

Creating a new primary Organization Affiliation record on a Contact is not updating the existing primary Affiliation record to Former status and adding End Date value as documented.

**Steps to Repro:**
1. Navigate to a Contact Record. 
2. If not existing, create an Organization Affiliation record (Affiliation #1). Make Primary checkbox = True and Status = Current. Set Start to a date in the past and leave End Date empty.
3. Once created, notice the Primary Affiliation field is populated on the Contact with that Organization name (Affiliation #1).
4. Create a second Organization Affiliation on the Contact record, for a different Organization (Affiliation #2). Make Primary checkbox = True and Status = Current. Set Start date to today and leave End Date empty.
5. Once created, notice the Primary Affiliation field is populated on the Contact with that new Organization name as expected (Affiliation #2).
6. View Affiliation #1 record.

 **What is expected:** 
- Primary Checkbox = False
- Status = Former
- End Date = Today's date

**What is happening:**
- Primary Checkbox = False (as expected)
- Status = Current
- End Date = blank

Affiliation #2 (Current):
screen shot 2018-12-06 at 12 05 41 pm

Affiliation #1 (Former):
screen shot 2018-12-06 at 12 06 07 pm

Documentation reference to changing both Status and End Date values:
screen shot 2018-12-06 at 12 08 32 pm

seanpat09:
# Critical Changes
When setting npe5__Status__c, uses hard coded values `Former` and `Current` instead of relying on the custom labels `DefaultFormerStatus` and `DefaultCurrentStatus` respectively to prevent label translation issues
# Changes

# Issues Closed
Fixes #3860
Fixes #2817

# New Metadata

# Deleted Metadata


seanpat09:
**lurch: attach W-030243

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQ78QAE" target="_blank">W-030243</a>

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -560,6 +560,9 @@ public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
             for (npe5__Affiliation__c affl : queryAffl) {
                 if (!currentPrimaryAffl.contains(affl.id)) {
                     affl.npe5__Primary__c = false;
+                    affl.npe5__Status__c = system.label.npe5.DefaultFormerStatus;
```
jjbennett: @force2b - In older code there are places were we are using custom labels to set field values. This class is using both the DefaultCurrentStatus and DefaultFormerStatus labels to set the Status field. We are treating custom labels as custom settings here and this has caused issues in the past with translations. We have set all translations to be the same value as our packaged value and have removed them from being translated in the future but that was a stop gap. We could still have confusion in an org that changed their default but didn't also update the translations and have translated users that cause this trigger to fire.
Since we are making changes to this code is now the time to address the issue? Some options:
1. Leave it alone and add comments explaining why we are setting field values using custom labels.
2. Add constants to the top of the page and hardcode the values. This could introduce issues if customers have used something besides 'Former' and 'Current' as their defaults.
3. Add 2 fields to the custom settings for affiliations for default values with Former and Current as the default's default.
My vote would be for option 3 :)

force2b: @jjbennett  Where we had issues was that the older code (pre API v39 that introduced picklist api values) would store the picklist label into the Custom Settings. Since the label gets translated, the Custom Settings value no longer matched. Plus, apex would be looking at the api name. Since our code is now all API v43, the "value" in the field is the picklist API value, not the picklist label. So, technically if we were to hardcode the API value in the code and the customers translates the label, everything will continue to work. I just tested this out in the Acknowledgements PR the other day.

In this particular case, I believe the intent is to continue to support the "Former" and "Current" picklist entries, while allowing a customer to translate the label part of those as required. If so, I think we can handle this by using constants for the picklist api values of "Former" and "Current". If the Customer has translated "Current" to "Corriente", their translated value will show up in the UI, while Apex will still see the value of "Current" when querying or setting the field value. Ideally we'd eliminate the two custom labels used for these, but those are in an underlying package. My preference would be to just replace all references to those custom labels with a reference to a constant var. 

This might be an example of how that can be tested: https://github.com/SalesforceFoundation/Cumulus/pull/3897#pullrequestreview-186669024

jjbennett: I like option 2 as well, the only concern was with customers that might have used a different value in the past so they might have to clean up their data and as you mentioned the underlying data could be different from their labels if they decided to change the labels. We would definitely need to do a critical changes in the PR to call this out.

force2b: It's a good point. If they only translated the label then it should be fine, but it's technically possible that someone might have changed the picklist value before Spring '17 so that the api name is no longer "Current" or "Former". I'm also not really sure if that would even happen given that it's a managed field. We could mention in the Critical Changes section. Basically, they just want to make sure that the picklist api name for those two values are still set properly. 
image

jjbennett: Yep I believe the original intent behind the custom labels was to give an end user an opportunity to change that default value. If they used translation workbench to override the English translation or any language translation, the stored value would be an inactive picklist value. #2817 - As you mentioned though with the addition of the api name / label on picklist values they can now change the label alone.

seanpat09: Sounds like we should go with option 2? If that's the case, I'll add a change to use a constant instead of the label and make a note in the Critical Changes section in this PR

jjbennett:
@seanpat09 - not sure if you saw the question in chatter but I believe we will also need to check if the automation is enabled before modifying the values.

src/classes/AFFL_Affiliations_TDTM.cls
force2b: Sounds great. Thanks @seanpat09

seanpat09:
@jjbennett @force2b I added the constants for Status and also added a check for the automatic affiliation management checkbox, this is ready for review again!

src/classes/AFFL_Affiliations_TDTM.cls
```diff
@@ -35,6 +35,11 @@
 * @description  Trigger handler for Affiliations related operations on Contacts, Accounts, or Affiliations.
 */
 public class AFFL_Affiliations_TDTM extends TDTM_Runnable {
+    @TestVisible
+    private static String DEFAULT_FORMER_STATUS = system.label.npe5.DefaultFormerStatus;
```
jjbennett: Can you actually switch this to 'Former' and 'Current' vs using the labels. If a user is working in another language the translated value would be pulled in here otherwise.

seanpat09: @jjbennett Haha that would definitely help fix the translation problem instead of just shuffling the variables around, just made the update. Good catch!

force2b:
Nice catch @jjbennett

seanpat09:
@rponti-sforg this is ready for testing!

rponti-sforg:
QE has been completed.  Tests have passed.

For further testing details, please see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fQ78QAE/view

PR #3900, opened 2018-12-18
Update post-install config to match trial experience

davisagli:
_tl;dr for review: Set up a scratch org using the `qa_org` and/or `install_prod` flows from this branch, compare it to an org created from the trial template, and make sure I'm aware of any discrepancies._

---

This adds a "trial" set of post-install metadata to configure NPSP similarly to the trial experience, including page layouts, record types, quick actions, adding a few unmanaged fields, and activating workflow rules. The new configuration will be used in these installation scenarios:
- QA org installed via CumulusCI or MetaCI (`qa_org` flow)
- Managed NPSP installed via CumulusCI or MetaCI (`install_beta` or `install_prod` flows)
- Yet-to-be-created installer for use with Trailhead orgs

These other scenarios are unchanged:
- Dev org installed via CumulusCI or MetaCI (`dev_org` flow)
- NPSP installed into an existing org using the mrbelvedere installer

Since the QA org configuration is used for automated Robot Framework tests, I've also updated the tests to work with the revised layouts.

The new `unpackaged/config/trial` metadata includes everything that used to be in `unpackaged/config/managed`, which I've removed. `unpackaged/config/qa` remains but now only includes extra fields used only for testing (it's deployed to QA orgs following the trial configuration).

Areas for future improvement:
- Make sure NPSP's Remote Site Settings are activated so that the modules at the bottom of the Getting Started page work. (depends on W-030318)
- Configure other profiles in addition to Admin

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


judisohn:
@davisagli THANK YOU! I'm deploying this branch to a scratch org using `qa_org` now ready to check it out.

In the meantime, after this is merged, how do you propose we keep this updated as trial changes are required going forward? Would we (CCE) also need to submit a PR against `unpackaged/config/trial` on each change to make sure they're in sync, or is there a better (read: automated) way?

jlantz:
We're going to have to work this out as we go.  It's much easier with our existing tooling capabilities to automate the trial like config in a fresh scratch org.  It's much harder to automate that for an existing org.  In the longer term, we want to have tooling that can do that but it will take some time to get in place.

Ideally, we should be thinking about qa_org config changes as part of our development process instead of as something separate.

davisagli:
@judisohn That question is on my mind too. Yes, we'll need to keep the metadata in `unpackaged/config/trial` up to date as we make changes to the trial org. I think for now it probably makes sense to continue making changes via the UI first and then retrieve them using cci. I can help with that. The hard part is figuring out which metadata needs to be updated. This will hopefully become easier to automate once source tracking is available in non-scratch orgs.

judisohn:
@jlantz Understood. I was referring more to how we would keep this config current and up to date for our internal purposes.

judisohn:
Thanks @davisagli - I think for now we can just make sure you're in the loop on the work items as we make the changes and we can coordinate from there (FYI @jesslopez-sf). There will be some big ones in the next release.

cumulusci.yml
```diff
@@ -42,25 +42,26 @@ tasks:
...
 
-    deploy_dev_config_managed:
-        description: Deploys the post install configuration for a managed DE org
+    deploy_trial_config:
+        description: Deploys the post-install configuration for a trial/new org.
         class_path: cumulusci.tasks.salesforce.Deploy
         options:
-            path: unpackaged/config/managed
+            path: unpackaged/config/trial
+            namespace_inject: $project_config.project__package__namespace
```
force2b: Out of curiosity, does this mean it always inserts the namespace where tokens are used? If so, then I'd assume this is only usable in a scratch or other org that has the npsp namespace.

davisagli: @force2b The Deploy task replaces the token with an empty string when the `unmanaged` option is True (the default) or the `namespace_inject` namespace when `unmanaged` is False. So as written here the deploy_trial_config task will set up unmanaged NPSP, but the `config_managed` flow uses this task with `unmanaged` overridden to False to get the namespace for managed NPSP.

unpackaged/config/qa/objects/Account.object
```diff
@@ -12,30 +12,6 @@
         <trackTrending>false</trackTrending>
         <type>Currency</type>
     </fields>
-    <fields>
```
force2b: These two fields used to be used in the levels robot tests, and more specifically I use them in the levels tests I expanded on -- It would validate that the "Current Level" field was populated after running the batch job.  For those tests to work, there would need to be a Level__c lookup field on the Account (or Contact) object. I can see where we might not want to have those fields in all orgs, though the `A_User_Defined_Rollup_Field__c` field is still being created in this `.object` file. 

I'm unclear on the usage of the different folders. If `unpackaged/config/qa` is for QA testing, then I'd expect these fields to be needed. But, maybe this means only manual QA testing?

unpackaged/config/qa/objects/Contact.object
```diff
@@ -68,29 +47,4 @@
         <type>Text</type>
         <unique>true</unique>
     </fields>
-    <webLinks>
```
force2b: These are the custom buttons we just added to work around namespacing issues in the child packages. Do they now exist in another flow/task?

unpackaged/config/trial/workflows/Contact.workflow
```diff
@@ -0,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workflow xmlns="http://soap.sforce.com/2006/04/metadata">
+    <rules>
+        <fullName>npe01__Contact%2EEmailChanged_Alternate</fullName>
```
force2b: Aren't these already packaged in the NPE01 child package? I didn't think we could deploy changes to the metadata if they're included in another managed package. Basically, these aren't subscriber editable so I think they're only creatable/updatable through the NPE01 package install/update.

unpackaged/config/qa/objects/Account.object
davisagli: The trial org includes these level fields for Account (although it calls the current one Level__c instead of Current_Level__c) so I moved them to unpackaged/config/trial. I'm thinking of unpackaged/config/trial as the standard NPSP configuration for new non-dev orgs. QA orgs get that PLUS a small amount of additional QA-only configuration in unpackaged/config/qa (basically some fields for testing that don't belong in customer orgs).

unpackaged/config/qa/objects/Contact.object
davisagli: Yes, I moved them to unpackaged/config/trial. This is one thing I wasn't sure about because they aren't needed for managed NPSP (but I did use namespace tokens so at least they will work there). But if we move them back to unpackaged/config/qa we don't currently have a good way to add them to the page layouts without duplicating a bunch of layout metadata, so I figured this is the best way for now.

unpackaged/config/trial/workflows/Contact.workflow
davisagli: These are here to activate the workflows because they don't get activated when installed as a managed package. It works because we aren't changing anything other than activation status compared to the npe01-packaged metadata. It would be nice if the MD-API would let us only include the change to the activation state here, but it complains if I leave other parts out.

davisagli:
@jesslopez-sf thanks, I'll go over those sometime this week. I'm surprised about those discrepancies in the page layouts because I retrieved this metadata from the trial source org, but it's certainly possible something got left out.

davisagli:
@jesslopez-sf I've added the Deliverables tab and the page layouts that I missed -- good catch!

I put the tab in the npsp package metadata with the other tabs. Let me know if there's a reason it was intentionally left out of the package and should go in the post-install metadata instead.

jjbennett:
@judisohn - Curious if you know if we intentionally did not include a tab for Grant Deadlines / Deliverables in the package? I'm wondering if it was because they aren't widely used and/or because they are only child records to specific opps like Grants?

jjbennett:
@davisagli - should we be able to run the flow multiple times on a scratch org? I get errors when I attempt to. I can work with a fresh scratch org but not existing scratch orgs, even if I run the same flow.

davisagli:
@jjbennett I'll investigate that but if you can report what the errors are that might save me some time.

davisagli:
@judisohn no problem, I moved the Deliverables tab out of the package

davisagli:
@judisohn @jesslopez-sf @force2b @jjbennett those of you who have looked at this, are there more things you're waiting for from me or that you want to check again before approving?

src/package.xml
```diff
@@ -2537,6 +2537,7 @@
         <members>DataImport__c.To_Be_Imported</members>
         <members>Engagement_Plan_Template__c.All</members>
         <members>General_Accounting_Unit__c.All</members>
+        <members>Grant_Deadline__c.All</members>
```
force2b: @davisagli I believe this adds the "All" listview for the GrantDeadline object to the main NPSP Package, however the package doesn't include a Tab for this object. I suspect this may have been added accidentally.

jjbennett:
@davisagli - This is not related to your change but the test_data_relationships should be an upsert vs an insert. If you want I can make a separate PR later and move that to a script file instead of having it in the yaml?

src/package.xml
davisagli: @force2b thanks, I missed taking this back out when I moved the tab out of the package because I hadn't quite made the connection that list views are only accessible in the context of a tab. I moved it.

jjbennett:
When I attempted to run install_prod, I got the following error.... 

MetadataComponentFailure: Update of CustomTab Grant_Deadline__c: Error: In field: fullName - no CustomObject named Grant_Deadline__c found

Update of Profile Admin: Error on line 104, col 22: In field: tab - no CustomTab named npsp__Batch_Gift_Entry found
Error: Failed: MetadataComponentFailure

davisagli:
@jjbennett Ah, I missed adding the namespace when moving the Deliverables tab to unpackaged metadata. Just fixed that.

The other one is because I've added metadata to configure BGE but there isn't a final release with BGE yet. `install_beta` should work for testing this config with managed NPSP, and I should wait to merge this until after I cut the release next Tuesday.

judisohn:
I started a review but got distracted by other things and now my org expired. Sorry about that. I'll spin up another one and check again, using `qa_org`, given @jjbennett's error report with `install_prod`

davisagli:
I'm going to merge this as is, and then add BGE-related configuration in a new PR after @jesslopez-sf is done setting it up in the TSO.

PR #3901, opened 2018-12-18
Prevent CSRF, Enforce CRUD and Add BGE Warning on BDI Delete Page

jjbennett:
# Critical Changes

# Changes

- We fixed a security issue with the deletion of NPSP Data Import records. After you click Delete All Data Import Records or Delete Imported Data Import Records, you will now be required to confirm deletion of records by clicking a Mass Delete button. Additionally, the Delete permission on the NPSP Data Imports object is now required to delete NPSP Data Import records. For information on deleting NPSP Data Import records, see the [Administrator's Guide to Importing Donor Data](https://powerofus.force.com/articles/Resource/NPSP-Administrators-Guide-to-Importing-Donor-Data).

# Issues Closed

# New Metadata

# Deleted Metadata


jjbennett:
**lurch: attach W-026535

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AYCDnAAP" target="_blank">W-026535</a>

jjbennett:
@smarovich - Attached are the 2 new messages that will appear. One to warn the users about deleting and one is for when the user does not have delete permissions. Can you please take a look when you have a moment to help with the wording? I wanted to also mention the 10k record limit but couldn't figure out the best way to explain that in a generic way.

delete bdi perm error
delete bdi warning

jjbennett:
@smarovich - This branch is ready with the changes applied. Thanks again for the help!

smarovich:
mass_delete

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -69,19 +96,39 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
...
-    PageReference deleteDIRecords(Boolean isDeleteAll) {
-        try {
-            if (isDeleteAll) {
-                delete [SELECT Id FROM DataImport__c LIMIT 10000];
-            } else {
-                delete [SELECT Id FROM DataImport__c WHERE Status__c = :BDI_DataImport_API.bdiImported LIMIT 10000];
-            }
+    private PageReference findDIRecords(Boolean isDeleteAll) {
+        if (isDeleteAll) {
+            dataImports = [SELECT Id FROM DataImport__c LIMIT 10000];
```
force2b: Rather than hardcode 10000, what about calculating this dynamically. Though theoretically the value should be 10000, this allows the action to evolve if Salesforce ever increases that limit.
`Integer maxRecords = Limits.getLimitDMLRows() - Limits.getDMLRows();`

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -41,23 +41,50 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
...
     /** @description Action type specified to delete all Data Import records */
     public static final String ACTION_DELETE_ALL = 'deleteAll';
-    
+
+    /** @description The list of found Data Import records to be deleted */
+    @TestVisible
+    private List<DataImport__c> dataImports;
+
+    /** @description True when user can delete data import records */
+    @TestVisible
```
lmeerkatz: @TestVisible and public are redundant here. Should this be private?

Could we simplify this as `public static final Boolean canDelete = SObjectType.DataImport__c.isDeletable();` ?

jjbennett: @lmeerkatz - Thanks for the review! It is a private set, but this allows the test to set it to validate what happens when a user does not have permissions without having to muck around with users/profiles in the test.

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
jjbennett: @force2b - good idea, fixed!

jjbennett:
@rponti-sforg - This is ready for QA.
FYI - I modified the page message component that are used on several of our pages to properly escape the message when we set escMsg to false. Only BDI Delete and the ContactDeleteOverride pages were setting this value to false. I closed the ContactDeleteOverride work item since the fix here also fixed that issue. Please let me know if you have any questions. Thank you!

jjbennett:
@smarovich - Release notes look awesome, thanks!

PR #3907, opened 2018-12-18
Updating BDI test results when ADV is installed

lparrott:
- Because BDI now allows the One-to-one Account model when ADV is installed, a few test results required some changes.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lparrott:
@lmeerkatz Sorry for spamming you with these today, this should be the last change we need to pass NPSP tests with ADV installed!

src/classes/BDI_DataImport_TEST.cls
```diff
@@ -357,18 +357,29 @@ public with sharing class BDI_DataImport_TEST {
...
         System.assertEquals('TestFirstname00 TestLastname00', listCon[0].Name);
 
         listDI = getDIs();
         System.assertEquals(1, listDI.size());
-        System.assertEquals(null, listDI[0].Contact1Imported__c);
-        System.assertEquals(null, listDI[0].Contact2Imported__c);
-        System.assertEquals(null, listDI[0].HouseholdAccountImported__c);
-        System.assertEquals(label.bdiErrorNonHHAccountContact, listDI[0].Contact1ImportStatus__c);
-        System.assertEquals(null, listDI[0].Contact2ImportStatus__c);
+        // the outcome will be different if Advancement is installed because non-household Accounts CAN be donors
```
lmeerkatz: I don't think this comment matches what this test is doing. Also, non-household accounts can be donors in NPSP.

src/classes/BDI_DataImport_TEST.cls
```diff
@@ -357,18 +357,29 @@ public with sharing class BDI_DataImport_TEST {
         ID ApexJobId = Database.executeBatch(bdi, 10);
         Test.stopTest();
     
+        Boolean usingAdv = ADV_PackageInfo_SVC.useAdv();
         // verify expected results
         List<Contact> listCon = getContacts();
-        System.assertEquals(1, listCon.size());
+        Integer numCons = usingAdv ? 2 : 1;
```
lmeerkatz: This doesn't make sense to me as a desirable outcome. I believe that this test is verifying that if you import a contact that matches an existing contact, that person doesn't get a second contact record. Are you interpreting this differently?

lparrott:
@lmeerkatz Alright, I think my latest changes should handle the situations we talked about:
- If either of the Contacts on a DI record is One-to-one, there can't be a second Contact on that record
- If the org in the One-to-one Account model, there can't be 2 new Contacts on a DI record

In both of those cases, I'm adding a new failure message and failing the import.

src/classes/BDI_ContactService.cls
```diff
@@ -309,13 +309,17 @@ public with sharing class BDI_ContactService {
...
         Integer iCx = strCx == 'Contact1' ? 1 : 2;
+        Boolean modelIsOneToOne = CAO_Constants.isOneToOne();
 
         for (DataImport__c di : bdi.listDI) {
             // skip di's that already have an error
             if (di.Status__c == BDI_DataImport_API.bdiFailed)
                 continue;
             List<String> listDiKey = listDIKeyCx(di, strCx);
             Contact con = ContactFromDiKeys(listDiKey);
+            Boolean bothContactsSpecified = (strCx == 'Contact1' && isContactSpecified(di, 'Contact2')) ||
```
lmeerkatz: Is there a scenario where you'd want to let Contact2 get through if Contact1 is blank?

src/classes/BDI_DataImport_TEST2.cls
```diff
@@ -1494,20 +1494,36 @@ private with sharing class BDI_DataImport_TEST2 {
         npe01__Contacts_and_Orgs_Settings__c contactSettingsForTests = UTIL_CustomSettingsFacade.getContactsSettingsForTests(
```
lmeerkatz: I'd consider moving the Adv logic to a separate test, as the method name doesn't match what you're testing in that context

src/classes/BDI_ContactService.cls
lmeerkatz: If so, it would be good to add a test to confirm that it will work when Contact2 alone is populated

src/labels/CustomLabels.labels
```diff
@@ -2291,6 +2291,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>bdiErrorNonHHAccountContactWithAdv</shortDescription>
         <value>This Contact matched an existing Contact that's associated with a non-Household, non-Administrative Account. The Gift Entry Tool can only update Contacts that are part of a Household or Administrative Account.</value>
     </labels>
+    <labels>
+        <fullName>bdiErrorOneToOneMultiContact</fullName>
+        <categories>NPSP Data Importer</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>bdiErrorOneToOneMultiContact</shortDescription>
+        <value>A Gift from a One-to-one or Administrative Contact cannot be associated with a second Contact.</value>
```
lmeerkatz: The scenario that you're testing doesn't involve gifts at all. I'd make this 'A Data Import for a One-to-one or Administrative Contact cannot be associated with a second Contact'

lparrott:
@lmeerkatz Thanks for taking another look, I just pushed some changes based on your most recent comments.

PR #3908, opened 2018-12-18
Add user-friendly labels to components available from App Builder

lmeerkatz:
It looks like this:
screen shot 2018-12-18 at 3 37 38 pm

I looked into using custom labels to support translation, but it looks like they're not available in .design files. "Label expressions in markup are supported in .cmp and .app resources only." - https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/components_config_for_app_builder_design_files.htm

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lmeerkatz:
**lurch:attach W-030366

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQHNQA2" target="_blank">W-030366</a>

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.design
```diff
@@ -1,4 +1,4 @@
-<design:component>
+<design:component label="Batch Gift Entry Configuration">
```
bethbrains: The Wizard doesn't need one, since it is only invoked from the tab or a batch record home. It had flexipage:availableForRecordHome because it used to be available on Template record home. So that can come out now.
I mentioned the BGE_BatchGiftEntryTab needing a design file, but I don't think it does, since we are shipping the tab and it doesn't appear in App Builder.

PR #3911, opened 2018-12-19
Fix CRLP unit test failure when multi-currency is enabled

force2b:
Fix a failure that did not show up until #3877 was merged into Master and the automated builds ran against a multi-currency org.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@jjbennett Would you mind doing a quick review on this? It addresses a unit test failure from a previous merge that did not show up until the tests were run in an org with multi-currency enabled.

PR #3934, opened 2018-12-20
Change icon for BGE tab to something unique

lmeerkatz:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davisagli:
@lmeerkatz your call on whether to actually proceed with this change. It does not seem to actually be necessary for packaging purposes despite what the docs said.

lmeerkatz:
@davisagli Okay. Let's leave it if it's not a problem.

PR #3906, opened 2018-12-18
BGE: Remove Process_Using_Scheduled_Job__c from BGE wizard

lmeerkatz:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

PR #3943, opened 2018-12-27
Feature/bge lc  revised batch options advanced

ptewson-sfdo:
Includes the 'Show/Hide Advanced Options' link to selectively reveal the advanced options on step #5 of the Batch Config Wizard. Also dynamically disables fields for date range + others when they are rendered irrelevant by selections of match fields or match options.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.css
```diff
@@ -1,6 +1,11 @@
 .THIS.slds-tabs_card {
     background: white;
 }
+
+/*remove focus button formatting*/
+.THIS .slds-button:focus {
```
randi274: Can this use a class instead? We're not using any other buttons currently, but would rather this was a more specific selector just in case we do.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardController.js
```diff
@@ -97,7 +98,22 @@
...
+            if (step === '5') {
+                model.getBatchMetadata().toggleShowAdvanced();
+                model.getBatchInfo().load(component.get('v.batchInfo'));
+            }
+        } else if (task === 'donationMatchingBehavior') {
+            var step = component.get('v.batchMetadata.progressIndicatorStep');
+            if (step === '5') {
+                model.getBatchInfo().load(component.get('v.batchInfo'));
+            }            
+        } else if(task === 'donationMatchingRule') {
```
randi274: Needs a space after the if

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -417,7 +426,12 @@
                 this.donationMatchingOptions = info.donationMatchingOptions;
                 this.donationMatchingRule = info.donationMatchingRule;
                 this.donationDateRange = info.donationDateRange;
+                this.labels = info.labels;
                 this.postProcessClass = info.postProcessClass;
+                this.processUsingScheduledJob = info.processUsingScheduledJob;
+                this.noMatchSelected = (info.donationMatchingBehavior === this.labels.DoNotMatch);
+                this.noMatchOnDate = (info.donationMatchingRule.indexOf("donation_date__c") < 0);
```
randi274: In my namespaced org, donation date is always disabled because the value is equal to npsp__donation_date__c instead of donation_date__c. This might need some more complex logic to work here.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardController.js
```diff
@@ -97,7 +98,22 @@
             } else {
                 model.getBatchMetadata().showError(possibleError);
             }
+        } else if (task === 'showAdvanced') {
```
randi274: Is it possible to break this out into using another function? This one gets a lot of noise already because it's also listening for events from the footer (definitely something we'll change with the rewrite!). Maybe change these 3 functions to one that's handleProcessingSettingsChange?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -417,7 +426,12 @@
                 this.donationMatchingOptions = info.donationMatchingOptions;
                 this.donationMatchingRule = info.donationMatchingRule;
                 this.donationDateRange = info.donationDateRange;
+                this.labels = info.labels;
```
randi274: This usage breaks the pattern a bit (insofar as I understand it), and is one of the reasons we believe we need to rewrite ourselves out of this pattern. To work around it here though, I would put this logic into the markup instead of the controller since we're only relying on it for user input rather than in any functions. Then you can remove the label assignment here, and in the View module too.

Example: where you've written 'disabled="{!v.batchInfo.noMatchSelected}"', instead you could use disabled="{!v.batchInfo.donationMatchingBehavior == v.batchMetadata.labels.DoNotMatch}"

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -338,76 +339,103 @@
...
+                                                                type="checkbox" variant="label-hidden" class="inputCheckbox"
+                                                                checked="{!v.batchInfo.runOpportunityRollupsWhileProcessing}"/>
+                                                <label class="slds-form-element__label" for="runOpportunityRollupsWhileProcessing">
+                                                    {!v.batchMetadata.labels.runOpportunityRollupsWhileProcessingLabel}
+                                                </label>
+                                                <lightning:helptext content="{!$Label.c.bgeBatchRunOpportunityRollupsWhileProcessingHelp}" />
+                                            </lightning:layoutItem>
+                                            <lightning:layoutItem class="slds-p-around_x-small" size="12">
+                                                <lightning:input aura:id="processUsingScheduledJob"
+                                                                label="{!v.batchMetadata.labels.processUsingScheduledJobLabel}"
```
randi274: I think this checkbox should be removed, since it was previously removed with this PR https://github.com/SalesforceFoundation/Cumulus/pull/3906

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -338,76 +339,103 @@
...
+                                                                type="checkbox" variant="label-hidden" class="inputCheckbox"
+                                                                checked="{!v.batchInfo.processUsingScheduledJob}"/>
+                                                <label class="slds-form-element__label" for="processUsingScheduledJob">
+                                                    {!v.batchMetadata.labels.processUsingScheduledJobLabel}
+                                                </label>
+                                                <lightning:helptext content="{!$Label.c.bgeBatchProcessUsingScheduledJobHelp}" />
```
randi274: Re: previous comment, this helptext can/should be removed too: bgeBatchProcessUsingScheduledJobHelp

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.css
ptewson-sfdo: Good thought! I like that way better. 

**NB! I do want to solicit a bit of broader feedback on this topic though - this css is basically in here to make the base button look/behave more like a link, where the principal difference that I noticed is the focus/active behavior showing the shadow/border around the un-styled button. I'm curious if the team thinks that it's actually just a bad idea to try and make a button act like a link, and if we should use one of the LDS button variants without additional theming. 

Despite appearing like a link in the mockup, this is in fact a button for two reasons - one, the LDS style guide suggests using links only when the user is directed to a new url, which is not the case here. The second is that the way all the previous user interaction is being handled in the controller is by identifying the source of the user action but the aura:id attribute, which isn't available on links.

I think the intent of showing this more like a link visually in the mock-up was to make this feature more subtle than a button, so that users who want a straightforward user experience aren't distracted into clicking it and getting confused by the advanced options. However, for obvious reasons "slds-not-very-obvious-button" is not one of the brand variants. So, in summary: 

- Maybe we shouldn't style this button at all and just use the SLDS base (or neutral?) button variants, so we aren't fighting the design system to make it look like a link? 

- If we want to make it look like a link, @randi274's suggestion is definitely the way to go.

- We shouldn't actually implement it with an <a> link however

Thoughts?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
ptewson-sfdo: This is great feedback, and fwiw here's my justification: 

My preference is to avoid putting logic directly into the markup except for convenience in cases where the logic is straightforward. The logic for enabling/disabling the date range field gets more complex since it depends on the selected fields in the matching model selection above, so that ends up being more like:

disabled="{!or(v.batchInfo.donationMatchingBehavior == v.batchMetadata.labels.DoNotMatch, v.batchInfo.donationMatchingRule.indexOf("donation_date__c") < 0)}"

or 

disabled="{!or(v.batchInfo.donationMatchingBehavior == v.batchMetadata.labels.DoNotMatch, v.batchInfo.noMatchOnDate)}"

The latter one is kind of a hybrid approach. 

All that said, I didn't feel SUPER great about copying the labels into the BatchInfo/BatchInfoView module, but for sort of complex reasons wasn't sure if it was 'breaking the model' per se, since it's information that we now need to accurately render the view.

Ultimately I think this comes down to another vote for re-factoring, and I'm probably going to take your suggestion of just moving the logic into the markup since it gets the job done and makes the delivery quite a bit simpler.  I'm a tiny bit concerned that behavior could possibly drift between the state of the 'real' data in the info module and what's currently in the rendered markup, but I think that's probably not a valid concern. 

How do people write short comments? Asking for a friend.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
ptewson-sfdo: Yup. Good call.

ptewson-sfdo:
**lurch: attach W-029479

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPB0QAM" target="_blank">W-029479</a>

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.css
ptewson-sfdo: (Going with using css class because everyone likes it)

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -338,76 +339,94 @@
...
+                                                                    value="{!v.batchInfo.donationDateRange}"/>
+                                            </lightning:layoutItem>
+                                            <lightning:layoutItem class="slds-p-around_x-small" size="12">
+                                                <label class="slds-form-element__label" for="donationMatchingClass">
+                                                    {!v.batchMetadata.labels.donationMatchingClassLabel}
+                                                </label>
+                                                <lightning:helptext content="{!$Label.c.bgeBatchDonationMatchingClassHelp}" />
+                                                <lightning:input aura:id="donationMatchingClass"
+                                                                label="{!v.batchMetadata.labels.donationMatchingClassLabel}"
+                                                                disabled="{!v.batchInfo.donationMatchingBehavior == v.batchMetadata.labels.DoNotMatch}"
```
lmeerkatz: I think we've already discussed, but `v.batchMetadata.labels.DoNotMatch` is coming from the service class and not from an actual label, correct?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -449,9 +456,10 @@
                 donationMatchingRule: [],
                 donationDateRange: '',
                 postProcessClass: '',
-
+                processUsingScheduledJob: false,
```
lmeerkatz: This shouldn't be here - we've removed all references to processUsingScheduledJob in BGE.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -418,6 +422,9 @@
                 this.donationMatchingRule = info.donationMatchingRule;
                 this.donationDateRange = info.donationDateRange;
                 this.postProcessClass = info.postProcessClass;
+                this.processUsingScheduledJob = info.processUsingScheduledJob;
+                this.noMatchOnDate = (info.donationMatchingRule.indexOf("donation_date__c") < 0 && 
+                    info.donationMatchingRule.indexOf("npsp__donation_date__c") < 0);
```
lmeerkatz: You would never reach ` info.donationMatchingRule.indexOf("npsp__donation_date__c") < 0`

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.css
```diff
@@ -1,6 +1,11 @@
 .THIS.slds-tabs_card {
     background: white;
 }
+
+/*remove focus button formatting*/
+.THIS .bgeLinkButton:focus {
+    box-shadow: 0 0 3px transparent;
```
lmeerkatz: We should be able to see when we've got focus - add an underline to the text

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
lmeerkatz: Oh wait - it's && so you would reach it, but ... you would never get a true in an unnamespaced org. Either way, not quite right.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
ptewson-sfdo: Yes, exactly

PR #3948, opened 2018-12-28
BGE: BDI only checks FLS on relevant fields when processing a batch

bethbrains:
# Critical Changes

# Changes
- BDI only checks FLS on relevant fields when processing a batch

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -36,6 +36,82 @@
 */
 public with sharing class BGE_BatchGiftEntry_UTIL {
 
+    /*******************************************************************************************************
+    * @description returns a list of DataImport__c fields the Batch Gift Entry UI needs in SOQL
+    * @param batchId the batch for which to get all fields for soql
+    * @return List<String> list of DataImport__c field api names
+    */
+    public static List<String> getDataImportFields(Id batchId) {
```
randi274: What's the thought behind having these overloaded methods? Does this follow a certain pattern? I found it difficult to read, but if this is just a larger convention then it might just be me!

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -36,6 +36,82 @@
...
+    * @return list of field api names
+    */
+    public static List<String> getActiveFieldNamesFromBatch(Id batchId) {
+
+        DataImportBatch__c batch = [SELECT Active_Fields__c FROM DataImportBatch__c WHERE Id = :batchId];
+        String activeFieldsJSON = batch.Active_Fields__c;
+        List<String> activeFieldNames = new List<String>();
+
+        if (activeFieldsJSON != null) {
+            List<BGE_ConfigurationWizard_CTRL.BGEField> activeFields = (List<BGE_ConfigurationWizard_CTRL.BGEField>)JSON.deserialize(activeFieldsJSON, List<BGE_ConfigurationWizard_CTRL.BGEField>.class);
```
randi274: Deserializing in Apex like this makes me think we should reconsider child objects...

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -36,6 +36,82 @@
...
+
+        List<String> fields = getDataImportFields(includeRelationshipFields);
+        fields.addAll(getActiveFieldNamesFromBatch(batchId));
+
+        return fields;
+    }
+
+    /*******************************************************************************************************
+    * @description returns the subset of DataImport__c fields that are part of every batch
+    * @param includeRelationshipFields whether to include relationship fields in the returned list
```
randi274: I'm a little confused about why relationship fields would or would not be included. Can you elaborate on this?

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -36,6 +36,82 @@
...
+                'DonationImported__r.Name',
+                'PaymentImported__r.Name'
+            });
+        }
+
+        return fields;
+    }
+
+    /*******************************************************************************************************
+    * @description
```
randi274: Missing description

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -350,7 +329,7 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
     */
     public static List<DataImport__c> getDataImportRecordsByBatchId(Id batchId) {
 
-        List<String> dataImportFields = getDataImportFields(batchId);
+        List<String> dataImportFields = BGE_BatchGiftEntry_UTIL.getDataImportFields(batchId);
```
randi274: 👏 👏 for moving more to the UTIL class!

src/classes/BDI_DataImportService.cls
```diff
@@ -81,7 +81,11 @@ global with sharing class BDI_DataImportService {
 
         List<DataImportBatch__c> dataImportBatches = BDI_DataImportBatch_SEL.selectByIds(new Set<Id>{ batchId });
         if (dataImportBatches.size() == 1) {
-            return dataImportBatches[0];
+            DataImportBatch__c batch = dataImportBatches[0];
+            if (batch.Active_Fields__c != null) {
```
lmeerkatz: isBlank() might be better here

src/classes/BDI_DataImportService.cls
```diff
@@ -81,7 +81,11 @@ global with sharing class BDI_DataImportService {
...
+                relevantDataImportFields = BGE_BatchGiftEntry_UTIL.getDataImportFields(batchId, false);
```
lmeerkatz: I'm not comfortable with creating a dependency on BGE_BatchGiftEntryUtil here. Should we get the selected fields as part of diSettingsFromDiBatch(DataImportBatch__c dataImportBatch) instead?

bethbrains: That method returns a Data_Import_Settings__c object, which doesn't have Active Fields on it. Is there another way we could get around this? Should the dependency go the other way? BGE is the only place we use the Active Fields JSON, so parsing it in the BGE UTIL seemed logical.

lmeerkatz: I can live with it.

src/classes/BGE_BatchGiftEntry_UTIL.cls
bethbrains: When BDI_DataImportService uses this list, it wants only the fields directly on DataImport__c. You end up with an error in BDI processing if it tries to parse a `__r.Name` because it tries to get a DFR with that from the UTIL_Describe, and it's an invalid field. 
When BGE uses this list, it needs the related fields in the SOQL query so we can use them in the UI. 
I'm certainly not sure this is the right/best or even a good way to design this. This is a big part of what I wanted to talk through. Looking forward to further thoughts!

src/classes/BGE_BatchGiftEntry_UTIL.cls
bethbrains: renamed one for clarity - it didn't need to have the same name. the other two just provide some defaulting.

lmeerkatz:
Closing without merge as we're not currently working on this

PR #3950, opened 2018-12-31
Adding TDTM Trigger for Cascade delete of Payment Allocations

### LINKED ISSUE #3949: Cascade delete Payment Allocations when related Payment is deleted
When a Payment record is deleted its related Allocations records become orphaned. This can cause incorrect GAU rollup calculations.

lparrott:
# Critical Changes

# Changes

- We have added a TDTM trigger that deletes all related Payment Allocations when a Payment is deleted.

# Issues Closed

- Fixes #3949

# New Metadata

- PMT_CascadeDeleteLookups_TDTM.cls
- PMT_CascadeDeleteLookups_TEST.cls

# Deleted Metadata

### DIFF
--- a/src/classes/CMT_FilterRuleEvaluation_TEST.cls
+++ b/src/classes/CMT_FilterRuleEvaluation_TEST.cls
@@ -88,11 +88,11 @@ private class CMT_FilterRuleEvaluation_TEST {
                 CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId2, 'Group2.Rule4', 'Opportunity', 'CreatedDate', 'Equals', 'TODAY') + ',' +
                 CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId2, 'Group2.Rule5', 'Opportunity', 'CreatedDate', 'Greater', 'YESTERDAY') + ',' +
 
-                CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId3, 'Group3.Rule1', 'Opportunity', 'CreatedDate', 'Equals', DateTime.Now().format('YYYY-MM-dd')) + ',' +
-                CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId3, 'Group3.Rule2', 'Opportunity', 'CreatedDate', 'Greater_Or_Equal', DateTime.Now().addDays(-1).format('YYYY-MM-dd')) + ',' +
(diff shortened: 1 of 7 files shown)


njjc:
@lparrott a test is failing in both managed and unmanaged builds, do you mind fixing it before I review?

force2b:
@njjc and @lparrott Out of curiosity, what happens to Allocation records attached to Payments if the Opportunity record is deleted (as opposed to deleting the individual Payment)? I don't see that in the unit tests.

lparrott:
@force2b My assumption (a dirty word, I know, we should have added a test) is that since it's Master-Detail, the Payment would also get deleted, triggering deletion of the Payment's Allocations.

force2b:
That's my concern. The trigger will not fire in that scenario. https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_triggers_ignoring_operations.htm

lparrott:
@force2b Good catch! I'll update our Work Item and push to get this additional case looked at ASAP.

force2b:
Thanks. I was looking for the WI linked to this to use for comments, but there wasn't one listed. Also, this new scenario should be added as a new WI rather than updating the closed WI. We generally link our PR's to WI's using lurch: attach.

lparrott:
**lurch: attach W-027724

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fN0UAAU" target="_blank">W-027724</a>

smarovich:
Hi @lparrott @njjc, I noticed this in the release notes and thought we should probably either remove the release note OR add a sentence saying something like "This trigger was created to support a future feature and shouldn't be used until that feature is released. Please ignore the trigger for now. " (like we did with the other Payment Allocations stuff). Let me know how you want to proceed. Thanks!

lparrott:
@smarovich Removing that line from the Release notes is fine, but this trigger WILL be automatically activated for any new installs of NPSP after this release. It won't hurt anything, it should honestly have been included in the original merge of Payment Allocations, but we didn't want to delay it anymore!

smarovich:
Thanks for the context @lparrott. I think in that case, we should include the note but also add the sentence about this trigger being used for a future feature (since Payment Allocations is still hidden, right?).

njjc:
I think nothing in the release notes is best! 
But that's, like, just my opinion, man.

judisohn:
Given that it's a trigger shouldn't fire for any customer who hasn't enabled payment allocations in custom settings, can we say something like "This trigger was created to support a future feature and should have no visible impact." 99% of admins already ignore Triggers as Triggers anyway until it does DML.

smarovich:
Sure thing, @judisohn! I'll revise the release note. 

@njjc, I hear you, but I've learned that it's best to mention things that people might notice (even if they're not being used yet).

PR #3979, opened 2019-01-09
Removing hardcoded permission references in test class to avoid dependencies.

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/USER_UserService_TEST.cls
+++ b/src/classes/USER_UserService_TEST.cls
@@ -291,13 +291,15 @@ public with sharing class USER_UserService_TEST {
         // on which CustomizeApplication settings depends are set as well
         PermissionSet permission = new PermissionSet(
             Name = 'PermissionTest',
-            Label = 'PermissionTest',
-            PermissionsManageCustomPermissions = true,
-            PermissionsViewRoles = true, 
-            PermissionsViewSetup = true,
-            PermissionsManageTranslation = true,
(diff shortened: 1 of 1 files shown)


src/classes/USER_UserService_TEST.cls
```diff
@@ -291,13 +291,17 @@ public with sharing class USER_UserService_TEST {
...
-            PermissionsCustomizeApplication = true
+            Label = 'PermissionTest'
         );
+
+        // Dynamically adding permissions to avoid dependencies, previously our package failed to deploy
+        // to an org without translations enabled, due to the field not existing.
+        permission.put('PermissionsManageCustomPermissions',true);
+        permission.put('PermissionsViewRoles',true);
+        permission.put('PermissionsViewSetup',true);
+        permission.put('PermissionsManageTranslation',true);
```
force2b: @jjbennett Is it necessary to have this particular permission defined?

davisagli: @force2b In quip she said: so it looks like if translations are enabled then you have to have that set or it won't let you insert System.DmlException: Insert failed. First exception on row 0; first error: FIELD_INTEGRITY_EXCEPTION, Permission CustomizeApplication depends on permission(s): ManageTranslation: []

force2b: Makes sense. Thanks.

PR #3984, opened 2019-01-10
Fix test that fails in orgs without Translations enabled.

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/USER_UserService_TEST.cls
```diff
@@ -287,20 +287,24 @@ public with sharing class USER_UserService_TEST {
...
-        // Dynamically adding permissions to avoid dependencies, previously our package failed to deploy
-        // to an org without translations enabled, due to the field not existing.
-        permission.put('PermissionsManageCustomPermissions',true);
-        permission.put('PermissionsViewRoles',true);
-        permission.put('PermissionsViewSetup',true);
-        permission.put('PermissionsManageTranslation',true);
-        permission.put('PermissionsCustomizeApplication',true);
+        // Dynamically adding translation permission to avoid dependencies. Previously our package failed to deploy
+        // to an org without translations enabled, since the field did not exist.
+        if (PermissionSet.getSObjectType().getDescribe().fields.getMap().containsKey(translationPerm)) {
```
davisagli: When I tried this in the debug console the key it returned was lowercase.

jjbennett:
@force2b - can you review this change to the unit test fix. Although we removed the dependency on translations, the test will fail in orgs without translations enabled. @davisagli suggested a different approach which fixes both the dependency and the test failure. This is needed so that our scratch orgs that run beta tests can have translations not enabled to simulate customer orgs without translations.

src/classes/USER_UserService_TEST.cls
jjbennett: @davisagli  A field map is special and is case insensitive. They keyset itself is case sensitive but I switched to containsKey vs checking the keyset.

davisagli: Got it!

PR #3989, opened 2019-01-11
Feature/patrick xss ep template

patrick-yan-sf:
Add escaping to the labels that are used in the dynamic button.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


patrick-yan-sf:
**lurch: Attach W-030041

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPqqQAE" target="_blank">W-030041</a>

patrick-yan-sf:
@force2b This is ready for review

src/classes/EP_ManageEPTemplate_CTRL.cls
```diff
@@ -455,10 +455,11 @@ public with sharing class EP_ManageEPTemplate_CTRL {
         Component.Apex.OutputPanel formPanel = new Component.Apex.OutputPanel();
     
         //creates a box around the form with a label for the hierarchy level, and a color bar to more easily grock which level we're at
+        // uses escape = false but the labels are escaped
```
force2b: Is there more this comment should state as to why the labels are escaped, but `escape=false` is included on the component?

patrick-yan-sf: @force2b - yeah I just wanted to add some comments so that future readers who see the escape = false know that it was intentional but that we escaped the inputs.  Thanks

PR #3990, opened 2019-01-11
Update trial metadata for BGE

davisagli:
Updates to post-install metadata to match trial experience:
- Add BGE permission set (not actually assigned to anyone since the scratch org admin should already have access via the Admin profile)
- Add fields to NPSP Data Import Batch Page Layout
- Add Gift Batch field to NPSP Data Import Batch List Views (I put this one in the package)

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/objects/DataImportBatch__c.object
```diff
@@ -360,6 +360,7 @@
         <columns>NAME</columns>
         <columns>LAST_UPDATE</columns>
         <columns>Batch_Description__c</columns>
+        <columns>GiftBatch__c</columns>
```
jjbennett: Is this adding a field to the DataImportBatch object and do we want to do that?

davisagli: The field was added previously for BGE. This adds it to the "All" list view of DataImportBatch. In the TSO @jesslopez-sf did that as a customization to the packaged list view, but I thought it would be simpler to make the change in the packaged list view rather than duplicate the list view definition with a different list of columns in unpackaged/trial. Is there a reason not to make this change in the package?

PR #3991, opened 2019-01-14
[DO NOT MERGE] Feature/jbrandolini/map opp stages to finalized and committed

jbrandolinisf:
# Critical Changes

# Changes
* You may notice two new fields on the Household Settings custom settings page: Opportunity Committed Stage and Opportunity Finalized Stage. These fields were created to support a future feature and shouldn't be used until the feature is released. Please ignore the fields for now.

# Issues Closed

# New Metadata

Classes:
* OPP_StageToStateMappingService
* OPP_StageToStateMappingService_TEST

Fields:
* npo02__Households_Settings__c.Opportunity_Committed_Stages__c
* npo02__Households_Settings__c.Opportunity_Finalized_Stages__c

# Deleted Metadata


jbrandolinisf:
**lurch: attach W-030114

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPwjQAE" target="_blank">W-030114</a>

jbrandolinisf:
@annejackson - Testing Notes:

To invoke API from Dev Console,

System.debug(OPP_StageToStateMapping_API.determineOppStateFromOppStage(<<Opportunity Stage>>);

For example,

System.debug(OPP_StageToStateMapping_API.determineOppStateFromOppStage('Closed Won');

Note, the following:

Until the settings fields are populated, everything will return UNCOMMITTED.  Since we are not exposing this via the UI, you will need to go to the Custom Settings area, and click the Manage button for the setting labeled "Household Settings".  Enter Opportunity Stage names, separated by ;  (to allow multiple values)

Nothing at this point enforces what you can/cannot set here, so it is possible to have a value of John defined as an Opportunity Committed or Finalized Stage, even though it is not an actual Opportunity Stage.  The Settings UI to be built will enforce this.  Further, you could have the same value in both fields.  If this occurs, it will return FINALIZED when you call with that value.

Additionally, the method does not check to see if it is a valid Opportunity Stage Name, but instead assumes it is.

I recommend testing the API method with no stages defined in the two settings fields (the default), one value, and multiple values-- with and without overlap, confirming it selects the right value FINALIZED if present in the Opportunity Finalized Stage(s) field, COMMITTED if present in . the Opportunity Committed Stage(s) field (but not in the Opportunity Finalized Stage(s) field, and UNCOMMITTED in every other case (including if null/completely made up).

jbrandolinisf:
@weizsforg - Go for it!  Looking forward to your feedback!!!

src/classes/OPP_StageToStateMapping_API.cls
```diff
@@ -0,0 +1,149 @@
...
+    * NOTE: Opportunity Stage is treated as Case Insensitive, so Closed Won CLOSED WON and closed won are
+    * all the same for this purpose.
+    * @param String opportunityStage The opportunity stage to check the state of
+    * @return OPPORTUNITY_STATE enum value representing the current opportunity state
+    ********************************************************************************************************/
+    global static OPPORTUNITY_STATE determineOppStateFromOppStage(String opportunityStage) {
+
+        initializeStates();
+
+        String oppStageToCheck = String.isNotBlank(opportunityStage) ? opportunityStage.toLowerCase(opportunityStage) : null;
```
weizsforg: Just curious, any specific reason that we use toLowerCase(opportunityStage) instead of toLowerCase() here?

jbrandolinisf: Nope.  I will correct that.  It was me being redundant.  It was me being redundant.  It was me being redundant.

jbrandolinisf: Corrected.

PR #3992, opened 2019-01-15
Feature/lvl custom setting batch size

ptewson-sfdo:
Provide a new custom settings object and setting for providing a configurable batch size for the scheduled level calculation.

# Critical Changes

# Changes

- You can now configure the batch size for the Level Assignments nightly scheduled job using the new Level Assignments Batch Size field located in NPSP Settings | Bulk Data Processes | Batch Process Settings. If this field is left blank, it will default to 200.

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/STG_SettingsService.cls
```diff
@@ -159,6 +159,17 @@ public with sharing class STG_SettingsService {
         private set;
     }
 
+    /*********************************************************************************************************
+    * @description The Levels Settings object
+    */
+    public Levels_Settings__c stgLevels {
+        get {
+            if (stgLevels == null) stgLevels = UTIL_CustomSettingsFacade.getOrgLevelsSettings();
```
lmeerkatz: Our style guide doesn't allow inline ifs

src/objects/Levels_Settings__c.object
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: @ptewson-sfdo I'm not sure it's worth adding a new custom object for this single field. If anything I would recommend a BatchSettings object so we can add other batch job size settings going forward. However, given that all the Rollup batch job size settings are in the CustomizableRollups object and some of the other batch job size settings are in other settings objects, I would suggest this just be added as a field to the Contacts & Organizations or Households custom settings object. It would eliminate a lot of the code and metadata/

lmeerkatz: @force2b We discussed this and it sounds like there is potential for additional Levels settings, so decided to add this now even if it's a bit thin.

src/classes/STG_SettingsService.cls
```diff
@@ -231,6 +242,7 @@ public with sharing class STG_SettingsService {
 	        upsert stgErr;
 	        upsert stgAllo;
```
lmeerkatz: If you felt like fixing the weird alignment here, I wouldn't hate it.

src/objects/Levels_Settings__c.object
force2b: Oh. What other level settings are you thinking of? I don't really see anything in the backlog for Levels. If we do want to keep that object then the PR will need the translation files.

lmeerkatz: We were thinking of a setting for filtering records included in the batch - see W-030738. 

Good call on translations!

src/pages/STG_PanelSchedule.page
```diff
@@ -84,6 +84,18 @@
                     </div>
                 </div>
             </div>
+
+            <div class="slds-grid slds-gutters_small">
+                <div class="slds-col slds-has-flexi-truncate slds-grid" role="listitem">
+                    <div class="slds-form-element">
+                        <apex:outputLabel value="{!$ObjectType.Levels_Settings__c.Fields.Level_Assignment_Batch_Size__c.Label}" for="tbxLA" styleClass="slds-form-element__label" />
```
lmeerkatz: tbxLA?

src/classes/LVL_LevelAssign_SCHED.cls
```diff
@@ -46,6 +49,13 @@ public without sharing class LVL_LevelAssign_SCHED implements Schedulable {
     * @description Starts one batch job per level object.
     */
     public void runBatch() {
+        Integer batchSize;
+        try {
+            batchSize = levelsSettings.Level_Assignment_Batch_Size__c.intValue();
+        } catch(Exception e) {
```
lmeerkatz: Would this cause something other than a null pointer exception?

src/objects/Levels_Settings__c.object
force2b: If that’s the wi Jess just added today, there are no settings needed on that wi, at least we didn’t identify any in refinement.

force2b: Just to close the loop from the WI posting, let's keep with this object. I will update the other WI to reference this.

src/classes/LVL_LevelAssign_SCHED.cls
ptewson-sfdo: Given that the field is defined as a Number field with scale 0, probably not ... this direction is a bit of paranoia that a user might find a creative way to figure that out, but in truth I don't believe its possible for it to be anything but a whole number (a negative/nonzero check might be a good idea though?)

lmeerkatz: In general, we should be specific about the exceptions we're catching. If we catch something unrelated in a weird spot it's going to mask unexpected problems.

src/objects/Levels_Settings__c.object
ptewson-sfdo: Thanks for the hot tip on translations! Will include that asap.

src/classes/STG_SettingsService.cls
```diff
@@ -153,12 +164,27 @@ public with sharing class STG_SettingsService {
     */
     public Customizable_Rollup_Settings__c stgCRLP {
         get {
-            if (stgCRLP == null) stgCRLP = UTIL_CustomSettingsFacade.getOrgCustomizableRollupSettings();
+            if (stgCRLP == null) {
+                stgCRLP = UTIL_CustomSettingsFacade.getOrgCustomizableRollupSettings();
+            }
```
lmeerkatz: Thank you for the {}!

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -910,4 +912,42 @@ public without sharing class UTIL_CustomSettingsFacade {
...
+                if(!Test.isRunningTest()) {
+                    upsert orgLevelsSettings;
+                }
+            }
+        }
+        return orgLevelsSettings;
+    }
+
+    private static void configLevelsSettings(Levels_settings__c ls) {
+        // New default settings for the Customizable Rollups engine
```
lmeerkatz: Copied comment from somewhere else?

ptewson-sfdo:
lurch: **attach W-017997

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000005KmZhAAK" target="_blank">W-017997</a>

jesslopez-sf:
I pulled this into a scratch org yesterday, entered 2000 as the batch size,  and confirmed the Bulk Data Process for Levels used my new batch size. I also let the scheduled job run overnight and confirmed this morning that it also used the new batch size. Looks great. Thank you!

smarovich:
Release notes are updated! Let me know if you'd like me to many any changes @ptewson-sfdo

src/objects/Levels_Settings__c.object
force2b: @ptewson-sfdo Holding my final approval until the translation files are in to avoid the build failures in master. CC @rponti-sforg

ptewson-sfdo: @force2b @rponti-sforg - are we looking for more translation files than the files like "Levels_Settings__c-nl_NL.objectTranslation" that are already included?

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x80000005KmZhAAK/view

Merging PR and and deleting branch.

PR #4004, opened 2019-01-17
Adds filter on SystemModStamp to Levels batch job

seanpat09:
# Critical Changes

# Changes

* To improve the performance of the Level Assignment Batch Job in larger data volume organizations, we updated the batch job to only process Accounts or Contacts that have been updated since the last time the related Level Assignment Batch job had run. This filtering is ignored if Levels reference any formula fields.

# Issues Closed

# New Metadata

# Deleted Metadata


seanpat09:
**lurch: attach W-030738

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQlNQAU" target="_blank">W-030738</a>

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -270,32 +287,106 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
-                setFlds = new Set<String>();
-                mapStrObjectToQueryFields.put(strObj, setFlds);
+        return levelWrappers; 
+    }
+
+    /*********************************************************************************************************
+    * @description Wrapper class for the Level__c object
+    */
+    public class Level {
+        public String target { get { return levelRecord.Target__c; } }
```
force2b: Should these be separated onto different lines rather than a single line?

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -270,32 +287,106 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
+        private Level__c levelRecord;
+        private Boolean targetIsAccount;
+        private Boolean targetIsContact;
+
+        /*********************************************************************************************************
+        * @description Constructor
+        */
+        public Level(Level__c levelRecord) {
+            this.levelRecord = levelRecord;
+            targetIsAccount = target == String.valueOf(Account.SObjectType);
```
force2b: Rather than set the var in the constructor, could the var be defined as 
```
public Boolean targetIsAccount { 
   get { 
      return target == String.valueOf(Account.SObjectType);
   }
   private set; 
} 
```

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -270,32 +287,106 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
+                datetimeToFilter = levelsSettings.LastJobStartTimeContact__c;
+            } else {
+                return '';
+            }
+
+            return ' WHERE SystemModStamp > ' + datetimeToFilter.formatGMT(DATETIME_SOQL_FORMAT);
+        }
+
+        /*********************************************************************************************************
+        * @description Determines is SystemModStamp filter should be built
```
force2b: Change `is` to `if` and maybe add more information as to the logic used for the determination

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -270,32 +287,106 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
+            this.levelRecord = levelRecord;
+            targetIsAccount = target == String.valueOf(Account.SObjectType);
+            targetIsContact = target == String.valueOf(Contact.SObjectType);
+        }
+
+        /*********************************************************************************************************
+        * @description Builds query to get object data need to calculate and assign levels
+        * @returns SOQL string for querying SObjects for level calculation
+        */
+        public String buildQuery() {
```
force2b: I'm not clear how this works when there is more than one Level record for an Account or Contact. This method is for a single Level record, however there could be an unlimited number of Level record for the Account or Contact object.

src/classes/LVL_LevelAssign_SCHED.cls
```diff
@@ -60,23 +60,11 @@ public without sharing class LVL_LevelAssign_SCHED implements Schedulable {
...
-                // construct our query
-                String strSoql = 'select Id';
-                for (String strFld : setFlds) {
-                    strSoql += ', ' + strFld;
-                }
-                strSoql += ' from ' + strObj;
-                
-                // kick off our batch
-                LVL_LevelAssign_BATCH batch = new LVL_LevelAssign_BATCH(strSoql, strObj);
+            for (LVL_LevelAssign_BATCH.Level level : levels) {
```
force2b: There should be one batch job that runs for all the Level records on the Account or Contact (i.e., a max of 2 Level jobs). However, this appears to be setup to run one job per Level record. The way it used to work was to use a Map of Levels by Object, so this for loop never had more than two iterations (`mapStrObjectToQueryFields.keySet()`).

seanpat09: Ah I completely misread the original query! I'll make the fix

seanpat09:
@force2b I updated the code to group the levels together by SObject/Target and also added a test to confirm that a batch is run for each Target as opposed to Level__c record. Thanks for catching that, I would have never noticed with the tests passing

src/classes/LVL_LevelAssignBATCH_TEST.cls
```diff
@@ -269,13 +278,17 @@ private class LVL_LevelAssignBATCH_TEST {
         sched.runBatch();
         Test.stopTest();
         
+        System.assertEquals(2, [SELECT COUNT() FROM AsyncApexJob WHERE JobType = 'BatchApex'],
+            'Two batches should run, one for Contact and one for Account'
+        );
+
```
jjbennett: Do we want to assert that both the Account and Contact fields were stamped here as well?

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -267,35 +284,149 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
-            if (setFlds == null) {
-                setFlds = new Set<String>();
-                mapStrObjectToQueryFields.put(strObj, setFlds);
+        return ladderByTarget.values();
+    }
+
+    /*********************************************************************************************************
+    * @description Stores all the Level__c records for an SObject
+    */
+    public class Ladder {
```
force2b: This took me a while to understand. I couldn't figure out why it was called a Ladder, but I see now that there are basically two instances of Ladder (one for Account and one for Contact) and that each contains a collection of Levels. Makes perfect sense now that I've read the code.

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -230,9 +239,17 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
     private class LevelException extends Exception {}
     
     /*********************************************************************************************************
-    * @description Batch process finish method, does nothing.
+    * @description Batch process finish method, stores last job start time for accounts and contacts
     */
-    public void finish(Database.BatchableContext BC) {}
+    public void finish(Database.BatchableContext BC) {
+        if (strObject == String.valueOf(Account.SObjectType)) {
+            levelsSettings.LastJobStartTimeAccount__c = batchStartTime;
```
jjbennett: Do we need to verify the batch succeeded in the finish method?

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -267,35 +284,149 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
+            DateTime datetimeToFilter;
+            if (targetIsAccount) {
+                datetimeToFilter = levelsSettings.LastJobStartTimeAccount__c;
+            } else if (targetIsContact) {
+                datetimeToFilter = levelsSettings.LastJobStartTimeContact__c;
+            } else {
+                return '';
+            }
+
+            return ' WHERE SystemModStamp > ' + datetimeToFilter.formatGMT(DATETIME_SOQL_FORMAT);
```
jjbennett: Should this be in a variable even if just at the top of the method?

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -267,35 +284,149 @@ public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable
...
+            } else {
+                return true;
+            }
+        }
+
+        /*********************************************************************************************************
+        * @description Determines if any source field is a formula field
+        * @returns Boolean true if any source field is a formula, otherwise false
+        */
+        private Boolean aSourceFieldIsAFormula() {
```
jjbennett: What do you think about isAnySourceFieldAFormula?

src/classes/LVL_LevelAssign_BATCH.cls
```diff
@@ -34,7 +34,12 @@
...
 public without sharing class LVL_LevelAssign_BATCH implements Database.Batchable<sObject>, Database.Stateful {
-    
+    /** @description Custom Settings record that defines Levels settings */
+    @TestVisible
+    private static Levels_Settings__c levelsSettings = UTIL_CustomSettingsFacade.getLevelsSettings();
+
+    @TestVisible
+    private static final String DATETIME_SOQL_FORMAT = 'yyyy-MM-dd\'T\'HH:mm:ss\'Z\'';
     /** @description The query for the batch process to run on.*/
     private String strSoql;
```
force2b: I know this is mostly out of the scope of this particular WI, but this is a small class. I think we should fix some of the glaring style issues:
* Remove hungarian notation
* Proper case data types
* Upper case SOQL keywords

src/objects/Levels_Settings__c.object
```diff
@@ -2,6 +2,24 @@
...
+        <description>The last time the Level assignment batch job ran successfully for the Account object</description>
+        <externalId>false</externalId>
+        <label>Last Job Start Time : Account</label>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>DateTime</type>
+    </fields>
+    <fields>
+        <fullName>LastJobStartTimeContact__c</fullName>
+        <description>The last the time the Levels assignment batch job ran for the Contact object</description>
```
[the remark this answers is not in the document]
seanpat09: Oh good point, I'll add the worse successfully and fix the descriptions so they match

src/classes/LVL_LevelAssign_BATCH.cls
force2b: Do you think this should be `>=` to catch Accounts/Contacts that might have been updated at exactly the same moment the job previously started?

src/classes/LVL_LevelAssign_BATCH.cls
seanpat09: that's a good idea, I'll do a query on the batch job and check if the status is Completed before setting the fields

src/classes/LVL_LevelAssign_BATCH.cls
seanpat09: That's probably safest the safest approach, I'll do that

src/classes/LVL_LevelAssign_BATCH.cls
seanpat09: I'll add a little more to the description to help clarify

src/classes/LVL_LevelAssign_BATCH.cls
seanpat09: `datetimeToFilter` won't be populated until this line, but I can put the rest of the of string in a variable like so:

```
String filter = ' WHERE SystemModStamp > {0}'

...

String formattedDateTime = datetimeToFilter.formatGMT(DATETIME_SOQL_FORMAT);
return String.format(filter, new List<String>{formattedDateTime});
```

jesslopez-sf:
@seanpat09 @force2b @jjbennett @rponti-sforg 
I did some proactive and extensive testing for this Pull Request. And I'm happy to report that all of my tests worked as expected. Thank you!! 👍

force2b:
@rponti-sforg  Lock errors can be an issue with parallel testing. I restarted the build

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fQlNQAU/view

PR #4006, opened 2019-01-18
Deprecate opp matching button.

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/pages/OPP_MatchingDonationsBTN.page
```diff
@@ -1,3 +1,3 @@
-<apex:page standardController="Opportunity" extensions="OPP_MatchingDonationsBTN_CTRL" action="{!MatchingDonationPage}">
-    <apex:pageMessages />
+<apex:page standardController="Opportunity">
+    This button has been deprecated.
```
davisagli: Is it worth using a label so it's translatable?

jjbennett: I was wondering that myself. I would think not since the message is temporary and probably wouldn't be translated before it was ultimately deleted.

davisagli: Keep in mind that even once we delete it from the managed package, my understanding is that the VF page will remain in existing subscriber orgs until an admin deletes it.

jjbennett: With buttons and fields we usually add DEPRECATED to the label before deletion, so I just made that change to the Visualforce page as well. However, let me know if you really feel this needs to be translatable and I can add a label. Maybe a generic DEPRECATED label that can be used down the road? I also created a PR in the underlying package to remove the button that calls this page.

davisagli: I like the idea of a generic "This component has been deprecated" label but I don't personally feel strongly about what we do here.

jjbennett:
@davisagli - Thanks for catching those, I was searching for the page vs the underlying package button! I also switched the hardcoded text to a label.

davisagli:
@jjbennett do you think we need to reference the label from Apex so it still gets spidered once the page is deleted? I'm not sure how that would be handled in a subscriber org.

jjbennett:
@davisagli I thought about proactively adding it, but I decided to wait until we delete the page. It is marked protected so we could technically delete it but if we want to be consistent and potentially have that label translated in the future I will just add it now with a note as to why we want to keep it when it isn't in use.

jjbennett:
@rponti-sforg - This one is ready, we will need to swap out the underlying Contacts and Org package and use the beta package instead. Please let me know when you are ready to test and I can either create a scratch org for you or walk you through the process.

rponti-sforg:
> @rponti-sforg - This one is ready, we will need to swap out the underlying Contacts and Org package and use the beta package instead. Please let me know when you are ready to test and I can either create a scratch org for you or walk you through the process.

@jjbennett Thank you Jenny.  So, to test this, does the scratch org need to have the **install_beta** or **release_beta** flow applied or is there something more altogether?  I have installed those types of flows before.  But, if it's something more involved, I would like to step through that process.  If you are available after stand-up today, we can do it then.

jjbennett:
@rponti-sforg - To test this one, we just need to verify the page now shows the deprecated message. Since the button to the page has been removed the easiest way to verify is to go to Setup -> Visualforce Pages, click on the page name "OPP_MatchingDonationsBTN.page" then click on the Preview button.

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x80000005QChJAAW/view

PR #4015, opened 2019-01-23
Feature/robot  relationships updates

skristem:
# Critical Changes
Updated relationships test case to check reciprocal relation and capture screenshot in order to manually verify the relationship viewer 
# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


skristem:
@davisagli Load Related List Keyword seems to work on my local machine but keeps failing in builds. I have used this keyword twice so far and both test cases fail on this step. Can u plz take a look

davisagli:
@skristem I added "Select Tab  Related" before "Load Related List" and now it passes.

PR #4020, opened 2019-01-28
CRLP: Add contact filter to OCR and PSC subqueries

### LINKED ISSUE #4014: Customizable Rollups Soft Credit Queries Pull All OCRs and Partial Soft Credits vs. for Single Entity
When our Customizable Rollups run on soft credits, the [CRLP_Batch_Base](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/CRLP_Batch_Base.cls#L287-L289) class pulls all OCRs and Partial Soft Credits for a list of Opportunity Ids vs. just for the related entity Ids. This is happening for both single Contacts/Accounts using the Recalculate Rollups button and all Contacts/Accounts using the batch jobs. 

**Real World Example:**
Donor One gives monthly through a workplace giving program. There is an Opportunity for each monthly donation (Account = the workplace giving Account) and hundreds of OCRs/Partial Soft Credits for each donor on each Opportunity. When we calculate Donor One's Soft Credits, we find all her Opportunities and then all of her Opportunities' OCRs and Partial Soft Credits, rather than just hers.

This results in the following errors:
1. Inline query has too many rows for direct assignment, use FOR loop
2. First error: npsp:Too many query rows: 50001

Stacktrace available at the time of saving the record - 
`Class.npsp.CRLP_Batch_Base.buildPartialSoftCreditRecordsFromOCRs: line 290, column 1 Class.npsp.CRLP_Batch_Base_Skew.execute: line 289, column 1`

This is the line of code that @force2b pointed as the culprit, in [CRLP_Batch_Base (Lines 287-289)](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/CRLP_Batch_Base.cls#L287-L289): 
`    String oppQuery = CRLP_Query_SEL.buildDetailObjectQueryForRollup(Opportunity.SObjectType,
                new List<SObjectType>{ OpportunityContactRole.SObjectType, Partial_Soft_Credit__c.SObjectType }) +
                ' WHERE Id IN :oppIds';`

**_Which then looks like this:_**
`SELECT Id,IsWon,Amount,RecordTypeId,RecordType.DeveloperName,IsClosed,CloseDate,CampaignId,npe01__Member_Level__c,npe01__Membership_Origin__c,npe01__Membership_End_Date__c,npe01__Membership_Start_Date__c,AccountId,npsp__Primary_Contact__c,npe03__Recurring_Donation__c, (SELECT Id,ContactId,Contact.AccountId,OpportunityId,Role FROM OpportunityContactRoles), (SELECT Id,npsp__Role_Name__c,npsp__Amount__c,npsp__Opportunity__c,npsp__Contact__c,npsp__Contact__r.AccountId FROM npsp__Partial_Soft_Credits__r), Account.npe01__SYSTEMIsIndividual__c FROM Opportunity WHERE Id IN :oppIds
`

bethbrains:
# Critical Changes

# Changes

# Issues Closed
Fixes #4014 
# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CMT_FilterRuleEvaluation_TEST.cls
+++ b/src/classes/CMT_FilterRuleEvaluation_TEST.cls
@@ -161,10 +161,15 @@ private class CMT_FilterRuleEvaluation_TEST {
         insert psc;
 
         // Query the Opportunity and the child OCR and Payment records to use in the evaluation
-        List<Id> contactIds = new List<Id>{con.Id};
+        List<Id> contactIds = new List<Id>();
         String soql = CRLP_Query_SEL.buildDetailObjectQueryForRollup(Opportunity.SObjectType,
                 new List<SObjectType>{OpportunityContactRole.SObjectType, npe01__OppPayment__c.SObjectType, Partial_Soft_Credit__c.SObjectType});
         opp = database.query(soql);
+        System.assertEquals(0, opp.OpportunityContactRoles.size());
+
+        contactIds = new List<Id>{con.Id};
+        opp = database.query(soql);
+        System.assertEquals(1, opp.OpportunityContactRoles.size());
 
         Test.startTest();
 
--- a/src/classes/CMT_FilterRuleEvaluation_TEST.cls
+++ b/src/classes/CMT_FilterRuleEvaluation_TEST.cls
@@ -164,35 +164,35 @@ private class CMT_FilterRuleEvaluation_TEST {
         List<Id> contactIds = new List<Id>();
         String soql = CRLP_Query_SEL.buildDetailObjectQueryForRollup(Opportunity.SObjectType,
                 new List<SObjectType>{OpportunityContactRole.SObjectType, npe01__OppPayment__c.SObjectType, Partial_Soft_Credit__c.SObjectType});
-        opp = database.query(soql);
+        opp = Database.query(soql);
         System.assertEquals(0, opp.OpportunityContactRoles.size());
 
         contactIds = new List<Id>{con.Id};
-        opp = database.query(soql);
+        opp = Database.query(soql);
         System.assertEquals(1, opp.OpportunityContactRoles.size());
 
         Test.startTest();
 
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
+         System.Assert(CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
                  opp
          }, filterGroupId1));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
+         System.Assert(CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
                  opp
          }, filterGroupId2));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
+         System.Assert(CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
                  opp
          }, filterGroupId3));
 
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(
+         System.Assert(CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(
                  new List<SObject>{
                          opp, opp.OpportunityContactRoles[0]}, filterGroupId2));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(
+         System.Assert(CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(
                  new List<SObject>{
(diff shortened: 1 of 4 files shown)


bethbrains:
**lurch: attach W-030871

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQwLQAU" target="_blank">W-030871</a>

src/classes/CRLP_Query_SEL_TEST.cls
```diff
@@ -54,6 +54,7 @@ private class CRLP_Query_SEL_TEST {
         List<Account> accts = Database.query(soql);
         System.AssertEquals(0, accts.size(), 'There should be no results, but the query should work');
 
+        List<Id> contactIds = new List<Id>();
```
force2b: I think this should include a note as to why the var is needed.

force2b: I also think might be helpful to assert that the query string includes the additional WHERE filter against contactIds.

src/classes/CMT_FilterRuleEvaluation_TEST.cls
```diff
@@ -136,49 +136,60 @@ private class CMT_FilterRuleEvaluation_TEST {
...
                 new List<SObjectType>{OpportunityContactRole.SObjectType, npe01__OppPayment__c.SObjectType, Partial_Soft_Credit__c.SObjectType});
-        o = database.query(soql);
+        opp = database.query(soql);
 
         Test.startTest();
 
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId1));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId2));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId3));
+         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
```
force2b: This is good. I think you can go one step further by adding two more asserts:

1. Set contactIds to an empty list and assert that that the query returns 0 OCR records.
2. Set contactIds as you have above and assert the that the query returns the expected number of OCR records.

src/classes/CMT_FilterRuleEvaluation_TEST.cls
```diff
@@ -136,49 +136,65 @@ private class CMT_FilterRuleEvaluation_TEST {
...
+                Amount__c = opp.Amount
         );
         insert psc;
 
         // Query the Opportunity and the child OCR and Payment records to use in the evaluation
+        List<Id> contactIds = new List<Id>();
         String soql = CRLP_Query_SEL.buildDetailObjectQueryForRollup(Opportunity.SObjectType,
                 new List<SObjectType>{OpportunityContactRole.SObjectType, npe01__OppPayment__c.SObjectType, Partial_Soft_Credit__c.SObjectType});
-        o = database.query(soql);
+        opp = database.query(soql);
```
lmeerkatz: database -> Database (also on 171)

src/classes/CMT_FilterRuleEvaluation_TEST.cls
```diff
@@ -136,49 +136,65 @@ private class CMT_FilterRuleEvaluation_TEST {
...
 
         Test.startTest();
 
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId1));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId2));
-         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{o}, filterGroupId3));
+         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
+                 opp
+         }, filterGroupId1));
+         System.AssertEquals(True, CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup(new List<SObject>{
```
lmeerkatz: You could use System.assert() instead of System.assertEquals(true, ...)

src/classes/CRLP_Query_SEL.cls
```diff
@@ -220,7 +220,16 @@
      * @return part of a SOQL string
      */
     private static String buildChildQuery(SObjectType childObject, String relationshipName) {
-        String childQuery = ', (' + buildObjectQueryForRollup(childObject).split(' FROM ')[0] + + ' FROM ' + relationshipName + ')';
+        String childQuery = ', (' + buildObjectQueryForRollup(childObject).split(' FROM ')[0] + + ' FROM ' + relationshipName;
```
lmeerkatz: What is the double plus doing here? ``` + +```

force2b: Good catch @lmeerkatz. Totally missed the two pluses.

PR #4024, opened 2019-01-30
Update UTIL_CurrencyCache to check for null rates

AndersonTarren:
# Critical Changes

# Changes
Disabled currencies come back with a null value in cachedMapOfRatesByCurrencyCode.  Adding a null check to avoid errors for this case.

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/UTIL_CurrencyCache.cls
```diff
@@ -77,6 +77,12 @@ public class UTIL_CurrencyCache implements UTIL_CurrencyCache.Interface_x {
             // If a set of rates was retrieved, find the effective rate based on the date
             if (cachedMapOfRatesByCurrencyCode.containsKey(currCode)) {
                 UTIL_CurrencyCache.CurrencyData currData = cachedMapOfRatesByCurrencyCode.get(currCode);
+                
+                //In this case the currency was likely disabled and the rates are null e.g. {AUD=null} 
+                //Return 1, to align with existing logic.
+                if(currData == null){
```
dvenor: Being super pedantic here but our coding standard has a space between the `)` and the `{`.  We should check out Uncrustify for our code formatting, or you might know of something even better.

src/classes/UTIL_CurrencyCache.cls
```diff
@@ -77,6 +77,11 @@ public class UTIL_CurrencyCache implements UTIL_CurrencyCache.Interface_x {
             // If a set of rates was retrieved, find the effective rate based on the date
             if (cachedMapOfRatesByCurrencyCode.containsKey(currCode)) {
                 UTIL_CurrencyCache.CurrencyData currData = cachedMapOfRatesByCurrencyCode.get(currCode);
+                
+                //Currency data is null.  Return 1, to align with existing logic and avoid NPE.
```
dvenor: This comment is probably not entirely necessary.  It kinda just rehashes the code block below it.  Also, I don't know what NPE is.  Maybe something like this might explain it a little better:
> If the currency data can't be found in the cache, return a default rate of 1

src/classes/UTIL_CurrencyCache_TEST.cls
```diff
@@ -86,6 +90,27 @@ private class UTIL_CurrencyCache_TEST {
         System.AssertEquals(0, UTIL_CurrencyCache.cachedMapOfRatesByCurrencyCode.size(), 'The cache should be empty');
     }
 
+    /**
+     * @description Validate that the retrieveAndCacheExchangeRates method doesn't map null CurrencyData values when queried.
+     */
+    static testMethod void test_NullCurrencyDataValues() {
```
dvenor: I like this test.  Addresses the real issue.  I'd maybe change the name of this test to a more active name that includes a verb. For example `test_NullCurrencyDataIsNotCached`.  I know that naming pattern is not happening elsewhere in this class but it's the style we've been using lately in more recent code so I think new tests should adopt it.

src/classes/UTIL_CurrencyCache.cls
AndersonTarren: Daniel - after a review this morning, I'm going to lobby for the original solution of a simple null check.  

In this class, we have a couple calls to check the map keys like this : 
`if (!cachedMapOfRatesByCurrencyCode.containsKey(currCode))`

By not mapping null values, we are forcing that call to dig deeper, each check,  to see if currency data exists - which can lead to additional queries.  By mapping those nulls, we can tell future calls that we've already determined the value - so don't check again.

There's another method using this setup called getCurrDecimalPlaces, where another similar null check is already being made.

Thoughts?

dvenor: I did not notice that we have the same situation in `getCurrDecimalPlaces()` method and previously decided that caching a null CurrencyData is a valid scenario.  And you make a good argument for why:  Yes, the currency code may be invalid or just plain garbage, but at least we *know* that is the case, and don't have to recheck it.  I buy it.  Well done.  I'd still ditch the comment though.

dvenor:
**Lurch: attach W-30515

dvenor:
**lurch: attach W-030515

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQTOQA2" target="_blank">W-030515</a>

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fQTOQA2/view

rponti-sforg:
@dvenor Do we want to run another build on this PR so that the Feature Test - Namespaced tests pass?  I can still merge the PR, but wanted to check with you first about the Feature Test - Namespaced tests.

dvenor:
@rponti-sforg, yes that's a good idea. It looks like a transient error that won't be show up on the next build, but I'll kick it off again just to be sure.  It's definitely better when it's all green!

PR #4029, opened 2019-01-31
Feature/bge clone

kslewis:
# Critical Changes

# Changes

- Row action in table on Batch Gift Entry tab allows Users to copy setup configuration from an existing Batch to a new Batch. 

# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabHelper.js
```diff
@@ -87,16 +100,17 @@
 
     /**
      * @description: checks that user has all necessary permissions and then launches modal or displays error
+     * @param: sourceBatchId Id of Batch record to be copied - optional
      */
-    checkFieldPermissions: function(component) {
+    checkFieldPermissions: function(component, sourceBatchId) {
```
ptewson-sfdo: This is a suggested refactor for something that predates this work item, but to address the confusion of opening the wizard model by calling a function called "checkFieldPermissions", I wonder if we can pass in the openNewBatchWizard() as a callback to checkFieldPermissions - 

Called with: 
    checkFieldPermissions(component, this.openNewBatchWizard(component, sourceBatchId));

And the function def here: 
    checkFieldPermissions: function(component, onSuccess) {

Honestly not sure if this is worth it just to make it a bit clearer how hitting that action opens the Wizard after permissions check passes, but wanted to bring it up

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabHelper.js
```diff
@@ -87,16 +100,17 @@
...
         var action = component.get('c.checkFieldPermissions');
         action.setCallback(this, function (response) {
             var state = response.getState();
             if (state === 'SUCCESS') {
                 if (!component.get('v.modalOpen')) {
```
ptewson-sfdo: Followup on above comment about passing in callback, this if block would become:
   onSuccess()

And we'd want to move the if(modalOpen) bit into the openNewBatchWizard function I think

randi274: In my double click testing, the ifModalOpen bit had to go here or it wouldn't work (multiple modals could still be opened on a quick double click). It might be different if it's structured as a callback though?

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabHelper.js
randi274: We haven't used that as a pattern before, but I'd be interested to see if this works and still prevents a double click! If it doesn't, we could name things more explicitly "checkFieldPermissionsToOpenWizard" ?

ptewson-sfdo:
**lurch: attach W-028589

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fNyeAAE" target="_blank">W-028589</a>

smarovich:
@kslewis @ptewson-sfdo I updated the release notes for this PR. Please let me know if I should make any further changes. TIA!

PR #4030, opened 2019-01-31
Fix test failure in CRLP_RollupContact_TEST 

lmeerkatz:
This only fails on certain dates. Tricksy!

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupContact_TEST.cls
+++ b/src/classes/CRLP_RollupContact_TEST.cls
@@ -349,7 +349,8 @@ private class CRLP_RollupContact_TEST {
                 Type = 'New'
         ));
         if (tt == TestType.TestWithAlwaysRollupToPrimary) {
-            donationYears.add(orgDonationDate.year().format());
+            String donationYr = UTIL_String.removeNonNumericCharacters(orgDonationDate.year().format());
+            donationYears.add(donationYr);
         }
 
         // create 4 opps for con4 to test single result opp edge cases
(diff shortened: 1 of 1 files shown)


src/classes/CRLP_RollupContact_TEST.cls
```diff
@@ -348,6 +348,9 @@ private class CRLP_RollupContact_TEST {
                 npe01__Do_Not_Automatically_Create_Payment__c = true,
                 Type = 'New'
         ));
+        if (tt == TestType.TestWithAlwaysRollupToPrimary) {
+            donationYears.add(orgDonationDate.year().format());
```
force2b: This will need to use the same `UTIL_String.removeNonNumericCharacters()` method as used on line 318.

lmeerkatz: Why can't I use the actual CloseDate value? I thought this was what you were suggesting in chat as a replacement for removeNonNumeric ... ?

force2b: `orgDonationDate.year().format()` will return `2,019`.

lmeerkatz: Ah, okay. I wonder how the test was passing with that in there!

force2b: Line 318 uses the method to strip out the comma from the year before adding it to the `donationYears` set. Line 352 has to add the year in the same format.

PR #4031, opened 2019-01-31
Attempt to fix time-based test failures in CRLP_RollupAccount_TEST

lmeerkatz:
This is a speculative change that may fix some tests that were failing yesterday. Fortunately/unfortunately those tests are not failing today, which makes it difficult to know if this change is doing anything useful.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupAccount_TEST.cls
+++ b/src/classes/CRLP_RollupAccount_TEST.cls
@@ -67,7 +67,9 @@ private class CRLP_RollupAccount_TEST {
         String filterGroupId1 = CMT_UnitTestData_TEST.getNewRecordId();
         String filterGroupId2 = CMT_UnitTestData_TEST.getNewRecordId();
 
-        String dtValue = DateTime.now().addMonths(-100).addDays(-1).format('yyyy-MM-dd');
+        DateTime dt = DateTime.now().addMonths(-100).addDays(-1);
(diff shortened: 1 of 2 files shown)


src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -67,7 +67,9 @@ private class CRLP_RollupAccount_TEST {
         String filterGroupId1 = CMT_UnitTestData_TEST.getNewRecordId();
         String filterGroupId2 = CMT_UnitTestData_TEST.getNewRecordId();
 
-        String dtValue = DateTime.now().addMonths(-100).addDays(-1).format('yyyy-MM-dd');
+        DateTime dt = DateTime.now().addMonths(-100).addDays(-1);
+        String dtValue = dt.format('yyyy-MM-dd', UserInfo.getTimezone().toString());
+
```
bethbrains: This looks perfect! I just tested with a ton of debug/anon stuff and I believe this will successfully avoid future failures based on the date being off by one because of time zone. The only thing I'd suggest is adding a comment to remind future devs (future us) why we have to go through the gymnastics and can't use System.today() that already accounts for time zone. 
Something like... 
// Must use DateTime to get .format(), so we have to explicitly ensure it's in the same time zone as System.today() dates used elsewhere that are already in user's time zone

PR #4032, opened 2019-01-31
BGE - Infinite Scroll Datatable

randi274:
# Critical Changes

# Changes
- Performance improvement and infinite scroll support for the Batch Gift Entry lightning datatable
# Issues Closed

# New Metadata

# Deleted Metadata


src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -9,7 +9,6 @@
             if (state === 'SUCCESS') {
```
randi274: @ptewson-sfdo this is the file that I'd like to reformat to match the helpful grouping we did with the Config Wizard. If you see anything else that needs reformatting/reordering, feel free to call it out!

randi274:
Discussed with @ptewson-sfdo that this will require 2 reviews; the first to review the code changes for the PR, and the second to improve the layout and formatting of the BGE_DataImportBatchEntryHelper. This will hopefully make the WI changes easier to review before I reorder the whole helper!

randi274:
***lurch: attach W-029995*

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fPn8QAE" target="_blank">W-029995</a>

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -260,6 +287,40 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
         return columns;
     }
 
+    /*******************************************************************************************************
+    * @description gets updated amount total and row count for the batch
+    * @param batchId: ID of the NPSP_Data_Import_Batch__c
+    * @return DataImportModel updated model with total amount and number of rows
+    */
+    private static DataImportModel getUpdatedTotals(Id batchId) {
```
bethbrains: I find this method name confusing given that it returns an intact Model. `getUpdatedModel` maybe?

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -416,11 +467,32 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
+    * @return list of field API names
+    */
+    private static List<String> getActiveFieldNamesFromBatch(Id batchId) {
+
+        DataImportBatch__c batch = [SELECT Active_Fields__c FROM DataImportBatch__c WHERE Id = :batchId];
+        String activeFieldsJSON = batch.Active_Fields__c;
+        List<String> activeFieldNames = new List<String>();
+
+        if (activeFieldsJSON != null) {
+            List<BGE_ConfigurationWizard_CTRL.BGEField> activeFields = (List<BGE_ConfigurationWizard_CTRL.BGEField>)JSON.deserialize(activeFieldsJSON, List<BGE_ConfigurationWizard_CTRL.BGEField>.class);
```
bethbrains: split onto multiple lines

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -534,8 +608,8 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
                 this.errors = errors;
 
             }
-            catch (Exception e) {
-                throw new AuraHandledException(e.getMessage());
+            catch (Exception ex) {
+                throw new AuraHandledException(ex.getMessage());
```
bethbrains: 👍

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -322,72 +383,62 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
     }
 
     /*******************************************************************************************************
-    * @description runs BDI Dry Run against a list of DataImport__c Ids
-    * based on the processing settings of a given batch
-    * @return the full data set for the batch
+    * @description prepare the data import record for dry run by clearing certain fields or retaining them
+    * based on if user has already selected a match, or if they want to let matching run. Either way, we want to use Dry Run
+    * to check for potential errors.
+    * @return List<DataImport__c> because BDI expects a list
```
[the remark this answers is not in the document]
randi274: We could, but then it's just really easy to pass directly to the processDataImportRecords. 

Since I named this 'prepareDataImportForDryRun', I feel this is an acceptable weird thing, because I'd rather make it strange here than change that BDI method. 

That said, if we think this is really weird, then I'd make another BDI_DataImport_API method like 'processSingleDataImportRecord' and remove the list processing there. Thoughts?

src/classes/BGE_DataImportBatchEntry_CTRL.cls
randi274: Yes, this felt awkward to me too. What I *really* wanted to do was make a new model that just returns the totals + a single DI record, but wasn't sure if that was making the controller too bulky. But that's really what I'm looking for when I'm calling this, and just using the model as a way to pass those values back together. I need the totals + DI record for run dry run on a new record or edited record, and then just need the totals for when delete is called.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -37,6 +37,11 @@
 
 public with sharing class BGE_DataImportBatchEntry_CTRL {
 
+    /*******************************************************************************************************
+    * @description The number of additional Data Import Records to load on each infinite scroll query
+    */
+    private static Integer loadSize = 50;
```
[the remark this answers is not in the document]
randi274: Yeah, I felt that too. I think we originally said 200 records initially, then loading in batches of 50, but I liked using a single variable here. What about if we load in batches of 200?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntry.cmp
```diff
@@ -180,16 +181,18 @@
                 <!--End expected totals-->
 
                 <aura:if isTrue="{!notequals(v.data.length, 0)}">
-                    <div class="{!if(lessthan(v.data.length, 4), 'datatableContainer', '')+' slds-p-top_small'}">
+                    <div class="slds-p-top_small" style="height: 350px">
```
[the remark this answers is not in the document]
randi274: Let's discuss in review! (I just took this from the docs and ran with it, but do generally prefer classes.)

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -213,6 +267,60 @@
...
+    },
+
+    /**
+     * @description: finds the index of the row and replaces it with the updated row
+     * @param tableRow: flattened row for insert into the table
+     */
+    replaceTableRow: function(component, tableRow) {
+        let rows = component.get('v.data');
+        for (let i = 0; i < rows.length; i++) {
+            if (rows[i].Id === tableRow.Id) {
```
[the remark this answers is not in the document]
randi274: Yes, I really wanted a better way to do this too, and gave it some thought. The issue is that when a user makes a change, onCellChange is fired and only gives us the draftValues. This is the Id of the row and the field that has changed by API name. 

In contrast, when I delete something, I can match for the indexOf in the table, because handleRowAction passes me the entire row and I can get an exact match on it.

Maybe there's a way to have something running in JS at the same time the action is firing? I'm a little nervous about race conditions, but that would make things more efficient.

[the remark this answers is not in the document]
randi274: Did a little further testing in onCellChange and confirmed that's all we have to work with:

`console.log(JSON.parse(JSON.stringify(event.getParams())));
{draftValues: Array(1)}
   draftValues: Array(1)
      0: {npsp__Donation_Amount__c: "16", Id: "a0S5C000006ugREUAY"}
      length: 1
   __proto__: Array(0)
__proto__: Object
`

src/classes/BGE_DataImportBatchEntry_CTRL.cls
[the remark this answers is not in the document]
lmeerkatz: Have you tried 200? I recall that being slower than I’d like. I’d lean toward having a smaller initial load (50?) and incrementing by a larger number (100?) to avoid jumpiness, but play around with some sample data and see what feels good. :)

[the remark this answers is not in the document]
randi274: Sounds like a fair compromise, and I'm sure 200 ends up being pretty slow (though hopefully less so now that we're not counting all the records to get the totals!). I'll give 50/100 load a try!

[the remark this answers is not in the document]
randi274: I tried bumped them both up to 100 (initial load and infinite scroll load), and the wait is not visibly more significant than it was with 50 records. Let me know if you test and find otherwise, and I can still trying switching to 50/100 (and we can verify with Steph if that's ok too). 

(...Also it's late in the day and my brain was overcomplicating the 50/100 split, so this is what I've got for now 😅 )

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -84,46 +106,74 @@
...
+            let errorSize = 0;
+            let errorIds = Object.keys(tableErrors.rows);
+            errorIds.forEach(function(oldErrorId) {
+                if (oldErrorId !== rowId) {
+                    updatedErrors.rows[oldErrorId] = tableErrors.rows[oldErrorId];
+                    errorSize++;
+                }
+            });
+            updatedErrors.size = errorSize;
+            component.set('v.errors', updatedErrors);
```
ptewson-sfdo: Is updatedErrors just a duplicated list of errors except you aren't adding the error at rowId into it? Would it be possible to just remove the error at tableErrors.rows[rowId]?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -194,17 +244,21 @@
      * @description: starts the BDI dry run to verify DataImport__c matches
      * @param recordIds: list of DataImport__c RecordIds to check for dry run
      */
-    runDryRun: function(component, recordIds) {
+    runNewRecordDryRun: function(component, recordId) {
```
ptewson-sfdo: Is there a lot of overlap between this function and handleRowEdit()? Could they share a function and pass in the action (which is the main difference I think?) as a parameter?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
randi274: @ptewson-sfdo thanks for nudging me on this one, I was overthinking it. I needed to call delete on it, because rowId is an attribute on the row Object, and was having a tough time because I thought it was a function on the object!

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
ptewson-sfdo: What caught my eye is that the callbacks between runNewRecordDryRun and handleRowEdit are nearly identical except for the toast behavior so maybe there was a way to split that out into updateTableFromModel() or whatever name makes sense - but it's like, what six lines of code? I'm honestly not sure if the juice is worth the squeeze here.

randi274: @ptewson-sfdo I reworked this a bit to combine them, including moving the use of 'afterDryRun' to be the callback function used and renaming the old 'afterDryRun' to be more specific to what we want to happen with the datatable rows. What do you think?

ptewson-sfdo: I love it! Nice!

PR #4033, opened 2019-02-01
WIP: Move Contact Merge Future Cleanup to Queueable

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Store Session Record  Contact  ${winning_id}
+    Go To Object Home  Contact
+    Page Should Contain  Robot One
+    Page Should Contain  Robot Two
+    Go To Setup Home
+    Open App Launcher
+    Input Text  //input[@placeholder='Search apps or items...']  Contact Merge
+    Click Link  Contact Merge
+    Select Frame with ID  vfFrameId
+    Input Text  //input[@placeholder='Search Contacts']  Robot
```
davisagli: I had to add this before this line:

        Wait Until Page Contains Element  //input[@placeholder='Search Contacts']

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Click Element  //input[@value='Search']
+    Click Element  (//span[@id='fauxCBSelect'])[1]
+    Click Element  (//span[@id='fauxCBSelect'])[2]
+    Click Element  //input[@value='Next']
+    Click Button  name:Merge
```
davisagli: I had to change this one to:

        Click Element  (//button[text()='Merge'])[1]

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+Resource        tests/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Merge Contacts
+    # Manually creating contact records, sosl doesn't always return api generated contacts
+    Go To Object Home  Contact
+    Click Link  title:New
```
skristem: How about using an existing keyword Click Object Button       New

tests/browser/contacts_accounts/merge_contacts.robot
skristem: Use existing keyword 'Click Managehh Button__Merge'

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Input Text  //input[@placeholder='Search apps or items...']  Contact Merge
+    Click Link  Contact Merge
+    Select Frame with ID  vfFrameId
+    Input Text  //input[@placeholder='Search Contacts']  Robot
+    Click Element  //input[@value='Search']
+    Click Element  (//span[@id='fauxCBSelect'])[1]
+    Click Element  (//span[@id='fauxCBSelect'])[2]
+    Click Element  //input[@value='Next']
+    Click Button  name:Merge
+    Click Element  //input[@value='Merge']
```
skristem: Use 'Click Button__Merge'

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Go To Setup Home
+    Open App Launcher
+    Input Text  //input[@placeholder='Search apps or items...']  Contact Merge
+    Click Link  Contact Merge
+    Select Frame with ID  vfFrameId
+    Input Text  //input[@placeholder='Search Contacts']  Robot
+    Click Element  //input[@value='Search']
+    Click Element  (//span[@id='fauxCBSelect'])[1]
+    Click Element  (//span[@id='fauxCBSelect'])[2]
+    Click Element  //input[@value='Next']
```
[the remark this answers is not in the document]
skristem: This suggestion is optional. In case u want to keep the robot file clean, u can create a generic locator for checkbox and add it to locator_w19.py file and create a keyword for it in NPSP.py file. Here is the generic locator '//tbody/tr[./td[2]/a[contains(text(),"{}")]]/td//input[@type="checkbox"]'. To your keyword you would pass the contact name in order to select checkbox next to it. If you are not comfortable I can clean this up later.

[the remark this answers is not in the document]
davisagli: @skristem that locator doesn't look very generic; it looks very tied to the exact DOM structure of the page. This way is at least targeting by id, which is less likely to break. I agree it would be ideal to locate based on the actual contact name rather than position though.

tests/browser/contacts_accounts/merge_contacts.robot
davisagli: @skristem there are multiple buttons on the page with "Merge" as their text, so using the value attribute is necessary here to pick the right one.

tests/browser/contacts_accounts/merge_contacts.robot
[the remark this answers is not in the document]
skristem: @davisagli i agree but if there is another contact created by earlier robot test that had robot in the name and was not deleted, then there is a possibility of merging wrong contacts.

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Populate Form
+    ...                       First Name=Robot
+    ...                       Last Name=One
+    Click Modal Button        Save & New
+    Populate Form
+    ...                       First Name=Robot
+    ...                       Last Name=Two
+    Click Modal Button        Save
+    ${winning_id} =  Get Current Record Id
+    Store Session Record  Contact  ${winning_id}
```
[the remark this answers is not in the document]
jjbennett: I will try this again, I thought I remembered getting an error when it attempted to delete a record that was already deleted.

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Go To Object Home  Contact
+    Page Should Contain  Robot One
+    Page Should Contain  Robot Two
+    Go To Setup Home
```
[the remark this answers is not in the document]
jjbennett: I was having issues clicking the Contact Merge link, I kept getting an error if I clicked the App Launcher from anywhere besides the Setup Home page. I will try again and report back the error.

tests/browser/contacts_accounts/merge_contacts.robot
```diff
@@ -0,0 +1,44 @@
...
+    Open App Launcher
+    Input Text  //input[@placeholder='Search apps or items...']  Contact Merge
+    Click Link  Contact Merge
+    Select Frame with ID  vfFrameId
+    Input Text  //input[@placeholder='Search Contacts']  Robot
+    Click Element  //input[@value='Search']
```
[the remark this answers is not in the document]
jjbennett: I wasn't able to get Click Button to work since these are actually input elements and not button elements.

tests/browser/contacts_accounts/merge_contacts.robot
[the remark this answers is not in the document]
davisagli: It logs a message in that case but it's not actually a problem.

tests/browser/contacts_accounts/merge_contacts.robot
jjbennett: @davisagli - assuming you had to change it because the current version didn't work for you? It's working for me the way it is, so I'm curious what / why it wasn't working for you.

tests/browser/contacts_accounts/merge_contacts.robot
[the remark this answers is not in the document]
davisagli: Hmm, `Click Button` works with both `input` and `button` elements, so maybe it was a different problem.

tests/browser/contacts_accounts/merge_contacts.robot
davisagli: There isn't a button with `name="Merge"`. You can see the log of the failed MetaCI build here: https://mrbelvedereci.herokuapp.com/tests/result/32058172

jjbennett: Thanks, I just realized I never actually checked the robot tests in MetaCI!

jjbennett:
@davisagli @skristem - Thank you both for the feedback. I have incorporated both of your suggestions into my now passing robot test.
@davisagli - I noticed that when using the store session record that only the 2nd record is being removed when the merge fails. Is there a trick I am missing as to adding multiple record ids?

davisagli:
@jjbennett looks good. I think you just found a bug in the Delete Session Records keyword -- it's removing items from the same list it's iterating over, so ends up skipping every other one! I'll add a CumulusCI issue for that and it doesn't need to hold up this PR.

jjbennett:
Going a different route moving the robot tests to PR4049.

PR #4038, opened 2019-02-05
BGE - Required field error message fix

randi274:
- Fix error message in Config Wizard modal to only show name as a required field.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
@ptewson-sfdo this was approved and should've gone in with the earlier work on the record type picklist branch. It was unrelated to that change, but was a small fix we were hoping to include with it. Let me know how you'd like to handle this!

PR #4043, opened 2019-02-06
Feature/jbrandolini/rebrand advancement to gem

jbrandolinisf:
Updated label of npe01__Contacts_And_Orgs_Settings__c.Advancement Namespace field to GEM Namespace field
Updated code in setup() method of ADV_PackageInfo_SVC class to check for both legacy Advancement_Setting custom metadata (with Use_Advancement__c field) and GEM API Settings metadata with (Use_GEM__c field)

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


annejackson:
Should we assign this to a NPSP dev first before QE-based on comments above.... @lparrott and @jbrandolinisf ?

ptewson-sfdo:
Hi @jbrandolinisf - just for context, do I have this right - this work covers some changes to settings and configuration, where the config information is in the GEM_Setting metadata now instead of the Advancement_Setting? But because customer orgs could be using either one there is some logic in this code to check for both? 

Do we need to keep an eye out for any near- or far-future work to finalize deprecation of old settings, or is that covered already? 

Thanks! Just trying to stay up to speed!

jbrandolinisf:
Your understanding of the context is correct.  It will check for both the
original and updated metadata, and will work with either.

We have 3 PRs for this.  One is in HEDA, and has been merged.  There is the
work to refactor the new GEM config.  Once we have a final namespace, we
will update HEDA and NPSP again for that.  Then, this will be complete.

We will likely just rename the labels in the Cumulus space, rather than
deprecate the fields.

Let me know if you need anything else!

--
John Brandolini
Phone: 781-460-3479

jbrandolinisf:
@ptewson-sfdo - Do you need anything else on this, or can we get it merged?

PR #4044, opened 2019-02-08
BDI - Campaign Lookup Support

randi274:
# Critical Changes

# Changes
- Created the Donation Campaign Source field on the Data Import object in to order to specify by ID which Campaign to use as the Primary Campaign ID on an opportunity.
- In the case of Campaign Name matching, we preserve the Id of the campaign that was used as the Primary Campaign Source and keep a record of this ID on the Data Import record.
# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
**lurch: attach W-031061

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fRBfQAM" target="_blank">W-031061</a>

judisohn:
This will make many folks happy. Thank you.

src/classes/BDI_DataImportService.cls
```diff
@@ -288,7 +288,7 @@ global with sharing class BDI_DataImportService {
     //public Map<String, Contact> mapDIKeyToContact = new Map<String, Contact>();
     public Map<String, Account> mapDIKeyToAccount = new Map<String, Account>();
     private Map<Id, Address__c> mapDIIdToAddr = new Map<Id, Address__c>();
-    public Map<String, Id> mapStrNameToCampaignId = new Map<String, Id>();
+    public Map<String, Id> campaignNameToCampaignId = new Map<String, Id>();
```
lmeerkatz: 🙌

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,66 +1725,81 @@ global with sharing class BDI_DataImportService {
...
-                setStrCampaign.add(dataImport.Donation_Campaign_Name__c);
+                uniqueCampaignNames.add(dataImport.Donation_Campaign_Name__c);
             }
 
         }
 
         // add the existing campaigns to our map
-        for (Campaign cmp : [SELECT Id, Name FROM Campaign WHERE Name IN :setStrCampaign]) {
-            mapStrNameToCampaignId.put(cmp.Name, cmp.Id);
+        for (Campaign cmp : [SELECT Id, Name FROM Campaign WHERE Name IN :uniqueCampaignNames]) {
```
lmeerkatz: I know you didn't write the logic here, but I feel like we should at least add `ORDER BY CreatedDate DESC` to make the selection predictable (though flawed). I'd also consider moving the query out of the for loop for clarity.

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,66 +1725,81 @@ global with sharing class BDI_DataImportService {
...
+            campaignNameToCampaignId.put(cmp.Name, cmp.Id);
         }
 
         // create any new campaigns
-        List<Campaign> listCmpNew = new List<Campaign>();
-        List<DataImport__c> listDINew = new List<DataImport__c>();
-        setStrCampaign.clear();
+        List<Campaign> campaignsToInsert = new List<Campaign>();
+        List<DataImport__c> dataImportsWithNewCampaigns = new List<DataImport__c>();
+        uniqueCampaignNames.clear();
```
lmeerkatz: Do we need to clear this out?

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,66 +1725,81 @@ global with sharing class BDI_DataImportService {
...
+            if (Test.isRunningTest() && campaignName != null && campaignName.startsWith('FailTest')) {
                 cmp.Name = null;
             }
         }
-        //insert listCmpNew;
-        List<Database.UpsertResult> listUR = UTIL_DMLService.upsertRecords(listCmpNew, false);
-        for (Integer i = 0; i < listUR.size(); i++) {
-            Database.UpsertResult ur = listUR[i];
+
+        List<Database.UpsertResult> upsertResults = UTIL_DMLService.upsertRecords(campaignsToInsert, false);
```
lmeerkatz: When are we updating campaigns? Can this just be an insert?

src/classes/BDI_DataImportService.cls
```diff
@@ -1816,7 +1831,7 @@ global with sharing class BDI_DataImportService {
             // only create CMs here for rows that don't have an opportunity
             if (dataImport.Donation_Campaign_Name__c != null && dataImport.DonationImported__c == null) {
                 // assume errors would have already been thrown
-                Id idCmp = mapStrNameToCampaignId.get(dataImport.Donation_Campaign_Name__c);
+                Id idCmp = campaignNameToCampaignId.get(dataImport.Donation_Campaign_Name__c);
```
lmeerkatz: idCmp -> campaignId

src/classes/BDI_Donations.cls
```diff
@@ -506,6 +508,15 @@ public class BDI_Donations {
             } else if (dataImport.PaymentImportStatus__c == System.Label.bdiMatchedNone) {
                 dataImport.PaymentImportStatus__c = bdi.statusMatchedNone();
             }
+
+            // campaigns -- list is shorter since we're limiting matching behavior
```
lmeerkatz: Consider breaking the three sections of this method into three methods - `getCampaignImportStatus(String labelValue)`, etc.

lmeerkatz: Also, is this really the best way to do this? Set status THEN see if we're running dry run then change it?

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,66 +1725,81 @@ global with sharing class BDI_DataImportService {
...
-            if (strName != null && setStrCampaign.add(strName)) {
-                if (mapStrNameToCampaignId.get(strName) == null) {
-                    cmp = new Campaign(Name=strName, IsActive=true);
-                    listCmpNew.add(cmp);
-                    listDINew.add(dataImport);
+
+            if (campaignId != null) {
+                // User specified a Campaign lookup in BGE lookup or with an ID on import
+                dataImport.DonationCampaignImportStatus__c = System.Label.bdiMatchedId;
+            } else if (campaignName != null && uniqueCampaignNames.add(campaignName)) {
```
kslewis: Just curious - why the uniqueCampaignNames.add(campaignName) here?  I don't see any subsequent usage of the uniqueCampaignNames Set.

src/classes/BDI_DataImportService.cls
randi274: This previously was an insert, and you'll see I removed that commented out line in the diff. I'm not sure why this was changed, but it seemed to be made intentionally. I could dig into the previous diffs to see why this change may have been made and if it could just be an insert.

src/classes/BDI_DataImportService.cls
randi274: Great question, and I actually removed this at one point because it seemed unnecessary and it broke a test! Since uniqueCampaignNames is a set, it returns a boolean that lets us know if it could be added and wasn't in the list yet, or if it couldn't be added because we've already queued up a campaign with that same name. This stops us from making duplicate campaigns.

lmeerkatz: Would it be clearer to separate the addition from the check? 

```
else if (campaignName != null && !uniqueCampaignNames.contains(campaignName)) {
    uniqueCampaignNames.add(campaignName)
```

src/classes/BDI_DataImportService.cls
randi274: Yes, this is part of the logic the code relies on below. As we loop through each match we add them back to this list, and it prevents duplicate campaigns from being created. It could be refactored, but as-is this part is necessary.

src/classes/BDI_Donations.cls
randi274: That's a great question re: the latter part, and I think this written to avoid additional complexity during the writing of this code, or as a catch-all. I definitely would rather not rework the first two sections, but would be comfortable changing the campaigns section if you'd like to see that fixed here. Otherwise, I can break them out into separate methods if that's preferable.

lmeerkatz: Let's take the opportunity to improve code where we can.

lmeerkatz: I think this code is equivalent and avoids repeating the ifs:

```
Map<String, String> importStatusToDryRunStatus = new Map<String, String>{
	BDI_DataImport_API.bdiFailed => bdi.statusDIFailed(),
	System.Label.bdiMatched => bdi.statusMatchedExisting(),
	System.Label.bdiMatchedBest => bdi.statusMatchedBest(),
	System.Label.bdiMatchedNone => bdi.statusMatchedNone(),
	System.Label.bdiMatchedId => bdi.statusMatchedById()
}

private void setDryRunDonationStatuses(List<DataImport__c> dataImportDonations) {

    for (DataImport__c dataImport : dataImportDonations) {
        // opps
        dataImport.DonationImportStatus__c = importStatusToDryRunStatus.get(dataImport.DonationImportStatus__c);

        // payments
        dataImport.PaymentImportStatus__c = importStatusToDryRunStatus.get(dataImport.PaymentImportStatus__c);

        // campaigns -- list is shorter since we're limiting matching behavior
        dataImport.DonationCampaignImportStatus__c = importStatusToDryRunStatus.get(dataImport.DonationCampaignImportStatus__c);
    }
}
```

randi274: Love it! I'll test this out to be sure, and will clean this up. Great suggestion!

src/classes/BDI_DataImportService.cls
randi274: Yeah, I think so @lmeerkatz. A little more code, but definitely easier to read.

src/classes/BDI_Donations.cls
randi274: So I did quite a bit of testing, and for some reason putting importStatusToDryRunStatus outside of the setDryRunDonationStatuses as a private or public class level variable caused all kinds of test failures. I've wrapped it into this method in this next commit, and think it is still clearer to read than before. I also needed to factor in a null check, since we're not using else blocks in the code and need to be prepared for an import status not being in the map (there was exactly 1 test failure that helped catch this!).

lmeerkatz: Yay for helpful tests!

src/classes/BDI_DataImport_TEST.cls
```diff
@@ -1501,15 +1619,15 @@ public with sharing class BDI_DataImport_TEST {
     **********************************************************************************************************/
     private static List<DataImport__c> getDIs() {
```
kslewis: Just stylistic, but we might consider changing this to getDataImports() ?

src/classes/BDI_DataImportService.cls
lmeerkatz: @randi274 I'd still love to know what's going on here. At the very least, it would be good to change the variable name to reflect what's actually happening, but as I don't see a use case for updating campaigns I'm more curious about why we'd be doing an upsert at all.

bethbrains: I dug back in the git blame, and it looks like Upsert goes all the way back to the origins of the feature. I know this doesn't explain much, but just providing some historical context. Not sure why it was originally designed to upsert instead of insert. Maybe/likely because other objects are being necessarily upserted and this was copied?

src/classes/BDI_DataImport_TEST.cls
randi274: But if you give a mouse a cookie, the mouse will have to update all the listDI references too 😆 fair suggestion though, and the diff is quite a bit larger now!

src/objects/DataImport__c.object
```diff
@@ -718,6 +718,27 @@
         <trackTrending>false</trackTrending>
         <type>Currency</type>
     </fields>
+    <fields>
+        <fullName>DonationCampaignImported__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
```
lmeerkatz: Should we map to primary campaign source in help text here?

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,68 +1723,141 @@ global with sharing class BDI_DataImportService {
...
+            setCampaignIdIfMatchedByName(dataImport, campaignName);
+
+            // Create a new Campaign during processing if no match on Name
+            if (!isDryRun && dataImport.DonationCampaignImportStatus__c == System.Label.bdiMatchedNone) {
+
+                createNewCampaignIfUnique(campaignName, campaignNameToNewCampaign);
+                associateCampaignWithDataImport(dataImport, campaignName, newCampaignNameToDataImports);
+
             }
-            if (Test.isRunningTest() && strName != null && strName.startsWith('FailTest')) {
```
randi274: I looked into this and also a full BDI test without this, and we don't need it. We don't have any 'FailTest's hard coded in there, and there weren't any before. This was something in the previous block, but I think it was just added in here since it's in all the other import blocks.

src/classes/BDI_Donations.cls
```diff
@@ -479,32 +467,28 @@ public class BDI_Donations {
     */
     private void setDryRunDonationStatuses(List<DataImport__c> dataImportDonations) {
 
+        Map<String, String> importStatusToDryRunStatus = new Map<String, String>{
```
lmeerkatz: You could move this outside the method to avoid reconstructing it each time the method is called.

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,68 +1723,141 @@ global with sharing class BDI_DataImportService {
...
 
+    /*******************************************************************************************************
+     * @description Track duplicate campaign names so we only insert 1 campaign of the same name, but we associate
+     * that campaign with all the DIs that reference it
+     * @param dataImport the Data Import object to check
+     * @param campaignName name of the Campaign in the Donation_Campaign_Name__c field on the Data Import
+     * @param newCampaignNameToDataImports reference to the running list of campaign names and the Data Imports that reference them
+     * @return void
+     */
+    private void associateCampaignWithDataImport(DataImport__c dataImport, String campaignName, Map<String, List<DataImport__c>> newCampaignNameToDataImports) {
```
lmeerkatz: I'm inclined to make campaignNameToDataImports an instance variable instead of passing it in. What are the advantages of making it a parameter instead?

src/classes/BDI_DataImportService.cls
```diff
@@ -1725,68 +1723,141 @@ global with sharing class BDI_DataImportService {
...
+        String campaignName = dataImport.Donation_Campaign_Name__c;
+        String currentImportStatus = dataImport.DonationCampaignImportStatus__c; 
+        Id campaignId = dataImport.DonationCampaignImported__c;
+
+        // if we have a campaign Id, and it isn't populated from previous dry-run match, set the status
+        if (campaignId != null && currentImportStatus != System.Label.bdiDryRunMatched) {
+            dataImport.DonationCampaignImportStatus__c = System.Label.bdiMatchedId;
+        } else if (campaignName != null) {
+            Boolean campaignNameFound = campaignNameToCampaignId.containsKey(campaignName);
+            if (campaignNameFound) {
```
lmeerkatz: Nice! Very readable. :)

src/classes/BDI_Donations.cls
randi274: @lmeerkatz I did try this last time, and had some really surprising test failures! I'm not sure why, but it seems to be happiest in this method 😬

src/classes/BDI_DataImportService.cls
randi274: Yeah, I had a similar feeling. Mostly to avoid cluttering the class, but I could try moving it!

PR #4046, opened 2019-02-08
NPSP Callable Api to disable all TDTM Triggers

patrick-yan-sf:
Create CallableApi class that implements System.Callable to allow external packages to turn off TDTM triggers without direct references to NPSP.

# Critical Changes

# Changes

- Created a new Global Class called CallableApi that implements the System.Callable interface.  It currently supports the 'TDTM.disableAllTriggers' action.
- Added a new public method disableAllTriggers to the TDTM_Config_API class to allow users to disable all TDTM triggers for the current execution context.

# Issues Closed

# New Metadata
- CallableApi.cls
- CallableApi_TEST.cls
- CallableApiMethodNotImplemented Custom label

# Deleted Metadata


patrick-yan-sf:
*** lurch: Attach W-030998

patrick-yan-sf:
**lurch: Attach W-030998

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fR6aQAE" target="_blank">W-030998</a>

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description API class that exposes a callable class for users
+*/
+global class CallableApi implements System.Callable {
+
+    /*******************************************************************************************************
+    * @description Exception class for Malformed Method Invocations
+    ********************************************************************************************************/
+    global class MalformedMethodInvocationException extends Exception {}
```
force2b: I'm pretty sure this exception can be public. What I'm not 100% sure on is how that shows up in a try/catch block in the calling class. I'm positive it would work with a simple `catch (Exception ex)` block though. I think the `getMessage()` method would give the calling class the right message too.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+    * action handler based on the action text
+    * @param action string of the target action
+    * @param args name value map of arguments
+    * @return Object returned will depend on the results of the target function
+    ********************************************************************************************************/
+    global Object call(String action, Map<String, Object> args) {
+       
+        CallableApiInfo actionInfo = validateAction(action, args);
+        
+        if (actionInfo == null) {
```
force2b: I'm not sure there's a lot of value gained by validating the action and then calling another method with the switch statement. I think you can collapse/simplify all of that by just using a single switch statement in this method. The `when else` part would cover an invalid action. With that model you can eliminate the map (which also eliminates duplicating the action names), the wrapper class and the other two methods. I think it would be a lot easier to read as well.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+            throw new MalformedMethodInvocationException(action + ' is not valid');
+        }
+
+       return dispatchAction(actionInfo); 
+    }
+
+    /*******************************************************************************************************
+    * @description Helper class to keep track of the information needed for actions
+    ********************************************************************************************************/
+    public class CallableApiInfo {
```
force2b: This is something else that I don't think adds value. A simple switch statement in the `call` methods would handle separating the actions.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+
+        return generateResponse(actionInfo);
+    }
+
+    /*******************************************************************************************************
+    * @description Generate a generic response to the call for actions that do not have an explicit response
+    ********************************************************************************************************/
+    private Map<String, Object> generateResponse(CallableApiInfo actionInfo) {
```
force2b: This heavily limits the return options for our api. The beauty of the callable is that it can return absolutely anything - a List<SObject>, JSON String, Map, etc. This implementation removes all of that capability.

For errors/exceptions, our api or the delegated method should throw the appropriate exception so that it can be handled as needed by the calling package.class.method

src/classes/CallableApi_TEST.cls
```diff
@@ -0,0 +1,115 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Callable API 
+* @group-content ../../ApexDocContent/CallableAPI.htm
+* @description Tests specific to Callable API. 
+*/
+@IsTest
+
```
force2b: The blank line should be removed, and this class should just be defined as `private class CallableApi_TEST`.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description API class that exposes a callable class for users
+*/
+global class CallableApi implements System.Callable {
```
force2b: I recommend implementing this class as `inherited sharing`

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    /*******************************************************************************************************
+    * @description call function implementation of the callable Interface will dispatch to the appropriate
+    * action handler based on the action text
+    * @param action string of the target action
+    * @param args name value map of arguments
+    * @return Object returned will depend on the results of the target function
+    ********************************************************************************************************/
+    global Object call(String action, Map<String, Object> args) {
```
force2b: I know the docs use `args` in their sample code, but I think a better variable name is `params` or `actionParms`

src/classes/CallableApi_TEST.cls
```diff
@@ -0,0 +1,115 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT 
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Callable API 
+* @group-content ../../ApexDocContent/CallableAPI.htm
```
force2b: This should be `ApexDocContent/API.htm`. Plus that html file should be updated accordingly to reflect the new API class.

src/classes/TDTM_Config_API.cls
```diff
@@ -96,4 +96,17 @@ global class TDTM_Config_API {
         TDTM_ProcessControl.toggleTriggerState('npe01__OppPayment__c', 'CRLP_Rollup_TDTM', false);
     }
 
+
+    /*******************************************************************************************************
+    * @description Method to disable all TDTM triggers for the current code execution context
+    * @return void
+    ********************************************************************************************************/
+    global static void disableAllTriggers() {
```
force2b: I would leave this as a public method so that we're not locked into forever. We've exposed the ability to call the method already through our new callable api, so it's not necessary to have another entry point as a global method.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,144 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description API class that exposes a callable class for users
```
force2b: How about `Global api class that exposes a System.Callable interface for external applications and customers to leverage for future NPSP api's`.

src/classes/CallableApi_TEST.cls
```diff
@@ -0,0 +1,115 @@
...
+
+public with sharing class CallableApi_TEST {
+
+    private static final String TDTM_TRIGGER_ACTION = 'TDTM.disableAllTriggers';
+    
+    /*******************************************************************************************************
+    * @description verify the TDTM Callable action returns an Object
+    ********************************************************************************************************/
+    @IsTest
+    public static void verifyCallableReturnsAnObject() {
```
force2b: There should not be a `public` declaration on test methods.

src/classes/CallableApi_TEST.cls
```diff
@@ -0,0 +1,115 @@
...
+
+        Callable npspApi = (Callable) Type.forName('CallableApi').newInstance();
+        Map<String, Object> triggerArgs = new Map<String, Object>();
+        Boolean methodExceptionThrown = false;
+
+        try {
+            Object result = npspApi.call('InvalidFormat', triggerArgs);
+
+        } catch (CallableApi.MalformedMethodInvocationException apiException) {
+            methodExceptionThrown = true;
```
force2b: A simpler alternative to using a var to flag that the exception is thrown is to add the following in the `try` block above:
```System.assert(false, 'The exception was not thrown');```

File: src/classes/TDTM_Config_API.cls [hunk 13]
```diff
@@ -96,4 +96,17 @@ global class TDTM_Config_API {
...
+
+        TDTM_TriggerHandler.disableTDTM = true;
```
force2b: Good idea to set this flag as well.

patrick-yan-sf:
@force2b thanks for the suggestions.  Overall the code is much simpler now.  

I changed the TDTM_Config_API call to return a Boolean as opposed to void - and thus the callableApi will just pass that along as the return object.

ApexDocContent/API.htm
```diff
@@ -12,7 +12,8 @@ <h2>Nonprofit Success Pack API (Application Programming Interface)</h2>
 	
 		<p>The classes involved with the public interface include:</p>
 		<ul>
-		    <li>BDI_DataImport_API</li>
+			<li>BDI_DataImport_API</li>
+			<li>CallableApi</li>
 			<li>ERR_Handler_API</li>
```
patrick-yan-sf: @sam-knox - I simply added the new class to this HTML page - is there more info that we need to do - with regards to informing the users of the new API?

force2b: FYI: We still keep these files up to date, but technically they're only used by the currently non-functional ApexDoc system. The hope is that one day ApexDocs will start generating docs for our app again.

sam-knox: @patrick-yan-sf - That should be sufficient. As Mike says, ApexDoc is defunct so the repo itself is the best place for people to look.

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,65 @@
...
+    * @description call function implementation of the callable Interface will dispatch to the appropriate
+    * action handler based on the action text
+    * @param action string of the target action
+    * @param args name value map of arguments
+    * @return Object returned will depend on the results of the target function
+    ********************************************************************************************************/
+    global Object call(String action, Map<String, Object> params) {
+       
+        switch on (action) {
+            when 'TDTM.disableAllTriggers' {
```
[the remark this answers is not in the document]
patrick-yan-sf: @force2b Yes switch is case sensitive.  I've changed it to be lower case so that the calling user shouldn't have to worry about the case of the text.  

I've made the other changes you've suggested as well.  Thanks

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,65 @@
...
+    * action handler based on the action text
+    * @param action string of the target action
+    * @param args name value map of arguments
+    * @return Object returned will depend on the results of the target function
+    ********************************************************************************************************/
+    global Object call(String action, Map<String, Object> params) {
+       
+        switch on (action.toLowerCase()) {
+            when 'tdtm.disablealltriggers' {
+                return TDTM_Config_API.disableAllTriggers();
```
force2b: One more change. Rather than force an arbitrary return value from all delegated methods, I recommend that 'TDTM_Config_API.disableAllTriggers()' be defined as `void`. The end of this method can then return `true` or `'success'` as a standardized response when the delegated method does not return actual data.

force2b:
@patrick-yan-sf One more little change request. In the PR description, can you use bullets in the Changes and New Metadata sections? That'll ensure these get formatted properly in the release notes.

src/classes/CallableApi.cls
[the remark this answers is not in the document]
force2b: @patrick-yan-sf Making this lower case doesn't actually fix the issue of the case. If the callable is passed a value of "TDTM.Disablealltriggers" that would still fail. To be fully case insensitive this method would need to cast the action value to lower case first:
`action = (action.contains('.') ? action.split('\\.')[0] + '.' + action.split('\\.')[1].toLowerCase() : action);`

src/classes/CallableApi_TEST.cls
```diff
@@ -0,0 +1,83 @@
...
+
+        Test.stopTest();
+
+        Object callResult = result;
+        System.assertNotEquals(null, callResult, 'No result was returned from call');
+    }
+
+
+    /*******************************************************************************************************
+    * @description Call the API with an invalid method format and make sure an exception is thrown
```
force2b: @patrick-yan-sf  I think this test method is actually testing that the action is not valid rather than it doesn't match a specific format.

patrick-yan-sf:
@force2b 

I've made the changes suggested but for the lowercase one.  I think it is working.  I'll add a comment on the actual line and maybe you can see if I'm missing something.  Thanks

src/classes/CallableApi.cls
```diff
@@ -47,19 +47,22 @@ global with sharing class CallableApi implements System.Callable {
...
     * @param args name value map of arguments
-    * @return Object returned will depend on the results of the target function
+    * @return Object returned will depend on the results of the target function otherwise it returns true
     ********************************************************************************************************/
     global Object call(String action, Map<String, Object> params) {
        
         switch on (action.toLowerCase()) {
             when 'tdtm.disablealltriggers' {
-                return TDTM_Config_API.disableAllTriggers();
+                TDTM_Config_API.disableAllTriggers();
```
patrick-yan-sf: @force2b 
If action = 'TDTM.DisableAllTriggers' then action.toLowerCase() should convert it to 'tdtm.disablealltriggers' .  Unless I'm missing something?

a null action would throw an error but we are expecting an action

force2b: That can work to. I was just trying to separate the two parts of the action name (UPPERCASE.lowercase). Doesn't really matter though. This meets the need perfectly.

Do you mean a null response would throw an exception?

patrick-yan-sf: yeah because action.toLowerCase() will throw an error if action is null right?  Should we check for that before doing the toLowerCase()?

force2b: Ah. The inline reassignment example I had from earlier could catch that, or we could just let the null value throw the npe exception. I like the idea of us throwing our custom exception ("(null) is not implemented", or something along those lines).

patrick-yan-sf: Added a check for the text and formatted the exception message to put in the ( parentheses )

force2b: Thanks @patrick-yan-sf. The parenthesis was only for the null value. It would be better as just wrapping action in single quotes, so a null or empty string would thrown an exception of `'' is not implemented`.

src/classes/TDTM_Config_API_Test.cls
```diff
@@ -92,5 +92,31 @@ private class TDTM_Config_API_Test {
...
+    private static void triggersAreDisabledForCurrentContext() {
+        // skip the test if Advancement is installed
+        if (ADV_PackageInfo_SVC.useAdv()) {
+            return;
+        }
+
+        // Get list of TDTM modified dates
+        Integer accountCountBefore = Database.countQuery('SELECT COUNT() FROM Account');
+
+        TDTM_Config_Api.disableAllTriggers();
```
force2b: @patrick-yan-sf One more question. Should this method call be `CallableApi.run('TDTM.DisableAllTriggers', null)` so that the test can validate the entire process?

patrick-yan-sf: I'm not sure.   It makes sense that we want to run the entire process.  But putting CallableApi in to the TDTM_Config_Api_Test would mean the TDTM class references the CallableApi class.  And the CallableApi class already references the TDTM class.  So there would be a circular type of reference.  I think we talked about this yesterday.  We might want to look at an API test suite?  Something that just test the global APIs that we provide.

patrick-yan-sf: Actually this is just the TEST class that references the CallableApi class.  So maybe it's ok.  I can add it if you think that is fine.

force2b: This is the test method that currently calls TDTM_Config_Api directly. We could either change this method to call the CallableApi class, or change the `CallableApi_Test` class to validate that the triggers are disabled by calling the run method with `TDTM.DisableAllTriggers`. Writing this out, I think it makes sense to add this validation logic to the `CallableApi_Test` class rather than this class.

patrick-yan-sf: I've added the check on CallableApi_Test to make sure the triggers were indeed disabled

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,69 @@
...
+    ********************************************************************************************************/
+    global Object call(String action, Map<String, Object> params) {
+        String cleanActionText = (String.isBlank(action) ? '' : action.toLowerCase());
+
+        switch on (cleanActionText) {
+            when 'tdtm.disablealltriggers' {
+                TDTM_Config_API.disableAllTriggers();
+            }
+            when else {
+                throw new MalformedMethodInvocationException('\'' + action + '\' is not implemented');
```
force2b: @sam-knox This is a new error message that would be thrown in Apex if a non-supported api is called. Since it's not really something an End User would see, I was thinking it's fine to have this as a hard coded (English) string. What do you think?

patrick-yan-sf:
@rponti-sforg This is ready for testing

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,69 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description Global API class that exposes a System.Callable interface for external
+* applications and customers to leverage for future NPSP api's
+*/
+global with sharing class CallableApi implements System.Callable {
```
lmeerkatz: Can we cap API? (CallableAPI instead of CallableApi)

src/classes/CallableApi.cls
```diff
@@ -0,0 +1,71 @@
...
+        String cleanActionText = (String.isBlank(action) ? '' : action.toLowerCase());
+
+        switch on (cleanActionText) {
+            when 'tdtm.disablealltriggers' {
+                TDTM_Config_API.disableAllTriggers();
+            }
+            when else {
+                throw new MalformedMethodInvocationException(
+                    String.format(System.Label.CallableApiMethodNotImplemented, new List<String> { action }) 
+                );
```
patrick-yan-sf: @force2b  Can you review the change to use a custom label for the exception text?  
I had asked Rick to hold off on QA testing until this got checked again.

src/labels/CustomLabels.labels
```diff
@@ -2971,6 +2971,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>campaignMemberStatusResponded</shortDescription>
         <value>campaignMemberStatusResponded</value>
     </labels>
+    <labels>
+        <fullName>CallableApiMethodNotImplemented</fullName>
+        <categories>API</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>CallableApi Method Not Implemented</shortDescription>
+        <value>''{0}'' is not implemented</value>
```
force2b: @patrick-yan-sf  The double single quotes are not needed here. It could also be added as `&apos;{0}&apos;` to escape the single quotes

src/classes/CallableApi.cls
force2b: Reviewed below.

patrick-yan-sf:
@force2b thanks - if custom label is ok then I'll ask Rick to test now

ApexDocContent/API.htm
patrick-yan-sf: @force2b - I think we forgot to change the API doc page with the updated Callable_API name.

rponti-sforg:
@patrick-yan-sf Do you want to re-run a build on this before I merge, or do you want me to merge?

force2b:
@rponti-sforg The Robot test can be ignored

PR #4047, opened 2019-02-11
BGE Wizard - Defaults - Date Picker not showing for Donation Date

mdeleon0:
Using the old getLightningDataType function from the controller instead of the ones in the utils class.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BGE_ConfigurationWizard_CTRL.cls
+++ b/src/classes/BGE_ConfigurationWizard_CTRL.cls
@@ -36,10 +36,6 @@
 */
 public with sharing class BGE_ConfigurationWizard_CTRL {
 
-    private static String LIGHTNING_FIELD_TYPE_NUMBER = 'number';
-    private static String LIGHTNING_FIELD_TYPE_TEXT = 'text';
-    private static String LIGHTNING_FIELD_TYPE_CHECKBOX = 'checkbox';
-
     /*******************************************************************************************************
     * @description These fields are required only if at least 1 field from its object group is selected
     */
@@ -152,7 +148,7 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
             }
 
             String rawFieldType = dfr.getType().name().toLowerCase();
-            field.type = getLightningFieldType(rawFieldType);
+            field.type = UTIL_LightningComponentSupport.getLightningFieldType(rawFieldType);
             if (rawFieldType == 'currency') {
                 field.formatter = 'currency';
             }
@@ -237,20 +233,6 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         return dfrs;
     }
 
-/*******************************************************************************************************
-    * @description Returns the corresponding Lightning field type from an Apex field type
-    */
-    public static String getLightningFieldType(String apexFieldType) {
-        if (apexFieldType == Schema.DisplayType.Currency) {
-            return LIGHTNING_FIELD_TYPE_NUMBER;
-        } else if (apexFieldType == Schema.DisplayType.String) {
-            return LIGHTNING_FIELD_TYPE_TEXT;
-        } else if (apexFieldType == Schema.DisplayType.Boolean) {
-            return LIGHTNING_FIELD_TYPE_CHECKBOX;
-        } else {
-            return apexFieldType;
-        }
-    }
 
     /*******************************************************************************************************
     * @description Retrieves the default active fields for a new Batch or Template in a JSON String
@@ -274,7 +256,7 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         donationAmount.isActive = true;
         donationAmount.requiredInEntryForm = true;
         donationAmount.hide = false;
-        donationAmount.type = getLightningFieldType('number');
+        donationAmount.type = UTIL_LightningComponentSupport.getLightningFieldType('number');
         donationAmount.formatter = 'currency';
         donationAmount.sortOrder = 0;
         donationAmount.alwaysRequired = true;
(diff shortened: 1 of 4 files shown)


src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -233,6 +233,20 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         return dfrs;
     }
 
+/*******************************************************************************************************
+    * @description Returns the corresponding Lightning field type from an Apex field type
+    */
+    public static String getLightningFieldType(String apexFieldType) {
+        if (apexFieldType == 'currency') {
```
ptewson-sfdo: Can we use "Schema.DisplayType.Currency" (or whatever the correct system type constant) is here instead of the hard-coded string? And actually probably pass in the DisplayType directly instead of the lowercased string.

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -233,6 +233,20 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         return dfrs;
     }
 
+/*******************************************************************************************************
+    * @description Returns the corresponding Lightning field type from an Apex field type
+    */
+    public static String getLightningFieldType(String apexFieldType) {
```
ptewson-sfdo: If this method also supports lightning components, should we move it into UTIL_LightningComponetSupport alongside getLightningDataType()?

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -233,6 +233,20 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
...
+        if (apexFieldType == 'currency') {
+            return 'number';
+        } else if (apexFieldType == 'string') {
+            return 'text';
+        } else if (apexFieldType == 'boolean') {
+            return 'checkbox';
```
ptewson-sfdo: Are there defined constants for the lightning field types to avoid hardcoding those strings as well?

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -233,6 +233,20 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         return dfrs;
     }
 
+/*******************************************************************************************************
+    * @description Returns the corresponding Lightning field type from an Apex field type
+    */
+    public static String getLightningFieldType(String apexFieldType) {
+        if (apexFieldType == 'currency') {
+            return 'number';
+        } else if (apexFieldType == 'string') {
```
ptewson-sfdo: Do we need to convert more than currency?

src/classes/BGE_ConfigurationWizard_CTRL.cls
mdeleon0: I haven't found any but I can create them.

src/classes/UTIL_LightningComponentSupport.cls
```diff
@@ -36,23 +36,43 @@
 
 public with sharing class UTIL_LightningComponentSupport {
 
+    public static final String LIGHTNING_FIELD_TYPE_NUMBER = 'number';
```
randi274: Let's rename these to be a little more clear about these being a data type, eg: "LIGHTNING_DATA_TYPE_NUMBER"

src/classes/UTIL_LightningComponentSupport.cls
```diff
@@ -36,23 +36,43 @@
...
-        rawType = rawType.toLowerCase();
-        if (rawType == 'datetime') {
-            return 'date';
-        } else if (rawType == 'date') {
-            return 'date-local';
-        } else if (rawType == 'double') {
-            return 'number';
-        } else if (rawType == 'string') {
-            return 'text';
+    public static String getLightningDataType(Schema.DisplayType fieldDisplayType) {
```
randi274: Could we also rename this to getLightningDataTableDataType, so we're super clear what this is in the future? We should also update the param in the comment above to reflect the Schema.DisplayType, and the description to reflect only lightning data table.

src/classes/UTIL_LightningComponentSupport.cls
```diff
@@ -36,23 +36,43 @@
...
+            return LIGHTNING_FIELD_TYPE_TEXT;
+        } else {
+            return fieldDisplayType.name().toLowerCase();
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description Returns the corresponding Lightning field type from an Apex field type
+    */
+    public static String getLightningFieldType(Schema.DisplayType fieldDisplayType) {
```
randi274: Same comment here, could we go with getLightningInputDataType and add the same param? Thanks!

lmeerkatz:
**lurch:attach W-031192

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fRMEQA2" target="_blank">W-031192</a>

src/classes/UTIL_LightningComponentSupport.cls
```diff
@@ -36,23 +36,45 @@
...
 
+    public static final String LIGHTNING_DATA_TYPE_NUMBER = 'number';
+    public static final String LIGHTNING_DATA_TYPE_TEXT = 'text';
+    public static final String LIGHTNING_DATA_TYPE_CHECKBOX = 'checkbox';
+    public static final String LIGHTNING_DATA_TYPE_DATE = 'date';
+    public static final String LIGHTNING_DATA_TYPE_DATE_LOCAL = 'date-local';
+
+    /*******************************************************************************************************
+    * @description Converts Apex field data type to type expected by lightning:datatable.
+    * @param rawType: the Schema.DisplayType from DescribeFieldResult
```
randi274: Ok last change -- can you rename the param to fieldDisplayType? Then we're all good!

PR #4048, opened 2019-02-12
CRLP: Incremental Rollups for Account/Contact Hard Credit Non-Skew

kslewis:
- To help with the performance of nightly Customizable Rollups batch jobs for our larger customers, we're introducing custom settings to enable incremental processing of Account and Contact Hard Credit Rollups (Non-Skew).  The new settings to enable incremental support are disabled by default and, for now, can be enabled by editing the Customizable Rollup Settings under Custom Settings in Salesforce Setup. The incremental mode option fields will be added to NPSP Settings in an upcoming release.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -323,4 +322,84 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+            return false;
+        }
+    }
+
+    /**
+     * @description Produces the where clause that should be used for the inner join when in incremental mode
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    private String getIncrementalModeWhereClause() {
+        String dateFieldNameForTesting = Test.isRunningTest() ? 'CloseDate' : 'SystemModstamp'; //Set to CloseDate for dev, will be set to SystemModstamp for testing/prod
```
kslewis: I'll need some help figuring out what to do here - I was thinking we might be able to test this better if we use an alternate date field for testing that would actually yield different results, since the SysModstamp will not actually filter down the Accounts in Test context.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -76,18 +76,33 @@ public abstract class CRLP_Batch_Base {
...
 
+    /** @description Whether this class was constructed from within a scheduled job */
+    protected Boolean isScheduledJob = false;
+
+    /**
+     * @description Testing API to mock isScheduledJob value when running tests
+     * @param isScheduledJob Boolean value to which isScheduledJob will be set if called
+     * from within a running test
+     */
+    @TestVisible protected void setIsScheduledJob(Boolean isScheduledJob){
```
lmeerkatz: Since this is only expected to be visible in tests, we could use `private` here instead of `protected`. The @TestVisible annotation should go on its own line.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -323,4 +322,84 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+        Integer mDays = this.getMaxDaysAgo();
```
lmeerkatz: The `m` in `mDays` and `mYears` could be clearer. Replace with `maxDaysAgo` or something along those lines?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +359,45 @@ public abstract class CRLP_Batch_Base {
...
 
+    /**
+     * @description returns the max Integer for all active time-bound Days_Back rollups,
+     * or null if there are no active time-bound Days_Back rollups for this summary object
+     */
+    private Integer maxDaysAgo = null;
+    protected Integer getMaxDaysAgo(){
+        if (maxDaysAgo == null) {
+            for (Rollup__mdt rollup : statefulCacheOfRollupsToBeProcessed) {
+                if (rollup.Active__c == true && rollup.Integer__c != null) {
```
ptewson-sfdo: Is it worthy mooshing these three ifs into one? That's a big gnarly bunch of logic so not honestly sure if that's cleaner

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +359,45 @@ public abstract class CRLP_Batch_Base {
...
+
+    /**
+     * @description returns the max Integer for all active time-bound Years_Ago rollups,
+     * or null if there are no active time-bound Years_Ago rollups for this summary object
+     */
+    private Integer maxYearsAgo = null;
+    protected Integer getMaxYearsAgo() {
+        if (maxYearsAgo == null) {
+            for (Rollup__mdt rollup : statefulCacheOfRollupsToBeProcessed) {
+                if (rollup.Active__c == true && rollup.Integer__c != null) {
```
ptewson-sfdo: Similar question about combining if's

src/classes/CRLP_Batch_Base_NonSkew.cls
lmeerkatz: I'm wondering if we should be able to pass in a date to facilitate manual testing. Either way, I'd change `dateFieldNameForTesting` to `dateFieldName` since the variable is used in tests and in production.

lmeerkatz:
@kslewis Should have mentioned earlier - This code is gorgeous! Very readable. Thank you!

src/classes/CRLP_Batch_Base.cls
lmeerkatz: We're checking for maxDaysAgo == null on 368 and 373, so that could help? Generally I'd try to break complex logic like this into a separate method.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -120,10 +120,9 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
                 // Only use the related records filter if not disabled in settings and not running unit tests
                 if (customizableRollupSettings.Disable_Related_Records_Filter__c == false && !Test.isRunningTest()) {
```
kslewis: @force2b Michael, can you provide any info as to why we are skipping the block below when in Test context?

src/classes/CRLP_Batch_Base.cls
kslewis: Ok, I'm going to put @TestVisible on its own line and switch to getter/setter format.  I tried to switch to Private but it actually needs to stay Protected (getter) since it is referenced from the extending classes (I'll make the setter explicitly private).

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -323,4 +322,84 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+        return standardWhereClause;
+    }
+
+    /**
+     * @description Produces the where clause that should be used for the inner join
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    @TestVisible private String getWhereClause() {
+        Boolean incrementalModeActivatedForThisJobType = isIncrementalMode(jobType);
+        Boolean hasActiveIncrementalModeRollups = (maxDaysAgo != null || maxYearsAgo != null);
```
force2b: I think this is more subtle in what we can support for this. For example, recurring donations do not have a native rollup that uses Days or Years, so in that case both of these would be null. However, that doesn't necessarily mean that an incremental rollup wouldn't be beneficial. I recommend that if both fields are null, the range default to 1 year (365 days). This way something like Recurring Donation rollups would see improved performance with its rollups.

NOTE: Since RD's aren't covered in this iteration, how about if the code make include a TODO comment to reflect the fact that we want to modify this to support rollups where there is no time period.

src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -527,4 +532,162 @@ private class CRLP_RollupAccount_TEST {
         }
 
     }
+
+    @isTest
```
force2b: I recommend moving the tests that are not actually running the full rollup jobs into a new class, maybe `CRLP_Batch_Base_TEST`. I think that test class could also use the IsParallel=true annotation.

lmeerkatz:
Thanks for your review, @force2b!

lmeerkatz:
**lurch:attach W-030929

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fR11QAE" target="_blank">W-030929</a>

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -323,4 +322,84 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    private String getIncrementalModeWhereClause() {
+        String whereClauseDateField = Test.isRunningTest() ? 'CloseDate' : 'SystemModstamp'; // Set to CloseDate
+        // or another Date field for testing.
+        // Will be set to SystemModstamp for prod.
+
+        String whereClause = 'Id IN (' +
+                'SELECT ' + keyField + ' ' +
+                'FROM ' + detailObjectType + ' ALLROWS ' +
```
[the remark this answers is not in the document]
kslewis: Thats what I thought also!  But it only compiles and runs this way when using ALLROWS within the join query

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -323,4 +322,84 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
+                'WHERE ';
+
+        if (maxDaysAgo != null) {
+            whereClause += whereClauseDateField + ' >= LAST_N_DAYS:' + maxDaysAgo;
+        }
+        if (maxYearsAgo != null) {
```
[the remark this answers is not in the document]
kslewis: Ok done!

src/classes/CRLP_RollupAccount_TEST.cls
```diff
@@ -56,7 +56,7 @@ private class CRLP_RollupAccount_TEST {
     /**
      * @description Create Rollup__mdt records to test rolling up from the Opp and Payment objects to the Account
      */
-    private static void mockRollupCMTValues() {
+    public static void mockRollupCMTValues() {
```
kslewis: I had to change this to public in order to access from the new batch base test class

src/classes/CRLP_RollupContact_TEST.cls
```diff
@@ -238,7 +238,8 @@ private class CRLP_RollupContact_TEST {
                 Rollups_Contact_SkewMode_Batch_Size__c = 200,
                 Rollups_Contact_Soft_Credit_Batch_Size__c = 200,
                 Rollups_Account_Batch_Size__c = 20,
-                Rollups_Contact_Batch_Size__c = 20
+                Rollups_Contact_Batch_Size__c = 20,
+                Disable_Related_Records_Filter__c = tt == TestType.TestBatch ? true : false
```
kslewis: Need to specifically set Disable Related Records Filter to true in the TestBatch batch type in order for the code block on line 499 of this class to pass

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,332 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Customizable Rollups Operations Services
+* @description Unit Tests for the Opportunity to Account and Contact Rollups
+*/
+@IsTest (isParallel=true)
+private class CRLP_Batch_Base_TEST {
+
+    @isTest
+    private static void testAccountHardCreditGetWhereClause_Standard(){
```
force2b: How about `standardWhereClauseShouldIncludeInnerJoin`?

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,332 @@
...
+        CRLP_Account_BATCH accountBatchJob = new CRLP_Account_BATCH(new List<Id>());
+        accountBatchJob.initializeJobData();
+
+        String whereClause = accountBatchJob.getWhereClause();
+        System.assert(whereClause.contains('WHERE') == false, 'The standard where clause for the inner join ' +
+                'should NOT contain a where filter on the Opportunities.');
+    }
+
+    @isTest
+    private static void testAccountHardCreditGetWhereClause_IncrementalMode(){
```
force2b: How about `standardWhereClauseShouldIncludeInnerJoinIncremental`?

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,332 @@
...
+        ));
+
+        // Create the test data
+        CRLP_RollupAccount_TEST.mockRollupCMTValues();
+
+        CRLP_Account_BATCH accountBatchJob = new CRLP_Account_BATCH(new List<Id>());
+        accountBatchJob.initializeJobData();
+
+        String whereClause = accountBatchJob.getWhereClause();
+        System.assert(whereClause.contains('WHERE') == false, 'The standard where clause for the inner join ' +
```
force2b: `.contains('(SELECT Id FROM')`

force2b:
@smarovich Would you mind reviewing the PR comment for this? I don't think we need to list the new custom settings fields under new metadata since it's mostly for internal use.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * @description Produces the where clause that should be used for the inner join when in incremental mode
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    protected String getIncrementalModeWhereClause() {
```
force2b: This can be private since it's not used outside this class. I also recommend renaming this to `buildIncrementalModeWhereClause`.

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,378 @@
...
+     * @return The configured Customizable_Rollup_Settings__c record
+     */
+    private static Customizable_Rollup_Settings__c configureIncrementalMode(CRLP_RollupProcessingOptions.RollupType jobType, Boolean isIncrementalMode) {
+        Customizable_Rollup_Settings__c settings;
+        if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit) {
+            settings = UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
+                    Customizable_Rollups_Enabled__c = true,
+                    AccountHardCreditNonSkew_Incremental__c = isIncrementalMode
+            ));
+        } else if (jobType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit) {
```
force2b: It's hard to see this correctly in github review, but can you add a blank line befe and after the `else if` blocks in all of these methods to break up the lines a bit. It'll make the methods easier to read.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+        set;
+    }
+
+    /**
+     * @description Produces the where clause that should be used for the inner join when NOT in incremental mode
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    protected String getStandardWhereClause() {
+        String standardWhereClause = null; //will return null if this method is called from an unsupported rollup type
+        if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit
```
force2b: It's hard to see this correctly in github review, but can you add a blank line befe and after the `else if` blocks in all of these methods to break up the lines a bit. It'll make the methods easier to read.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+        get {
+            return userDefinedLastNDaysField != null && userDefinedLastNDays != null;
+        }
+    }
+
+    /**
+     * @description Value to use with LAST_N_DAYS: in the inner join query
+     * when the org has the user-defined inner join for incremental rollups
+     */
+    private Integer userDefinedLastNDays {
```
force2b: How about `overrideForLastNDaysValue`?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+            return userDefinedLastNDays;
+        }
+        set;
+    }
+
+    /**
+     * @description API name of the date field to use as in the inner join query
+     * when the org has the user-defined inner join for incremental rollups
+     */
+    private String userDefinedLastNDaysField {
```
force2b: How about `overrideForLastNDaysField`?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+            }
+        }
+        return whereClause += ')';
+    }
+
+    /**
+     * @description Returns true if the org has configured a
+     * user-defined inner join for incremental rollups
+     */
+    private Boolean hasUserDefinedInnerJoin {
```
force2b: How about `hasIncrementalFieldOverrides`?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -344,4 +353,195 @@ public abstract class CRLP_Batch_Base {
...
+            standardWhereClause = 'Id IN (SELECT npe03__Recurring_Donation__c FROM Opportunity)';
+        }
+        return standardWhereClause;
+    }
+
+    /**
+     * @description Produces the where clause that should be used for the inner join
+     * @return The WHERE clause, as a string, that should be used to filter the detail records
+     */
+    @TestVisible protected String getWhereClause() {
```
force2b: Since this is the main called outside this class to get the inner join portion for the main query locator, I think the method name should be something like `getMainQueryInnerJoinFilter()`.

kslewis: Ok, moved and renamed!

src/classes/CRLP_Batch_Base.cls
```diff
@@ -326,6 +335,214 @@ public abstract class CRLP_Batch_Base {
...
+        get {
+            if (maxYearsAgo == null) {
+                maxYearsAgo = getMaxInteger(CRLP_Operation.TimeBoundOperationType.Years_Ago);
+            }
+            return maxYearsAgo;
+        }
+        private set;
+    }
+
+    private Integer getMaxInteger(CRLP_Operation.TimeBoundOperationType timeBoundType) {
```
force2b: Needs an apex doc comment

force2b: `getMaxIntegerValueFromRollupsForType`

src/classes/CRLP_Batch_Base.cls
```diff
@@ -326,6 +335,214 @@ public abstract class CRLP_Batch_Base {
...
+                    && rollup.Integer__c != null) {
+                if (maxInteger == null || rollup.Integer__c > maxInteger) {
+                    maxInteger = (Integer) rollup.Integer__c;
+                }
+            }
+        }
+        return maxInteger;
+    }
+
+    private Boolean isIncrementalMode {
```
force2b: Needs apexdoc

force2b: Combine with `getIncrementalModeEnabled` as `getIsIncrementalModeEnabled`

src/classes/CRLP_Batch_Base.cls
```diff
@@ -326,6 +335,214 @@ public abstract class CRLP_Batch_Base {
...
+            whereClause = buildIncrementalModeWhereClause();
+        } else {
+            if (customizableRollupSettings.Disable_Related_Records_Filter__c == false) {
+                whereClause = buildNonIncrementalModeWhereClause();
+            }
+        }
+        return whereClause;
+    }
+
+    /**
```
force2b:
```
// Helper Methods
```

src/classes/CRLP_Batch_Base.cls
```diff
@@ -326,6 +335,214 @@ public abstract class CRLP_Batch_Base {
...
+     */
+    private Boolean getIncrementalModeEnabled() {
+        if (isScheduledJob == null || isScheduledJob == false) {
+            return false;
+        }
+
+        // Only run incremental mode when the custom settings are configured
+        // to enable incremental mode for this job type
+        Customizable_Rollup_Settings__c customizableRollupSettings = UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
+        if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit
```
force2b: This needs to check for skew/non-skew
`jobMode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode`

force2b:
```
if (jobMode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode) {
   switch (jobJtype) {
       when (CRLP_RollupProcessingOptions.RollupType.ContactHardCredit) {
           return customizableRollupSettings.ContactHardCreditNonSkew_Incremental__c;
       }
else {
}
```

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,378 @@
...
+            settings = UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
+                    Customizable_Rollups_Enabled__c = true,
+                    ContactHardCreditNonSkew_Incremental__c = isIncrementalMode
+            ));
+        }
+        return settings;
+    }
```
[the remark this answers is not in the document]
kslewis: good call!  done!

src/classes/CRLP_Batch_Base.cls
```diff
@@ -326,6 +335,226 @@ public abstract class CRLP_Batch_Base {
...
+    private Boolean getIsIncrementalModeEnabled() {
+        if (isScheduledJob == null || isScheduledJob == false) {
+            return false;
+        }
+
+        switch on (jobMode) {
+
+            when NonSkewMode {
+
+                Customizable_Rollup_Settings__c customizableRollupSettings =
```
force2b: @kslewis  The assignment for `customizableRollupSettings` can be above the first `switch`.

src/classes/CRLP_Batch_Base_TEST.cls
```diff
@@ -0,0 +1,361 @@
...
+    private static Customizable_Rollup_Settings__c configureIncrementalMode(
+            CRLP_RollupProcessingOptions.RollupType jobType,
+            Boolean isIncrementalMode) {
+
+        Customizable_Rollup_Settings__c settings;
+
+        switch on (jobType) {
+
+            when AccountHardCredit {
+                settings = UTIL_CustomSettingsFacade.getRollupSettingsForTests(
```
force2b: @kslewis  I think the creation of the settings can be above the switch. Then within each switch you can set the appropriate field to true:
`settings.AccountHardCreditNonSkew_Incremental__c = isIncrementalMode;`

src/classes/CRLP_Batch_Base.cls
kslewis: Ok, updated!

src/classes/CRLP_Batch_Base_TEST.cls
kslewis: Ok, updated!

PR #4049, opened 2019-02-12
Contact Merge TDTM Refactor

jjbennett:
// This is the parent feature branch for moving the future async method to a queueable async method. A child feature branch will be created with the change to queueable.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/CON_ContactMergeTDTM_TEST.cls
```diff
@@ -36,7 +36,8 @@
 
 @isTest
```
force2b: This is a big test class and I think it just doubled. Do you think it's worth testing the `IsParallel=true` annotation to see if the tests in this class will support parallel testing? Alternatively, could we split the tests out so that one of the test classes is capable of running in parallel while the other isn't?

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -105,93 +112,253 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
-        for (Contact con : listContactWinner) {
-        	if (con.Account.npe01__SYSTEM_AccountType__c == CAO_Constants.HH_ACCOUNT_TYPE)
-        	   listAccIdHH.add(con.AccountId);
-            if (con.Account.npe01__SYSTEM_AccountType__c == CAO_Constants.HH_ACCOUNT_TYPE ||
-                con.Account.npe01__SYSTEM_AccountType__c == CAO_Constants.ONE_TO_ONE_ORGANIZATION_TYPE ||
-                con.Account.npe01__SYSTEM_AccountType__c == CAO_Constants.BUCKET_ORGANIZATION_TYPE) {
-                mapContactIdAccountIdWinner.put(con.Id, con.AccountId);        
+
+        /*******************************************************************************************************
+        * @description Returns the account Id of accounts with household account types
```
RadSFDeveloper: Missing `@param` and `@return` for this method and some other methods as well.

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -105,93 +112,253 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
+        ********************************************************************************************************/
+        @TestVisible
+        private List<Id> getHouseholdAcctIds(List<Account> accounts) {
```
RadSFDeveloper: Suggestion: `getHouseholdIds() ` or `extractHouseholdIds()`? Usually `get` is reserved to indicate method getting data from database.

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -105,93 +112,253 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
-        	if (listAccountIdLoser[i] != null) {
-                Account accLoser = mapAccountIdAccountLoser.get(listAccountIdLoser[i]);
-                if (accLoser == null)
+
+        /*******************************************************************************************************
+        * @description Adds the account Id of the losing contacts account when it
+        * is any type of individual account
+        ********************************************************************************************************/
+        @TestVisible
+        private Map<Id, Id> getOldAcctIdForOppsByWinningContactId(List<Account> losersAccounts) {
```
RadSFDeveloper: Not sure how to change the name so it is shorter but still descriptive. These suggestions are too long: `extractOldIndividualAccountIdsByMasterContactId()`, or `mapOldIndividualAccountIdsToMasterContactIds()`.  Or can it be just `extractOldIndividualAccountIds()` while the result containing `Map` var can be more descriptive?

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -105,93 +112,253 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
+            return acctType == CAO_Constants.HH_ACCOUNT_TYPE;
+        }
+
+        /*******************************************************************************************************
+        * @description Determines if the account type provided matches any of the Individual Account Types
+        *
+        * @param acctType The account type to compare
+        * @return Boolean true if the string passed in matches any of the Individual Account Types
+        ********************************************************************************************************/
+        private Boolean isAnyIndividualType(String acctType) {
```
RadSFDeveloper: Suggestion: `isIndividualAccountType()`?

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -49,53 +60,49 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Contacts 
     * @return dmlWrapper.  always null.
     ********************************************************************************************************/
-    public override DmlWrapper run(List<SObject> listNew, List<SObject> listOld, 
+    public override DmlWrapper run(List<SObject> listNew, List<SObject> listOld,
```
RadSFDeveloper: Suggestions:
`listNew` -> `contacts' or `newContacts`
`listOld` -> `oldContacts`

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -49,53 +60,49 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
+
     /*******************************************************************************************************
     * @description returns a unique key for a tracking a losing Contact, their account, and the winning Contact.
     * @param contactIdLoser The losing Contact's Id
     * @param accountIdLoser The losing Contact's Account Id
     * @param contactIdWinner The winning Contact's Id
     * @return string
     ********************************************************************************************************/
-    private static string mergeKey(string contactIdLoser, string accountIdLoser, string contactIdWinner) {
+    private String mergeKey(string contactIdLoser, string accountIdLoser, string contactIdWinner) {
```
RadSFDeveloper: `getMergeKey()` or `buildMergeKey()`?

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -105,93 +112,253 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
+        mergeProcessor.fixup(mergeKeys);
+    }
+
+    /*******************************************************************************************************
+    * @description Fixes records related to merged contacts
+    ********************************************************************************************************/
+    @TestVisible
+    private class ContactMergeProcessor {
+        @TestVisible
+        private List<String> losingContactIds;
```
RadSFDeveloper: Even though `ContactMergeProcessor` is an inner class and the `fixup()` method called from the outer class, let's init collection vars when they are declared.

src/classes/CON_ContactMergeTDTM_TEST.cls
jjbennett: None of the new tests that were added committed any data to the database so it shouldn't cause the test to run long. It also shouldn't cause the debug logs to get too big, but I can take a look at moving things around.

jjbennett: I changed the tests in this class to use test setup data so that we can run in parallel. It is quite a bit more changes and there is one test that I had to comment out parts because when I made the changes there were unexpected results. I will comment in the code the line that is giving me grief.

src/classes/CON_ContactMerge_TDTM.cls
jjbennett: I went with build on all of the sets and map generators, thanks for the suggestion and please let me know if that reads better now.

src/classes/CON_ContactMergeTDTM_TEST.cls
```diff
@@ -269,65 +239,318 @@ public with sharing class CON_ContactMergeTDTM_TEST {
...
+            ORDER BY Id
+        ];
+        // This query causes the asserts to fail, the current address on the contact is being changed to the default account address during merge
+        // cannot replicate in the UI
+        // overrideAddress = [SELECT Id, Household_Account__c, MailingStreet__c, MailingCity__c FROM Address__c WHERE Id = :overrideAddress.Id];
+        overrideAddress = [SELECT Id, Household_Account__c, MailingStreet__c, MailingCity__c FROM Address__c WHERE Id = :masterContact.Current_Address__c];
+
+        System.assertEquals(3, addresses.size(), 'Expected both accounts to have an address each with an additional address for the master contacts address override.');
+        System.assert(masterContact.is_Address_Override__c, 'The master contact should still be set to override the address.');
+        System.assertEquals(masterContact.Current_Address__c, overrideAddress.Id, 'The master contact should still be set to the override address.');
```
jjbennett: @force2b - Previously we tested to make sure the master contact was still set to override and then asserted a count of default addresses but we did not verify the address on the master contact was still the override address and it turns out in the test it gets set to the accounts default address during merge. I tried to replicate in the UI but cannot. I could switch the test back to just asserting there are 2 default addresses but I don't want to cover up a potential bug or misconception in our tests.

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -39,159 +39,333 @@
...
+
+            return mergeProcessor;
+        }
+        set;
+    }
+
     /*******************************************************************************************************
     * @description Trigger Handler on Contact that handles detecting when an Contact Merge 
     * has completed, in order to fixup an assortment of affected information.
     * @param listNew the list of Contacts from trigger new.
```
RadSFDeveloper: `listNew` param has changed to `newContacts`.

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -39,159 +39,333 @@
...
-                con.Account.npe01__SYSTEM_AccountType__c == CAO_Constants.BUCKET_ORGANIZATION_TYPE) {
-                mapContactIdAccountIdWinner.put(con.Id, con.AccountId);        
+
+        /*******************************************************************************************************
+        * @description Returns the account Id of accounts with household account types
+        * @param accounts - The accounts to return the Id for when they are a household account
+        * @return List<Id> - The list of household account Ids
+        ********************************************************************************************************/
+        @TestVisible
+        private List<Id> buildHouseholdIds(List<Account> accounts) {
```
RadSFDeveloper: Suggestion: since `build` usually indicates building an sObject but not inserting it into the DB, maybe we can use `extract` or any other keyword indicating Ids of specific input Accounts will be returned?

jjbennett:
@rponti-sforg - I have 2 PRs that are related to work item, W-030823. I did a bit of work refactoring some of the code and wanted to ensure that the refactor itself didn't break any existing behavior. This work item has been approved and is ready for testing. The 2nd part of this work item is switching to Queueable which is done in PR 4062 which is currently in review. 4062 is pointing to this branch so if we merge this first, I will repoint 4062 to master. This work item is basically a regression test for the Contact Merge feature, there should be no change in behavior. Please let me know if you have any questions. Thank you!

jjbennett:
@rponti-sforg - to recap our conversation from this morning. This branch is now pointing to the branch were the fix for the crlp will be added. When you have a chance please feel free to merge. Thanks again!

rponti-sforg:
QE testing is complete for this PR.  Contact Merge regression tests have passed.  

For further testing information, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fQsOQAU/view

davisagli:
@rponti-sforg don't merge today; we are currently in merge freeze.

rponti-sforg:
@jjbennett Going forward, is safe to ignore the Feature Test - Robot failure and just go ahead and do a merge and delete branch?  Or do you want to run another build when this happens?

jjbennett:
@rponti-sforg - I like to look at them first and decide based on what's failing. I have looked at this test and it is just a glitch so this can be merged without it passing first. Since we are not merging into master, it should be fine to merge during the freeze. Thanks!

davisagli:
Oops, I didn't see that this is targetting a different feature branch rather than master. Yes, it is fine to go ahead and merge then.

PR #4053, opened 2019-02-13
Refactored Opportunity Mapping Stage To State API To Use NPSP Callabl…

jbrandolinisf:
# Critical Changes

# Changes
* You may notice two new fields on the Household Settings custom settings page: Opportunity Committed Stage and Opportunity Finalized Stage. These fields were created to support a future feature and shouldn't be used until the feature is released. Please ignore the fields for now.
# Issues Closed

# New Metadata
Classes:
* OPP_StageToStateMapping_API
* OPP_StageToStateMapping_API_TEST

Fields:
* npo02__Households_Settings__c.Opportunity_Committed_Stages__c
* npo02__Households_Settings__c.Opportunity_Finalized_Stages__c
# Deleted Metadata


jbrandolinisf:
@annejackson, @beAGoodQE  - Testing Notes:

Revised way to invoke API from Dev Console:

Callable npspApi = (Callable) Type.forName('CallableApi').newInstance();
System.debug((String)npspApi.call('OPP.MapStageToState', new Map<String, Object> { 'OpportunityStage' => 'Closed Won' }));

Note, the following:

Until the settings fields are populated, everything will return UNCOMMITTED. Since we are not exposing this via the UI, you will need to go to the Custom Settings area, and click the Manage button for the setting labeled "Household Settings". Enter Opportunity Stage names, separated by ; (to allow multiple values)

Nothing at this point enforces what you can/cannot set here, so it is possible to have a value of John defined as an Opportunity Committed or Finalized Stage, even though it is not an actual Opportunity Stage. The Settings UI to be built will enforce this. Further, you could have the same value in both fields. If this occurs, it will return FINALIZED when you call with that value.

Additionally, the method does not check to see if it is a valid Opportunity Stage Name, but instead assumes it is.

I recommend testing the API method with no stages defined in the two settings fields (the default), one value, and multiple values-- with and without overlap, confirming it selects the right value FINALIZED if present in the Opportunity Finalized Stage(s) field, COMMITTED if present in . the Opportunity Committed Stage(s) field (but not in the Opportunity Finalized Stage(s) field, and UNCOMMITTED in every other case (including if null/completely made up).

jbrandolinisf:
@annejackson, @beAGoodQE - FYI, just waiting for Wei and Mike to review it, then will pass it back to you!

weizsforg:
@jbrandolinisf Looks good in my side!

src/classes/CallableApi.cls
```diff
@@ -55,15 +55,16 @@ global with sharing class CallableApi implements System.Callable {
         switch on (cleanActionText) {
             when 'tdtm.disablealltriggers' {
                 TDTM_Config_API.disableAllTriggers();
-            }
-            when else {
+            } when 'opp.mapstagetostate' {
```
force2b: Can you add a blank line between the two when's to make this a bit more readable as it grows?
```
when 'tdtm.disablealltriggers' {
    // do something

} when 'opp.mapstagetostate' {
   // do something else

} when else {
   // throw exception

}
```

src/classes/OPP_StageToStateMapping_API.cls
```diff
@@ -0,0 +1,148 @@
+/*
```
force2b: Is this still an `API` class? Generally we apply the `API` suffix to classes that are externally visible, but that's not the case here. I think this is probably just fine at `OPP_StageToStateMapping`, though will this class ever support functionality beyond just "Stage to State Mapping"? Basically, I'm wondering if this is more related to Abacus utilities than just "Stage to State Mapping"

Could also be something like:
* `OPP_StageMappingUtil`
* `OPP_StageMapper`
* `OPP_AbacusUtil`

src/classes/OPP_StageToStateMapping_API.cls
```diff
@@ -0,0 +1,148 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Opportunity
+* @group-content ../../ApexDocContent/Opportunity.htm
+* @description API Class to allow caller to determine Opportunity's State (Uncommitted, Committed. or Finalized) using the Opportunities Stage and Household Settings
+*/
+public class OPP_StageToStateMapping_API {
```
force2b: I think this class would be better implemented as an instance class rather than all static classes, so something along the lines of:
```
    OPP_StageToStateMapping stageMapper = new OPP_StageToStateMapping(String stageName);
    String oppState = stageMapper.getOppState();
```

That said, depending on how this will be used in Abacus, maybe the class method is better as:
`String oppState = stageMapper.getOppStateFromStageName(String stageName);`

src/classes/OPP_StageToStateMapping_API.cls
```diff
@@ -0,0 +1,148 @@
...
+    /******************************************************************************************************
+    * @description Represents the overall state of the Opportunity.
+    * UNCOMMITTED -- Donor has not yet agreed to make the donation.  Donation is unpaid.  This is useful
+    * for soliciting or staging gifts.
+    * COMMITTED --  Donor has agreed to make a donation.  Donation can be unpaid or partially paid.
+    * More money is still expected from the donor.
+    * FINALIZED -- Donation is completely received.  Donation can be fully or partially paid.  No more
+    * money is expected for the donation.
+    */
+    public static final String UNCOMMITTED_STATE_VAL = 'UNCOMMITTED';
```
force2b: Do these need to be `public`?

src/classes/OPP_StageToStateMapping_API_TEST.cls
```diff
@@ -0,0 +1,261 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Opportunity
+* @group-content ../../ApexDocContent/Opportunity.htm
+* @description Test Coverage for OPP_StageToStateMapping_API class
+*/
+@isTest
```
force2b: I recommend marking this class as `IsParallel=true`

src/objectTranslations/npo02__Households_Settings__c-de.objectTranslation
```diff
@@ -50,5 +50,15 @@
         <label><!-- Use Dated Conversion Rates if available --></label>
         <name>Use_Dated_Conversion_Rates__c</name>
     </fields>
+    <fields>
```
force2b: The translation files are necessary for new fields, just new objects - unless you're actually providing the final translations for the new fields in this PR.

src/classes/OPP_StageToStateMapping_API.cls
```diff
@@ -0,0 +1,148 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Opportunity
+* @group-content ../../ApexDocContent/Opportunity.htm
```
force2b: Though our apexdoc generator is offline, if an html file is listed here it would also need to be updated to reflect that it's referenced in the class.

At the same time, I'm wondering if you need a new html file for Abacus related classes. I'm not tied to either approach, just thinking out loud based on what new classes/functionality may be coming down the road.

src/classes/OPP_StageToStateMapping_API.cls
jbrandolinisf: So, @pedrodalcol  is still working this out this some of the NGO team, but my understanding is that some of the Stage to State functionality may be desirable for NPSP as well as Abacus.  For example, it may be desirable to expose a setting that would allow a customer to automatically write off unpaid payments when an Opportunity becomes "Finalized".  I can refactor it to Util, but do not think it should become an Abacus centric class.

src/classes/OPP_StageToStateMapping_API.cls
jbrandolinisf: Just curious.  Why do you believe this to be better?  We would be calling it over and over for different opportunities, so it would be extra overhead to repeatedly instantiate to get there...

src/classes/OPP_StageToStateMapping_API.cls
jbrandolinisf: Maybe not for now, but it we would/extend functionality in Cumulus we will need to.  If you feel strongly on this, I will make them private/test visible for now.

src/classes/OPP_StageToStateMapping_API_TEST.cls
jbrandolinisf: I can do that, but will there be any problem with the multiple ones initiating household settings different ways?

src/objectTranslations/npo02__Households_Settings__c-de.objectTranslation
jbrandolinisf: Do you want me to remove these then?  I was just being overzealous, I guess.

src/classes/OPP_StageToStateMapping_API.cls
jbrandolinisf: Ok, I will update the Opportunity.htm file

jbrandolinisf:
@force2b - Can you review my responses?  I had a few counterpoints, and a question or two...

src/classes/OPP_StageToStateMapping_API.cls
force2b: Thanks. Based on your details I agree this is not an abacus centric class.

I still think a different class name is needed though. Maybe `OPP_StageMapper` or `OPP_StageMappingUtil`; the former if we go with an instance class, the latter if it stays static - if you agree.

src/classes/OPP_StageToStateMapping_API.cls
force2b: I suppose this depends on the use case. If you're making this call in a loop, I'd go with the model that instantiates the class once outside the loop (i.e., no parameters in the constructor), and then have a class method for `getOppStateFromStageName` that accepts a stage name string. Instantiation that doesn't do much in the constructor isn't terribly expensive either, unless of course it's happening 10000 times in a loop.

On the instance vs. static class, I think the instance class will offer more flexibility down the road if there is other related logic. Static classes should generally be for utility methods. To me this isn't necessarily a _utility_ method. 

The above noted, I could see where the class constructor accepts an Opportunity object and has a method to return the OppStage. That could eventually allow for new methods that return other needed details about the Opp that require some related translation (all TBD based on usage).

src/classes/OPP_StageToStateMapping_API.cls
force2b: If they're not needed currently, let's keep them at `private`. It's easy enough to make public when the time comes.

src/classes/OPP_StageToStateMapping_API_TEST.cls
force2b: The settings objects aren't committed in unit tests.

src/objectTranslations/npo02__Households_Settings__c-de.objectTranslation
force2b: No need to remove them unless you really want to.

src/classes/OPP_StageToStateMapping_API.cls
jbrandolinisf: I would like to keep it a UTIL Class.  I tend to think of instance classes when different instances of the class (no pun intended) have/need their own private variables.  This does not/and it is undesirable.  The only variables supporting this process are static.  If you feel super strongly I will go the other route, but it feels like a static utility class to me...

I do get that there may come a time where we might want to do decorate an Opportunity with additional functionality (passing it into a constructor), but it is really a different use case than this is built for (and a separate class could be built for that, but still use the util class for this piece).

force2b: Good explanation. Let's go with `OPP_StageMappingUtil` as the class name then.

src/objectTranslations/npo02__Households_Settings__c-de.objectTranslation
jbrandolinisf: It's good to know.  I started doing that as part of my process.  I will discontinue, and only worry about it for objects then.

src/classes/OPP_StageMapping_UTIL.cls
```diff
@@ -0,0 +1,148 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Opportunity
+* @group-content ../../ApexDocContent/Opportunity.htm
+* @description API Class to allow caller to determine Opportunity's State (Uncommitted, Committed. or Finalized) using the Opportunities Stage and Household Settings
+*/
+public class OPP_StageMapping_UTIL {
```
force2b: @jbrandolinisf  I was going to say that ending classes in `_UTIL` is not something we have in NPSP, but there are 5 classes with that suffix. I think that `OPP_StageMappingUtil` and `OPP_StageMappingUtil_TEST` are easier to read (particularly the test class name since there are no examples of classes with a name ending in `_[*]_TEST`), but I'm OK if you'd like to leave them as is.

Also, can you add `with sharing` to this class definition?

jbrandolinisf: Your call.  I was going with what was in the repo, but I am ok with whatever.

force2b: > _Naming is an art, and I'm not an artist._ :)

Anyway, let's go with `OPP_StageMappingUtil` and `OPP_StageMappingUtil_TEST` if you don't mind.

jbrandolinisf:
@force2b - As always, you are the best!  Thanks for working with me on this!!!

beAGoodQE:
Testing is complete.  Merging this PR into feature/patrick-npsp-api

PR #4055, opened 2019-02-13
BGE: Donation Campaign Source

bethbrains:
# Critical Changes

# Changes
- Batch Gift Entry supports Donation Campaign Source

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
**lurch: add W-031066

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fRC4QAM" target="_blank">W-031066</a>

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007x7FyQAI" target="_blank">W-031312</a>

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -212,27 +199,39 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
...
-    }
+        List<String> sortedFieldLabels = new List<String>(dfrMap.keySet());
+        sortedFieldLabels.sort();
+
+        List<Schema.DescribeFieldResult> sortedDFRs = new List<Schema.DescribeFieldResult>();
+        for (String fieldLabel : sortedFieldLabels) {
+            sortedDFRs.add(dfrMap.get(fieldLabel));
+        }
 
+        return sortedDFRs;
```
bethbrains: Moved this down here from above so that we can sort on Field Label instead of Field API Name.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -502,6 +462,13 @@
         // getActives grabs allFields, returns those isActive, sorted.
         let activeFields = this.getActives(component);
 
+        activeFields.forEach(function(currentField) {
+            if (currentField.defaultValue && currentField.type == 'reference') {
+                // lookups in recordeditform store as an array of IDs; need to flatten
```
ptewson-sfdo: Do we know why they are stored in an array? A lookup would only ever point to a single Id.... right?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -44,14 +44,23 @@
...
+            editable: false,
+            typeAttributes: {label: {fieldName: 'donorName'}}});
 
         dataColumns.forEach(function(col){
-            columns.push({label: col.label,
-                fieldName: col.fieldName,
-                type: col.type,
-                editable: !col.readOnly,
-                typeAttributes: JSON.parse(col.typeAttributes)});
+            if (col.type !== 'reference') {
```
ptewson-sfdo: I'm missing the nuance of why aren't we including reference fields in the columns ... since I remember seeing them there! 

Unless it's because of the comment below for setDataImportFields() where there's a distinction between the read-only fields and editable fields in the grid?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -532,13 +547,20 @@
...
      */
-    flattenDataImportRow: function(currentRow) {
+    flattenDataImportRow: function(namespace, currentRow) {
         let row = currentRow.record;
         row.donorName = currentRow.donorName;
         row.donorLink = currentRow.donorLink;
         row.matchedRecordUrl = currentRow.matchedRecordUrl;
         row.matchedRecordLabel = currentRow.matchedRecordLabel;
         row.errors = currentRow.errors;
+        // reformat lookup info to display as a url in datatable
```
ptewson-sfdo: I wonder if it might be worth refactoring the DataImportRow class in apex to support the translation for reference Id to the link+label needed to build the data table link below, especially as we move towards support any lookup fields generically. 

Even here, the operation for the Campaign Link seems similar in theory to the donor link or the matched record link, but the logic for the first two is handled in the apex DataImportRow class, and the campaign link logic is here in javascript. Is there a reason not to do all the three in the same place, either in the DataImportRow class or in this method?

ptewson-sfdo: Follow-up thought, if we include a list of relationship field info from the Apex side of things, would that be a good setup to move towards handling generic lookup fields, as opposed to handling each specific case differently?

src/classes/BDI_DataImportService.cls
```diff
@@ -728,6 +751,11 @@ global with sharing class BDI_DataImportService {
         DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(dataImportObjectName, sourceFieldName);
         String targetField = dfr.inlineHelpText != null ? dfr.inlineHelpText.split('\\.')[1] : null;
 
+        // Donation Campaign is a special case because its mapping is hard-coded, not in help text
```
lmeerkatz: How is this handled for Contact1Imported, Contact2Imported, etc? Those are all hard-coded somewhere, right?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -180,77 +182,82 @@
...
+                                                                  value="{!field.defaultValue}"
+                                                                  variant="label-hidden"
+                                                                  class="hideLabel">
+                                                    <option value=""></option>
+                                                    <aura:iteration items="{!field.options}" var="opt">
+                                                        <option value="{!opt.value}">{!opt.label}</option>
+                                                    </aura:iteration>
+                                                </lightning:select>
+                                                <aura:set attribute="else">
+                                                    <!-- TODO: ensure these are accessible -->
```
randi274: Is there additional testing required here?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -124,9 +124,11 @@
                 currentField.requiredInEntryForm = activeFieldMap.get(currentField.id).requiredInEntryForm;
                 currentField.sortOrder = activeFieldMap.get(currentField.id).sortOrder;
                 currentField.type = activeFieldMap.get(currentField.id).type;
-                currentField.formatter = activeFieldMap.get(currentField.id).formatter;
                 currentField.options = activeFieldMap.get(currentField.id).options;
                 currentField.alwaysRequired = activeFieldMap.get(currentField.id).alwaysRequired;
+                if (currentField.defaultValue && currentField.type == 'reference') {
```
randi274: Maybe add a comment here for why this is necessary?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -146,7 +148,7 @@
         var allFieldsBySObject = this.groupFieldsBySObject(everyField);
 
         const opportunitySObjectName = "Opportunity";
```
randi274: Could/should we get these from our labels file? Also, can you change these both to single quotes?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -349,20 +351,6 @@
         return isValid;
     },
 
-    /**
```
randi274: 👏 love👏 deleting 👏  stale 👏  code 👏

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
randi274: According to the LWC side of the docs, they allude to the possibility of multiple lookups someday. We're guessing it was easier for them to always assume an array in their code?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
randi274: This could be a good thing to discuss in a group code review... but your assumption is correct! See the comment below on 89 for clarity on that.

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
randi274: The logic for the matching URL/label is a good bit more complex. I'm for any future lookup fields being handled in the JS, but would push back a bit on the system-provided matching field and the lookup fields added by the user being semantically similar. This might also be something to refactor in group review.

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -243,8 +239,8 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         String dataImportApiName = UTIL_Namespace.StrTokenNSPrefix('DataImport__c');
         String opportunityName = 'Opportunity';
         String opportunityLabel = UTIL_Describe.getObjectLabel(opportunityName);
-        String paymentName = 'Payment'; // bdi object name, not api name
-        String paymentLabel = UTIL_Describe.getObjectLabel('npe01__OppPayment__c');
+        String paymentName = 'npe01__OppPayment__c';
```
randi274: Maybe these should also be schema values instead of hardcoded strings?

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -553,7 +549,6 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
         @AuraEnabled public Boolean alwaysRequired; // always required in BGE (e.g., Donation Amount and Donation Date)
         @AuraEnabled public Boolean conditionallyRequired; // required if any other field on its object is selected (e.g., Account Name and Contact Last Name)
         @AuraEnabled public String defaultValue;
-        @AuraEnabled public String formatter;
```
randi274: 😮 👍

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -346,28 +347,40 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
-
                         col.options = BGE_BatchGiftEntry_UTIL.getPicklistOptions(col.fieldName);
 
                         columns.add(col);
 
-                        String targetObject = BDI_DataImportService.getTargetObject(col.fieldName);
-                        if (targetObject == 'Payment') {
-                            targetObject = 'npe01__OppPayment__c';
+                        if (dataImportDFR.getType().name() == 'reference') {
+                            // manually construct EXTRA column for campaign link
```
randi274: 😢

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -346,28 +347,40 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
+                            Column linkCol = new Column();
+                            linkCol.label = dataImportDFR.getLabel();
+                            String fieldLabel = dataImportDFR.name + 'label';
+                            String fieldName = dataImportDFR.name + 'link';
+                            linkCol.fieldName = fieldName;
+                            linkCol.readOnly = true;
+                            linkCol.required = field.requiredInEntryForm;
+                            linkCol.type = 'url';
+                            linkCol.typeAttributes = '{"label":{"fieldName":"'+fieldLabel+'"},"target":"_blank"}';
```
randi274: This blank tab was also not working for me. Maybe we need to retest this and find out why?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
bethbrains: There is a multi-lookup design pattern in core and in SLDS, so yeah, as Randi said, it looks like they're just working ahead toward a base component for that.

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
bethbrains: There are two separate "Column" objects -- one for the actual lookup in the form and one for the link output in the grid, since there is no native handling for reference datatype in datatable and it just displays as a raw id. We create our own separate "link" url-type Column and display that one in the datatable instead.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
ptewson-sfdo: Coolbeans. Carry on!

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
ptewson-sfdo: I think if this is part of a strategic switch to handling in javascript, it's probably ok to use this as a bridge opportunity. Just as long as we are being thoughtful/intentional about it!

bethbrains: I wrote this section with an eye toward being able to parse any number of user-defined lookup fields in a loop and construct them here. I'll back Randi up on the fact that the donor and donation ones have additional logic behind them (Account or Contact? Opportunity or Payment?) that likely make them better left back in apex. I'd love to review this as a group and see if there's a better solution!

src/classes/BDI_DataImportService.cls
bethbrains: Great question. I'm not sure there's a central spot for all of the lookups because of how they're used historically as outputs, not inputs. I'll look around a bit. Because we're making Account1Imported and Contact1Imported always available, we're handling them elsewhere for now, but this is certainly something to look at at *least* when/if we get to offering more of those lookups as inputs.

src/classes/BDI_DataImportService.cls
```diff
@@ -704,17 +704,40 @@ global with sharing class BDI_DataImportService {
...
+    * @return String the api name of the target object for that field, derived from the help text
     */
     public static String getTargetObject(String sourceFieldName) {
         sourceFieldName = sourceFieldName.toLowerCase();
         String dataImportObjectName = UTIL_Namespace.StrTokenNSPrefix('DataImport__c');
         DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(dataImportObjectName, sourceFieldName);
         String targetObject = dfr.inlineHelpText != null ? dfr.inlineHelpText.split('\\.')[0] : null;
 
-        return targetObject;
+        // Donation Campaign is a special case because its mapping is hard-coded, not in help text
```
ptewson-sfdo: What's the reason we aren't using a help text mapping like other fields?

bethbrains: Because it needs to work both as an input from BGE and as an output from BDI if they use the legacy campaign name matching.

lmeerkatz: Even legacy records should work using the help text mapping - as soon as they run matching (through dry run or import) existing records should be updated to populate the Id fields. /cc @randi274 for confirmation

bethbrains: Can we discuss this on a call? I'm not understanding the question. The campaign name-matching feature doesn't use help text mapping either.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
randi274: Update: testing is done, and we do need to add an aria label here because the field label is being lost in this format.

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
```diff
@@ -145,8 +148,8 @@
         // returns map of sobject name => list of fields
         var allFieldsBySObject = this.groupFieldsBySObject(everyField);
 
-        const opportunitySObjectName = "Opportunity";
-        const paymentSObjectName = "Payment";
+        const opportunitySObjectName = component.get('v.labels.opportunitySObjectName');
```
lmeerkatz: This seems like one place we wouldn't want a label. We're talking about API names not object label, right?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
ptewson-sfdo: It's not totally intuitive to think of the fields in the form as "columns" but I think I can let it go

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizardHelper.js
bethbrains: Yep I have fixed this just not pushed yet...

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -256,6 +256,7 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
             'donationAmountField' => Schema.SObjectType.DataImport__c.fields.Donation_Amount__c.getName(),
             'expectedCountField' => Schema.SObjectType.DataImportBatch__c.fields.Expected_Count_of_Gifts__c.getName(),
             'expectedTotalField' => Schema.SObjectType.DataImportBatch__c.fields.Expected_Total_Batch_Amount__c.getName(),
+            'namespacePrefix' => UTIL_Namespace.StrTokenNSPrefix(''),
```
ptewson-sfdo: Why not UTIL_Namespace.getNamespace()?

src/classes/BGE_DataImportBatchEntry_CTRL.cls
ptewson-sfdo: I like that Randi's review is at least 50% in emoji form :)

src/classes/BDI_DataImportService.cls
bethbrains: Removed, as we've adjusted things to rely on help text for Campaign ID mapping.

src/classes/BDI_DataImportService.cls
bethbrains: Removed, since we are treating Donation Campaign as a normal mapped field now with help text mapping.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
bethbrains: That returns the namespace without the `__`. This method uses that method and concatenates the namespace AND the `__` to the string you pass it, so this is the way to get `npsp__` (OR just the empty string back if there is no namespace) which is what we need in JS to prepend field names dynamically.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
bethbrains: We called this quirkiness known and out of scope for this ticket, but it would be nice to track down eventually.

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -115,44 +115,27 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
...
-            if (conditionallyRequiredFields.contains(unnamespacedField)) {
+            field.sObjectName = BDI_DataImportService.getTargetObject(dfr.getName());
+            field.sObjectLabel = UTIL_Describe.getObjectLabel(field.sObjectName);
+
+            if (conditionallyRequiredFields.contains(field.name)) {
                 field.conditionallyRequired = true;
             }
 
-            String rawFieldType = dfr.getType().name().toLowerCase();
             field.type = UTIL_LightningComponentSupport.getLightningInputDataType(dfr.getType());
```
ptewson-sfdo: Do we need type in this context?

randi274: Yep, still do! It's used to see if it's a reference and we have to do some converting in 2 places in the JS, so I'd vote to keep this for now, or make this a separate ticket. Maybe part of a ticket to increase the speed of the load, since it seems to have gotten slow again?

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.css
```diff
@@ -55,5 +55,17 @@
...
 /* hide rogue label spacing on lightning:select when variant="label-hidden" */
 .THIS .label-hidden > label {
-    display: none;
+    position: absolute!important;
+}
+/*Workaround since label-hidden variant on lightning:inputField is inconsistent in hiding labels for types of fields.
+Revist this Summer '19 and after to see if it's still necessary or the CSS has changed*/
+.THIS .hideLabelFix lightning-grouped-combobox .slds-form-element__label {
+    /*partial copy from .slds-assistive-text*/
+    position: absolute!important;
```
ptewson-sfdo: I'm assuming there's no alternative to "!important" here?

randi274: That's a good question; I really just copied/pasted it over and didn't check, but would assume so! This might be specific enough to override without the !important, but would need to test that again.

PR #4059, opened 2019-02-14
Feature/crlp incremental hard credit rollups  user defined inner join

kslewis:
# Critical Changes

# Changes

- Adds custom setting fields that allow configurable date field and value to be used when selecting summary objects to be rolled up in nightly scheduled batch job

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Batch_Base.cls
+++ b/src/classes/CRLP_Batch_Base.cls
@@ -467,8 +467,8 @@ public abstract class CRLP_Batch_Base {
                 Customizable_Rollup_Settings__c customizableRollupSettings =
                         UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
 
-                if (customizableRollupSettings.Incremental_Last_N_Days_Field_Override__c != null) {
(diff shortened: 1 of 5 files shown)


src/classes/CRLP_Batch_Base.cls
```diff
@@ -433,24 +433,64 @@ public abstract class CRLP_Batch_Base {
...
-            whereClause += '>= LAST_N_DAYS:' + maxDaysAgo;
-        } else if (maxYearsAgo != null) {
-            whereClause += '>= LAST_N_YEARS:' + maxYearsAgo;
         }
         return whereClause += ')';
     }
 
+    private Boolean hasUserDefinedInnerJoin {
+        get {
+            return userDefinedLastNDaysField != null && userDefinedLastNDays != null;
```
force2b: @kslewis I was originally thinking that the customer can able to define or or both of these fields independently rather than require them to define both, but I think this fine as is for the MVP

PR #4062, opened 2019-02-15
Change contact merge fixup from future to queueable

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


jjbennett:
@force2b @RadSFDeveloper - figured I'd tag you both on the actual work item PR, it should be short and sweet to review. No rush, since I do not believe we'd be able to QA and merge this sprint.
@force2b - we are checking for customizable rollups for the contact rollup fixup but not for the account rollup fixup. Should we add a work item for that?

jjbennett:
**lurch: attach W-030823

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQsOQAU" target="_blank">W-030823</a>

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -106,23 +106,32 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
...
     *   - recalculate Opportunity Rollups
     *
-    * @param mergeKeys The list of mergeKeys that provide oldContactId, oldAccountId, and masterContactId
-    * @return null
     ********************************************************************************************************/
-    @Future
-    private static void handleContactMergeFixupsFuture(List<String> mergeKeys) {
-        if (mergeKeys == null || mergeKeys.isEmpty()) {
-            return;
+    private class AsyncContactMergeFixup implements Queueable {
```
RadSFDeveloper: Suggestion: `ContactMergeFixupQueueable`?

jjbennett: I actually went with async intentionally so no matter what we decided to use we could keep the name the same... thoughts? I also noticed our error class was named with Async.

RadSFDeveloper: It implements Queueable, so I thought it would help to be clear and descriptive in the class name, similarly how we add `future` or `batch` in respective method/class names. It is optional.

jjbennett:
@RadSFDeveloper - thanks for the feedback, after our discussion I did change that inner class name so that it would not be ambiguous!
@force2b - did you have a chance to look this one over yet?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -422,19 +422,23 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
     * @return void
     */ 
-    public static void rollupAccounts(list<Id> listAccountId) {
-        if (listAccountId.size() == 1 || system.isFuture() || system.isBatch()) {
-            RLLP_OppRollup_UTIL oppRollupUtil = new RLLP_OppRollup_UTIL(); 
-            map<id, Account> amap = new map<id, account>((list<Account>)Database.query(oppRollupUtil.buildAccountQuery() + 
-                                                           ' where id IN :listAccountId')); 
-            RLLP_OppRollup rg = new RLLP_OppRollup();
-            rg.rollupAccounts(amap);                    
+    public static void rollupAccounts(List<Id> accountIds) {
```
force2b: It looks like this method only works with legacy rollups. If that's the case, we should fix this so that it calls either Legacy or Customizable Rollups based on what's enabled. The `CON_ContactMerge_TDTM` class has logic that does this

src/classes/CON_ContactMerge_TDTM.cls
```diff
@@ -77,7 +77,7 @@ public class CON_ContactMerge_TDTM extends TDTM_Runnable {
                 mergeKeys.add(generateMergeKey(oldContact.Id, oldContact.AccountId, oldContact.MasterRecordId));
             }
 
-            handleContactMergeFixupsFuture(mergeKeys);
+            System.enqueueJob(new ContactMergeFixupQueueable(mergeKeys));
```
force2b: Though I'm pretty sure that Customizable Rollups run synchronously when called from this class, it's worth testing  this with CRLP both enabled and disabled to ensure that the rollups run as expected without any errors.

force2b: Worse case, it is feasible to change queueables so that if needed ContactMergeFixupQueueable() could execute another queueable to handle the Contact rollups.

src/classes/ACCT_IndividualAccounts_TDTM.cls
jjbennett: @force2b - I agree but I think it should be a separate work item.

force2b: Fair enough. Can you create one for this?

jjbennett:
@force2b - I created the issue / work item for the account rollup call. Is there anything else that needs to be changed in this PR or is this ready for QA?

jjbennett:
Thanks @force2b - I believe both of those scenarios are covered in the tests Sean added in TDTM_TEST2. 
contactMergeShouldNotUseLegacyRollupWhenCRLPEnabled
contactMergeShouldUseLegacyRollupWhenCRLPNotEnabled

force2b:
Thanks @jjbennett. Though, looking at the `contactMergeShouldNotUseLegacyRollupWhenCRLPEnabled` test, I'm not really sure that is properly setup. Normally customizable rollup metadata has to be mocked and injected to be used in a test. If I had to guess, this code relies on the CRLP metadata existing in the org that the test is running in.

jjbennett:
@force2b - I don't think we need a full integration test, we just need to know that the correct rollup process was called.

PR #4067, opened 2019-02-20
BGE - Donation Matching Rule scroll fix

randi274:
# Critical Changes

# Changes
- Fixed a scroll issue in the Donation Matching Rule dual picklist
# Issues Closed

# New Metadata

# Deleted Metadata


randi274:
**lurch: attach W-031292

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007x7EMQAY" target="_blank">W-031292</a>

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -116,15 +116,15 @@
...
+                <aura:if isTrue="{!v.wizardMetadata.progressIndicatorStep == '1' }">
 
                     <div class="slds-m-around_medium">
 
                         <aura:if isTrue="{!greaterthan(v.batchInfo.recordCount,0)}">
                             <lightning:layout horizontalAlign="center">
                                 <lightning:layoutItem size="12" class="slds-grid slds-grid_align-center slds-p-bottom_medium">
                                     <div class="slds-notify slds-notify_alert slds-theme_alert-texture slds-theme_warning" role="alert">
-                                        <span class="slds-assistive-text">warning</span>
+                                        <span class="slds-assistive-text">{!$Label.c.PageMessagesWarning}</span>
```
ptewson-sfdo: 👍

src/aura/BGE_ConfigurationWizard/BGE_ConfigurationWizard.cmp
```diff
@@ -256,157 +255,160 @@
...
-                        <lightning:layoutItem size="12" class="slds-p-bottom_small">
-                            <lightning:layout horizontalAlign="spread" multipleRows="true">
-                                <lightning:layoutItem class="slds-p-horizontal_x-small" size="12">
-                                    <div class="slds-section">
-                                        <h3 class="slds-section__title slds-theme_shade">
-                                                <span class="slds-p-left_small slds-truncate">
-                                                    {!$Label.c.bdiSettingsSectionDonationMatching}
-                                                </span>
-                                        </h3>
+                <div class="{!if(v.wizardMetadata.progressIndicatorStep == '3', '', 'slds-hide')}">
```
ptewson-sfdo: I'm understanding that this change is at the crux of fixing this issue, but I think this will include a ton more markup in the form - basically all the step 3 markup will be delivered in every step of the form, just hidden by css, as opposed to to the "aura:if" which would actually remove this block from the dom. Does that have performance, or maybe even functional considerations? 

I am thinking something like values from hidden form fields could actually get accessed and written to the record or something akin to that...

randi274: Yes, I understand the concern. Luckily, I was able to remove the slds-hide and replace it with an aura:if on the second step (also unclear why that worked, but that's a good trade-off!), and in step 3, there are large blocks of aura:if code that we rely on to keep things smaller too. I've added some specific aura:ifs on the progress indicator too, so we can still keep this light and return scroll functionality!

randi274: https://github.com/SalesforceFoundation/Cumulus/pull/4067/commits/5f5dc37a8dcde282b70be16a95259d23caae6f1d

PR #4070, opened 2019-02-21
Fix robot library initialization to be able to generate keyword documentation

boakley:
This includes a fix for NPSP.py; it was failing to be instantiated
when robot tries to import it when generating documentation. That is
because the __init__ tries to get a variable from a running test,
and no test is running when we generate the documentation.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/tests/NPSP.py
+++ b/tests/NPSP.py
@@ -2,7 +2,7 @@ import logging
 import re
 import time
 
-from robot.libraries.BuiltIn import BuiltIn
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
 from selenium.common.exceptions import ElementNotInteractableException
 from selenium.common.exceptions import StaleElementReferenceException
 from selenium.common.exceptions import NoSuchElementException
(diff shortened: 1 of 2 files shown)


tests/NPSP.py
```diff
@@ -42,10 +42,16 @@ def __init__(self, debug=False):
...
+        try:
+            client = self.cumulusci.tooling
+            response = client._call_salesforce(
+                'GET', 'https://{}/services/data'.format(client.sf_instance))
+            latest_api_version = float(response.json()[-1]['version'])
+        except RobotNotRunningError:
+            # We aren't part of a running test, likely because we are
+            # generating keyword documentation. If that's the case, assume
+            # the latest supported version
+            latest_api_version = max(locators_by_api_version.keys())
```
davisagli: @boakley This seems like a good workaround. I've actually been intending to move the code for getting the API version from here into the CumulusCI or Salesforce library. Do you think we should do that now or separately?

tests/NPSP.py
```diff
@@ -749,9 +755,11 @@ def click_object_manager_button(self,title):
 
     def page_scroll_to_locator(self, path, *args, **kwargs):
         locator = self.get_npsp_locator(path, *args, **kwargs)
+        import sys; sys.stdout=sys.__stdout__
+        import pdb; pdb.set_trace()
```
davisagli: stray pdb

boakley: d'oh! Thanks for catching that.

tests/NPSP.py
boakley: I think it should probably be a separate issue.

boakley: I didn't put a ton of thought into this workaround. We can probably do better, but I didn't want that to derail getting the cci task finished.

tests/NPSP.py
boakley: I've removed them and pushed the changes

tests/NPSP.py
boakley: I created issue #W-031439 to create the function.

PR #4081, opened 2019-02-26
BDI: Record type works even when unmapped

bethbrains:
# Critical Changes

# Changes
--Donation Record Type works in all orgs, regardless of help text mapping.

# Issues Closed

# New Metadata

# Deleted Metadata


bethbrains:
**lurch: attach W-031346

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007x7IiQAI" target="_blank">W-031346</a>

bethbrains:
> Is the help text mapping a user-facing, documented feature? Is there ever a use case where a User would want to configure this field to map somewhere other than to Opportunity.RecordTypeId? With this change, we override whatever value is in the help text for this field on the Data Import object, so we might either want to check the help text value first to see if there is a User-configured alternate mapping, or document the fact that this Data Import field is now mapped directly to Opportunity.RecordTypeId regardless of whatever value is/isn't in the help text for this field.

@kenslewis  et al! 
summarizing discussion from 2/27:
We are leaving this refactor as-is. BDI does not respect any custom mapping an admin may have included in the DonationRecordTypeName field. It is unclear why the help text was added at all. When BDI loops through mappings to update the opp, this field is explicitly ignored: https://github.com/SalesforceFoundation/Cumulus/blob/910ea10dcec4c10da4613cad7038b4d8e0734015/src/classes/BDI_Donations.cls#L440 and when writing Opportunity record type, it is used without regard for its mapping: https://github.com/SalesforceFoundation/Cumulus/blob/910ea10dcec4c10da4613cad7038b4d8e0734015/src/classes/BDI_Donations.cls#L380

PR #3947, opened 2018-12-28
BGE: Prevent multiple new batch modals from opening

randi274:
# Critical Changes

# Changes
- Prevent multiple modals from opening in Batch Gift Entry
# Issues Closed

# New Metadata

# Deleted Metadata

PR #4094, opened 2019-03-04
Fix multiple clicks opening multiple modals

psadie:
- Apply multiple modals fix to BGE 'Edit' button and 'Review Donations' anchor.
  - [Fix Source](https://github.com/SalesforceFoundation/Cumulus/pull/3947#issuecomment-450432787)

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


psadie:
**lurch: attach W-031525

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008Ll8KQAS" target="_blank">W-031525</a>

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -442,7 +442,10 @@
         action.setCallback(this, function (response) {
             var state = response.getState();
             if (state === 'SUCCESS') {
-                this.openBatchWizard(component, event);
+                if (!component.get('v.modalOpen')) {
```
ptewson-sfdo: This comment arguably falls outside of this work, but we recently did some refactor to similar code in BGE_BatchGiftEntryTabHelper.js and I wonder if it's worth bringing the same update in here as long as we are opening the hood - if you look at the "openNewBatchWizard" and "checkFieldPermissions" functions in there, we switched to passing in the 'onSuccess' behavior as an argument to this method, in an effort to avoid hard-coding the wizard open behavior as an outcome of checking the field permissions. 

Let me know if that all makes sense!

psadie: Got it. Checking that out now

src/aura/BGE_EntryForm/BGE_EntryFormController.js
```diff
@@ -104,20 +106,26 @@
                         header: component.get('v.donationModalHeader'),
                         body: newcomponent,
                         showCloseButton: true,
-                        cssClass: 'slds-modal_large'
+                        cssClass: 'slds-modal_large',
+                        closeCallback: function() {
+                            component.set('v.modalOpen', false);
```
kslewis: Since "component.set('v.modalOpen', false);" is used three times here (also on lines 115 and 123), we could explore defining this as a named function within this function and then call it directly on these lines, like:

let setModalOpenToFalse = function() {
    component.set('v.modalOpen', false);
}

then you could reference it as the closeCallback function on line 110, and call it below using setModalOpenToFalse()

src/aura/BGE_EntryForm/BGE_EntryFormController.js
```diff
@@ -104,20 +106,26 @@
...
+                        }
                     });
                 } else if (status === 'INCOMPLETE') {
```
kslewis: These two else if blocks could be somewhat combined within an outer "else".  The declaration of the message variable, the "component.set('v.modalOpen', false);" (or new setModalOpenToFalse()) and the call to helper.sendMessage could then be written only once within the outer "else" block, while the contents of the message can be generated within the inner if block that checks the error status...

psadie: Hey Ken, what do you think of something like this inside the outer "else"?

setModalOpenToFalse();
const message = {
    title: $A.get('$Label.c.PageMessagesError'),
    errorMessage: status === 'ERROR' ? errorMessage : $A.get('$Label.c.stgUnknownError')
 };
helper.sendMessage('onError', message);

kslewis: Hey Prince, Yep, structured like this would be great!

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -481,9 +485,13 @@
...
                         footer: modalFooter,
                         showCloseButton: true,
-                        cssClass: 'slds-modal_large'
+                        cssClass: 'slds-modal_large',
+                        closeCallback: function() {
+                            component.set('v.modalOpen', false);
+                        }
                     })
                 } else {
+                    component.set('v.modalOpen', false);
```
kslewis: What do you think about doing the same thing here and using a named function in both places?

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryHelper.js
```diff
@@ -467,7 +471,7 @@
         ];
 
         $A.createComponents([
-                ['c:BGE_ConfigurationWizard', {sObjectName: 'DataImportBatch__c', recordId: batchId}],
+                ['c:BGE_ConfigurationWizard', {sObjectName: 'DataImportBatch__c', recordId: component.get('v.recordId')}],
```
kslewis: :clap:

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryController.js
```diff
@@ -63,7 +63,9 @@
      * @description: checks that user has all necessary permissions and then launches modal or displays error
      */
     onEditClick: function(component, event, helper) {
-        helper.checkFieldPermissions(component, event, helper);
+        helper.checkFieldPermissions(component, function () {
+            return helper.openBatchWizard(component);
+        }.bind(helper));
```
kslewis: I know I just used .bind recently, but what do you think about a passing a named function here also?  I've been researching it a bit and it has a few benefits - let me know if you want to discuss

src/aura/BGE_EntryForm/BGE_EntryFormController.js
```diff
@@ -104,20 +110,19 @@
...
+                        cssClass: 'slds-modal_large',
+                        closeCallback: setModalOpenToFalse
                     });
-                } else if (status === 'INCOMPLETE') {
+                } else {
+                    setModalOpenToFalse();
                     const message = {
                         title: $A.get('$Label.c.PageMessagesError'),
-                        errorMessage: $A.get('$Label.c.stgUnknownError')
+                        errorMessage: status === 'ERROR' ? errorMessage : $A.get('$Label.c.stgUnknownError')
```
kslewis: This function looks great - nice work!

src/aura/BGE_DataImportBatchEntry/BGE_DataImportBatchEntryController.js
psadie: Would love to hear more about your thoughts around this and how it'd look

sam-knox:
@kenlewis-sfdo - Does this change any behavior for the end user? Do we need to say anything is release notes on this one?

PR #4103, opened 2019-03-07
CRLP: Api to retrieve Rollup Definition Collection for a specific Rollup Type

force2b:
Implement a Callable_API action for `CRLP.getDefinitionsForRollupType` that accepts a string with the type of rollup (ex: "AccountHardCredit") and returns a `List<CRLP_Rollup>` collection. This new api will be used by the DSO-NPSP Bridge to generate rollup aggregation for archived data. 

It is currently not included in the release notes because as an Api it has a limited use for customers until further Customizable Rollup api are exposed through the Callable_API class.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
** lurch:attach W-031334

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007x7HkQAI" target="_blank">W-031334</a>

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @description Validate the getRollupDefsForRollupType returns an exception if the RollupType is null or invalid
+     */
+    @IsTest
+    static void validateGetRollupDefsForRollupTypeInvalidParameter() {
+        CRLP_ApiService crlpApiSvc = new CRLP_ApiService();
+
+        Test.startTest();
+        try {
```
RadSFDeveloper: Since the same conditions are tested for various params, we can apply the same pattern to reduce number of statements. Also the verified error message should check for the param name (that probably should be a `CRLP_ApiService` `@TestVisible` constant).
```
for (String param : new String[] { null, 'NonExistentType' }) {
    Exception actualException = null;
    try {
       crlpApiSvc.getRollupDefsForRollupType(param);
    } catch (Exception e) {
        actualException = e;
    }
    System.assertNotEquals(null, actualException, 'An exception should be thrown for invalid param: ' + param);
    System.assert(actualException.getMessage().containsIgnoreCase('RollupType'), 'Message should contain "RollupType" param: ' + actualException.getMessage());
}
```

src/classes/CRLP_ApiService.cls
```diff
@@ -0,0 +1,62 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Service methods for the Customizable Rollups API
+*/
+public inherited sharing class CRLP_ApiService {
```
RadSFDeveloper: When the class is viewed in VSCode, the spacing is set to 2 spaces.

src/classes/CRLP_RollupProcessingOptions.cls
```diff
@@ -51,6 +51,24 @@ public class CRLP_RollupProcessingOptions {
         RecurringDonations
     }
 
+    /** @description Map the string name for each of the different rollup types enum vales */
+    public static Map<String, RollupType> rollupTypesEnumByString {
```
RadSFDeveloper: Naming suggestion: `rollupTypeByName`.

src/classes/CRLP_RollupProcessingOptions.cls
```diff
@@ -51,6 +51,24 @@ public class CRLP_RollupProcessingOptions {
...
+        get {
+            if (rollupTypesEnumByString == null) {
+                rollupTypesEnumByString = new Map<String, RollupType>{
+                        RollupType.ContactHardCredit.name().toLowerCase() => RollupType.ContactHardCredit,
```
RadSFDeveloper: Indentation is off.

src/classes/CRLP_ApiService.cls
force2b: I'm just not seeing that in IntelliJ. The file uses 4 space indentations. Even if I look at the raw file in Github it appears as 4 spaces -- https://raw.githubusercontent.com/SalesforceFoundation/Cumulus/aa42a999aa0e55238f9485e2a0b5db11c754f9c1/src/classes/CRLP_ApiService.cls

src/classes/CRLP_ApiService.cls
```diff
@@ -0,0 +1,62 @@
...
+    * @param rollupTypeString String representation of the Rollup Type enum
+    * @return List<CRLP_Rollup>
+    */
+    public List<CRLP_Rollup> getRollupDefsForRollupType(String rollupTypeString) {
+        CRLP_RollupProcessingOptions.RollupType rollupType;
+        try {
+            rollupType = CRLP_RollupProcessingOptions.rollupTypesEnumByString.get((String)rollupTypeString.toLowerCase());
+        } catch (Exception ex) { }
+        if (rollupType == null) {
+            throw new CRLPApiException(String.format( System.Label.CallableApiParameterError, new List<String>{'RollupType'} ));
```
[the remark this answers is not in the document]
force2b: The label used for the error is `Error parsing {0} parameter`, where the intent was to specify which parameter was unable to be parsed. In the particular example, we only have a single parameter to parse, but the next api will support multiple parameters in the `Map<String,Object>` var. I think it's useful for this error to indicate which parameter had an issue.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,185 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @description Unit Tests for the Customizable Rollups Api Service class
+*/
+@IsTest(IsParallel=true)
+private class CRLP_ApiService_TEST {
```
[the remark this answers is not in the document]
force2b: With the exception of the indentations in the `mockRollupDefinitions` method to create the json string, this shows as a 4 character indentation to me, even in the raw file - https://raw.githubusercontent.com/SalesforceFoundation/Cumulus/aa42a999aa0e55238f9485e2a0b5db11c754f9c1/src/classes/CRLP_ApiService_TEST.cls

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,185 @@
...
+    static void validateGetRollupDefsForRollupTypeReturnValues() {
+        // Call the mocking utilities to build a bunch of rollup definitions and filter groups/rules fo
+        mockRollupDefinitions();
+
+        CRLP_ApiService crlpApiSvc = new CRLP_ApiService();
+
+        Test.startTest();
+
+        // These first three have mocked CMT's for rollup definitions
+        List<CRLP_Rollup> rollups = crlpApiSvc.getRollupDefsForRollupType(CRLP_RollupProcessingOptions.RollupType.AccountHardCredit.name());
```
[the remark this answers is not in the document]
force2b: Good point. My original test had different return values for each, but that's not the case any more, so your code makes a lot more sense.

src/classes/CRLP_ApiService_TEST.cls
force2b: Great idea!

force2b:
Thanks fo the review @RadSFDeveloper. Most requested changes were made, with just a couple of comments/notes.

src/classes/CRLP_ApiService.cls
```diff
@@ -0,0 +1,65 @@
...
+
+    public static final String PARAM_ROLLUPTYPE = 'RollupType';
+
+
+    /**
+    * @description Generates the list of Rollup Definitions for the specified rollup type
+    * @param rollupTypeString String representation of the Rollup Type enum
+    * @return List<CRLP_Rollup>
+    */
+    public List<CRLP_Rollup> getRollupDefinitions(String rollupTypeString) {
```
RadSFDeveloper: - Introducing vertical distance between different concepts in the method would benefit readability.
- Let's change input argument not to contain the data type in the name, ie `rollupType` instead of `rollupTypeString`.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,159 @@
...
+            System.assertEquals(2, rollups[0].rollupsMdt.size(), 'There should be two Rollup__mdt records associated with the CRLP_Rollup instance');
+        }
+    }
+
+    /**
+     * @description Validate the getRollupDefsForRollupType returns an exception if the RollupType is null or invalid
+     */
+    @IsTest
+    private static void shouldThrowExceptionWhenParameterIsInvalid() {
+        CRLP_ApiService crlpApiSvc = new CRLP_ApiService();
```
RadSFDeveloper: Naming suggestion to `service` instead of `crlpApiSvc`.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,159 @@
...
+                    'Message should contain "' + CRLP_ApiService.PARAM_ROLLUPTYPE + '" param: ' + actualException.getMessage());
+        }
+    }
+
+    /**
+     * @description Mock the Rollup__mdt records for use within this unit test class
+     */
+    static void mockRollupDefinitions() {
+
+        String filterGroupId1 = CMT_UnitTestData_TEST.getNewRecordId();
```
RadSFDeveloper: - Since we do not have another `filterGroupId` can we remove `1` from the name?
- Does CRLP have a test API  that we can use to extract the `rollupJSON` instead of creating it every time it is required to test rollups? Also, if someone new to rollup functionality and code would need to write a test method to create data, is there a guide on which fields can be used for which rollup definition? Having a CRLP test API would be greatly of benefit.

src/classes/Callable_API.cls
```diff
@@ -59,6 +59,10 @@ global with sharing class Callable_API implements System.Callable {
             } when 'opp.mapstagetostate' {
                 return OPP_StageMappingUtil.determineOppStateFromOppStage((String) params.get('OpportunityStage'));
 
+            } when 'crlp.getrollupdefsforrolluptype' {
```
RadSFDeveloper: Can we sync in external callable API call to the internal naming, ie use `crlp.getrollupdefinitions` instead of `crlp.getrollupdefsforrolluptype`?

src/classes/CRLP_ApiService.cls
force2b: `rollupType` is used already within the method to represent the converted value, which is why I went with `rollupTypeString` to indicate the difference. Drawing a blank for an alternative name though.

src/classes/CRLP_ApiService_TEST.cls
force2b: There isn't an API to create these. The best I have is the `CMT_UnitTestData_TEST.createRollupRecord` method, which as you point out is complex and has a lot of parameters. It would be a lot easier if we could just instantiate `Rollup__mdt` records, but that's not yet supported when there are Entity Relationship fields on the CMT object.

What would be a lot more readable is to rebuild the `CMT_UnitTestData_TEST.createRollupRecord` logic as more of a class structure:

```
List<Rollup__mdt> = new List<Rollup__mdt>{
   CMT_UnitTestData_TEST.newRollup('Total Donations')
     .withSummaryField('Contact.npo02__TotalOppAmount__c')
     .withFilterGroupId(filterGroupId)
     .withDetailField('Opportunity.Amount')
     .withAmountField('Opportunity.Amount')
     .withDateField('Opportunity.CloseDate')
     .withOperation(CRLP_Operation.RollupType.Sum)
     .withTimeFrame(CRLP_Operation.TimeBoundOperationType.All_Time)
     .getMdtRecord()
};
```

I wouldn't mind taking that on, but I'm not sure I've got the bandwidth to redo that (and all the classes that use it) right now

src/classes/Callable_API.cls
force2b: I like that it specifies what it getting the definitions for, but `rollups` is repeated in there twice. What do you think about `crlp.getrollupdefinitionsfortype` or `crlp.getdefinitionsforrolluptype`?

src/classes/CRLP_ApiService.cls
RadSFDeveloper: How about `rollupTypeName`?

src/classes/Callable_API.cls
RadSFDeveloper: Either new naming works fine. Good point on keeping the rollup type in the action name. This is also a good example when external action naming differs from internal method name and both are descriptive for their usage.

src/classes/CRLP_ApiService_TEST.cls
RadSFDeveloper: Thanks for the feedback. Perhaps the best action is to create a new WI to check into building a CRLP (or any specific domain) test API that we can call from our other test classes and keep the high level code in unit tests.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,170 @@
...
+        mockRollupDefinitions();
+
+        UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
+                Customizable_Rollups_Enabled__c = true
+        ));
+
+        CRLP_ApiService service = new CRLP_ApiService();
+
+        Test.startTest();
+        System.debug(CRLP_Rollup_SEL.cachedRollups.size());
```
RadSFDeveloper: Let's remove `System.debug`.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * @description Validate the getDefinitionsForRollupType returns the expected collection of CRLP_Rollup instances.
+     * Calls the mocking utilities to build a bunch of rollup definitions and filter groups/rules fo
+     */
+    @IsTest
+    private static void shouldReturnRollupDefinitionsWhenParameterIsValid() {
+        mockRollupDefinitions();
+
+        UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
+                Customizable_Rollups_Enabled__c = true
```
RadSFDeveloper: Indentation is off.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * @description Validate the getDefinitionsForRollupType returns an exception if the RollupType is null or invalid
+     */
+    @IsTest
+    private static void shouldThrowExceptionWhenParameterIsInvalid() {
+        CRLP_ApiService service = new CRLP_ApiService();
+
+        System.assert(service.getRollupDefinitions(null).isEmpty(),
```
RadSFDeveloper: Let's separate testing of rollup definition retrieval when CRLP is disabled into another unit test.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,181 @@
...
+        Test.startTest();
+        for (String param : new String[] { null, 'NonExistentType' }) {
+            Exception actualException = null;
+            try {
+                service.getRollupDefinitions(param);
+            } catch (Exception e) {
+                actualException = e;
+            }
+            System.assertNotEquals(null, actualException,
+                    'An exception should be thrown for invalid param: ' + param);
```
RadSFDeveloper: Indentation is off.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * @description Validate the getRollupDefinitions returns an empty list<> when CRLP is not enabled
+     */
+    @IsTest
+    private static void shouldThrowReturnEmptyListWhenCRLPNotEnabled() {
+        CRLP_ApiService service = new CRLP_ApiService();
+
+        System.assert(service.getRollupDefinitions(null).isEmpty(),
+                'When CRLP is disabled, then call to getRollupDefinitions should return an empty list');
```
RadSFDeveloper: Indentation is off.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -0,0 +1,181 @@
...
+            System.assert(actualException.getMessage().containsIgnoreCase(CRLP_ApiService.PARAM_ROLLUPTYPE),
+                    'Message should contain "' + CRLP_ApiService.PARAM_ROLLUPTYPE + '" param: ' + actualException.getMessage());
+        }
+    }
+
+    /**
+     * @description Validate the getRollupDefinitions returns an empty list<> when CRLP is not enabled
+     */
+    @IsTest
+    private static void shouldThrowReturnEmptyListWhenCRLPNotEnabled() {
```
RadSFDeveloper: `shouldReturnEmptyListWhenCRLPNotEnabled()` (remove "Throw").

src/classes/CRLP_ApiService_TEST.cls
force2b: This is the default indentation for a second line (at least in IntelliJ). The style guide does not specify the exact indentation for continued lines. I think we should leave this up to the developer - as long as it meets the readability requirements.

RadSFDeveloper:
@force2b , changes look good. This is ready for testing.

force2b:
@rponti-sforg This story is ready for QA. Instructions on testing are in the WI description:  W-031334

PR #4104, opened 2019-03-07
Call correct rollup process for Individual Accounts

jjbennett:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/RLLP_OppRollup.cls
```diff
@@ -80,109 +91,199 @@ public without sharing class RLLP_OppRollup {
...
+        }
+
+        if (contactById.isEmpty()) {
+            return;
+        }
+
+        RLLP_OppRollup oppRollup = new RLLP_OppRollup();
+        oppRollup.rollupContacts(contactById);
+
+        if (!householdById.isEmpty()) {
```
force2b: Might be useful to add a comment here for:
```
// Rollups to the old Household object use legacy rollups only.
```

src/classes/RLLP_OppRollup.cls
```diff
@@ -80,109 +91,199 @@ public without sharing class RLLP_OppRollup {
...
+     * @param accountIds Ids for the accounts that require rollups to be recalculated
+     */
+    public void rollupAccounts(Set<Id> accountIds) {
+        if (accountIds == null || accountIds.isEmpty()) {
+            return;
+        }
+
+        if (isCustomizableRollupsEnabled) {
+            runCustomizableRollups(accountIds);
+
```
force2b: Is the blank line needed here?

src/classes/RLLP_OppRollup.cls
```diff
@@ -80,109 +91,199 @@ public without sharing class RLLP_OppRollup {
...
+    /**
+     * @description Calls the future or immediate customizable rollups depending on current context
+     * @param  ids The ids of the sObjects that require rollups to be recalculated
+     */
+    private void runCustomizableRollups(Set<Id> ids){
+        if (isImmediate(ids)) {
+            CRLP_RollupQueueable.runRollupsForIds(new List<Id>(ids));
+
+        } else {
+            System.enqueueJob(new CRLP_RollupQueueable(new List<Id>(ids)));
```
force2b: Is there a possibility that this method will be called twice when `ids.size() > 1`? If so, that will throw an exception when `enqueueJob` is called the second time. Reading through the code I don't believe this is possible, but I just want to make sure.

src/classes/RLLP_OppRollup.cls
```diff
@@ -80,109 +91,199 @@ public without sharing class RLLP_OppRollup {
...
+        }
+    }
+
+    /**
+     * @description Determines if the context is already async or if there is only 1 record to update
+     * @param ids The set of ids that need their rollups refreshed
+     * @return Boolean - true when the rollups should happen immediately
+     */
+    private Boolean isImmediate(Set<Id> ids) {
```
force2b: How about `canRunImmediate` for this?

src/classes/RLLP_OppRollup.cls
```diff
@@ -80,109 +91,199 @@ public without sharing class RLLP_OppRollup {
...
         // use the contact roles to find the contacts and households
-        RLLP_OppRollup_UTIL oppRollupUtil = new RLLP_OppRollup_UTIL();
-        for (OpportunityContactRole r : Database.query(oppRollupUtil.buildOpportunityContactRoleQuery() +
-                                            ' where Opportunity.Id IN :OppIds and isPrimary = true ALL ROWS ')){   
-                    
-            cmap.put(r.ContactId, r.Contact);
-                
-            if (r.Contact.npo02__Household__c != null){
-                hhmap.put(r.Contact.npo02__Household__c, r.Contact.npo02__Household__r);
+        String query = new RLLP_OppRollup_UTIL().buildOpportunityContactRoleQuery() +
```
force2b: I'm not necessarily sure this is going to matter 99.9% of the time, but it's worth noting that there is a difference in how legacy rollups identify "Primary" for the Contact Hard Credit rollups:
* Legacy Rollups use the `OpportunityContactRole.IsPrimary` field to determine which Contact gets hard credit. If there are dupe OCR"s with IsPrimary=True (which is technically possible) this could lead to two Contacts getting Hard Credit for the same OCR. There's a fixup job in NPSP Settings to address the dupe OCR thing.
* Customizable Rollups use the `Opportunity.Primary_Contact__c` field to determine which Contacts get Hard Credit.

So, for this job to determine which Contacts will get Hard Credit (and thus need a recalc) by querying the OCR for specific Opportunities could - though extremely unlikely - lead to a situation where the actual Primary Contact (the one specified by `Opportunity.Primary_Contact__c`) is not recalculated. This won't mis-calculate amounts because Customizable Rollups will still use the correct logic to make the Hard Credit determination. However, if for some reason the Contact identified by `Opportunity.Primary_Contact__c` does not have a matching OpportunityContactRole record with IsPrimary set to true, then that PrimaryContact would have their values recalculated by this job. Again, this is extremely unlikely and even if it does happen the Contact would get recalculated in the nightly job.

src/classes/RLLP_OppRollup_TEST.cls
```diff
@@ -1094,4 +1094,27 @@ public with sharing class RLLP_OppRollup_TEST {
             System.AssertEquals ( 0 , UpdatedCon.npo02__OppAmount2YearsAgo__c );
     }
 
+    @IsTest
```
force2b: ApexDoc comments missing for these

src/classes/RLLP_OppRollup.cls
jjbennett: This change was strictly formatting and should not have changed any logic. Anything related to household rollups should be handled in a separate work item.

src/classes/RLLP_OppRollup.cls
jjbennett: The only possibility would be once for account and then once for contact but I'm assuming it'd only result in an error if it was the same object type?

src/classes/RLLP_OppRollup.cls
jjbennett: Going with should since it technically always can :)

src/classes/RLLP_OppRollup.cls
force2b: It's not the object that would be an issue, but attempting to call `System.enqueueJob()` twice that would throw an exception.

jjbennett: @force2b - I added a unit test for this scenario and I ran a test in execute anonymous after loading test data, I am unable to get an exception:
`Set<Id> accountIds = new map<id, account>([select id from account]).keyset();
Set<Id> contactIds = new map<id, contact>([select id from contact]).keyset();

RLLP_OppRollup oppRollup = new RLLP_OppRollup();

oppRollup.rollupAccounts(accountIds);
oppRollup.rollupContacts(contactIds);`

force2b: You're right. There definitely _was_ a limit at some point, but according to the docs: `You can add up to 50 jobs to the queue with System.enqueueJob in a single transaction.`

jjbennett:
@force2b - is this ready for testing then?

src/classes/RLLP_OppRollup.cls
force2b: @jjbennett What do you think about adding this comment?

jjbennett:
**lurch: attach W-031429

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008Ll0VQAS" target="_blank">W-031429</a>

force2b:
Thanks @jjbennett. Agreed on the merge. I'm still going to attempt to debug this to figure out why that one unit test is passing.

jjbennett:
@rponti-sforg - this is part 2 of this work item and should resolve the need to go and manually run the bulk rollups. Thank you!
P.S. This is ok to merge into the parent branch but we are still holding off on merging the parent branch into master.

rponti-sforg:
> @rponti-sforg - this is part 2 of this work item and should resolve the need to go and manually run the bulk rollups. Thank you!
> P.S. This is ok to merge into the parent branch but we are still holding off on merging the parent branch into master.

@jjbennett Thank you.  This one has a robot feature test that failed.  I will let you review that and then I will merge once you give the all clear.

davisagli:
@jjbennett oh we renamed that "Populate Address" robot keyword. It's now "Populate Field By Placeholder"

jjbennett:
@rponti-sforg - thanks, David pointed me to the fix for that robot test. Also, just do double check, were you able to confirm the fix?

rponti-sforg:
> @rponti-sforg - thanks, David pointed me to the fix for that robot test. Also, just do double check, were you able to confirm the fix?

@jjbennett Testing is currently underway for feature/jen-manual-acct-crlp-enabled__fix.  Post an update here and on W-031429 when testing is complete.   Should be finished testing in next 20-30 min.

rponti-sforg:

For further testing details, see https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000008Ll0VQAS/view

Merging this PR into parent branch.  Deleting child branch.

sam-knox:
@jjbennett - Is there any release note needed here?

jjbennett:
@sam-knox - no this was a child branch going into a parent branch that is not yet going into master. Thanks for checking!

PR #4105, opened 2019-03-07
Feature/pt bdi support checkbox fields

ptewson-sfdo:
Bring more responsible support for mapping to checkbox fields to BDI/BGE. Mapped field on 
Data Import must be picklist with None/True/False or text field for full support, but current user processes that use checkbox-to-checkbox mappings will be unaffected by this change.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_DataImportService.cls
```diff
@@ -878,6 +878,16 @@ global with sharing class BDI_DataImportService {
                 return false;
             }
 
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - None, True, False - or a text with either 
+            // blank, "True", or "False"
+            String destinationSObjectname = destinationRecord.getSObjectType().getDescribe().getName(); 
+            if ((UTIL_Describe.getFieldDisplayType(destinationSObjectname, destinationField) ==
```
ptewson-sfdo: Field name must be LOWERCASE. Also namespace support.

src/classes/BDI_DataImportService.cls
```diff
@@ -878,6 +878,18 @@ global with sharing class BDI_DataImportService {
                 return false;
             }
 
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - None, True, False - or a text with either 
+            // blank, "True", or "False"
+            String destinationSObjectname = destinationRecord.getSObjectType().getDescribe().getName(); 
+            if (!(value instanceof Boolean) &&
```
kslewis: I know were nesting a few ifs here, but I might consider adding one more and putting the `destinationSObjectname` var inside the first `value instanceof Boolean` if block, since we don't need to get the SObject name to inspect this unless the field is Boolean.

src/classes/BDI_DataImportService.cls
```diff
@@ -878,6 +878,18 @@ global with sharing class BDI_DataImportService {
                 return false;
             }
 
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - None, True, False - or a text with either
```
kslewis: Will this allow for mapping from picklists with any values?  if so we might not want to specifify "None" here, since admins can configure whatever PL values they want.

src/classes/BDI_DataImportService.cls
```diff
@@ -878,6 +878,18 @@ global with sharing class BDI_DataImportService {
...
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - None, True, False - or a text with either 
+            // blank, "True", or "False"
+            String destinationSObjectname = destinationRecord.getSObjectType().getDescribe().getName(); 
+            if (!(value instanceof Boolean) && 
+                (UTIL_Describe.getFieldDisplayType(UTIL_Namespace.StrTokenNSPrefix(destinationSObjectname), 
+                 UTIL_Namespace.StrTokenNSPrefix(destinationField.toLowerCase())) == 
+                Schema.DisplayType.BOOLEAN)) {
+
+                value = Boolean.valueOf(value);
```
kslewis: While you are inside this method, it might be nice to rename the `value` var here to something like `sourceField` but just my preference 😄 .

src/classes/BDI_DataImportService.cls
ptewson-sfdo: None just means null or no value ... it's important for the behavior but I've been struggling with how to type that to make it clear that it's not an actual picklist option with text "none" but that there IS a valid option in the picklist that is neither true or false and evaluates to null

ptewson-sfdo: Maybe I should just say "null" ... 🤔

ptewson-sfdo: And if the PL values aren't parse-able to a boolean value with Boolean.valueOf() it will fail

src/classes/BDI_DataImportService.cls
kslewis: nm - I realize why it is named this way now!

src/classes/BDI_DataImportService.cls
kslewis: I see!  Ok, yeah, very cool.

src/classes/BDI_DataImportService.cls
kslewis: Another option could be to extract the comparison to a little utility method, for readability - something like `if(booleanMappingIsValid(value, destinationSObjectname)`.  This is good - I'm just trying to think through it if it could be a little more readable what with multiple calls to those UTIL classes here.

cpose:
Lurch: W-031084

src/classes/BDI_DataImportService.cls
```diff
@@ -878,13 +878,44 @@ global with sharing class BDI_DataImportService {
                 return false;
             }
 
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - null, True, False - or a text with either 
+            // null, "True", or "False"
+            if (!(value instanceof Boolean) && 
+                fieldIsBooleanDisplayType(destinationRecord, destinationField)) {
+
```
kenlewis-sfdo: 👍 much easier to read!!

PR #4106, opened 2019-03-07
Update Advancement Namespace default value

bdvorachek:
- We've updated an internal setting to support a namespace change for our sister product, GEM. This should have no impact on NPSP functionality.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/objects/npe01__Contacts_And_Orgs_Settings__c.object
+++ b/src/objects/npe01__Contacts_And_Orgs_Settings__c.object
@@ -2,7 +2,7 @@
 <CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
     <fields>
         <fullName>Advancement_Namespace__c</fullName>
-        <defaultValue>&quot;adv1&quot;</defaultValue>
+        <defaultValue>&quot;gem0&quot;</defaultValue>
         <description>For system use only: Namespace for cross package functionality.</description>
         <externalId>false</externalId>
         <inlineHelpText>For system use only: Namespace for cross package functionality.</inlineHelpText>
--- a/src/classes/BDI_DataImport_TEST2.cls
+++ b/src/classes/BDI_DataImport_TEST2.cls
@@ -561,6 +561,9 @@ private with sharing class BDI_DataImport_TEST2 {
     *    all contacts matched
     **********************************************************************************************************/            
     static testMethod void ManyDIUniqueIdNotMatchingExistingContacts() {
+        //skip the test if Advancement is installed
+        if(ADV_PackageInfo_SVC.useAdv()) return;
+        
         //Since Data Import API does not have other non-encryptable fields to test matching on,
         //we need to exit the test if the matching is done on the encrypted field. 
         if (!Schema.sObjectType.Contact.fields.Title.isFilterable()) return;


bdvorachek:
Hello NPSP team, this change should not affect the behavior of anything in Cumulus unless GEM is installed in the target org.

Quip doc explaining some of the reasoning behind this change: https://salesforce.quip.com/R3wkAbQGa5Xf

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
```diff
@@ -2,7 +2,7 @@
 <CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
     <fields>
         <fullName>Advancement_Namespace__c</fullName>
-        <defaultValue>&quot;adv1&quot;</defaultValue>
```
force2b: The change itself is fine, however any new installs of NPSP since this field was first implemented would have a value in the field of "adv1". Is that an issue, or is it assumed that existing NPSP customers (even ones that were spun up since the field was introduced) would not likely be GEM users as well?

force2b: @bdvorachek Just read through the quip doc and you're already considering that the field value will be "adv1" for current NPSP customers. Approving. Is this going to QA?

bdvorachek: We know who all has the GEM pilot installed, this shouldn't break any existing installs since they'll have the "adv1" namespaced GEM installed still. New pilot installs that happen in the window between this being released and GEM being updated are the only risk here, and we should know about those too.

I'm not sure what the QA plan for this would be, since we need to also make a change in GEM to support this.

bdvorachek: @annejackson

annejackson: @force2b - for this change, are there certain things you'd like to see tested in a Cumulus scratch org? We can do some testing in GEM with the beta_dep flow after merging this. Please let me know if you have some areas of concern (either testing functionality in the UI like making sure logging a donation works fine, or checking something in custom settings, or if it's along the lines of just doing some tests via dev console)? CC: @bdvorachek

lparrott:
Sorry to crash the party, but we also need to skip a Data Import test when GEM is installed, here are the details about why (if anyone is curious): https://salesforce.quip.com/nclIAUenLPyj

@force2b I just pushed the change if you want to take a look, it's the same pattern we've used for other tests that fail with HEDA installed.

src/objects/npe01__Contacts_And_Orgs_Settings__c.object
force2b: I was more curious than anything @annejackson. I wasn't sure if this needed to go through some tests in an org that had a previous value in that field (though I suppose that wouldn't be possible for unmanaged testing). I guess if you really wanted to test the variations that  @bdvorachek mentioned it could be done post-merge using a beta package.

bdvorachek:
**lurch: attach W-028785

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fOF1AAM" target="_blank">W-028785</a>

lparrott:
@force2b Not yet, we started a conversation about possibly skipping some tests via cci instead of requiring a code change, but for now we'll keep an eye on the "beta dependency" builds to catch test failures as soon as we can.

annejackson:
Hey @smarovich and @markross-sfdo, 
This change is ready to be merged as the real testing will have in a GEM beta_dep org. Please see the Changes section in this PR and let us know what you think. @force2b feel free to weigh in on the changes notes as well. There should be no action needed from NPSP users.

lparrott:
**lurch: detach

LurchTheButler:
Detached <a href="https://foundation.my.salesforce.com/a2x80000007fOF1AAM" target="_blank">W-028785</a>

annejackson:
**lurch: attach W-031792

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LlThQAK" target="_blank">W-031792</a>

annejackson:
Merging PR.

PR #4107, opened 2019-03-08
Tag a test that is frequently failing as unstable

davisagli:
@skristem and @boakley, @force2b pointed out that this test is frequently failing. I looked at it for a few minutes and did not immediately understand what is wrong (deleting the contact manually appears to work), so I think we should at least mark it as unstable so it's not breaking builds.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


skristem:
yes, I noticed this. It seems like the account name takes some time to get updated. I am modifying this test little bit to verify the account name. For now we can mark it as unstable

PR #4111, opened 2019-03-08
Add configuration for MetaDeploy installation plans

davisagli:
This adds configuration for publishing two MetaDeploy installation plans:
- `existing_org` (equivalent to the existing installer, installs NPSP but does not configure)
- `new_org` (equivalent to running the `install_prod` flow, installs NPSP with trial configuration)

Note: Using this configuration requires changes to the metadeploy publish task which are pending in https://github.com/SFDO-Tooling/CumulusCI/pull/1020

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -370,6 +370,9 @@ plans:
                 task: install_managed
             3:
                 task: deploy_post
+                ui_options:
+                    first:
+                        name: NPSP Config for Salesforce Mobile App
     new_org:
         slug: trial
         title: Install NPSP Trial
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -352,3 +352,40 @@ orgs:
         enterprise:
             config_file: orgs/enterprise.json
 
+plans:
+    existing_org:
+        slug: install
+        title: Install NPSP
+        tier: primary
+        steps:
+            1:
+                flow: dependencies
+                ui_options:
+                    deploy_pre:
+                        account_record_types:
+                            name: Account Record Types
+                        opportunity_record_types:
+                            name: Opportunity Record Types
+            2:
+                task: install_managed
+            3:
+                task: deploy_post
+    new_org:
+        slug: trial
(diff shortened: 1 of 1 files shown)


cumulusci.yml
```diff
@@ -352,3 +352,40 @@ orgs:
...
+                ui_options:
+                    deploy_pre:
+                        account_record_types:
+                            name: Account Record Types
+                        opportunity_record_types:
+                            name: Opportunity Record Types
+            2:
+                task: install_managed
+            3:
+                task: deploy_post
```
davisagli: @jlantz what user-facing name should we give this step? The current installer has `NPSP Config for Salesforce1` which is pretty ambiguous.

jlantz: NPSP Config for Salesforce Mobile App?

Might want to inquire with the NPSP dev team to get their thoughts though.

davisagli: Ohh.... (lightbulb goes on) ... those components are focused the mobile experience, aren't they. Now I understand the existing name better.

cumulusci.yml
```diff
@@ -352,3 +352,40 @@ orgs:
...
+        slug: install
+        title: Install NPSP
+        tier: primary
+        steps:
+            1:
+                flow: dependencies
+                ui_options:
+                    deploy_pre:
+                        account_record_types:
+                            name: Account Record Types
```
jlantz: We should make these steps recommended.  Sometimes users have issues with those steps but want to deploy everything else.

davisagli: @jlantz Are you talking about the issues with deploying the NPSP_Default opportunity record type if the org doesn't have the right set of Stage picklist values? Or other issues too? We have a plan for how to avoid the former and I'd rather do that than make the step optional (leaving the box could also have issues if opportunity record types aren't enabled, right?)

PR #4112, opened 2019-03-10
Customizable Rollups: Chunk mode processing for Account/Contact Hard & Soft Credits

kenlewis-sfdo:
- Adds feature for organizations with millions of Account and Contact records to run Customizable Rollups as smaller jobs. The rollups will also restart themselves if there are more records to process.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


kslewis:
The one significant change in this PR is the relocation of the logic to build the start query from inside the start method in the `CRLP_Batch_Base_NonSkew` class to an inner class called `BatchQueryBuilder` in the `CRLP_Batch_Base` class.  This allows for more unit-like testing against the query string, since we can now build and inspect it from within the test methods.  There is some significant refactoring of methods, reduction of code, etc. also, but the main functionality is to add the "ORDER BY", "LIMIT", and "Id > <Id>" clauses to the query string when appropriate.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -118,6 +122,14 @@ public abstract class CRLP_Batch_Base {
         this.recordIds = recordIds;
     }
 
+    /*******************************************************************************************************
+     * @description Constructs the class - Rollup is for a specific set of Summary Object Id's
+     */
+    public CRLP_Batch_Base(CRLP_RollupProcessingOptions.RollupType jobType) {
+        this();
+        this.jobType = jobType;
```
force2b: I noticed that on line 101 above, it passes `this.jobType` into the `CRLP_RollupBatch_SVC.getSummaryObjectType` method. However, in this constructor the `this.JobType` var isn't defined until after that call is made.

src/classes/CRLP_RollupBatch_SVC.cls
```diff
@@ -176,42 +176,63 @@ public class CRLP_RollupBatch_SVC {
...
-            } else if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
-               database.executeBatch(new CRLP_AccountSkew_SoftCredit_BATCH(recordIds, lastIdProcessed), batchSize);
-            } else if (jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
-               database.executeBatch(new CRLP_ContactSkew_SoftCredit_BATCH(recordIds, lastIdProcessed), batchSize);
-            } else if (jobType == CRLP_RollupProcessingOptions.RollupType.GAU) {
-               database.executeBatch(new CRLP_GAU_BATCH(recordIds, lastIdProcessed), batchSize);
-            } else if (jobType == CRLP_RollupProcessingOptions.RollupType.RecurringDonations) {
-               database.executeBatch(new CRLP_RDSkew_BATCH(recordIds, lastIdProcessed), batchSize);
+
+        switch on (jobMode) {
```
force2b: So much easier to read now!

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -913,6 +913,11 @@ public without sharing class UTIL_CustomSettingsFacade {
         customizableRollupSettings.Rollups_IncrementalLastNDays_FldOverride__c = mySettings.Rollups_IncrementalLastNDays_FldOverride__c;
         customizableRollupSettings.Rollups_IncrementalLastNDays_ValOverride__c = mySettings.Rollups_IncrementalLastNDays_ValOverride__c;
 
+        // Default values
+        if (customizableRollupSettings.Rollups_Limit_on_Attached_Opps_for_Skew__c == null) {
```
force2b: Was this needed for something?

src/objects/Customizable_Rollup_Settings__c.object
```diff
@@ -199,6 +199,45 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Rollups_Contact_HardCredit_ChunkSize__c</fullName>
+        <description>Only used by the new Customizable Rollups engine in LDV Orgs and only when required for LDV orgs</description>
```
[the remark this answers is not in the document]
smarovich: I would recommend giving a bit more detail as to what this is for. Can we combine the description and help text and just use the same value for both. "LDV ORGS ONLY: If chunking is needed to avoid a Query Timeout in the batch job, specifies the number of Contacts processed in each batch. Used by the Customizable Rollups engine."

src/classes/CRLP_Batch_Base.cls
```diff
@@ -335,13 +347,126 @@ public abstract class CRLP_Batch_Base {
...
+            this.chunkSize = chunkSize;
+            this.fieldsToCheckForNonZero = fieldsToCheckForNonZero;
+        }
+
+        /**
+         * @description Uses the local and base class variables to construct the batch class
+         * start query.
+         * @return The query string.
+         */
+        String buildQuery() {
```
force2b: I really like the new implementation that uses a class to build out the batch job start query. My only comment is that this logic is unique to NonSkew mode. Long term, I'd like to see the entire start() method of the `CRLP_Batch_Base_Skew` class refactored the same way you did the `CRLP_Batch_Base_NonSkew.start()` method. However, I'm not going to push that for this release.

What I would say is that this class does belong here or possibly be abstracted out into its own class, however I think the methods should be refactored a bit to plan for the eventual implementation of a SkewMode Batch Query Builder. The `buildQuery` method already has a simple switch that splits between the two modes. What I would do is abstract this out further so that all the logic in the `when NonSkewMode` block is in a method or methods of its own. This way the `buildQuery` method is small and easy to follow.

It's also possible that some logic can be raised up or down a level, such as checking for `if (base.hasActiveRollupsForThisJob() == false)` to determine if the query should even return any records at all.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -335,13 +347,126 @@ public abstract class CRLP_Batch_Base {
...
+                }
+            }
+        }
+    }
+
     /**
      * @description Produces the where clause that should be used for the inner join
      * @return The WHERE clause, as a string, that should be used to filter the detail records
      */
     @TestVisible protected String getMainQueryInnerJoinFilter() {
```
force2b: Is there a need to have these methods outside of the `BatchQueryBuilder` inner class?

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -36,6 +36,13 @@
 */
 public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
 
+    /*******************************************************************************************************
+    * @description When Chunking mode is used for an LDV org, this is the last Id retrieved
+    * in the previous batch. When the job is chained, it'll query records with Ids greater than this Id.
+    */
+    @TestVisible
+    private Id chunkModeLastProcessedId;
```
force2b: What would you think of `lastIdProcessed` for this, and then not check for `isChunkEnabled` on line 151 before assigning to this var?

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -75,6 +87,24 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
         super(jobType, jobFilter);
     }
 
+    /*******************************************************************************************************
+     * @description Constructs the class - Rollup is for a specific set of Summary Object Id's
+     * @param jobType Defines the type of rollup job
+     * @param recordIds Specific set of parent record id's to roll up to
+     */
+    public CRLP_Batch_Base_NonSkew(CRLP_RollupProcessingOptions.RollupType jobType, Id lastIdProcessedForChunking) {
```
force2b: Both var names could be `lastIdProcessed`.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -93,45 +123,11 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
-            if (!whereClauses.isEmpty()) {
-                query += ' WHERE ' + String.join(whereClauses, ' AND ');
-            }
-
-            // If there are no active rollups for the object type, then force the query to return zero records.
-            if (hasActiveRollupsForThisJob() == false) {
-                // This is needed because can't return null as a valid iterator.
-                query += ' LIMIT 0';
-            }
+            String query = getBatchStartQuery(
```
force2b: So much easier to read. Great work abstracting this out.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -298,6 +304,31 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
             );
             insert logEntry;
         }
+
+        // If chunk mode is enabled and we (probably) have more records to process,
+        // submit another batchable job
+        if (isChunkModeEnabled && batchIteration > 1) {
```
force2b: There's a train of thought on handling this a bit differently. Curious what you think. Rather than just restart the batch for the next chunk if more than one child batch was processed, the finish() method would query for the next Id to determine if there is one.

`SELECT Count() FROM Contact WHERE Id > :lastIdProcessed LIMIT 1`

If that returns back a number > 0, then the job is resubmitted. Otherwise it's not. The thought is that this would avoid queuing a new job that will not select any records. In the end the overall work still finishes at the same time, but this avoids the extra job by sacrificing a query. 

My thought is that it might work. `Id` is indexed, and the `LIMIT 1` clause will force the query to be quick.

What do you think?

src/objects/Customizable_Rollup_Settings__c.object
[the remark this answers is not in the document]
force2b: Thanks @smarovich. How about the following?
* Field Name: `LDV: Contact Hard Credit Chunk Size`
* API Name: `ChunkSize_Contact_HardCredit__c`
* Inline Help Text: `FOR LDV ORGS ONLY WITH SFDO GUIDANCE. To avoid a Query Timeout in the rollup batch job in orgs with many millions of Contacts, this specifies the maximum number of Contacts processed in each batch.`

src/classes/CRLP_Batch_Base.cls
kslewis: We could consider moving the BatchQueryBuilder out to its own top-level class, but I kind of like storing within the base class itself.  Just kind of feels like the right place to me, all self-contained here instead of in another file/class.  I added a specific method for `buildQueryForNonSkewJob` so that later on we can add a similar one for Skew jobs.

src/classes/CRLP_Batch_Base.cls
kslewis: Nope, good call - moved them inside!  Let me know what you think.

src/classes/CRLP_Batch_Base_NonSkew.cls
kslewis: I like the name change - updated!  We need the `isChunkModeEnabled`  check on 145 because in the first `execute` method in the first batch job, `lastProcessedId` will be null here, so in that specific context we can only know if we are in chunk mode by checking that the custom setting value is configured (and if its not configured than we don't need to store the Id).

src/classes/CRLP_Batch_Base_NonSkew.cls
kslewis: I like it - updated!

src/objects/Customizable_Rollup_Settings__c.object
```diff
@@ -200,11 +200,11 @@
...
     <fields>
-        <fullName>Rollups_Contact_HardCredit_ChunkSize__c</fullName>
+        <fullName>ChunkSize_Contact_HardCredit__c</fullName>
         <description>Only used by the new Customizable Rollups engine in LDV Orgs and only when required for LDV orgs</description>
         <externalId>false</externalId>
-        <inlineHelpText>LDV ORGS ONLY: If chunking is needed to avoid a Query Timeout in the batch job, this specifies the number of Contacts processed in each batch.</inlineHelpText>
-        <label>Contact Hard Credit Chunk Size</label>
+        <inlineHelpText>FOR LDV ORGS ONLY WITH SFDO GUIDANCE. To avoid a Query Timeout in the rollup batch job in orgs with many millions of Contacts, this specifies the maximum number of Contacts processed in each batch.</inlineHelpText>
+        <label>LDV: Contact Hard Credit Chunk Size</label>
         <precision>7</precision>
```
force2b: @kslewis One more change. The precision on these three fields be set to 8 instead of 7 to allow for a number over 10,000,000.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -94,40 +94,81 @@ public abstract class CRLP_Batch_Base {
...
+
         this.jobType = jobType;
+        if (jobFilter != null) {
+            this.jobFilter = jobFilter;
+        }
+        this.recordIds = recordIds;
+
+        this.summaryObjectType = CRLP_RollupBatch_SVC.getSummaryObjectType(this.jobType);
+        this.statefulCacheOfRollupsToBeProcessed = CRLP_Rollup_SEL.getRollupDefinitionsToUse(this.jobType, this.jobFilter);
+        this.isScheduledJob = System.isScheduled();
```
force2b: @kslewis Have you verified that `System.isScheduled();` still returns true here when the jobs are executed from the scheduled context (since it's being set from a different location)?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -350,19 +353,22 @@ public abstract class CRLP_Batch_Base {
...
+     * accessible to the inner class. Used in the start method of the CRLP_Batch_Base_NonSkew class.
+     * @param lastIdProcessed The last processed Id from the previous chunk.
      * @param chunkSize The size of each chunk - used in the LIMIT statement in the query.
      * @param fieldsToCheckForNonZero A list of fields that were checked for non-zero values,
      * used in the query.
      * @return The query string to be used by the batch start method.
      */
     @TestVisible
-    protected String getBatchStartQuery(Id chunkLastId, Integer chunkSize, List<String> fieldsToCheckForNonZero) {
+    protected String getBatchStartQuery(Id lastIdProcessed, Integer chunkSize, List<String> fieldsToCheckForNonZero) {
```
force2b: @kslewis  Since `fieldsToCheckForNonZero` is a specific name that describes the usage (which is not necessarily known in this wrapper class), how about changing the var name here (and in the inner class) to `addlFieldsToQuery`.

force2b: Also, does this need to pass in `chunkSize` as a parameter or is thing something the wrapper class could determine on its own using the current job type?

force2b:
Thanks for the updates @kslewis. A few more comments and suggestions added.

src/classes/CRLP_Batch_Base.cls
```diff
@@ -335,224 +350,348 @@ public abstract class CRLP_Batch_Base {
...
 
-            when RecurringDonations {
-                whereClause = 'Id IN (SELECT npe03__Recurring_Donation__c FROM Opportunity)';
+            // If there are no active rollups for the object type, then force the query to return zero records.
+            if (base.hasActiveRollupsForThisJob() == false) {
+                // This is needed because can't return null as a valid iterator.
+                query += ' LIMIT 0';
+            } else if (base.recordIds == null) {
+                if (chunkSize != null) {
+                    // In an LDV Org, there may be too many records to query in the main query. Core support has recommended
```
ptewson-sfdo: Super minor comment, but can we keep these comments on one line?

kslewis: I'd say this is safe to remove now anyway.  I removed it but Mike let me know if we should keep it.

src/classes/CRLP_Batch_Base.cls
kslewis: Yes, it wouldn't need to be passed in, but I'm also using the constructor param to pass in a limit for the query that checks for additional records in `hasAdditionalRecordsToProcess`. 

I renamed the fieldsToCheckForNonZero variable!

src/classes/CRLP_Batch_Base.cls
kslewis: I'll try to schedule and test in a scratch org but it should still be set to true since it is still called from within the schedulable class execute method.

src/objects/Customizable_Rollup_Settings__c.object
kslewis: Ok, updated!

src/classes/CRLP_Batch_Base.cls
force2b: I do like the comments just to explain the "why" behind chunk mode, but it's also not critical to be there. What I might suggest is to put this particular comment (that explains the why) into the apexDoc comment of the `isChunkModeEnabled` method.

sam-knox:
@melissabarber - I think this one needs a release note written for it.

force2b:
@sam-knox and @melissabarber Possibly, though keep in mind that we're not including the Chunking Mode settings fields in the NPSP Settings UI. This new feature should only be enabled by customers after consulting with CCE to address specific performance needs in the largest customers.

sam-knox:
@force2b - Ok, if there's no setting available in this release then perhaps we don't need to call it out, especially given that it's for a limited audience at this point.

melissabarber:
@force2b & @sam-knox I say we keep this one out of this release. This is one Sarah mentioned before she was OOO that shouldn't be included in these release notes.

PR #4116, opened 2019-03-12
Feature/robot  automation cleanup

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/tests/NPSP.py
+++ b/tests/NPSP.py
@@ -852,7 +852,7 @@ class NPSP(object):
     def click_link_with_text(self, text):
         self.builtin.log("This test is using the 'Click link with text' workaround", "WARN")
         element = self.selenium.driver.find_element_by_link_text(text)
-        self.selenium.driver.execute_script('arguments[0].click()', element)
+        self.selenium.driver.execute_script('arguments[0].click()', element)  
     
     def verify_expected_batch_values(self, batch_id,**kwargs):
         """To verify that the data in Data Import Batch matches expected value provide batch_id and the data u want to verify"""    
@@ -868,3 +868,23 @@ class NPSP(object):
         locator=self.get_npsp_locator(path, *args, **kwargs)  
         self.selenium.click_element(locator)      
             
+    def wait_for_record_to_update(self, id, value):
+        """Waits for specified record header to be updated by checking every second for 10 times.
+        """
+        el = None
+        i = 0
+        while el is None:
+            i += 1
+            if i > 10:
+                raise AssertionError(
+                    "Timed out waiting for account name to be {} .".format(value)
+                )
+            self.salesforce.go_to_record_home(id)
+            try:
+                self.verify_header(value)
+                break
+            except :
+                time.sleep(1)
+                continue     
+            
+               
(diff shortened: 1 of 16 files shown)


tests/browser/affiliations/remove_primary_affiliation.robot
```diff
@@ -33,20 +35,16 @@ Remove Primary Affiliation for Contact2
     Scroll Element Into View  text:Do Not Contact
     Click Button    title:Edit Primary Affiliation
     Wait For Locator  record.edit_form
-    #Sleep    5
-    Scroll Page To Location    100    500
+    Page Scroll To Locator  detail_page.edit_mode.section_header    Contact Information
```
davisagli: Can you do `Scroll Element Into View    text:Contact Information` so that we don't need to add a locator?

skristem: that doesn't work as there is Contact Information twice on the page and this one is second one.

davisagli:
Looks like Add Existing Contact to Existing Household through Manage Household Page is still failing sometimes. Can we at least mark it unstable for now?

skristem:
@davisagli want to give 1 last try before i mark it as unstable

tests/NPSP.py
```diff
@@ -867,4 +867,24 @@ def click_element_with_locator(self, path, *args, **kwargs):
...
         self.selenium.click_element(locator)      
-            
\ No newline at end of file
+            
+    def wait_for_record_to_update(self, id, value):
+        """Waits for specified record header to be updated by checking every second for 10 times.
+        """
+        el = None
+        i = 0
+        while el is None:
```
davisagli: `el` will always be None, because there's nothing that sets it. You can just write `while True:` and get rid of the `el = None` line.

tests/NPSP.py
```diff
@@ -867,4 +867,24 @@ def click_element_with_locator(self, path, *args, **kwargs):
...
+            i += 1
+            if i > 10:
+                raise AssertionError(
+                    "Timed out waiting for account name to be {} .".format(value)
```
davisagli: This should probably say record name rather than account name since there's no reason this keyword would not work with other objects.

tests/NPSP.py
```diff
@@ -867,4 +867,24 @@ def click_element_with_locator(self, path, *args, **kwargs):
...
+                )
+            self.salesforce.go_to_record_home(id)
+            try:
+                self.verify_header(value)
+                break
+            except :
```
davisagli: We should always catch a specific exception with the Python `except` keyword so that we don't catch the KeyboardInterrupt that is raised when the user presses Ctrl+C to exit the program. If you don't know which exception to catch, at least write `except Exception:`. `KeyboardInterrupt` is a subclass of BaseException rather than Exception so that will avoid catching it.

tests/NPSP.py
```diff
@@ -867,4 +867,24 @@ def click_element_with_locator(self, path, *args, **kwargs):
...
+                time.sleep(1)
+                continue
```
davisagli: There's no need for the `continue` statement. Since this is the last line of the `while` loop, it will continue anyway!

PR #4124, opened 2019-03-15
adding 10,000,000 record chunking to Opportunity Naming batch job

### LINKED ISSUE #4087: Opportunity Naming Batch Hits 50 Million Row QueryLocator Limit
The OPP_OpportunityNaming_BATCH currently runs for all Opportunities, so if an org has more than 50 million Opportunities, they will hit the governor limit.

50 million row error

This is not a nightly scheduled job, but it is one that a customer can schedule. Requesting that we implement batch chunking for this job.

seanpat09:
# Critical Changes

# Changes

# Issues Closed
Fixes #4087

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/OPP_OpportunityNaming_BATCH.cls
+++ b/src/classes/OPP_OpportunityNaming_BATCH.cls
@@ -34,21 +34,38 @@
 * @group-content ../../ApexDocContent/Opportunity.htm
 * @description Batch class creates names all Opportunities per the naming spec.
  */
-public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>, Schedulable {
-    
+public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>, Schedulable, Database.Stateful {
     /** @description The query for the batch process to run on.*/
-    String query;
-    
+    private String query;
+
+    @TestVisible
+    private static Integer defaultQueryLimit= 10000000;
+
+    @TestVisible
+    private Id lastOppIdProcessed;
+
     /** @description The batch process constructor; creates opportunity query for all opportunities.*/
     public OPP_OpportunityNaming_BATCH() {
-        query = 'SELECT Id, Name FROM Opportunity';
+        query = 'SELECT Id, Name FROM Opportunity ORDER BY Id LIMIT ' + defaultQueryLimit;
+    }
+
+    /** @description Constructor that accepts Id offset*/
+    public OPP_OpportunityNaming_BATCH(Id idToOffset) {
+        query = buildOffsetQuery(idToOffset, defaultQueryLimit);
+    }
+
+    /** @description Builds query string with given Id to offset and query limit*/
+    private String buildOffsetQuery(Id idToOffset, Integer queryLimit) {
+        return String.format(
+            'SELECT Id, Name FROM Opportunity WHERE Id > \'\'{0}\'\' ORDER BY Id LIMIT {1}',
+            new List<String>{ idToOffset, String.valueOf(queryLimit) }
+        );
     }
 
     /** @description Batch process start method.*/
     public Database.QueryLocator start(Database.BatchableContext BC) {
         return Database.getQueryLocator(query);
     }
-
 
     /** @description Schedulable execute method.*/
     public void execute(SchedulableContext context) {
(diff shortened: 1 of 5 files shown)


seanpat09:
**lurch: attach W-031598

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LlEDQA0" target="_blank">W-031598</a>

src/pages/STG_PanelOppNamingBatch.page
```diff
@@ -12,7 +12,7 @@
                 strBatchComponentLabel="{!$Label.stgLabelOppNamingRefreshTitle}"
                 cNumberOfJobs="1"
                 startPolling="True"
-                stopPollingOnComplete="True"
+                stopPollingOnComplete="False"
```
seanpat09: @RadSFDeveloper @force2b I changed this to false so that it'll continue to poll for the job even after the job is complete, so as batch jobs chain, the latest batch job gets displayed. For example, let's say there 15,000,000 Opportunities in an org.

* When the job start, the progress container will say something along the lines of "1 out of 5000 batches processed", which is the first 10,000,000 Opportunities.
* When that batch is done, it will chain another job for the remaining 5,000,000 opportunities
* The progress container will then be replaced by this next job, saying something along the lines of "1 out of 2500 batches processed".

I imagine a better scenario would be to display several progress containers so you can see the progress of the chain, though with this being a P1 bug, maybe that should go into another work item so that this fix can go out sooner?

force2b: I don't see it being a major issue, though if I'm not mistaken the way that component works is that it gets the latest batch job running and displays its status - not specifically the job that was executed. It's possible that the component may pick up a different job after this one completes, even if that job isn't the same OppNaming job. 

The only other thing I'd note is that I would very surprised if an admin in a large org (one with more than 10M Opps) would actually stay on the NPSP Settings page for that many hours to monitor progress. If that's true, then I'd say we don't need to modify the component.

seanpat09: Ah good point on the it getting the latest batch job. You are probably right about admins actually staying on the page, I'll revert this change. If it does become an issue later, we can address this with another work item

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -34,22 +34,39 @@
...
-public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>, Schedulable {
-    
+public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>, Schedulable, Database.Stateful {
     /** @description The query for the batch process to run on.*/
-    String query;
-    
+    private String query;
+
+    @TestVisible
+    private static Integer defaultQueryLimit = 10000000;
```
force2b: What would you think of making this a constant?
`private static final Integer DEFAULT_QUERY_LIMIT = 10000000;`

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -34,22 +34,39 @@
 * @group-content ../../ApexDocContent/Opportunity.htm
 * @description Batch class creates names all Opportunities per the naming spec.
  */
-public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>, Schedulable {
```
force2b: I think it would be helpful to update the block comment on the class to explain that the batch job now uses a chunking pattern.

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -59,27 +76,45 @@ public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>,
     * @description Batch process execute method. Names and updates all opportunities in the current batch.
     */
     public void execute(Database.BatchableContext BC, List<sObject> scope) {
+        List<Opportunity> oppsToProcess = scope;
```
force2b: An alternative to this is to change the method signature to:
`public void execute(Database.BatchableContext BC, List<Opportunity> oppsToProcess)`

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -59,27 +76,45 @@ public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>,
...
+        }
+    }
+
+    /** @description Returns whether or not another batch should be chained*/
+    private Boolean shouldChainNextBatch() {
+        if (lastOppIdProcessed == null) {
+            return false;
+        }
+
+        String hasMoreQuery = buildOffsetQuery(lastOppIdProcessed, 1);
```
force2b: Perfect!

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -59,27 +76,45 @@ public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>,
...
+        List<Opportunity> oppsForUpdate = new List<Opportunity>();
+        for (Opportunity opp : oppsToProcess) {
+            if (opp.Name != originalOppNamesById.get(opp.id)) {
                 oppsForUpdate.add(opp);
+            }
         }
 
         if (!oppsForUpdate.isEmpty()) {
-            database.update(oppsForUpdate, false);
+            Database.update(oppsForUpdate, false);
```
force2b: Do you think we should use `UTIL_DMLService.updateRecords(oppsForUpdate, false)` instead?

src/classes/OPP_OpportunityNaming_BATCH_TEST.cls
```diff
@@ -0,0 +1,247 @@
...
+        Id expectedId = testOpps[NUM_OPPS - 1].Id;
+        System.assertEquals(expectedId, batch.lastOppIdProcessed,
+            'The execute method should track the Id of the last opportunity processed');
+    }
+
+    /**
+     * @description Confirms that the finish method chains the next batch with an offset
+     * of the last Opportunity Id processed
+     */
+
```
force2b: extra blank line here

src/classes/OPP_OpportunityNaming_BATCH.cls
seanpat09: I didn't make this a constant only because I wanted to be able to change it during tests so that I can test the chunking in `OPP_OpportunityNaming_BATCH.cls.shouldRenameOpportunitiesInChunks`

src/classes/OPP_OpportunityNaming_BATCH.cls
seanpat09: I didn't realize you could do that, I thought it would violate the signature for the `Database.Batchable<sObject>` interface. Thanks!

src/classes/OPP_OpportunityNaming_BATCH.cls
seanpat09: Yeah I think that makes sense, I'll make the change

src/classes/OPP_OpportunityNaming_BATCH.cls
force2b: Very good point.

seanpat09:
@rponti-sforg this is ready for QA!

PR #4131, opened 2019-03-19
Customizable Rollups: API to Retrieve Rollup Definitions - serialization fix

force2b:
After testing with the bata package, it's not possible to serialize the response when it's returned as an object that is not global. As a result, the api needs to return a serialized string rather than List<CRLP_Rollup> so that the contents can be maintained stateful in the bridge as well as written into the Big Object for persistent storage of the summaries.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


sam-knox:
@force2b - I'm guessing this doesn't need any release notes. Is that correct?

force2b:
@sam-knox Correct. The api is an early release for the DSO Bridge. I'd prefer that we not include it in the release notes until we're positive that it's finalized and usable by customers.

PR #4136, opened 2019-03-20
Move reports & dashboards to Cumulus

davisagli:
This moves the NPSP reports & dashboards metadata from the CumulusReports repository into Cumulus. This makes it possible to include them in the config_qa and config_managed flows and to test that they deploy successfully in feature builds. They used to be in the separate repository partly because that was necessary to have a separate installer for them, but MetaDeploy makes it possible for us to publish multiple installation plans from the same repository.

Some of the details of this include:
- I added a `deploy_reports` task which deploys the metadata stored in `unpackaged/config/reports`
- The contents of that folder are taken verbatim from CumulusReports master, except I did replace the `npsp__` prefix with the `%%%NAMESPACE%%%` token so that it can be deployed with both managed and unmanaged NPSP using CCI's namespace injection.
- I configured the `deploy_reports` task so that it will show up as an optional step in the new MetaDeploy installer plans, as well as a separate installation plan for reports & dashboards only that is more like the current one.
- I realized that the metadata in `unpackaged/config/trial` was inadvertently disabling reports and activities for some objects by virtue of leaving `enableReports` and `enableActivities` out of the metadata instead of explicitly continuing to set it to true (see https://success.salesforce.com/issues_view?id=a1p3A000000IXGkQAO&title=excluding-enableactivities-and-enablereports-settings-from-customobject-metadata-incorrectly-disables-the-settings-in-target-org). This is now fixed so that these settings will stay the same as in the packaged metadata.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@davisagli I've never heard of this package, so I'm not sure I'm really qualified to "approve" it, other than to say that the metadata seems to make sense. Might need someone who is familiar with the reports to verify that it installs and works as expected.

davisagli:
@force2b that's ok, I mostly tagged you to make sure you're aware this is going in.

force2b:
Ah. Makes sense. Thanks @davisagli

PR #4140, opened 2019-03-22
BDI: Prevent multiple process clicks on Data Import page

psadie:
- Bug Fix: Add Action and JS functions to handle 'Process Button' disable and form post

# Critical Changes

# Changes
- 'Begin Data Import Process' button will grey out and be disabled while form is posting.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/pages/BDI_DataImport.page
+++ b/src/pages/BDI_DataImport.page
@@ -38,11 +38,6 @@
             dominant-baseline: middle;
             fill: #00396B;
         }
-
-        .slds-button_extension-disabled {
-            pointer-events: none;
-            cursor: default;
-        }
     </style>
 
     <script>
--- a/src/pages/BDI_DataImport.page
+++ b/src/pages/BDI_DataImport.page
@@ -38,26 +38,20 @@
             dominant-baseline: middle;
             fill: #00396B;
         }
+
+        .slds-button_extension-disabled {
+            pointer-events: none;
+            cursor: default;
+        }
     </style>
 
     <script>
-        function buttonsEnabled(enabled) {
+        function enableButton(enabled) {
             let targetButton = document.getElementById('{!$Component.idForm.idDataImportProcessButton}');
             let enableProcessButton = targetButton.setAttribute('disabled', null);
             let disableProcessButton = targetButton.setAttribute('disabled', 'disabled');
 
             enabled === true ? enableProcessButton : disableProcessButton;
-        }
-
-        function tryImportData() {
-            // Call action function to post form
-            importDataActionFunction();
-
-            // Disable button to prevent spam saving
-            buttonsEnabled(false);
-
-            // Return false to prevent the click from posting the form a second time
-            return false;
         }
     </script>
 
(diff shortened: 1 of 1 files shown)


psadie:
**lurch: attach W-031416

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LkzDQAS" target="_blank">W-031416</a>

melissabarber:
@psadie is there an issue in Github we can link to this PR under the Issues Closed header?

psadie:
> @psadie is there an issue in Github we can link to this PR under the Issues Closed header?

@melissabarber Searched through for this one as well and didn't find a related issue we could close

src/pages/BDI_DataImport.page
```diff
@@ -83,7 +104,20 @@
...
                                         <div class="slds-form-element__control">
-                                            <apex:commandButton value="{!$Label.bdiRunBtn}" action="{!importData}" immediate="false" styleClass="slds-button slds-button_brand" rerender="idForm"/>
+                                            <apex:actionFunction name="importDataActionFunction"
+                                                                 action="{!importData}"
+                                                                 oncomplete="buttonsEnabled(true);"
+                                                                 rerender="idForm">
+                                            </apex:actionFunction>
+
+                                            <apex:commandButton value="{!$Label.bdiRunBtn}"
+                                                                action="{!importData}"
```
ptewson-sfdo: Does having both the button action call {!importData} and the actionFunction call {!importData} mean that the controller method is being called twice? I haven't being in visualforce land in forever but it seems like that might be an issue

src/pages/BDI_DataImport.page
```diff
@@ -83,7 +104,20 @@
...
+                                                                immediate="false"
+                                                                styleClass="slds-button slds-button_brand"
+                                                                rerender="idForm"
+                                                                id="idDataImportProcessButton"
+                                                                onclick="return tryImportData();"/>
```
ptewson-sfdo: I've been thinking about this a bit more - I think you may be able to simplify this overall by specifying:
  onclick="buttonsEnabled(false);"
  onComplete="buttonsEnabled(true);"/>

in the attributes of the command button, and I think you can avoid the actionFunction entirely, as well as the tryImportData() javascript function. That seems a bit more robust as it calls the apex controller in the more traditional visualforce method instead of from the javascript intermediary - however, my visualforce is pretty rusty so feel free to let me know if you are ahead of me and have justification for this setup that I wasn't are of yet.

src/pages/BDI_DataImport.page
psadie: Good point! I believe the action on the commandButton isn't actually doing anything at this point as returning false within its onclick function overrides the commandButton's normal behavior. So at the very least there's no point in having it there. Tested this out by dropping a sys debug in the controller's 'importData' method. I'm looking into your follow up comment below now. Currently under the impression that disabling the button before the post completes actually scrubs the request of which button was actually clicked and along with it which action to execute.

melissabarber:
Release note not needed.

PR #4142, opened 2019-03-22
Telemetry for Batch Job Performance and RD Statistics

### LINKED ISSUE #4119: UTIL_OrgTelemetry_SObject_BATCH Hits 50 Million QueryLocator Limit
Orgs with more than 50 million Opportunities associated with open-ended Recurring Donations will see an error when the UTIL_OrgTelemetry_SObject_BATCH runs its query on Lines 78-79.

`First error: npsp:Too many query locator rows: 50000001`

seanpat09:
Adds the following telemetry:
* Data_CountRdsWithDiffAmount
* Data_CountRecurringDonationsAll
* Org_LongestRunningBatchLast7Days
* Org_CountTimedOutBatchJobsLast7Days

# Critical Changes

# Changes
## Telemetry Enhancements
* Organizations with more than 50 million Opportunities associated were seeing an error when the UTIL_OrgTelemetry_SObject_BATCH ran. We've implemented batch chunking to resolve these errors.

* NPSP Feature Telemetry collects feature and usage metrics when customers use our products. We use this information to understand the level of usage and adoption of various NPSP features. No personally identifiable information is collected through NPSP Feature Telemetry. We've added the following to NPSP Feature Telemetry:
    - Count of Recurring Donations with at least one Opportunity where the amount differs from the Recurring Donation amount
    - Count of all Recurring Donations
    - Count of  batch jobs that failed with a TIME_OUT error within the last 7 days
    - Count (in minutes) of the  batch job with the longest execution time within the last 7 days
    - For more information about NPSP Feature Telemetry, read: https://powerofus.force.com/articles/Resource/NPSP-Feature-Telemetry

# Issues Closed
* Fixes #4119

# New Metadata
- ApexClass: UTIL_FeatureManagement
- ApexClass UTIL_FeatureManagement_TEST
- FeatureParameterInteger: Data_CountRdsWithDiffAmount
- FeatureParameterInteger: Data_CountRecurringDonationsAll
- FeatureParameterInteger: Org_LongestRunningBatchLast7Days
- FeatureParameterInteger: Org_CountTimedOutBatchJobsLast7Days

# Deleted Metadata


src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -74,10 +88,15 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
      */
     public Database.QueryLocator start(Database.BatchableContext context) {
         switch on (jobMode) {
-            when RECURRING_DONATIONS {
+            when RECURRING_DONATIONS_OPPS {
```
seanpat09: I switched the original job mode to use RECURRING_DONATIONS_OPPS since it was iterating over opportunities instead of Recurring donations

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -113,14 +136,33 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
...
 
+    /**
+     * @description Process RecurringDonation counters for the specific Opportunity
+     */
+    private void getRecurringDonationsCounts(npe03__Recurring_Donation__c recurringDonation) {
+        rdAllCount++;
+
+        for (Opportunity opp : recurringDonation.npe03__Donations__r) {
+            if (opp.Amount != recurringDonation.npe03__Amount__c) {
+                rdDiffCount++;
```
seanpat09: The quip dock mentioned using the original job to count the recurring donation diffs, but I thought it would be better to do this in the new job so we don't have to keep track of the unique recurring donations.

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -36,26 +36,40 @@
...
 private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
 
+    private static UTIL_FeatureManagement_TEST.Mock featureManagementMock = new UTIL_FeatureManagement_TEST.Mock();
+
     /**
      * @description Create test data needed for the unit tests in this class
      */
     @TestSetup
     static void testDataSetup() {
+        TDTM_Config_API.disableAllTriggers();
```
seanpat09: With the FeatureManagement mock, I thought it be easier to set up the test data manually instead of relying on the triggers to create the opportunities so that it's easier to track the counts.

seanpat09:
**lurch: attach W-031524

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008Ll8FQAS" target="_blank">W-031524</a>

src/featureParameters/Data_CountRdsWithDiffAmount.featureParameterInteger
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FeatureParameterInteger xmlns="http://soap.sforce.com/2006/04/metadata">
+    <dataflowDirection>SubscriberToLmo</dataflowDirection>
+    <masterLabel>Data: Count Of Recurring Donations With A Differing Opportunity Amount</masterLabel>
```
seanpat09: @sam-knox I'm not sure how important these labels are in these feature parameters, but let me know if you want me to change any of them

force2b: @seanpat09 and @sam-knox These labels are not visible to Customers at all. They're actually not visible to us either. They appear in the Cumulus Packaging Org as the description for the Feature Parameter.

src/classes/UTIL_FeatureManagement_TEST.cls
```diff
@@ -0,0 +1,84 @@
...
+        new UTIL_FeatureManagement().setPackageIntegerValue(
+            UTIL_OrgTelemetry_SVC.TelemetryParameterName.Data_CountRdOppsAll.name(),
+            1
+        );
+    }
+
+    /*******************************************************************************************************************
+    * @description Mock for UTIL_FeatureManagement instance
+    */
+    public class Mock implements System.StubProvider {
```
force2b: Can I just say that this is brilliant. When we were having trouble testing these initially it never occurred to me to stub them this way for tests. Great work!!

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -74,10 +88,15 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
...
     public Database.QueryLocator start(Database.BatchableContext context) {
         switch on (jobMode) {
-            when RECURRING_DONATIONS {
+            when RECURRING_DONATIONS_OPPS {
                 String soql = 'SELECT Id, npe03__Recurring_Donation__r.npe03__Open_Ended_Status__c FROM Opportunity';
                 return Database.getQueryLocator(soql);
             }
+
+            when RECURRING_DONATIONS {
+                String soql = 'SELECT npe03__Amount__c, (SELECT Amount FROM npe03__Donations__r) FROM npe03__Recurring_Donation__c';
```
force2b: Using a sub-query in a queryLocator, especially in a large org, can cause query timeout errors. The recommended best practice for this is to query the main object -- `npe03__Recurring_Donation__c` -- in the QueryLocator and then execute a second query in the `execute()` method to retrieve the related records.

The one concern I have is that we could hit a governor limit in the execute() method if there are more than 50K related Opportunities for the collection of 2000 Recurring Donation Id's. To that end, I recommend changing the batch size of the RecurringDonations job to 200 instead of 2000. That would allow for up to 250 Opportunities per RD, assuming every RD in the batch had that many. Plus generally there are significantly less RD's in an Org than Opps, so a batch size of 200 should still finish pretty fast.

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
force2b: Agreed. Good idea.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -195,7 +212,79 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+
+        }
+
+        featureManager.setPackageIntegerValue(TelemetryParameterName.Org_LongestRunningBatchLast7Days.name(), longestJobLength);
+    }
+
+    /**
+     * @description Calculates the number of minutes between two date times
+     */
+    private Integer getMinutesBetween(DateTime startDateTime, DateTime endDateTime) {
```
force2b: Not sure this is correct. I tried the following code. The result should be 11 hours (660 minutes), however the method returns 2100 minutes (35 hours).
```
DateTime startDate = DateTime.newInstanceGMT(2019, 03, 01, 23, 00,00);
DateTime endDate = DateTime.newInstanceGMT(2019, 03, 02, 10, 00,00);
System.debug(startDate + ' -> ' + endDate);
System.debug(getMinutesBetween(startDate, endDate));
```

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -195,7 +212,79 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+
+        featureManager.setPackageIntegerValue(TelemetryParameterName.Org_CountTimedOutBatchJobsLast7Days.name(), countTimeouts);
+    }
+
+    /**
+     * @description Determins whether batch has a time out error using the ExtendedStatus field
+     */
+    private Boolean batchHasTimeoutError(AsyncApexJob job) {
+        String requestTooLongMessge = 'request_running_too_long';
+        String cpuTimeOutMessage = 'cpu time limit';
```
force2b: Were you able to validate the message for `cpu time limit`?

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -36,6 +36,7 @@
 @IsTest(IsParallel=false)
 private class UTIL_OrgTelemetry_TEST {
 
+    private static UTIL_FeatureManagement_TEST.Mock featureManagementMock = new UTIL_FeatureManagement_TEST.Mock();
```
force2b: You may be run this in `parallel=true` mode now that you're mocking the feature parameter part. Worth trying if you haven't already.

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
force2b: Since this actually iterates over Opportunities, even though it's counting RD related stuff, it probably make sense to rename this enum to `OPPORTUNITIES`.

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -85,11 +86,17 @@ public without sharing class UTIL_OrgTelemetry_BATCH implements Database.Batchab
     public void finish(Database.BatchableContext info) {
         Integer cnt = Database.countQuery('SELECT Count() FROM npe03__Recurring_Donation__c LIMIT 1');
         if (cnt == 1) {
+            Database.executeBatch(new UTIL_OrgTelemetry_SObject_BATCH(
+                    UTIL_OrgTelemetry_SObject_BATCH.TelemetryBatchJobMode.RECURRING_DONATIONS_OPPS), 2000);
+
             Database.executeBatch(new UTIL_OrgTelemetry_SObject_BATCH(
```
force2b: See previous note on the recurring donations job batch size.

src/classes/UTIL_OrgTelemetry_SVC.cls
seanpat09: Yup, the actual full message if you get a time out during a a batch is `First error: Apex CPU time limit exceeded`

src/classes/UTIL_OrgTelemetry_SVC.cls
seanpat09: Ah for some reason I was removing a day from the start date in the method, probably from when I was originally testing the logic. Good catch! I'll add a unit test, date/time calculations are confusing enough to warrant one even if it is a private method

src/classes/UTIL_OrgTelemetry_TEST.cls
```diff
@@ -73,80 +74,249 @@ private class UTIL_OrgTelemetry_TEST {
...
+    /**
+     * @description Validates calculation in getMinutesBetween()
+     */
+    @IsTest
+    private static void shouldCalculateMinutesBetweenTwoDateTimes() {
+        UTIL_OrgTelemetry_SVC service = new UTIL_OrgTelemetry_SVC();
+        DateTime startDate = DateTime.newInstanceGMT(2019, 03, 01, 23, 00,00);
+        DateTime endDate = DateTime.newInstanceGMT(2019, 03, 02, 10, 00,00);
+
+        System.assertEquals(660, getMinutesBetween(startDate, endDate),
```
force2b: @seanpat09 The `getMinutesBetween` method needs the class name to compile.

seanpat09:
@force2b made the suggested changes, thanks again for the feedback. Let me know if there is anything else!

src/classes/UTIL_FeatureManagement_TEST.cls
RadSFDeveloper: I second that, brilliant work on using stubs and testing overall, @seanpat09 !

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -67,7 +81,166 @@ private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
...
+    /**
+     * @description Builds test opportunity records with the given field values
+     */
+    private static List<Opportunity> buildOpps(Integer numOpps, Map<Schema.SObjectField, Object> fieldValues) {
+        List<Opportunity> opps = new List<Opportunity>();
+
+        for (Integer i = 0; i < numOpps; i++) {
+            Opportunity opp = new Opportunity(
+                Name = 'Test Opp ' + i,
+                StageName = 'Closed Won',
```
RadSFDeveloper: Let's use `UTIL_UnitTestData_TEST.getClosedWonStage()` instead of the value.

src/classes/UTIL_OrgTelemetry_BATCH.cls
RadSFDeveloper: Should we have a comment (or local var and description, or property) here indicating why certain batch sizes are used so anyone changing the code down the road does not change the batch size?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -195,7 +222,80 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+     */
+    private void handleLongestRunningBatchLast7Days() {
+        Integer longestJobLength = 0;
+
+        for (AsyncApexJob job : [
+            SELECT CreatedDate, CompletedDate
+            FROM AsyncApexJob
+            WHERE CreatedDate = LAST_N_DAYS:7
+                AND JobType = 'BatchApex'
+                AND Status IN ('Aborted', 'Completed', 'Failed')
```
RadSFDeveloper: Suggestion: use either status constants, or have a class property that is `Set<String> completedBatchStatuses = new Set<String>{  'Aborted', 'Completed', 'Failed' }` since they are used in more than one query.  So the query here would be `AND Status IN :completedBatchStatuses`? 
Similar comment is for the namespaces.

src/classes/UTIL_OrgTelemetry_BATCH.cls
seanpat09: Good idea, I'll add a constant with a description

seanpat09:
**lurch: attach W-031902

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LlcUQAS" target="_blank">W-031902</a>

patrick-yan-sf:

https://salesforce.quip.com/8aKBACw5QLqb

PR #4147, opened 2019-03-27
Feature/robot  reorganization

boakley:
This branch has a reorganization of the tests folder, based on my earlier proposal. The main difference between this and the original proposal is that instead of tests/Cumulus/tests I used robot/Cumulus/tests. 

In addition to moving things around, I modified cumulusci.yml to save the robot_libdoc output to robot/Cumulus/doc, and test output to robot/Cumulus/results. We probably want to add those folders to .gitignore, though I haven't yet done that. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cumulusci.yml
```diff
@@ -67,10 +67,16 @@ tasks:
         options:
             path: unpackaged/config/delete
 
+    robot:
+        options:
+            suites: robot/Cumulus/tests
+            options:
+                outputdir: robot/Cumulus/results
```
davisagli: yay, I've been wanting this. Please do go ahead and add this folder to .gitignore.

robot/Cumulus/tests/browser/batch_gift_entry/bge_default_batch.robot
```diff
@@ -1,6 +1,7 @@
 *** Settings ***
 
-Resource        tests/NPSP.robot
+#Resource        robot/Cumulus/resources/NPSP.robot
```
skristem: looks like you forgot to remove the commented line

boakley: Thanks for catching that. I've removed it and pushed the change.

boakley: Thanks for catching that. I've removed it and pushed it.

PR #4151, opened 2019-03-29
Feature/bdi co  pgt field mapping object hierarchy

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_FieldMapping_BASE.cls
```diff
@@ -0,0 +1,97 @@
...
+    * @description flag to allow us to test out field level security exception
+    */
+    public Boolean failFieldLevelSecurity = false;
+
+
+    public BDI_FieldMapping_BASE() {
+
+    }
+
+    public static BDI_FieldMapping_BASE getInstance() {
```
ptewson-sfdo: The commented-out code here demonstrates how we would provide the different mapping behaviors based on settings, once the custom-metadata variant is created

src/classes/BDI_SettingsUI_CTRL.cls
```diff
@@ -91,8 +91,8 @@ public with sharing class BDI_SettingsUI_CTRL {
                 listSODonationFields = new list<SelectOption>();
                 
                 set<String> setFields = new set<String>();
-                setFields.addAll(BDI_Donations.dataImportFieldToOpportunityField.keySet());
-                setFields.addAll(BDI_Donations.dataImportFieldToPaymentField.keySet());
+                //setFields.addAll(BDI_Donations.dataImportFieldToOpportunityField.keySet());
```
ptewson-sfdo: Need to do something here, but this dependence on these field lists needs to shift with our new scheme I think anyway

ptewson-sfdo:
**lurch: attach W-032261

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008Lm6FQAS" target="_blank">W-032261</a>

src/classes/BDI_DataImportService.cls
```diff
@@ -771,89 +773,31 @@ global with sharing class BDI_DataImportService {
...
-
-    /*******************************************************************************************************
-    * @description Checks if this is one of the objects where we want to have a currency mapped.
-    * As of now, that object is Opportunity. Payment CurrencyISOCode is inherited from the Opportunity.
-    * @param strBaseObj the underlying object the bdi object is (ie, Contact, Account, etc)
-    * @return Boolean if the strBaseObj is an Opportunity.
-    */
-    private static Boolean isCurrencyMapped(String objectName) {
-        return objectName.equalsIgnoreCase('Opportunity');
+    global static Map<String, String> mapFieldsForDIObject(String dataImportObjectName,
```
psadie: I think this method is still being used in BDI_MatchDonations. Would that be an issue if this now always only returns an empty map?

src/classes/BDI_SettingsUI_CTRL.cls
```diff
@@ -128,7 +133,7 @@ public with sharing class BDI_SettingsUI_CTRL {
                 listSOAccountUniqueID = listSOForObject('Account');
             }
             return listSOAccountUniqueID;
-        }
+        }
```
psadie: Just some extra spaces

src/classes/BDI_DataImportService.cls
callen-sfdo: Yeah, I think one of two things needs to be done:  1) once the custom setting field is in place to control mapping we could modify this to automatically construct the right version the Data import service based on the setting.  

Or 2)We need to remove all references to this method in our code and throw an error if it is called.

ptewson-sfdo: Good find - I'm in favor of pulling all our internal references to the method for SURE, and I updated this PR to (I think) take out all the remaining calls. I do think it might be wise to support this method for the short term just in case a customer is calling it, but we should probably add the  "Deprecated" annotation, which I believe is designed for this use case  - https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_annotation_deprecated.htm

PR #4155, opened 2019-04-01
BDI/BGE: Update FLS to check only populated fields for DI process/dry-run

psadie:
This updates the way we handle FLS during the mapping phase of processing/dry-running. Previously we checked FLS on all the mapped fields. Now we'll run FLS only on fields that are actually being populated/otherwise being updated in the DI records. If any DI record in a Batch has a failing FLS check, then the entire batch is failed.

# Critical Changes

# Changes
- We updated how Batch Gift Entry and Data Import handle field level security (FLS). Previously, Dry Run in BGE and Data Import would throw a permissions error even if restricted fields weren't included in the batch.  Now, Dry Run only verifies that a user has Edit permissions on fields that are included in the batch.

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -36,6 +36,108 @@
...
+    }
+
+    /*******************************************************************************************************
+    * @description Returns the subset of DataImport__c fields that are part of every batch
+    * @param includeRelationshipFields whether to include relationship fields in the returned list
+    * @return list of DataImport__c field api names
+    */
+    private static List<String> getCoreDataImportFields(Boolean includeRelationshipFields) {
+        List<String> fields = new List<String> {
+            'Id',
```
ptewson-sfdo: Is there any reason that Donation_Amount__c, Donation_Date__c, and NPSP_Data_Import_Batch__c aren't included in this list?

src/classes/BDI_DataImportService.cls
```diff
@@ -78,10 +78,14 @@ global with sharing class BDI_DataImportService {
...
     */
     private static DataImportBatch__c loadBatch(Id batchId) {
-
         List<DataImportBatch__c> dataImportBatches = BDI_DataImportBatch_SEL.selectByIds(new Set<Id>{ batchId });
         if (dataImportBatches.size() == 1) {
-            return dataImportBatches[0];
+            DataImportBatch__c currentBatch = dataImportBatches[0];
+            if (!String.isBlank(currentBatch.Active_Fields__c)) {
+                relevantDataImportFields = BGE_BatchGiftEntry_UTIL.getDataImportFields(batchId, false);
+                batchFromBGE = true;
```
ptewson-sfdo: Want to avoid references to "BGE" if we can describe this property more generically ... "batchDefinesActiveFieldList" or somethign

src/classes/BDI_DataImportService.cls
```diff
@@ -78,10 +78,14 @@ global with sharing class BDI_DataImportService {
     * @return DataImportBatch__c
     */
     private static DataImportBatch__c loadBatch(Id batchId) {
-
         List<DataImportBatch__c> dataImportBatches = BDI_DataImportBatch_SEL.selectByIds(new Set<Id>{ batchId });
         if (dataImportBatches.size() == 1) {
-            return dataImportBatches[0];
+            DataImportBatch__c currentBatch = dataImportBatches[0];
+            if (!String.isBlank(currentBatch.Active_Fields__c)) {
+                relevantDataImportFields = BGE_BatchGiftEntry_UTIL.getDataImportFields(batchId, false);
```
ptewson-sfdo: This was the thing that made me the most uncomfortable about the original PR, but I am thinking we may be able to avoid it - applications calling BDI (like BGE) should ideally not be defining BDI behaviors, else we could build up all these conditionals and exceptions built in for each application that's calling into it. HOWEVER, I think we may be able to avoid this

src/classes/BGE_BatchGiftEntry_UTIL.cls
psadie: Yeah, these get added in by the "getActiveFieldNamesFromBatch" method as these are always required and are also considered "selected" fields in the BGE.

src/classes/BDI_DataImportService.cls
psadie: @ptewson-sfdo - Would moving this logic (along with other the field collection/map building) from the BGE util to the BDI_DataImportService class make sense? The one thing that immediately sticks out to me is that the `getActiveFieldNamesFromBatch` method from `BGE_BatchGiftEntry_UTIL` relies on the `BGE_ConfigurationWizard_CTRL.BGEField` class. If the above is acceptable, would moving `BGEField` into its own file make sense so we don't rely on the config wizard controller in BDI here as well?

src/classes/BDI_DataImportService.cls
```diff
@@ -674,6 +692,28 @@ global with sharing class BDI_DataImportService {
         }
     }
 
+    /*******************************************************************************************************
+    * @description flag to let us know if we came from BGE
+    */
+    global static Boolean batchFromBGE { get; set; }
```
ptewson-sfdo: avoid global!

src/classes/BDI_DataImportService.cls
```diff
@@ -674,6 +692,28 @@ global with sharing class BDI_DataImportService {
...
+
+    /*******************************************************************************************************
+    * @description list of the fields in the Data Import object that we care about for FLS checks because
+    * they are present in the Batch we are processing
+    * This list gets set in the loadBatch method.
+    * If no batch is loaded, we use the full list of all DI fields listStrDataImportFields
+    */
+    global static List<String> relevantDataImportFields {
```
ptewson-sfdo: avoid global!

src/classes/BGE_BatchGiftEntry_UTIL.cls
ptewson-sfdo: nm let's leave this alone

src/classes/BDI_DataImportService.cls
ptewson-sfdo: We believe we can pull the call to BGE_BatchGiftEntry_UTIL.getDataImportFields(batchId, false) entirely - these fields would only be used for FLS when the help text maps to a target object field and then the check would be on the target field. None of these fields have default mappings or map logically to target fields and would not be checked anyway, and we can remove an unneeded dependency on BGE by pulling this out.

src/classes/BGE_ConfigurationWizard_CTRL.cls
```diff
@@ -200,16 +200,20 @@ public with sharing class BGE_ConfigurationWizard_CTRL {
...
-            Boolean fieldAllowed = (allowedObjects.contains(dataImportFieldMappedObject)
+            Boolean fieldAllowed = (
+                UTIL_Permissions.canUpdate(dfr, false)
+                && ((allowedObjects.contains(dataImportFieldMappedObject)
                 && allowedDataTypes.contains(dfr.type)
                 && !bannedFields.contains(dataImportFieldApiName)
                 && dfr.getInlineHelpText() != null)
-                || allowedFields.contains(dataImportFieldApiName);
+                || allowedFields.contains(dataImportFieldApiName))
+            );
```
ptewson-sfdo: Is the call to "BGE_BatchGiftEntry_UTIL.checkFieldPermissions(filteredDFRMap.values());" below on line 223 redundant now, since the field has to be updatable to get in the map in the first place?

psadie: Yeah, it is redundant as all the fields passed in will always pass. What do you think about the following scenario?

If a user without perms for a field (Campaign Source) accesses a DI Batch that has Campaign Source in its Opportunity Selected Fields... with this update they'll receive an error on the DI Batch page but still be able to click "Edit" and open the Config Wizard. The page will tell them they don't have access to field Campaign Source, the page will not render the New Gift form and not render the Gifts table. They'll still be able to open the Config Wizard, but Campaign Source will not be visible for them in the Opportunity Selected Fields.

ptewson-sfdo: This is what we discussed earlier today (Monday 4/8) right?  I think this is ok, we may look to providing a more informative blocking message when trying to edit a batch with an inaccessible field, but this is great of this story.

ptewson-sfdo: And I believe we should still remove the redundant call to "BGE_BatchGiftEntry_UTIL.checkFieldPermissions(filteredDFRMap.values());" on line 223-ish?

psadie: Yup, exactly. Completely agree, I'll go ahead and remove the call to the BGE util

src/classes/BDI_DataImportService.cls
```diff
@@ -812,8 +826,8 @@ global with sharing class BDI_DataImportService {
 
                         if (failFieldLevelSecurity ||
                             fieldDescribe == null ||
-                            !fieldDescribe.isCreateable() ||
-                            (!isAuditField && !fieldDescribe.isUpdateable())) {
+                            (!fieldDescribe.isCreateable() && populatedDataImportFields.contains(dataImportField)) ||
```
ptewson-sfdo: format this block's linebreaks a little more cleanly?

ptewson-sfdo:
**lurch: attach W-030471

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQPqQAM" target="_blank">W-030471</a>

PR #4158, opened 2019-04-02
[PYMNT ALLOC] unit test coverage and bug fixes

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/ALLO_PaymentSync_SVC.cls
+++ b/src/classes/ALLO_PaymentSync_SVC.cls
@@ -32,7 +32,7 @@
 * @date 2019
 * @group Allocations
 * @group-content ../../ApexDocContent/Allocations.htm
-* @description Service syncs changes between Opportunity and Payment 
+* @description Service syncs changes between Opportunity and Payment
 *              Allocations.
 */
 public class ALLO_PaymentSync_SVC {
@@ -46,6 +46,10 @@ public class ALLO_PaymentSync_SVC {
     private class SynchronizationContext {
         Opportunity opportunity;
         Map<Id, npe01__OppPayment__c> payments = new Map<Id, npe01__OppPayment__c>();
+
+        //Orig allocations from trigger context, arranged by gau Id
+        //This is needed for deletes, where deleted allocation is not available in other context vars
+        Map<Id, Allocation__c> triggerAllocationsByGAUId = new Map<Id, Allocation__c>();
 
         //list of allocations by GAU
         Map<Id, List<Allocation__c>> opportunityAllocationsByGAU = new Map<Id, List<Allocation__c>>();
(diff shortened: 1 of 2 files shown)


ptewson-sfdo:
@jbrandolinisf - this should be ready for your review!

src/classes/ALLO_PaymentSync_SVC.cls
```diff
@@ -199,14 +219,16 @@ public class ALLO_PaymentSync_SVC {
                 Decimal totalOpportunityAllocations = context.totalOpportunityAllocationsByGAU.get(gau);
                 Decimal adjustedTotalPaymentAllocations = totalPaymentAllocations * context.scheduleRatio;
                 
-                System.debug('njjc totalOpportunityAllocations ' + totalOpportunityAllocations + ' adjustedTotalPaymentAllocations ' + adjustedTotalPaymentAllocations);
+                System.debug('njjc totalOpportunityAllocations ' + totalOpportunityAllocations + ' adjustedTotalPaymentAllocations ' +
```
jbrandolinisf: Remove me please...:-)

src/classes/ALLO_PaymentSync_SVC.cls
```diff
@@ -199,14 +219,16 @@ public class ALLO_PaymentSync_SVC {
...
+                System.debug('njjc totalOpportunityAllocations ' + totalOpportunityAllocations + ' adjustedTotalPaymentAllocations ' + 
+                    adjustedTotalPaymentAllocations);
 
                 //check if they need updating
                 if (adjustedTotalPaymentAllocations != totalOpportunityAllocations) {
                     for (Allocation__c oppAllocation : context.opportunityAllocationsByGAU.get(gau)) {
                         //we want to keep the ratios amongst duplicate opportunity allocations
                         Decimal duplicateAllocationPaidUnpaidAmountRatio = oppAllocation.Amount__c / totalOpportunityAllocations;
-                        System.debug('njjc duplicateAllocationPaidUnpaidAmountRatio ' + duplicateAllocationPaidUnpaidAmountRatio + ' calculatedAmount ' + totalPaymentAllocations * duplicateAllocationpaidUnpaidAmountRatio);
+                        System.debug('njjc duplicateAllocationPaidUnpaidAmountRatio ' + duplicateAllocationPaidUnpaidAmountRatio +
```
jbrandolinisf: This is part of my war on System.debug...it's gotta go...

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -221,14 +221,143 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+        //original payment allocations
+        assertAllocation(queryAllocations[0], 200, defaultGau.Id, false);
+        assertAllocation(queryAllocations[1], 300, gaus[1].Id, false);
+        assertAllocation(queryAllocations[2], 500, gaus[0].Id, false);
+    }
+
+
+    /*******************************************************************************************************
+    * @description When changing the amount of Opportunity allocations connected to paid payments
+    * Payment Allocations for paid payments will not be changed
```
jbrandolinisf: It is good either way, but maybe could you add one additional payment that is not paid, and confirm that the changes do flow to that payment?

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -221,14 +221,143 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+
+        //original payment allocations
+        assertAllocation(queryAllocations[0], 200, defaultGau.Id, true);
+        assertAllocation(queryAllocations[1], 300, gaus[1].Id, true);
+        assertAllocation(queryAllocations[2], 500, gaus[0].Id, true);
+        assertAllocation(queryAllocations[3], origAllocAmount, defaultGau.Id, false);
+    }
+
+    /*******************************************************************************************************
+    * @description When attempting to lower or delete an opp allocation for a GAU that has paid payment allocations:
```
jbrandolinisf: Do we need to confirm the other case (where there is some paid, and some unpaid), and the allocations (up to the amount of the unpaid payment)?

src/classes/ALLO_PaymentSync_TEST.cls
ptewson-sfdo: It's a good point - I could definitely be flexible, but I'm trying to keep my unit tests more focused and atomic for what they are testing in general (and in this case stick to the specific tests that Nic was requesting). I might consider that as a separate test method if anything, more than putting that test in here.

src/classes/ALLO_PaymentSync_TEST.cls
ptewson-sfdo: Perhaps, although this wasn't listed specifically in the list of tests Nic left us with... not that we have to anchor ourselves to those specific tests, but in the interest of limiting the scope of each individual test it might be another one I would lobby for an additional test method.

jbrandolinisf: Ok.  Why don't we throw these tests in a future Work Item.  Maybe they are good (cutting teeth) Work Items for a new developer.

src/classes/ALLO_PaymentSync_TEST.cls
jbrandolinisf: @ptewson-sfdo - Maybe we log these as a future work item, and a good candidate for a new developer to cut their teeth on.

PR #4160, opened 2019-04-03
CRLP: Execute Rollups Api

force2b:
* Processing methods logic previously in `CRLP_RollupProcessor_SVC` have been refactored out into an instance class of `CRLP_RollupProcessor`.
* Much of query building logic that was spread across multiple classes has been consolidated into an instance class of `CRLP_QueryBuilder`. It currently only supports detail queries. Longer term this class should support building out the QueryLocator and Related Record queries. For now that logic still exists in other locations.
* Long methods in the batch and processor classes have been abstracted out into smaller methods to consolidate logic into methods that do one (or mostly one) thing.

# Critical Changes

# Changes

* We've introduced a new Callable_Api action for CRLP.ExecuteRollups. This is reserved for future use and can be ignored for now.

# Issues Closed

# New Metadata

* Apex Class: CallableApiParameters
* Apex Class: CRLP_RollupProcessor
* Apex Class: CRLP_QueryBuilder
* Apex Class: CRLP_ApiExecuteRollups

# Deleted Metadata


src/classes/CRLP_RollupProcessor.cls
```diff
@@ -0,0 +1,538 @@
...
+
+    /**
+    * @description Returns the Parent Object Field by parsing the parentKeyFieldFull property
+    */
+    private String parentObjectField {
+        get {
+            if (parentObjectField == null) {
+                // The parentKeyField field may be have dot-notation (ex: Contact__r.AccountId)
+                // If this is the case, need to parse out the two components to use
+                String keyField = parentRelationshipField;
```
RadSFDeveloper: No need for `keyField` var, can use `parentRelationshipField`.

If the `parentRelationshipField` is just `AccountId` or `Account__c`, then `parentObjectField` will always be null. Thus, it is best to set this field to empty string by default, and whenever the value is checked (line 152), use `String.isBlank(parentObjectField)` and not use `parentObjectField == null`. This is applicable for all strings, let's use String methods.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -0,0 +1,538 @@
...
+        } set;
+    }
+
+    /**
+     * @description Determine the SObjectType of the Detail Records based on the current rollup type
+     */
+    private String parentRelationshipField {
+        get {
+            if (parentRelationshipField == null) {
+                parentRelationshipField = CRLP_Rollup_SVC.getParentRelationshipFieldForProcessingJob(rollupType);
```
RadSFDeveloper: `getParentRelationshipField()`

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -0,0 +1,538 @@
...
+
+    /**
+    * @description When PreventDupesInRollup is true, this method will purge the list of detail Opps in the detailsForParent
+    * collection to remove any that were previously processed. This is used in AccountContactSoftCredit rollups to
+    * prevent an Account from both getting soft credit for the same opportunity more than one - when there is more than
+    * one Household Contact that got soft credit for that Opportunity.
+    */
+    private List<SObject> cleanDetailOppsListForSoftCreditMode(Id parentId, List<SObject> detailsForParent) {
+        List<SObject> oppsToRollup = new List<SObject>();
+        for (Integer itemNum = 0, detailListSize = detailsForParent.size(); itemNum < detailListSize; itemNum++) {
```
RadSFDeveloper: Lines 384-385 can be: `for (SObject detail : detailsForParent) {`

src/classes/CRLP_RollupQueryBuilder.cls
```diff
@@ -0,0 +1,312 @@
...
+            // Force the details query to sort by Ascending CreatedDate so that the records are process
+            // in the order in which they were created in the database. Thus the 'first' is truly the first
+            // and the last is truly the last.
+            orderByFields.add('CreatedDate ASC');
+        }
+
+        if (!whereClauses.isEmpty()) {
+            query += ' WHERE ' + String.join(whereClauses, ' AND ');
+        }
+        if (customWhereClauses != null && !customWhereClauses.isEmpty()) {
```
RadSFDeveloper: What is the purpose of the `customWhereClauses`? Can the query builder have both `whereClauses` and `customWhereClauses`?

src/classes/CRLP_RollupQueryBuilder.cls
```diff
@@ -0,0 +1,312 @@
...
+        String detailKeyField;
+        List<SObjectType> childObjects = new List<SObjectType>();
+
+        SObjectType summaryObjectType = CRLP_Rollup_SVC.getSummaryObjectType(rollupType);
+
+        // The specific set of 'detail' records is fixed based on the target summary object.
+        if (summaryObjectType == Account.SObjectType) {
+
+            if (rollupType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit
+                || rollupType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit
```
RadSFDeveloper: The method is called for hard credits, but the line above has the check for the Account Soft Credit. is additional refactoring required?

src/classes/CRLP_Batch_Base.cls
```diff
@@ -134,93 +134,16 @@ public abstract class CRLP_Batch_Base {
...
-                this.detailObjectType = OpportunityContactRole.SObjectType;
-                this.keyField = 'Contact.AccountId';
-            }
-            when RecurringDonations {
-                this.detailObjectType = Opportunity.SObjectType;
-                this.keyField = SObjectType.Opportunity.fields.npe03__Recurring_Donation__c.getName();
-            }
-        }
+        this.detailObjectType = CRLP_Rollup_SVC.getDetailObjectType(jobType);
+        this.keyField = CRLP_Rollup_SVC.getParentRelationshipFieldForQueryReference(jobType);
```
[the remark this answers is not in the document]
force2b: There are two of these that return slightly different things:
* `getParentRelationshipFieldForQueryReference`
* `getParentRelationshipFieldForProcessingJob`

The only real difference is the field removed for Contact and AccountContact Soft Credits. This is because in these two rollup operations, the detail object changes from OCR (the object queried) to Partial Soft Credit (the object processed). I was trying to find a way to name these so it was clear what was being returned based on intent - "ForQuery" versus "ForProcessing"

I could try something like if think that's clearer:
* `getParentFieldNameForQuery`
* `getParentFieldNameForProcessing`

src/classes/CRLP_Batch_Base.cls
```diff
@@ -306,50 +199,65 @@ public abstract class CRLP_Batch_Base {
      * @return List<Partial_Soft_Credit__c>
      */
     protected List<Partial_Soft_Credit__c> buildPartialSoftCreditRecordsFromOCRs(List<OpportunityContactRole> detailRecords) {
+        // These are both referenced in the SOQL created by buildDetailObjectQueryForRollup().
```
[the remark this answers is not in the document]
force2b: I added the comment because IntelliJ is smart enough to recognize that the "contactIds" var is never referenced in the code here. However, that's because it's added to the SOQL statement created by `buildDetailObjectQueryForRollup`. So, when I removed the _unused_ var the first time it ended up throwing an exception and took me a few minutes to track down.

Overall I really want to finish rebuilding `CRLP_Query_SEL`, which would probably help here, but time is starting to run out on the refactoring.

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -152,123 +153,43 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
...
-                    parentIdHasDetailRecords.add((String)detailRecords[n].get(parentKeyField));
-                } else {
-                    SObject dtlRecord = detailRecords[n].getSObject(parentObjectField);
-                    parentIdHasDetailRecords.add((String)dtlRecord.get(parentKeyField));
-                }
+            // Convert the list of OpportunityContactRole detail records into a list of PartialSoftCredit
+            // records complete with the parent Opportunity details (CloseDate, Amount, ec.)
+            if (jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit ||
+                    jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit) {
+                detailRecords = buildPartialSoftCreditRecordsFromOCRs(detailRecords);
```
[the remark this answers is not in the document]
force2b: Very good suggestion

src/classes/CRLP_Batch_Base_NonSkew.cls
```diff
@@ -311,6 +223,72 @@ public abstract class CRLP_Batch_Base_NonSkew extends CRLP_Batch_Base {
         }
     }
 
+    /**
+     * @description Loop through the Summary records and remove any that do not have Detail Objects to be rolled up
+     * @param summaryRecords
+     * @param detailRecords
+     * @return a cleaned list of summary records
+     */
+    private List<SObject> cleanseSummaryRecordsList(List<SObject> summaryRecords, List<SObject> detailRecords) {
```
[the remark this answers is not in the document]
force2b: Nice re-write. How about `excludeSummaryWithMissingDetailRecords` as the method name?

src/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -348,70 +223,112 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
...
-                    Full_Message__c = ('Total Batch Iterations: ' + this.batchIteration + '\n' +
-                            'Total DML Operations: ' + this.totalCommits + '\n' +
-                            'Total Records Updates: ' + this.totalRecordsModified + '\n' +
-                            'Parent Ids Commited in Finish: ' + (this.parentIdsNotCommittedWithBatch !=  null ?
-                                '(' + this.parentIdsNotCommittedWithBatch.size() + ')\n' +
-                                        JSON.serializePretty(this.parentIdsNotCommittedWithBatch) : 'none') + '\n' +
-                            'Rollups State:\n' + CRLP_Debug_UTIL.getCurrentRollupState()).left(32768)
-            );
-            insert logEntry;
+            String additionalInfo = 'Parent Ids Commited in Finish: ' + (this.parentIdsNotCommittedWithBatch !=  null ?
```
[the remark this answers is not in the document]
force2b: Better. Thanks.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -0,0 +1,538 @@
...
+     * @description used to receive and return the last Parent record id processed by startRollupProcessing().
+     * The value is saved as a stateful var in the calling batch process
+     */
+    private Id lastParentIdProcessed;
+
+    /**
+     * @description Used to return a true to the calling LDV class when the last parent Id was removed from
+     * the list of updated SObjects
+     */
+    public Boolean lastParentIdRemovedFromUpdatesList = false;
```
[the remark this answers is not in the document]
force2b: How about `processor.wasLastParentIdCompleted()`?

src/classes/CRLP_RollupProcessor.cls
force2b: What would you think about leaving the `get` format, but changing the checks for `null` into `String.isNotBlank(parentRelationshipField)`?

src/classes/CRLP_RollupProcessor.cls
force2b: The reason that these loops use an integer iterator over a Sobject is the result of performance analysis that originated here: https://paulbattisson.com/blog/2014/force-com-cast-episode-6-faster-loops/. At the time of building CRLP I ran tests of my own and confirmed that this is actually faster. If the loop was only performed once it probably wouldn't make any difference, but given the nature of CRLP many of these loops are performed hundreds or thousands times depending on the amount of data being rolled up.

src/classes/CRLP_RollupProcessor_SVC.cls
```diff
@@ -37,297 +37,38 @@
...
     /**
      * @description Returns the Type of the appropriate CRLP_RollupXXXX_SVC class to use for handling the
      * rollups based on the SObjectType of the Summary object
-     * @param summObjType Summary Object Type
-     * @param detailObjType Detail Object Type
+     * @param rollupType Rollup Type
      * @return ApexClassType
      */
-    private static Type getHanderClassType(CRLP_RollupProcessingOptions.ProcessingOptions options) {
+    public static Type getHanderClassType(CRLP_RollupProcessingOptions.RollupType rollupType) {
```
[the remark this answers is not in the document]
force2b: Nice catch!

src/classes/CRLP_RollupQueryBuilder.cls
force2b: `customWhereClauses` was meant to allow the caller to pass a specific additional where clause into the builder that isn't automatically created by the logic in this method. It's not actually implemented anywhere at the moment though. This is a good example of where I started thinking ahead to future refactors, but decided to not take it that far quite yet.

src/classes/CRLP_RollupQueryBuilder.cls
force2b: Hmm. 🤷‍♂️

src/classes/CRLP_RollupProcessor.cls
RadSFDeveloper: Yes, the `get` format stays for the `parentRelationshipField `, and we use `String` methods when checking String throughout (for this and other String properties).

src/classes/CRLP_RollupProcessor.cls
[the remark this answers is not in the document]
RadSFDeveloper: Let's use `is`, `has` and `should` prefixes for Boolean output methods?

src/classes/CRLP_RollupProcessor.cls
RadSFDeveloper: Thanks for letting me, I was not aware of the performance issue. If so, let's be consistent in naming and reduce the variable names throughout code, for example: `for (Integer i = 0, size = detailsForParent.size(); i < size; i++) {`? Sometimes the index is `n`,  `itemNum` or other, so let's consistently name vars in `CRLP` namespace?
Similarly `detailsForParent` are just `detailRecords`?
Or if we specifically know where are getting OCRs or PSCs, let's name those detail record vars and methods as such?

src/classes/CRLP_RollupQueryBuilder.cls
RadSFDeveloper: Let's remove code if it is not required at this point?

src/classes/CRLP_RollupQueryBuilder.cls
RadSFDeveloper: It is good that refactoring is done while all tests are intact. Probably further refactoring and code extracting (if any) will be done when this class is completely tested on its own.

src/classes/CRLP_RollupProcessor.cls
force2b: Makes sense.

src/classes/CRLP_RollupQueryBuilder.cls
force2b: That's fine if you're good with not digging too much further into the query refactoring at this point.

force2b:
@RadSFDeveloper Thank you for your review. All changes made. Unit Tests still pending though. I'll probably include these in the child PR that has the API work.

force2b:
**lurch: attach W-031888

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LlbRQAS" target="_blank">W-031888</a>

force2b:
Thanks @seanpat09. I was thinking the same thing about merge conflict resolution. I get the feeling there may be at least one other branch with merge conflicts as a result of this PR.

force2b:
**lurch: attach W-031888

RadSFDeveloper:
@force2b , creating a new or editing an existing rollup does not work. It works on `master`. The error:

Screen Shot 2019-05-12 at 11 16 56 AM

force2b:
@RadSFDeveloper Interesting. This PR shouldn't have changed anything at all related to the UI. I just tested this same change in another org (same branch) and it worked just fine. Could there be another factor that is affecting the save?

RadSFDeveloper:
> @RadSFDeveloper Interesting. This PR shouldn't have changed anything at all related to the UI. I just tested this same change in another org (same branch) and it worked just fine. Could there be another factor that is affecting the save?

Thanks for verifying @force2b! The error is issued in a branch built on top of this one. When creating a new scratch org for this branch, I did not notice that checkout did not follow through due to modified files. Sorry for the false alarm.

force2b:
@RadSFDeveloper No problem. Thanks for looking into it.

rponti-sforg:
@force2b Looks like just the Feature Test - Robot Prerelease test failed.  When I click on Details, 403 Forbidden is displayed in new browser tab.  

Do we want to look into the Robot Prerelease tests, or would you prefer to just merge?  Just let me know.

force2b:
@rponti-sforg Looks like the robot pre-release failed with a missing metadata error, which has been very common on that build. It can be ignored. Feel free to merge when you're ready.

rponti-sforg:
@seanpat09 @force2b  Fyi ... this PR / branch has been merged.

PR #4162, opened 2019-04-04
Making the BGE Donation Selector global

lparrott:
- Making the BGE Donation Selector component global so GEM can access it from Single Gift Entry.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lparrott:
Hey @ptewson-sfdo, since we switched from a "Lightning Page Tabs" to a "Lightning Component Tab", we can no longer access the Donation Selector component from Single Gift Entry. How do you feel about making this component global so we can re-use it in the GEM package?

PR #4168, opened 2019-04-09
CCI Flow: Deploy_Rollup_Testing

force2b:
Change the `config_acctsc_rollups` flow to `deploy_rollup_testing` and expand to include Payment Rollups and the 'CRLP_TEST_VALIDATE_ROLLUPS` apex class used to fully validate all rollup operations using Apex Anonymous. This has been tested as a complete testing solution against the rollup engine functionality. It is a complete flow that creates Account-Contact Soft Credit Rollups, Account Soft Credit Rollups, and Payment Rollups along with the test apex class needed to validate the entire engine. 

Note: There is no change to the Cumulus package code in this PR.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


unpackaged/config/acctsc_rollups/customMetadata/Rollup.NPSP_AcctAccConSCCount.md
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomMetadata xmlns="http://soap.sforce.com/2006/04/metadata" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-    <label>Account: AccConSC - Last SC Amt</label>
+    <label>Account: AccConSC - Total Nomber Of</label>
```
davisagli: Nom nom nomber

force2b:
Thanks @davisagli. The CRLP Testing Script was already there and has been used on occasion, though primarily by me. Instructions for using the class are in the comments at the top of the class. This PR actually expands on what was there into a more complete and functional test suite. It creates all the necessary metadata to use the test script. To my knowledge nobody was using the old 'config_acctsc_rollups` flow. The new flow creates all of those Account Soft Credit rollups along with Payment Rollups, fixes the MembershipRollup Filter (for scratch orgs). One question I have is whether I should just combine the two folders (accsc_rollups and payment_rollups) into one?

davisagli:
@force2b yeah I think you're right about no one using the config_acctsc_rollups flow -- I added that as a demonstration of how to have a separate flow to handle deploying metadata selectively for QA purposes.

I agree this is a step forward and am okay with taking that step. I think the ideal end goal would be to run this as part of the ci_feature flow (so it runs during feature builds) and figure out how to have failures be reported to MetaCI similarly to other test failures. But, how much additional time would it add to the build?

Not approving the PR since someone else should probably review the Apex.

force2b:
Thanks @davisagli. I don't think this validation script will ever be one that we can automate. There are a lot of steps that take an indeterminate amount of time to complete, and each has to be executed as a separate code context - i.e., run one line at a time and wait for it to complete. It was built to be a manual validation script only. It would probably take least 10 to 15 minutes to complete, assuming that there were a way to automate.

@ptewson-sfdo Would you mind doing a quick review the apex class and approving. The class isn't in our code base. It's just in the unpackaged folder. Same class you previously reviewed. I just moved it to a different folder and fixed some problems.

davisagli:
@force2b what are the conditions in which it should be run? I'm wondering if we can at least automate reminding the developer that it exists under appropriate conditions.

force2b:
It's only needed when anything related to Customizable Rollups is modified. It doesn't currently cover every single configuration option in CRLP either. For example, it's not testing Chunk Mode processing (which might be hard with only 2 Accounts anyway) or Incremental (which also might be hard given that all records are created as part of the test). It does run Queueable, Skew and NonSkew jobs though.

The reason I'm checking in this new PR is that I was using the script to validate significant changes I'm making in another PR (crlp-api). As part of that I realized that there was a lot of setup work I had to do to get the script to run. This PR addresses that by creating the needed metadata (fields, rollups, filters) and fixes some inconsistencies I found in the test script. Basically this will get run every time we make any changes to CRLP (that's always been the case) Now it'll just be easier to run.

unpackaged/config/payment_rollups/classes/CRLP_TEST_VALIDATE_ROLLUPS.cls
```diff
@@ -680,140 +701,7 @@ public class CRLP_TEST_VALIDATE_ROLLUPS {
         rd.npe03__Next_Payment_Date__c = null;
         update rd;
     }
-
```
ptewson-sfdo: Is the idea that this removed createCustomRollups() method is being replaced by the cci task that sets up the rollups for you?

force2b: Yes. Most of that cci task metadata already exists. I just added a few extra types to it.

PR #4169, opened 2019-04-09
Updating the Donation Selector component to have global attributes

lparrott:
- Updating the Donation Selector component to have global attributes so GEM can access them.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


lparrott:
@ptewson-sfdo Making this component global got me halfway there... I still can't set the attributes from GEM without making them global 👎 

Let me know if you have questions or concerns about this change!

PR #4176, opened 2019-04-12
Feature/tdtm mutable dmlwrapper

rogeramitchell:
# Context

TDTM allows for users to create execution chains of logic, however customer implementations of TDTM are unable to interact with any aspects of the `TDTM_Runnable.DmlWrapper` collection that will have inserts, updates, and deletes performed. This creates a critical issue where customer implementations of TDTM must workaround any situation where NPSP's TDTM implementations have already added a record into the `objectsToUpdate` collection; not handling this will present a duplicate record in collection exception, and handling this often means leveraging asynchronous updates or performing DML separately from `TDTM_TriggerHandler`. Furthermore, any customer implementations of TDTM might break if NPSP updates its TDTM logic to start adding a record into an `objectsToUpdate` (which was not encountered when the customer implementation was initially designed and deployed).

This PR enables NPSP and customer implementations of TDTM to gain access to the global `DmlWrapper` that `TDTM_TriggerHandler` maintains during a transaction. The feature is considered "opt in" because it relies upon an entirely new interface to implement.

**Example:**

- Customer wishes to set fields on Account when a Contact is created in the `AfterInsert` context
- Customer adds the Contact's Account into `DmlWrapper.objectsToUpdate` and returns this from their `TDTM_Runnable.run` implementation
- When DML is performed, Salesforce throws an exception because an NPSP TDTM implementation had already added that Account into `DmlWrapper.objectsToUpdate`
- Customer updates their class to implement `TDTM_RunnableMutable`, checks whether the Account retrieved is part of `DmlWrapper.objectsToUpdate`, and either updates that record or places the Account retrieved into that collection

**Benefits:**

- Existing TDTM implementations will not break, as any existing classes implement the `TDTM_Runnable` interface
- Classes implementing the `TDTM_RunnableMutable` interface can be smaller implementations of code, as developers can reliably retrieve a record that was processed in a prior context from the transaction's global `DmlWrapper`
- Customers are able to adopt TDTM for use cases that were previously managed with a different framework or workaround

**Risks:**

- Mutating the transaction's `DmlWrapper` collection could remove records for insertion or update, or overwrite previous changes by putting a record into the collection that does not reflect changes made from TDTM logic fired previously in transaction

**Thought Leaders + Contributors:** @njjc, @force2b, @matthewpoe

# Critical Changes

1. `TDTM_TriggerHandler` class
	1. Updated `runClass` to accept an additional parameter, which is the `TDTM_Runnable.DmlWrapper` declared in the `run` method
	1. Updated `runClass` to check whether `classInstance` is an instance of `TDTM_RunnableMutable`, in which case that class's `run` method is called

# New Metadata

1. `TDTM_RunnableMutable` class
	1. Has global access to allow developers to implement this interface for creating TDTM logic that has access to the transaction's `DmlWrapper` collection
1. `TDTM_RunnableMutableExample_TEST` class
	1. Required for test coverage to demonstrate how the same class can be called multiple times, and where that class is responsible for checking whether `dmlWrapper` already contains a record for update
1. `TDTM_RunnableMutable_TEST` class
	1. Verifies that an implementation of `TDTM_RunnableMutable` interacts with the transaction's `DmlWrapper` within the method, and expected updates based on iteratively running the same class multiple times to increment `Account.NumberOfEmployees`


coriobriensfdo:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmWrQAK" target="_blank">W-032586</a>

coriobriensfdo:
**lurch: detach

LurchTheButler:
Detached <a href="https://foundation.my.salesforce.com/a2x1E000008LmWrQAK" target="_blank">W-032586</a>

coriobriensfdo:
**lurch: attach W-030244

coriobriensfdo:
Thank you for your contribution @rogeramitchell & @njjc!!

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQ7DQAU" target="_blank">W-030244</a>

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -182,7 +182,19 @@ public class TDTM_TriggerHandler {
                                 describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');
                         return classToRun.run(newlist, oldlist, thisAction, describeObj);
                     }
-	        	} else {
+	        	} else if(classInstance instanceof TDTM_RunnableMutable) {
```
force2b: @rogeramitchell I'm confused on this. It looks like this `if` is in the `else` that would run the old code. I would expect the check of `} else if (classInstance instanceof TDTM_RunnableMutable) {` to be after the call to `runAsync()`.

src/layouts/Trigger_Handler__c-Trigger Handler Layout.layout
```diff
@@ -99,4 +99,10 @@
     <showInteractionLogPanel>false</showInteractionLogPanel>
     <showRunAssignmentRulesCheckbox>false</showRunAssignmentRulesCheckbox>
     <showSubmitAndAttachButton>false</showSubmitAndAttachButton>
+    <summaryLayout>
```
force2b: Can you remove this file from the PR?

src/objects/Trigger_Handler__c.object
```diff
@@ -48,6 +48,7 @@
     <enableBulkApi>true</enableBulkApi>
```
force2b: Can you remove this file from the PR?

src/package.xml
```diff
@@ -1,2723 +1,2725 @@
...
-        <members>Addr_Verify_Endpoint</members>
-        <members>Addr_Verify_Google_Help</members>
-        <members>Addr_Verify_Settings</members>
-        <members>Addr_Verifying</members>
-        <members>Address_Not_Found</members>
-        <members>Address_StateCountry_Invalid_Error</members>
-        <members>Address_Verification_Limit</members>
-        <members>CMT_DeleteConfirm</members>
-        <members>CMT_FilterGroupActiveRollupsHelpText</members>
-        <members>CMT_FilterGroupDelet
```
force2b: Can you remove this file from the PR, with the exception of the three new classes?

src/classes/TDTM_RunnableMutable_TEST.cls
```diff
@@ -0,0 +1,69 @@
...
+        insert UTIL_UnitTestData_TEST.CreateMultipleTestContacts(3);
+    }
+
+    /*************************************************************************** 
+    * @description Verifies that when calling the same Trigger Handler multiple
+    * times where that class implements TDTM_RunnableMutable
+    **/
+    @isTest
+    private static void testAllowMutation()
+    {
```
force2b: Can you move the `{` up to the method?

src/classes/TDTM_RunnableMutableExample_TEST.cls
```diff
@@ -0,0 +1,60 @@
...
+ * @group-content ../../ApexDocContent/TDTM.htm
+ * @description Sample implementation of TDTM_RunnableMutable to allow for testing
+ * expected behavior and assisting with test coverage of TDTM_RunnableMutable. For
+ * each time this class runs, the expectation is that the Account's NumberOfEmployees
+ * field will increment by 1.
+ */
+@isTest
+public class TDTM_RunnableMutableExample_TEST extends TDTM_RunnableMutable {
+    
+    public override void run(List<SObject> newList, List<SObject> oldList,
```
force2b: Do you think this could be accomplished with a `System.mock` instead of as a separate class?

src/objects/Trigger_Handler__c.object
rogeramitchell: @force2b done

src/classes/TDTM_RunnableMutableExample_TEST.cls
rogeramitchell: @force2b I think this implementation is more appropriate considering `Test.setMock` is typically reserved for callouts.

src/classes/TDTM_TriggerHandler.cls
rogeramitchell: @force2b it's helpful to see the full extent of the code here. Existing classes that implement `TDTM_Runnable` will fall into the `if` block on line 167, and will either have `runAsync` called or the instance of that class's `run` method. The `else if` block handles classes implementing the new `TDTM_RunnableMutable` interface, which only has one method `run`.

```apex
    private static TDTM_Runnable.DmlWrapper runClass(SObject classToRunRecord, List<Sobject> newList, List<Sobject> oldList, 
    TDTM_Runnable.Action thisAction, Schema.DescribeSobjectResult describeObj, TDTM_Runnable.DmlWrapper dmlWrapper) {
    	        
        if(classToRunRecord != null) {
	        String classToRunName = String.valueOf(classToRunRecord.get('Class__c'));
	        Type classType = Type.forName(classToRunName);
	        
	        if(classType != null) {   	
	        	Object classInstance = classType.newInstance();
	        	
	        	//The objects we create dynamically need to implement the TDTM_Runnable interface
	        	if(classInstance instanceof TDTM_Runnable) {
			        TDTM_Runnable classToRun = (TDTM_Runnable)classInstance;
			        Boolean async = Boolean.valueOf(classToRunRecord.get('Asynchronous__c'));
			        boolean isBefore = (thisAction == TDTM_Runnable.Action.BeforeInsert || 
			             thisAction == TDTM_Runnable.Action.BeforeUpdate || thisAction == TDTM_Runnable.Action.BeforeDelete); 
			        
			        //A class can only run asynchronously (calling a future) if we have the ids of the
                    //records we want to affect. This means we can only run it for after in insert. 
                    //Also,a future cannot be called from a future or batch method.
                    if(async && !isBefore && !System.isFuture() && !System.isBatch()) {
                        UTIL_Debug.debug(LoggingLevel.WARN, '****Calling asynchronously: ' + classToRunName + ' for ' +
                                describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');
                        runAsync(classToRun, classToRunName, newList, oldList, thisAction, describeObj);
                    } else {
                        UTIL_Debug.debugWithInfo('****Calling synchronously: ' + classToRunName + ' for ' +
                                describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');
                        return classToRun.run(newlist, oldlist, thisAction, describeObj);
                    }
	        	} else if(classInstance instanceof TDTM_RunnableMutable) {
                    // alternatively, objects created dynamically can implement the TDTM_RunnableMutable
                    // interface if the class intends to directly modify the global DmlWrappper
                    TDTM_RunnableMutable classToRun = (TDTM_RunnableMutable) classInstance;
                    UTIL_Debug.debugWithInfo('****Calling synchronously: ' + classToRunName + ' for ' +
                        describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');

                    // we do not check whether class is identified as Asynchronous__c because the
                    // underlying assumption to mutating DmlWrapper is that the class modifies
                    // the collection directly, and TDTM_TriggerHandler.run will perform DML on
                    // whatever exists in the collection after iterating through all classes to run
                    classToRun.run(newList, oldList, thisAction, describeObj, dmlWrapper);
                } else {
	        		UTIL_Debug.debug(LoggingLevel.WARN, '****The class does not implement the required interface.');
	        	}
	        } else {
	        	UTIL_Debug.debug(LoggingLevel.WARN, '****ClassType was null.');
	        }
        } else {
        	UTIL_Debug.debug(LoggingLevel.WARN, '****ClassToCall was null');
        }
        return null;
    }
```

src/package.xml
rogeramitchell: @force2b done!

src/classes/TDTM_RunnableMutable_TEST.cls
rogeramitchell: @force2b done!

src/layouts/Trigger_Handler__c-Trigger Handler Layout.layout
rogeramitchell: @force2b done!

rogeramitchell:
@force2b reviewed your requested changes, replied to each, and removed `testdata/dev_org/test_data.db`.

src/classes/TDTM_RunnableMutableExample_TEST.cls
force2b: Sorry, mean to says `System.stub`. I was thinking more about using `System.StubProvider' to mock the class and functionality. In the end though it might still come out the same and not provide any extra capability since both are extending from a global abstract class. A mocked version of this would look pretty much the same as this class that extends the global abstract version. Maybe the only advantage might be that you could have this as an inner class in the `TDTM_RunnableMutable_TEST` class.

force2b:
@rogeramitchell I think when removing some of the extraneous files from the PR they ended up getting removed from the branch entirely. For example, the PR is trying to delete the `Trigger_Handler__c` object, its related layout, etc. from the repo instead of just not having any reference to those files in the PR.

rogeramitchell:
@force2b added back `Trigger_Handler__c.object` and `Trigger_Handler__c-Trigger Handler Layout.layout` from Cumulus/master, which resolves the diff that was included as part of this PR.

src/classes/TDTM_RunnableMutableExample_TEST.cls
rogeramitchell: @force2b keeping it as a separate class will likely cut down effort for your documentation team, as it exists as a standalone example (provided you update scope to `global` and remove `@isTest`.

rogeramitchell:
@force2b the requested changes have been made to this PR. What are our next steps?

force2b:
@rogeramitchell We reviewed this yesterday in backlog refinement, so it is slowly moving forward. I have some other more concepts I want to be sure of as we figure out how this will be clearly documented, but otherwise we should see it move in the new couple of weeks. 

Full disclosure on our process though. Once we bring this into our sprint to work on, this PR will get closed (without merging) and we'll create a new feature branch for the work so it can fit into our CI processes.

rogeramitchell:
@force2b completely makes sense. If there's a need for community involvement on the QA and documentation side, let me know.

src/classes/TDTM_RunnableMutable.cls
```diff
@@ -0,0 +1,25 @@
...
+     * @description Method to be implemented by the classes the extend this abstract class. This is the method
+     * that will be called by the trigger handler.
+     * @param newlist The records that were passed to the trigger as trigger.new.
+     * @param oldlist The records that were passed to the trigger as trigger.old.
+     * @param action The event (before insert, after insert,...) that caused the class to run.
+     * @param objectResult The type of SObject the class runs for.
+     * @param dmlWrapper the global TDTM_Runnable.DmlWrapper collection that is passed from TDTM_TriggerHandler
+     * @return void
+    */
+    global abstract void run(List<SObject> newlist, List<SObject> oldlist, TDTM_Runnable.Action triggerAction,
```
force2b: @rogeramitchell  I'm just noticing that this is a void method. That would imply that the dmlWrapper collection passed to this is collection is never returned and the customer is responsible for implementing their own version of the TDTM_TriggerHandler.processDml() method. The objective should be for the customer implementation to only handle the fixing of the dml wrapper to avoid duplicate entries.  Why not just return the _fixed_ dmlWrapper collection and allow the Trigger handler to use the processDml() method to handle inserts/updates/deletes/undeletes as normal? Without doing that, the customer implementation would lose all of the functionality that the package has built in around handling dml and parsing errors.

force2b:
Closing this as it has been replaced by #4372

PR #4180, opened 2019-04-15
Enforce pay date and scheduled dates on payments. 

KietTruongSFDO:
# Critical Changes

# Changes
* We've created a new setting, “Enforce Accounting Data Consistency”, that requires users to set a Payment Date on paid payments and a Scheduled Date on unpaid payments. It can be found in Setup > Custom Settings > Contacts And Orgs Settings. This setting is also likely to control additional data validation and automation in the future.
# Issues Closed

# New Metadata
* Added a new field on "Contact and Orgs Setting" called "Enforce Accounting Rules"
* Added a new validation rule on Payment called "Require_Pay_Date_on_Paid_Payments"
* Added a new validation rule on Payment called "Require_Scheduled_Date_on_Unpaid_Payment"

# Deleted Metadata


force2b:
@KietTruongSFDO Using a validation rule isn't a pattern that we generally recommend in the NPSP and/or underlying packages. For example, in this case any customer can easily disable the Validation Rule in which case the Custom Settings that is supposed to control this wouldn't actually work as expected. Not to mention that you cannot guarantee that the VR's are active or functioning properly from a unit test since they cannot be controlled other than via the Custom Setting. 

We are just going through a relatively similar validation on the Payment object. In this case, it's important that both the Paid and Written-Off checkboxes are not checked at the same time. Though technically it would be possible to implement this as a validation rule, we quickly ruled that out because of the potential for customers being able to disable that rule - which would then cause all sort of accounting issues. Instead, this change will be implemented directly in the Payments trigger. I highly recommend you take the same approach for this. More information can be found at  [W-032340.](https://foundation--agf.na70.visual.force.com/a2x1E000008LmCcQAK)

cc @pedrodalcol

KietTruongSFDO:
Hi @force2b, thanks for letting me know. I'll update the code accordingly and refrain from using validation rules in the future.

force2b:
Thanks @KietTruongSFDO. Can you also @ mention the Wi for this on the W-032340 WI just in case work begins at the same time (to avoid/handle conflicts)? Thanks.

markross-sfdo:
@KietTruongSFDO I've updated the Changes text here. Please verify if it looks good!

jbrandolinisf:
Closing this pull request as invalid.  It was created with validation rules, rather than triggers.

PR #4189, opened 2019-04-19
Feature/BDI-CO migration utility

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,342 @@
...
+            String jobId = getUniqueJobId();
+            CMT_MetadataAPI.deployMetadata(jobId, queuedMetadataTypesForDeploy.clone(), true);
+
+            return jobId;
+        }
+
+        // Queue is empty, but we still need to update Data Import settings
+        updateCustomDataImportSettings('Data Import Field Mapping');
+
+        return 'Queue is empty';
```
ptewson-sfdo: Is this worth a custom label?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,342 @@
...
+                    pendingFieldMappings.add(fieldMapping.getMetadataRecord());
+                }
+            }
+        }
+
+        // Don't enqueue custom help text mappings
+        //queuedMetadataTypesForDeploy.addAll(pendingFieldMappings);
+
+        // Debugging
+        System.debug(pendingFieldMappings.size());
```
ptewson-sfdo: We want to pull out the debugs before merging for sure

src/classes/CMT_MetadataAPI.cls
```diff
@@ -99,6 +99,15 @@ public class CMT_MetadataAPI {
      * @param mtToDeploy List of CustomMetadata objects to deploy
      */
     public static void deployMetadata(String jobId, List<Metadata.CustomMetadata> mtToDeploy) {
```
psadie: @ptewson-sfdo - Do you think overloading the deployMetadata method to add a flag for the result handler to use later is a good idea? HandleResult method needs a way to know which custom settings to update

ptewson-sfdo: Ah I see - I think there may be a better way to handle this but I will have to chew on it. At very least I think there may be a better name for that variable than "isHelpTextReleated", but there's some options ... if you can hunt down the author of this comment:

// THIS WILL NEED TO BE REFACTORED IF THIS CMT DEPLOYMENT CODE IS USED BY SOMETHING OTHER THAN ROLLUPS

... maybe they would have ideas? One option would actually be to define a class hierarchy for "Metadataday callback support" or whatever but that feels like massive overkill, so maybe just passing in a variable to indicate what we are doing there like you are already is best. 

Would that make sense as an enumeration instead of a boolean just in case we support additional types in the future? And maybe the contents of the separate branches could move into their own methods just so it's a bit more obvious about how that's getting handed off?

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -0,0 +1,249 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Rollups
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Wrapper classes for the Help Text CustomMetadataType objects used to create the
+* default Custom Metadata records
+*/
+public class BDI_HelpTextCMT {
```
ptewson-sfdo: Do we want to name this "HelpText"? Isn't it more like a "FieldMappingCMT"? I want to make sure these names make sense long after the help text mapping is a forgotten memory

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -0,0 +1,249 @@
...
+            this.isDeleted = false;
+        }
+
+        public Metadata.CustomMetadata getMetadataRecord() {
+            Metadata.CustomMetadata md = new Metadata.CustomMetadata();
+
+            String mdTypeName = UTIL_Namespace.StrTokenNSPrefix(BDI_HelpTextCMT.MetadataObject.Data_Import_Field_Mapping.Name() + '__mdt');
+
+            // Setup custom metadata to be created or updated in the subscriber org.
+            Metadata.CustomMetadata customMd = new Metadata.CustomMetadata();
```
ptewson-sfdo: Do we need both the 'md' and the 'customMd' instances? What's the thinking there?

src/classes/BDI_HelpTextCMT.cls
ptewson-sfdo: (As I'm reading through this, the naming is making more sense! Still might have thoughts, but the original reaction is likely not going to hold)

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -0,0 +1,249 @@
...
+
+        if (name.length() > 30) {
+            String firstFifteen = name.substring(0, 15);
+            String lastFifteen = name.substring(name.length() - 15, name.length());
+            name = firstFifteen + lastFifteen;
+        }
+
+        if (makeUnique == true) {
+            // Create a GUID format unique string
+            Blob b = Crypto.GenerateAESKey(128);
```
ptewson-sfdo: Is the key generated by GenerateAESKey() guaranteed unique? Or do we run the risk of incredibly-unlikely duplicates?

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -0,0 +1,249 @@
...
+            name = name.left(30).removeEnd('_') + '_' + h.left(9);
+        } else {
+            name = name.left(40);
+        }
+
+        return name;
+    }
+
+    /** @description Map of object names from the data import help text field to actual object names */
+    private static final Map<String, String> DI_OBJECT_MAP = new Map<String, String>{
```
ptewson-sfdo: Even though most of our static finals are all caps this seemed a bit unusual for a collection (instead of a string or integer). But I honestly don't know why this wouldn't follow the convention of other static finals, aka essentially constants

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * @description Utility to convert a name into a valid 'DeveloperName' quality value (or unique GUID)
+     * for the CMT record
+     * @param name String
+     * @param makeUnique Boolean
+     * @return DeveloperName format string
+     */
+    private static String generateRecordName(String name, Boolean makeUnique, String objectLabel, String objectAPIName) {
```
ptewson-sfdo: The names this generates in our mappings from this method look really good - I think we landed on a great solution here!

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,349 @@
...
+
+    /**
+     * @description Go through all fields in Data Import object and collect relevant help text mappings
+     * @return A collection of help text field names mapped to an instance of MappingDetail which holds relevant values for building
+     * the corresponding Metadata Type record (i.e. Source Object Name, Target Object Name, Target Field API Name, etc)
+     */
+    public Map<String, BDI_HelpTextCMT.MappingDetail> getAllMappedHelpTextFields() {
+        Map<String, BDI_HelpTextCMT.MappingDetail> helpTextFieldMap = new Map<String, BDI_HelpTextCMT.MappingDetail>();
+        Map<String, BDI_HelpTextCMT.ImportStatusAndLookupParts> importStatusAndLookupMap = new Map<String, BDI_HelpTextCMT.ImportStatusAndLookupParts>();
+        Map<String, Schema.DescribeFieldResult> allDFRs = UTIL_Describe.getAllFieldsDescribe('npsp__DataImport__c');
```
ptewson-sfdo: Move 'npsp__DataImport__c' to a defined constant? Also looks like this might fail in non-namespaced orgs (although I pretty much universally fail to predict that accurately)

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+     * @description Queries for default Object Mappings and builds a map of DeveloperName to Object Mapping record
+     * Used to check for existance of Object Mapping record without loop
+     * @param objectMappingSetId Id of the default Object Mapping Set
+     * @return Collection of Object Mapping DeveloperName to Object Mapping record
+     */
+    private Map<String, Data_Import_Object_Mapping__mdt> getDefaultObjectMappings(Id objectMappingSetId) {
+        List<Data_Import_Object_Mapping__mdt> existingObjectMappings = [SELECT Id,
```
ptewson-sfdo: I can't remember exactly what the standard says, but I think we generally don't tab the field lists quite so far over? I think it's either 2 tabs for the fields and 1 tab for the FROM/WHERE etc, or maybe 1 additional tab in from that

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,349 @@
...
+            if(defaultFieldMappingSet == null) {
+                return getDefaultFieldMappingSet();
+            }
+            return defaultFieldMappingSet;
+        } private set;
+    }
+
+    /** @description Queries for default Object Mapping Set, assume existence */
+    private static Data_Import_Object_Mapping_Set__mdt getDefaultObjectMappingSet() {
+        return [SELECT Id, DeveloperName, Label FROM Data_Import_Object_Mapping_Set__mdt WHERE DeveloperName = 'Default_Object_Mapping_Set'];
```
ptewson-sfdo: Can we move 'Default_Field_Mapping_Set' to a class-level variable or defined constant?

ptewson-sfdo: And 'Default_Object_Mapping_Set'

src/pages/STG_SettingsManager.page
```diff
@@ -136,6 +136,8 @@
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelRelNav"><a href="#" onclick="ShowPanel('idPanelRel');return false;" class="slds-grow slds-m-left_large">{!$Label.stgNavRelationships}</a></li>
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelRelReciprocalNav"><a href="#" onclick="ShowPanel('idPanelRelReciprocal');return false;" class="slds-grow slds-m-left_large">{!$Label.stgNavRelReciprocal}</a></li>
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelRelAutoNav"><a href="#" onclick="ShowPanel('idPanelRelAuto');return false;" class="slds-grow slds-m-left_large">{!$Label.stgNavRelAutoCreate}</a></li>
+                                <!-- HELP TEXT MAPPING -->
```
ptewson-sfdo: If we can figure out where we want to slot this in the settings now, maybe let's see if we can figure that out

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,349 @@
...
+            diFieldNameToFieldMappingMap.put(fieldMapping.Source_Field_API_Name__c, fieldMapping);
+        }
+
+        return diFieldNameToFieldMappingMap;
+    }
+
+    /** @description Default Object Mapping Set record */
+    public static Data_Import_Object_Mapping_Set__mdt defaultObjectMappingSet {
+        get {
+            if(defaultObjectMappingSet == null) {
```
ptewson-sfdo: I think our code standard demands a space between "if" and the paren - so "if (defaultObjectMappingSet == null) {"

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,349 @@
...
+                                                                            Required__c,
+                                                                            Source_Field_API_Name__c,
+                                                                            Target_Field_API_Name__c,
+                                                                            Target_Object_Mapping__c,
+                                                                            Is_Deleted__c
+                                                                        FROM Data_Import_Field_Mapping__mdt
+                                                                        WHERE Data_Import_Field_Mapping_Set__c = :fieldMappingSetId];
+        Map<String, Data_Import_Field_Mapping__mdt> diFieldNameToFieldMappingMap = new Map<String, Data_Import_Field_Mapping__mdt>();
+
+        for(Data_Import_Field_Mapping__mdt fieldMapping : existingFieldMappings) {
```
ptewson-sfdo: I think our code standard demands a space between "for" and the paren - so "for (Data_Import_Field_Mapping__mdt fieldMapping : existingFieldMappings) {"

ptewson-sfdo:
**lurch: attach W-032582

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmWXQA0" target="_blank">W-032582</a>

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     * @description Queries for default Field Mappings and builds a map of DeveloperName to Field Mapping record
+     * Used to check for existance of Field Mapping record without loop
+     * @param fieldMappingSetId Id of the default Field Mapping Set
+     * @return Collection of Field Mapping DeveloperName to Field Mapping record
+     */
+    private Map<String, Data_Import_Field_Mapping__mdt> getDefaultFieldMappings(Id fieldMappingSetId) {
+        List<Data_Import_Field_Mapping__mdt> existingFieldMappings = [SELECT Id,
```
ptewson-sfdo: This isn't going to block here, but can we format these SOQL queries .... slightly less far to the right? I think there's some guidance in the style guide

PR #4197, opened 2019-04-23
Outbound requests to retrieve rollup states from the DSO-NPSP bridge package

seanpat09:
# Critical Changes

# Changes
* Changes related to integration with the forthcoming Data Storage Optimizer product:
    * Adds integration API to retrieve archived rollup state data from the DSO-NPSP bridge package
    * Includes rollup state data fetched from Bridge in CRLP calculations
    * If the Org has the DSO-NPSP package installed and Customizable rollups are enabled in NPSP, then changes to the Customizable Rollup definitions will mark the related rollup type(s) (Account Hard Credit / Contact soft Credit / etc) as _stale_ in the DSO-NPSP Bridge.

# Issues Closed

# New Metadata
Apex Classes:
* UTIL_IntegrationConfig.cls
* UTIL_IntegrationConfig_TEST.cls
* CRLP_RollupCMT_TEST

# Deleted Metadata


src/classes/UTIL_IntegrationConfig.cls
```diff
@@ -0,0 +1,147 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @description Config for external packages
+*/
+public abstract with sharing class UTIL_IntegrationConfig {
+    @TestVisible
+    private static final String DEFAULT_NPSP_DSO_BRIDGE_NAMESPACE = 'dso_npsp';
```
seanpat09: I went with hardcoding this since I can't think of a an actual scenario (outside of our development) where the bridge is installed unmanaged since it is closed source. Is this ok or should we try to put this in a custom setting?

src/classes/UTIL_IntegrationConfig.cls
```diff
@@ -0,0 +1,147 @@
...
+        */
+        public DsoNpspPackage() {
+            callableApiClassName = 'BridgeCallable_API';
+        }
+
+        /**********************************************************************************************************
+        * @description Checks if the package is installed
+        */
+        protected override Boolean initIsInstalled() {
+            return System.Type.forName(prefixNamespace('ArchiveRollupSummary__b')) != null;
```
seanpat09: Where this kind of falls flat is if the customer decides to create their own `ArchiveRollupSummary__b` - then this would return true if the bridge was not installed. Is there a better way to check if an managed app is installed?

seanpat09:
**lurch: attach W-032420

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmJEQA0" target="_blank">W-032420</a>

src/classes/UTIL_IntegrationConfig.cls
patrick-yan-sf: You can look at UserInfo. isCurrentUserLicensed( string namespace) - that will tell you if the namespace is there.  It will throw an exception if the namespace doesn't exist. For unmanaged setups, you can't have two objects called ArchiveRollupSummary__b anyway.

src/classes/Integration_API.cls
```diff
@@ -0,0 +1,76 @@
...
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description Integration from NPSP to defined external API's.
+*/
+public with sharing class Integration_API {
+    @TestVisible
+    private static final String DSO_BRIDGE_GET_ROLLUP_STATE_ACTION = 'dso.getRollupStateForRecords';
```
RadSFDeveloper: Can we rename the constant and its content to be more generic? NPSP will send request to an archive bridge solution to get the base rollup state. Also, we do not have a prefix for the Callable API calls in the archive solution nor its bridge Callable. 
How about `ACTION_ARCHIVE_BRIDGE_GET_ROLLUP_STATE = 'getRollupStateForRecords';`?

src/classes/Integration_API.cls
```diff
@@ -0,0 +1,76 @@
...
+
+    @TestVisible
+    private UTIL_IntegrationConfig dsoNpspPackage {
```
RadSFDeveloper: What do you think about making this class more generic without using namespaces in the names (if possible at all)? `archiveBridgeConfig`, or `archiveBridgePackage`?

src/classes/Integration_API.cls
```diff
@@ -0,0 +1,76 @@
...
+        return dsoNpspPackage.isInstalled();
+    }
+
+    /**
+    * @description Confirms if Rollup State is enabled by checking if the DSO-NPSP bridge is installed
+    * @param recordIds Ids of records that you want to retrieve archived rollup data for
+    * @param rollupType The rollup type that you want to retrieve archived rollup data for
+    * @return Map<Id, List<CRLP_Rollup>> List of CRLP_Rollup data for each specified record Id
+    */
+    public Map<Id, List<CRLP_Rollup>> getBaseRollupStateForRecords(Set<Id> recordIds, CRLP_RollupProcessingOptions.RollupType rollupType) {
```
RadSFDeveloper: I think we need to call `isRollupStateEnabled()` in the method to prevent calls if the package is not installed. Thoughts?

src/classes/Integration_API_TEST.cls
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * @description Builds mock rollups
+     */
+    private static List<CRLP_Rollup> buildMockRollups() {
+        String filterGroupId = CMT_UnitTestData_TEST.getNewRecordId();
+
+        String rollupsJSON = '[' +
+
+            CMT_UnitTestData_TEST.createRollupRecord('Total Donations All Time', filterGroupId,
+                    CMT_UnitTestData_TEST.RollupRecordType.OppToAcct,
```
RadSFDeveloper: Indentation is off (8 spaces).

src/classes/UTIL_IntegrationConfig.cls
```diff
@@ -0,0 +1,147 @@
...
+
+        return namespace + '__' + entity;
+    }
+
+    public class DsoNpspPackage extends UTIL_IntegrationConfig {
+        /**********************************************************************************************************
+        * @description Constructor
+        */
+        public DsoNpspPackage() {
+            callableApiClassName = 'BridgeCallable_API';
```
RadSFDeveloper: What do you think about naming a class without underscores, ie `BridgeCallableApi` (or `BridgeCallableAPI`)?

src/classes/UTIL_IntegrationConfig.cls
RadSFDeveloper: Can we support unmanaged external packages as well? Other package solutions have Custom Settings to support both managed and unmanaged packages as well as change of the namespace. Is it overhead to use the same approach here or there is a more elegant way to support (un)managed packages?

src/classes/Integration_API.cls
```diff
@@ -0,0 +1,76 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description Integration from NPSP to defined external API's.
+*/
+public with sharing class Integration_API {
```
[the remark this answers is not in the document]
RadSFDeveloper: There are good reasons to use either `CRLP_ApiService` or `Integration_API` work for the outbound calls, both work. The `CRLP_ApiService` already indicates calls are applicable to rollups only which in this case seems more applicable instead of using generic `Integration_API`. @force2b , thoughts?

src/classes/UTIL_IntegrationConfig.cls
```diff
@@ -0,0 +1,147 @@
...
+        */
+        protected override Boolean initIsInstalled() {
+            return System.Type.forName(prefixNamespace('ArchiveRollupSummary__b')) != null;
+        }
+
+        /**********************************************************************************************************
+        * @description Determines the package namespace
+        */
+        protected override String initNamespace() {
+            if (System.Type.forName(DEFAULT_NPSP_DSO_BRIDGE_NAMESPACE + '__ArchiveRollupSummary__b') != null) {
```
RadSFDeveloper: Can the constant be pulled into the inner class (without `static` or used here as `final String` local var)?

src/classes/Integration_API.cls
seanpat09: That's a good idea - I was thinking the implementation would first call `isRollupStateEnabled()` then `getBaseRollupStateForRecords`, but it would nice to get rid of that overhead

src/classes/UTIL_IntegrationConfig.cls
seanpat09: Actually, good catch, `BridgeCallableApi` is the name of the class in DSO-NPSP: https://github.com/SalesforceFoundation/DSO-NPSP/pull/4/files#diff-f65a45e6c2e8a4b24105d50c33c16292R13

src/classes/Integration_API.cls
seanpat09: I like using `ACTION_ARCHIVE_BRIDGE_GET_ROLLUP_STATE`, I'll make the update. As for the action string itself, in https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/Callable_API.cls, the actions have the prefixes, so I was just following that pattern. @force2b do we have a preference?

src/classes/Integration_API_TEST.cls
```diff
@@ -0,0 +1,237 @@
...
+        System.assert(actualRollupType instanceof String, 'The RollupType param value should be a String');
+        System.assertEquals(expectedRollupType, actualRollupType,
+            'The correct RollupType should be passed as the RollupType param value');
+    }
+
+    /**
+     * @description Confirms that getBaseRollupStateForRecords returns an empty map if not installed
+     */
+    @isTest
+    private static void getBaseRollupStateForRecordsShouldReturnAnEmptyMapIfNotIntsalled() {
```
RadSFDeveloper: Typo in the test name.

src/classes/Integration_API.cls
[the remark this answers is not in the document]
force2b: Great points. I agree that there are probably good reasons for either, however given that the new functionality is very specific to CRLP (and wouldn't be used outside of that context), I think this needs to be split out a bit.

* `Integration_API` - This could be a generic class that enables calling _some_ external Callable api. It might (eventually) support retrieving the api from a CMT object, but for now it can hardcode the api reference. Maybe the constructor would accept some value (enum?) to specify what type of api service it should return (when it can eventually return others).
* `CRLP_ApiService` - This class should contain any of the method specific to CRLP logic, such as the `getBaseRollupStateForRecords` method.

src/classes/UTIL_IntegrationConfig.cls
force2b: @seanpat09 They way that was addressed in DSO was to check for two objects rather than just one, figuring that the chances of a customer creating two custom objects of their own with the exact same name of the objects in our package are really slim. Alternatively you could check for a field in an object (using describe).

seanpat09:
@RadSFDeveloper @force2b @patrick-yan-sf 
Thanks you all for the feedback. I removed the `Integration_API` class and moved the logic into the `CRLP_ApiService` class. I also added a static method to `UTIL_IntegrationConfig` to build the correct `UTIL_IntegrationConfig` extension based on an enum. I also added additional checks for the bridge installation. Can you review again?

src/classes/UTIL_IntegrationConfig.cls
```diff
@@ -0,0 +1,184 @@
...
+* @group API
+* @description Config for external packages
+*/
+public abstract with sharing class UTIL_IntegrationConfig {
+    public Enum Integration {
+        ArchiveBridge
+    }
+
+    @TestVisible
+    private static final String DEFAULT_ARCHIVE_BRIDGE_NAMESPACE = 'dso_npsp';
```
force2b: @seanpat09  There is a protected Custom Settings object (`Package_Settings__c`) that has the NPSP Namespace in it. I'm pretty sure this was added to allow for GEM testing with unmanaged NPSP code (though there is an `Advancement_Namespace__c` field on the `Contacts_Organizations_Settings__c` object). Could we add a field to the `Package_Settings__c` object for `DSO_Bridge_Namespace__c` to hold this value dynamically to allow for unmanaged code testing, or is that overkill for this because we could just modify the this class in an unmanaged context to change the namespace?

seanpat09: I wanted to avoid adding another custom setting and figured that a combination of the `initNamespace` method and the `isInstalled` method could dynamically handle both the managed and unmanaged bridge installations.

* If the managed bridge package is installed, then `initNamespace` would use the hard coded namespace `dso_npsp` and `isInstalled` would return true

* If the unmanaged bridge package is installed, `initNamespace` would use a blank namespace since it won't find `dso__ArchiveRollupSummary__b`. Then `isInstalled` would check for a non-namespaced `ArchiveRollupSummary__b` object and for the `RollupType__c` field, find both and return true

* If it is not installed at all, then `isInstalled` shouldn't find the `ArchiveRollupSummary__b` at all and return false

So I think that covers handling both managed and unmanaged packages. The only thing it doesn't allow for is defining a custom namespace for the bridge package, but I don't think that's necessary, at least at this point.

seanpat09:
@RadSFDeveloper thanks for catching the issues with parsing the callable action's return value! For some reason I thought it was returning a serialized string instead of a `Map<Id, String>`. I changed the casting of the return value and it looks like it's working now.

src/classes/UTIL_IntegrationConfig.cls
force2b: Ah. Nice implementation. That can work.

seanpat09:
Wait for #4160 to merge before merging this one. There will be some some conflicts between that branch and this one in `CRLP_ApiService` and it will be easier to manage the conflicts in this smaller branch

seanpat09:
@rponti-sforg This is ready for testing!

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -35,6 +35,10 @@
 */
 @IsTest(IsParallel=true)
 private class CRLP_ApiService_TEST {
+    private static UTIL_IntegrationConfig_TEST.MockPackage mockPackage;
+    private static MockCallable callableMock;
+
+    private static final String ROLLUP_DEFINITIONS = '[{"useFiscalYear":false,"timeBoundOperation":"All_Time","theSum":0,"theCount":0,"textValue":[null,null,null,null],"summaryObject":"General_Accounting_Unit__c","sumByYear":{}}]';
```
force2b: This is ok as long as this is not going to be re-cast as `List<CRLP_Rollup>`.

seanpat09: I'll double check - if it is, I'll make a similar fix to as what I did in the child branch

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -200,13 +198,14 @@ private class CRLP_ApiService_TEST {
      */
     @isTest
     private static void getBaseRollupStateForRecordsShouldBuildCRLPRollupMap() {
+        final String MOCK_ROLLUP_DEFINITIONS = '[{"mockRollupDefinition":"randomData"}]';
```
seanpat09: @force2b turns out as long as this is valid JSON, it will still cast without breaking. I updated the variable to make it more clear that it is just mock data and not dependent on an actual serialized `List<CRLP_Rollup>`

seanpat09:
@force2b updated the PR comments, just waiting for the builds to pass. Once that's done, should we merge into `master`? If so, I can take care of that once the builds complete

PR #4206, opened 2019-04-26
Update Trialforce Template Id for NPSP 3.153

davidmreed:
This PR updates the `trial.json` org definition for NPSP 3.153.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davisagli:
@davidmreed we need to delete the branch from github just after merging so that CCI won't keep merging changes from master to the branch and triggering new builds. I'll delete it now.

PR #4207, opened 2019-04-26
Add new batch size for Account Soft Credit Rollups

### LINKED ISSUE #4166: Customizable Rollups: Account Soft Credit Jobs Using Default 200 Batch Size
If a customer has many rollups/Filter Groups and many Accounts with thousands of Account Soft Credits, they may hit heap size errors when the CRLP_Account_AccSoftCredit_BATCH and CRLP_AccountSkew_AccSoftCredit_BATCH run. Typically we would suggest reducing the batch size, but we don't have one for this job. The suggestion is to either create its own or share the Rollups_Account_Soft_Credit_Batch_Size__c field (for non-skew at least).

Either solution should be set in: [CRLP_RollupBatch_SVC.cls, Lines 265-296](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/CRLP_RollupBatch_SVC.cls#L265-L296)

**Sample Stack Trace:**
```
Class.npsp.CMT_FilterRule.isIncluded: line 149, column 1
Class.npsp.CMT_FilterRuleEvaluation_SVC.evaluateFilterGroup: line 115, column 1
Class.npsp.CRLP_RollupAccSoftCredit_SVC.processRollups: line 80, column 1
Class.npsp.CRLP_RollupProcessor_SVC.completeRollupForParent: line 303, column 1
lass.npsp.CRLP_RollupProcessor_SVC.startRollupProcess: line 115, column 1
Class.npsp.CRLP_Batch_Base_NonSkew.execute: line 271, column 1

```

See Work Item for full Investigation.

patrick-yan-sf:
Modified the custom settings for customizable rollup batch sizes.  

# Critical Changes

# Changes
- We introduced a new batch size setting for Customizable Rollups called Contact Soft Credit to Acct Batch Size that will be used for Account Contact Soft Credit Batch Sizes. The existing Account Soft Credit Batch Size will now be used for Account Soft Credits (Non-Skew mode).  The current value from Account Soft Credit Batch Size will be copied to the new Contact Soft Credit to Acct Batch Size so that your existing settings are not modified. For Skew Mode, the Account Skew Mode Batch Size will also be used for Account Soft Credit Skew mode execution.

# Issues Closed
- Fixes #4166

# New Metadata
CustomSetting
- Customizable_Rollup_Settings__c.Rollups_AcctContactSoftCredit_Batch_Size__c

# Deleted Metadata

### DIFF
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -317,7 +317,7 @@ global without sharing class STG_InstallScript implements InstallHandler {
     * added
     */
     private void createNewDefaultCustomSettingRecords() {
-        UTIL_CustomSettingsFacade.getOrgCustomizableRollupSettings();
+        updateDefaultRollupSettings();
         updateDefaultOrgContactsSettings();
     }
 
@@ -331,6 +331,30 @@ global without sharing class STG_InstallScript implements InstallHandler {
             npe01Settings.Contact_Role_for_Organizational_Opps__c = UTIL_CustomSettingsFacade.DEFAULT_OPPORTUNITY_CONTACT_ROLE_SOFT_CREDIT;
             update npe01Settings;
         }
+    }
+
+    /*******************************************************************************************************
+    * @description Ensure that any default Org Rollup Settigns are updated for custom settings 
+    * objects accordingly.
+    */
+    public static void updateDefaultRollupSettings() {
+        Customizable_Rollup_Settings__c rollupSettings = UTIL_CustomSettingsFacade.getOrgCustomizableRollupSettings();
+        Boolean isUpdateNeeded = false;
+
+        if (rollupSettings.Rollups_AccSoftCredit_Skew_Batch_Size__c == null) {
+            rollupSettings.Rollups_AccSoftCredit_Skew_Batch_Size__c = UTIL_CustomSettingsFacade.DEFAULT_ROLLUP_SKEW_MODE_BATCH_SIZE;
+            isUpdateNeeded = true;
+        }
+
+        if (rollupSettings.Rollups_AccSoftCredit_Batch_Size__c == null) {
+            rollupSettings.Rollups_AccSoftCredit_Batch_Size__c = UTIL_CustomSettingsFacade.DEFAULT_ROLLUP_BATCH_SIZE;
+            isUpdateNeeded = true;
+        }
+        
+        if (isUpdateNeeded) {
+            update rollupSettings;
+        }
+        
     }
 
     /*******************************************************************************************************
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -335,19 +335,20 @@ global without sharing class STG_InstallScript implements InstallHandler {
 
     /*******************************************************************************************************
     * @description Ensure that any default Org Rollup Settigns are updated for custom settings 
-    * objects accordingly.
+    * objects accordingly.  There was a swap of custom setting fields to get the account soft credit settings
+    * pointed to the corresponding setting
     */
     public static void updateDefaultRollupSettings() {
         Customizable_Rollup_Settings__c rollupSettings = UTIL_CustomSettingsFacade.getOrgCustomizableRollupSettings();
         Boolean isUpdateNeeded = false;
+        Decimal currentAccountSoftCreditBatchSize = UTIL_CustomSettingsFacade.DEFAULT_ROLLUP_BATCH_SIZE;
 
-        if (rollupSettings.Rollups_AccSoftCredit_Skew_Batch_Size__c == null) {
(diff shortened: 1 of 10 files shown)


src/objects/Customizable_Rollup_Settings__c.object
```diff
@@ -107,6 +107,20 @@
...
+        <trackTrending>false</trackTrending>
+        <type>Number</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
+        <fullName>Rollups_AcctContactSoftCredit_Batch_Size__c</fullName>
+        <defaultValue>200</defaultValue>
+        <description>Only used by the new Customizable Rollups engine</description>
+        <externalId>false</externalId>
+        <inlineHelpText>Number of records per batch to process when recalculating soft credit donor rollups to the Account object.</inlineHelpText>
```
patrick-yan-sf: @sam-knox Can you review the text for descriptions and such.  I just copied the text from the previous custom setting over.

src/pages/STG_PanelSchedule.page
```diff
@@ -49,7 +49,19 @@
...
+                        <apex:outputLabel value="{!$ObjectType.Customizable_Rollup_Settings__c.Fields.Rollups_AcctContactSoftCredit_Batch_Size__c.Label}" for="tbxAccContactSoft" styleClass="slds-form-element__label" />
+                        <div class="slds-form-element__control">
+                            <apex:outputField value="{!stgService.stgCRLP.Rollups_AcctContactSoftCredit_Batch_Size__c}" rendered="{!isReadOnlyMode}" styleClass="slds-form-element__static" />
+                            <apex:inputField value="{!stgService.stgCRLP.Rollups_AcctContactSoftCredit_Batch_Size__c}" rendered="{!isEditMode}" id="tbxAccContactSoft" styleClass="slds-input" />
+                        </div>
+                    </div>
+                </div>
+            </div>
+
+            <div class="slds-grid slds-gutters_small">
```
patrick-yan-sf: @sam-knox Let me know if I positioned it in the correct spot.  We can move it up or down

patrick-yan-sf:
**lurch: attach W-032439

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmKlQAK" target="_blank">W-032439</a>

src/classes/STG_InstallScript.cls
```diff
@@ -333,6 +333,31 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+            rollupSettings.Rollups_AcctContactSoftCredit_Batch_Size__c = currentAccountSoftCreditBatchSize;
+            isUpdateNeeded = true;
+        }
+        
+        if (isUpdateNeeded) {
+            update rollupSettings;
+        }
+        
+    }
+
```
patrick-yan-sf: This code will copy the retain the existing Account Contact Soft Credit Batch Size but it will also use that value for the new Account Soft Credit.  I think that is ok.  The alternative could possibly cause future upgrades to overwrite values if the customer accidentally blanks out one of them.

patrick-yan-sf:
@force2b Can you review? Thanks

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -425,6 +431,32 @@ public with sharing class STG_InstallScript_TEST {
         }
     }
 
+    /*********************************************************************************************************
+    @description 
+        Verify that the Customizable Rollup Settings matches the expected Configuration
+    **********************************************************************************************************/ 
+    private static void assertRollupSettingsEqual(Customizable_Rollup_Settings__c expectedConfig) {
```
force2b: Is it worth splitting this out since it's only referenced one time?

patrick-yan-sf: I think if we were to put into the actual push() test function, the basic structure of that test method would be different from the other ones.  I've added the new constants for the other values though.

force2b: Fair enough. It also makes the test method more readable. I would suggest updating the class to use `@IsTest` and new method names to bring in line with our current standards.

patrick-yan-sf: Thanks, I've updated the test class.  Can you also review the critical changes section?  Radmila brought up a good point - because we are switching settings - do we need to make our customers aware of this change?

patrick-yan-sf:
@sam-knox Can you also review the changes/critical changes section.  Just want to make sure our customers know that we are switching the custom setting - incase they were directly referencing it.

force2b:
@patrick-yan-sf and @sam-knox The one change I'd make to the Critical Changes and Changes sections is to replace the field api names with field labels since Customer's won't see the api name in NPSP Settings - only the field label.

patrick-yan-sf:
@sam-knox Thanks.

@rponti-sforg  - this should be ready to for testing.  Thanks

PR #4212, opened 2019-04-30
Replacement of NAMESPACE placeholder to use npsp prefix

Clint-Chester:
# Critical Changes
N/A
# Changes
Update of the Global Layout metadata and the Account field Number of Household Members to reference the global actions in the Nonprofit Success Pack (npsp) package instead of the NAMESPACE placeholder.

I'm not sure if all the metadata in the post install step is still relevant, but since these files haven't been touched since 2014 wanted to help out 😄 
# Issues Closed
N/A
# New Metadata
N/A
# Deleted Metadata
N/A

### DIFF
--- a/unpackaged/post/first/layouts/Global-Global Layout.layout
+++ b/unpackaged/post/first/layouts/Global-Global Layout.layout
@@ -5,7 +5,7 @@
             <quickActionName>FeedItem.TextPost</quickActionName>
         </quickActionListItems>
         <quickActionListItems>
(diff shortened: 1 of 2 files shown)


davisagli:
@Clint-Chester the use of the namespace token is intentional so that this metadata can be deployed with both managed and unmanaged variations of the NPSP package. The supported way of installing NPSP is using CumulusCI, which takes care of replacing the token during deployment.

Clint-Chester:
Apologies my mistake, makes total sense.

PR #4219, opened 2019-05-01
Add unpackaged metadata to support regression testing of BDI

davidmreed:
This branch introduces new unpackaged metadata, `config/bdi`, to support regression testing, and adds the new flow `config_bdi_qa`, to be run following `qa_org`.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davisagli:
@davidmreed hmm. I was thinking of this as additional metadata that would be added to the existing unpackaged/config/qa directory to be included for any org set up using the qa_org flow, rather than a separate set of metadata. The goal is for qa_org and install_regression to produce orgs that are ready for running the automated robot tests without the need to run additional flows manually. Happy to discuss though.

davidmreed:
That may be an "optimizing for the wrong quality" error on my part. I was assuming that the Page Layouts were needed to allow the tester to view the custom fields on a record, and didn't want to blanket-overwrite the packaged layouts in all QA orgs.

If the custom fields are to be used only in Robot tests of the BDI interface itself, then the layouts probably aren't needed and the other metadata just folded into the `qa` suite.

@skristem and @deepmalamehta, is that the better way to go from your perspective?

davisagli:
I suspect the fields do need to be in the layouts so that the robot can see them to verify that the data was mapped as expected during the import. But I think it's okay to just add them to the layouts for all QA orgs. config_managed (used by install_prod and install_beta) deploys the trial metadata but not the qa metadata, so there's still a way to get an org with everything configured except for the extra QA-related stuff. We probably want a new flow to do something similar (trial metadata but not qa metadata) but with unmanaged NPSP -- only reason I didn't do that already is I wasn't sure what to name it.

davidmreed:
Per conversation with @skristem and @deepmalamehta, the Page Layouts are not needed for the Robot tests they plan to run. I will remove them and fold the Custom Field metadata into the standard `qa` bundle.

PR #4221, opened 2019-05-01
Feature/bdi co  migration utility

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


ptewson-sfdo:
@psadie - I had to re-create this PR, long story!

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /**
+     * @description Queries for default Field Mappings and builds a map of DeveloperName to Field Mapping record
+     * Used to check for existance of Field Mapping record without loop
+     * @param fieldMappingSetId Id of the default Field Mapping Set
+     * @return Collection of Field Mapping DeveloperName to Field Mapping record
+     */
+    private Map<String, Data_Import_Field_Mapping__mdt> getDefaultFieldMappings(Id fieldMappingSetId) {
+        List<Data_Import_Field_Mapping__mdt> existingFieldMappings = [SELECT
```
ptewson-sfdo: @psadie Do we want to do any error handling here if there for some reason isn't a default field mapping set up? Not sure if that really can happen ...

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /**
+     * @description Queries for default Object Mappings and builds a map of DeveloperName to Object Mapping record
+     * Used to check for existance of Object Mapping record without loop
+     * @param objectMappingSetId Id of the default Object Mapping Set
+     * @return Collection of Object Mapping DeveloperName to Object Mapping record
+     */
+    private Map<String, Data_Import_Object_Mapping__mdt> getDefaultObjectMappings(Id objectMappingSetId) {
+        List<Data_Import_Object_Mapping__mdt> existingObjectMappings = [SELECT
```
ptewson-sfdo: @psadie similar here, do we want to do any handling of the case where the object mapping may not exist?

src/classes/BDI_MigrationMappingUtility.cls
psadie: @ptewson-sfdo - I'll have a commit up shortly this morning for this. I'd rather have it covered even if we are including these defaults in our package. Maybe someone deletes a set and subsequently gets an unfriendly error. Was focused on resolving a couple of issues Deepa found yesterday :)

PR #4223, opened 2019-05-02
Use Ensure Record Types task against Opportunity in dependencies flow

davidmreed:
# Critical Changes

# Changes

 - Ensure that NPSP's default record type can deploy successfully on customized orgs.

# Issues Closed

# New Metadata

# Deleted Metadata


davidmreed:
This PR is dependent upon CumulusCI [PR 1064](https://github.com/SFDO-Tooling/CumulusCI/pull/1064).

cumulusci.yml
```diff
@@ -73,6 +73,13 @@ tasks:
         class_path: cumulusci.tasks.salesforce.Deploy
         options:
             path: unpackaged/config/delete
+    
+    ensure_record_type:
```
davisagli: The name and description here will be the default text shown for this step in the installer, so let's make sure it's user-friendly. Something like:

    name: Ensure Opportunity Record Types
    description: This will make sure record types are enabled for the Opportunity object in your org before installing NPSP. If there are no Opportunity record types yet, it will create one called NPSP Default.

(Check with docs/CCE for preferred language.)

And may as well set `sobject: Opportunity` here too, since that's the one we care about for NPSP.

cumulusci.yml
```diff
@@ -344,6 +351,13 @@ flows:
                 options:
                     test_name_match: LDV_%
 
+    dependencies:
+        steps:
+            3:
+                task: ensure_record_type
+                options:
+                    sobject: Opportunity
+
```
davisagli: There is also a `beta_dependencies` flow that we should add the same step to.

davidmreed:
@jesslopez-sf, requesting your review on the language for describing this install step:

        name: Ensure Opportunity Record Types
        description: This will ensure Record Types are enabled for the Opportunity object in your org before installing NPSP. If there are no Opportunity record types yet, it will create one called NPSP Default.

davisagli:
@davidmreed want to trigger a rebuild of this now?

btw we should add `minimum_cumulusci_version: 2.4.4` at the top of cumulusci.yml so that it will error out early if run with an older version.

force2b:
@davidmreed  Since this change only impacts unpackaged deployment, should there a be a note in the release notes for this change given that it is not visible to customers?

davidmreed:
@force2b You're right - it shouldn't. Thanks for catching that; I will revise.

PR #4229, opened 2019-05-08
Mark customized rollups as changed 

patrick-yan-sf:
If NPSP has the bridge package installed and Customizable rollups are enabled then changes to the Customizable rollups will mark the rollup type (Account Hard Credit / Contact soft Credit / etc) as stale.

# Critical Changes

# Changes

# Issues Closed

# New Metadata
ApexClass
- CRLP_RollupCMT_TEST

# Deleted Metadata


patrick-yan-sf:
**lurch: attach W-032622

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmaAQAS" target="_blank">W-032622</a>

src/classes/CMT_MetadataAPI.cls
```diff
@@ -93,39 +93,60 @@ public class CMT_MetadataAPI {
...
+
+    /******************************************************************************************************************
+     * @description Interface for Metadata Callback Handler
+     * Is used to allow metadata deployments to have customized on result handlers
+     *****************************************************************************************************************/
+    public interface MetadataCallbackHandler {
+        void performSuccessHandler(Map<String, Object> params);
+    }
+
+    /******************************************************************************************************************
```
patrick-yan-sf: In order to execute the invalidation cache after a successful CMT deployment - I created an interface and added it to the Custom Callback class we already created.  Right now it only handles on success but we can extend it to provide on error handler as well.

src/classes/CMT_MetadataAPI.cls
```diff
@@ -159,11 +180,17 @@ public class CMT_MetadataAPI {
             crlpSettings.CMT_API_Status__c = JSON.serializePretty(statusMap).left(255);
             // THIS WILL NEED TO BE REFACTORED IF THIS CMT DEPLOYMENT CODE IS USED BY SOMETHING OTHER THAN ROLLUPS
             crlpSettings.Customizable_Rollups_Enabled__c = (isSuccess || crlpSettings.Customizable_Rollups_Enabled__c == true);
-            if (!Test.isRunningTest()) {
-                upsert crlpSettings;
-            }
+
```
patrick-yan-sf: I didn't touch this code above but as it says if we ever use the deployment for stuff other than CMT we'd probably want to refactor it

src/classes/CRLP_RollupCMT.cls
```diff
@@ -408,6 +462,127 @@ public class CRLP_RollupCMT {
...
+            if (otherRollup.isDeleted != this.isDeleted) {
+                return true;
+            }
+            if (otherRollup.useFiscalYear != this.useFiscalYear) {
+                return true;
+            }
+            if (otherRollup.intValue != this.intValue) {
+                return true;
+            }
+
```
patrick-yan-sf: Is there a better way to compare?

src/classes/CRLP_RollupCMT.cls
```diff
@@ -408,6 +462,127 @@ public class CRLP_RollupCMT {
...
+                if (this.summaryObject == accountApiName) {
+                    return CRLP_RollupProcessingOptions.RollupType.AccountHardCredit;
+                }
+                if (this.summaryObject == contactApiName) {
+                    return CRLP_RollupProcessingOptions.RollupType.ContactHardCredit;
+                }
+            }
+
+            return null;
+        }
```
patrick-yan-sf: I was looking for something that provided me this information but it doesn't look like there was one.  I took the logic from a couple of spots - this needs to be reviewed to make sure it is correct.

src/classes/CRLP_RollupCMT_TEST.cls
```diff
@@ -0,0 +1,229 @@
...
+     * @description verifies that different rollups are marked as different
+     *****************************************************************************************************************/
+    @isTest
+    private static void differentRollupsAreMarkedAsDifferent() {
+
+        CRLP_RollupCMT.Rollup newRollup = generateRollup('1');
+        CRLP_RollupCMT.Rollup existingRollup = generateRollup('2');
+        newRollup.operation = 'Count';
+
+        System.assertEquals(true, newRollup.isDifferent(existingRollup), 'Different Rollups are marked as different');
```
seanpat09: For the sake of code coverage of all the if statements, is it worth going through each property (i.e. `filterGroupRecordName`, `operation`, `timeBoundOperationType`, etc.) to make sure a difference in each one would return true?

src/classes/CRLP_RollupCMT.cls
```diff
@@ -122,6 +134,36 @@ public class CRLP_RollupCMT {
...
+            }
+
+            Map<String, FilterRule> rulesByHash = new Map<String, FilterRule>();
+
+            for (Integer otherIndex=0; otherIndex < otherFilterGroup.rules.size(); otherIndex++) {
+                FilterRule otherRule = otherFilterGroup.rules[otherIndex];
+                rulesByHash.put(otherRule.generateHash(), otherFilterGroup.rules[otherIndex]);
+            }
+
+            for (Integer currentIndex=0; currentIndex < this.rules.size(); currentIndex++) {
```
seanpat09: Spacing here:
```suggestion
            for (Integer currentIndex = 0; currentIndex < this.rules.size(); currentIndex++) {
```

src/classes/CRLP_RollupCMT.cls
```diff
@@ -122,6 +134,36 @@ public class CRLP_RollupCMT {
...
         }
+
+    /******************************************************************************************************************
+     * @description Determines if a Filter Group is computationally different from another Filter Group
+     * changes in name or label or description are not considered different
+     * filter rules are hashed because the sort order of the list is not guaranteed
+     * @param otherFilterGroup FilterGroup to compare against
+     * @return boolean to indicate of the filter group is different 
+     *****************************************************************************************************************/
+        public Boolean isDifferent(FilterGroup otherFilterGroup) {
```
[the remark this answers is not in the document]
patrick-yan-sf: Yeah I thought about using that - my hesitation is that - it is not really an equal comparison.  I'm not checking if the labels/descriptions are the same - so the two objects could have different values yet not be different - computationally.  But I'm open to making that change since It seems like that is the intent.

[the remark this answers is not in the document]
seanpat09: Oh that's a good point, let's leave this as is then. Implementing equals and hashCode may be more trouble than it's worth

src/classes/CRLP_RollupCMT.cls
```diff
@@ -408,6 +462,127 @@ public class CRLP_RollupCMT {
...
+                if (this.summaryObject == accountApiName) {
+                    return CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit;
+                }
+                if (this.summaryObject == contactApiName) {
+                    return CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit;
+                }
+            }
+
+            if (this.amountObject == accountSoftCreditApiName) {
+                if (this.detailObject == opportunityApiName || this.detailObject == accountSoftCreditApiName) {
```
force2b: Technically if the Amount Object is AccountSoftCredit then it's always an AccountSoftCredit rollup, no matter what the Detail object is. However, if there were a Count type rollup the Amount object might not be set (because it's ignored). I suggest using `detailObject == accountSoftCreditApiName || amountObject == accountSoftCreditApiName`.

src/classes/CRLP_RollupCMT.cls
```diff
@@ -408,6 +462,127 @@ public class CRLP_RollupCMT {
...
+            // work down from most specific
+            if (this.summaryObject == recurringDonationsApiName) {
+                return CRLP_RollupProcessingOptions.RollupType.RecurringDonations;
+            }
+
+            if (this.summaryObject == allocationApiName) {
+                return CRLP_RollupProcessingOptions.RollupType.GAU;
+            }
+
+            if (this.amountObject == partialSoftCreditApiName) {
```
force2b: I suggest using `detailObject == partialSoftCreditApiName || amountObject == partialSoftCreditApiName`

src/classes/CRLP_RollupCMT.cls
force2b: Other than my suggestions above, this is good. There was no single method that determined this previously.

src/classes/CRLP_RollupCMT_TEST.cls
```diff
@@ -0,0 +1,273 @@
...
+        System.assertEquals(true, newRollup7.isDifferent(existingRollup), 'Different Rollups are marked as different for detailField');
+        System.assertEquals(true, newRollup8.isDifferent(existingRollup), 'Different Rollups are marked as different for amountObject');
+        System.assertEquals(true, newRollup9.isDifferent(existingRollup), 'Different Rollups are marked as different for amountField');
+        System.assertEquals(true, newRollup10.isDifferent(existingRollup), 'Different Rollups are marked as different for dateObject');
+        System.assertEquals(true, newRollup11.isDifferent(existingRollup), 'Different Rollups are marked as different for dateField');
+        System.assertEquals(true, newRollup12.isDifferent(existingRollup), 'Different Rollups are marked as different for isActive');
+        System.assertEquals(true, newRollup13.isDifferent(existingRollup), 'Different Rollups are marked as different for isDeleted');
+        System.assertEquals(true, newRollup14.isDifferent(existingRollup), 'Different Rollups are marked as different for intValue');
+
```
force2b: extra blank lines

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -715,13 +720,33 @@ public with sharing class CRLP_RollupUI_SVC {
                         && !Test.isRunningTest()) {
                     throw new RollupException(Label.CRLP_DuplicateTargetField);
                 }
+
+                if (isRollupStateEnabled && rlp.DeveloperName == rollupCMTObj.recordName) {
+                    isRollupStale = rollupCMTObj.isDifferent(rlp);
+                    if (isRollupStale) {
+                        previousRollupType = new CRLP_RollupCMT.Rollup(rlp).getRollupType();
+                    }
+                    break;
```
force2b: Why `break` here? Wouldn't that exit out of the loop before it finishes checking for duplicates?

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -706,6 +706,11 @@ public with sharing class CRLP_RollupUI_SVC {
             // CMT data will generate a unique RecordName to use for the new record.
             CRLP_RollupCMT.Rollup rollupCMTObj = (CRLP_RollupCMT.Rollup) JSON.deserialize(rollupCMT, CRLP_RollupCMT.Rollup.class);
 
+            Boolean isRollupStateEnabled = new CRLP_ApiService().isRollupStateEnabled();
+            Boolean isRollupStale = false;
```
force2b: `Stale` is a concept that is sort of unique to the DSO-NPSP Bridge. From the perspective of NPSP, maybe this could be described as `rollupChanged` which triggers an operation of something like `rollupChangedEvent` that eventually ends up in the api which knows what to do with the `rollupChanged` or `filterRuleChanged` event.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -706,6 +706,11 @@ public with sharing class CRLP_RollupUI_SVC {
             // CMT data will generate a unique RecordName to use for the new record.
             CRLP_RollupCMT.Rollup rollupCMTObj = (CRLP_RollupCMT.Rollup) JSON.deserialize(rollupCMT, CRLP_RollupCMT.Rollup.class);
 
+            Boolean isRollupStateEnabled = new CRLP_ApiService().isRollupStateEnabled();
```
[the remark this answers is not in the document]
force2b: The abstraction of the logic is helpful, though I would (referring to comment below) say that the `shouldMarkRollupStale()` method might either not exist in this context or be renamed `shouldTriggerRollupChanged()`. I think it would be best if the save methods simply call the `rollupChangedTrigger` (or `filterRuleChangedTrigger`) and leave the logic that checks to see if a _trigger` (i.e., the bridge) is implemented in those methods..

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -970,6 +1010,95 @@ public with sharing class CRLP_RollupUI_SVC {
...
+        Map<Id, List<Filter_Rule__mdt>> rulesByGroup = CMT_FilterRule_SEL.getFilterRulesByGroup();
+        Map<Id, List<Rollup__mdt>> rollupsByFilterGroup = CRLP_Rollup_SEL.getRollupsByFilterGroup();
+
+        Set<String> modifiedRollupNames = new Set<String>();
+
+        for (Filter_Group__mdt existingFilterGroupMdt : allFilterGroupsMdt) {
+            CRLP_RollupCMT.FilterGroup existingFilterGroup = new CRLP_RollupCMT.FilterGroup(existingFilterGroupMdt);
+            
+            if (newFilterGroup.recordId == existingFilterGroup.recordId) {
+                Boolean isFilterGroupStale = newFilterGroup.isDifferent(
```
force2b: Same note as above - `Stale` is a DSO-NPSP bridge concept. From NPSP, it's more about identifying when the Filter Group/Rules has changed and taking some action.

unpackaged/config/trial/package.xml
```diff
@@ -131,6 +131,7 @@
         <members>LeadSource</members>
         <members>OpportunityStage</members>
         <members>OpportunityType</members>
+        <members>Salutation</members>
```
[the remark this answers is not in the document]
force2b: Interesting. I wonder if this branch accidentally got some of the changes that David is working on for the trial builds?

src/classes/CRLP_RollupUI_SVC.cls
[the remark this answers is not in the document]
RadSFDeveloper: I agree "stale" concept is not known within NPSP. In order to avoid ambiguity and since the Bridge is not using triggers to process stale records, perhaps we can keep the name generic and implement details in the method calling the integration API? I have updated the suggested code for the method name: `shouldSendChangeEvent()`. 
I agree there might be better to check `isRollupStateEnabled` in the lower levels closer to calling DSO API, if MetadataAPI can be changed. However, we do not want to differentiate the rollup with all existing rollups if the `isRollupStateEnabled` is false, so there might a need to keep the flag in this class. We still need to traverse existing rollups to determine if the target field is duplicate, but nothing more if the `isRollupStateEnabled` is false.

src/classes/CRLP_RollupUI_SVC.cls
RadSFDeveloper: Good catch. The `break` should not be there. The suggested refactoring does not contain it.

src/classes/CMT_MetadataAPI.cls
```diff
@@ -93,39 +93,59 @@ public class CMT_MetadataAPI {
...
-    /**
+    /******************************************************************************************************************
      * @description Initiate a deployment of custom metadata type records.
-     * @param jobId Unique string to act as a "jobId" when writing the final deployment status to the Custom Settings
      * @param mtToDeploy List of CustomMetadata objects to deploy
-     */
-    public static void deployMetadata(String jobId, List<Metadata.CustomMetadata> mtToDeploy) {
+     * @param callback a MetadataCallBack to handle the deployment result
+     *****************************************************************************************************************/
+    public static void deployMetadata(List<Metadata.CustomMetadata> mtToDeploy, MetadataCallBack callback) {
```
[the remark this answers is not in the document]
patrick-yan-sf: @force2b - no I was not aware they were making changes - I'll check with @psadie

unpackaged/config/trial/package.xml
```diff
@@ -131,7 +131,6 @@
         <members>LeadSource</members>
         <members>OpportunityStage</members>
         <members>OpportunityType</members>
-        <members>Salutation</members>
```
force2b: Something is still not right here. The package.xml shouldn't be in the PR at all as changed in any way.

orgs/trial.json
```diff
@@ -1,6 +1,6 @@
 {
   "orgName": "NPSP - Trial Org",
-  "template": "0TT1I000006uGCr",
```
force2b: Something is still not right here. The `trial.json` shouldn't be in the PR at all as changed in any way.

force2b: Sorry. You can ignore these last two comments.

unpackaged/config/trial/standardValueSets/Salutation.standardValueSet
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: Ok, the last two comments were wrong, but this file definitely shouldn't be in the PR.

patrick-yan-sf: @force2b - yeah I think what happened was that this branch originally was merging into master and thus took in some changes (david's changes on the trial org) before switching to merge to sean's main branch

src/classes/CRLP_ApiService.cls
```diff
@@ -37,10 +37,12 @@
 public inherited sharing class CRLP_ApiService {
 
     public static final String PARAM_ROLLUPTYPE = 'RollupType';
+    public static final String PARAM_ROLLUP_TYPES = 'RollupTypes';
```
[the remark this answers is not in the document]
patrick-yan-sf: I think part of the problem is with the current name is the inconsistent underscoring.  It looks like PARAM_ROLLUPTYPE and PARAM_ROLLUP_TYPES 

For now I've made it 
```
    public static final String PARAM_ROLLUP_TYPE = 'RollupType';
    public static final String PARAM_ROLLUP_TYPES = 'RollupTypes';
```
So hopefully it is easier to read

patrick-yan-sf:
I believe I've gotten to all the comments.  For the null input to JSON error that Radmila saw. I was able to reproduce it - but not consistently.  I believe it is related to the javascript using the same name for the parameter as a javascript object.  When I change the name to something else it works.  Also when I turn on lightning debugging then it works.
By giving the parameter a different name it consistently works.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -693,76 +693,92 @@ public with sharing class CRLP_RollupUI_SVC {
...
             }
 
             // Queue up the deployment
-            CRLP_ConfigBuilder_SVC.queueRollupConfigForDeploy(new List<CRLP_RollupCMT.FilterGroup>{filterGroupCMTObj});
+            CRLP_ConfigBuilder_SVC.queueRollupConfigForDeploy(new List<CRLP_RollupCMT.FilterGroup>{cmtFilterGroup});
+
+            FilterGroupUtil util = new FilterGroupUtil();
+            
+            String jobId;
+            if (util.isRollupStateEnabled()) {
```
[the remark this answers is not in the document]
RadSFDeveloper: When a filter group is cloned and saved, the change event is not sent. The same is applicable to the filter group deletion. Can you please check if you can reproduce behaviour?

src/classes/CRLP_RollupCMT.cls
force2b: @patrick-yan-sf The second `if` is not needed here. If the amount or detail object is AccountSoftCredit, then it's always AccountSoftCredit so there's no need to check the detail object a second time.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -970,6 +986,229 @@ public with sharing class CRLP_RollupUI_SVC {
...
+                }
+                return isRollupStateEnabled;
+            }
+            set;
+        }
+
+        /******************************************************************************************************************
+        * @description Determines if the archive bridge rollup state is enabled.
+        */
+        public Boolean isRollupStateEnabled() {
```
force2b: "Rollup State" is a concept that, like "Stale", is unique to the DSO Bridge. From the viewpoint of NPSP, it needs to know if there is some extension installed that needs to know when a Rollup or FilterRule is changed. How about something like `isRollupChangeTriggerEnabled' or `isCrlpChangeTriggerEnabled`, at least within this class. Have to look at the usage of the method in the parent branch to see if a name change is warranted there.

unpackaged/config/trial/package.xml
force2b: I think there is still an issue with the trial/package.xml and standardValueSet. The PR shouldn't have any reference to these two files at all. If they're included here it's because the PR is trying to modify them which it shouldn't be doing at all. The files should be left as they are.

src/classes/CRLP_RollupUI_SVC.cls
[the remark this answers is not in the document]
patrick-yan-sf: Correct me if I'm wrong but when a filter group is cloned (a new one created) there are no referenced rollups yet - so no rollup has changed.  That's why we don't send any change event.  Also when a rollup is deleted (it can only be deleted when there are no referenced rollups) so there's no change event as well.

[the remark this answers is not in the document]
RadSFDeveloper: Good point! Thanks for the clarification. Let's ensure we have these scenarios in a test charter indicating no change event will be created.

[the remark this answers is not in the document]
force2b: Great questions. Absolutely right on both. The only changed event on a Filter that would need t be sent is when a Rule is added, deleted or modified on a Filter Group that is attached to one or more active Rollups. In that case, all the Rollup Definitions associated with that Filter Group would be considered changed.

src/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -693,76 +693,84 @@ public with sharing class CRLP_RollupUI_SVC {
...
+    public static String saveRollup(String jsonRollup) {
         try {
             // For new records, the RecordName property will be null by default. The process of deploying the
             // CMT data will generate a unique RecordName to use for the new record.
-            CRLP_RollupCMT.Rollup rollupCMTObj = (CRLP_RollupCMT.Rollup) JSON.deserialize(rollupCMT, CRLP_RollupCMT.Rollup.class);
-
-            // Check for duplicate Target fields on active Rollups
+            CRLP_RollupCMT.Rollup cmtRollup = (CRLP_RollupCMT.Rollup) JSON.deserialize(jsonRollup, CRLP_RollupCMT.Rollup.class);
+            
             for (Rollup__mdt rlp : CRLP_Rollup_SEL.cachedRollups) {
```
RadSFDeveloper: Similarly `isDuplicate()` can traverse through rollups and compare them to the `cmtRollup`:
```
if (cmtRollup.isDuplicate()) {
    throw new RollupException(Label.CRLP_DuplicateTargetField);
}
```

where `CRLP_RollupCMT` method:
```
* @description Determines if the target rollup is a duplicate
*/
public Boolean isDuplicate() {
    for (Rollup__mdt rlp : CRLP_Rollup_SEL.cachedRollups) {
        if (rlp.DeveloperName != this.recordName
            && rlp.Active__c
            && this.summaryObject == rlp.Summary_Object__r.QualifiedApiName
            && this.summaryField == rlp.Summary_Field__r.QualifiedApiName
            && !Test.isRunningTest()
        ) {
            return true;
        }
    }
    
    return false;
}
```

src/classes/CRLP_ApiService.cls
```diff
@@ -194,8 +218,257 @@ public without sharing class CRLP_ApiService {
...
+            return new Set<String> { rollupType.name(), oldRollupType.name() };
+        }
+    }
+
+    /******************************************************************************************************************
+    * @description Utility class for Filter Group processing
+    */
+    public class FilterGroupUtil {
+    
+        List<Filter_Group__mdt> allGroups = CMT_FilterRule_SEL.getFilterGroups();
```
RadSFDeveloper: Let's use lazy load for `allGroups` and `rollupsById`.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -228,6 +229,109 @@ private class CRLP_ApiService_TEST {
...
+        };
+
+        Test.startTest();
+        service.setRollupStateAsStale(expectedRollupTypes);
+        Test.stopTest();
+
+        System.assertNotEquals(
+            null, 
+            callableMock.actionCalled, 
+            'The callabled action should be called'
```
seanpat09: Small typo here
```suggestion
            'The callable action should be called'
```

src/classes/CRLP_ApiService.cls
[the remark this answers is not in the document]
force2b: @patrick-yan-sf I think these constants should be in the `CallableApiParameters` class.

src/classes/CRLP_ApiService_TEST.cls
```diff
@@ -228,6 +229,109 @@ private class CRLP_ApiService_TEST {
...
+     * bridge is not installed
+     */
+    @isTest
+    private static void markRollupStateAsStaleWillNotThrowExceptionWhenNoBridgeInstalled() {
+        UTIL_CustomSettingsFacade.getRollupSettingsForTests(new Customizable_Rollup_Settings__c (
+            Customizable_Rollups_Enabled__c = true
+        ));
+
+        CRLP_ApiService service = new CRLP_ApiService();
+        setupMockIntegration(service);
```
seanpat09: Does `mockPackage.isInstalled` need to be set to `false` for this test case?
```suggestion
        setupMockIntegration(service);
        mockPackage.isInstalled = false;
```

src/classes/CRLP_ApiService.cls
```diff
@@ -194,8 +218,274 @@ public without sharing class CRLP_ApiService {
...
+                }
+                return rollupType;
+            }
+            set;
+        }
+
+        private Boolean isNew {
+            get {
+                if (isNew == null) {
+                    isNew = String.isBlank(cmtRollup.recordName);
```
force2b: I'm pretty sure that `recordName` gets assigned immediately in the constructor for `CRLP_RollupCMT.Rollup`, so it should never be blank. What you could probably do is add a Boolean to the Rollup, FilterGroup, FilterRule inner classes for `isNew` that gets set to true in the constructor that accepts a String label value.

src/classes/CRLP_ApiService.cls
```diff
@@ -194,8 +218,274 @@ public without sharing class CRLP_ApiService {
...
+        }
+
+        /**
+        * @description Inserts an error log into the Error table
+        * @param ex Exception object that is to be logged
+        */
+        private void logError(Exception ex) {
+            Error__c rollupHandlerError = new Error__c(
+                Error_Type__c = 'Rollup Metadata Handler Error',
+                Object_Type__c = 'CustomMetadata',
```
force2b: Can you add `Context_Type__c = ERR_Handler_API.Context.CRLP.name()` to this?

RadSFDeveloper:
@force2b , @seanpat09 : Changes have been applied based on the review. Can you please verify?

RadSFDeveloper:
@rponti-sforg , PR for creating events on rollup and filter group modifications is ready to be tested.

src/classes/CRLP_ApiService.cls
RadSFDeveloper: Changes for the rollup `isNew` have been reverted back. Creating new or cloning existing rollup work only when used from `RollupUtil` as is. Please let me know if there are questions.

force2b: @RadSFDeveloper Not sure I follow this. In the `CRLP_RollupCMT.Rollup` inner class, the constructor - when creating a new rollup - is the following (line 283):

```
        public Rollup(String label) {
            //check to see if UDR; these need a unique name
            Boolean isUDR = (label.left(3) == 'UDR');
            this.label = label.replace('Number', 'Num').left(40).trim();
            this.recordName = generateRecordName(this.label, isUDR);
            this.useFiscalYear = false;
            this.isActive = true;
        }
```

Based on the above, I don't believe the `recordName` property would ever be empty. If true, then `isNew` could never return true.

src/classes/CMT_MetadataAPI.cls
```diff
@@ -153,16 +174,14 @@ public class CMT_MetadataAPI {
...
 
-            // Write the deployment status to the custom settings object
-            Customizable_Rollup_Settings__c crlpSettings = UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
-            crlpSettings.CMT_API_Status__c = JSON.serializePretty(statusMap).left(255);
-            // THIS WILL NEED TO BE REFACTORED IF THIS CMT DEPLOYMENT CODE IS USED BY SOMETHING OTHER THAN ROLLUPS
-            crlpSettings.Customizable_Rollups_Enabled__c = (isSuccess || crlpSettings.Customizable_Rollups_Enabled__c == true);
-            if (!Test.isRunningTest()) {
-                upsert crlpSettings;
-            }
             if (isSuccess) {
```
patrick-yan-sf: This is the big change here - removed this CRLP specific code - as other calls of the CMT don't need / want this code running when deploying other Custom Metadata Types.

src/classes/CRLP_ApiService.cls
```diff
@@ -194,8 +213,329 @@ public without sharing class CRLP_ApiService {
...
+
+    /*
+    * @description handle changes to customizable rollups and send back a callback handler if 
+    * there is processing that needs to be executed after deloyment
+    * @param rollups is a List of Rollups
+    * @return a MetadataHandler that can be used for deployment callbacks
+    */
+    public RollupMetadataHandler sendChangeEvent(List<CRLP_RollupCMT.Rollup> rollups) {
+
+        Set<String> changedRollupTypes = new Set<String>();
```
RadSFDeveloper: These two lines can be rewritten as:
```
Set<String> changedRollupTypes = buildChangedRollupTypes(rollups);
```

src/classes/CRLP_ApiService.cls
```diff
@@ -194,8 +213,329 @@ public without sharing class CRLP_ApiService {
...
+    }
+
+    /*
+    * @description handle changes to customizable filter groups and send back a callback handler if
+    * there is processing that needs to be executed after deployment
+    * @param filterGroups is a list of filtergroups
+    * @return a MetadataHandler that can be used for deployment callbacks
+    */
+    public RollupMetadataHandler sendChangeEvent(List<CRLP_RollupCMT.FilterGroup> filterGroups) {
+        Set<String> changedRollupTypes = new Set<String>();
```
RadSFDeveloper: These two lines can be rewritten as:
```
Set<String> changedRollupTypes = buildChangedRollupTypes(filterGroups);
```

patrick-yan-sf: Thanks - yes I have combined them.  Good catch.

patrick-yan-sf:
@rponti-sforg This is ready for Testing

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000008LmaAQAS/view

rponti-sforg:
@patrick-yan-sf  Let me know if you want to run the build again or if it's okay to merge with Feature Test - Robot Prerelease failed.

patrick-yan-sf:
@rponti-sforg - I've triggered the rebuild but I don't think it should be a problem.  Those failed test are not related to the work item but we can wait and see if the rebuild fixes it.

rponti-sforg:
> @rponti-sforg - I've triggered the rebuild but I don't think it should be a problem. Those failed test are not related to the work item but we can wait and see if the rebuild fixes it.

@patrick-yan-sf Looks like we have same failures.  Just let me know what you would like me to do.

PR #4238, opened 2019-05-14
Include Archive Rollup Summaries in CRLP calculations

seanpat09:
# Critical Changes

# Changes
Includes rollup state data fetched from Bridge in CRLP calculations
# Issues Closed

# New Metadata

# Deleted Metadata


seanpat09:
**lurch: attach W-032786

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uLaQAI" target="_blank">W-032786</a>

src/classes/CRLP_RollupProcessor_TEST.cls
```diff
@@ -302,6 +302,218 @@ private class CRLP_RollupProcessor_TEST {
...
+            'The number of closed opportunities for this Account should be 11, which includes the 10 from external data');
+        System.assertEquals(1100, updatedAcct.npo02__TotalOppAmount__c,
+            'The total of closed opportunities for this Account should be 1100, which includes the 1000 from external data');
+    }
+
+    /**
+     * @description Builds a mock List<CRLP_Rollup> to be used as for external rollup data
+     */
+    private static List<CRLP_Rollup> buildMockCRLPRollup() {
+        String mockRollupJSON =
```
seanpat09: Is there a better way to create this list?

force2b: Definitely, plus this is fragile (and is in fact no longer valid as per PR 4232). The best way might be to use `CMT_UnitTestData_TEST.mockRollupCmtForAccountHC()` to mock a set of `Rollup__mdt`records for Account Hard Credit. Then use the `CRLP_ApiService.getRollupDefinitions()` method to convert that into `List<CRLP_Rollup>`.

seanpat09: I updated the method to use the CRLP_Processor to create some mock Rollups, I think that should make this more robust

seanpat09:
@RadSFDeveloper @force2b I added a check in the Non Skew batch that check if a summary record has rollup state so that summary records with no details are not skipped over

seanpat09:
@rponti-sforg this is ready for QA

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -202,6 +236,15 @@ public inherited sharing class CRLP_RollupProcessor {
         return this;
     }
 
+    /**
+     * @description Specifies that external data should not be used at all with this processor. Necessary when
+     * external package is using this processor to calculate rollups in order to not recursively include external data
+     */
+    public CRLP_RollupProcessor withExcludeExternalData() {
```
force2b: Assuming that "External Data" here refers to retrieving the rollup state, what do you think about one of these alternate method names:
* `withoutRetrieveExternalRollupStateData`
* `withoutExternalRollupStateData` 
* `withExcludeExternalRollupStateData`

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -93,6 +93,19 @@ public inherited sharing class CRLP_RollupProcessor {
      */
     private Map<Id,Set<Id>> processedOppIdsByParentId = new Map<Id,Set<Id>>();
 
+    private Boolean shouldFetchExternalData = true;
```
force2b: How about `shouldRetrieveExternalRollupStateData`?

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -546,5 +590,21 @@ public inherited sharing class CRLP_RollupProcessor {
...
+        if (!shouldFetchExternalData) {
+            return;
+        }
+
+        Set<Id> recordIds = new Set<Id>();
+        for (SObject parentRecord : parentRecords) {
+            recordIds.add(parentRecord.Id);
+        }
+
+        rollupsDefsByParentId = apiService.getBaseRollupStateForRecords(recordIds, rollupType);
```
force2b: If the `withRollupDefinitions()` method was used to inject rollup definitions into this process (such as in Skew Mode), it might only do so for some of the ParentId's being rolled up. For example, Skew Mode might pass in 10 Account Summary Records, where the first has Rollup Definitions from a previous process (child batch iteration). In this example, `withRollupDefinitions` would get just the definitions for the one parent record. 

There would still be a need to retrieve the external rollup states for the other Accounts that did not have rollup defintion data injected. I think this affects this method as well as the two `withRollupDefinitions` methods.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -546,5 +590,21 @@ public inherited sharing class CRLP_RollupProcessor {
         return handler;
     }
 
+    /**
+     * @description Initializes rollupsDefsByParentId with external rollup data
+     */
+    private void initializeExternalData() {
```
force2b: What do you think about `initializeExternalRollupStateData`?

src/classes/CRLP_RollupProcessor.cls
seanpat09: I like `withoutRetrieveExternalRollupStateData`, it reads a little better. I'll make the change

src/classes/CRLP_RollupProcessor.cls
seanpat09: ahh good point, I'll make that change.

@rponti-sforg If you're currently testing this, there might be an issue with skew mode that I need to fix. I'll tag you again when it's back up

seanpat09:
@force2b Made the requested changes, I believe that should take care of the skew mode issue you pointed out, but let me know if I've missed something

src/classes/CRLP_RollupProcessor.cls
rponti-sforg: > ahh good point, I'll make that change.
> 
> @rponti-sforg If you're currently testing this, there might be an issue with skew mode that I need to fix. I'll tag you again when it's back up

@seanpat09 Do I or should I create a new scratch org to test with?

rponti-sforg:

For further testing details, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E0000087uLaQAI/view

rponti-sforg:
@seanpat09 @force2b I did not merge this because Mike looks to have 1 change request.  I am not sure if this is outdated or not.  Did not want to merge until I received confirmation from you.

I will not be able to get to this since I will be on vacation.  If someone could merge this for me, I would appreciate it.

force2b:
@rponti-sforg Thanks for testing and for the heads-up. I approved the changes and will merge this.

PR #4243, opened 2019-05-15
Draft: Automate bdi benchmark test

prescod:
# Changes

Adds tasks, flows and Robot keywords for BDI testing.

Guided tour:

1. First we need the data, so I converted David Glick's class into a Task:

https://github.com/SalesforceFoundation/Cumulus/pull/4243/files#diff-c6088a9879682c1af25379e1729b8258

2. The instructions for how to map to the Org are in mapping.yml:

https://github.com/SalesforceFoundation/Cumulus/pull/4243/files#diff-8266b04adf776303d15b501884c92485

3. Next we need to be able to run the actual test, so I created a Robot test:

https://github.com/SalesforceFoundation/Cumulus/blob/1ef8407fa38e83ffcae0b9093a6bd1621b048bed/robot/Cumulus/perf/BDITest.robot

4. The robot test needs a new keyword for doing BDI imports:

https://github.com/SalesforceFoundation/Cumulus/pull/4243/files#diff-85e4e256ee474963612d941be48b86e3R926

Sorry: my editor made a LOT of whitespace changes in that file (NPSP.py) and I didn't see what it had done until I got to github...

5. Finally I need to glue it together with flows and tasks:

https://github.com/SalesforceFoundation/Cumulus/pull/4243/files#diff-b0e88920e25a1ba7fb4868790bbe6be8

### DIFF
--- a/robot/Cumulus/perf/BDITest.robot
+++ b/robot/Cumulus/perf/BDITest.robot
@@ -1,0 +1,8 @@
+*** Settings ***
+
+Resource        robot/Cumulus/resources/NPSP.robot
+
+*** Test Cases ***
+
+Do a Batch Load via API
+    Api Batchdata Load
--- a/robot/Cumulus/resources/NPSP.py
+++ b/robot/Cumulus/resources/NPSP.py
@@ -16,6 +16,11 @@ from cumulusci.robotframework.utils import selenium_retry
 import sys
 from email.mime import text
 
+from cumulusci.robotframework.CumulusCI import CumulusCI
+from cumulusci.tasks.apex.anon import AnonymousApexTask
+from cumulusci.core.config import TaskConfig
+from cumulusci.tasks.apex.batch import BatchApexWait
+
 from locators_44 import npsp_lex_locators as locators_44
 from locators_45 import npsp_lex_locators as locators_45
 locators_by_api_version = {
@@ -27,7 +32,7 @@ npsp_lex_locators = {}
 
 @selenium_retry
 class NPSP(object):
-    
+
     ROBOT_LIBRARY_SCOPE = 'GLOBAL'
     ROBOT_LIBRARY_VERSION = 1.0
 
@@ -37,7 +42,7 @@ class NPSP(object):
         self._session_records = []
         self.val=0
         self.payment_list= []
-        # Turn off info logging of all http requests 
+        # Turn off info logging of all http requests
         logging.getLogger('requests.packages.urllib3.connectionpool').setLevel(logging.WARN)
         self._init_locators()
 
@@ -104,7 +109,7 @@ class NPSP(object):
         button = self.selenium.get_webelement(locator)
         button.click()
         time.sleep(5)
-        
+
     def select_tab(self, title):
         """ Switch between different tabs on a record page like Related, Details, News, Activity and Chatter
             Pass title of the tab
@@ -113,8 +118,8 @@ class NPSP(object):
         self.selenium.set_focus_to_element(locator)
         button = self.selenium.get_webelement(locator)
         button.click()
-        time.sleep(5)    
-        
+        time.sleep(5)
+
     def click_special_related_list_button(self, heading, button_title):
         """ To Click on a related list button which would open up a new lightning page rather than a modal.
             Pass the list name and button name"""
@@ -123,7 +128,7 @@ class NPSP(object):
             heading, button_title
         )
         self.selenium.click_link(locator)
-        
+
     def click_dropdown(self, title):
         locator = npsp_lex_locators['record']['list'].format(title)
         self.selenium.set_focus_to_element(locator)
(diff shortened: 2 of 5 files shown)


davidmreed:
@prescod Can you resolve the merge conflicts?

prescod:
> @prescod Can you resolve the merge conflicts?

Actually, some of them seem to have nothing to do with my code, so I'm not sure how to resolve them. I was picking up from a branch that Glick did and those conflicts must bet between that branch and Master. 

After I get some feedback on this PR I'll recreate it from scratch off of Master.

robot/Cumulus/perf/BDITest.robot
```diff
@@ -0,0 +1,8 @@
+*** Settings ***
+
+Resource        robot/Cumulus/resources/NPSP.robot
+
+*** Test Cases ***
+
+Do a Batch Load via API
+    Api Batchdata Load
```
davisagli: It looks to me like you can use the existing `Run Task Class` keyword rather than adding a new keyword:

    Run Task Class | cumulusci.tasks.apex.anon.AnonymousApexTask | apex=BDI_DataImport_API.importData();
    Run Task Class | cumulusci.tasks.apex.batch.BatchApexWait | class_name= BDI_DataImport_BATCH

robot/Cumulus/resources/NPSP.py
```diff
@@ -16,6 +16,11 @@
 import sys
 from email.mime import text
 
+from cumulusci.robotframework.CumulusCI import CumulusCI
```
davisagli: unused import

davisagli:
@prescod the conflicts don't look entirely unrelated to your changes and they look fairly straightforward to resolve. Here are some pointers:

1. The conflict in cumulusci.yml is because you added a new flow in the same location that @davidmreed did on master. Resolve it by including both new sections in whatever order you like.

2. Some of the conflicts in NPSP.py are because you made whitespace fixes to code that was also changed in master (this is one good reason to do whitespace cleanup in a separate PR that can be approved and merged without discussion so there's less time for conflicts). Resolve them by keeping the change from master (and reapplying the whitespace cleanup if you want).

3. There's also a conflict in NPSP.py similar to number 1, where you added the api_batchdata_load keyword in the same place Sravani added other new keywords. I'll post my other review comments shortly which are going to suggest that you don't need to add this keyword after all...

tasks/generate_bdi_data.py
```diff
@@ -0,0 +1,302 @@
...
+from sqlalchemy import Unicode
+from sqlalchemy.ext.automap import automap_base
+from sqlalchemy.orm import create_session
+
+START_DATE = date(2019, 1, 1)
+
+
+class GenerateBDIData(BaseSalesforceApiTask):
+    def _run_task(self):
+        self.tempdir = tempfile.mkdtemp()
```
davisagli: There's a handy util you can use as a context manager that will take care of deleting the directory:

    from cumulusci.utils import temporary_dir
    with temporary_dir() as path:
        # it'll take care of changing to the directory within this block too

prescod:
You are right that this PR introduced the conflicts. I forgot about the whitespace changes that I made by running black.

Regardless, it was never my intention to merge this PR because it is based off of a branch that does not pass tests, and I assume that *that* branch was never intended for merging to master either.

tasks/generate_bdi_data.py
```diff
@@ -0,0 +1,302 @@
...
+            )
+            self.session.add(payment)
+
+    def make_records(
+        self, model, name, key_field, start, end, amount, paid, payment_amount
+    ):
+        date = START_DATE
+        for i in range(start + 1, end + 1):
+            parent = model(name=name + " " + str(i))
+            print(name + " " + str(i))
```
davisagli: If this is for temporary debugging, take it out; if it's something you want to see in MetaCI, use self.logger

tasks/generate_bdi_data.py
```diff
@@ -0,0 +1,302 @@
...
+                lower(hex(randomblob(16))), -- account_name -> npe01__Account1_Name__c
+                lower(hex(randomblob(16))), -- contact1_lastname -> npe01__Contact1_Lastname__c
+                lower(hex(randomblob(16))), -- donation_name ->  npe01__Donation_Name__c
+                donation_stage, -- donation_stage -> npe01__Donation_Stage__c
+                donation_amount, -- donation_amount -> npe01__Donation_Amount__c
+                donation_date -- donation_date -> npe01__Donation_Date__c
+        -- disabled   FALSE -- do_not_automatically_create_payment -> npe01__Do_Not_Automatically_Create_Payment__c
+        FROM npsp__DataImport__c
+            """
+        )
```
davisagli: This part is clever, in a good way.

davisagli:
@prescod I did intend for feature/bdi-benchmark-test-data to be merged; that's why I made a PR for it.  It needs to be merged so that it's available to run on other branches where the Honeybees team has worked on changes to BDI.

It looks like the failing tests are just the prerelease builds which are still pretty flaky across all branches.

robot/Cumulus/perf/BDITest.robot
prescod: That's true, but is it necessarily better? Do we want Apex code in Robot code? Seems like its punching through layers to me, though I don't feel strongly about it.

Also: I suspect its going to get more complicated in the future. Dry Runs? Field Mappings? If so, it might be easier to maintain in Python.

davisagli: I can see a Python task making more sense if it does get more complicated.

Taking a look at the big picture, it seems dumb to me that we have to use Robot Framework for this at all, just to track the times. We're basically just running a flow. Maybe MetaCI needs a way to record times for the tasks in the flows it runs or something like that.

tasks/generate_bdi_data.py
prescod: > I did intend for feature/bdi-benchmark-test-data to be merged; that's why I made a PR for it.

Okay, thanks for the clarification. We'll need to coordinate what specifically should be merged then. Does the generate.py command line tool need to exist anymore? If so, we need to refactor to avoid the code duplication. Do they need an actual test_data.db file in the directories? etc.

I would advise against merging the parent branch until we've decided on these things.

prescod:
> I did intend for feature/bdi-benchmark-test-data to be merged; that's why I made a PR for it.

Okay, thanks for the clarification. We'll need to coordinate what specifically should be merged then. Does the generate.py command line tool need to exist anymore? If so, we need to refactor to avoid the code duplication. Do they need an actual test_data.db file in the directories? etc.

I would advise against merging the parent branch until we've decided on these things.

davisagli:
@prescod Yeah, I like the direction you're going here. With this we don't need or want a non-task version of generate.py, nor do we want the db in the repo. I think it makes sense to close #4200, merge master into this PR, and make this PR target master.

PR #4247, opened 2019-05-16
Automated Bulk Data Import Testing

prescod:
New Pull Request including some design suggestions from Jason.

There is no more flow. Setup is done in the Robot Setup.

The flow we use in MetaCI will be pasted into the database.

Data deletion is not automatic anymore. If you need it you can do it on the command line or in a flow.

There is now just one task for performance tasks. It runs them all. You can override it if you want to be more specific.


cumulusci.yml
```diff
@@ -106,6 +106,22 @@ tasks:
                 - record_type: Opportunity.NPSP_Default
                   default: true
 
+    test_data_bdi:
+        description: 'Generate and load test data for BDI benchmarks into an org'
+        class_path: tasks.generate_bdi_data.GenerateBDIData
+        options:
+            mapping_yaml: 'testdata/bdi_benchmark/mapping.yml'
+            num_records: 2500
```
davisagli: Does this remain for manual testing, or should we remove it now that the robot suite setup does the same thing with a direct reference to the task class?

robot/Cumulus/resources/NPSP.py
```diff
@@ -967,4 +971,17 @@ def verify_expected_values(self,ns_ind,obj_api,rec_id,**kwargs):
             table=obj_api
        rec=self.salesforce.salesforce_get(table,rec_id)
        for key, value in kwargs.items():
-           self.builtin.should_be_equal_as_strings(rec[key], value)   
\ No newline at end of file
+           self.builtin.should_be_equal_as_strings(rec[key], value)
+
+    def batch_data_import(self):
```
davisagli: Every Robot Framework keyword should have a docstring so that it has a meaningful description in the generated docs.

tasks/generate_bdi_data.py
```diff
@@ -0,0 +1,312 @@
...
+class GenerateBDIData(BaseSalesforceApiTask):
+
+    task_docs = """
+    Use the `num_records` option to specify how many records to generate.
+    Use the `mappings` option to specify a mapping file.
+    """
+
+    task_options = {
+        "num_records": {
+            "description": "how many records to generate",
```
davisagli: more like how many records of each variation, right?

cumulusci.yml
```diff
@@ -149,6 +165,8 @@ tasks:
                 - Allocation__c
                 - General_Accounting_Unit__c
                 - Campaign
+                - npe01__OppPayment__c
```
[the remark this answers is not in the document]
prescod: Fixed in 5cea997c

cumulusci.yml
prescod: Yeah I left it for manual (including UI) testing.

robot/Cumulus/resources/NPSP.py
prescod: Makes sense. Fixed in 8178a46e1

tasks/generate_bdi_data.py
prescod: Clarified:

https://github.com/SalesforceFoundation/Cumulus/pull/4247/commits/b8c77eea40072c84958d57ddb6084d4730d5c7aa#diff-c6088a9879682c1af25379e1729b8258L31

robot/Cumulus/resources/NPSP.py
```diff
@@ -1006,5 +1009,23 @@ def verify_field_values(self,**kwargs):
             res=self.selenium.get_webelement(locator).text
             assert value == res, "Expected {} value to be {} but found {}".format(key,value,res)
 
-           
+    def batch_data_import(self, batchsize):
+        """"Do a BDI import using the API and wait for it to complete"""
+
+        code = """Data_Import_Settings__c diSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
```
prescod: @ptewson-sfdo These are the lines for you to review.

PR #4252, opened 2019-05-16
Adjust the main installer steps to use the ensure_record_type task instead of deploy_pre

davisagli:
I've tested this by deploying to the staging copy of MetaDeploy and running the installation.

One thing I'm wondering is whether we should adjust EnsureRecordTypes to include all active stages instead of just one.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -394,16 +394,20 @@ plans:
         tier: primary
         steps:
             1:
-                flow: dependencies
-                ui_options:
-                    deploy_pre:
-                        account_record_types:
-                            name: Account Record Types
-                        opportunity_record_types:
-                            name: Opportunity Record Types
+                task: update_dependencies
             2:
-                task: install_managed
+                task: deploy
+                options:
+                    path: unpackaged/pre/account_record_types
+                ui_options:
+                    name: Account Record Types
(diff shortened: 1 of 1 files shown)


davidmreed:
If the purpose of the `NPSP_Default` record type is just to get the Record Types feature enabled, I don't think it matters how many stages go into its associated Business Process. Using just one makes the generation a little easier.

NPSP's `update_admin_profile` marks `NPSP_Default` as the default Opportunity record type (in the `trial` metadata, that's changed to `Donation`), but other than that, it's used only in the `dev` config metadata. In an installer context, I don't see any downside to having only one picklist entry. The record type would not be exposed to any profiles other than the admin by the installer, right?

davisagli:
@davidmreed yeah I agree that it should not matter for the installer context; I was thinking of what you ran into with the `dev` metadata. But I guess that means it can wait until later.

davidmreed:
Ah, I misunderstood. Making that change would fix the issue with `dev` metadata, but I want to think about whether there's a better way to fix `dev` so that it's not dependent on what's intended to be an install artifact in the first place.

PR #4256, opened 2019-05-17
Feature/bdi co  create additional objects in bdi

callen-sfdo:
# Critical Changes
- Added ability to Create and Update additional objects as part of Batch Data Import.
# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,46 @@
+public with sharing virtual class BDI_ObjectMappingLogic {
```
ptewson-sfdo: Need the standard comment header on all class files

src/classes/BDI_ObjectWrap.cls
```diff
@@ -0,0 +1,17 @@
+public with sharing class BDI_ObjectWrap {
```
ptewson-sfdo: Standard class file header comments

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Implements the ability to import additional objects during batch data import.
+*/
+
+public with sharing class BDI_AdditionalObjectService {
+
+    /*******************************************************************************************************
+    * @description the data import service we are invoked from
+    */
+    private BDI_DataImportService bdi;
```
ptewson-sfdo: I think we should name these more descriptively (both "bdi" and "bdiFM") - perhaps "importService" or "dataImportService" and then .... cmtFieldMapping or something?

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+        this.bdiFM = BDI_FieldMappingCustomMetadata.getInstance();
+    }
+
+
+    /*******************************************************************************************************
+    * @description Main method to call to begin the additional object import process.
+    */
+    public void importAdditionalObjects(){
+
+        this.generateTiersOfObjMappings();
```
ptewson-sfdo: Do we need "this." here?

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,84 @@
...
+                System.debug('Target Describe isUpdateable: ' + targetDescribe.isUpdateable());
+                System.debug('Target Describe existingSobjectId: ' + objWrap.existingSobjectId);
+                System.debug('Target Describe sobj Id: ' + objWrap.sObj.Id );
+                //Confirm that it is real field, and that it is accessible to the running user.
+                if (targetDescribe != null && targetDescribe.isAccessible()) {
+                    //Confirm that either the sObj Id is null (ie new record) or that the field is updatable if it is not new.
+                    if (objWrap.sObj.Id == null || targetDescribe.isUpdateable()) {
+                        System.debug('Passed checks for field update with: ' + targetFieldName);
+                        Object value = objWrap.di.get(sourceFieldName);
+                        System.debug('Value is: ' + value);
```
kenlewis-sfdo: Do we want to remove the debugs in this class?

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,46 @@
...
```
kenlewis-sfdo: Do we want to remove the debugs in this class?

src/classes/BDI_AdditionalObjectService.cls
ptewson-sfdo: I think I would prefer to spell out generateTiersOfObjectMappings() and in similar cases where it's just saving us a few characters. That's obviously not going to work everywhere (like data import object mapping)

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+
+        //Iterate through the map of tiers, starting with the first tier that actually has a predecessor (ie 2nd tier), and
+        //validate and create each object as appropriate.
+        for (Integer i = 1; i < tierToObjMappings.size(); i++) {
+
+            Map<String,BDI_ObjectWrap[]> objAPINameToObjWraps = new Map<String,BDI_ObjectWrap[]>();
+
+            Map<String,Data_Import_Object_Mapping__mdt> currentTier = tierToObjMappings.get(i);
+
+            for (Data_Import_Object_Mapping__mdt diom : currentTier.values()) {
```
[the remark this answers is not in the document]
ptewson-sfdo: I like objectMapping too

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                Map<Id,DataImport__c> validDIsForObjMapping = new Map<Id,DataImport__c>(bdi.listDI);
+
+                Data_Import_Object_Mapping__mdt predecessor = bdiFM.diomDevNameToDiomMap.get(diom.Predecessor__c);
+                Data_Import_Field_Mapping__mdt[] fieldMappings = bdiFM.diomDevNameToDifmsMap.get(diom.DeveloperName);
+
+                //Only proceed if at least one field mapping exists
+                if (fieldMappings != null && fieldMappings.size() > 0) {
+                    //Map<Id,DataImport__c> disForUpdate = new Map<Id,DataImport__c>();
+                    BDI_ObjectWrap[] objWrapsForProcessing = new BDI_ObjectWrap[]{};
+                    for (DataImport__c di: bdi.listDI) {
```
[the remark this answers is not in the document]
ptewson-sfdo: I agree with preference for 'dataImport' over 'di'

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                        }
+
+                        BDI_ObjectWrap objWrap = new BDI_ObjectWrap(di,diom,fieldMappings);
+
+                        //null out any existing value in the status field for this object mapping
+                        di.put(UTIL_Namespace.adjustNPSPNamespace(diom.Imported_Record_Status_Field_Name__c),'');
+                        String lcFieldName = UTIL_Namespace.adjustNPSPNamespace(diom.Imported_Record_Field_Name__c).toLowerCase();
+
+
+                        //If the imported record field for the object is not populated then create a new Sobject, otherwise update existing.
```
ptewson-sfdo: This isn't a big thing, but these line wraps on the comments can make the code harder to read

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+    /*******************************************************************************************************
+    * @description Validates the field mappings to make sure that the required ones are not null and that at least one non imported record
+    * field is populated.
+    * @param objWrap The BDI_ObjectWrap that should have its field mappings validated.
+    */
+    private Boolean validateFields(BDI_ObjectWrap objWrap){
+
+        Boolean result = false;
+        Boolean anyNonImportedRecordFieldsPopulated = false;
+        Boolean allReqFieldsPopulated = true;
```
ptewson-sfdo: prefer "allRequiredFieldsPopulated" here if only because "req" can be required, requested, etc

src/classes/BDI_ObjectWrap.cls
ptewson-sfdo: Prefer to spell out "BDI_ObjectWrapper" for class name

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+
+    /*******************************************************************************************************
+    * @description the data import service we are invoked from
+    */
+    private BDI_DataImportService bdi;
+    private BDI_FieldMappingCustomMetadata bdiFM;
+
+    private Map<Integer,Map<String,Data_Import_Object_Mapping__mdt>> tierToObjMappings = new Map<Integer,Map<String,Data_Import_Object_Mapping__mdt>>();
+    private Map<String,Map<String,Data_Import_Object_Mapping__mdt>> predecessorDevNameToSuccessors = new Map<String,Map<String,Data_Import_Object_Mapping__mdt>>();
+    private Map<Id,String> diIdToError = new Map<Id,String>();
```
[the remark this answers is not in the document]
callen-sfdo: yep, unused, removing

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+    private Set<String> importedRecordFieldNames = new Set<String>();
+
+    public Set<String> additionalObjectDiomDevNames = new Set<String>();
+
+    private String namespace = UTIL_Namespace.getNamespace();
```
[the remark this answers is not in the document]
callen-sfdo: Yep, from a previous version, removing

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: No, we don't need this, I was experimenting with something and left it in on accident.  I'll make the name longer in this case

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: I'll change it to dataImportService and fieldMapping.

src/classes/BDI_ObjectWrap.cls
callen-sfdo: Changed this.

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: Added

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,66 @@
...
+            // If the payment is specified and payment allocations are enabled, always use that first.
+            if ( paymentValue != null && settings.Payment_Allocations_Enabled__c) {
+                objWrap.sObj.put(UTIL_Namespace.adjustNPSPNamespace(PAYMENT_FN),paymentValue);
+            } else if ( opportunityValue != null ) {
+                objWrap.sObj.put(UTIL_Namespace.adjustNPSPNamespace(OPPORTUNITY_FN),opportunityValue);
+            } else if ( campaignValue != null ) { 
+                objWrap.sObj.put(UTIL_Namespace.adjustNPSPNamespace(CAMPAIGN_FN),campaignValue);
+            }
+
+            System.debug('BDI_CustObjMappingGAUAllocation at end SObj is: ' + objWrap.sObj);
```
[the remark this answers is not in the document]
callen-sfdo: removed, hadn't done a final pass to clean them all up yet.

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,66 @@
+public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMappingLogic {
+
+    public override BDI_ObjectWrap[] populateObjects(BDI_ObjectWrap[] objWraps) {
+
+        Allocations_Settings__c settings = UTIL_CustomSettingsFacade.getAllocationsSettings();
+
+        System.debug('BDI_CustObjMappingGAUAllocation ');
```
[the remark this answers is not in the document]
callen-sfdo: removed, hadn't done a final pass to clean them all up yet.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+
+            Map<String,BDI_ObjectWrap[]> objAPINameToObjWraps = new Map<String,BDI_ObjectWrap[]>();
+
+            Map<String,Data_Import_Object_Mapping__mdt> currentTier = tierToObjMappings.get(i);
+
+            for (Data_Import_Object_Mapping__mdt diom : currentTier.values()) {
+
+                BDI_ObjectMappingLogic objMappingLogic = getObjectMappingLogicClass(diom);
+
+                Map<Id,DataImport__c> validDIsForObjMapping = new Map<Id,DataImport__c>(bdi.listDI);
```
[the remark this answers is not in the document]
callen-sfdo: Yep, it was left over from a previous version, removed

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                        // Skip over records that have already failed
+                        if (di.Status__c == BDI_DataImport_API.bdiFailed) {
+                            continue;
+                        }
+
+                        BDI_ObjectWrap objWrap = new BDI_ObjectWrap(di,diom,fieldMappings);
+
+                        //null out any existing value in the status field for this object mapping
+                        di.put(UTIL_Namespace.adjustNPSPNamespace(diom.Imported_Record_Status_Field_Name__c),'');
+                        String lcFieldName = UTIL_Namespace.adjustNPSPNamespace(diom.Imported_Record_Field_Name__c).toLowerCase();
```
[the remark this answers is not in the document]
callen-sfdo: Yes, left over from a previous version, removing.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+
+                        Database.SaveResult[] updateSaveResults = UTIL_DMLService.updateRecords(sObjsForUpdate,false);
+                        //Cycle through all the upsert results and update the data import records with status appropriately.
+                        for (Integer t = 0; t < updateSaveResults.size(); t++) {
+                            Database.SaveResult result = updateSaveResults[t];
+
+                            BDI_ObjectWrap objWrap = objWrapsForUpdate[t];
+
+                            if (result.isSuccess()) {
+                                objWrap.di.put(UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Imported_Record_Status_Field_Name__c),System.label.bdiUpdated);
```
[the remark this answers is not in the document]
callen-sfdo: I would prefer not to increase heap size by storing duplicate values as new variables in the wrapper.  I had considered just replacing all the values in the cmt records themselves (but not trying to update obviously) but I had concerns about whether the code would look ambiguous in places and we wouldn't have a way of knowing whether a namespace had been removed without adding a way to track that.  I know it looks messy but the current way is the most explicit about what is going on.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                                } else {
+                                    objAPINameToObjWraps.put(diom.Object_API_Name__c,new BDI_ObjectWrap[]{objWrap});
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            //If there are any records to create, then combine them into
```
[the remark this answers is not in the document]
callen-sfdo: Fixed this

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: This line wrapping is done by github because of the insanely narrow width they have for viewing code.  Are you suggesting we format everything around this limit?

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: Updated

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                                objWrap.di.put(UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Imported_Record_Field_Name__c),objWrap.sObj.Id);
+                            } else { 
+                                bdi.LogBDIError(objWrap.di, result.getErrors()[0].getMessage(),UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Imported_Record_Status_Field_Name__c));
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
callen-sfdo: I've broken it down into prepareObjectWrappers and importObjectWrappers, which should make it less monolithic

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: removed

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: removed

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,367 @@
...
+                BDI_ObjectMappingLogic objMappingLogic = getObjectMappingLogicClass(diom);
+
+                Map<Id,DataImport__c> validDIsForObjMapping = new Map<Id,DataImport__c>(bdi.listDI);
+
+                Data_Import_Object_Mapping__mdt predecessor = bdiFM.diomDevNameToDiomMap.get(diom.Predecessor__c);
+                Data_Import_Field_Mapping__mdt[] fieldMappings = bdiFM.diomDevNameToDifmsMap.get(diom.DeveloperName);
+
+                //Only proceed if at least one field mapping exists
+                if (fieldMappings != null && fieldMappings.size() > 0) {
+                    //Map<Id,DataImport__c> disForUpdate = new Map<Id,DataImport__c>();
```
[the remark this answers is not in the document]
callen-sfdo: removed

src/classes/BDI_AdditionalObjectService.cls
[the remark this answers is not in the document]
callen-sfdo: changed this to objMapping which matches iteration variables elsewhere

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,66 @@
...
+        String PAYMENT_FN = 'npsp__Payment__c';
+        String OPPORTUNITY_FN = 'npsp__Opportunity__c';
+        String CAMPAIGN_FN = 'npsp__Campaign__c';
+
+        for (BDI_ObjectWrap objWrap : objWraps) {
+            Map<String, Schema.DescribeFieldResult> fieldDescribeMap = UTIL_Describe.getAllFieldsDescribe(UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Object_API_Name__c));
```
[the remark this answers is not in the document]
callen-sfdo: See other comment about adjusting namespaces, I would either want to do this in the CMT or not at all for heap size reasons.

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,66 @@
+public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMappingLogic {
+
+    public override BDI_ObjectWrap[] populateObjects(BDI_ObjectWrap[] objWraps) {
+
+        Allocations_Settings__c settings = UTIL_CustomSettingsFacade.getAllocationsSettings();
+
+        System.debug('BDI_CustObjMappingGAUAllocation ');
+        String PAYMENT_FN = 'npsp__Payment__c';
+        String OPPORTUNITY_FN = 'npsp__Opportunity__c';
+        String CAMPAIGN_FN = 'npsp__Campaign__c';
```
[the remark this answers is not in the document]
callen-sfdo: expanded

src/classes/BDI_AdditionalObjectService.cls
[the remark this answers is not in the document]
callen-sfdo: made this change

callen-sfdo:
@kenlewis-sfdo @ptewson-sfdo Hey guys, I have made the vast majority of the changes you recommended. Please review the changes, and the remaining open conversations that I responded to.

src/classes/BDI_AdditionalObjectService.cls
ptewson-sfdo: I mean, when possible - it's not an absolute requirement but it helps. For comments especially, where we have a lot more flexibility around breaking lines. 

I've always assumed the narrow review window was actually intentional to enforce good practice around line breaks, but I might just be speculating?

ptewson-sfdo: Just out of curiosity here, I checked with our style guide: "Try to keep lines under 120 characters to ensure better readability." There's good research that backs that up for readability as well. I think the github window is 120 characters?

src/classes/BDI_AdditionalObjectServiceTest.cls
```diff
@@ -0,0 +1,751 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Batch Data Import
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Unit tests to cover the creation of additional objects during batch data import.
+*/
+
+@isTest
```
ptewson-sfdo: If it doesn't interfere with the test execution, can we add IsParallel=true to the class definition? We really should be investing time to back-update our tests but at least let's try and get in the habit of writing our new tests that way (minimizing our test execution time has big wins for our CI process)

callen-sfdo: I've made it parallel, we'll see if it causes any problems.

src/classes/BDI_DataImportService.cls
```diff
@@ -41,16 +41,20 @@ global with sharing class BDI_DataImportService {
     public static final String FM_HELP_TEXT = 'Help Text';
     public static final String FM_DATA_IMPORT_FIELD_MAPPING = 'Data Import Field Mapping';
 
-    private BDI_FieldMapping fieldMapper;
+    public BDI_FieldMapping fieldMapper;
```
ptewson-sfdo: Just curious, what was the motivation to switch this to public?

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,391 @@
...
+    */
+
+    private void importObjectWrappers (Map<String,BDI_ObjectWrapper[]> objAPINameToObjWraps) {
+        //If there are any records to create, then combine them into lists for update
+        if (objAPINameToObjWraps.size() > 0) {
+            for (String objApiName : objAPINameToObjWraps.keySet()) {
+
+                BDI_ObjectWrapper[] objWraps = objAPINameToObjWraps.get(objApiName);
+                
+                Sobject[] sObjsForUpdate = new Sobject[]{};
```
ptewson-sfdo: Style guide suggest "SObject" over "Sobject"

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,92 @@
...
+
+    public virtual BDI_ObjectWrapper[] populateObjects(BDI_ObjectWrapper[] objWraps) {
+        if (objWraps == null) {
+            return objWraps;
+        }
+
+        for (BDI_ObjectWrapper objWrap : objWraps) {
+            Map<String, Schema.DescribeFieldResult> fieldDescribeMap = UTIL_Describe.getAllFieldsDescribe(UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Object_API_Name__c));
+
+            objWrap.sObj = Schema.getGlobalDescribe().get(UTIL_Namespace.adjustNPSPNamespace(objWrap.objMapping.Object_API_Name__c)).newSObject();
```
ptewson-sfdo: is this worth supporting as a method in UTIL_Describe?

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,92 @@
...
+
+            // If the object already has an existing sobject Id, then set it in the sObj
+            if (objWrap.existingSobjectId != null) {
```
ptewson-sfdo: existingSObjectId

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,92 @@
...
+                //Confirm that it is real field, and that it is accessible to the running user.
+                if (targetDescribe != null && targetDescribe.isAccessible()) {
+
+                    //Confirm that either the sObj Id is null (ie new record) or that the field is updatable if it is not new.
+                    if (objWrap.sObj.Id == null || targetDescribe.isUpdateable()) {
+
+                        Object value = objWrap.di.get(sourceFieldName);
+
+                        //only copy values that are not null to avoid overwriting data when updating existing sobjects.
+                        if (value != null) {
```
ptewson-sfdo: Do we need to include the logic to map convert Boolean from DI Picklist to support checkbox fields?

src/classes/BDI_ObjectWrapper.cls
```diff
@@ -0,0 +1,56 @@
...
+* @description Wrapper class used to contain information about a particular object mapping and data import record combination.
+*/
+
+public with sharing class BDI_ObjectWrapper {
+    public String wrapId;
+
+    public Data_Import_Object_Mapping__mdt objMapping;
+    public Data_Import_Object_Mapping__mdt predecessorObjMapping;
+    public Data_Import_Field_Mapping__mdt[] fieldMappings;
+    public DataImport__c di;
```
ptewson-sfdo: dataImport

src/classes/BDI_DataImportService.cls
```diff
@@ -41,16 +41,20 @@ global with sharing class BDI_DataImportService {
...
     * @description constructor
     * @param isDryRun whether to run in Dry Run mode or commit records.
     */
     public BDI_DataImportService(Boolean isDryRun, BDI_FieldMapping fieldMapper) {
-        contactService = new BDI_ContactService(this);
         this.isDryRun = isDryRun;
         this.fieldMapper = fieldMapper;
+
+        contactService = new BDI_ContactService(this);
+        if (fieldMapper != null && String.valueOf(fieldMapper).split(':')[0] == 'BDI_FieldMappingCustomMetadata'){
```
ptewson-sfdo: I see why we need to check if we can support the additional objects here, but could we at least use "instanceOf" rather than a string comparison?

Even better is if we can find a slicker way to handle this behavior consistently. Help text mapping is inconsistent with any of the additional objects because it doesn't have the predecessor relationship and other additional metadata available, right? This wants something less brute force I think.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,391 @@
...
+
+        Boolean result = false;
+        Boolean anyNonImportedRecordFieldsPopulated = false;
+        Boolean allRequiredFieldsPopulated = true;
+
+        String requiredFieldErrorMsg = System.label.bdiAdditionalObjRequiredFieldsNull;
+
+        // First determine if there is at least one field populated that was not automatically populated by the import of another object.
+        // This is done to ensure that the user actually intended to attempt the creation of this object.
+        for (Data_Import_Field_Mapping__mdt fieldMapping : objWrap.fieldMappings) {
```
ptewson-sfdo: I got tripped up by the duplication of the fieldMapping variable name in here and the class variable "BDI_FieldMappingCustomMetadata fieldMapping". Recognizing you may have renamed this one based on previous feedback, it would still be nice not to have that naming overlap

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,102 @@
...
+
+            if (objWrap.existingSobjectId != null) {
+                objWrap.sObj.put('Id', objWrap.existingSobjectId);
+            }
+
+            Object campaignValue;
+            Object paymentValue;
+            Object opportunityValue;
+
+            for (Data_Import_Field_Mapping__mdt fieldMapping : objWrap.fieldMappings) {
```
ptewson-sfdo: Similar here with the repeat use of "fieldMapping"

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,391 @@
...
+* @description Implements the ability to import additional objects during batch data import.
+*/
+
+public with sharing class BDI_AdditionalObjectService {
+
+    /*******************************************************************************************************
+    * @description the data import service we are invoked from
+    */
+    private BDI_DataImportService dataImportService;
+    private BDI_FieldMappingCustomMetadata fieldMapping;
```
ptewson-sfdo: Would it be possible to set this up as a generic BDI_FieldMapping and define the necessary behaviors in the base class and implement them in both sub-classes? If the help text variant returns empty collections for those behaviors in a way that skips the additional object logic, that would allow us to execute this inline without having to test the type of the mapping class and branch accordingly

src/classes/BDI_DataImportService.cls
callen-sfdo: It needs to be accessible to the Additional object service for parsing out the tiers.

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: Which part? the construction of the sobject?

src/classes/BDI_DataImportService.cls
callen-sfdo: I think I can change the method of class comparison, I forget the reason I went this way.  In general though there is no way to additional object mapping with helptext so there needs to be a switch somewhere at some level.

ptewson-sfdo: What about a method in the base class, "Boolean supportsAdditionalObjects()"? Then the help text variant returns false and the CMT variant returns true?

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: The problem is that for the sake of efficiency I am attempting to query and sort out the data import custom metadata only once when the BDI_FieldMappingCustomMetadata is initially called, and not do it again in the additional object service, which means that I need those variables to be present and accessible in the version of the mapping class that gets to the AdditionalObjectService.

I could add those variables to base class, but that would be muddling the different implementations since it would be referring to metadata that the helptext mapping doesn't use.  In that situation I would then need to check whether those maps were populated before attempting to process.  In the end we are still doing a check, just on a different thing.

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: updated to fieldMappingCustomMetadata

src/classes/BDI_CustObjMappingGAUAllocation.cls
callen-sfdo: updated to fieldMappingCustomMetadata

src/classes/BDI_ObjectWrapper.cls
callen-sfdo: updated

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: I've added a version of the logic that was in the DataImportService to do this.

src/classes/BDI_DataImportService.cls
callen-sfdo: We could do that, but we will still need to check the class type unless we move the map variables that contain the mappings into the base field mapping class (see my other comment for more detail)

callen-sfdo:
@ptewson-sfdo @kenlewis-sfdo Made all the renaming changes requested and also added in support for mapping string values to booleans.  Please review.

src/classes/BDI_CustObjMappingGAUAllocation.cls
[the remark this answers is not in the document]
kenlewis-sfdo: What about adjusting the namespace in the BDI_ObjectWrapper internally?  That way classes that use an instance of the BDI_ObjectWrapper could reference its objectAPIName using the `objWrap.objectAPIName` attribute without needing to explicitly check the namespace.

In BDI_ObjectWrapper:
```
    public String objectAPIName {
        get {
            if (objectAPIName == null) {
                objectAPIName = UTIL_Namespace.adjustNPSPNamespace(objMapping.Object_API_Name__c);
            }
            return objectAPIName;
        }
    }
```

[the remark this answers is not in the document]
kenlewis-sfdo: Or maybe just a public method on the BDI_ObjectWrapper if trying to save the heap space?

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,124 @@
...
+
+            // If the destination field is a Boolean and the DI field intermediate requires interpretation to Boolean
+            // This allows for mapping checkboxes to a tri-state picklist - null, True, False - or a text with either 
+            // null, "True", or "False"
+            if (targetFieldDescribe.getType() == Schema.DisplayType.BOOLEAN && !(value instanceof Boolean)) {
+                value = Boolean.valueOf(value);
+            }
+
+            // If the destination field is an integer, and the source field is a double with a scale of 0 then
+            // convert the
```
kenlewis-sfdo: This comment could use an ending

src/classes/BDI_AdditionalObjectService.cls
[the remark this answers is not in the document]
callen-sfdo: Per our discussion, I've moved the removal of the npsp namespace to the BDI_FieldMappingCustomMetadata class in order to clean up the code appearance and slightly improve performance.

src/classes/BDI_CustObjMappingGAUAllocation.cls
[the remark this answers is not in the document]
callen-sfdo: Per our discussion, I've moved the removal of the npsp namespace to the BDI_FieldMappingCustomMetadata class in order to clean up the code appearance and slightly improve performance.

callen-sfdo:
@kenlewis-sfdo @ptewson-sfdo I've pushed a change that should do all the npsp namespace removal up front in the BDI_FieldMappingCustomMetadata class as needed.  This should make the code clearer and slightly improve performance.

src/labels/CustomLabels.labels
```diff
@@ -1913,6 +1913,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>bdiAccountCustomIdError</shortDescription>
         <value>The Account Custom Unique ID setting field {0} must have matching fields in the NPSP Data Import object whose API Names are {1} and {2}.</value>
     </labels>
+    <labels>
+        <fullName>bdiAdditionalObjPredNotFound</fullName>
+        <categories>Data Importer</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>bdiAdditionalObjPredNotFound</shortDescription>
+        <value>Error: Predecessor object not found so record could not be created.</value>
```
melissabarber: Is it possible to call out the object dependency field (field label, not field API name)?

Error: record not created. This record is dependent on another record which couldn't be found. The field with this object dependency is [dependency field].

melissabarber: After reviewing with Charles, decided to go with the following error message:

Error: record not created. This record is dependent on another record which couldn't be found.

src/classes/BDI_DataImportService.cls
```diff
@@ -41,16 +41,20 @@ global with sharing class BDI_DataImportService {
...
     * @description constructor
     * @param isDryRun whether to run in Dry Run mode or commit records.
     */
     public BDI_DataImportService(Boolean isDryRun, BDI_FieldMapping fieldMapper) {
-        contactService = new BDI_ContactService(this);
         this.isDryRun = isDryRun;
         this.fieldMapper = fieldMapper;
+
+        contactService = new BDI_ContactService(this);
+        if (fieldMapper instanceof BDI_FieldMappingCustomMetadata){
```
ptewson-sfdo: Thanks for updating this from the text comparison to the name - would still prefer this was handled by a method defined in BDI_FieldMapping in the form of "public Boolean supportsAdditionalObjects()" that could allow the field mapper classes to describe their behavior.

callen-sfdo:
@ptewson-sfdo @kenlewis-sfdo  Pushed a few changes to the additional objects branch.  One will remove the "Campaign" field mapping from the GAU Allocation object mappings.  This is because its not really needed, and it was causing issues with detecting if someone was trying to create a GAU.  Also I made "Precent" for GAU Allocations not required, and changed the error message for the predecessor not existing based on Melissa's feedback.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,449 @@
...
+
+    public Set<String> objMappingDevNames = new Set<String>();
+
+    /*******************************************************************************************************
+    * @description Constructor
+    * @param dataImportService The data import service instance that invokes us
+    */
+    public BDI_AdditionalObjectService(BDI_DataImportService dataImportService) {
+        this.dataImportService = dataImportService;
+        this.fieldMappingCustomMetadata = BDI_FieldMappingCustomMetadata.getInstance();
```
ptewson-sfdo: Suggestion: get the field Mapper that is a member of the data import service rather than accessing through the singleton

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: fixed

src/classes/BDI_ObjectMappingLogic.cls
callen-sfdo: Yep, done.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,449 @@
...
+    /*******************************************************************************************************
+    * @description the data import service we are invoked from
+    */
+    private BDI_DataImportService dataImportService;
+    private BDI_FieldMappingCustomMetadata fieldMappingCustomMetadata;
+
+    private Map<Integer,Map<String,Data_Import_Object_Mapping__mdt>> tierToObjMappings = 
+        new Map<Integer,Map<String,Data_Import_Object_Mapping__mdt>>();
+
+    private Map<String,Map<String,Data_Import_Object_Mapping__mdt>> predecessorDevNameToSuccessors =
```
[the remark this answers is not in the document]
callen-sfdo: changed

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,103 @@
...
+                        } else if (fieldMapping.Target_Field_API_Name__c == CAMPAIGN_FIELDNAME) {
+                            campaignValue = value;
+                        } else {
+                            castAndCopyField(objWrap, sourceFieldName, sourceFieldDescribe, targetFieldName, targetFieldDescribe);
+                        }
+                    }
+                }
+            }
+
+            // If the payment is specified and payment allocations are enabled, always use that first.
```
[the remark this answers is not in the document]
callen-sfdo: If the payment allocations are not enabled then it will use opportunity instead regardless of payment value. I've verified this with Alison as the desired behavior.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,449 @@
...
+
+public with sharing class BDI_AdditionalObjectService {
+
+    /*******************************************************************************************************
+    * @description the data import service we are invoked from
+    */
+    private BDI_DataImportService dataImportService;
+    private BDI_FieldMappingCustomMetadata fieldMappingCustomMetadata;
+
+    private Map<Integer,Map<String,Data_Import_Object_Mapping__mdt>> tierToObjMappings =
```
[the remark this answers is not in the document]
callen-sfdo: updated to objMappingsByTier, don't think target is necessary here since all object mappings are target object mappings.

src/classes/BDI_FieldMappingCustomMetadata.cls
```diff
@@ -69,52 +77,89 @@ public with sharing class BDI_FieldMappingCustomMetadata implements BDI_FieldMap
         Map<String,String> dataImportFieldToTargetField = new Map<String,String>();
 
         String namespace = UTIL_Namespace.getNamespace();
```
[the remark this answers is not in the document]
callen-sfdo: Yep, it was from a previous version, removing.

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,103 @@
...
+* the Opportunity__c, Payment__c and Campaign__c fields should be populated.
+*/
+
+public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMappingLogic {
+
+    public override BDI_ObjectWrapper[] populateObjects(BDI_ObjectWrapper[] objWraps) {
+
+        Allocations_Settings__c settings = UTIL_CustomSettingsFacade.getAllocationsSettings();
+
+        Map<String, Schema.DescribeFieldResult> sourceFieldDescribeMap = UTIL_Describe.getAllFieldsDescribe(UTIL_Namespace.adjustNPSPNamespace('npsp__DataImport__c'));
```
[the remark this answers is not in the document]
callen-sfdo: Sure, that looks a bit cleaner.

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,103 @@
...
+
+        String PAYMENT_FIELDNAME = UTIL_Namespace.adjustNPSPNamespace('npsp__Payment__c');
+        String OPPORTUNITY_FIELDNAME = UTIL_Namespace.adjustNPSPNamespace('npsp__Opportunity__c');
+        String CAMPAIGN_FIELDNAME = UTIL_Namespace.adjustNPSPNamespace('npsp__Campaign__c');
+
+        for (BDI_ObjectWrapper objWrap : objWraps) {
+            Map<String, Schema.DescribeFieldResult> fieldDescribeMap = UTIL_Describe.getAllFieldsDescribe(objWrap.objMapping.Object_API_Name__c);
```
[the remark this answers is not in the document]
callen-sfdo: Changed this to targetFieldDescByFieldName and the other one to sourceFieldDescByFieldName

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,103 @@
...
+
+            objWrap.sObj = Schema.getGlobalDescribe().get(objWrap.objMapping.Object_API_Name__c).newSObject();
+
+            if (objWrap.existingSObjectId != null) {
+                objWrap.sObj.put('Id', objWrap.existingSObjectId);
+            }
```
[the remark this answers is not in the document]
callen-sfdo: I thought about that, but I wanted to have it done as part of the overridable populateObjects method in field mapping to provide more flexibility about how the sobject is initialized (for instance they could query the existing object and initialize the sobject from there).

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -0,0 +1,103 @@
...
+
+                Schema.DescribeFieldResult sourceFieldDescribe = sourceFieldDescribeMap.get(sourceFieldName.toLowerCase());
+                Schema.DescribeFieldResult targetFieldDescribe = fieldDescribeMap.get(targetFieldName.toLowerCase());
+
+                //Confirm that it is real field, and that it is accessible to the running user.
+                if (targetFieldDescribe != null && targetFieldDescribe.isAccessible()) {
+                    //Confirm that either the sObj Id is null (ie new record) or that the field is updatable if it is not new.
+                    if (objWrap.sObj.Id == null || targetFieldDescribe.isUpdateable()) {
+                        // If the target field mapping is one of the key parent objects, then extract the values for later logic
+                        if (fieldMapping.Target_Field_API_Name__c == PAYMENT_FIELDNAME) {
```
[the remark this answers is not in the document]
callen-sfdo: The field mappings being looped through here are only for one object mapping.

src/classes/BDI_ObjectMappingLogic.cls
```diff
@@ -0,0 +1,124 @@
...
+        
+        for (BDI_ObjectWrapper objWrap : objWraps) {
+            Map<String, Schema.DescribeFieldResult> targetFieldDescribeMap = UTIL_Describe.getAllFieldsDescribe(objWrap.objMapping.Object_API_Name__c);
+
+            objWrap.sObj = Schema.getGlobalDescribe().get(objWrap.objMapping.Object_API_Name__c).newSObject();
+
+            // If the object already has an existing SObject Id, then set it in the sObj
+            if (objWrap.existingSObjectId != null) {
+                objWrap.sObj.put('Id', objWrap.existingSObjectId);
+            }
```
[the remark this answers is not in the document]
callen-sfdo: See other comment, this is intentionally an overrideable method to increase flexibility.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -0,0 +1,449 @@
...
+                Map<String,Data_Import_Object_Mapping__mdt> currentTier = 
+                    new Map<String,Data_Import_Object_Mapping__mdt>();
+
+                Map<String,Data_Import_Object_Mapping__mdt> nextTier = 
+                    new Map<String,Data_Import_Object_Mapping__mdt>();
+
+                if (tierToObjMappings.get(i) != null) {
+                    currentTier = tierToObjMappings.get(i);
+
+                    for (String objMappingDevName:currentTier.keyset()) {
```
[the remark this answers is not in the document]
callen-sfdo: Updated

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: This was updated.

src/classes/BDI_CustObjMappingGAUAllocation.cls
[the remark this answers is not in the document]
kenlewis-sfdo: 👍🏽

src/classes/BDI_AdditionalObjectService.cls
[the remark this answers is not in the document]
kenlewis-sfdo: ok cool!

src/classes/BDI_ObjectMappingLogic.cls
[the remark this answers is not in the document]
kenlewis-sfdo: Ok, thanks.  Final question: would it make sense to put this in the else block of the `prepareObjectWrappers` method within the `BDI_AdditionalObjectService` class around [line 157](https://github.com/SalesforceFoundation/Cumulus/blob/029797fcd536f5cf1343618c8f712c6e3121d209/src/classes/BDI_AdditionalObjectService.cls#L157)? Since at that point the objWrapper has an existing SObjectId?

src/classes/BDI_FieldMappingCustomMetadata.cls
```diff
@@ -39,6 +39,12 @@ public with sharing class BDI_FieldMappingCustomMetadata implements BDI_FieldMap
 
     public static final String DEFAULT_DATA_IMPORT_FIELD_MAPPING_SET_NAME = 'Default_Field_Mapping_Set';
 
+    public Data_Import_Field_Mapping_Set__mdt diFieldMappingSet;
+    public String fieldMappingSetName;
+
+    public Map<String,Data_Import_Object_Mapping__mdt> objMappingDevNameToObjMapping = new Map<String,Data_Import_Object_Mapping__mdt>();
```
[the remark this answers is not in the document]
callen-sfdo: Updated

src/classes/BDI_FieldMappingCustomMetadata.cls
```diff
@@ -39,6 +39,12 @@ public with sharing class BDI_FieldMappingCustomMetadata implements BDI_FieldMap
...
+    public Map<String,Data_Import_Field_Mapping__mdt[]> objMappingDevNameToFieldMappings = new Map<String,Data_Import_Field_Mapping__mdt[]>();
```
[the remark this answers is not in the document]
callen-sfdo: Updated

src/classes/BDI_FieldMappingCustomMetadata.cls
```diff
@@ -69,52 +77,89 @@ public with sharing class BDI_FieldMappingCustomMetadata implements BDI_FieldMap
         Map<String,String> dataImportFieldToTargetField = new Map<String,String>();
```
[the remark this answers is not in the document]
callen-sfdo: updated

src/classes/BDI_ObjectMappingLogic.cls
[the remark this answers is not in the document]
callen-sfdo: I'd prefer to just keep it with all the other mapping logic inside that method in case there is some edge case reason that a client would want to do it differently.

callen-sfdo:
@kenlewis-sfdo @ptewson-sfdo pushed another few commits that should resolve the latest round of feedback. Please do a final pass on review.

cpose:
***lurch: attach W-032311

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmAHQA0" target="_blank">W-032311</a>

PR #4257, opened 2019-05-17
My experiments with custom fields.

prescod:
I'm attempting to create custom fields for Do_Not_Automatically_Create_Payment__c but I'm not sure of some of the details and my experiments were not working for me. I need to switch to another project so I don't have time for Trailhead this week. :( 

Please look for the question marks and clarify what I should do. Do I need a valueSet to go with a pickList?


unpackaged/config/perf/objects/___NAMESPACE___DataImport__c.object
```diff
@@ -0,0 +1,25 @@
...
+        <required>false</required>
+        <trackFeedHistory>false</trackFeedHistory>
+        <trackTrending>false</trackTrending>
+        <type>Date</type>
+    </fields>
+    <fields>
+        <fullName>npe01__Do_Not_Automatically_Create_Payment__c</fullName>
+        <description>????</description>
+        <externalId>false</externalId>
+        <inlineHelpText>npe01__Do_Not_Automatically_Create_Payment__c</inlineHelpText>
```
davisagli: Should be Opportunity.npe01__Do_Not_Automatically_Create_Payment__c

unpackaged/config/perf/objects/___NAMESPACE___DataImport__c.object
```diff
@@ -0,0 +1,25 @@
...
+        <label>Do Not Automatically Create Payment</label>
+        <required>false</required>
+        <trackFeedHistory>false</trackFeedHistory>
+        <trackTrending>false</trackTrending>
+        <type>????</type>
```
davisagli: Something like this:

        <type>Picklist</type>
        <valueSet>
            <valueSetDefinition>
                <sorted>false</sorted>
                <value>
                    <fullName>True</fullName>
                    <default>false</default>
                    <label>True</label>
                </value>
                <value>
                    <fullName>False</fullName>
                    <default>false</default>
                    <label>False</label>
                </value>
            </valueSetDefinition>
        </valueSet>

prescod:
I did try that yesterday (and again today). New error message:

$ cci task run deploy_qa_config -o unmanaged False -o namespace_inject npsp -o path unpackaged/config/perf

Error: Update of CustomField npsp__DataImport__c.npe01__Do_Not_Automatically_Create_Payment__c: Error on line 14, col 13: Cannot create a new component with the namespace: npe01.  Only components in the same namespace as the organization can be created through the API

prescod:
I pushed the version of the code that generates this error message.

davisagli:
@prescod I thought you might get something like that. The field you're adding to DataImport__c should not have the npe01__ prefix. (It should remain for the field you're mapping to on Opportunity, since that one was actually added by the npe01 package)

PR #4262, opened 2019-05-17
BGE: Remove "Update Opportunity" link in BGE donation matching portal when payments are enabled

kenlewis-sfdo:
- Batch Gift Entry will now only allow Users to select "Apply New Payment" on an open Opportunity if Automatic Payment Creation is enabled in the NPSP Settings.  Likewise, Users will only be allowed to select "Update Opportunity" if Automatic Payment Creation is disabled.

# Critical Changes

# Changes

- Batch Gift Entry now provides a more customized user experience when reviewing possible matching donations. If Automatic Payment Creation is disabled in NPSP Settings, your only option when you click Review Donations will be to Update Opportunity. Previously, users with Automatic Payment Creation disabled were also given the option to Apply New Payment.

# Issues Closed

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
lurch: attach W-033025

src/aura/BGE_DonationSelector/BGE_DonationSelector.cmp
```diff
@@ -42,6 +42,7 @@
     <aura:attribute name="unpaidPayments" type="List" description="List of unpaid Payments" access="global" />
     <aura:attribute name="openOpportunities" type="List" description="List of open Opportunities" access="global" />
     <aura:attribute name="labels" type="Map" description="Labels retrieved in Apex related to Schema information." access="global"/>
+    <aura:attribute name="isPaymentCreationEnabled" type="Boolean" description="True if Automatic Payment Creation is enabled." access="public" default="false" />
```
ptewson-sfdo: @lparrott - just want to call your attention to this new attribute in case it causes any concern. You were using this component for GEM, right?

lparrott: We sure are, this change shouldn't be an issue though, thanks for pointing it out!

kenlewis-sfdo:
**lurch: attach W-033025

PR #4270, opened 2019-05-21
Refactor *_org flows and unpackaged metadata

davidmreed:
This pull request:

 - Refactors the regression and trial org flows to match the pattern established by SAL. (Note that `install_regression` users should now use `regression_org`).
 - Migrates test data sets to `datasets/`, again to match directory structure established by SAL.
 - Creates tasks to use in configuring unmanaged orgs to use CRLPs.
 - Breaks `config/dev` metadata's dependency on the `NPSP_Default` business process installed in `pre`, to improve compatibility with the `EnsureRecordTypes` task in the future.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


cumulusci.yml
```diff
@@ -290,18 +299,26 @@ flows:
...
-            6:
+            2:
                 task: update_admin_profile
                 options:
                     managed: true
 
+    config_trial:
+        steps:
+            1:
+                flow: config_qa
```
davisagli: At some point we probably want to refactor it so that you can get the trial-like configuration with or without the additional QA-specific metadata from unpackaged/config/qa

cumulusci.yml
```diff
@@ -290,18 +299,26 @@ flows:
...
+            2:
+                task: enable_customizable_rollups
+            3:
+                task: enable_incremental_rollups
```
davisagli: We could arrange for these to happen with a single anonymous Apex call instead of 2, but it's not a big deal.

cumulusci.yml
```diff
@@ -290,18 +299,26 @@ flows:
...
 
-    install_regression:
+    config_regression:
         steps:
-            # Add QA metadata needed for regression
-            5:
+            1:
                 task: deploy_qa_config
                 options:
                     unmanaged: False
```
davisagli: The table you're working on made me realize that this refactoring lost the `config_managed` flow here (which was the last step of install_regression in CumulusCI). The automated tests used for regression need the full configuration applied by config_managed, plus the additional metadata added by deploy_qa_config.

jstvz: Sorry, I'm confused. Here's what `cci flow info install_regression` says on my machine:

```yaml
description: Install the latest beta dependencies and upgrade to the latest beta version from the most recent production version
steps:
    1:
        flow: beta_dependencies
    2:
        task: install_managed
        options:
            version: previous
    3:
        task: install_managed_beta
    4:
        flow: config_managed
```

Are we moving `config_managed` out of `install_regression` in CumulusCI?

davisagli: Ah right. I have proposed that change in https://github.com/SFDO-Tooling/CumulusCI/pull/1088, but was forgetting that this was written before that.

jstvz:
Since we're pushing for uniformity, what about anonymous Apex? SAL has its scripts in `unpackaged/anon_apex` while Cumulus has its in `scripts/`. Do we have a preference? I think `scripts/` makes slightly more sense if we want to keep `unpackaged` for metadata only, but I don't have a strong preference here.

davidmreed:
@jstvz I prefer `scripts` since Anonymous Apex isn't going to be deployed to any org.

jstvz:
I've got some refactoring to do 😏

cumulusci.yml
```diff
@@ -359,6 +372,15 @@ flows:
...
                 task: ensure_record_type
+    
+    trial_org:
+        steps:
+            1:
+                flow: dependencies
+            2:
+                task: install_managed
+            3:
+                flow: config_trial
```
davisagli: I'm not clear on the use case for adding this flow. It's the same as `install_prod`, isn't it?

davidmreed: Yeah, it has less utility after admitting defeat on being able to activate CRLPs in code. I'll remove it for now but hope to bring it back later.

PR #4274, opened 2019-05-22
Add flow to configure BDI QA metadata in a namespaced scratch org

davidmreed:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davidmreed:
@skristem Can you validate that this branch fixes the gack you were observing related to help text mapping?

@davisagli I'd love to know if there's a better way to achieve this - for deployment into a namespaced org, we need to merge the org's namespace *only* on the Help Text mappings for BDI. As far as I could tell I can't use `NAMESPACED_ORG` tokens without also injecting other tokens we don't want injected in that context.

cumulusci.yml
```diff
@@ -256,7 +256,17 @@ flows:
...
+
+    qa_org_namespaced:
+        description: Execute the qa_org flow against a namespaced scratch org and patch help text.
+        steps:
+            1:
+                flow: qa_org
+            2:
+                task: deploy
+                options:
+                    path: unpackaged/config/qa_namespaced
```
davisagli: You shouldn't need a separate metadata folder for this. Put `%%%NAMESPACED_ORG%%%` in unpackaged/config/qa in the places where we need the npsp__ prefix for namespaced orgs but no prefix for unnamespaced orgs. Then use the `deploy_qa_config` task with the `namespaced_org: True` option set.

davisagli: @davidmreed oh, just saw your comment. What other tokens were getting injected?

davidmreed: @davisagli The existing `qa` metadata bundle has other tokens like, for example, `___NAMESPACE___Data_Import__c`. If those are injected with `npsp__`, they won't deploy to the namespaced org.  `zip_inject_namespace()` maps the `%%%NAMESPACED_ORG%%%` token to blank if `managed` is not `True`, but if `managed` is `True`, all other tokens also get mapped to the namespace.

Does that make sense, or am I missing something critical?

davisagli: @davidmreed that may be a bug in the namespace injection code. I should take a closer look; in the meantime I'm okay with this workaround.

skristem: Quick update, I am still getting the error while trying to create a bge batch. Will troubleshoot and update

unpackaged/config/qa/objects/___NAMESPACE___DataImport__c.object
```diff
@@ -1,11 +1,11 @@
...
     <fields>
-        <fullName>custom_campaign__c</fullName>
+        <fullName>new_lookup_campaign__c</fullName>
         <deleteConstraint>SetNull</deleteConstraint>
         <externalId>false</externalId>
         <inlineHelpText>Opportunity.custom_lookup__c</inlineHelpText>
-        <label>custom_campaign</label>
+        <label>new_lookup_campaign</label>
         <referenceTo>Campaign</referenceTo>
         <relationshipLabel>NPSP Data Imports (custom_campaign)</relationshipLabel>
```
skristem: @davidmreed does this line needs to be updated too?

davidmreed: No, the relationship label is cosmetic only, and the target field doesn't appear to cause the issue.

PR #4281, opened 2019-05-23
Feature/bdi co migration utility handle conflicts with defaults

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


psadie:
**lurch: attach W-032615

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LmZbQAK" target="_blank">W-032615</a>

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -36,65 +36,473 @@
...
-     * @description Update the custom data import settings field mapping method field
-     * @param fieldMappingMethod String
+     * @description Save the metadata deployment id in data import custom settings
+     * @param deploymentId Metadata deployment id
      */
-    private void updateCustomDataImportSettings(String jobId, String fieldMappingMethod) {
-        Map<String, String> statusMap = new Map<String, String>{jobId => 'Queue was empty'};
-        CMT_MetadataAPI.setHelpTextMigrationSettings(statusMap, true);
+    @future
+    public static void saveDeploymentId(String deploymentId) {
```
psadie: Move to Visualforce Page controller

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -36,65 +36,473 @@
...
+            List<Metadata.CustomMetadata> updatedCMTs = updateCustomMetadataRecords(
+                MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME,
+                conflictingMappings
+            );
+
+            queuedMetadataTypesForDeploy.addAll(updatedCMTs);
+        }
+
+        //TODO: handle GAUs and Opportunity Contact Roles
+        handleGAUAndOpportunityContactRoles(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
```
psadie: Don't pass in field mapping set name

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -36,65 +36,473 @@
...
+
+        if (!Test.isRunningTest()) {
+            deployCustomMetadata();
+        }
+    }
+
+    /**
+    * @description Method for handling special cased mappings (GAU and Opportunity Contact Role)
+    * because they aren't mapped in regular help text
```
psadie: Expand this description, why do we need to do this?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -36,65 +36,473 @@
...
+    public void handleGAUAndOpportunityContactRoles(String fieldMappingSet) {
+        System.debug(LoggingLevel.ERROR, '*** handleGAUAndOpportunityContactRoles');
+        // Collect our default GAU and Opportunity Contact Role metadata mappings
+        Map<String, Data_Import_Field_Mapping__mdt> specialCasedMappings =
+            new Map<String, Data_Import_Field_Mapping__mdt>();
+
+        Map<String, Data_Import_Field_Mapping__mdt> tempMap =
+            new Map<String, Data_Import_Field_Mapping__mdt>();
+
+        tempMap = helper.getMappingByFuzzyNameAndFieldSetName(
```
psadie: Create method to stop repeating process

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -164,61 +571,55 @@ public class BDI_MigrationMappingUtility {
...
+     * @return Metadata.DeployResult JSON
      */
-    private static String getUniqueJobId() {
-        return UserInfo.getUserId().left(15).right(4).toUpperCase() +
-                String.valueOf(Integer.valueOf(math.rint(math.random()*5000000))).right(6);
+    public String checkDeployStatus(String deploymentId) {
+        Http h = new Http();
+        HttpRequest request = new HttpRequest();
+        request.setEndpoint(Url.getOrgDomainUrl().toExternalForm() +
+            '/services/data/v44.0/metadata/deployRequest/' +
```
psadie: Hey @ptewson-sfdo - Is it alright to hardcode the api version in this endpoint?

src/classes/BDI_HelpTextCMT.cls
```diff
@@ -48,7 +48,40 @@ public class BDI_HelpTextCMT {
...
+
+        public Metadata.CustomMetadata getMetadataRecord() {
+            String mdTypeName = UTIL_Namespace.StrTokenNSPrefix(BDI_HelpTextCMT.MetadataObject.Data_Import_Field_Mapping_Set.Name() + '__mdt');
+
+            // Setup custom metadata to be created or updated in the subscriber org.
+            Metadata.CustomMetadata customMd = new Metadata.CustomMetadata();
+            customMd.label = this.label.left(40);
+            this.recordName = generateRecordName(this.label, false, '', '');
+            customMd.fullName = mdTypeName + '.' + this.recordName.left(40).removeEnd('_');
+            customMd.protected_x = IS_PROTECTED;
```
psadie: Don't protect custom field mapping set

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,81 +32,486 @@
...
+
+    /**
+    * @description Boolean flag to determine if a custom field mapping set exists 
+    * Set as static because it's used in the static method updateCustomSettings
+    */
+    public static Boolean customFieldMappingSetExists {
+        get {
+            List<Data_Import_Field_Mapping_Set__mdt> fieldMappingSets =
+                [SELECT Id
+                FROM Data_Import_Field_Mapping_Set__mdt
```
psadie: Fix indents

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,277 @@
...
+            // Org had no conflicting field mappings nor any net new field mappings in their Data Import help text
+            isPolling = false;
+            throwMessage(ApexPages.Severity.CONFIRM, Label.stgDeploymentCustomMetadataMappingEnabled);
+        }
+
+        if (timeoutCount > MAX_TIME_OUT) {
+            // We timed out, reset instance of migration mapping utility
+            migrationMappingUtility = new BDI_MigrationMappingUtility(new BDI_MigrationMappingHelper());
+            isPolling = false;
+            throwMessage(ApexPages.Severity.WARNING, Label.stgDeploymentTimedOut + ' ' + Label.stgDeploymentTryAgain);
```
kenlewis-sfdo: What do you think about making the `throwMessage` method static so another instance of the mapping utility does not need to be instantiated here?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,277 @@
...
+
+        return null;
+    }
+
+    /*******************************************************************************************************
+    * @description Checks to see if we have a saved custom metadata deployment id,
+    * inspects the deployment status, and sets the hasDeploymentInProgress boolean
+    * @return Metadata.DeployResult
+    */
+    private Metadata.DeployResult checkDeploymentStatus() {
```
kenlewis-sfdo: This method might be broken up into two methods, this one could be called 'getDeployResult', since it returns a `DeployResult' and the `hasDeploymentInProgress` member variable could be changed to an attribute that inspects the `deployResult` when called.  Not necessary, but that would allow each method/attribute to do one thing only.

If you did that, then, for example at the top of the `enableDataImportFieldMapping` method, instead of 
```
// Check the deployment status and set hasDeploymentInProgress value
        deploymentResult = checkDeploymentStatus();

        if (hasDeploymentInProgress == true) {
``` 
It could be:
```
// Get the status of the latest deployment
        setDeploymentResult();

        if (hasDeploymentInProgress == true) {
```

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,277 @@
...
+
+    /*******************************************************************************************************
+    * @description ActionPoller count
+    */
+    private Integer timeoutCount = 0;
+
+    /*******************************************************************************************************
+    * @description Get custom data import settings
+    */
+    public Data_Import_Settings__c dataImportSettings { get; set; }
```
kenlewis-sfdo: Can this member variable be made private?  Or does it need to be public?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,277 @@
...
+
+    /*******************************************************************************************************
+    * @description Custom metadata deployment result
+    */
+    public Metadata.DeployResult deploymentResult { get; set; }
+
+    /*******************************************************************************************************
+    * @description Boolean that determines if the toggle should be disabled
+    */
+    public Boolean hasDeploymentInProgress { get; set; }
```
kenlewis-sfdo: I think `deploymentResult ` and `hasDeploymentInProgress` should be made private also, since they are only used internally.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,81 +32,481 @@
...
+        return updatedCMTs;
+    }
+
+    /**
+    * @description Method for handling special cased mappings (GAU and Opportunity Contact Role).
+    * Since GAUs and Opportunity Contact Roles aren't mapped in helptext, we need to inspect and collect
+    * our existing packaged data import field mappings in order to clone them for the migrated custom
+    * field mapping set.
+    */
+    public void handleGAUAndOpportunityContactRoleMappings() {
```
ptewson-sfdo: I wanted to dive in to this a bit more deeply before the end of the day, but could we approach this more like "for any object mappings that don't exist in the help text mapping, bring over all the corresponding field mappings"? My concern here is that a) it's a bit hard-coded to OCR and GAU alloc and b) we might add new default mappings without help text mapping equivalent and it would be nice to bring those in during migration without having to adjust the code. 

Second thought, do we just want to copy in any default CMT field mappings that don't clash with their help text mappings (Including GAU and OCR) to create the new customized set? I don't want to surprise customers with changed config, but am I right in thinking that adding new field mappings should be reasonably safe?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
ptewson-sfdo: Don't they need to be public for visualforce page access?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
psadie: I think we can just toss out the line instantiating a new instance of migration util here. It's not actually useful anymore. Left over from before I was clearing the lists holding pending metadata records on every attempted migration.

src/classes/BDI_MigrationMappingUtility.cls
psadie: 100% makes total sense. I think it absolutely makes sense to bring over any and all default field mappings that don't clash with their help text mappings. Pretty much what I wanted to do, it just so happens in this case everything is mapped via help text _except_ GAUs and OCRs. I have a good idea of where I can update this 👍

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
kenlewis-sfdo: Yep, but `deploymentResult`, `migrationMappingUtility` and `dataImportSettings` are all only referenced in the controller and not from the page - so I think they can be private.

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
kenlewis-sfdo: 👍🏽

src/labels/CustomLabels.labels
```diff
@@ -2145,6 +2145,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>bdiDataImporterTitle</shortDescription>
         <value>NPSP Data Import</value>
     </labels>
+    <labels>
+        <fullName>bdiDeploymentInProgress</fullName>
+        <categories>Data Importer</categories>
+        <language>en_US</language>
+        <protected>false</protected>
+        <shortDescription>bdiDeploymentInProgress</shortDescription>
+        <value>There is a deployment in progess.</value>
```
melissabarber: @psadie this will show when a user tries to enable in a different tab when a deployment is already in progress, right?

If so, please update this message to read:
"Help Text mappings are currently being converted to Custom Metadata."

Is there anything else we should add to that message other than, "Please go put on your patience pants." :) But seriously, should we say something along the lines of, "This page will automatically reload when the conversion is complete."

melissabarber: Disregard the last paragraph. Error should read, "Help Text mappings are currently being converted to Custom Metadata."

src/labels/CustomLabels.labels
```diff
@@ -2233,6 +2241,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <shortDescription>error when updating data import custom settings</shortDescription>
+        <value>There was an error attempting to update data import custom settings.</value>
+    </labels>
+    <labels>
+        <fullName>bdiErrorDeploymentFailed</fullName>
+        <categories>Data Importer</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>error deploying metadata during migration from help text to metadata mapping</shortDescription>
+        <value>There was an error attempting to deploy custom metadata.</value>
```
melissabarber: Please update to "There was an error attempting to deploy Custom Metadata."

src/labels/CustomLabels.labels
```diff
@@ -2233,6 +2241,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>error when a batch is missing a name</shortDescription>
         <value>Batch Name is a required field.</value>
     </labels>
+    <labels>
+        <fullName>bdiErrorDataImportCustomSettings</fullName>
+        <categories>Data Importer</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>error when updating data import custom settings</shortDescription>
+        <value>There was an error attempting to update data import custom settings.</value>
```
melissabarber: Update to be, "There was an error attempting to update the Custom Settings named Data Import Settings."

When error is exposed to user, please also append the System Error at the end.

src/classes/BDI_MigrationMappingUtility.cls
ptewson-sfdo: I've made my peace with it! Although, maybe we can define the string as a class variable?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,277 @@
...
+        return null;
+    }
+
+    /*******************************************************************************************************
+    * @description Checks to see if we have a saved custom metadata deployment id,
+    * inspects the deployment status, and sets the hasDeploymentInProgress boolean
+    * @return Metadata.DeployResult
+    */
+    private Metadata.DeployResult checkDeploymentStatus() {
+        dataImportSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
```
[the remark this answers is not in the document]
psadie: I think we still need this here as the instance of data import settings initiated in the constructor may be stale at the time we need to poll for the latest deployment status

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,282 @@
...
+            throwMessage(ApexPages.Severity.WARNING, Label.stgDeploymentTimedOut + ' ' + Label.stgDeploymentTryAgain);
+            return null;
+        }
+
+        return null;
+    }
+
+    /*******************************************************************************************************
+    * @description Checks to see if we have a saved custom metadata deployment id,
+    * inspects the deployment status, and sets the hasDeploymentInProgress boolean
```
psadie: Update comment, no longer setting hasDeploymentInProgress

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,282 @@
...
+    * @description Wrapper for Metadata.DeployResult
+    */
+    private void throwMessage(ApexPages.Severity severity, String message) {
+        ApexPages.addMessage(new ApexPages.Message(severity, message));
+    }
+
+    /*******************************************************************************************************
+    * @description Wrapper for Metadata.DeployResult
+    */
+    public class DeployResultWrapper {
```
psadie: Remove wrapper, no longer used

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -584,7 +584,7 @@ public with sharing class BDI_MigrationMappingUtility {
      * @description Check the status of custom metadata deployment
      * @return Metadata.DeployResult JSON
      */
-    public String getDeploymentResult(String deploymentId) {
+    public Metadata.DeployResult getDeploymentResult(String deploymentId) {
```
psadie: Change to Id from String

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -96,17 +96,17 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     /*******************************************************************************************************
     * @description Custom metadata deployment result
     */
-    private Metadata.DeployResult deploymentResult { get; set; }
```
psadie: Keep full Metadata.DeployResult

src/classes/BDI_FieldMappingHelpText.cls
```diff
@@ -150,4 +155,183 @@ public with sharing class BDI_FieldMappingHelpText implements BDI_FieldMapping {
...
+
+        public void setInstance(Data_Import_Object_Mapping__mdt objectMapping,
+            String fieldMappingSetName) {
+
+            this.targetObjectMappingLabel = objectMapping.Label;
+            this.targetObjectMappingRecordName = objectMapping.DeveloperName;
+            this.dataImportFieldMappingSetRecordName = fieldMappingSetName;
+        }
+
+        public void setInstance(Data_Import_Object_Mapping__mdt objectMapping,
```
kenlewis-sfdo: These setInstance methods can be combined to one with a three param signature that handles null being passed in for the targetFieldAPIName.

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -82,18 +82,18 @@ public class BDI_MigrationMappingHelper {
     /*******************************************************************************************************
     * @description List of bad/malformed help text mappings
     */
-    public List<BDI_HelpTextCMT.HelpTextParts > badHelpTextFieldDescribeResultsByFieldName =
-        new List<BDI_HelpTextCMT.HelpTextParts >();
+    public List<BDI_MigrationMappingUtility.HelpTextFieldMapping > badHelpTextFieldDescribeResultsByFieldName =
+        new List<BDI_MigrationMappingUtility.HelpTextFieldMapping >();
```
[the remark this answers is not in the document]
kenlewis-sfdo: Also, variable is now a list and can be renamed - maybe `badHelpTextFieldDescribeResults`?

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -112,7 +112,7 @@ public class BDI_MigrationMappingHelper {
             Boolean helpTextIsUsable = String.isNotBlank(helpText) && !helpText.contains(' ');
 
             if (helpTextIsUsable) {
-                BDI_HelpTextCMT.HelpTextParts helpTextParts = new BDI_HelpTextCMT.HelpTextParts(sourceField);
+                BDI_MigrationMappingUtility.HelpTextFieldMapping helpTextParts = new BDI_MigrationMappingUtility.HelpTextFieldMapping(sourceField);
```
kenlewis-sfdo: How about `helpTextFieldMapping` rather than helpTextParts?

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -123,8 +123,8 @@ public class BDI_MigrationMappingHelper {
                 if (fieldNameIsUsable && objectNameIsUsable && objectAPINameIsUsable) {
```
kenlewis-sfdo: All of the `isUsable` checks can be done inside the `HelpTextFieldMapping` inner class.  So you should be able to instantiate the inner class by passing in the sourceField DFR after line 110.  Then you can call 
`if(helpTextFieldMapping.isUsable()` on line 114.  If it is usable, instantiate the `DataImportFieldMapping` instance.  If not, add the `HelpTextFieldMapping` to your list `badHelpTextFieldDescribeResults`.  Let me know what you think!

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -37,125 +37,131 @@
...
-
-        mappingDetail.importedRecordField = importStatusAndLookupParts.importLookupFieldName;
-        mappingDetail.importedRecordStatusField = importStatusAndLookupParts.importStatusFieldName;
+    /*******************************************************************************************************
+    * @description Method to get list of good help text mappings. Used in conjuction with
+    * populateListsFromHelpTextMappings()
+    *
+    * @return List of DataImportFieldMappings
+    */
+    public List<BDI_MigrationMappingUtility.DataImportFieldMapping> getDataImportFieldMappingsFromHelpText() {
```
kenlewis-sfdo: This should be housed in the Utility - and maybe renamed to `getDataImportFieldMappings` since the caller doesnt know what the source was necessarily, and to be consistent with the name of the `getBadHelpTextMappings` method below.

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -37,125 +37,131 @@
...
-        }
-
-        return importRelatedField;
+    /*******************************************************************************************************
+    * @description Method to get list of bad help text mappings. Used in conjuction with
+    * populateListsFromHelpTextMappings()
+    *
+    * @return List of HelpTextFieldMappings
+    */
+    public List<BDI_MigrationMappingUtility.HelpTextFieldMapping> getBadHelpTextMappings() {
```
kenlewis-sfdo: This should be in the Utility.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+            } else {
+                name = name.left(40);
+            }
+
+            return name;
+        }
     }
 
+    /*******************************************************************************************************
+    * @description Wrapper class for help text field
```
kenlewis-sfdo: How about "Wrapper class for field mapping information stored in the inline help text of a field"?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+    /*******************************************************************************************************
+    * @description Wrapper class for Data_Import_Field_Mapping__mdt used to create
+    * the default CMT data as well as for use in the Custom UI.
+    */
+    public class DataImportFieldMapping {
+        public String recordName;
+        public String label {get; set;}
+        public String dataImportObjectName {get; set;}
+        public String dataImportFieldMappingSetRecordName;
+        public String dataImportFieldMappingSetLabel;
```
kenlewis-sfdo: This variable appears to be unused. Can it be removed?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+        public String sourceFieldAPIName;
+        public String targetFieldAPIName {get; set;}
+        public String targetObjectMappingRecordName;
+        public String targetObjectMappingLabel;
+        public String targetObjectAPIName;
+        public String isRequired;
+        public Boolean isDeleted;
+        public String importedRecordField;
```
kenlewis-sfdo: This variable appears to be unused. Can it be removed?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+        public String importedRecordStatusField;
```
kenlewis-sfdo: This variable appears to be unused. Can it be removed?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+
+        public DataImportFieldMapping(String label, String sourceFieldAPIName, String dataImportObjectName,
```
kenlewis-sfdo: This is the signature that can be updated to accept two objects as opposed to these 5 values.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+            this.label = label.trim();
+            this.sourceFieldAPIName = sourceFieldAPIName;
+            this.dataImportObjectName = dataImportObjectName;
+            this.targetFieldAPIName = targetFieldAPIName;
+            this.targetObjectAPIName = targetObjectAPIName;
+        }
 
-    /** @description Name of default packaged Field Mapping Set */
-    private static final String DEFAULT_FIELD_MAPPING_SET_RECORD_NAME = 'Default_Field_Mapping_Set';
+        public DataImportFieldMapping(Data_Import_Field_Mapping__mdt sourceFieldMapping, String fieldMappingSet) {
```
kenlewis-sfdo: This signature is only used twice, each time passing in the migrated mapping set name.  Will it ever be called passing in a different value? If not, can it be simplified?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+            }
+        }
 
-    /**
-     * @return A 10 character unique 'jobId' to use for
-     */
-    private static String getUniqueJobId() {
-        return UserInfo.getUserId().left(15).right(4).toUpperCase() +
-                String.valueOf(Integer.valueOf(math.rint(math.random()*5000000))).right(6);
+        public void setRecordName(String recordName, Boolean isUnique) {
```
kenlewis-sfdo: This should be broken up into two methods: `setRecordName` and `getUniqueRecordName` .  `setRecordName` should set the value without doing any internal manipulation to the string.  If a unique record name is needed, then it should be generated separately beforehand and passed to `setRecordName`.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+            if (isUnique) {
+                this.recordName = generateRecordName(this.label, true, this.targetObjectMappingLabel, this.dataImportObjectName);
```
kenlewis-sfdo: `generateRecordName` should have an empty signature, since it is only operating on its instance member variable's values.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,614 @@
...
+                this.recordName = recordName;
+            }
+        }
+
+        public Metadata.CustomMetadata getMetadataRecord() {
+            String mdTypeName = 'Data_Import_Field_Mapping__mdt';
+
+            // Setup custom metadata to be created or updated in the subscriber org.
+            Metadata.CustomMetadata customMd = new Metadata.CustomMetadata();
+            if (this.targetObjectMappingLabel != null) {
```
kenlewis-sfdo: `this.` should be used when assigning values to the inner classes' local member variables inside of the constructor only.  When not assigning values inside the constructor, `this.` can be removed.  This can be applied to the other inner classes also.

src/classes/CMT_MetadataAPI.cls
```diff
@@ -170,68 +192,29 @@ public class CMT_MetadataAPI {
...
-
-        /**
-        * @description Update Customizable Rollups Settings
-        */
-        private void setCustomizableRollupsSettings(Map<String, String> statusMap, Boolean isSuccess) {
             // Write the deployment status to the custom settings object
             Customizable_Rollup_Settings__c crlpSettings = UTIL_CustomSettingsFacade.getCustomizableRollupSettings();
             crlpSettings.CMT_API_Status__c = JSON.serializePretty(statusMap).left(255);
-
+            // THIS WILL NEED TO BE REFACTORED IF THIS CMT DEPLOYMENT CODE IS USED BY SOMETHING OTHER THAN ROLLUPS
```
kenlewis-sfdo: Can this comment be removed? Or is it still needed?

src/classes/CMT_MetadataAPI.cls
```diff
@@ -170,68 +192,29 @@ public class CMT_MetadataAPI {
...
             crlpSettings.Customizable_Rollups_Enabled__c = (isSuccess || crlpSettings.Customizable_Rollups_Enabled__c == true);
             if (!Test.isRunningTest()) {
                 upsert crlpSettings;
             }
+            if (isSuccess) {
```
kenlewis-sfdo: Is this change meant to be included in this PR?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,279 @@
...
+                Label.bdiErrorDataImportCustomSettings + ' ' + e.getMessage()));
+        }
+    }
+
+    /*******************************************************************************************************
+    * @description Action Method used by Poller to check status of CMDT deployment
+    * @return void
+    */
+    public PageReference startPolling() {
+        System.debug('***startPolling()');
```
kenlewis-sfdo: Can this debug be removed?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,279 @@
...
+    /*******************************************************************************************************
+    * @description Checks to see if we have a saved custom metadata deployment id,
+    * inspects the deployment status, and sets the hasDeploymentInProgress boolean
+    */
+    private void refreshCustomSettings() {
+        dataImportSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
+    }
+
+    /*******************************************************************************************************
+    * @description Wrapper for Metadata.DeployResult
```
kenlewis-sfdo: Looks like this comment needs to be updated.

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -0,0 +1,279 @@
...
+    */
+    private void throwMessage(ApexPages.Severity severity, String message) {
```
kenlewis-sfdo: How about `addPageMessage` rather than throwMessage since it is not actually throwing an exception?

src/classes/CMT_MetadataAPI.cls
psadie: This comment was left by the original dev and I think leaving it might be a good idea for the next person to hop into this class. I believe we might have have incoming changes from the Abacus folks as well

src/classes/CMT_MetadataAPI.cls
psadie: Hmm I'm pretty sure this mirrors what we have in master currently. Does git think this is a new addition because I have previous commits that moved things around in here?

src/classes/CMT_MetadataAPI.cls
kenlewis-sfdo: 👍🏽

src/classes/CMT_MetadataAPI.cls
kenlewis-sfdo: Possibly - it looks unchanged to me, other than being pushed down in the file because of your additions above this line.  But strange that it is included in your PR.  Maybe try checking the file out from Master and then re-applying your changes to this file on top and see if this goes away?

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -37,125 +37,113 @@
...
+        Map<String, Schema.DescribeFieldResult> describeFieldResultsByFieldName =
+            UTIL_Describe.getAllFieldsDescribe(UTIL_Namespace.StrAllNSPrefix(
+                BDI_MigrationMappingUtility.DATA_IMPORT_OBJECT_NAME));
 
-        for (String dataImportFieldName : allDFRs.keySet()) {
+        for (String dataImportFieldName : describeFieldResultsByFieldName.keySet()) {
 
-            // check if field is importstatus or imported
             Boolean isImportStatus = dataImportFieldName.contains('importstatus');
             Boolean isImportLookup = dataImportFieldName.contains('imported');
```
kenlewis-sfdo: The code should probably check against a specific list of fields here since customers will be able to map from their own newly created custom fields on DataImport__c and could potentially give their new fields names that include these strings...

src/classes/BDI_MigrationMappingHelper.cls
```diff
@@ -37,125 +37,34 @@
...
-
-        BDI_HelpTextCMT.ImportStatusAndLookupParts importStatusAndLookupParts =
-                importStatusAndLookupMap.get(mappingDetail.dataImportObjectName);
-
-        mappingDetail.importedRecordField = importStatusAndLookupParts.importLookupFieldName;
-        mappingDetail.importedRecordStatusField = importStatusAndLookupParts.importStatusFieldName;
+    /*******************************************************************************************************
+    * @description Default packaged Object Mapping Set metadata record
+    */
+    public Data_Import_Object_Mapping_Set__mdt getDefaultObjectMappingSet(String developerName) {
```
kenlewis-sfdo: Is this method still used now?  If not maybe we can remove it.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,668 @@
...
+        if (customFieldMappingSetExists) {
+            fieldMappingSetName = MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME;
+        }
+
+        Map<String, Data_Import_Field_Mapping__mdt> fieldMappingsByFieldMappingSetDeveloperName =
+            helper.getFieldMappingsByFieldMappingSetDeveloperName(fieldMappingSetName);
+
+        for (DataImportFieldMapping fieldMapping : dataImportFieldMappings) {
+
+            String fieldMappingKey = fieldMapping.targetObjectAPIName +
```
kenlewis-sfdo: This should be an internal method: `fieldMapping.getKey()`

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,668 @@
...
+                if (customFieldMappingSetExists) {
+                    // Migrated Custom Field Mapping Set exists
+                    DataImportFieldMapping overwrittenMapping =
+                        new DataImportFieldMapping(
+                            existingDefaultFieldMapping,
+                            MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
+
+                    overwrittenMapping.setFieldMappingSet(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
+                    overwrittenMapping.setObjectMapping(existingDefaultObjectMapping);
+                    overwrittenMapping.setTargetField(fieldMapping.targetFieldAPIName);
```
kenlewis-sfdo: 👍🏽👍🏽👍🏽

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,668 @@
...
+                } else {
+                    // Check if we have a conflict and update the target field based on the help text mapping
+                    if (fieldMapping.hasBadDefaultValues(existingDefaultFieldMapping)) {
+                        fieldMapping.setFieldMappingSet(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
+                        fieldMapping.setObjectMapping(existingDefaultObjectMapping);
+                        fieldMapping.setTargetField(fieldMapping.targetFieldAPIName);
+                    } else {
+                        fieldMapping.setFieldMappingSet(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
+                        fieldMapping.setObjectMapping(existingDefaultObjectMapping);
+                    }
```
kenlewis-sfdo: `fieldMapping.setFieldMappingSet(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
fieldMapping.setObjectMapping(existingDefaultObjectMapping);`
Can be moved above the if block and only performed once since they are doing the same thing in each block.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,668 @@
...
 
-        Map<String, Data_Import_Object_Mapping__mdt> existingObjectMappings =
-                helper.getDefaultObjectMappings(DEFAULT_OBJECT_MAPPING_SET.Id);
+        // TODO: Need to move this into the helper so we can mock the result
+        // Replace stubbedMethod getDataImportFieldMappingsFromHelpText
+        // When this method was in the helper, we had another get method
+        // used by this utility class on line 238
+        // i.e. List<DataImportFieldMapping> helpTextFieldMappings = helper.getDataImportFieldMappingsFromHelpText();
+        Map<String, Schema.DescribeFieldResult> describeFieldResultsByFieldName =
+            UTIL_Describe.getAllFieldsDescribe(UTIL_Namespace.StrAllNSPrefix(DATA_IMPORT_OBJECT_NAME));
```
kenlewis-sfdo: Yep, this could be `helper.getAllFieldsDescribe(DATA_IMPORT_OBJECT_NAME)`

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -32,193 +32,668 @@
...
-    /**
-     * @description Collect all Help Text field mappings and compare against default Field Mapping Metadata 
-     * Type records reserve all non-standard NPSP Help Text mappings. What to do with reserved list of Help 
-     * Text mappings TBD.
+    /*******************************************************************************************************
+     * @description Go through all fields in Data Import object and populate relevant lists from help text
+     * mappings
      */
-    public void buildCustomMetadataRecordsFromHelpText(BDI_MigrationMappingHelper helper) {
+    public void populateListsFromHelpTextMappings() {
```
kenlewis-sfdo: This method should be private since it is only used in the Utility class, with an @TestVisible tag so that it can be called from the tests.

src/classes/BDI_MigrationMappingHelper.cls
psadie: Nope! Removed

src/classes/BDI_MigrationMappingUtility.cls
psadie: @kenlewis-sfdo - I was trying this out yesterday and I don't think this would actually be useful. Since that would need to return a Map<String, Schema.DescribeFieldResult>. Am I wrong in that there isn't a way for us to mock or otherwise construct instances of Schema.DescribeFieldResults?

PR #4293, opened 2019-05-30
Feature/robot  payments writeoff

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/NPSP.py
```diff
@@ -126,6 +126,17 @@ def click_special_related_list_button(self, heading, button_title):
         )
         self.selenium.click_link(locator)
         
+    def click_related_list_dd_button(self, heading, dd_title, button_title):
+        """ To Click on a related list dropdown button.
+            Pass the list name, dd name and button name"""
```
Safiyab: Wondering what "dd" means. I guess anyone working on NPSP would know. But if you think elaborating that is good, please go for it.

skristem: I didn't want to make the name too long so I the comment I mentioned dropdown

Safiyab: Ohhh !! Its dropdown ! Got it. Sorry after 6 pm brain does that to me :)

PR #4296, opened 2019-05-30
Remove Opportunity related list blocking prerelease builds

davidmreed:
This PR removes the Opportunities related list from the NPSP's Campaign page layout. Prerelease orgs include the Customizable Campaign Influence feature, which is incompatible with this related list and blocks NPSP deployments. At present, we are not aware of a programmatic route to deactivate this feature.

As a consequence of this PR, if accepted, manual NPSP installs through MetaDeploy would require an additional manual setup step to configure this related list if it is desired and if Customizable Campaign Influence is not in use.

Merging this PR would allow us to resume performing prerelease builds and tests of NPSP as we approach the release of Summer '19. It would also ensure that NPSP installs are compatible with orgs created in Summer '19 that may have this feature active. We do not know at this time exactly which newly created orgs will be so configured.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/layouts/Campaign-NPSP Campaign Layout.layout
+++ b/src/layouts/Campaign-NPSP Campaign Layout.layout
@@ -194,13 +194,6 @@
         <relatedList>Allocation__c.Campaign__c</relatedList>
     </relatedLists>
     <relatedLists>
-        <fields>OPPORTUNITY.NAME</fields>
-        <fields>OPPORTUNITY.STAGE_NAME</fields>
-        <fields>OPPORTUNITY.AMOUNT</fields>
-        <fields>OPPORTUNITY.CLOSE_DATE</fields>
-        <relatedList>RelatedOpportunityList</relatedList>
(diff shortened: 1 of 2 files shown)


davidmreed:
cc @jlantz.

We've raised this issue with Core in W-6187295.

davidmreed:
@jesslopez-sf The `qa_org` flow installs the `trial` unpackaged metadata, which includes a reference to the same page layout. I just pushed a new commit to remove that one as well; there are no other remaining references.

davisagli:
@force2b I hear you, but keep in mind that unless we make this change, _no one_ will be able to install NPSP in new orgs created on the Summer '19 release (which reaches R1 on June 7) unless they first manually disable Customizable Campaign Influence. Also, unless we make this change, we are currently blocked from running our automation to make sure that Summer '19 doesn't cause other issues.

PR #4297, opened 2019-05-30
Added Opportunity Stage to State Mapping custom metadata type and updated the code to utilize the new CMT

KietTruongSFDO:
# Critical Changes

# Changes
- We introduced a new Custom Metadata Type called Opportunity Stage to State Mapping. This feature is intended for use in a future release and should have no visible impact on your org.
# Issues Closed

# New Metadata
Custom Metadata Type: "Opportunity_Stage_to_State_Mapping__mdt"
Layout: "Opportunity  Stage to State Mapping Layout"

# Deleted Metadata


src/classes/OPP_StageMappingUtil.cls
```diff
@@ -93,8 +96,118 @@ public with sharing class OPP_StageMappingUtil {
...
+        // Only create the finalized opp stage to state record if there's no existing finalized record. 
+        Integer finalizedCount = [Select Count() From Opportunity_Stage_To_State_Mapping__mdt Where Opportunity_State__c =: FINALIZED_STATE_VAL];
+        
+        // Take the value, if there is one, and create a finalized Opp Stage to State custom metadata type record. 
+        if (finalizedCount == 0 && 
+            (contactAndOrgCustomSetting.Payments_Auto_Close_Stage_Name__c != null && 
+            contactAndOrgCustomSetting.Payments_Auto_Close_Stage_Name__c != '')) {
+            
+            String mdTypeName = UTIL_Namespace.StrTokenNSPrefix(OPP_STAGE_TO_STATE_METADATA_NAME);
+            String recordName = FINALIZED_STATE_VAL;
```
jbrandolinisf: Recommend Making this FINALIZED_STATE_VAL + '-' + <<Opportunity StageName>>, or 
FINALIZED_STATE_VAL + '-' + now.getTime() + '-' + <<RandomVal>>)

Probably the second one, because we will need to prevent dupes.  Lastly, I would do two things: 

1) Keep this method, but have it call a general method passing a Map of Stage to State Names
2)The general method should make sure it deletes any other records for that Opportunity Stage

Keep in mind you will need additional (method(s)) that are more general when you do the next work item to let them edit these via the UI.

jbrandolinisf: I would recommend a clear all, a delete by Stage, and an addAllIn as methods

src/classes/STG_InstallScript.cls
```diff
@@ -63,6 +64,8 @@ global without sharing class STG_InstallScript implements InstallHandler {
             
             executeAction(InstallAction.RUN_NEW_ORG_SCRIPT, logger);
 
+            // Create the Opportunity Stage to State mapping record for finalized
+            executeAction(InstallAction.CREATE_FINALIZED_OPP_STAGE_TO_STATE_RECORD, logger);
         // Upgrade of existing Cumulus installation
```
jbrandolinisf: If there are no states mapped at all, it should map Committed to Pledged, if a Pledged Stage exists.
If there is no finalized states mapped and the Opp Stage for When Fully Paid has a value, it should map that

If the criteria are not met, it should not attempt to execute these.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -93,8 +96,118 @@ public with sharing class OPP_StageMappingUtil {
...
         // This will be updated in March 2019 timeframe with revised code to retrieve from Custom Metadata instead
-        COMMITTED_STATE_SET = new Set<String> {'closed won'};
-        FINALIZED_STATE_SET = new Set<String> {'finalized'};
+        // COMMITTED_STATE_SET = new Set<String> {'closed won'};
+        // FINALIZED_STATE_SET = new Set<String> {'finalized'};
+
+        // If either the committed state set or the finalized state set is null, query for everything and 
+        // reinitialize everything. 
+        if (COMMITTED_STATE_SET == null || FINALIZED_STATE_SET == null) {
+            COMMITTED_STATE_SET = new Set<String>();
```
jbrandolinisf: If you switch to the approach I outlined, you can eliminate initialization here for the two sets.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -93,8 +96,118 @@ public with sharing class OPP_StageMappingUtil {
...
+
+            List<Opportunity_Stage_To_State_Mapping__mdt> stageToStateRecords = [Select Id, Opportunity_Stage__c, Opportunity_State__c FROM Opportunity_Stage_To_State_Mapping__mdt];
+
+            for (Opportunity_Stage_To_State_Mapping__mdt mappingRecord : stageToStateRecords) {
+                String oppStage = mappingRecord.Opportunity_Stage__c.trim().toLowerCase();
+
+                if (mappingRecord.Opportunity_State__c.trim().toLowerCase() == 'committed') {
+                    COMMITTED_STATE_SET.add(oppStage);
+                } else {
+                    FINALIZED_STATE_SET.add(oppStage);
```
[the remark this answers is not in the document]
jbrandolinisf: Technically, you could just use the map, and never worry about the sets, but you would have to modify the check to handle when no values are present, because the set would not be populated then.  Your call, though.   I am fine either way.  

The main reason I like to do this kind of thing this way, is that it reduces moving parts (meaning conditional branching), which is harder to write tests for.

robot/Cumulus/resources/NPSP.py
```diff
@@ -126,6 +126,17 @@ def click_special_related_list_button(self, heading, button_title):
         )
         self.selenium.click_link(locator)
         
+    def click_related_list_dd_button(self, heading, dd_title, button_title):
```
jbrandolinisf: Is this stuff you added, or are the two branches out of sync with each other?

robot/Cumulus/resources/locators_45.py
```diff
@@ -26,6 +26,7 @@
             'item':"//article[contains(@class, 'forceRelatedListCardDesktop')][.//img][.//span[@title='{}']]//h3//a",
             'field_value': '//a[text()="{}"]/ancestor::li//div[contains(@class, "slds-item--detail")]//*[text()="{}"]',
             'link':"//article[contains(@class, 'forceRelatedListCardDesktop')][.//img][.//span[@title='{}']]//table[contains(@class,'forceRecordLayout')]/tbody/tr[.//th/div/a[contains(@class,'textUnderline')]][.//td//span[text()='{}']]/th//a",
+            'dd-link':'//div[contains(@class,"actionMenu")]//a[@title="{}"]',
          },
```
jbrandolinisf: Is this stuff you added, or are the two branches out of sync with each other?

robot/Cumulus/tests/browser/donations_payments/create_donation_and_payment.robot
```diff
@@ -76,4 +76,28 @@ Verify values in Create one or more Payments for this Opportunity page
...
+    ...    Remaining Balance=$550.00  
+    Page Should Contain    You are preparing to write off 2 Payment(s) totaling $550.00
+    Choose Frame    Write Off Remaining Balance
+    Click Button    Cancel
+    Select Tab    Related
+    Load Related List    Payments
+    Click ViewAll Related List    Payments
+    Verify Details
+    ...    Unpaid=2
+    ...    Written Off=2
```
jbrandolinisf: Is this stuff you added, or are the two branches out of sync with each other?

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,180 @@ public with sharing class OPP_StageMappingUtil {
...
+        // Only create the finalized opp stage to state record if there's no existing finalized record. 
+        Integer finalizedCount = [Select Count() From Opportunity_Stage_To_State_Mapping__mdt Where Opportunity_State__c =: FINALIZED_STATE_VAL];
+        
+        // Take the value, if there is one, and create a finalized Opp Stage to State custom metadata type record. 
+        if (finalizedCount == 0 && 
+            (contactAndOrgCustomSetting.Payments_Auto_Close_Stage_Name__c != null && 
+            contactAndOrgCustomSetting.Payments_Auto_Close_Stage_Name__c != '')) {
+            
+            String mdTypeName = UTIL_Namespace.StrTokenNSPrefix(OPP_STAGE_TO_STATE_METADATA_NAME);
+            String recordName = mdTypeName + '_' + FINALIZED_STATE_VAL + '_' + datetime.now().format('yyMMddHHmmss');
```
jbrandolinisf: Take mdTypeName out.  It should not be part of the Record Name.  Is there are reason you are not using DateTime.now().getTime() ?

As stated before, I would prefer a factory method to create these, taking stage and state as it's only parameter.

This section would then be:  

medataToInsert.add(createStageToStateMapping (contactAndOrgCustomSetting.Payments_Auto_Close_Stage_Name__c, FINALIZED_STATE_VAL));

Everything, including the fieldMapping, recordName, etc and setup of it, should be inside the createStageToStateMapping Method.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,180 @@ public with sharing class OPP_StageMappingUtil {
...
+        }
+
+
+        /**
+        * @description Inserts an error log into the Error table
+        * @param ex Exception object that is to be logged
+        */
+        private void logError(Exception ex) {
+            Error__c rollupHandlerError = new Error__c(
+                Error_Type__c = 'Opportunity Stage to State Metadata Handler Error',
```
jbrandolinisf: If anyone will ever see this, it will need to be added to the Catalogue section of the Work Item, and converted to a Label.

src/classes/STG_InstallScript.cls
```diff
@@ -369,6 +377,21 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+    * @description Creates the finalized opp stage to state record based on what value is set as the 
+    * "Opportunity Stage for When Fully Paid" 
+    */
+    private void createOppStageToStateRecordsIfNecessary() {
+
+        List<Metadata.CustomMetadata> metadataToInsert = new List<Metadata.CustomMetadata>();
+        Map<String, Object> params = new Map<String, Object>();
+
+        metadataToInsert = Opp_StageMappingUtil.createFinalizedOppStageToStateRecord(metadataToInsert);
+        metadataToInsert = Opp_StageMappingUtil.createPledgedOppStageToStateRecord(metadataToInsert);
```
jbrandolinisf: Given the requirements, you will probably need to swap these.  Also, ideally, it should only add the committed mapping if the Pledged Stage exists in the OpportunityStage table.

KietTruongSFDO: I'll switch the order of these two. The check for the pledge stage is done inside of createPledgedOppStageToStateRecord.

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: I can change it so that the createStageToStateMapping method just takes in the stage and state. I'm using dateTime.now().format() because the api name can only have alphanumerics and underscore, dateTime.now() has dashes.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,180 @@ public with sharing class OPP_StageMappingUtil {
...
+        
+        Integer overallCount = [Select Count() From Opportunity_Stage_To_State_Mapping__mdt];
+
+        // Get all opportunity stages to see if there is a pledged stage
+        Schema.DescribeFieldResult fieldResult = Opportunity.StageName.getDescribe();
+        List<Schema.PicklistEntry> picklistEntryList = fieldResult.getPicklistValues();
+        Boolean hasPledgedValue = false;
+
+        for (Schema.PicklistEntry picklistEntry : picklistEntryList) {
+            if (picklistEntry.getValue().toUpperCase() == PLEDGED_STAGE) {
```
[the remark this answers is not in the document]
KietTruongSFDO: There's a check above this that is querying for all records of the mappings and if there isn't any mappings then it goes in and creates the pledge/committed mapping. Do we create the pledge/committed only when there's no other committed or finalized mapping instead? If so, I can update the query to reflect that.

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: Turns out we need the mdTypeName, its for creating the actual metadata record. This is an example of the one in GEM: 

"GEM_API_Settings.GEM_Config.md"

jbrandolinisf: dateTime.now().getTime() is the number of milliseconds since 1970.

src/classes/STG_InstallScript.cls
jbrandolinisf: And again, it should not check for Pledge Stage.  Just whether the "Pledged Stage" exists, and whether there are any metadata records yet.

src/classes/OPP_StageMappingUtil.cls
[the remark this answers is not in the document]
jbrandolinisf: Right.  Only if there are no mappings period.

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: Oh sorry, I misread it. I thought you had meant just dateTime.now() instead of dateTIme.now().getTime(). I'll change it to getTime().

jbrandolinisf: Thanks!  And thanks for working through this stuff with me!!!

src/classes/CMT_MetadataAPI.cls
```diff
@@ -177,7 +185,30 @@ public class CMT_MetadataAPI {
 
             if (isSuccess) {
                 if (!Test.isRunningTest()) {
-                    UTIL_OrgTelemetry_SVC.asyncProcessNpspSettingsTelemetry();
+                    Set<String> metadataObjectTypeSet = new Set<String>();
```
KietTruongSFDO: Hey @patrick-yan-sf This is the start of the changes we discussed.

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: As of this moment, it should never be shown. Should we preemptively convert it into a label?

src/classes/CMT_MetadataAPI.cls
patrick-yan-sf: @KietTruongSFDO - yup that's the general idea.  Here are some thoughts.

Do you think the metadataFullName.split call might have issues with namespaces?  I can't recall if the fullname will be npsp.opp_state_to_stage_mapping__mdt or just npsp__opp_state_to_stage_mapping__mdt.

I'm not sure about passing in the telemetry code version of the names.  That would couple this logic with that class.  Can we just pass in the full name and have the UTIL_OrgTelemetry_SVC class do that parsing?

Also let's move this logic into a separate function.

Something like 
Set<String> buildChangedMetada(DeployResult results) 

Then the call can just be Util_OrgTelemetry_SVC.asyncProcessCMTChange(buildChangedMetadata(results));

src/classes/OPP_StageMappingUtil.cls
jbrandolinisf: Lets convert it.  One less thing for later.

KietTruongSFDO: Actually, this is for the "type" of error, does it really need to be a label? Also, will users typically look at the error log or is that primarily for internal usage?

.flake8
```diff
@@ -0,0 +1,6 @@
+# Black-compatible flake8 config
```
force2b: @KietTruongSFDO I think this file was accidentally committed. If so, would you mind removing from the repo?

.gitignore
```diff
@@ -31,6 +31,7 @@ Referenced Packages
 **/.sfdx/
 **/.vscode/
 **/.idea/
+**/.mypy_cache/
```
force2b: @KietTruongSFDO I think this is also something that is unique to your dev environment. If you need it to be in _your_ gitignore, you can add it to your global gitignore file.

.flake8
KietTruongSFDO: I didn't add this file, it might just be the branches are out of sync. I'll pull and re-update this PR.

.gitignore
KietTruongSFDO: Same as above, I didn't make this change and I think the branches are out of sync. I'll pull and update again.

force2b:
@KietTruongSFDO and @jbrandolinisf I was taking a look at this PR just because it hits the CRLP Metadata API classes, but I noticed some files in there that seem completely unrelated to what this is supposed to be. There's a CCI task and some related files for loading BDI data, some other file, a change to gitignore. Is there a possibility that this branch has some extraneous stuff in it that should not be here? Maybe a manual merge was done off an incorrect branch (I think that would explain what I'm seeing). Also, can the PR name be updated to reflect the functionality instead of just the branch name?

KietTruongSFDO:
> @KietTruongSFDO and @jbrandolinisf I was taking a look at this PR just because it hits the CRLP Metadata API classes, but I noticed some files in there that seem completely unrelated to what this is supposed to be. There's a CCI task and some related files for loading BDI data, some other file, a change to gitignore. Is there a possibility that this branch has some extraneous stuff in it that should not be here? Maybe a manual merge was done off an incorrect branch (I think that would explain what I'm seeing). Also, can the PR name be updated to reflect the functionality instead of just the branch name?

@force2b I've updated the PR name to be more descriptive. In regards to the extra files, I think it's because another branch got merged in and I hadn't updated this PR yet. I just did a pull and updated this branch and those files are no longer in the PR.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -108,6 +108,28 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
     }
 
+    /**
+     * @description Determines if we should execute the telemetry code based on if the created custom metdata types
+     * are NPSP metadata type objects. 
+     * @param customMetadataTypeSet
+     */
+    @Future
+    public static void asyncProcessCMTChange(Set<String> customMetadataTypeSet) {
+
```
KietTruongSFDO: @patrick-yan-sf  I Just updated the code, is this more in line with with you were thinking? I no longer do a split on a period, instead I just get the whole name and use the enum values to see if it's contained.

patrick-yan-sf: @KietTruongSFDO - Thanks.  Yes that's what I was thinking of.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,198 @@ public with sharing class OPP_StageMappingUtil {
...
+            finalizedRecord.Opportunity_State__c = FINALIZED_STATE_VAL;
+
+            Opportunity_Stage_To_State_Mapping__mdt pledgedRecord = new Opportunity_Stage_To_State_Mapping__mdt();
+            pledgedRecord.Opportunity_Stage__c = 'Pledged';
+            pledgedRecord.Opportunity_State__c = COMMITTED_STATE_VAL;
+
+            metadataRecordList.add(finalizedRecord);
+            metadataRecordList.add(pledgedRecord);
+        } else {
+            metadataRecordList = [Select Id, Opportunity_Stage__c, Opportunity_State__c FROM Opportunity_Stage_To_State_Mapping__mdt];
```
force2b: Can this be an upper case `SELECT`?

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,198 @@ public with sharing class OPP_StageMappingUtil {
...
+
+    /*******************************************************************************************************
+    * @description Creates the finalized state record if there are no finalized records. It utilizes the 
+    * "Opportunity Stage when Fully Paid" field as the Opportunity stage for the finalized state. 
+    ********************************************************************************************************/
+    public static List<Metadata.CustomMetadata> createFinalizedOppStageToStateRecord(List<Metadata.CustomMetadata> metadataToInsert) {
+        npe01__Contacts_And_Orgs_Settings__c contactAndOrgCustomSetting = UTIL_CustomSEttingsFacade.getOrgContactsSettings();
+
+        // Only create the finalized opp stage to state record if there's no existing finalized record. 
+        Integer finalizedCount = [Select Count() From Opportunity_Stage_To_State_Mapping__mdt Where Opportunity_State__c =: FINALIZED_STATE_VAL];
```
force2b: Same here. Upper case `SELECT`.

force2b: I've also noticed that SOQL against the CMT object is slower than with SObjects. Definitely something to keep in mind if you're doing any queries against this in a loop.

jbrandolinisf:
Not in a loop, and it is why we are initializing sets from the metatada for
calls to the API.
--
John Brandolini
Phone: 781-460-3479

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +95,198 @@ public with sharing class OPP_StageMappingUtil {
...
+        }
+
+        return metadataRecordList;
+    }
+
+    /*******************************************************************************************************
+    * @description Creates the finalized state record if there are no finalized records. It utilizes the 
+    * "Opportunity Stage when Fully Paid" field as the Opportunity stage for the finalized state. 
+    ********************************************************************************************************/
+    public static List<Metadata.CustomMetadata> createFinalizedOppStageToStateRecord(List<Metadata.CustomMetadata> metadataToInsert) {
```
force2b: @jbrandolinisf The description for this class is `API Class to allow caller to determine Opportunity's State (Uncommitted, Committed. or Finalized) using the Opportunities Stage`. Given the single responsibility for this class, would it make sense to move all of the methods related to creating and deploying CMT's for this custom object into a different class such as `OPP_StageMappingCMT` or `OPP_StageMappingConfigBuilder`?

src/classes/STG_InstallScript.cls
```diff
@@ -369,6 +377,21 @@ global without sharing class STG_InstallScript implements InstallHandler {
         }
     }
 
+    /*******************************************************************************************************
+    * @description Creates the finalized opp stage to state record based on what value is set as the 
+    * "Opportunity Stage for When Fully Paid" 
+    */
+    private void createOppStageToStateRecordsIfNecessary() {
```
force2b: We've never tested creating new CMT records as part of the install script process. This will need to be tested in a beta package to ensure that it works as expected. There's always the possibility that this method needs to be handled as a Queueable executed from the PostInstallScript.

jbrandolinisf: @force2b - Would you prefer we start out doing this as a Queuable, rather than the current approach, or are you ok with trying it first.  Agree on how we will have to test it, unfortunately.

src/classes/OPP_StageMappingUtil.cls
jbrandolinisf: @force2b - Yep, Kiet and I talked about it.  Originally I was thinking we would add other support to this class (to add/remove stage to state mappings, but I am ok with moving this to another class, and @KietTruongSFDO is doing this now.

src/classes/OPP_StageMappingUtil.cls
jbrandolinisf: We cache it in the Stage to State API for that reason.  So only the first call is a cost, but it is a great callout.

src/classes/STG_InstallScript.cls
force2b: You can try it first, though it might be because MixedDML rules apply with calls to the FeatureManagement namespace I suspect that it will need to be handled asynchronously. So might be easier to just start out as async.

KietTruongSFDO: Hi @force2b We're utilizing the "Metadata.Operations.enqueueDeployment()" which is already async. Since it's already async, I don't think the MixedDML rules should be an issue. I'm curious to see if this will actually work as expected in the post install script process though. This is a reference link: 

https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_metadata_retrieve_deploy.htm

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: I moved it into a new class named "OPP_StageMappingCMT"

src/classes/OPP_StageMappingUtil.cls
KietTruongSFDO: Changed to upper case

jbrandolinisf:
@force2b - Anything further we need to do with this, or can you approve it?

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+        CMT_MetadataAPI.deployMetadata(listOfMetadataToInsert, callback);
+    }
+
+    public class OppStageToStateMetadataHandler implements CMT_MetadataAPI.MetadataCallbackHandler {
+
+        public Map<String, Object> params;
+
+
+        /**************************************************************************************************************
+        * @description success handler - No implementation as of right now, will be implemented later as needed.
```
force2b: If these are not used is there any benefit in including them in the code base? It's extra code that needs to be tested, but it's not currently providing any value to the application. If there is no use case for it now and the comment says "as needed" I'm assuming there is no future plans to implement it either. If true, I would just remove the unused methods.

src/classes/OPP_StageMappingCMT_TEST.cls
```diff
@@ -0,0 +1,91 @@
...
+    * @return void
+    **********************************************************************************************************/
+    @isTest
+    private static void testFinalizedOppStageToStateRecordCreation() {
+        String closedWonStage = UTIL_UnitTestData_TEST.getClosedWonStage();
+        npe01__Contacts_And_Orgs_Settings__c configRecord = new npe01__Contacts_And_Orgs_Settings__c();
+        configRecord.Payments_Auto_Close_Stage_Name__c = closedWonStage;
+
+        npe01__Contacts_And_Orgs_Settings__c contactAndOrgSetting = UTIL_CustomSettingsFacade.getContactsSettingsForTests(configRecord);
+        UTIL_CustomSettingsFacade.getContactsSettings().Payments_Auto_Close_Stage_Name__c = closedWonStage;
```
force2b: I think these three lines can be reduced into one more readable line:
```
UTIL_CustomSettingsFacade.getContactsSettingsForTests(
    new npe01__Contacts_And_Orgs_Settings__c(
       Payments_Auto_Close_Stage_Name__c = closedWonStage
    )
);
```

src/classes/OPP_StageMappingCMT_TEST.cls
```diff
@@ -0,0 +1,91 @@
...
+
+    /*********************************************************************************************************
+    * @description Tests the createPledgedOppStageToStateRecord Method and creates the pledged record. 
+    * verifies that it returns the appropriate values
+    * @return void
+    **********************************************************************************************************/
+    @isTest
+    private static void testPledgedOppStageToStateRecordCreation() {
+        String closedWonStage = UTIL_UnitTestData_TEST.getClosedWonStage();
+        npe01__Contacts_And_Orgs_Settings__c configRecord = new npe01__Contacts_And_Orgs_Settings__c();
```
force2b: Same change here to reduce the 3 lines into one.

src/classes/OPP_StageMappingCMT.cls
jbrandolinisf: I will defer to @KietTruongSFDO  on this, but I believe we are required to have a success handler for the callback, even if it does not do anything.  If not, I agree we can eliminate it.

force2b: I was looking for that method name being used in the PR, but didn't see it referenced.

KietTruongSFDO: It's required to be implemented. They're both methods of the interface "CMT_MetadataAPI.MetadataCallbackHandler"

src/classes/OPP_StageMappingCMT_TEST.cls
KietTruongSFDO: Updated

src/classes/OPP_StageMappingCMT_TEST.cls
KietTruongSFDO: Updated

KietTruongSFDO:
Hey @force2b, any other changes necessary for this PR?

force2b:
@KietTruongSFDO Looks to be fine. I can re-review after all of John's requested changes have been made if you'd like.

KietTruongSFDO:
@force2b  I believe I've taken care of @jbrandolinisf 's requested changes already. John, let me know if I've missed something.

KietTruongSFDO:
@force2b I misread the PR, I thought you were a required reviewer. Sorry about that!

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
+/*
```
jbrandolinisf: For any new files, can you update ApexDocContent?  For the new files for this, I believe they would go in the Opportunity.htm file, in the Stage to State Section at the bottom.

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+    private static final String OPP_STATE_FIELD = 'Opportunity_State__c';
+    private static final String PLEDGED_STAGE = 'PLEDGED';
+
+    public OPP_StageMappingCMT() {
+
+    }
+
+    /*******************************************************************************************************
+    * @description Creates the finalized state record if there are no finalized records. It utilizes the 
+    * "Opportunity Stage when Fully Paid" field as the Opportunity stage for the finalized state.
```
jbrandolinisf: The overly precise monster in me, wants you to put something about if the "Opportunity Stage when Fully Paid" field is populated, but it is up to you.

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+
+        for (Schema.PicklistEntry picklistEntry : picklistEntryList) {
+            if (picklistEntry.getValue().toUpperCase() == PLEDGED_STAGE) {
+                normalPledgeValue = picklistEntry.getValue();
+                hasPledgedValue = true;
+                break;
+            }
+        }       
+        
+        // Take the value, if there is one, and create a finalized Opp Stage to State custom metadata type record.
```
jbrandolinisf: Can you fix the comment?  It says finalized, when it should say Committed

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+    * @description Creates the initial Opportunity Stage to State record if there are no existing records. 
+    * When there are no existing records, it will create a new "Committed" state record for a "Pledged" stage
+    * if there is one. 
+    ********************************************************************************************************/
+    public static List<Metadata.CustomMetadata> createPledgedOppStageToStateRecord(List<Metadata.CustomMetadata> metadataToInsert) {
+        npe01__Contacts_And_Orgs_Settings__c contactAndOrgCustomSetting = UTIL_CustomSEttingsFacade.getOrgContactsSettings();
+
+        
+        Integer overallCount = [SELECT Count() FROM Opportunity_Stage_To_State_Mapping__mdt];
+
```
jbrandolinisf: Why not return here if a record is present, rather than iterating over the the picklist entries?  Also...since we are using this only to tell whether or not there is one or more records present, I usually do a 

select Id from <<Object>>> limit 1, and then check for 1 or zero records.  It is more performant than a count()

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+        // Only create the pledged opp stage to state record if there's no existing opportunity stage to state records at all. 
+        if (overallCount == 0 && hasPledgedValue) {
+            metadataToInsert.add(createCustomMetadata(normalPledgeValue, COMMITTED_STATE_VAL));
+        }
+
+        return metadataToInsert;
+    }
+
+    /*******************************************************************************************************
+    * @description Creates a custom metadata record based on the parameters provided to the method.
```
jbrandolinisf: Please add the @param entries for the parameters and the @return entry for the method.

src/classes/OPP_StageMappingCMT.cls
jbrandolinisf: Needs ApexDoc to be completed here as well.

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+        fieldMapping.put(OPP_STATE_FIELD, oppState);
+
+        Metadata.CustomMetadata md = new Metadata.CustomMetadata();
+        md = CMT_MetadataAPI.addFieldsToMetadataObject(customMd, fieldMapping);
+
+        return md;
+    }
+
+    /*******************************************************************************************************
+    * @description Deploys the Custom metadata records using the OppStageToStateMetadataHandler as the handler
```
jbrandolinisf: ApexDoc

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+        String jobId = CMT_MetadataAPI.getUniqueJobId();
+        CMT_MetadataAPI.MetadataCallbackHandler oppCallbackHandler = new OppStageToStateMetadataHandler();
+        CMT_MetadataAPI.MetadataCallBack callback = new CMT_MetadataAPI.MetadataCallBack(jobId);
+        callback.callbackHandler = oppCallbackHandler;
+        callback.callbackParams = params;
+
+        CMT_MetadataAPI.deployMetadata(listOfMetadataToInsert, callback);
+    }
+
+    public class OppStageToStateMetadataHandler implements CMT_MetadataAPI.MetadataCallbackHandler {
```
jbrandolinisf: Maybe a comment here, as well explaining the purpose

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+
+        public Map<String, Object> params;
+
+
+        /**************************************************************************************************************
+        * @description success handler - No implementation as of right now, will be implemented later as needed. 
+        /*************************************************************************************************************/
```
jbrandolinisf: Complete the doc header for the methods please.

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+            try {
+                // TODO: Implement if needed. 
+                // Process a successful 
+            } catch (Exception ex) {
+                logError(ex);
+            }
+        }
+
+        /**************************************************************************************************************
+        * @description error handler - No implementation as of right now, will be implemented later as needed.
```
jbrandolinisf: ApexDoc

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,193 @@
...
+        public void performErrorHandler(Map<String, Object> params, String status) {
+            if (params == null || params.isEmpty()) {
+                return;
+            }
+
+            try {
+                // TODO: Implement if needed. 
+                // Handle error 
+            } catch (Exception ex) {
+                logError(ex);
```
jbrandolinisf: Why don't we remove the logError call and the logError method for now.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -91,10 +91,53 @@ public with sharing class OPP_StageMappingUtil {
...
+    * return an artificially created list. Otherwise, get it from the database
+    ********************************************************************************************************/
+    private static List<Opportunity_Stage_To_State_Mapping__mdt> getOpportunityStageToStateMappings() {
+        List<Opportunity_Stage_To_State_Mapping__mdt> metadataRecordList = new List<Opportunity_Stage_To_State_Mapping__mdt>();
+
+        if (Test.isRunningTest()) {
+            
+            // If it's a test running, return artificial records. This ensures consistency in the testing. 
+            Opportunity_Stage_To_State_Mapping__mdt finalizedRecord = new Opportunity_Stage_To_State_Mapping__mdt();
+            finalizedRecord.Opportunity_Stage__c = 'Closed Won';
```
jbrandolinisf: Can you make these static final Strings at the top, and call them TEST_COMMITTED_STAGE and TEST_FINALIZED_STAGE?

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -112,6 +112,28 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+        // We need to parse and check so that only NPSP specific actions will have telemetry execution.
+        // We don't want this to run for Abacus.
```
jbrandolinisf: What is the comment about not wanting to run for Abacus referencing?

force2b:
No problem @KietTruongSFDO

src/classes/OPP_StageMappingCMT.cls
KietTruongSFDO: Added OPP_StageMappingCMT to the Opportunity.htm file

src/classes/OPP_StageMappingCMT.cls
KietTruongSFDO: Updated

src/classes/UTIL_OrgTelemetry_SVC.cls
KietTruongSFDO: Updated the comment to be more clear

src/classes/OPP_StageMappingCMT.cls
```diff
@@ -0,0 +1,196 @@
...
+    private static final String PLEDGED_STAGE = 'PLEDGED';
+
+    public OPP_StageMappingCMT() {
+
+    }
+
+    /*******************************************************************************************************
+    * @description Creates the finalized state record if there are no finalized records. It utilizes the 
+    * "Opportunity Stage when Fully Paid" field (If it's populated) as the Opportunity stage for the finalized state. 
+    * @params List<Metadata.CustomMetadata> metadataToInsert A running list of custom metadata that is going to be inserted,
```
jbrandolinisf: @param, not @params and @return not @returns

src/classes/OPP_StageMappingCMT_TEST.cls
jbrandolinisf: Looks like the original line is still in there as well.  Technically, this whole thing (lines 43 - 51) could be come:

UTIL_CustomSettingsFacade.getContactsSettingsForTests(
    new npe01__Contacts_And_Orgs_Settings__c(
       Payments_Auto_Close_Stage_Name__c = UTIL_UnitTestData_TEST.getClosedWonStage()
    )
);

src/classes/OPP_StageMappingCMT_TEST.cls
```diff
@@ -0,0 +1,97 @@
...
+    private static void testPledgedOppStageToStateRecordCreation() {
+        String closedWonStage = UTIL_UnitTestData_TEST.getClosedWonStage();
+
+        npe01__Contacts_And_Orgs_Settings__c contactAndOrgSetting = UTIL_CustomSettingsFacade.getContactsSettingsForTests(
+                                                                        new npe01__Contacts_And_Orgs_Settings__c(
+                                                                            Payments_Auto_Close_Stage_Name__c = closedWonStage
+                                                                        )
+                                                                    );
+
+        UTIL_CustomSettingsFacade.getContactsSettings().Payments_Auto_Close_Stage_Name__c = closedWonStage;
```
jbrandolinisf: see previous

ApexDocContent/Opportunity.htm
```diff
@@ -54,7 +54,8 @@ <h2>Opportunity Classes</h2>
         <p> The classes involved with Opportunity Stage to State Mapping include:</p>
         <ul>
             <li>OPP_StageMappingUtil</li>
-            <li>OPP_StageMappingUtil_TEST</li>
+			<li>OPP_StageMappingUtil_TEST</li>
```
jbrandolinisf: The indentation looks off, but it could be my browser.  Can you confirm?

KietTruongSFDO: I just pulled the code locally and the spacing looks fine. Not sure why it looks like this on the diff.

KietTruongSFDO:
Hi @davisagli, why was this PR closed?

davisagli:
@KietTruongSFDO my apologies, I deleted the feature/scuevo/crlp-api-retrieve-rollup-state branch because it didn't have any changes compared to master, but that automatically closed this PR (I didn't realize that there was an open PR targeting that branch).

I'll try to undo that but is there a reason this PR is targeting the feature/scuevo/crlp-api-retrieve-rollup-state branch instead of master?

KietTruongSFDO:
@davisagli I was targeting that branch because at the time, I needed some changes to "CMT_MetadataAPI" that Patrick Yan had changed at the time. That was many weeks ago, so maybe those changes are already in master at this point since you mentioned there's no diff compared to master.

davisagli:
@KietTruongSFDO it looks like feature/scuevo/crlp-api-retrieve-rollup-state was merged into master on June 20: https://github.com/SalesforceFoundation/Cumulus/pull/4197

jbrandolinisf:
We will need to recreate this PR from master, right @davisagli

davisagli:
@jbrandolinisf Based on my understanding of the situation, it would make sense for this PR to target master. You can do that by editing the PR and changing its base branch; you don't need to open a new PR.

jbrandolinisf:
Done, and thanks for the tip!

KietTruongSFDO:
@jbrandolinisf thanks for switching the base to master!

PR #4300, opened 2019-06-03
fix reports for salesforce console

### LINKED ISSUE #4018: Household Mailing List Button Not Working in Console App
When the Household Mailing List button is clicked while in a Lightning Console App, the report is not filtering properly. This is happening as a result of the way we construct the URL for the report. On Line [120 of the HH_CampaignDedupeBTN_CTRL class](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/HH_CampaignDedupeBTN_CTRL.cls#L120), we are checking to see if the User is in Lightning ("Theme4d" -- but the console is actually "Theme4u"). If so, we construct the URL one way. If not, we construct it for Classic. 

```
if (USER_UserService.isUserRunningLightning()) {
                    newPageUrl = USER_UserService.getLightningBaseURL(campaignMailingListReportId) + '?fv0=' + ActiveID;
                } else {
                    newPageUrl = '/' + campaignMailingListReportId + '?retURL=/' + ActiveID + '&pv0=' + ActiveID;
                }
```

USER_UserService's isUserRunningLightning method determines if a User is in Lightning by looking for UserInfo.getUiThemeDisplayed(). If it returns Theme4d, they are:
```

public static Boolean isUserRunningLightning() {
        return (UserInfo.getUiThemeDisplayed() == 'Theme4d');
    }
```

**The issue is that the Console app returns _Theme4u_ and not Theme4d, so we attempt to construct the Classic URL and the report just doesn't work.**

---

### **Steps to Repeat.**
1. Create 2 Contacts

2. Create a Campaign and add the two Contacts as Campaign Members.

3. Go to the Campaign and click the Household Mailing List button (note: if you're in a scratch org with unmanaged code, you'll need to create a new button with no namespace and add that button to your page layout instead).

4. Results: the report loads properly.

5. Create a Console App:

- Go to Setup --> App Manager. 
- New Lightning App
- Name = My Console Test
- Choose Console Navigation in App Options
- Add Campaigns, Contacts, Accounts to Navigation Items.
- Add the System Admin Profile to User Profiles
- Save.

6. Switch to the My Console Test app

7. Go to your Campaign and click your Household Mailing List button. The report shows 0 results.

patrick-yan-sf:
Updated the lightning experience UI check to include Lightning console UIs and explicitly unlocked the Campaign Filter Criteria in an NPSP report.

# Critical Changes
- We changed the Campaign ID filter in the NPSP Campaign Household Mailing List V2 report from a locked filter to an unlocked filter so that the report URL would continue to populate the Campaign ID dynamically. However, **in orgs that existed before NPSP version 3.157, an Admin must manually make this change**.  The steps are:

  1. Click the Reports tab.
  2. Choose the All Reports filter from the left column.
  3. Search for "NPSP Campaign Household Mailing List V2" and open it.
  4. Click Edit.
  5. Click the Filters tab in the left column.
  6. Choose the Campaign ID filter.
  7. Deselect the Locked checkbox and click Apply.
  8. Save the Report.
  9. NOTE: It's important that the Campaign ID filter is the first custom filter in the report. It should appear just below the Show Me filter in the report builder.

**Customers starting from a new trial in NPSP version 3.157 don't need to take any action as this change is already included.**

# Changes

# Issues Closed
Fixes #4018 

# New Metadata

# Deleted Metadata


patrick-yan-sf:
**lurch: attach W-030943

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fR29QAE" target="_blank">W-030943</a>

patrick-yan-sf:
@sam-knox @force2b - how do we document the need for the users to manually change the report?  Is that in the critical changes section?

PR #4301, opened 2019-06-04
Psadie/bdi co  migration utility preserve standard field mapping set

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -127,6 +121,23 @@ public class BDI_MigrationMappingUtility {
         FROM Data_Import_Field_Mapping_Set__mdt
         WHERE DeveloperName = :DEFAULT_FIELD_MAPPING_SET_RECORD_NAME];
 
+    /**
+    * @description Boolean flag to determine if a custom field mapping set exists 
+    */
+    public Boolean customFieldMappingSetExists {
```
ptewson-sfdo: Are you still using this method?

psadie: Yep, still using this in a couple of places. One at the beginning of a migration to check whether or not we're going to be comparing against the default field set or the custom field set and when we're updating the data import custom settings (my thinking here is that when Advanced Mapping is turned on through the migration utility, we always point to the custom field mapping set if it exists).

ptewson-sfdo: Oh right - not a method, variable with a getter. Duh.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -198,19 +202,22 @@ public class BDI_MigrationMappingUtility {
...
+    public static Data_Import_Settings__c upsertCustomSettings(String fieldMappingMethod, String status) {
         Data_Import_Settings__c dataImportSettings = UTIL_CustomSettingsFacade.getDataImportSettings();
         dataImportSettings.Field_Mapping_Method__c = fieldMappingMethod;
 
-        if (migrationType == CUSTOM_MAPPING) {
+        List<Data_Import_Field_Mapping_Set__mdt> fieldMappingSets = [SELECT Id
+            FROM Data_Import_Field_Mapping_Set__mdt
+            WHERE DeveloperName = :MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME];
+
+        Boolean customFieldMappingSetExists = fieldMappingSets.size() > 0 ? true : false;
```
ptewson-sfdo: This local variable has the same name as the class variable - if you truly need both is it possible to make the distinction clear in the names?

src/classes/BDI_MigrationMappingUtility.cls
ptewson-sfdo: I notice below you have a local variable with the same name that just checks the list size > 0 to set the boolean - I just want to be sure a) we don't get tripped up with the two variables having the same name and b) aren't querying twice, once to see if it exists and again to pull the value.

psadie: 100%. Looking at this again what do you think of the following...

Right now I have both a class and local var because upsertCustomSettings is static. I use it in BDI_MigrationMappingUtility, DeploymentCallback, and STG_PanelDataImportAdvancedMapping_CTRL (vfp controller). Doesn't really make sense now but I think it must've at point before the changes.

I think I'd prefer to unstatic-fy upsertCustomSettings, toss out that local var, and use the class var. I'll just create a new instance of BDI_MigrationMappingUtility in the DeploymentCallback class where I need to call upsertCustomSettings. Not needed in STG_PanelDataImportAdvancedMapping_CTRL since already creating an instance of BDI_MigrationMappingUtility in the constructor and can use that.

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -159,53 +159,55 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
...
-                migrationMappingUtility.buildCustomMetadataRecordsFromHelpText(DEFAULT_FIELD_MAPPING_SET_RECORD_NAME);
+                // Collect and inspect Help Text mappings
+                // This method will populate the lists pendingMappings, pendingStandardMappings, and conflictingMappings
+                //migrationMappingUtility.buildCustomMetadataRecordsFromHelpText(DEFAULT_FIELD_MAPPING_SET_RECORD_NAME);
 
-                if (!migrationMappingUtility.conflictingMappings.isEmpty()) {
-                    isConflict = true;
-                    isPolling = false;
+                if (migrationMappingUtility.customFieldMappingSetExists == true) {
+                    // Custom field mapping set exists
```
kenlewis-sfdo: This comment can be removed.

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -159,53 +159,55 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
...
-                            new ApexPages.Message(
-                                ApexPages.Severity.CONFIRM,
-                                Label.stgDeploymentCustomMetadataMappingEnabled
-                            )
-                        );
-                    } else {
-                        Id deploymentId = migrationMappingUtility.deployCustomMetadata();
-                        saveDeploymentId(deploymentId);
-                    }
+                    // Custom field mapping set doesn't exist
```
kenlewis-sfdo: This comment can be removed.

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -159,53 +159,55 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     public PageReference enableDataImportFieldMapping() {
         STG_SettingsManager_CTRL.idPanelCurrent = idPanel();
 
+        // Check the deployment status and set toggleDisabled value
         deployment = checkDeploymentStatus();
 
         if (toggleDisabled == false || toggleDisabled == null) {
```
kenlewis-sfdo: I think toggleDisabled should be changed to be a positive, and more descriptive to what it actually represents.  The actual toggle on the page UI can respond to the attribute value, which should be called something like `hasDeploymentInProgress`

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -47,6 +47,12 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     private static final String DEFAULT_FIELD_MAPPING_SET_RECORD_NAME =
```
kenlewis-sfdo: These static final vars that hold details about mapping sets, etc., should probably be stored elsewhere, not in the controller for this page.  Maybe in the BDI_FieldMapping class?

src/classes/BDI_MigrationMappingUtility.cls
ptewson-sfdo: I think this totally makes sense - I'm guessing that the DeploymentCallback was probably operating in a new static context anyway, since it's asynchronous. Let me know if you run into hitches building that out.

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -282,66 +278,127 @@ public class BDI_MigrationMappingUtility {
...
             // Update existing mappings to point to custom field mapping set
+            List<BDI_HelpTextCMT.DataImportFieldMapping> updatedConflictingMappings =
+                new List<BDI_HelpTextCMT.DataImportFieldMapping>();
+
             for (BDI_HelpTextCMT.DataImportFieldMapping dataImportFieldMapping : conflictingMappings) {
-                dataImportFieldMapping.dataImportFieldMappingSetRecordName =
-                    MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME;
-                dataImportFieldMapping = dataImportFieldMapping;
+
+                if (dataImportFieldMapping.dataImportFieldMappingSetRecordName == DEFAULT_FIELD_MAPPING_SET_RECORD_NAME) {
```
ptewson-sfdo: Can we break this line to avoid wrapping around?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -282,66 +278,127 @@ public class BDI_MigrationMappingUtility {
...
+                            MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME
+                        );
+                        queuedMetadataTypesForDeploy.add(clonedFieldMapping.getMetadataRecord());
+                } else {
+                    updatedConflictingMappings.add(dataImportFieldMapping);
+                }
             }
 
+            String fieldMappingSetToCompareAgainst =
+                (customFieldMappingSetExists == true ? MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME : DEFAULT_FIELD_MAPPING_SET_RECORD_NAME);
```
ptewson-sfdo: Can we break this line to avoid wrapping around?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -282,66 +278,127 @@ public class BDI_MigrationMappingUtility {
...
         }
+    }
+
+    /**
+    * @description Method updates existing custom metadata records in memory.
+    * @param fieldMappingSetName Name of parent Field Mapping Set
+    * @param fieldMappings List of field mappings with new field values we want to keep
+    * @return List of custom metadata type records
+    */
+    public List<Metadata.CustomMetadata> updateCustomMetadataRecords(String fieldMappingSetName,
```
ptewson-sfdo: Can we break this line to avoid wrapping around?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -159,53 +159,55 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
...
 
         if (toggleDisabled == false || toggleDisabled == null) {
             Savepoint sp = Database.setSavepoint();
             try {
                 isPolling = true;
 
-                // Check to see if we have any existing help text mappings that conflict with our default cmt mappings
-                migrationMappingUtility.buildCustomMetadataRecordsFromHelpText(DEFAULT_FIELD_MAPPING_SET_RECORD_NAME);
+                // Collect and inspect Help Text mappings
+                // This method will populate the lists pendingMappings, pendingStandardMappings, and conflictingMappings
```
ptewson-sfdo: Can we break this line to avoid wrapping around?

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -159,53 +159,55 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
...
+                //migrationMappingUtility.buildCustomMetadataRecordsFromHelpText(DEFAULT_FIELD_MAPPING_SET_RECORD_NAME);
 
-                if (!migrationMappingUtility.conflictingMappings.isEmpty()) {
-                    isConflict = true;
-                    isPolling = false;
+                if (migrationMappingUtility.customFieldMappingSetExists == true) {
+                    // Custom field mapping set exists
+                    migrationMappingUtility.buildCustomMetadataRecordsFromHelpText(MIGRATED_CUSTOM_FIELD_MAPPING_SET_RECORD_NAME);
```
ptewson-sfdo: Can we break this line to avoid wrapping around?

PR #4311, opened 2019-06-11
BGE Donation Selector Component: Add ability to override BGE donation matching portal labels

kenlewis-sfdo:
This pull request adds global text attributes to the component, enabling override of the values when the component is used in different contexts.  Each attribute has a default value that is set to the Label that was previously used for each string.  

This change enables future implementations to override the values without needing to make any modifications to how the component is currently configured elsewhere.

# Critical Changes

# Changes
- We updated our Batch Gift Entry Donation Selector component to allow labels to be overridden. There’s no change to the Batch Gift Entry labels, but this allows SFDO developers to reuse the component for other features with different labels.

# Issues Closed

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-033054

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uhCQAQ" target="_blank">W-033054</a>

PR #4314, opened 2019-06-12
Telemetry: New feature parameters for BDI and BGE

psadie:
Adding new integer feature parameters Data_CountBDIRowsLast30 and Data_CountBGERowsLast30

# Critical Changes

# Changes
## Telemetry Enhancements
NPSP Feature Telemetry collects feature and usage metrics when customers use our products. We use this information to understand the level of usage and adoption of various NPSP features. No personally identifiable information is collected through NPSP Feature Telemetry. We've added the following to NPSP Feature Telemetry:
- Count of Batch Gift Entry batches within the last 30 days. 
- Count of NPSP Data Import records created within the last 30 days.

# Issues Closed

# New Metadata
## Telemetry Enhancements
FeatureParameterInteger:
- Data_CountBDIRowsLast30
- Data_CountBGERowsLast30
# Deleted Metadata


src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -108,4 +117,23 @@ public without sharing class UTIL_OrgTelemetry_BATCH implements Database.Batchab
...
         }
     }
+
+    /**
+     * @description Batch job to count the number of Data Imports.
+     */
+    private void handleDataImportCounts() {
+        Integer cnt = Database.countQuery('SELECT Count() FROM DataImport__c LIMIT 1');
+        if (cnt == 1) {
+            Database.executeBatch(
```
force2b: @psadie If you're just counting a small number of records (i.e., DataImport records created in the last 30 days) there probably isn't a need to use a batch job. You can just issue the count() query in this method. The batch job is used to handle counts against really large objects like Opportunity or Recurring Donation. Not that the batch job doesn't work, but it's not always necessary. For example, we're not using the batch job to count error rows, even when there could be millions of error row records in the org.

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -70,6 +70,9 @@ private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
         opps.addAll(buildOpps(12, fieldValues));
 
         insert opps;
+
+        // Build data import records
```
seanpat09: This comment probably isn't necessary, it's clear from the method name what's happening here

src/classes/UTIL_OrgTelemetry_BATCH.cls
psadie: Ah that makes a lot of sense. I'll make that update. Thank you!

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -377,6 +380,104 @@ private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
...
+
+        String diBatchQuery = diBatch.start(null).getQuery();
+        List<DataImport__c> dis = Database.query(diBatchQuery);
+        Integer allDis = [SELECT count() FROM DataImport__c];
+
+        Test.startTest();
+        diBatch.execute(null, dis);
+        Test.stopTest();
+
+        System.assertEquals(dis[dis.size() - 1 ].Id, diBatch.lastProcessedId,
```
seanpat09: extra space here
```suggestion
        System.assertEquals(dis[dis.size() - 1].Id, diBatch.lastProcessedId,
```

src/classes/UTIL_OrgTelemetry_SObject_BATCH_TEST.cls
```diff
@@ -403,6 +504,32 @@ private class UTIL_OrgTelemetry_SObject_BATCH_TEST {
...
+     * @description Builds and inserts test data import records
+     */
+    private static void createDataImports() {
+        Integer recordsToCreate = 5;
+        List<DataImport__c> dis = new List<DataImport__c>();
+        DataImportBatch__c diBatch = new DataImportBatch__c(GiftBatch__c = true);
+
+        insert diBatch;
+
+        for (Integer i = 0; i < recordsToCreate; i++) {
```
seanpat09: Instead of creating 5 `DataImport__c` records and then adjusting those, it would be clearer to have 3 distinct `DataImport__c` lists.

- data imports that do NOT have a `NPSP_Data_Import_Batch__c` lookup populated
- data imports that do have a `NPSP_Data_Import_Batch__c` lookup populated
- data imports that were created more than 30 days ago

This way, the 3 lists could have different sizes, which you could put into constants and then the assertions would be a little clearer. 

For example, this is a little clearer as opposed to using a hardcoded `2` so the reader doesn't have to figure out where 2 came from

```
System.assertEquals(NUM_BGE_ROWS_30_DAYS_AGO, diBatch.BGERowsLast30DaysCount,
    'The number of data imports should be tracked correctly');
```

src/classes/UTIL_OrgTelemetry_BATCH.cls
psadie: @force2b - Could I do something like below in UTIL_OrgTelemetrySVC and potentially not have to touch any of batch related classes?

```
/**
* @description Total number of BDI and BGE records created in the last 30 days
*/
private void handleDataImportCount() {
    try {
        Integer bdiCount = Database.countQuery('SELECT count() FROM DataImport__c WHERE CreatedDate = Last_n_days:30');
        featuremanager.setPackageIntegerValue(TelemetryParameterName.Data_CountDataImportsBDILast30Days.name(), bdiCount);

        Integer bgeCount = Database.countQuery('SELECT count() FROM DataImport__c WHERE CreatedDate = Last_n_days:30 AND NPSP_Data_Import_Batch__r.GiftBatch__c = true');
        featuremanager.setPackageIntegerValue(TelemetryParameterName.Data_CountDataImportsBGELast30Days.name(), bgeCount);
    } catch (Exception e) {
        // Default these to -1 so we can determine (in the LMO) if there was an error retrieving them
        featuremanager.setPackageIntegerValue(TelemetryParameterName.Data_CountDataImportsBDILast30Days.name(), -1);
        featuremanager.setPackageIntegerValue(TelemetryParameterName.Data_CountDataImportsBGELast30Days.name(), -1);
    }
}
```

force2b: Absolutely. My only comment is that this assumes both count queries can be completed in a single execution context. That may be a perfectly fine assumption given expected row counts in those objects. However if you feel that issuing both count queries back to back in the same context could run a little long this could be split into two methods (which is how you've split up the categories anyway - `Data_CountBDIRows` and `Data_CountBGERows` - in the `processTelemetryType()` method.

psadie:
**lurch: attach W-033417

psadie:
**lurch: attach W-033189

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087vATQAY" target="_blank">W-033417</a>

psadie:
**lurch: attach W-033189

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087usFQAQ" target="_blank">W-033189</a>

PR #4320, opened 2019-06-14
W-033232 - Adding the ability to create parent additional objects

callen-sfdo:
- Added the ability to have additional objects created as a parent of their predecessor, and for the id of the new record to populated in the designated relationship field on the child predecessor.
- Also started using an existing utility method to generate new SObjects which should reduce  metadata describe calls.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_AdditionalObjectService.cls
+++ b/src/classes/BDI_AdditionalObjectService.cls
@@ -193,28 +193,66 @@ public with sharing class BDI_AdditionalObjectService {
 
             BDI_ObjectWrapper[] objWrapsForChildPredecessorUpdate = new BDI_ObjectWrapper[]{};
 
+            List<BDI_ObjectWrapper[]> objWrapsForUpdateList = new List<BDI_ObjectWrapper[]>();
+            BDI_ObjectWrapper[] currentObjWrapsForUpdate = new BDI_ObjectWrapper[]{};
+
+            List<BDI_ObjectWrapper[]> objWrapsForInsertList = new List<BDI_ObjectWrapper[]>();
+            BDI_ObjectWrapper[] currentObjWrapsForInsert = new BDI_ObjectWrapper[]{};
+
+            List<SObject[]> sObjsForUpdateLists = new List<SObject[]>();
+            SObject[] currentSObjectForUpdateList = new SObject[]{};
+
+            List<SObject[]> sObjsForInsertLists = new List<SObject[]>();
+            SObject[] currentSObjectForInsertList = new SObject[]{};
+
+            Integer objTypeCount = 1;
             for (String objApiName : objAPINameToObjWraps.keySet()) {
 
                 BDI_ObjectWrapper[] objWraps = objAPINameToObjWraps.get(objApiName);
-                
-                SObject[] sObjsForUpdate = new SObject[]{};
-                SObject[] sObjsForInsert = new SObject[]{};
-
-                BDI_ObjectWrapper[] objWrapsForUpdate = new BDI_ObjectWrapper[]{};
-                BDI_ObjectWrapper[] objWrapsForInsert = new BDI_ObjectWrapper[]{};
 
                 for (BDI_ObjectWrapper objWrap : objWraps) {
                     if (objWrap.sObj.Id != null) {
-                        sObjsForUpdate.add(objWrap.sObj);
-                        objWrapsForUpdate.add(objWrap);
+                        currentSObjectForUpdateList.add(objWrap.sObj);
+                        currentObjWrapsForUpdate.add(objWrap);
                     } else {
-                        sObjsForInsert.add(objWrap.sObj);
-                        objWrapsForInsert.add(objWrap);
+                        currentSObjectForInsertList.add(objWrap.sObj);
+                        currentObjWrapsForInsert.add(objWrap);
                     }
                 }
 
-                if (sObjsForUpdate.size() > 0) {
+                //If the loop is on its 10th run then add the lists to the array 
+                // and start a new list and reset the counter.
+                if (objTypeCount == 10) {
+                    sObjsForUpdateLists.add(currentSObjectForUpdateList);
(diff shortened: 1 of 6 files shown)


callen-sfdo:
**lurch: attach W-033232

callen-sfdo:
@melissabarber Hi Melissa, there is one new custom label to review in the CustomLabels.labels file.  Please let me know if you have any questions.

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uvdQAA" target="_blank">W-033232</a>

src/labels/CustomLabels.labels
```diff
@@ -1913,6 +1913,13 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>bdiAccountCustomIdError</shortDescription>
         <value>The Account Custom Unique ID setting field {0} must have matching fields in the NPSP Data Import object whose API Names are {1} and {2}.</value>
     </labels>
+    <labels>
+        <fullName>bdiAdditionalObjChildNotUpdated</fullName>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>bdiAdditionalObjChildNotUpdated</shortDescription>
+        <value>Error when updating child record lookup field(s) with new parent record id. Error message is:</value>
```
```suggestion
        <value>Unable to update the child record's lookup field with new parent record's ID. It was unable to update because of this system error:</value>
```

src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -50,7 +50,7 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
         for (BDI_ObjectWrapper objWrap : objWraps) {
             Map<String, Schema.DescribeFieldResult> targetFieldDescByFieldName = UTIL_Describe.getAllFieldsDescribe(objWrap.objMapping.Object_API_Name__c);
 
-            objWrap.sObj = Schema.getGlobalDescribe().get(objWrap.objMapping.Object_API_Name__c).newSObject();
+            objWrap.sObj = UTIL_Describe.getPrototypeObject(objWrap.objMapping.Object_API_Name__c);
```
ptewson-sfdo: nice!

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -279,7 +291,90 @@ public with sharing class BDI_AdditionalObjectService {
...
+
+        Map<String,Map<Id,SObject>> childSObjectMapByObjectAPIName = new Map<String,Map<Id,SObject>>();
+        Map<Id,BDI_ObjectWrapper[]> successorObjWrapsByPredecessorId = new Map<Id,BDI_ObjectWrapper[]>();
+
+        for (BDI_ObjectWrapper objWrap : objWrapsForChildPredecessorUpdate) {
+
+            Id predecessorId = (Id)objWrap.dataImport.get(objWrap.predecessorObjMapping.Imported_Record_Field_Name__c);
+            SObject sObjForUpdate;
+
+            Map<Id,SObject> tempSObjMap = new Map<Id,SObject>();
```
ptewson-sfdo: Even though this is internal to this loop, can we rename this "sObjectById" just for clarity of what's being stored? Or even tempSObjectById if we want to label it as a temporary collection

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -279,7 +291,90 @@ public with sharing class BDI_AdditionalObjectService {
...
+
+            // Store the successor object wraps by their predecessor id for later use in recording errors.
+            if (successorObjWrapsByPredecessorId.get(predecessorId) != null) {
+                successorObjWrapsByPredecessorId.get(predecessorId).add(objWrap);
+            } else {
+                successorObjWrapsByPredecessorId.put(predecessorId, new BDI_ObjectWrapper[]{objWrap});
+            }
+        }
+
+        if (childSObjectMapByObjectAPIName.size() > 0) {
```
ptewson-sfdo: I think we can skip this "if" statement - if the childSObjectMapByObjectAPIName is empty, the for loop should just never iterate on the empty keyset, right?

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -279,7 +291,90 @@ public with sharing class BDI_AdditionalObjectService {
...
+            for (String objAPIName : childSObjectMapByObjectAPIName.keySet()) {
+
+                SObject[] sObjsForUpdate = childSObjectMapByObjectAPIName.get(objAPIName).values();
```
ptewson-sfdo: Since we are handling this as a generic list of sobjects anyway, and it didn't look like there was any state being tracked throughout the iterations, could we aggregate all the values from childSObjectMapByObjectAPIName's various objects into the the sObjsForUpdate list (instead of instantiating new) and then close the for loop before going into the DML block below this? It would prevent having to do DML inside the for loop, and would generally reduce CPU and # of DML operations. 

We might have to consider the 10 "chunk" limit on DML if we did this, although it would be a pretty crazy setup to get there I think

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: updated the name.

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: I've made this update here, and also for the main sObject update and insert for additional objects.  Please review.

src/classes/BDI_AdditionalObjectService.cls
callen-sfdo: removed

callen-sfdo:
@ptewson-sfdo I've made the changes grouping DML and made the other updates you requested.  Please review and let me know if you see any issues.

PR #4325, opened 2019-06-16
More robust in the face of API change-overs.

prescod:
# Changes

Allows NPSP robot to run on newer releases before they have locator libraries.


robot/Cumulus/resources/NPSP.py
```diff
@@ -51,7 +51,8 @@ def _init_locators(self):
             response = client._call_salesforce(
                 'GET', 'https://{}/services/data'.format(client.sf_instance))
             latest_api_version = float(response.json()[-1]['version'])
-        except RobotNotRunningError:
+            assert latest_api_version  in locators_by_api_version
+        except (RobotNotRunningError, AssertionError):
```
davisagli: I'd prefer an explicit

    if latest_api_version not in locators_by_api_version:
        latest_api_version = max(locators_by_api_version)

`assert` anywhere other than in pytest always makes me spend extra brain cycles convincing myself no one is going to run python with -O. But more to the point, I don't want to accidentally mask other assertions that could exist in the code invoked by _call_salesforce.

We should also probably log something in this case. Actually it wouldn't hurt to always log which version of the locators was chosen.

prescod: What logger is appropriate in NPSP.py? A new one?

davisagli: I would use self.builtin.log() which I see used elsewhere in the module

robot/Cumulus/resources/NPSP.py
```diff
@@ -51,6 +52,9 @@ def _init_locators(self):
             response = client._call_salesforce(
                 'GET', 'https://{}/services/data'.format(client.sf_instance))
             latest_api_version = float(response.json()[-1]['version'])
+            if not latest_api_version in locators_by_api_version:
+                warnings.warn("Could not find locator library for API %d" % latest_api_version)
```
davisagli: @prescod I asked for using `self.builtin.log()` because that will end up in the Robot Framework log output which is presumably where people will be looking to debug robot tests.

@boakley let me know if I'm mistaken

prescod: I tested it and it doesn't seem to end up in any file. Most likely because it happens before the testing has started.

davisagli: Oh, that seems plausible. Okay, let's leave it for now.

PR #4332, opened 2019-06-18
Telemetry for opportunity payments

seanpat09:
# Critical Changes

# Changes
 - We added two data points to NPSP Telemetry:
   - Count of Opportunities with Multiple Payments
   - Is a custom Recurring Donation schedule in use

# Issues Closed

# New Metadata

Apex Classes:
* UTIL_OrgTelemetry_SVC_TEST.cls

FeatureParameters:
* Data_CountOppsWithMultiplePayments.featureParameterInteger
* DataCount_ONETIME_Use.featureParameterInteger

# Deleted Metadata


seanpat09:
**lurch: attach W-033188

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087usAQAQ" target="_blank">W-033188</a>

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -56,13 +56,22 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
...
+    
     /**
      * @description BatchJobMode is to add support for other query types in this same batch job
      * that aren't yet needed, but may be in the future
      */
     public enum TelemetryBatchJobMode {
         RECURRING_DONATIONS,
-        OPPORTUNITIES
+        OPPORTUNITIES,
+        PAYMENTS
```
force2b: I know this complicates the job a bit, but since the job to count the number of payments is effectively querying all Opportunity records and then doing a sub-query on Payments, I think this should be combined with the OPPORTUNITIES mode. That mode queries all Opportunities. In an org with 30M Opportunities, I'd much rather have one job iterating over all 30M than two.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -190,6 +201,23 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+    /**
+     * @description Checks if custom installments are enabled and sets feature parameter. Enabled defined as having
+     * custom installment settings and having at least one recurring donation that has one of those custom installments
+     */
+    private void handleCustomInstallmentSettings() {
+        Set<String> customInstallments = UTIL_ListCustomSettingsFacade.getMapCustomInstallmentSettings().keySet();
+
+        Boolean hasCustomInstallments =
+            !customInstallments.isEmpty()
+            && ![SELECT Id FROM npe03__Recurring_Donation__c WHERE npe03__Installment_Period__c IN :customInstallments].isEmpty();
```
force2b: I suggest adding `LIMIT 1` to this query

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -133,6 +137,7 @@ public without sharing class UTIL_OrgTelemetry_SVC {
             // operation.
             when NPSP_Settings {
                 handleNPSPSettings();
+                handleCustomInstallmentSettings();
```
force2b: Because the `handleCustomInstallmentSettings` method has a query against the RD object which could (though unlikely) timeout, it might be better to separate this action into a different `TelemetryBatchCategory`.

src/classes/UTIL_OrgTelemetry_BATCH.cls
```diff
@@ -101,11 +104,66 @@ public without sharing class UTIL_OrgTelemetry_BATCH implements Database.Batchab
...
+
+    /**
+    * @description Determines if payments telemetry batch should run. Is test visible so this can be tested outside of
+    * the finish method without executing UTIL_OrgTelemetry_SObject_BATCH. Executing batch can cause internal errors
+    * to get thrown when tests in the same class set the same Feature Parameters
+    */
+    @TestVisible
+    private Boolean shouldExecutePaymentsTelemetryBatch() {
+        List<AggregateResult> results = [
+            SELECT count(Id) numPayments
```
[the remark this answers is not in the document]
seanpat09: If we can't check if if any opp has more than one payment, if any payments exist, should we run the batch regardless of whether Automatic Payments are enabled?

seanpat09:
@force2b Thanks again for the feedback - I made the suggested changes, let me know if you see anything else!

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -48,14 +48,15 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
     /**
      * @description Batch size for OPPORTUNITY job mode
      */
-    public static final Integer OPPORTUNITIES_BATCH_SIZE = 2000;
+    public static final Integer OPPORTUNITIES_BATCH_SIZE = 200;
```
force2b: I don't think this needs to be 200. You're never going to query more than 2 payment records per Opportunity, so even if every Opp in a batch of 2000 had 2 payments that would only be 4000 rows out of a max of 50000 rows.

src/classes/UTIL_OrgTelemetry_SVC_TEST.cls
```diff
@@ -0,0 +1,52 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Utilities
+* @description Unit Test for UTIL_OrgTelemetry_SVC
+*/
+
+@isTest
```
force2b: Do you think this can use `IsParallel=true`?

seanpat09: I think so, I ran all the tests in parallel a few times and it they still passed

force2b: Do you mean you added the `IsParallel=true` annotation or that enable parallel testing in the org? We're still not able to do the latter on our builds.

src/classes/UTIL_OrgTelemetry_Batch_TEST.cls
```diff
@@ -109,10 +109,13 @@ private class UTIL_OrgTelemetry_Batch_TEST {
         batch.finish(null);
 
         System.assertEquals(null, batch.countByBatchMode.get(UTIL_OrgTelemetry_SObject_BATCH.TelemetryBatchJobMode.OPPORTUNITIES),
-            'The OPPORTUNITIES should not be executed');
+            'The OPPORTUNITIES batch batch should not be executed');
```
force2b: `batch` is duplicated here.

src/classes/UTIL_OrgTelemetry_SVC_TEST.cls
seanpat09: I enabled parallel testing in the org. Just to be safe, I'll add `IsParallel=false` to this test

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
```diff
@@ -245,6 +254,26 @@ public class UTIL_OrgTelemetry_SObject_BATCH implements Database.Batchable<SObje
         return oppsByRecurringDonationId;
     }
 
+    /**
+     * @description Handles payment counts
+     */
+    private void countOppPayments(List<Opportunity> opps) {
+        for (Opportunity opp : [
+            SELECT
```
force2b: @seanpat09 I'm not sure why I didn't think of this before, but there is a rollup summary field on the Opportunity for `npe01__Number_of_Payments__c` that is a count of all payment records attached to the Opp. I think you can just use that rather than have to execute the subquery against the child object.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -68,8 +70,11 @@ public without sharing class UTIL_OrgTelemetry_SVC {
         IsEnabled_AddressVerification,
         IsEnabled_AutomaticPayments,
         IsEnabled_CustomizableRollups,
+        IsEnabled_CustomRDSchedules,
```
force2b: @seanpat09 After discussing this more with @mpusto, we felt that knowing about Custom RD Schedule usage is more of a one-time transient need rather than something we need to follow long term. As such, we'd like to change the `IsEnabled_CustomRDSchedules` boolean to the `DataCount_ONETIME_Use` integer feature parameter (returning 1 for orgs that have CustomRDSchedules in use).

Given that there is still a theoretical 25 feature parameter limit per package, the intent is that the `DataCount_ONETIME_Use` feature parameter will be re-used over time to get telemetry that is on a one-time/short need basis. This will be our first implementation of it.

src/featureParameters/IsEnabled_CustomRDSchedules.featureParameterBoolean
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: As described in the other note, this will be come `DataCount_ONETIME_Use.featureParameterInteger`. I think a label such as "Data Count: Short Term Use Telemetry" will work. It's not visible outside of the Packaging Org anyway.

src/classes/UTIL_OrgTelemetry_SVC_TEST.cls
force2b: Actually what I was suggesting is to add `IsParallel=true` to the class so it will run in parallel during builds.

src/classes/UTIL_OrgTelemetry_SVC.cls
seanpat09: That's a good idea, I'll make the change

src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls
seanpat09: haha nice catch! That's a lot more convenient, I will make the necessary updates

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -51,12 +51,13 @@ public without sharing class UTIL_OrgTelemetry_SVC {
      * against a table (such as a really huge Opportunity object) were to fail.
      */
     public enum TelemetryBatchCategory {
+        CustomInstallment_Settings,
         NPSP_Settings,
         Org_Environment,
         Data_MaxNumRelatedOpps,
         Data_CountRdOpps,
         Data_CountRecurringDonations,
-        Data_CountRds,
```
seanpat09: This actually wasn't being used anywhere, I accidentally left this in, it was a duplicate for `Data_CountRecurringDonations`

seanpat09:
@force2b Made the requested changes and resolved merge conflicts. Can you review again?

seanpat09:
> @seanpat09 Sorry for the back and forth on this stuff. After some discussion with Mary and Ali on usage, we've agreed to hold on the `Data_OppPaymentsWithoutPaymentsEnabled` feature parameter for now. I think we'll come back to it at some point though. The
> 
> I recommend removing the feature parameter metadata file and reference in package.xml and then commenting out any references to that in the code rather than deleting totally. This way we can just uncomment the code when we're ready to go live with that feature parameter, which I think we'll happen once we're done with the Custom RD Schedules one-time feature parameter.
> 
> We're keeping the `Data_CountOppsWithMultiplePayments` as is.

No problem, I'll remove the `Data_OppPaymentsWithoutPaymentsEnabled` and comment out the code and keep the `Data_CountOppsWithMultiplePayments`

seanpat09:
@rponti-sforg this is ready for QA!

sam-knox:
@seanpat09 - Can you check the Changes section and make sure what I have there is accurate? I think you also need to remove one of the Feature Parameters in the Metadata section of the release notes.

seanpat09:
> @seanpat09 - Can you check the Changes section and make sure what I have there is accurate? I think you also need to remove one of the Feature Parameters in the Metadata section of the release notes.

@sam-knox Looks good! Good catch, I removed the extra feature parameter from the Metadata section. I also updated the Changes to say `We added two data points to NPSP Telemetry:` as opposed to `three data points`

PR #4334, opened 2019-06-19
UnitTests: Force Unique Contact.LastName in Data Factory Methods

force2b:
Purpose: An attempt to *eliminate* or *seriously reduce* the number of build failures due to Lock errors in unit tests. These are a result of how Salesforce handles data rollbacks in their test methods. When two Contacts with the same LastName exist and one is being rolled back by the system, it creates a lock for that LastName in the system Index which prevents another Contact with that same LastName from being inserted (into the Index). By forcing everything to a unique LastName, the hope is that this will drastically reduce occurrences of this error.

This PR has no customer facing changes. All code modifications are at the Unit Tests level only.
* Force a unique LastName when creating a Contact and a unique Name when creating other types of records.
* Update a BDI Unit Test Method that was looking for a hardcoded LastName based on a data factory that was updated to force a unique LastName value.
* Style and other cleanups as well.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_UnitTestData_TEST.cls
+++ b/src/classes/UTIL_UnitTestData_TEST.cls
@@ -163,7 +163,7 @@ public class UTIL_UnitTestData_TEST {
 
         for (Integer i=0; i<n; i++) {
             Contact newCon = getContact();
-            newCon.FirstName = CAO_Constants.CONTACT_FIRSTNAME_FOR_TESTS + i;
+            newCon.FirstName = CAO_Constants.CONTACT_FIRSTNAME_FOR_TESTS + getUniqueString();
             newCon.LastName = CAO_Constants.CONTACT_LASTNAME_FOR_TESTS;
             if (RLLP_OppRollup_UTIL.isMultiCurrency()) {
                 newCon.put(RLLP_OppRollup_UTIL.mcFieldValues.get('Contact'), RLLP_OppRollup_UTIL.currCorporate);
@@ -304,12 +304,12 @@ public class UTIL_UnitTestData_TEST {
      * @param n The number of Engagement Plan Templates to be created
      * @return List of Engagement Plan Templates
      */
-    public static List<Engagement_Plan_Template__c> createEPTemplates (integer n) {
+    public static List<Engagement_Plan_Template__c> createEPTemplates (Integer n) {
         // Holds the list of EP Templates to be returned
-        List<Engagement_Plan_Template__c> EPTemplatesToAdd = New List<Engagement_Plan_Template__c>();
+        List<Engagement_Plan_Template__c> EPTemplatesToAdd = new List<Engagement_Plan_Template__c>();
         
-        for (integer i=0 ;i<n; i++) {
-            EPTemplatesToAdd.add(new Engagement_Plan_Template__c(Name = 'Test ' + i));
+        for (Integer i=0 ;i<n; i++) {
+            EPTemplatesToAdd.add(new Engagement_Plan_Template__c(Name = 'Test ' + getUniqueString()));
         }
         
         return EPTemplatesToAdd;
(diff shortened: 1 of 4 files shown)


src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -106,195 +138,297 @@ public class UTIL_UnitTestData_TEST {
...
-                npe01__Private__c = false,
-                npe01__WorkEmail__c = CAO_Constants.CONTACT_EMAIL_FOR_TESTS, 
-                npe01__Preferred_Email__c = CAO_Constants.CONTACT_PREFERRED_EMAIL_FOR_TESTS,
-                npe01__WorkPhone__c = CAO_Constants.CONTACT_PHONE_FOR_TESTS,
-                npe01__PreferredPhone__c = CAO_Constants.CONTACT_PREFERRED_PHONE_FOR_TESTS,
-                OtherCity = 'Seattle'
-            );
+        return new Contact (
+            FirstName = CAO_Constants.CONTACT_FIRSTNAME_FOR_TESTS,
+            LastName = CAO_Constants.CONTACT_LASTNAME_FOR_TESTS + getUniqueString(),
```
force2b: This is one of the main changes for uniqueness in this class.

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -106,195 +138,297 @@ public class UTIL_UnitTestData_TEST {
...
-                npe01__WorkPhone__c = CAO_Constants.CONTACT_PHONE_FOR_TESTS,
-                npe01__PreferredPhone__c = CAO_Constants.CONTACT_PREFERRED_PHONE_FOR_TESTS,
-                OtherCity = 'Seattle'
-            );
-            //since we're running rollup code on a manually created contacts in RLLP_OppRollup_TEST, we have to manually set currency field in multicurrency
+        List<Contact> contactsToAdd = new List<Contact> ();
+
+        for (Integer i=0; i<n; i++) {
+            Contact newCon = getContact();
+            newCon.FirstName = CAO_Constants.CONTACT_FIRSTNAME_FOR_TESTS + getUniqueString();
```
force2b: This is one of the main changes for uniqueness in this class.

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -106,195 +138,297 @@ public class UTIL_UnitTestData_TEST {
...
+        insert listAccT;
+
+        for (Integer i = 0; i < cAcc; i++) {
+            for (Integer j = 0; j < cCon; j++) {
+                Integer iCon = (i * cCon) + j;
+                String unique = getUniqueString();
+
+                Contact con = listConT[iCon];
+                con.FirstName = 'TestFirstName' + iUnique + iCon;
+                con.LastName = 'TestLastName' + iUnique + iCon + unique;
```
force2b: This is one of the main changes for uniqueness in this class.

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -436,14 +573,21 @@ public class UTIL_UnitTestData_TEST {
         return buildUser(buildUniqueLastName(), profileName);
     }
 
+    /**
+      * @description Generate a unique string to append to indexed fields to make them unique
+      */
+    public static String getUniqueString() {
+        return String.valueOf(Math.abs(Crypto.getRandomLong()));
```
force2b: This method is intended to generate a completely unique string value to add to a Contact.FirstName, Contact.LastName, Account.Name, etc. In testing, I found that using `getRandomInteger()` could still return a duplicate value if the method was called repeatedly in a loop of 50000 iterations. That did not happen using `getRandomLong()` even in a loop of 100000 iterations.

force2b:
**lurch: attach W-033961

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087vskQAA" target="_blank">W-033961</a>

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -106,195 +138,297 @@ public class UTIL_UnitTestData_TEST {
...
+    * @param iUnique the number to start with when naming the contacts and addresses
+    * @return  void
+    **********************************************************************************************************/
+    public static void createAccountContactTestData(String accType, Integer cAcc, Integer cCon, Integer iUnique) {
+
+        listConT = UTIL_UnitTestData_TEST.createMultipleTestContacts(cAcc * cCon);
+        listAccT = UTIL_UnitTestData_TEST.createMultipleTestAccounts(cAcc, accType);
+        insert listAccT;
+
+        for (Integer i = 0; i < cAcc; i++) {
```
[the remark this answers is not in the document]
force2b: I really thought about changing all the var names, but held back just because there were so many to update.

force2b:
Thanks @RadSFDeveloper. Unfortunately running everything in parallel is a rabbit hole with NPSP. I tried changing the build to enable parallel tests twice before and found there are far too many unit tests that require fixing. The lists of failed classes changes nearly every single time. That's a battle for a **much** larger work item.

PR #4336, opened 2019-06-19
LDV Chunking Support in Common Abstract Base Batch Class

jbrandolinisf:
1. Abstract Base class to ease support LDV Chunking situations.

1. Updates to OPP_OpportunityNaming_BATCH using base class instead of original implementation.

# Critical Changes

# Changes

# Issues Closed

# New Metadata
**Classes**
- UTIL_AbstractCommon_BATCH
- UTILAbstractCommon_BATCH_TEST
- UTIL_SalesforceID
- UTIL_SalesforceID_TEST

# Deleted Metadata


jbrandolinisf:
Hi @force2b and @pedrodalcol,

This is the LDV Base Class work I will be going through on Monday.  I am tagging you as reviewers so you can look at it in advance (if you want to), and have a place to comment afterwards if you see stuff that will require changes.

I did make a POC version of the OpportunityNamingBatch that uses it instead of what it was doing, but for now I created it under a new name.  If you agree with everything, and want to switch even that to use it, it is easy enough for me to make the changes.

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
```diff
@@ -0,0 +1,100 @@
+/*
+    Copyright (c) 2015, Salesforce.org
```
force2b: The year should be updated here and below.

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
```diff
@@ -0,0 +1,100 @@
+/*
```
force2b: Why the new class instead of just refactoring the existing class? I'm not sure we'd want to have both the old and new classes in the repo at the same time.

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,453 @@
...
+* each batch, if executing in LDV Chunking Mode, and the main batch query if executing in non LDV
+* Chunking mode.
+*
+* Future: Add exception handling, and logging support so implementing clases can inherit.
+* Recommend registering a default handler using an interface that can be switched out/overriden
+* as desired for specific use cases.
+*/
+public abstract with sharing class UTIL_AbstractCommon_BATCH implements Database.Batchable<SObject>, Database.Stateful {
+
+    private static final Integer MAX_LDV_CHUNK_SIZE = 99999;
```
force2b: 99K records doesn't seem like a lot for a chunk size if the job is iterating over 80M records. Is there a reason this couldn't be 1M or 10M? 

There's a balance to had over the total run time of the job versus the number of jobs that continue to chain over time. In a more reasonable org of 30M Opportunities, chunking at 99K record would mean that there 301 individual batch jobs executed, each with around 200 child batches (assuming batch size of 500). That's still a lot of actual batch jobs.

If this were to use a chunk size of 1M, then there would only be 30 batch jobs, each with around 2000 child batches. Depending on the complexity of the processing, one could expect each job to finish in around 1-5 hours.

The way that Salesforce dequeues jobs from the stack is generally to wait a little longer between each job as it's executed (when they chain together). That's going by some anecdotal discussions. I've personally seen where jobs that are in holding state will stay in a holding state longer as there are more jobs to process. Anyway, I think there's an advantage to a lower number of total jobs being executed with a balance to the number of records processed by each.

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,453 @@
...
+    @testVisible private String getQueryOrderByANDLimitClause() {
+        return this.ldvMode ? ' ORDER BY Id ASC LIMIT ' + getValidLDVChunkSize() : '';
+    }
+
+    /**
+     * @description SOQL Query Fragment: Calculates the where clause for the batch driver in LDV Chunking and non LDV mode
+     * @return String Where clause fragment
+     */
+    @testVisible private String getQueryWhereClause() {
+        String ldvWhere = isFirstChunk() ? ' WHERE Id >= \'' + startId + '\'' : ' WHERE Id > \'' + startId + '\'';
```
force2b: The ternary conditions are getting a little hard to read here and in many places below. I recommend breaking these up into smaller individual var assignments and then concatenating the results so it's easier to read.

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
```diff
@@ -0,0 +1,100 @@
...
+        if (!oppsForUpdate.isEmpty()) {
+            UTIL_DMLService.updateRecords(oppsForUpdate, false);
+        }
+    }
+
+   /**
+     * @description Returns Opportunity for the main object we are querying
+     * @return String value of SOQLObjectName
+     */
+    public override String getSOQLObjectName() {
```
force2b: It might be better to deal with this as an `SObjectType` rather than a string this way the implementing code doesn't need to worry about conditionally applying a namespace to the string.

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,453 @@
...
+        return 'SELECT Id FROM ' + getSOQLObjectName() + ' LIMIT 0';
+    }
+
+    /**
+     * @description Determines the query to execute factoring all the different params for where clause, ldv versus non ldv, etc.
+     * @param startId starting id of the chunk
+     * @return String queryString to execute for Batch
+     */
+    @testVisible private String getQuery(Id startId) {
+        return 'SELECT ' + getQueryFieldList() + ' FROM ' + getSOQLObjectName() + getQueryWhereClause() + getQueryOrderByAndLimitClause();
```
force2b: Have you thought about using the `UTIL_Query` and `UTIL_Where` classes to handle building the queries?

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
jbrandolinisf: I took this from the original OPP_OpportunityNaming_NEW_BATCH.cls, which we will replace.  Do you still want it changed?

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
jbrandolinisf: I wasn't sure we were going to keep it, so I did it strictly as a proof.  I can replace the original with it, and eliminate the _NEW no problem!

src/classes/UTIL_AbstractCommon_BATCH.cls
jbrandolinisf: So the max chunking size SF lists for their bulk api is 250K.  I refactored for that.  I am basing this on [PK Chunking Header](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/async_api_headers_enable_pk_chunking.htm), but if you have counter/other info, I am just going by what is published.

jbrandolinisf: I do think if we want to go beyond that, I should refactor it into a constructor param, as it will become more of a tuning thing than a default/override situation.

src/classes/OPP_OpportunityNaming_NEW_BATCH.cls
jbrandolinisf: Makes sense, I will refactor for this.

src/classes/UTIL_AbstractCommon_BATCH.cls
jbrandolinisf: I hadn't.  When I started this (in the Abacus repo), those classes did not exist (there).  I think UTIL_Query is an easier fit than UTIL_Where, but I only glanced it over.  Perhaps looking at a few classes that use it if you have ready examples would change my mind.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -0,0 +1,467 @@
...
+
+        List<AsyncApexJob> aaJobs = queryBatchJobs();
+        System.assertEquals (1, aajobs.size(), '# of jobs present');
+    }
+
+    /*******************************************************************************************************************
+    * @description Test Batch Execution with inner test batch class to iterate over and count Accounts while simulating
+    * chunking conditions
+    */
+    @isTest
```
force2b: Indent issue here

src/classes/UTIL_AbstractCommon_BATCH.cls
jbrandolinisf: Keeping this as is, since neither is easy to adapt to all the use cases, and this class will need to exist in Abacus (where the other classes do not exist) and Cumulus.

jbrandolinisf:
@force2b - Want to do one more pass, and approve if all set?  I know we cannot merge until next build.

jbrandolinisf:
@force2b - Finished the changes we talked about this AM.

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,519 @@
...
+     */
+    public UTIL_AbstractCommon_BATCH(Integer batchSize) {
+        // batchSize, ldvMode, additionalWhereClause
+        this(batchSize, false, '');
+    }
+
+    /**
+     * @description Constructor that assumes  that ldvMode will be
+     * determined by record distance between min and max Id for main object of batch
+     * @param additionalWhereClause if present, will be applied to the inner query, or the main query if executing in non ldv mode.  If a main where clause is also present both will be encapsulated in () and anded.
```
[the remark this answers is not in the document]
jbrandolinisf: @force2b - I can do that, but typically don't with @param because I read (like 4 years ago) ApexDoc required params to be on one line.  If we don't care, though I am good with that!

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,519 @@
...
+    @testVisible private String getLdvWhereFirstChunk() {
+        return ' WHERE Id >= \'' + startId + '\'';
+    }
+
+    /**
+     * @description SOQL Query Fragment: Where LDV Where Clause for all chunks beyond the first should not
+     * include the startId of the chunk, as it was already processed
+     * @return String Where clause fragment
+     */
+    @testVisible private String getLdvWhereAfterFirstChunk() {
```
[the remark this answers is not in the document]
jbrandolinisf: @force2b - It is Not.  After the first chunk, it Id >, rather than I >=.  I wish there was a more elegant way to do this...

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,519 @@
...
+     * exceeds either the maximum non LDV size allowed for the job or the LDV Chunk Size, it will return true, and false otherwise.  If the Ids are null, meaning there are no records present,
+     * computeRecordDistance returns a negative number.  In all other cases, the record distance is a positive number.
+     * @param startId Starting Id of the all records present in the object
+     * @param endId Ending Id of the records in the object
+     * @return Boolean true, if it appears to be an LDV Chunking circumstance, and false otherwise
+     */
+    @testVisible private boolean isLDVCircumstance(Id startId, Id endId) {
+        Long recordDistance = UTIL_SalesforceID.computeRecordDistance(startId, endId);
+
+        return (recordDistance > getMaxNonLDVSize() || recordDistance > getLDVChunkSize());
```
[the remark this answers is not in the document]
jbrandolinisf: Eh...you caught me...I added both so I wouldn't have to update the tests that were using a smaller size, but would still allow us to do what we wanted...my reasoning was that if the chunksize is smaller than the number of records, we should still also run ldv...I will update it...

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -0,0 +1,472 @@
...
+
+    /*******************************************************************************************************************
+     * @description Test Batch Execution with inner test batch class to test non chunking conditions
+     */
+    @isTest
+    public static void executeTestBatch() {
+
+        List<Account> accs = new List<Account>();
+
+        for (Integer i = 0; i < TEST_ACCOUNT_CNT; i++) {
```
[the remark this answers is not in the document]
jbrandolinisf: @force2b - I didn't use a TestSetup Method, because I didn't want records present for certain tests.  I could refactor to use the UnitTestData_Test class.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -0,0 +1,472 @@
...
+                new Account(
+                    Name = '1-Test Account #' + i
+                )
+            );
+        }
+        insert accs;
+
+        Test.startTest();
+
+        UTIL_AbstractCommon_BATCH_TEST tab = new UTIL_AbstractCommon_BATCH_TEST(10);
```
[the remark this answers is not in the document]
jbrandolinisf: @force2b - I wanted to...it is not supported/allowed.  This monstrosity was the best i could do, short of a separate class.  It gave a wacky error, that it turned out to be was because it was an inner class.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -0,0 +1,472 @@
...
+
+        UTIL_AbstractCommon_BATCH_TEST tab = new UTIL_AbstractCommon_Batch_TEST(10, 'Name = \'Test Account #1\' or Name=\'Test Account #2\' or Name=\'Test Account #3\'');
+        tab.setTestWhereClause('Name equals \'Test Account #0\' or Name equals \'Test Account #1\' or Name equals \'Test Account #2\'');
+
+        System.assertNotEquals(null, tab.validateQuery(), 'Exception expected');
+    }
+
+    /* Test Harness for Batch Implementation Here... */
+
+    /**
```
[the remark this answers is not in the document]
jbrandolinisf: See prior comment about why it is not an inner class

jbrandolinisf:
@force2b I am done with all the feedback I did not have responses on.  Can you review my responses, and respond?

jbrandolinisf:
@force2b - Let me know if you have any further feedback on this, or if you are ready to approve it!  Thanks!

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -35,56 +35,29 @@
...
-        return Database.getQueryLocator(query);
+    /** @description Constructor to allow specifying batch size.*/
+    public OPP_OpportunityNaming_BATCH(Integer batchSize) {
+        super(batchSize);
     }
 
     /** @description Schedulable execute method.*/
     public void execute(SchedulableContext context) {
-        Database.executeBatch(new OPP_OpportunityNaming_BATCH(), 200);
+        Database.executeBatch(new OPP_OpportunityNaming_BATCH(200), 200);
```
force2b: Minor request. Can you make the 200 batch size a constant at the top of this class?
`private static final Integer BATCH_SIZE = 200;`

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -107,21 +80,28 @@ public class OPP_OpportunityNaming_BATCH implements Database.Batchable<sObject>,
...
-    private Boolean shouldChainNextBatch() {
-        if (lastOppIdProcessed == null) {
-            return false;
-        }
+    /**
+     * @description LDV Chunk Size to use
+     * @return Integer chunk size to use when running in LDV Chunking mode
+     */
+    public override Integer getLDVChunkSize() {
+        return 10000000;
```
force2b: Can you make this one a constant as well (I think it was previously)?

src/classes/OPP_OpportunityNaming_BATCH_TEST.cls
```diff
@@ -67,137 +68,24 @@ private with sharing class OPP_OpportunityNaming_BATCH_TEST {
...
-    private static void shouldChainNextBatchOffsetByLastRecordProcessed() {
-        List<Opportunity> testOpps = [SELECT Id FROM Opportunity ORDER BY Id];
-
-        OPP_OpportunityNaming_BATCH batch = new OPP_OpportunityNaming_BATCH();
-        batch.lastOppIdProcessed = testOpps[0].Id;
 
         Test.startTest();
-        batch.finish(null);
+        String chron = '0 0 23 * * ?';
+        System.schedule('Test Sched', chron, batch);
```
force2b: Does this test method need an Assert?

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,521 @@
...
+    @testVisible private Boolean ldvMode = null;
+    @testVisible private String additionalWhereClause;
+
+    /**
+     * @description No argument constructor that assumes default batch size of 500, and that ldvMode will be
+     * determined by record distance between min and max Id for main object of batch
+     */
+    public UTIL_AbstractCommon_BATCH() {
+        // batchSize, ldvMode, additionalWhereClause
+        this(DEFAULT_BATCH_SIZE, false,'');
```
force2b: Does the `additionalWhereClause` value have to be passed as an empty string instead of a null?

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -0,0 +1,521 @@
...
+        return 'SELECT Id FROM ' + getSObjectName() + ' LIMIT 0';
+    }
+
+    /**
+     * @description Determines the query to execute factoring all the different params for where clause, ldv versus non ldv, etc.
+     * @param startId starting id of the chunk
+     * @return String queryString to execute for Batch
+     */
+    @testVisible private String getQuery(Id startId) {
+        return 'SELECT ' + getQueryFieldList() + ' FROM ' + getSObjectName() + getQueryWhereClause() + getQueryOrderByAndLimitClause();
```
force2b: Rather than have to add a blank space before every occurrence of "WHERE" in the code below, why not just add the space into this query builder so it looks like:
```
'SELECT ' + getQueryFieldList() + ' FROM ' + getSObjectName() + ' ' + getQueryWhereClause() + ' ' + getQueryOrderByAndLimitClause();
```

src/classes/UTIL_AbstractCommon_BATCH.cls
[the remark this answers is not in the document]
force2b: I see your point.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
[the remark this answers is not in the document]
force2b: Might be because the Abstract class implements Batchable and Schedulable, and those cannot be implemented in an Inner class. What about moving this into a different class - `UTIL_AbstractCommon_MOCK extends UTIL_AbstractCommon_BATCH` and then keep just the test methods in this class?

rponti-sforg:
@jbrandolinisf @beAGoodQE  Looks like this needs a kick or further investigation.

rponti-sforg:
> @jbrandolinisf @beAGoodQE Looks like this needs a kick or further investigation.

@davidmreed @davisagli Adding you as well.  I went ahead and clicked the 'Update branch' button which kicked off another build.  This is a WI / PR we would like to include in NPSP 3.161.  

I apologize for the extreme delay.  I assumed Brian Kayser was going to Close the WI and merge the PR.  I did not see his message via Quip that he was going to let me do that.  I had to pick my son up and did not see his message until I came back.  Lines of communication crossed.

PR #4342, opened 2019-06-24
Duplicate GAU Allocation Bug

KietTruongSFDO:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -547,9 +548,13 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
                 (sObj.get('CampaignId') != null && mapWrapper.containsKey((Id)sObj.get('CampaignId'))))) {
 
                 alloWrapper wrap;
+                Map<Id, Allocation__c> GAUtoAlloMapping;
                 //get wrapper for parent Payment.
-                if (sObj instanceof npe01__OppPayment__c)
+                if (sObj instanceof npe01__OppPayment__c) {
                     wrap = mapWrapper.get((Id)sObj.get('npe01__Opportunity__c'));
+                    GAUtoAlloMapping = new Map<Id, Allocation__c>();
```
jbrandolinisf: This would be easier to follow (further down), if you set an indicator like isPaymentAllocation.  Checking for null/not null for the GAUtoAlloMapping made the intent less clear.

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -585,22 +596,60 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
+                            }
                         }
                         allo.put('CurrencyIsoCode', sObj.get('CurrencyIsoCode'));
                     }
 
                     if (allo.Percent__c != null) {
                         allo.Amount__c = (amount * allo.Percent__c * .01).setScale(2);
+                        
+                        if (alloInMap != null) {
+                            alloInMap.Amount__c += allo.Amount__c;
```
jbrandolinisf: The way you have this and the one on 601  laid out, you will inflate the GAU amount (more than it should be) if it is a multicurrency org, because both sections will be entered.  Should account for the one on 621, as well, to make sure it does not duplicate in either case.  Btw, you need to make sure the isMultiCurrency stuff carries over to your Consolidated GAUs as well.

src/classes/ALLO_Allocations_TDTM.cls
jbrandolinisf: Would also be fine with the check sObj instanceof npe01__OppPayment__c in all cases, as this is clearer.

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -188,6 +188,49 @@ private with sharing class ALLO_Allocations_TEST {
         System.assertEquals(8, queryPmts[0].npe01__Payment_Amount__c, 'The payment amount should remain unchanged after the exception is caught.');
     }
 
+    /*******************************************************************************************************
+    * @description When changing the amount of a Payment with Allocations:
+    * Increasing the amount will adjust the percentage allocations.
+    * Decreasing the amount so that the total allocation amount is below the payment amount will adjust
```
jbrandolinisf: Does not reflect actual test.

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -577,30 +590,64 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
                     }
 
+                    if (allo.Percent__c != null) {
+                        allo.Amount__c = (amount * allo.Percent__c * .01).setScale(2);
+                        
+                    }
+
+                    //Calculate proportional amount for non-percentage based Payment allocations
+                    if (sObj instanceof npe01__OppPayment__c && allo.Percent__c == null) {
+                        allo.Percent__c = allo.Amount__c / wrap.parentAmount * 100;
```
jbrandolinisf: I believe it should not calculate the Percentage for the Default GAU.  If you have to do it here, for some reason, we will need to make sure we remove the percentage for the default.  I looked the code over again, but it didn't seem like it was removing the Default GAU from the listAlloForInsert.

KietTruongSFDO: This doesn't calculate the percentage for the Default GAU. The Default GAU is not included in the listAlloForInsert. The listAlloForInsert is a clone of the list of Allocations in the mapWrapper, which excludes the default GAU when being populated in the "fillMapWrapper".

src/classes/ALLO_PaymentSync_SVC.cls
```diff
@@ -378,24 +397,43 @@ public class ALLO_PaymentSync_SVC {
...
+                            //we want to keep the ratios amongst duplicate opportunity allocations
+                            Decimal duplicateAllocationPaidUnpaidAmountRatio = oppAllocation.Amount__c / totalOpportunityAllocations;
+                            System.debug(loggingLevel.ERROR, 'duplicateAllocationPaidUnpaidAmountRatio: ' + duplicateAllocationPaidUnpaidAmountRatio);
 
-                        oppAllocation.Amount__c = newGAUAmtForOpp * duplicateAllocationpaidUnpaidAmountRatio;
-                        if (oppAllocation.Percent__c != null) {
-                            oppAllocation.Percent__c = (context.Opportunity.Amount != 0) ? oppAllocation.Amount__c / context.opportunity.Amount * 100 : 0;
-                        }
+                            // Not sure why we are multiplying by the duplicateAllocationPaidUnpaidAmountRatio??
+                            // oppAllocation.Amount__c = newGAUAmtForOpp * duplicateAllocationpaidUnpaidAmountRatio;
```
KietTruongSFDO: @jbrandolinisf  Why are we multiplying by the duplicateAllocationPaidUnpaidAmountRatio?

KietTruongSFDO:
Debugs will be cleaned up afterwards

src/classes/ALLO_PaymentSync_SVC.cls
jbrandolinisf: Basically, it is determining what percentage of the pie the GAU is of the total amount for that GAU.  It uses this to preserve the ratios of the allocations on the Opportunity if there are duplicates.  Where there are not duplicates, it is 1.  For example, I have 2 allocs for GAU 1.  One is for 25, and the other is for 75.  

Now, lets say the Allocations on the Payments for GAU 1 are doubled, and now total 200, but the Opportunity is not over or under scheduled.

When it processes the first entry for GAU 1, it will be come 200 * .25 = 50.  When it processes, the second entry in the opportunity allocations for GAU 1, it will be come 200 * .75 = 150.  150 + 50 = 200....

This works for the following reasons:

1) It is calculating the totalAmountAllocatedFor that GAU, before updating it, so it can use it as a ratio with the existing amount.
2) We are determining the total amount for a GAU across the paid and unpaid allocations.

Hope this helps to clarify what it is/should be doing and why.  If you believe it needs to change, we should probably talk it through.  This approach would continue to maintain both allocations on Opportunity, using their original ratios to decide how to split the updated amount.

KietTruongSFDO:
Hi @markross-sfdo  and @lianeyee, can you guys take a look at the change notes real quick?

PR #4343, opened 2019-06-25
new test case for verifying payment allocations is disabled

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
+++ b/robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
@@ -1,0 +1,17 @@
+*** Settings ***
+
+Resource        robot/Cumulus/resources/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Verify Payment Allocations is Disabled by Default
+    [tags]  unstable
+    Populate Field By Placeholder    Quick Find    Custom Settings
+    Click Element With Locator    custom_settings.subtree    Custom Settings
+    Select Frame And Click Element    Custom Settings    custom_settings.link    Allocations Settings    Manage
+    Choose Frame    Custom Setting Allocations Settings
+    Wait Until Element Is Visible  text:Default Organization Level Value
+    Checkbox Status    Payment Allocations Enabled    Not Checked
+    
--- a/robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
+++ b/robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
@@ -7,10 +7,11 @@ Suite Teardown  Delete Records and Close Browser
 *** Test Cases ***
 
 Verify Payment Allocations is Disabled by Default
-    [tags]  unstable
-    Populate Field By Placeholder    Quick Find    Custom Settings
-    Click Element With Locator    custom_settings.subtree    Custom Settings
+    [tags]  stable
+    Go To Setup Page    CustomSettings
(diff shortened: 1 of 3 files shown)


robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
```diff
@@ -0,0 +1,17 @@
...
+
+Resource        robot/Cumulus/resources/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Verify Payment Allocations is Disabled by Default
+    [tags]  unstable
+    Populate Field By Placeholder    Quick Find    Custom Settings
```
davidmreed: Question: There's a direct link available to the Custom Settings home, `/lightning/setup/CustomSettings/home`. Would it be more stable to navigate via relative URL rather than manipulating Quick Find?

robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
```diff
@@ -0,0 +1,17 @@
+*** Settings ***
+
+Resource        robot/Cumulus/resources/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Verify Payment Allocations is Disabled by Default
+    [tags]  unstable
```
davidmreed: This test passed on the build I am looking at. Do we have any way to identify the source of any instability and see if we can get it to stable?

skristem: the test is stable as far as I see but I always mark new tests as unstable to be on safe side to not let dev's panic. What do u suggest

davidmreed: Well, I'm just concerned about your hard work on new tests being put into a box called "unstable" where we don't care about, and maybe don't even look at, the results because they're not stable. Is there a process for reviewing unstable tests and moving them to stable if they are working consistently?

I am definitely not a Robot expert, but my general thought is that if we think a test works consistently based on what we can see, let's treat it as a real, stable test so that we get value from its results, and if we see that it's unstable then let's change it. But I'm open to being persuaded by the Robot team that that's not the right framework to apply here. Any thoughts, @Safiyab, or @boakley?

skristem: yes I agree with you and that's why I had brought up this topic couple of times in our robot meetings but we never came to a solution. But for now can u plz approve it so that I can merge for regression tomorrow? and later we based on what we decide I can change the status accordingly

boakley: I think the `unstable` tag is good as a temporary fix while a test is being developed, but there's no point in keeping unstable tests around. There's not much point in having a test we can't rely on. 

I don't think we have any sort of official policy about this, but we probably need to. For example, we may want to decide it's ok to push unstable tests on a branch, but a branch should never be merged into master if it has unstable tests.

davidmreed: I will go ahead and approve to let the QE team move this into operation, since it clearly does no harm, but let's definitely continue discussing this subject and pushing to get Robot tests marked as stable.

robot/Cumulus/tests/browser/npsp_settings/verify_payment_allocations_setting.robot
skristem: changed

PR #4352, opened 2019-06-26
PYALLOCB: 1-B : Update Error Object and Err_Handler class

jbrandolinisf:
# Critical Changes

# Changes
- We added two new Error fields, Record_Id__c and Retry_Pending__c, as well as additional background functionality. Together, these changes support future development of error logging functionality.   
# Issues Closed

# New Metadata
Error (Error__c) Object:
- Retry Pending? (Retry_Pending__c [Checkbox]) Added for future support.  If true, will indicate the associated process and record needs to be reprocessed.  After successful reprocessing, it will change to false [unchecked].  This field is not required.
- Record Id (Record_Id__c [Text(18)] Added to relate the Error to a specific record in the system.    This field is not required.
# Deleted Metadata


jbrandolinisf:
@force2b - just give me a rough, yeah or heck no...and we can talk.  I made most of the changes inside of a new inner class called RecordErrors.

src/classes/ERR_Handler.cls
```diff
@@ -420,15 +864,29 @@ public class ERR_Handler {
...
+
+    /*******************************************************************************************************
+    * @description Returns the URL of a record, if available.
+    * @param recordId The id of the record associated with the error
+    * @return The URL of the record. Warning: it may not necessarily be valid.
+    */
+    private static String getRecordURL(String recordId) {
+        String recordURL = URL.getSalesforceBaseUrl().toExternalForm() + '/' + recordId;
+
+        if (recordURL.startsWith('http:'))
```
force2b: Curly brackets are needed, though I'm really sure the `if` is needed for this.

src/classes/ERR_Handler.cls
```diff
@@ -449,10 +907,10 @@ public class ERR_Handler {
     */
     private static String getStatusCode(system.StatusCode status) {
         if (status == system.Statuscode.REQUIRED_FIELD_MISSING)
-            return 'Required Field Missing';   
+            return 'Required Field Missing';
         else if (status == system.Statuscode.FIELD_CUSTOM_VALIDATION_EXCEPTION)
             return 'Custom Validation Rule Failed';
         else
-            return 'Other';
+            return STATUS_CODE_OTHER_VAL;
```
force2b: Curly brackets needed

src/classes/ERR_Handler.cls
```diff
@@ -53,6 +55,273 @@ public class ERR_Handler {
         public List<Error__c> errorRecords = new List<Error__c>();
     }
 
+    /*******************************************************************************************************
+    * @description Inner class that groups the result of processing the result of multiple DML operations performed
+    * using the Database class, as well as manages Errors by Record Id (Salesforce Id of the record most directly
+    * associated with the operation).
+    */
+    public class RecordErrors {
```
force2b: I'm unclear on this inner class. I can't find any references to it being instantiated in this code. Two things come to mind though:
1. Class names are generally singular (`RecordError')
2. I wonder if there is enough code in this class that it needs to be abstracted out into another class such as `ERR_RecordError`.

For #2, I'm not really able to tell how this is used so it may be that it doesn't need to move. I'm just not sure of the connection between the existing methods in the ERR_Handler and this inner class.

src/classes/ERR_Handler.cls
jbrandolinisf: This method was there originally.  I can update it  accordingly.

src/classes/ERR_Handler.cls
jbrandolinisf: Another method that was here originally...I just refactored to use constants...I debated converting to Custom Labels but thought I would see what you thought...

src/classes/ERR_Handler.cls
jbrandolinisf: I am good with extracting the code out.  This is the kind of feedback I was particularly interested in.  I initially tried to stay with methods similar to what is in ERR_Handler, but I really needed stuff to handle/abstract the insert/update aspects of keeping a unique Record Error.

Should I make that change then?

src/classes/ERR_Handler.cls
force2b: Yes, sorry. If you're doing some significant work in the class and/or modifying methods, it's worth fixing legacy style errors.

force2b:
@jbrandolinisf Also, the `.cci/snapshot/dev.json` file should be removed from this PR.

src/classes/ERR_Handler.cls
force2b: My gut says yes. It's enough of a different use-case than the rest of the methods that I think it's worth separating concerns.

jbrandolinisf:
@force2b - Ready for Review 2

jbrandolinisf:
@force2b - I rolled most of the updates in ERR_Handler back, but there were three lines I felt I had to update, given I was adding Custom Labels for that text.  So I did update that since I was adding labels for those values....I stayed away from touching anything else in the class though...

src/classes/ERR_RecordError.cls
```diff
@@ -0,0 +1,536 @@
...
+        return updatedMessage;
+    }
+
+    /*******************************************************************************************************
+    * @description Returns the URL of a record, if available.
+    * @param recordId The id of the record associated with the error
+    * @return The URL of the record. Warning: it may not necessarily be valid.
+    */
+    private static String getRecordURL(String recordId) {
+        String recordURL = URL.getSalesforceBaseUrl().toExternalForm() + '/' + recordId;
```
force2b: Isn't this mostly the same as the version in `ERR_Handler`? If so, I recommend updating that method (to clean it up) and using that version rather that duplicating.

src/classes/ERR_RecordError.cls
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+
+    /*******************************************************************************************************
+    * @description Returns the status of the error that happened as a result of a DML operation performed through
+    *              the Database class as a string.
+    * @param status The status code.
+    * @return The error status as a string.
+    */
+    private static String getStatusCode(System.StatusCode status) {
```
force2b: Isn't this also the same as the version in `ERR_Handler`?

src/classes/ERR_RecordError_TEST.cls
```diff
@@ -0,0 +1,586 @@
...
+        System.assertNotEquals('Bar', err.Full_Message__c, 'Full Message');
+        System.assertEquals(true, err.Full_Message__c.contains('Bar'), 'Full Message');
+    }
+
+    /*******************************************************************************************************
+    * @description Unit Test for createRecordError and updateRecordError for invalid undelete operation
+    */
+    @IsTest
+    private static void createAndUpdateDMLRecordErrorsForInvalidUndelete() {
+        List<Account> accs = UTIL_UnitTestData_TEST.createMultipleTestAccounts(1, null);
```
force2b: Thank you for using the data factory methods

src/classes/ERR_RecordError_TEST.cls
```diff
@@ -0,0 +1,586 @@
...
+private class ERR_RecordError_TEST {
+
+    private class TestException extends Exception { }
+
+    /*******************************************************************************************************
+    * @description Unit Test of updateFullMessage, existing message is null, additional test for existing
+    * is blank.
+    */
+    @IsTest
+    private static void updateFullMessageWithoutExistingMessage() {
```
force2b: Would you mind change the method names to something like `shouldUpdateFullMessageWithNullExistingMessage` or `validateUpdateFullMessageWithNullExistingMessage`, etc.? That's more consistent with how we've been naming them in new test classes.

src/objects/Error__c.object
```diff
@@ -144,6 +144,29 @@
...
         <type>LongTextArea</type>
         <visibleLines>10</visibleLines>
+    </fields>
+     <fields>
+        <fullName>Record_Id__c</fullName>
+        <description>The Salesforce Id of the record associated with this error.</description>
+        <externalId>true</externalId>
+        <inlineHelpText>The Salesforce Id of the record associated with this error.</inlineHelpText>
+        <label>Record Id</label>
+        <length>255</length>
```
force2b: Is there a need for this field to be 255 instead of 18? I guess we can't configure a field as indexed unless it's unique and I can see why you wouldn't want this as unique. I recommend updating the Gem docs to include a note about indexing this field or the `Retry_Pending__c` field for performance **if** the batch job starts to time out.

src/objects/Error__c.object
```diff
@@ -144,6 +144,29 @@
         <trackTrending>false</trackTrending>
         <type>LongTextArea</type>
         <visibleLines>10</visibleLines>
+    </fields>
+     <fields>
+        <fullName>Record_Id__c</fullName>
+        <description>The Salesforce Id of the record associated with this error.</description>
```
force2b: Can you have docs review the Label, Description and and InlineHelpText attributes for the two new fields?

src/objects/Error__c.object
```diff
@@ -144,6 +144,29 @@
...
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
+        <fullName>Retry_Pending__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>The record associated with this error will be reprocessed the next time the process the error is associated with executes.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The record associated with this error will be reprocessed the next time the process the error is associated with executes.</inlineHelpText>
```
force2b: Should this state somewhere that it's for Gem reprocessing only (or something that indicates that not every error will support Retrying this way)?

src/classes/ERR_Handler.cls
```diff
@@ -449,10 +449,10 @@ public class ERR_Handler {
     */
```
force2b: Since you've added a new field for `Record_Id__c` I definitely recommend updating this class to populate the new field where possible.

src/classes/ERR_RecordError.cls
```diff
@@ -0,0 +1,536 @@
...
+/**
+* @author Salesforce.org
+* @date 2019
+* @description Helps manage the process of inserting and updating Record centered Errors.
+* @group ErrorHandling
+* @group-content ../../ApexDocContent/ErrorHandling.htm
+*/
+public class ERR_RecordError {
+
+   public class RecordErrorException extends Exception { }
```
force2b: What  do you think about calling this `RecordErrorHandlerException`?

src/classes/ERR_RecordError.cls
```diff
@@ -0,0 +1,536 @@
...
+    * operation performed through the Database class.
+    * @param result The result of the Database DML operation. It cannot be more specific than Object, because the result of
+    *               the different DML operations that the Database class can perform don't have a common parent.
+    * @param objectType The name of the type of object that caused the error.
+    * @param objId The id of the record that caused the error.
+    * @param context The context the exception was raised on.
+    * @return The error record.
+    */
+    @testVisible private static Error__c createRecordError(Object result, String objectType, Id objId, String context) {
+        Error__c error = new Error__c();
```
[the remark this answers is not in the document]
jbrandolinisf: I started there, and found it didn't lend itself well to this.  Keep in mind, sharing the methods will require that I make them public.  If you feel strongly about it, I will refactor for this, but it may make modifications slightly more complicated in the future.

src/classes/ERR_RecordError.cls
```diff
@@ -0,0 +1,536 @@
...
+    * @param objectType The type of all records in the job, only to be passed in if they all match.
+    * @return The error record.
+    */
+    @testVisible private static String updateFullMessage(Datetime ts, String existingMessage, String message) {
+
+        String updatedMessage = ts.format() + ':\r\n' + message;
+
+        if (String.isNotBlank(existingMessage)) {
+            updatedMessage += '\r\n' + '========================================\r\n' + existingMessage;
+            // Limit to 5K worth of Message History
```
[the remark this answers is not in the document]
jbrandolinisf: I was mostly doing this because I was concerned about an aggregate of heap size across several records, as I might be retrieving and updating hundreds of these in a batch depending...  If you don't think I need to worry about it, I will ease the restriction and change it to 32K. Granted there shouldn't be too many where there are large numbers of error records to retrieve/update.

src/classes/ERR_RecordError.cls
jbrandolinisf: If you are ok with changing the visibility on the original, I will do so.

src/classes/ERR_RecordError.cls
jbrandolinisf: If you are ok with changing the visibility on the original, I will do so.

src/objects/Error__c.object
jbrandolinisf: No...That was a screw up.  I will get it changed.  It is (supposed) to be Text(18) External Id (Non unique) which will index it.

The index should be across Category_Type__c, Record_ID__c, and Retry_Pending__c, I think.

src/objects/Error__c.object
jbrandolinisf: Already tagged them in the associated QUIP doc.

src/objects/Error__c.object
jbrandolinisf: I think it should absolutely say something, but I am not certain we should say GEM/Abacus/Payment Allocations only...

src/classes/ERR_Handler.cls
jbrandolinisf: Is it better/worse to populate this field, given that it is an External Id, albeit, non unique?

jbrandolinisf:
@force2b - If you want to comment on my comments, that is great, or we can get some time together.  Anything with a thumbs up emoji is stuff I will just address.

markross-sfdo:
@jbrandolinisf The Changes text is ready for your review. I am reviewing the UI text now.

src/classes/ERR_Handler.cls
force2b: I think it's better to populate the field. Customers would be able to more easily find errors associated with records. I probably would have found a use for this in my consulting days.

src/classes/ERR_Handler.cls
```diff
@@ -440,19 +447,19 @@ public class ERR_Handler {
...
     * @description Returns the status of the error that happened as a result of a DML operation performed through
     *              the Database class as a string.
     * @param status The status code.
     * @return The error status as a string.
     */
-    private static String getStatusCode(system.StatusCode status) {
+    public static String getStatusCode(system.StatusCode status) {
         if (status == system.Statuscode.REQUIRED_FIELD_MISSING)
-            return 'Required Field Missing';   
+            return System.Label.errErrorTypeRequiredFieldMissing;
```
force2b: Might as well add the curly brackets if you're modifying this method. Sorry I missed this one before.

src/classes/ERR_RecordError.cls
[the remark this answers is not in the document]
force2b: Very good point. They're close, but not exactly like this. A fuller refactor might have moved the `ERR_Handler` class into an instance class or a base class that we can inherit from, but I think that's a big haul for not as much gain.

src/classes/ERR_RecordError.cls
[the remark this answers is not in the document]
force2b: Yeah, this is true. If you're retrieving a bunch of records (> 1000) with a 32K field it might become a heap size issue.

src/classes/ERR_RecordError.cls
force2b: I think in this case it might be worth making it public.

src/objects/Error__c.object
force2b: Agreed. We can let docs come up something that makes sense.

force2b: ^^^ @markross-sfdo

jbrandolinisf:
@force2b - Do you have a recommendation on what we should use to test the existing functionality so we can get this merged in?  Thanks for all your help with this!

PR #4364, opened 2019-07-02
Include inactive GAUs when creating allocations for payments. 

KietTruongSFDO:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ report.html
 robot/Cumulus/results/
 datasets/dev_org/test_data.db
 *.db
+.cci
--- a/orgs/trial.json
+++ b/orgs/trial.json
@@ -1,6 +1,6 @@
 {
   "orgName": "NPSP - Trial Org",
-  "template": "0TT1I000006uGTc",
+  "template": "0TT1I000006q2fM",
   "settings": {
     "orgPreferenceSettings": {
       "chatterEnabled": true,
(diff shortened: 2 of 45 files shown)


src/objects/Allocation__c.object
```diff
@@ -86,12 +86,23 @@
         <label>General Accounting Unit</label>
         <lookupFilter>
             <active>true</active>
+            <booleanFilter>(1 AND 2) OR 3</booleanFilter>
```
jbrandolinisf: @KietTruongSFDO - My only question on this was why was it necessary to update the Allocation object as part of this?  Is this needed?  If so, why?

jbrandolinisf: Actually, I get it...this was there to block it.  I need to review this with Pedro...just making it work with Payments, only partly fixes it.  If we cannot update the Opportunity Allocations, it will not work when it tries to finalize it.

jbrandolinisf: @pedrodalcol and I discussed this item.  We will need to do further work on this when he returns, since we will error when attempting to update the opportunity allocations for payment allocations changes (when there is an inactive GAU present), but we agreed that was a) not a reason to hold this up, and b) erroring for now, while we figure out what we really want to have happen is probably ok.

PR #4370, opened 2019-07-05
Prevent GAU/RD delete if DSO archive data exists

seanpat09:
# Critical Changes

# Changes
Prevents deletion of GAUs and RDs if they are related to ArchiveKey data in DSO
# Issues Closed

# New Metadata
Apex Classes
* GAU_TDTM.cls
* GAU_TDTM_TEST.cls
* UTIL_IntegrationGateway.cls
* UTIL_IntegrationGateway_TEST.cls

Apex Triggers
* TDTM_GeneralAccountingUnit.trigger

# Deleted Metadata


seanpat09:
**lurch: attach W-033805

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087vgZQAQ" target="_blank">W-033805</a>

src/classes/GAU_GeneralAccountingUnits_TDTM_TEST.cls
```diff
@@ -0,0 +1,72 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group General Accounting Units
+* @description Test class for the GAU_GeneralAccountingUnits_TDTM class
+*/
+@isTest
+private with sharing class GAU_GeneralAccountingUnits_TDTM_TEST {
+    @isTest
+    private static void shouldDispatchGAUsToIntegrationGatewayOnBeforeDelete() {
+        List<General_Accounting_Unit__c> triggerOld = new List<General_Accounting_Unit__c> {
```
RadSFDeveloper: Suggestion: How about `gaus` instead `triggerOld` to indicate what the var is?

src/classes/GAU_GeneralAccountingUnits_TDTM.cls
```diff
@@ -0,0 +1,64 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group General Accounting Units
+* @description TDTM trigger handler for General_Accounting_Unit__c
+*/
+public with sharing class GAU_GeneralAccountingUnits_TDTM extends TDTM_Runnable {
```
RadSFDeveloper: Can we rename the class to indicate the processing or just name it `GAU_TDTM` since `GeneralAccountingUnits` is expended GAU?

src/classes/RD_RecurringDonations_TEST.cls
```diff
@@ -198,8 +198,42 @@ public class RD_RecurringDonations_TEST {
         installments3 = getOpps(r2.id);
         System.assertEquals(false,installments3[0].IsDeleted, 'Record should be undeleted.');
     }
-    
-    
+
+    @isTest
+    private static void shouldDispatchRDsToIntegrationGatewayOnBeforeDelete() {
+        List<npe03__Recurring_Donation__c> triggerOld = new List<npe03__Recurring_Donation__c> {
```
RadSFDeveloper: Suggestion: `rds`, or `recurringDonations` instead of `triggerOld`?

src/classes/UTIL_IntegrationGateway.cls
```diff
@@ -0,0 +1,81 @@
...
+            if (listeners == null) {
+                listeners = new List<UTIL_IntegrationConfig>{
+                    archiveBridgeConfig
+                };
+            }
+            return listeners;
+        }
+        set;
+    }
+
```
RadSFDeveloper: ApexDoc description missing for methods in this class and other classes and tests.

src/classes/UTIL_IntegrationGateway.cls
```diff
@@ -0,0 +1,81 @@
...
+    public void handleBeforeDelete(List<SObject> triggerOld, Map<Id,SObject> triggerOldMap) {
+        Map<String, Object> params = new Map<String, Object> {
+            PARAM_TRIGGER_OLD => triggerOld,
+            PARAM_TRIGGER_OLD_MAP => triggerOldMap
```
RadSFDeveloper: Can we just pass in the `triggerOld` and remove `triggerOldMap`? Any potential issue with that?

seanpat09: I think that should be fine, the TDTM framework doesn't seem to use `trigger.oldMap` either. I will update

seanpat09:
@RadSFDeveloper I've made the requested changes, can you review again? I also ran into the issue where GAU wasn't being marked as deleted even with the custom settings set up correctly. It turned out that since the GAU trigger is new, I didn't have a `Trigger_Handler__c` record for the GAU trigger, so the code wasn't firing on before delete. When I created a new scratch org, when the default `Trigger_Handler__c` records from `TDTM_DefaultConfig.cls` created it and the validation worked

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -352,6 +352,10 @@ public without sharing class TDTM_DefaultConfig {
                 Class__c = 'CRLP_Rollup_TDTM', Load_Order__c = 2, Object__c = 'npe01__OppPayment__c',
                 Trigger_Action__c = 'AfterInsert;AfterUpdate;AfterDelete;AfterUndelete'));
 
+        handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false,
```
seanpat09: @force2b @RadSFDeveloper 

Since we're adding a new trigger for GAU and `BeforeDelete` for RDs, what's the best way to make sure that customers that are already have TDTM installed get these records updated correctly? Should we query for these records in the Post Install class and update if they exist?

force2b: There is already a process for this in the `STG_InstallScript.insertTdtmDefaults()` method.

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -182,7 +182,7 @@ public without sharing class TDTM_DefaultConfig {
         // RecurringDonations on RecurringDonations
         handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false,
               Class__c = 'RD_RecurringDonations_TDTM', Load_Order__c = 1, Object__c = 'npe03__Recurring_Donation__c',
-              Trigger_Action__c = 'BeforeInsert;BeforeUpdate;AfterInsert;AfterUpdate'));
+              Trigger_Action__c = 'BeforeInsert;BeforeUpdate;AfterInsert;AfterUpdate;BeforeDelete'));
```
force2b: The list of actions needs be in alphabetical order - only for the reason of how the logic that compares the defaults to the current records works.

src/classes/UTIL_IntegrationGateway.cls
```diff
@@ -0,0 +1,85 @@
...
+     * @description Passes trigger records to listener integrations
+     * @param triggerOld records that were passed to the trigger as trigger.old
+     */
+    public void handleBeforeDelete(List<SObject> triggerOld) {
+        Map<String, Object> params = new Map<String, Object> {
+            PARAM_TRIGGER_OLD => triggerOld,
+            PARAM_TRIGGER_OLD_MAP => new Map<Id, SObject>(triggerOld)
+        };
+
+        for (UTIL_IntegrationConfig listener : listeners) {
```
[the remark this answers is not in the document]
seanpat09: Good point, I didn't take that into consideration. Luckily, the bridge delegates the query to DSO, which runs in a class that has `without sharing`, so we should be good.

seanpat09:
@RadSFDeveloper @force2b I've made the requested changes, thanks again for the feedback! Let me know if there's anything else you notice

src/classes/UTIL_IntegrationGateway.cls
RadSFDeveloper: @seanpat09 , change to the `triggerOldMap` have not been done yet. Do you think the suggestion is outdated and the param is still required? Everything else looks fine. If there are changes to these params, the bridge side needs to be changed too.

seanpat09: I'm not sure how I missed this, I'll get this updated!

src/classes/UTIL_IntegrationGateway.cls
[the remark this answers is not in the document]
force2b: Perfect. Thanks for checking.

src/classes/UTIL_IntegrationGateway.cls
```diff
@@ -0,0 +1,85 @@
...
+        set;
+    }
+
+    /*******************************************************************************************************
+     * @description Passes trigger records to listener integrations
+     * @param triggerOld records that were passed to the trigger as trigger.old
+     */
+    public void handleBeforeDelete(List<SObject> triggerOld) {
+        Map<String, Object> params = new Map<String, Object> {
+            PARAM_TRIGGER_OLD => triggerOld,
```
force2b: Is it necessary to pass both the Map and the List into the method? This could send just send the map (that is all that is needed in the Bridge api method). For eventual other implementations, those methods can easily use `.values()` to get the list version. Mainly I'm just looking at reducing the size of the payload.

src/classes/UTIL_IntegrationGateway.cls
```diff
@@ -0,0 +1,81 @@
...
+        set;
+    }
+
+    /*******************************************************************************************************
+     * @description Passes trigger records to listener integrations
+     * @param triggerOld records that were passed to the trigger as trigger.old
+     */
+    public void handleBeforeDelete(List<SObject> triggerOld) {
+        Map<String, Object> params = new Map<String, Object> {
+            PARAM_TRIGGER_OLD => triggerOld
```
seanpat09: @RadSFDeveloper removed the Trigger.oldMap parameter, thanks for catching that! I also needed to change the signature in the DSO-NPSP callable API, I'll tag you in that PR  too

src/classes/UTIL_IntegrationGateway.cls
seanpat09: I ended up just using the List into the method instead so that order is preserved if needed, and it can also be converted into a map if necessary.

rponti-sforg:

For further testing information, see: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E0000087vgZQAQ/view

PR #4372, opened 2019-07-08
TDTM Mutable

force2b:
# Context

TDTM allows for users to create execution chains of logic, however customer implementations of TDTM are unable to interact with any aspects of the `TDTM_Runnable.DmlWrapper` collection that will have inserts, updates, and deletes performed. This creates a critical issue where customer implementations of TDTM must workaround any situation where NPSP's TDTM implementations have already added a record into the `objectsToUpdate` collection; not handling this will present a duplicate record in collection exception, and handling this often means leveraging asynchronous updates or performing DML separately from `TDTM_TriggerHandler`. Furthermore, any customer implementations of TDTM might break if NPSP updates its TDTM logic to start adding a record into an `objectsToUpdate` (which was not encountered when the customer implementation was initially designed and deployed).

This PR enables NPSP and customer implementations of TDTM to gain access to the global `DmlWrapper` that `TDTM_TriggerHandler` maintains during a transaction. The feature is considered "opt in" because it relies upon an entirely new interface to implement.

**Example:**

- Customer wishes to set fields on Account when a Contact is created in the `AfterInsert` context
- Customer adds the Contact's Account into `DmlWrapper.objectsToUpdate` and returns this from their `TDTM_Runnable.run` implementation
- When DML is performed, Salesforce throws an exception because an NPSP TDTM implementation had already added that Account into `DmlWrapper.objectsToUpdate`
- Customer updates their class to implement `TDTM_RunnableMutable`, checks whether the Account retrieved is part of `DmlWrapper.objectsToUpdate`, and either updates that record or places the Account retrieved into that collection

**Benefits:**

- Existing TDTM implementations will not break, as any existing classes implement the `TDTM_Runnable` interface
- Classes implementing the `TDTM_RunnableMutable` interface can be smaller implementations of code, as developers can reliably retrieve a record that was processed in a prior context from the transaction's global `DmlWrapper`
- Customers are able to adopt TDTM for use cases that were previously managed with a different framework or workaround

**Risks:**

- Mutating the transaction's `DmlWrapper` collection could remove records for insertion or update, or overwrite previous changes by putting a record into the collection that does not reflect changes made from TDTM logic fired previously in transaction

# Critical Changes

# Changes

* Introduced a new global `TDTM_RunnableMutable` interface to support passing a mutable instance of the `DmlWrapper` to the `run` method. Customer TDTM classes that implement this new interface can include a new `DmlWrapper` parameter in the `run()` method of their TDTM class to support handling duplicate record updates. Thanks to @rogeramitchell, @njjc, and @matthewpoe for their work on design and implementation of this new feature. More information on using the new Mutable interface can be found at https://powerofus.force.com/s/article/NPSP-Deploy-Custom-Apex-TDTM#ariaid-title7 and https://github.com/rogeramitchell/tdtm-mutable-example

# Issues Closed

* Hub Idea: [Add method to update an item in the DMLWrapper collection](https://powerofus.force.com/hub-ideas/1386)

# New Metadata

**Apex Classes**
* TDTM_RunnableMutable
* TDTM_RunnableMutableExample_TEST
* TDTM_RunnableMutable_TEST


force2b:
**lurch: attach W-030244

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007fQ7DQAU" target="_blank">W-030244</a>

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -153,63 +145,121 @@ public class TDTM_TriggerHandler {
...
        classToRun.runFutureNonStatic(setNewId, setOldId, thisAction.name(), describeObj.getName(), classToRunName);
     }
+
+    /**
+     * @description Combine the DmlWrapper instance returned from an individual TriggerHandler run() call into
+     * the DmlWrapper instance for the entire TriggerHandler.run() method.
+     * @param dmlWrapper The shared instance
+     * @param dmlWrapperLocal The single instance returned from a TriggerHandler call.
+     */
+    private static void updateDmlWrapperWithTriggerResponse(TDTM_Runnable.DmlWrapper dmlWrapper, TDTM_Runnable.DmlWrapper dmlWrapperLocal) {
```
seanpat09: Nice pulling this out into it's own method, small incremental clean up like this goes a long way!

force2b:
Thanks @seanpat09. As per @rogeramitchell:

> Yes, the duplicate ID issue can still occur based on load order. The assumptions are that it can be mitigated with this new interface, and also that there are practically zero implementations today that would face this issue because it would have presented itself regardless of the new interface. 

My expectation was that customers who want to implement the Mutable interface to manipulate the DmlWrapper would tend to put their custom triggers at the end of the load order. However, an SI could really apply this anywhere and make it work if they're comfortable with the way that the triggers are fired.

PR #4374, opened 2019-07-10
WIP - W-033110 Opportunity Amount Sync

weizsforg:
# Critical Changes

# Changes

If Enforce Accounting Data Consistency is enabled, it will:
- If Opportunity is Finalized: Prevent changes to the Opportunity Amount that differ from the Sum of Paid Payments 
- If Opportunity is Not Finalized: Prevent the Opportunity Amount from being lowered below the Sum of the Paid Payments
- If the Opportunity is not Finalized: Increase the Opportunity Amount to the Sum of the Paid Payments, as changes to Payments occur
- If the Opportunity becomes Finalized (e.g. the Opportunity Stage is changed to a Finalized State): Change the Opportunity Amount to the sum of the Paid Payments, raising OR lowering it.  It will continue to be updated/recalculated if additions or changes to the Paid Payments occur

Definition of Finalized:  A closed Opportunity Stage with a corresponding Opportunity Stage to State Mapping of "Finalized".

# Issues Closed
Fixes [#4461](https://github.com/SalesforceFoundation/Cumulus/issues/4461)

# New Metadata
Custom Labels: 
- Opp_Amount_Is_Less_Error
- Opp_Amount_Not_Equal_Error

# Deleted Metadata


jbrandolinisf:
This one is on hold for the moment.

src/classes/OPP_OpportunityAmountSync_SVC.cls
```diff
@@ -0,0 +1,32 @@
...
+        for (Opportunity opp : opportunitiesByIds.values()) {
+            Decimal totalPaid = 0;
+
+            for (npe01__OppPayment__c pmt : opp.npe01__OppPayment__r) {
+                if (pmt.npe01__Paid__c) {
+                    totalPaid += pmt.npe01__Payment_Amount__c == null ? 0 : pmt.npe01__Payment_Amount__c;
+                }
+            }
+
+            Boolean addOppToList = (OPP_StageMappingUtil.determineOppStateFromOppStage(opp.StageName) == 'FINALIZED' && opp.Amount != totalPaid) ||
```
jbrandolinisf: I asked @KietTruongSFDO in his related PR to make a method in OPP_StageMappingUtil called either isFinalized or isOpportunityFinalized, depending on whether the Opportunity or the stage is passed (I originally indicated I don't care which, but in Abacus when we did this, we also checked the Opportunities isClosed field, and if it was not a Closed stage, we threw an exception.  I believe we should do the same here [though I could be convinced we should do this later if we are trying to get it in today]).  Can you work with him and agree on that, and get this in there?

src/classes/OPP_OpportunityAmountSync_SVC.cls
```diff
@@ -0,0 +1,32 @@
...
+                    (OPP_StageMappingUtil.determineOppStateFromOppStage(opp.StageName) != 'FINALIZED' && opp.Amount < totalPaid);
```
jbrandolinisf: If we do not move this into the Util, we should reference the constant in the Util class for Finalized.  You might have to change the visibility of the constants, if you do this...

src/classes/OPP_OpportunityAmountSync_SVC.cls-meta.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>44.0</apiVersion>
```
jbrandolinisf: Should be 45.0 at this point, I believe.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +159,33 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+            for (Opportunity opp : oppsById.values()) {
+                Decimal totalPaid = 0;
+
+                for (npe01__OppPayment__c pmt : opp.npe01__OppPayment__r) {
+                    if (pmt.npe01__Paid__c) {
+                        totalPaid += pmt.npe01__Payment_Amount__c == null ? 0 : pmt.npe01__Payment_Amount__c;
+                    }
+                }
+
+                if (OPP_StageMappingUtil.determineOppStateFromOppStage(opp.StageName) == 'FINALIZED'
```
jbrandolinisf: I assume this came from Kiet's work item...

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +159,33 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+                        && mapIdOppNew.get(opp.Id).Amount != totalPaid) {
+                    opp.addError('The Opportunity Amount must be equal to total paid Payment Amount amount!');
```
jbrandolinisf: Convert to Custom Label

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +159,33 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+                }
+
+                if (OPP_StageMappingUtil.determineOppStateFromOppStage(opp.StageName) != 'FINALIZED'
+                        && mapIdOppNew.get(opp.Id).Amount < totalPaid) {
+                    opp.addError('You are not allow to lower the Opportunity Amount to less than total Paid Payment amount!');
```
jbrandolinisf: Convert to Custom Label

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -212,7 +267,7 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
             oppIds.add(pmt.npe01__Opportunity__c);
         }
 
-        Map<Id, Opportunity> oppsById = getPaymentOpportunities(oppIds);
+        oppsById = getPaymentOpportunities(oppIds);
```
jbrandolinisf: Make sure we call out that they need to retest Multicurrency functionality, given that we changed (albeit minor) stuff here.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -837,6 +837,90 @@ private class PMT_Payment_TEST {
     }
 
 
+    /**
+     * @description Check that the opportunity amount is updated when the payed payments amounts
+     * exceed the opportunity amount
+     */
+    @isTest
+    private static void testOpportunityAmountSync() {
+        UTIL_CustomSettingsFacade.getContactsSettingsForTests(new npe01__Contacts_And_Orgs_Settings__c (npe01__Payments_Enabled__c = false, Enforce_Accounting_Data_Consistency__c = true));
```
jbrandolinisf: Why is Payments Enabled false?

src/classes/PMT_Payment_TEST.cls
```diff
@@ -837,6 +837,90 @@ private class PMT_Payment_TEST {
...
+                        npe01__Payment_Date__c = System.today()
+                )
+        };
+        insert payments;
+
+        Test.stopTest();
+
+        // check the opp amount
+        opp = [select Id, Amount from Opportunity where Id = :opp.Id];
+        System.assertEquals(20, opp.Amount, 'Finalized opportunity amount not updated when paid amounts are less than opportunity amount.');
```
jbrandolinisf: So we decided not to ever lower the Opp Amount, even if finalized? @pedrodalcol is that correct?

src/classes/OPP_OpportunityAmountSync_SVC.cls-meta.xml
force2b: You can go up to 46 now too. We're advising that all the new BDI classes be set at 46 as well.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +159,33 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+            Set<Id> oppIds = (new Map<Id, Opportunity>(newList)).keySet();
+
+            Map<Id, Opportunity> oppsById = getPaymentOpportunities(oppIds);
+
+            for (Opportunity opp : oppsById.values()) {
+                Decimal totalPaid = 0;
+
+                for (npe01__OppPayment__c pmt : opp.npe01__OppPayment__r) {
+                    if (pmt.npe01__Paid__c) {
+                        totalPaid += pmt.npe01__Payment_Amount__c == null ? 0 : pmt.npe01__Payment_Amount__c;
```
jbrandolinisf: why not make this 

totalPaid += (pmt.npe01__Paid__c && pmt.npe01__Payment_Amount__c != null) ? pmt.npe01__Payment_Amount__c : 0;

This eliminates the if, and puts are affirmative choice first.

src/classes/OPP_OpportunityAmountSync_SVC.cls
```diff
@@ -0,0 +1,32 @@
...
+     * @return
+     */
+    public List<Opportunity> syncOppAmountWithTotalPaidPaymentAmount(Map<Id, Opportunity> opportunitiesByIds) {
+        List<Opportunity> oppsToUpdate = new List<Opportunity>();
+        
+        for (Opportunity opp : opportunitiesByIds.values()) {
+            Decimal totalPaid = 0;
+
+            for (npe01__OppPayment__c pmt : opp.npe01__OppPayment__r) {
+                if (pmt.npe01__Paid__c) {
```
jbrandolinisf: why not make this

totalPaid += (pmt.npe01__Paid__c && pmt.npe01__Payment_Amount__c != null) ? pmt.npe01__Payment_Amount__c : 0;

This eliminates the if, and puts are affirmative choice first.

src/classes/OPP_OpportunityAmountSync_SVC.cls
jbrandolinisf: Will require updates if we decide to allow negative Opportunity Amounts.  We DO allow negative Payments.

src/classes/PMT_Payment_TEST.cls
vmartinov-sforg: Does it automatically create a payment when the opportunity is created if the Payments enables is true?

jbrandolinisf: This disables the whole PMT_PaymentCreator trigger.  There is other functionality in there we would want to run in a unit test to make sure there is not breakage between our change and that code.  A better option is to set the field npe01__Do_Not_Automatically_Create_Payment__c to true on the opportunity to tell it not to create the payment automatically if needed.

jbrandolinisf:
@vmartinov-sforg - Hey, Vlad, Pedro and I connected briefly.  There is one other thing we need to do with this.  This functionality should only be enabled if the Custom Setting Enforce Accounting Data Consistency is enabled/true.  I believe there is already a variable present in the PMT_Payment_TDTM trigger for this called isEnforceAccountingDataConsistency.  So you can wrap these in blocks with that if (isEnforceAccountingDataConsistency) { ... }

vmartinov-sforg:
@jbrandolinisf The methods runForPayments() and runForOpportunities()  already have checks for if (isEnforceAccountingDataConsistency) {..}. Is that what you meant?

jbrandolinisf:
Yeah, I just wanted to make sure it was behind that, meaning if that is false neither the validation nor the Amount update will occur.  I didn't look closely to see whether it was/wasn't so if I missed it in the 3rd pass over, I apologize!

src/classes/PMT_Payment_TEST.cls
```diff
@@ -837,6 +837,92 @@ private class PMT_Payment_TEST {
     }
 
 
+    /**
+     * @description Check that the opportunity amount is updated when the payed payments amounts
+     * exceed the opportunity amount
+     */
+    @isTest
+    private static void testOpportunityAmountSync() {
```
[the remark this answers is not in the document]
jbrandolinisf: Time Permitting, we would have the following tests:

- Tests with Enforce Accounting Data Consistency Enabled:
- Non Finalized Opportunity, Amount changed to more than Sum of Paid Payments: Result (Allowed)
- Non Finalized Opportunity,  Amount changed to less than Sum of Paid Payments: Result (Blocked)
- Finalized Opportunity, Amount same as Sum of Paid Payments: Result (Allowed)
- Finalized Opportunity, Amount < Sum of Paid Payments: Result (Blocked)
- Finalized Opportunity, Amount > Sum of Paid Payments: Result (Blocked)
- Non Final Opportunity inserted with unpaid payments totaling < than opportunity Amount.  Change Unpaid Payments to Paid.  Amount should remain the same
- Non Finalized Opportunity inserted with unpaid payments totaling > than opportunity Amount. Change Unpaid Payments to Paid.  Amount should increase to Sum of Paid Payments
- Non Finalized Opportunity inserted with unpaid payments totaling < Opportunity Amount.  Change Unpaid Payments to Paid.  No change to Opportunity Amount.  Change Opportunity to Finalized Stage.  Amount should change to Sum of Paid Payments-- I think as coded, this will instead throw a validation error, as it does not re-sum the Payment Amounts when going to finalized though). @pedrodalcol - Do I have the behavior we want correct here?  Basically, when an Opportunity Becomes finalized (but was not previously, it needs to recalculate the Amount based on the Paid Payments.  The logic as coded currently will not do that because we are only recalcing on payment changes..
- Non Finalized Opportunity inserted with unpaid payments.  Change to Finalized Opportunity.  Update the payments to Paid.  Confirm Opportunity Amount is now Sum of Paid Payments:  Note: I don't love this test because it will eventually conflict with write off on finalized behavior.
- Finalized Opportunity with Paid Payment equal to amount.  Add additional unpaid payment.  No change in  Opp amount.  Add additional paid payment confirm Opp AMount increases.
- Finalized Opportunity with Paid Payments equal to Amount.  Change one of the Payments to Unpaid or Written Off.  Confirm Opportunity Amount updates.

-- Tests with Enforce Accounting Data Consistency Not Enabled:
- Create Non Finalized Opportunity with Amount > Paid Payments.  Change Opp Amount to < Sum of Paid Payment Amounts.  Confirm no error.
- Create Non Finalized Opportunity with Amount < Unpaid Payments.  Make Payments Paid.  Confirm No change to Opp Amount.
- Create Finalized Opportunity With Amount Different than Paid Payments.  Confirm no error.   Update Amount.  Confirm no error.
- Create Finalized Opportunity With Amount and Unpaid Payments.  Change Unpaid Payments to Paid.  Confirm no change in amount.

I realize you will not be able to get to these in the time you have.  Can you add comments with a //TODO: to this so we can add these tests later?

jbrandolinisf:
@vmartinov-sforg - I confirmed with Pedro that we do need to recalculate the Opportunity Amount on Opportunity Update (if the Stage changes from a non Finalized to a Finalized State).  So we would call isFinalized on both the old and the new opportunity records from the trigger map.  If it was not finalized, but now is.  We should recalculate the amount.  Again, this too should be behind the Enforce Accounting Data Consistency setting.

vmartinov-sforg:
@jbrandolinisf Can we chat about the latest change request from Pedro?

jbrandolinisf:
Sure.  Let me know when.

jbrandolinisf:
@vmartinov-sforg  - There is a pending merge for this.  Can you fix that and I will review?

vmartinov-sforg:
@jbrandolinisf I don't see this branch being out-of-sync with master. Where do you see that?

jbrandolinisf:
@vmartinov-sforg - Never mind, I think I misread the status after I told it to go ahaed and do it's thing.  I will review now!

src/classes/OPP_OpportunityAmountSync_SVC.cls
```diff
@@ -0,0 +1,66 @@
+/*
```
jbrandolinisf: Please add this to the ApexDocContent/Opportunity.htm files, In a new section called Accounting and Data Consistency

src/classes/OPP_OpportunityAmountSync_SVC.cls
```diff
@@ -0,0 +1,66 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Payments
+* @description service class used by PMT_Payment_TDTM to set the opportunity amount to match the paid payment amounts
+*/
+
+public class OPP_OpportunityAmountSync_SVC {
+
+    /**
+     * updates the opportunity amount based on the paid payment amounts if there is a mismatch
```
jbrandolinisf: Not a huge deal, but can you add the @description tag?

jbrandolinisf: Also can you indicate that the List that is returned consists of the records to be updated, and that DML is not processed in this method?

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -45,8 +45,8 @@ public with sharing class OPP_StageMappingUtil {
     @testVisible private static final String UNCOMMITTED_STATE_VAL = 'UNCOMMITTED';
     @testVisible private static final String COMMITTED_STATE_VAL = 'COMMITTED';
     @testVisible private static final String FINALIZED_STATE_VAL = 'FINALIZED';
-    @testVisible private static final String TEST_FINALIZED_STAGE = 'Closed Won';
-    @testVisible private static final String TEST_COMMITTED_STAGE = 'Pledged';
+    @testVisible public static final String TEST_FINALIZED_STAGE = 'Closed Won';
```
jbrandolinisf: If this was from a test class, you did not need to make them public.  @testVisible would have made them accessible for this purpose.

src/classes/OPP_StageMappingUtil.cls
```diff
@@ -90,6 +90,15 @@ public with sharing class OPP_StageMappingUtil {
         return UNCOMMITTED_STATE_VAL;
     }
 
+    /*******************************************************************************************************
+    * @description Determines if the opportunity is in the FINALIZED state
+    * @param Opportunity 
+    * @return Boolean
```
jbrandolinisf: Maybe add returns true if Finalized, and false otherwise?

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +157,55 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+        // Prevent the user from modifying the Opp Amount
+        if (triggerAction == TDTM_Runnable.Action.AfterUpdate && isEnforceAccountingDataConsistency) {
+            Set<Id> oppIds = (new Map<Id, Opportunity>(newList)).keySet();
+            Map<Id, Opportunity> oppsById = getPaymentOpportunities(oppIds);
+
+            for (Opportunity opp : newList) {
+                // get the opp with payments
+                Opportunity oppWithPayment = oppsById.get(opp.Id);
+
+                Decimal totalPaid = 0;
```
jbrandolinisf: Calculating the total Paid payment amount could be a shared method in the OPP_OpportunityAmountSync_SVC class that it and this class use.(170 - 173)

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -132,6 +157,55 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+            Map<Id, Opportunity> oppsById = getPaymentOpportunities(oppIds);
+
+            for (Opportunity opp : newList) {
+                // get the opp with payments
+                Opportunity oppWithPayment = oppsById.get(opp.Id);
+
+                if (OPP_StageMappingUtil.isOpportunityFinalized(opp) && !OPP_StageMappingUtil.isOpportunityFinalized(mapIdOppOld.get(opp.Id))) {
+                    Decimal totalPaid = 0;
+
+                    for (npe01__OppPayment__c pmt : oppWithPayment.npe01__OppPayment__r) {
```
jbrandolinisf: See shared method comment

src/classes/PMT_Payment_TEST.cls
```diff
@@ -837,6 +837,871 @@ private class PMT_Payment_TEST {
...
+                )
+        };
+        insert payments;
+        
+        Test.stopTest();
+
+        // update the opp to a larger amount
+        try {
+            opp.Amount = 300;
+            update opp;
```
jbrandolinisf: This is fine, but the other way to handle this (that won't throw a DML exception on error) is to do the following:

Database.SaveResult sr = Database.update(opp, false);

You can then interrogate the SaveResult to see if it was successful, and or look further at the errors returned.

What you did is fine, just mentioning it in case it resonates for another use case, or in the future.

src/classes/PMT_Payment_TEST.cls
```diff
@@ -837,6 +837,871 @@ private class PMT_Payment_TEST {
...
+
+
+    /* Test with Enforce Accounting Data Consistency Enabled */
+    /* @description Non Finalized Opportunity inserted with unpaid payments totaling > than opportunity Amount. 
+                    Change Unpaid Payments to Paid. Amount should increase to Sum of Paid Payments
+    */
+    @isTest
+    private static void oppNonFinalizedUpdateHigherPaymentsToPaid() {
+        UTIL_CustomSettingsFacade.getContactsSettingsForTests(new npe01__Contacts_And_Orgs_Settings__c (npe01__Payments_Enabled__c = true, Enforce_Accounting_Data_Consistency__c = true));
+        Account acc = [SELECT Id FROM Account LIMIT 1];
```
jbrandolinisf: Remind me to show you some of the NPSP test helper methods for creating data.  They can be useful to make test classes that can run multiple tests at once, which can reduce test execution times.

PR #4376, opened 2019-07-10
BDI: Prevent field level security exception for individual Data Import records from failing entire Data Import Batch

kenlewis-sfdo:
This set of changes adds a service class that is used to perform Field Level Security checks early in the data import `process` method.  Data Import records with FLS exceptions are not processed further and are stamped with their respective FLS exception details at the end of the method.  It also adds a method to the BDI_FieldMapping interface (and implementations of the method in our current implementing classes) to enable retrieving the mapped target objects and fields in a unified format keyed by the Data Import source field name. 

# Critical Changes

# Changes
- NPSP Batch Data Import now performs field level security on a row-by-row basis and will successfully process Data Import records that have appropriate field level access, as opposed to failing the entire Data Import Batch if a single Data Import row has insufficient field level access.  This change also allows the Dry Run process to complete when a User has read access to the mapped target fields, whereas before the User needed read and write access to the mapped fields to successfully dry run the Batch.  

# Issues Closed

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-033365

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087v6MQAQ" target="_blank">W-033365</a>

src/classes/BDI_FieldMapping.cls
```diff
@@ -50,4 +50,5 @@ public interface BDI_FieldMapping {
             String targetObjectName, 
             List<String> dataImportFields);
 
+    Map<String, Map<String, Set<String>>> getTargetObjectsAndFieldsByDataImportField();
```
ptewson-sfdo: Is it worth considering an inner class to encapsulate the Map<String, Set<String>> inner part of this collection? That would avoid the nested maps

src/classes/BDI_DataImportService.cls
```diff
@@ -525,6 +529,20 @@ global with sharing class BDI_DataImportService {
...
+            this.flsValidator =
+                    new FieldLevelSecurityValidator(
+                            dataImports,
+                            fieldMapper.getTargetObjectsAndFieldsByDataImportField());
+
+            //From this point forward, only operate on the Data Import records that passed FLS
+            //validation.  Records that did not pass FLS validation will be updated with their
+            //respective FLS exceptions at the end of this block
+            if (isDryRun) {
+                this.listDI = flsValidator.getDataImportsWithAccessibleTargetFields();
```
ptewson-sfdo: It's a bit wordy, but I love how descriptive these names are - there's a lot of complexity to FLS in this arena and it's worthwhile to be very clear about source vs target, accessible/createable/updateable, etc. Nice!

src/classes/BDI_DataImportService.cls
```diff
@@ -525,6 +529,20 @@ global with sharing class BDI_DataImportService {
...
+            } else {
```
ptewson-sfdo: Do we really check different levels of access for dry run vs 'real' processing? You don't NEED write access to dry run, but would it be better to advertise the FLS error that you WOULD get in processing from dry run?

src/classes/BDI_DataImportService.cls
```diff
@@ -607,6 +629,144 @@ global with sharing class BDI_DataImportService {
...
+                    if (targetFieldsByObject == null) { //this source field is unmapped
+                        continue;
+                    }
+
+                    //Now loop through all of the target object + field combos for this source field
+                    for (String targetObjectName : targetFieldsByObject.keySet()) {
+                        for (String targetFieldName : targetFieldsByObject.get(targetObjectName)) {
+                            DescribeFieldResult fieldDescribe = UTIL_Describe.getFieldDescribe(targetObjectName, targetFieldName);
+                            if (fieldDescribe.isAccessible() == false) {
+                                nonAccessibleTargetFields.add(targetObjectName + '.' + targetFieldName);
```
ptewson-sfdo: If we define the string `targetObjectName + '.' + targetFieldName` before we hit these conditionals, I think we can reduce our memory footprint, and possible save a little bit of execution time by rebuilding the same field twice. Technically we might even be able to define our combined field strings outside of the DI loop (or else check against existence in a Set and only create new if its not present already)? As long as it's not an issue sharing instances of that string between collections, but I don't think we are modifying the strings after this point. And string arithmetic is rumored to be relatively expensive?

src/classes/BDI_FieldMappingHelpText.cls
```diff
@@ -69,6 +69,37 @@ public with sharing class BDI_FieldMappingHelpText implements BDI_FieldMapping {
...
+                new Map<String, Map<String, Set<String>>>();
+
+        BDI_MigrationMappingHelper helper = new BDI_MigrationMappingHelper();
+        for (BDI_MigrationMappingUtility.HelpTextFieldMapping helpTextFieldMapping : helper.getAllHelpTextFieldMappings()) {
+            if (!helpTextFieldMapping.isUsable()) {
+                continue;
+            }
+            if (targetObjectsAndFieldsByDataImportField.get(helpTextFieldMapping.dataImportFieldAPIName) == null) {
+                Set<String> targetFields = new Set<String>{helpTextFieldMapping.targetFieldAPIName};
+                Map<String, Set<String>> targetFieldsByTargetObject = new Map<String, Set<String>>{helpTextFieldMapping.targetObjectAPIName => targetFields};
```
ptewson-sfdo: Can we clean up the line breaks in this method? #petpeeves

src/classes/BDI_DataImportService.cls
```diff
@@ -607,6 +629,144 @@ global with sharing class BDI_DataImportService {
...
+        for (DataImport__c dataImport : dataImportsWithFlsExceptions) {
+            dataImport.Status__c = statusDIFailed();
+            dataImport.ApexJobId__c = apexJobId; //Needed to display the correct totals on the BDI_DataImport page
+            String errMessage;
+            if (isDryRun) {
+                Set<String> nonAccessibleFields = flsValidator.getFieldsWithoutReadAccess(dataImport.Id);
+                List<String> errorMsgObjects = new List<String>{String.join(new List<String>(nonAccessibleFields), ', ')};
+                errMessage = String.format(System.Label.flsReadAccessError, errorMsgObjects);
+            } else {
+                Set<String> nonCreateableUpdateableFields = flsValidator.getFieldsWithoutCreateUpdateAccess(dataImport.Id);
```
ptewson-sfdo: Can we clean up these line breaks (as well as line 645, 649)?

src/classes/BDI_DataImport_TEST2.cls
```diff
@@ -1384,25 +1384,28 @@ private with sharing class BDI_DataImport_TEST2 {
...
 
-        ID ApexJobId = Database.executeBatch(bdi, 10);
-        Test.stopTest();
-    
+        User readOnlyUser = UTIL_UnitTestData_TEST.createUserWithoutInsert('Read Only');
+        System.runAs(readOnlyUser) {
+            //run batch data import
+            Test.StartTest();
+            BDI_FieldMappingHelpText fieldMapper = BDI_FieldMappingHelpText.getInstance();
+            BDI_DataImportService.failFieldLevelSecurity = true;
```
ptewson-sfdo: Do we still need to set this? Looks like it may be getting dropped from DataImportService????

src/classes/BDI_FieldLevelSecurityAccessLevel.cls
```diff
@@ -0,0 +1,40 @@
+/*
```
ptewson-sfdo: As we talked about having this in its own class feels excessive ... can we move it into a member of UTIL_Permissions or perhaps BDI_DataImportFLSService if it works with the inheritance model?

src/classes/BDI_FieldLevelSecurityService.cls
```diff
@@ -0,0 +1,45 @@
+/*
```
ptewson-sfdo: I was thinking about this over the weekend and I think I'm leaning towards leaving this interface out of the package for now, and just having BDI_DataImportFLSService exist as a standalone.

It's not a super strong opinion and it's maybe subjective so I could flex, but I'm just not I can see the future clearly enough to see when we need to share this behavior. We can try to schedule a quick huddle with some of the other devs as well, yeah - I can be flexible, but my preference would be to avoid providing the interface until we know more about where else this behavior would see use.

src/classes/BDI_SObjectFieldWrapper.cls
```diff
@@ -0,0 +1,51 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Batch Data Import
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Wrapper used to hold the SObjectType related to an SObjectField since the SObjectField
+* token does not provide a reliable method to get its related SObjectType.
+*/
+public class BDI_SObjectFieldWrapper {
```
ptewson-sfdo: As of this point in time, I thiiiiiink this feels like it should be an inner convenience class for BDI_DataImportFLSService. I know that in the longer term we have discussed unifying around bring in more shared wrapper classes, but until we get there I believe we should keep this in the FLSService - the only class where it's currently referenced

src/classes/BDI_DataImportFLSService.cls
```diff
@@ -0,0 +1,150 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Batch Data Import
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Service class used to validate
+* Field Level Security of the target fields mapped in the data import process.
+*/
+public class BDI_DataImportFLSService {
+    Set<AccessLevel> accessLevels;
+    Map<SObjectField, BDI_TargetFields> fieldMap;
```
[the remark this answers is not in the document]
kenlewis-sfdo: Updated!

kenlewis-sfdo:
@melissabarber I updated the description and text in the Changes of this PR, let me know if you have any questions or suggestions!

melissabarber:
Thank you for the heads up, @kenlewis-sfdo, and for putting that information in the Changes section. Because this PR is going into a feature branch I won't refine the text you added, but it's super helpful for PRs to have a quick description of what's covered in this work!

PR #4380, opened 2019-07-11
Prevent the user from decreasing the Opportunity amount to an amount that is less than the sum of paid payment amounts. 

KietTruongSFDO:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


markross-sfdo:
@KietTruongSFDO Shouldn't the error message be in a custom label?

How about:
'You cannot change the amount on an Opportunity once it is ' StageName '. You can instead update the amounts on related Payments.'

KietTruongSFDO:
> @KietTruongSFDO Shouldn't the error message be in a custom label?
> 
> How about:
> 'You cannot change the amount on an Opportunity once it is ' StageName '. You can instead update the amounts on related Payments.'

@markross-sfdo I'll change it to a custom label. Would it be too confusing to say "finalized" instead of the stage name?

markross-sfdo:
@KietTruongSFDO Augh, I just saw your second email. I think changing it to Finalized is fine.

KietTruongSFDO:
@markross-sfdo  @melissabarber How does this Custom Label look for the error message?

markross-sfdo:
@KietTruongSFDO Should there not be a period at the end?

KietTruongSFDO:
> @KietTruongSFDO Should there not be a period at the end?

@markross-sfdo  My mistake. Added the period.

judisohn:
@markross-sfdo Can we make sure that when this is ready to go to master, the final PR calls this out in the critical changes section, in case there are any existing NPSP customers who are doing this intentionally? Thanks.

markross-sfdo:
Absolutely, @judisohn!

markross-sfdo:
@KietTruongSFDO Should this be linked to an Issue or WI?

KietTruongSFDO:
@markross-sfdo This isn't linked to an Issue. If I understand correctly, we don't make issues for NPSP, only the customers do. @judisohn, correct me if I'm wrong about the usage of Issues. This is for W-033109.

judisohn:
@kiettruongsfdo @markross-sfdo We (CCE) make issues, not customers. As this is something that no customer or partner has complained about and is being driven by the needs of the EDU team, I don't think it's appropriate to make an issue. But I'd like to see the potential data change called out in the release notes since we have no clue if customers are already doing this as part of a "normal for them" workflow. FYI @sam-knox @melissabarber

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -313,6 +316,52 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+    ********************************************************************************************************/
+    private void enforceFinalizedOpportunityAmountIsEqualToPayment(Map<Id, Opportunity> newMap, Map<Id, Opportunity> oldMap) {
+
+        // Get the opportunities with payments.
+        Map<Id, Opportunity> oppsById = getPaymentOpportunities(newMap.keySet());
+
+        for (Opportunity queriedOpp : oppsById.values()) {
+            Opportunity oldOpp = oldMap.get(queriedOpp.Id);
+            Opportunity updatedOpp = newMap.get(queriedOpp.Id);
+            if (oldOpp.Amount != updatedOpp.Amount && OPP_StageMappingUtil.determineOppStateFromOppStage(updatedOpp.StageName) == OPP_StageMappingUtil.getFinalizedState()) {
```
jbrandolinisf: Would be better if it passed in the stage to a method in that class that returned true or false, as we will do this in other spots...
Added to OPP_StageMappingUtil: 
public static Boolean isFinalized(String stageName) {
   return OPP_StageMappingUtil.determineOppStateFromOppStage(stageName) == FINALIZED_STATE_VAL;
}

Then it becomes (if (oldOpp.Amount != updatedOpp.Amount && OPP_StageMappingUtil.isFinalized(updateOpp.StageName))

Or you could even pass the Opportunity, and get the Opportunity stage inside the method.

src/labels/CustomLabels.labels
```diff
@@ -1222,6 +1222,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>OCR Remove Primary Duplicate Batch Error Text</shortDescription>
         <value>Unable to remove the duplicate primary OCRs from the Opportunity {0}. Review the Contact Roles on the Opportunity record, choose which is the correct one, and delete the others.</value>
     </labels>
+    <labels>
+        <fullName>OppAmountLessThanPaidAmountError</fullName>
+        <categories>Payment, Opportunity</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>OppAmountLessThanPaidAmountError</shortDescription>
```
jbrandolinisf: Should this be NotEqualTo rather than LessThan

src/classes/PMT_Payment_TEST.cls
```diff
@@ -836,6 +836,55 @@ private class PMT_Payment_TEST {
...
+        // Mark the payment as Paid
+        pmts[0].npe01__Payment_Amount__c = 500;
+        pmts[0].npe01__Paid__c = true;
+        update pmts;
+
+
+        try {
+            opp.Amount = 100;
+            update opp;
+        } catch (Exception e) {
```
jbrandolinisf: Need an assert or exception thrown after update as it should blow up if no exception was thrown.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -313,6 +316,52 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
+    /*******************************************************************************************************
+    * @description Validate that the opportunity has the appropriate values. This will check to make sure
+    * that the opportunity amount is not set to an amount lower than the paid payments.
+    * @param newlist The list of Opportunities from trigger new.
+    * @param oldlist The list of Opportunities from trigger old.
+    * @param triggerAction  Which trigger event (BeforeInsert, AfterInsert, etc.).
+    * @return void Does not return anything
+    ********************************************************************************************************/
+    private void validateOpportunity(Map<Id, Opportunity> newMap, Map<Id, Opportunity> oldMap, TDTM_Runnable.Action triggerAction, DmlWrapper dmlWrapper) {
+        if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
```
jbrandolinisf: I think we should do this on "AfterUpdate".   See https://andyinthecloud.com/2015/04/19/considerations-placing-validation-code-in-an-apex-triggers/ for an explanation of why.

KietTruongSFDO:
We will most likely close this PR in the next 3 weeks. Waiting to make sure this is covered in another PR/WI.

KietTruongSFDO:
**lurch: attach W-033109

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uldQAA" target="_blank">W-033109</a>

PR #4396, opened 2019-07-18
Added functionality to Enable/Disable Two way sync

KietTruongSFDO:
# Critical Changes

# Changes
- We added a custom setting to sync Payment Allocations and Opportunity Allocations. This feature supports future development and is disabled by default.

# Issues Closed

# New Metadata

# Deleted Metadata


jbrandolinisf:
@ptewson-sfdo - Tagging you on this for awareness.  This is the work item to make Payment Allocations - 2 Way Sync Controlled by a Setting, and not on by default.

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -794,6 +794,7 @@ public without sharing class UTIL_CustomSettingsFacade {
         allocationsSettings.Payment_Allocations_Enabled__c = mySettings.Payment_Allocations_Enabled__c;
         allocationsSettings.Rollup_N_Day_Value__c = mySettings.Rollup_N_Day_Value__c;
         allocationsSettings.Use_Fiscal_Year_for_Rollups__c = mySettings.Use_Fiscal_Year_for_Rollups__c;
+        allocationsSettings.Payment_Allocation_Two_Way_Sync_Enabled__c = true;
```
jbrandolinisf: It should grab the value from the mySettings object so we can make if true or false, as necessary.

jbrandolinisf: Please also update configAllocationsSettings to set it explicitly to false.

src/classes/ALLO_PaymentSync_TDTM.cls
```diff
@@ -50,6 +50,12 @@ public class ALLO_PaymentSync_TDTM extends TDTM_Runnable {
     public override DmlWrapper run(List<SObject> newList, List<SObject> oldList,
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
 
+        Allocations_Settings__c allocationsSettings = UTIL_CustomSettingsFacade.getAllocationsSettings();
+
+        if (allocationsSettings.Payment_Allocation_Two_Way_Sync_Enabled__c == false && !Test.isRunningTest()) {
```
jbrandolinisf: Please change this to 

if (!allocationSettings.Payment_Allocation_Two_Way_Sync_Enabled__c && !Test.isRunningTest()) {

this will cover it better because !true covers null as well as false...

src/objects/Allocations_Settings__c.object
```diff
@@ -49,6 +49,14 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Payment_Allocation_Two_Way_Sync_Enabled__c</fullName>
```
jbrandolinisf: I assume @lianeyee and @markross-sfdo are still working on the description and help text for this, or are we forgoing them for this field?

src/classes/ALLO_PaymentSync_TDTM.cls
jbrandolinisf: Please also add a comment indicating that the check for whether tests are running is strictly to cover Abacus until we can refactor the required tests so that they will either explicitly set two way sync enabled to true or are modified to no longer rely on the execution of 2 way sync.  Something like:

// Interim: The exclusion to this for tests is so that existing unit tests in external packages will not fail because they are expecting this to be present.  Once the Abacus tests are refactored and the Payment Allocations Two Way Sync tests are updated, this part of the condition will be removed.

A backlog item for that work item should be logged.

src/classes/UTIL_CustomSettingsFacade.cls
jbrandolinisf: Although it is not required now, if you have time to set the two way sync enabled flag to true in the two way sync tests, that would be very helpful as it would be less to do later.  Also, make sure you run all Allocation related tests, at a minimum...

src/classes/ALLO_PaymentSync_TDTM.cls
jbrandolinisf: This should be after the Payment Allocations Enabled check, not before it.  If Payment Allocations is not Enabled, this setting should not matter at all.  I know it doesn't matter (much) really, but it makes me feel better because it reflects the proper order the settings are/should be applied.

jbrandolinisf: Additionally, an item for removing the test shortcircuit, and making sure NPSP Allocations Test all run without it, should also be added.  That item should also explicitly add tests that confirm 2 way sync does/does not execute according to the schedule, since this work item cannot do so.

src/classes/UTIL_CustomSettingsFacade.cls
KietTruongSFDO: I ran all the test classes yesterday and then were all passing.

src/classes/ALLO_PaymentSync_TDTM.cls
KietTruongSFDO: > Please change this to
> 
> if (!allocationSettings.Payment_Allocation_Two_Way_Sync_Enabled__c && !Test.isRunningTest()) {
> 
> this will cover it better because !true covers null as well as false...

Using !someBoolVar will throw a null pointer if the boolean is actually null. Using someBoolVar == false actually covers nulls and true/false.

src/objects/Allocations_Settings__c.object
KietTruongSFDO: Just added the description and help text.

src/classes/UTIL_CustomSettingsFacade.cls
KietTruongSFDO: Updated the configAllocationsSettings to set the value to false

Updated to pull the value from mySettings object

jbrandolinisf:
This is on hold for now, and will likely be eliminated entirely in the next month.

PR #4397, opened 2019-07-18
BDI: Concatenate Additional Object Error Messages

kenlewis-sfdo:
This set of changes collects all error messages that occur within
the additional objects service class and concatenates them so that
there is one combined summary error message that is placed in the Data
Import Failure Information field.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_ObjectWrapper.cls
+++ b/src/classes/BDI_ObjectWrapper.cls
@@ -51,4 +51,19 @@ public with sharing class BDI_ObjectWrapper {
         this.fieldMappings = fieldMappings;
         this.dataImport = dataImport;
     }
+
+    public void setDataImportStatus(String status) {
+        dataImport.Status__c = status;
+    }
+
+    public void setImportedRecordStatusError(String errorMessage) {
+        String statusField = objMapping.Imported_Record_Status_Field_Name__c;
+        if (statusField != null && errorMessage != null) {
+            dataImport.put(statusField, errorMessage.left(255));
+        }
+    }
+
+    public void setFailureInformation(String failureInformation) {
+        dataImport.FailureInformation__c = failureInformation;
+    }
 }
--- a/src/classes/BDI_AdditionalObjectService.cls
+++ b/src/classes/BDI_AdditionalObjectService.cls
@@ -181,55 +181,6 @@ public with sharing class BDI_AdditionalObjectService {
         return objAPINameToObjWraps;
     }
 
-    Map<Id, String> errorMessagesByDataImportId;
-    void logError(BDI_ObjectWrapper objectWrapper, String errorMessage) {
-        DataImport__c dataImport = objectWrapper.dataImport;
-
-        //Set the overall data import record status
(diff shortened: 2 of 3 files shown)


kenlewis-sfdo:
**lurch: attach W-034098

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C6rkQAC" target="_blank">W-034098</a>

src/classes/BDI_ObjectWrapper.cls
```diff
@@ -51,4 +51,19 @@ public with sharing class BDI_ObjectWrapper {
...
     }
+
+    public void setDataImportStatus(String status) {
+        dataImport.Status__c = status;
+    }
+
+    public void setImportedRecordStatusError(String errorMessage) {
+        String statusField = objMapping.Imported_Record_Status_Field_Name__c;
+        if (statusField != null && errorMessage != null) {
+            dataImport.put(statusField, errorMessage.left(255));
```
ptewson-sfdo: Possibly consider truncating to 252 length and concatenating "..." just to make it clear that the message has been truncated?

kenlewis-sfdo:
@callen-sfdo the `BDI_DataImportService.LogBDIError` method always sets the `DataImport__c.FailureInformation__c` field to the value of the error message that you pass to it, which makes it difficult to pass it both a specific error for a given additional object (required field error, invalid predecessor error, update or insert dml error...) and a concatenated summary error for all additional object failures in the transaction.  This was an attempt to remedy that by tracking the errors across additional objects processed for any given tier and setting the values for the individual imported status fields and overall failure information field here.

src/classes/BDI_ObjectWrapper.cls
kenlewis-sfdo: @ptewson-sfdo happy to do this as a future WI - currently the entire message is included in the Failure Information field which is a long text area so the User will at least be presented with the entire error message there if it is truncated here.

callen-sfdo:
@kenlewis-sfdo Yes LogBDI error does do that, the original suggestion was just to change it so it concatenated when passed new errors instead of overwriting.  Because the additional object importAdditionalObjects already strips out failed di records between tiers all that was needed was to change logBdiError like this (did you try this and it didn't work?):
`
 public void LogBDIError(DataImport__c dataImport, String errorMessage, String statusField) {
        dataImport.Status__c = BDI_DataImport_API.bdiFailed;
        if (!dataImport.FailureInformation__c.isEmpty()) {
           dataImport.FailureInformation__c += '/n ' + errorMessage;
        } else {
        dataImport.FailureInformation__c = errorMessage;
       }
        
        if (statusField != null) {
            dataImport.put(statusField, errorMessage.left(255));
        }
    }

`

kenlewis-sfdo:
@callen-sfdo this method is used by both the BDI_ContactService and BDI_Donations classes also.  I didn't want to modify it without doing a deeper dive into how those classes are using it and whether the modifications for the additional objects service class should be applied to those classes usage also.

PR #4400, opened 2019-07-19
Feature/bdi co add field-side help text check for migration util

psadie:
Adds field-side checking of Help Text mapping during migration

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


psadie:
**lurch: attach W-034596

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C7wtQAC" target="_blank">W-034596</a>

psadie:
**lurch: attach W-034099

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C6rpQAC" target="_blank">W-034099</a>

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -174,7 +174,6 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
 
             } catch (Exception e) {
                 Database.rollback(sp);
-                ERR_Handler.processError(e, ERR_Handler_API.Context.STTG);
```
ptewson-sfdo: What's the thinking behind removing these calls to ERR_Handler.processError() (or conversely, what was the thinking behind including them initially?)

psadie: So the processErrors method from ERR_Handler takes in a an exception, a list of exceptions, or a list of Error__c. It's really an artifact from when I copied some of the code over from the CRLPs controller. Our error messages need the custom labels/custom label combinations instead of the exception straight up so I figured we don't really need it. Plus wherever they are in our controller we'll end up with two of the same error messages.

ptewson-sfdo: Oh yeah - let's remove that, good call!

PR #4410, opened 2019-07-22
PYALLOCB: 1-B : Update Error Object and Err_Handler class

jbrandolinisf:
# Critical Changes

# Changes
- We added two new Error fields, Related_Record_Id__c and Retry_Pending__c, as well as additional background functionality. Together, these changes support future development of error logging functionality.   
# Issues Closed

# New Metadata
Error (Error__c) Object:
- Retry Pending? (Retry_Pending__c [Checkbox]) Added for future support.  If true, will indicate the associated process and record needs to be reprocessed.  After successful reprocessing, it will change to false [unchecked].  This field is not required.
- Related Record Id (Related_Record_Id__c [Text(18)] Added to relate the Error to a specific record in the system.    This field is not required, and is not populated by most existing processes.
# Deleted Metadata


jbrandolinisf:
@force2b - I recreated this pr one more time with a new branch that has the record error support in it, directly to master.  It is no longer a child of feature/pmt-alloc

jbrandolinisf:
For reference, the original PR: #4352

beAGoodQE:
@markross-sfdo I see you have worked with John last month on the Quip Doc.  Did you get a chance to view this PR as well?

jbrandolinisf:
@force2b - All set, I believe.

src/objects/Error__c.object
```diff
@@ -144,6 +144,29 @@
...
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
+        <fullName>Retry_Pending__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>If checked, the record entered in Record Id will be reprocessed in the next available scheduled process. If the reprocessing is successful, this field is then unchecked.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>If checked, the record entered in Record Id will be reprocessed in the next available scheduled process.</inlineHelpText>
```
force2b: @jbrandolinisf What would you think about something like the following for the help text to explain that reprocessing is not a generally available feature in NPSP -- _"For specific types of errors on special processing jobs, if this field is checked, the record referenced by the Related Record Id will be reprocessed in the next available scheduled process"_. And maybe the Description (which we can update in future push upgrades) can specify that it's only used by Gem/Abacus.
@markross-sfdo 
CC @sam-knox

jbrandolinisf: I agree with what you are saying, except for the part about it being only GAbacus.  The Payment Allocations Batch is NPSP functionality that is independent of them, though we do rely/need it.  And if we do other batches in NPSP, we may want other processes to do this.

Opinion?

force2b: Very good point. The proposed InlineHelpText change doesn't mention GAbacus so that might be OK. The Description value can be push updated to customers so even if we wanted to note in their now that it's unique to GAbacus, we can change that at a later point if NPSP does implement reprocessing functionality. I'm ok not mentioning GAbacus there too. Overall I was trying to avoid the question from customers on what type of NPSP errors can be reprocessed.

jbrandolinisf: I see.  It is kinda true of the other field as well...The only section of existing NPSP code that will log the RecordId once this goes is the Recurring Donations code.  The rest of them all use the Exception logging stuff which is not associated with specific records (even if it is).

jbrandolinisf: @markross-sfdo , @sam-knox:  @force2b and I talked and he convinced me we should update the InlineHelpText for this.  I am going to tag you both on the associated Quip doc with the text.

jbrandolinisf:
@force2b - Ready for re-review.

PR #4413, opened 2019-07-23
W-017996 add seasonal address batch size setting

robertgalex:
# Critical Changes

# Changes
- We created a Seasonal Address Batch Size setting in NPSP Settings > Bulk Data Processes > Batch Process Settings. The default size is 10.

# Issues Closed

# New Metadata
Custom Field:
- npo02__Households_Settings__c.Seasonal_Addresses_Batch_Size__c

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-017996

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000005KmZcAAK" target="_blank">W-017996</a>

src/objects/npo02__Households_Settings__c.object
```diff
@@ -22,4 +22,18 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>Seasonal_Addresses_Batch_Size__c</fullName>
+        <defaultValue>10</defaultValue>
+        <description>Number of records per batch to process when setting seasonal address as the mailing address.</description>
```
RadSFDeveloper: @sam-knox , can you please review new field Description and Help Text as well as the PR "Changes" section? Thanks!

src/classes/ADDR_Addresses_TEST2.cls
```diff
@@ -466,7 +466,9 @@ public with sharing class ADDR_Addresses_TEST2 {
 
         // run the scheduled batch directly
         Test.startTest();
-        Database.executeBatch(new ADDR_Seasonal_SCHED(), 10);
+        final Integer BATCH_SIZE = (UTIL_CustomSettingsFacade.getHouseholdsSettings().Seasonal_Addresses_Batch_Size__c != null) ?
```
RadSFDeveloper: - Variables with all capital letters are reserved for constants. Local vars, even if they are `final`, use lowerCamelCase. 
- Let's set the Custom Settings field batch size in the test and then retrieve it in the test. (This comment is applicable to other unit test in this class executing `ADDR_Seasonal_SCHED` batch.)

src/classes/ADDR_Seasonal_SCHED.cls
```diff
@@ -35,7 +35,8 @@
 */ 
 public class ADDR_Seasonal_SCHED implements Database.Batchable<SObject>, Schedulable {
 
-    private static final Integer BATCH_SIZE = 10;
+    private static final Integer BATCH_SIZE = (UTIL_CustomSettingsFacade.getHouseholdsSettings().Seasonal_Addresses_Batch_Size__c != null) ?
```
RadSFDeveloper: - If the batch size is not provided, negative or null, default should be 10. No need for `Math.abs()`.

src/pages/STG_PanelSchedule.page
```diff
@@ -108,6 +108,18 @@
                     </div>
                 </div>
             </div>
+
+            <div class="slds-grid slds-gutters_small">
+                <div class="slds-col slds-has-flexi-truncate slds-grid" role="listitem">
+                    <div class="slds-form-element">
+                        <apex:outputLabel value="{!$ObjectType.npo02__Households_Settings__c.Fields.Seasonal_Addresses_Batch_Size__c.Label}" for="seasonalAddressBatchSize" styleClass="slds-form-element__label" />
+                        <div class="slds-form-element__control">
+                            <apex:outputField value="{!stgService.stgHH.Seasonal_Addresses_Batch_Size__c}" rendered="{!isReadOnlyMode}" styleClass="slds-form-element__static" />
```
RadSFDeveloper: The new "Seasonal Address Batch Size" is set to be the last field in the list. Is this fine or the field should be reordered? @force2b , @jesslopez-sf 

Screen Shot 2019-07-23 at 4 21 06 PM

src/pages/STG_PanelSchedule.page
```diff
@@ -108,6 +108,18 @@
...
+                            <apex:inputField value="{!stgService.stgHH.Seasonal_Addresses_Batch_Size__c}" rendered="{!isEditMode}" id="seasonalAddressBatchSize" styleClass="slds-input" />
```
RadSFDeveloper: - The first AC indicates "User can enter a batch size >0 for Seasonal Address on the NPSP Settings-->Batch Process Settings page". However, no other batch size on that page is validated on the input. @seanpat09 , can you please check if the input should be validated? Regardless of the input check, we still need to default the batch size to 10 if it is invalid since user can enter the value via Custom Settings page.
- If the batch size is null, should we default it to 10 when the batch size is displayed as either output or input?

seanpat09: I agree that some level of validation needs to happen, but I think it's beyond the scope of this work item. We should be fine with the back end default for now in case the user inputs invalid values in this setting.

src/classes/ADDR_Seasonal_SCHED.cls
```diff
@@ -35,13 +35,34 @@
...
+    private final static Integer DEFAULT_BATCH_SIZE = 10;
+
+    @TestVisible
+    private Integer batchSize {
+        get {
+            if (batchSize == null) {
+                npo02__Households_Settings__c householdSettings = UTIL_CustomSettingsFacade.getHouseholdsSettings();
+
+                if (householdSettings.Seasonal_Addresses_Batch_Size__c == null
+                    || householdSettings.Seasonal_Addresses_Batch_Size__c < 0
```
RadSFDeveloper: Let's check for 0 as well, ie `householdSettings.Seasonal_Addresses_Batch_Size__c <= 0`, thoughts?

src/pages/STG_PanelSchedule.page
RadSFDeveloper: Sounds good, thanks for verifying!

src/pages/STG_PanelSchedule.page
force2b: I think the last field is fine. If I had my preference, this field and Level Assignment Batch would have its own heading rather than be listed with the Rollup Batch Sizes.

src/classes/ADDR_Seasonal_SCHED.cls
seanpat09: Good catch - turns if you use a batch size of 0, the batch just runs with the default 200. I just made the update to include 0

src/pages/STG_PanelSchedule.page
jesslopez-sf: Thanks, @RadSFDeveloper ! I agree completely with @force2b .

seanpat09:
@ravensisco This is ready for testing!

PR #4429, opened 2019-07-24
Switching a payment to negative bug

KietTruongSFDO:
Updated the validation that verifies that the payment allocation amount cannot be more than the payment amount. This change will skip the validation in use cases where there is only the default allocation for the payment. If there is only a default allocation, the amount will never be more/less than the payment amount.  

# Critical Changes

# Changes
- We have updated a validation to exclude use cases where only the default General Accounting Unit has allocations. 

# Issues Closed

# New Metadata

# Deleted Metadata


kbromer:
Why are there no notes on this PR?

jbrandolinisf:
@KietTruongSFDO  - This looks good as far as it goes.  Do we need to handle anything else with this, or was this the only circumstance we needed to cover?

KietTruongSFDO:
> @KietTruongSFDO - This looks good as far as it goes. Do we need to handle anything else with this, or was this the only circumstance we needed to cover?

In terms of the WI that this came from, this was all that was covered.

KietTruongSFDO:
**lurch: attach W-032644

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uADQAY" target="_blank">W-032644</a>

PR #4430, opened 2019-07-24
Feature/bdi co  add unit tests

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
- BDI_ManageAdvancedMappingCtrlTest.cls

# Deleted Metadata


src/classes/BDI_ManageAdvancedMappingCtrl.cls
```diff
@@ -250,6 +250,10 @@ public class BDI_ManageAdvancedMappingCtrl {
             fieldMapping.Source_Field_API_Name__c = fieldMappingWrapper.Source_Field_API_Name;
             fieldMapping.Target_Field_API_Name__c = fieldMappingWrapper.Target_Field_API_Name;
 
+            if (Test.isRunningTest()) {
+                return JSON.serialize(fieldMapping);
```
ptewson-sfdo: I'm trying to think if it's weird to repurpose the return value to send the field mapping back to the test ... I'm not sure there's anything wrong with it, but maybe update the comment headers ("@return String: Deployment Id") to reflect that the return value is different in test context?

PR #4453, opened 2019-07-29
BDI: Prevent additional object dependency errors on dry run

kenlewis-sfdo:
Currently, Dry Run does no matching for additional objects, so this
update skips the additional object processing logic when the data import
dry run process runs, preventing "missing predecessor" errors that can
occur when an additional object is dependent on a predecessor object
being created.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-034884

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C8UPQA0" target="_blank">W-034884</a>

PR #4455, opened 2019-07-30
Updated the post install for Opportunity Stage to State

KietTruongSFDO:
# Critical Changes

# Changes
* We updated the installation and update flow for a new custom metadata type, Opportunity Stage to State Mapping. This feature is intended for a future release and should have no visible impact on your org. 
# Issues Closed

# New Metadata

# Deleted Metadata


KietTruongSFDO:
Hey @force2b, just to keep you in the loop, we had to update the post install. The value we were checking is not available at the time of the installation. So as an alternative, we are adding the CMT record itself into the unmanaged post install deploy. To be able to deploy the CMT, I needed to update the package.xml within unpackage/post/first from API 29 to API 43. Changing to API 43, the quick actions needed the optionsCreateFeedItem tag added as the tag is now required. 

To get the optionsCreateFeedItem value, I created a scratch org with the unchanged API 29 package.xml and repulled the quickactions using API 43. 

Let me know if you have any thoughts on this. Thanks!

force2b:
@KietTruongSFDO Anything in the unmanaged post install is not in the package and will not be deployed to any existing NPSP installation. Only new trial orgs can get unmanaged metadata - if it's in the right folder which @davidmreed can help with. However I was under the impression that your objective was to deploy these new CMT records to existing NPSP customers, even if just selectively.

KietTruongSFDO:
@force2b It was also creating the CMT records on install, not just update. One of the records would have been created no matter what on install. We are still keeping the upgrade logic, with a slight modification to add in the namespace to the fields of the record.

davidmreed:
@KietTruongSFDO @force2b  I am not sure that I completely follow the discussion above, but I want to confirm that all unpackaged metadata is not delivered via push upgrades. I'm happy to do a quick call to discuss the intent for this metadata if that would be helpful.

jbrandolinisf:
Hi @davidmreed , @force2b - This is a fall back, for the following: 

The existing install/update process we added to call the executeAction and update the metadata on install and update was not working...We don't know why, so this is our fallback for GA...we will still have to work out an alternate approach for this, and/or instruct people on setting them up manually.

PR #4473, opened 2019-08-01
fix namespacing of custom metadata type names in migration utility

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_MigrationMappingUtility.cls
+++ b/src/classes/BDI_MigrationMappingUtility.cls
@@ -37,16 +37,6 @@
 public with sharing class BDI_MigrationMappingUtility {
 
     /*******************************************************************************************************
-    * @description Enum of the supported MetadataType objects (without the __mdt suffix)
-    */
-    public enum MetadataObject {
-        Data_Import_Field_Mapping,
-        Data_Import_Object_Mapping,
-        Data_Import_Object_Mapping_Set,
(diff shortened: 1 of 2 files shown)


src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -426,7 +436,11 @@ public with sharing class BDI_MigrationMappingUtility {
     */
     private void createAndEnqueueMigratedCustomObjectMappingSet() {
         Metadata.CustomMetadata objectMappingSetMDT = new Metadata.CustomMetadata();
-        String mdTypeName = 'Data_Import_Object_Mapping_Set__mdt';
+        String mdTypeName =
+            UTIL_Namespace.StrTokenNSPrefix(
```
force2b: Though this pattern works and is what I had used in CRLP at the time because I think we had issues using Describe on the CMT back then, another way you can retrieve the fully qualified object name for the CMT's is using this:
`String mdTypeName = Data_Import_Object_Mapping_Set__mdt.SObjectType.getDescribe().getName()`

Both will work, but the latter might be a little more straight forward.

psadie: Just realized we're actually doing exactly that in our lwc controller after reading this. I'll go ahead and switch these over real quick. Thank you!

PR #4474, opened 2019-08-01
Advanced Mapping: fix migration utility and component url namespace bug

psadie:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
+++ b/src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
@@ -39,7 +39,7 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     /*********************************************************************************************************
     * @description URL of the Advanced Mapping Custom UI aura component container
     */
-    public static final String COMPONENT_URL = '/lightning/cmp/c__BDI_ManageAdvancedMapping';
+    public static String componentUrl = '/lightning/cmp/npsp__BDI_ManageAdvancedMapping';
 
     /*********************************************************************************************************
     * @description Returns the string Id of the Data Import Advanced Mapping panel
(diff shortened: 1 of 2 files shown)


src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -39,7 +39,7 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     /*********************************************************************************************************
     * @description URL of the Advanced Mapping Custom UI aura component container
     */
-    public static final String COMPONENT_URL = '/lightning/cmp/c__BDI_ManageAdvancedMapping';
+    public static String componentUrl = '/lightning/cmp/npsp__BDI_ManageAdvancedMapping';
```
force2b: Since this is in Apex, how about something like the following this way you're not hardcoding a namespace that only works in a namespaced org.
```suggestion
    public static String componentUrl = '/lightning/cmp/' + 
        (String.IsBlank(UTIL_Namespace.getNamespace()) ? 'c' : UTIL_Namespace.getNamespace()) +
        '__BDI_ManageAdvancedMapping';
```

src/classes/STG_PanelDataImportAdvancedMapping_CTRL.cls
```diff
@@ -139,8 +139,12 @@ public with sharing class STG_PanelDataImportAdvancedMapping_CTRL extends STG_Pa
     *
     * @return PageReference
     */
-    public PageReference navigate(){
-        PageReference advancedMapping = new PageReference(COMPONENT_URL);
+    public PageReference navigate() {
+        String namespace = UTIL_Namespace.getNamespace();
+        if (namespace != 'npsp') {
```
force2b: See above as the alternative to hardcoding the namespace.

PR #4480, opened 2019-08-05
Correct BDI unit tests for State & Country Picklists

davidmreed:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


ptewson-sfdo:
@callen-sfdo FYI! We need to be careful about our country values to avoid issues with state / country picklists

PR #4485, opened 2019-08-07
Advanced Mapping: namespace platform event channel name

psadie-sfdo:
# Critical Changes

# Changes
- Fix platform event listener component's channel name namespacing

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_Namespace.cls
+++ b/src/classes/UTIL_Namespace.cls
@@ -45,6 +45,7 @@ public class UTIL_Namespace {
     * @description Finds the namespace for the current context.
     * @return string The current namespace as a string, or a blank string if we're not in a namespaced context.
     ********************************************************************************************************/
+    @AuraEnabled
     public static string getNamespace() {       
         if (plainNamespace == null) {
             string withDotNotation = UTIL_Namespace.class.getName();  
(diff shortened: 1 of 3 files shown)


psadie-sfdo:
**lurch: attach W-035056

psadie-sfdo:
**lurch: attach W-035056

psadie-sfdo:
**lurch: attach W-035056

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C8x2QAC" target="_blank">W-035056</a>

src/classes/UTIL_Namespace.cls
```diff
@@ -45,6 +45,7 @@ public class UTIL_Namespace {
     * @description Finds the namespace for the current context.
     * @return string The current namespace as a string, or a blank string if we're not in a namespaced context.
     ********************************************************************************************************/
+    @AuraEnabled
```
ptewson-sfdo: @force2b - Do we have a policy of avoiding AuraEnabled in our util classes? Should we support this with a pass-through method in the LWC component controller?

force2b: I'm not sure there's an official policy, but I wouldn't recommend it. Your suggestion of a pass through would be better.

PR #4492, opened 2019-08-08
Manage Allocation bug on payments. 

KietTruongSFDO:
# Critical Changes

# Changes

- We fixed an issue that prevented Opportunity Allocation updates when an Opportunity was both fully scheduled and had a paid Payment. As part of this change, we updated GAU Allocation classes, including ALLO_PaymentSync_SVC.cls. This change is intended for use in a future release and should have no visible impact on your org.
 
# Issues Closed

# New Metadata

# Deleted Metadata


lianeyee:
@KietTruongSFDO Release note revised. Please sanity check?

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -832,6 +832,78 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+                Payment__r.npe01__Paid__c,
+                General_Accounting_Unit__c,
+                General_Accounting_Unit__r.Name,
+                Amount__c,
+                Percent__c
+        from Allocation__c
+        where Payment__c = :payments[1].Id
+        order by Amount__c DESC]);
+
+        System.assert(secondPaymentList.size() == 2, 'The allocation list size is incorrect. It should be 2, but it is actually: ' + secondPaymentList.size());
```
jbrandolinisf: Just reminding you if you did 
System.assertEquals(2, secondPaymentList.size(), 'Allocation List Size Check'); 
It would do the rest for you...

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -832,6 +832,78 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+                General_Accounting_Unit__c,
+                General_Accounting_Unit__r.Name,
+                Amount__c,
+                Percent__c
+        from Allocation__c
+        where Payment__c = :payments[0].Id
+        order by Amount__c DESC]);
+
+        System.assert(paidPaymentAllocationList.size() == 1, 'The allocation list size is incorrect. It should be 1, but it is actually: ' + paidPaymentAllocationList.size());
+        System.assert(paidPaymentAllocationList[0].Amount__c == 500, 'The allocation amount should be $250, but it is actually: ' + paidPaymentAllocationList[0].Amount__c);
```
jbrandolinisf: See comment down further about assertEquals.  Also if you look, the amount you list in your message is incorrect...This is the other reason I prefer the assertEquals approach for ones like this...one less place to change when updating...

src/classes/ALLO_PaymentSync_TEST.cls
KietTruongSFDO: Updated

src/classes/ALLO_PaymentSync_TEST.cls
KietTruongSFDO: Updated

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -832,6 +832,78 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+                Payment__r.npe01__Paid__c,
+                General_Accounting_Unit__c,
+                General_Accounting_Unit__r.Name,
+                Amount__c,
+                Percent__c
+        from Allocation__c
+        where Payment__c = :payments[0].Id
+        order by Amount__c DESC]);
+
+        System.assertEquals(1, paidPaymentAllocationList.size());
```
jbrandolinisf: Should still have the third part. 'Paid Payment Allocation Count' 'Paid Payment Amount', etc.  It just doesn't need to have the expected value as part of it.  E.g.
System.assertEquals(1, paidPaymentAllocationList.size(), 'Payment Payment Allocation Count');

src/classes/ALLO_PaymentSync_TEST.cls
```diff
@@ -832,6 +832,78 @@ private with sharing  class ALLO_PaymentSync_TEST {
...
+        where Opportunity__c = :opportunity.Id
+        AND General_Accounting_Unit__c =: gaus[0].Id
+        order by Amount__c DESC];
+
+        oppAllocation.Percent__c = 50;
+
+        Test.startTest();
+        update oppAllocation;
+
+        List<Allocation__c> paidPaymentAllocationList = new List<Allocation__c>([select Id,
```
jbrandolinisf: I personally prefer saying List, Set, Map, etc. as you did, but that is not our standard.  Our standard is just to pluralize collections e.g. paidPaymentAllocations, or if it were a map paidPaymentAllocationsById

KietTruongSFDO: Updated

src/classes/ALLO_PaymentSync_TEST.cls
KietTruongSFDO: Updated

KietTruongSFDO:
> @KietTruongSFDO Release note revised. Please sanity check?

I updated it from "fully allocated" to "fully scheduled".

lianeyee:
> > @KietTruongSFDO Release note revised. Please sanity check?
> 
> @lianeyee I updated it from "fully allocated" to "fully scheduled".

@KietTruongSFDO Could you clarify what's mean by "fully scheduled"?

KietTruongSFDO:
> > > @KietTruongSFDO Release note revised. Please sanity check?
> > 
> > 
> > @lianeyee I updated it from "fully allocated" to "fully scheduled".
> 
> @KietTruongSFDO Could you clarify what's mean by "fully scheduled"?

@lianeyee  Sure, "fully scheduled" means that the full amount of the donation is scheduled (they could be paid or unpaid). If there's an Opportunity for $1,000 and there's 10 scheduled payments of $100, then it would be fully scheduled. If the amount is that is scheduled is less than the opportunity amount, then it would be under scheduled and the opposite would be over scheduled.

lianeyee:
> > > > @KietTruongSFDO Release note revised. Please sanity check?
> > > 
> > > 
> > > @lianeyee I updated it from "fully allocated" to "fully scheduled".
> > 
> > 
> > @KietTruongSFDO Could you clarify what's mean by "fully scheduled"?
> 
> @lianeyee Sure, "fully scheduled" means that the full amount of the donation is scheduled (they could be paid or unpaid). If there's an Opportunity for $1,000 and there's 10 scheduled payments of $100, then it would be fully scheduled. If the amount is that is scheduled is less than the opportunity amount, then it would be under scheduled and the opposite would be over scheduled.

I guess I'm questioning the terminology. A set of Payments is simply "scheduled." There's no such thing as being partially scheduled versus fully scheduled, is there? I get that the sum of the total Payment amounts may not match the overall Opportunity amount--it's not recommended but it is possible. @markross-sfdo Is "over- and under-scheduling" a thing?

KietTruongSFDO:
> > > > > @KietTruongSFDO Release note revised. Please sanity check?
> > > > 
> > > > 
> > > > @lianeyee I updated it from "fully allocated" to "fully scheduled".
> > > 
> > > 
> > > @KietTruongSFDO Could you clarify what's mean by "fully scheduled"?
> > 
> > 
> > @lianeyee Sure, "fully scheduled" means that the full amount of the donation is scheduled (they could be paid or unpaid). If there's an Opportunity for $1,000 and there's 10 scheduled payments of $100, then it would be fully scheduled. If the amount is that is scheduled is less than the opportunity amount, then it would be under scheduled and the opposite would be over scheduled.
> 
> I guess I'm questioning the terminology. A set of Payments is simply "scheduled." There's no such thing as being partially scheduled versus fully scheduled, is there? I get that the sum of the total Payment amounts may not match the overall Opportunity amount--it's not recommended but it is possible. @markross-sfdo Is "over- and under-scheduling" a thing?

@lianeyee I think partially scheduled and under scheduled would be the same.

PR #4493, opened 2019-08-08
Adding filtering of the Link To Record picklist by object

callen-sfdo:
# Critical Changes

# Changes
Adding filtering to the picklist that is used for the "Link to Record" field on the Mapping Group modal. It will now only allow "Reference" fields that actually look up to the object defined in the Object Name field.  It will still allow Text fields for selection to support those standard objects that cannot be looked up to.

# Issues Closed

# New Metadata

# Deleted Metadata


ptewson-sfdo:
**lurch: attach W-034865

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C8SsQAK" target="_blank">W-034865</a>

PR #4504, opened 2019-08-10
BDI/BGE: update BGE to referenced advance mapping if enabled

ptewson-sfdo:
Previous to change, BGE was reading mapping information used for configuration and setup of a batch directly from the help text, which would render it incompatible with the new advanced mapping BDI engine that stores this configuring in custom metadata types. 

This change refactors the configuration and form setup to use shared behaviors of the new mapping classes, namely "Map<SObjectField, BDI_TargetFields> targetFieldsBySourceField" which is set up similarly for both mapping types. It also improves the responsiveness and performaces of the interface by storing the mapping information as a static variable so that the configuration is loaded once per session and can be referenced. 

Two methods of BDI_DataImportService - getTargetField() and getTargetObject() - are not needed after the change. 

Because the same source DI field can be mapped to multiple target fields, there is a scenario where a picklist could be mapped to two different picklist fields, say one on opportunity and one on payment. Furthermore these target fields could have different picklist options. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


annejackson:
@ptewson-sfdo when I added myself as reviewer it removed everyone else, I tried to assign who I remembered it being assigned to before, but it was so quick I didn't catch all the names. Sorry!

PR #4505, opened 2019-08-12
Add config_platformencryption flow

davidmreed:
This PR introduces a flow, `config_platformencryption`, and a new org type, `beta_platformencryption`, that can be used with `ci_beta` to configure a Beta - Platform Encryption org type.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/orgs/beta_platformencryption.json
+++ b/orgs/beta_platformencryption.json
@@ -1,7 +1,6 @@
 {
   "orgName": "NPSP - Beta Org - Platform Encryption",
   "edition": "Developer",
-  "instance": "cs46",
   "features": [
     "PlatformEncryption"
   ],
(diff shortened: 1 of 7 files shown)


orgs/beta_platformencryption.json
```diff
@@ -0,0 +1,17 @@
+{
+  "orgName": "NPSP - Beta Org - Platform Encryption",
+  "edition": "Developer",
+  "instance": "cs46",
```
davisagli: Is this intentionally a Winter '20 scratch org?

davidmreed: Yes, bu tit doesn't need to stay that way. Winter '20 did not change the Metadata API behavior I wanted it to change.

davidmreed:
Please note: 

- This PR uses the list of encrypted fields from W-025625.
- It appears that roughly 6 unit tests do fail in this configuration. Encrypting more fields results in a significantly larger number of failures. Some of the test failures are CPU timeouts whose relevance is unclear.

PR #4512, opened 2019-08-14
Add Beta Prerelease org

davidmreed:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/orgs/beta_prerelease.json
+++ b/orgs/beta_prerelease.json
@@ -1,0 +1,14 @@
+{
+  "orgName": "NPSP - Beta Test Org",
+  "instance": "cs46",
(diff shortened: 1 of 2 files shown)


orgs/beta_prerelease.json
```diff
@@ -0,0 +1,14 @@
+{
+  "orgName": "NPSP - Beta Test Org",
```
davisagli:
```suggestion
  "orgName": "NPSP - Prerelease Beta Test Org",
```

(so that it can be distinguished when looking at ScratchOrgInfo objects in the dev hub)

PR #4520, opened 2019-08-19
Updated TDTM_ObjectDataGateway class method to fix null pointer error in case trigger action is null.

### LINKED ISSUE #2742: Trigger Handler Null Pointer Exception when the Trigger Action is empty
The trigger action field is not required and can be nulled out by an end user. This will result in an error message for both Health Check and when the trigger is ran on the object: "System.NullPointerException: Attempt to de-reference a null object Class.TDTM_ObjectDataGateway.getClassesToCallForObject: line 74, column 1 External entry point Class.TDTM_TriggerHandler.run: line 89, column 1 Trigger.TDTM_Account: line 34, column 1"

manikrai29:
Updated TDTM_ObjectDataGateway class method (getClassesToCallForObject()) to fix null pointer error in case trigger action is null.

# Critical Changes

# Changes
Added and extra check (th.Trigger_Action__c != null) to make sure Trigger Action is not null and in case it is null we never call that particular class (as user never specified when to call it).

# Issues Closed
#2742

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/TDTM_ObjectDataGateway.cls
+++ b/src/classes/TDTM_ObjectDataGateway.cls
@@ -71,7 +71,7 @@ public without sharing class TDTM_ObjectDataGateway implements TDTM_iTableDataGa
         for (Trigger_Handler__c th : listTH) {
             Set<String> excludedUserNames = (th.Usernames_to_Exclude__c != null ?
                     new Set<String>(th.Usernames_to_Exclude__c.toLowerCase().split(';')) : new Set<String>());
-            if (th.Object__c == objectName && th.Trigger_Action__c.contains(strAction) && th.Active__c == true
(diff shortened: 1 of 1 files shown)


coriobriensfdo:
**lurch: attach W-020423

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000008F6wYAAS" target="_blank">W-020423</a>

coriobriensfdo:
Thanks @manikrai29! As soon as we get the signed CLA back we can move forward on this :)

RadSFDeveloper:
Thank you @manikrai29 for your contribution. This PR will be closed since we have pulled changes into new PR #4564 so it could go through process of review and QA.

PR #4521, opened 2019-08-19
#1795 - OldList searialized.

### LINKED ISSUE #1795: TDTM needs to serialize Old List and New List for asynchronous trigger handler invocations

manikrai29:
# Critical Changes

# Changes
In this fix I am serializing oldList for classes that are called in future context using TDTM framework. Also, I am not serializing newList to prevent unintentional overwriting of fields on the passed sObject. Followed by writing new test cases to check the new functionality.

Updated following files:
src\classes\TDTM_Runnable.cls
src\classes\TDTM_Runnable_TEST.cls
src\classes\TDTM_TriggerHandler.cls

Added following files:
src\classes\ERR_SerializeOldListForAsync_TEST.cls
src\classes\ERR_SerializeOldListForAsync_TEST.cls-meta.xml

# Issues Closed
#1795

# New Metadata
ERR_SerializeOldListForAsync_TEST.cls

# Deleted Metadata

### DIFF
--- a/src/classes/TDTM_Runnable.cls
+++ b/src/classes/TDTM_Runnable.cls
@@ -57,63 +57,63 @@ global abstract class TDTM_Runnable {
      * @description Calls the run method of the class, as a future. To be called dynamically from TDTM_TriggerHandler. 
      * We need this method because Apex won't let us call a non-static method from a separate object instance.
      * @param newIds The IDs of the records that were passed to the trigger as trigger.new.
-     * @param oldIds The IDs of the records that were passed to the trigger as trigger.old.
+     * @param serializedOldList The list of serialized records that were passed to the trigger as trigger.old.
      * @param action The event (before insert, after insert,...) that caused the class to run.
      * @param objectName The name of the SObject the class runs for.
      * @param classToRunName The name of the class to run.
      * @return void
      */
-    public void runFutureNonStatic(Set<Id> newIds, Set<Id> oldIds, String action, String objectName, String classToRunName) {
-        runFuture(newIds, oldIds, action, objectName, classToRunName);
+    public void runFutureNonStatic(Set<Id> newIds, List<String> serializedOldList, String action, String objectName, String classToRunName) {
+        runFuture(newIds, serializedOldList, action, objectName, classToRunName);
     }
     
     /*******************************************************************************************************
      * @description Calls the run method of the class, as a future.
      * @param newIds The IDs of the records that were passed to the trigger as trigger.new.
-     * @param oldIds The IDs of the records that were passed to the trigger as trigger.old.
+     * @param serializedOldList The list of serialized records that were passed to the trigger as trigger.old..
      * @param action The event (before insert, after insert,...) that caused the class to run.
      * @param objectName The name of the SObject the class runs for.
      * @param classToRunName The name of the class to run.
      * @return void
      */
     @future
-    public static void runFuture(Set<Id> newIds, Set<Id> oldIds, String action, String objectName, String classToRunName) {
-        
+    public static void runFuture(Set<Id> newIds, List<String> serializedOldList, String action, String objectName, String classToRunName) {
+
         Schema.DescribeSObjectResult objResult = UTIL_Describe.getObjectDescribe(objectName);
         TDTM_Runnable.Action triggerAction = TDTM_TriggerActionHelper.getTriggerActionFromName(action);
-        
+
         //Get all fields for the object
         List<Schema.SObjectField> allFields = objResult.fields.getMap().values();
-        
+
         //Getting the records the IDs refer to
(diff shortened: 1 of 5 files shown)


src/classes/TDTM_Runnable.cls
```diff
@@ -57,63 +57,63 @@ global abstract class TDTM_Runnable {
...
-        
+
         //Get all fields for the object
         List<Schema.SObjectField> allFields = objResult.fields.getMap().values();
-        
+
         //Getting the records the IDs refer to
-        String qn = 'SELECT ';
-        for (Integer i = 0; i < allFields.size() - 1; i++) { //add all fields, except last, to query
+        String qn = 'select ';
```
force2b: @manikrai29 Thank you for the submission. Very much appreciated. I haven't had a chance to look at the bulk of this PR, however I did notice that you undid our standard styling by changing SOQL to lower case, renaming vars, removing spaces, removing curly brackets, etc.. Can you please revert the style changes and limit the PR just just functional changes that address the issue? That will help us by not having to put these back as well as reduce the review time once one of our Scrum teams is able to pull this in. Thank you.

coriobriensfdo:
**lurch: attach W-008581

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000001gq4AAA" target="_blank">W-008581</a>

coriobriensfdo:
Also resolves Issue #4662

manikrai29:
@coriobriensfdo just curious on what is the status for this pull request?

coriobriensfdo:
@manikrai29 I don't have an update unfortunately, other than it's been checked by our developer/Product Manager teams and is sitting in the backlog. Once a scrum team has a chance to dive deeper into the code and merge it you'll see updates here.

force2b:
@manikrai29 I'm looking back through your PR and have a question about usage. Currently, the `TDTM_Runnable.runFuture()` method is not exposed as usable in customer orgs. Given that we currently do not support asynchronous TDTM handlers for customer-built TDTM, I was wondering if you submitted this PR so that we could support asynchronous handlers (with serialized content) for your use or if there was something else driving this? Basically, if we made this change as submitted you (or any customer) would not be able to use it. What I'm trying to determine is if there is a need to expand async tdtm as a global method (noting that NPSP does not use asynchronous triggers currently)?

manikrai29:
@force2b As far as my understanding goes we don't need TDTM_Runnable.runFuture() method to run them as future methods. All we need to do is 'Asynchronous__c = true' for that particular trigger handler record and then that class is called in future context. So, I opened up this pull request to fix the issue where I needed serialized records in a class running in future context using TDTM.

To back up my findings below is the peice of code that leads to the behaviour I just mentioned above. I got this code from TDTM_TriggerHandler.cls -> runClass() method.


 `if (classInstance instanceof TDTM_Runnable) {
                    TDTM_Runnable classToRun = (TDTM_Runnable)classInstance;
                    Boolean async = Boolean.valueOf(classToRunRecord.get('Asynchronous__c'));
                    Boolean isBefore = (thisAction == TDTM_Runnable.Action.BeforeInsert ||
                         thisAction == TDTM_Runnable.Action.BeforeUpdate || thisAction == TDTM_Runnable.Action.BeforeDelete);

                    //A class can only run asynchronously (calling a future) if we have the ids of the
                    //records we want to affect. This means we can only run it for after in insert. 
                    //Also,a future cannot be called from a future or batch method.
                    **if (async && !isBefore && !System.isFuture() && !System.isBatch()) {
                        UTIL_Debug.debug(LoggingLevel.WARN, '****Calling asynchronously: ' + classToRunName + ' for ' +
                                describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');
                        runAsync(classToRun, classToRunName, newList, oldList, thisAction, describeObj);**
                    } else {
                        UTIL_Debug.debug(LoggingLevel.WARN, '****Calling synchronously: ' + classToRunName + ' for ' +
                                describeObj.getName() + ': ' + thisAction + ' with ' + (newList != null ? newList : oldList).size() + ' records');
                        return classToRun.run(newList, oldList, thisAction, describeObj);
                    }
                }`

force2b:
Thanks @manikrai29. You're absolutely correct. I mis-read the method functionality. The `runAsync` method will call the `runFutureNonStatic` method which will in turn call the `run` method of the trigger handler.

manikrai29:
@force2b can you tell me wats the status of this PR? Its been more than 2 months and no feedback 😄.

force2b:
@manikrai29 Actually that's why I started asking some questions about this. We're looking to get this into an upcoming sprint. Now that I'm clearer on the objectives and changes, I can properly build out the acceptance criteria in our work item backlog.

PR #4523, opened 2019-08-26
Reverts prior change to make $0 Opportunity Amount behave like null Opportunity Amount

### LINKED ISSUE #4384: Default GAU Allocation of $0 not created when $0 Opportunity created
Before #4290 in the 3.157 release, default $0 GAU Allocations would be created when a $0 Opportunity was created. Release 3.157 stopped that default behavior.

jbrandolinisf:
- Reverts prior change to handling of $0 Opportunity Amounts concerning Allocations.
- It will create a default $0 Allocation if Default Allocations are enabled/configured, and the Opportunity Amount is $0.
- It will preserve percentage and fixed amount ($0) allocations if amount is changed from a non zero to a zero amount.
- New Opportunities with a null amount, will not have a default allocation, even if enabled.
- Updating an Opportunity Amount to null, will clear the Opportunities Allocations

# Critical Changes
## GAU Allocations
- We reverted the change made in [Release 3.157](https://github.com/SalesforceFoundation/Cumulus/releases/tag/rel%2F3.157) to GAU Allocation behavior for Opportunities with $0 in the Amount field. If your organization doesn’t create $0 Opportunities, or use GAU Allocations, this change won’t affect your org. If NPSP automatically created any $0 GAU Allocations in your org before the 3.157 release was implemented in Production orgs (July 3), your data may be inconsistent. To check your data, run an Opportunity report with the following filters:
  - Amount = $0
  - Cross Filter for Opportunities **without** GAU Allocations

# Changes

# Issues Closed
- Fixes #4384
# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/ALLO_Allocations_TDTM.cls
+++ b/src/classes/ALLO_Allocations_TDTM.cls
@@ -152,7 +152,7 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
             //INSERT case: Verify associated Campaign and Recurring Donations to auto create allocations
             if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
                 //don't do anything on insert if we don't have an amount
-                if (opp.Amount == null || opp.Amount == 0) {
+                if (opp.Amount == null) {
                     continue;
                 //if this new opportunity has a Campaign or RD, add to list for processing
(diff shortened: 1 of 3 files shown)


beAGoodQE:
Testing is complete.  Merging then deleting this branch.

PR #4548, opened 2019-08-30
Unit Test Fix for UTIL_AbstractCommon_BATCH_TEST

jbrandolinisf:
# Critical Changes

# Changes
- Switch Query Params from Name to Id to guarantee it is always unique, and not to interimitently return unexpected extra account results.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
+++ b/src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
@@ -264,8 +264,8 @@ public class UTIL_AbstractCommon_BATCH_TEST {
     @isTest
     public static void validateValidQuery() {
         // UserInfo.getUserId is not valid for Account, but doesn't matter for this test because we are only validating syntactical correctness.
-        UTIL_AbstractCommon_MOCK tab = new UTIL_AbstractCommon_MOCK(10, 'Id= \'' + UserInfo.getUserId() + '\' or Id=\'' + UserInfo.getUserId() + '\' or Id=\'' + UserInfo.getUserId() + '\'');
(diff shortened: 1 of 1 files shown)


src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -263,9 +263,9 @@ public class UTIL_AbstractCommon_BATCH_TEST {
      */
     @isTest
     public static void validateValidQuery() {
-
-        UTIL_AbstractCommon_MOCK tab = new UTIL_AbstractCommon_MOCK(10, 'Name = \'Test Account #1\' or Name=\'Test Account #2\' or Name=\'Test Account #3\'');
-        tab.setTestWhereClause('Name = \'Test Account #0\' or Name=\'Test Account #1\' or Name=\'Test Account #2\'');
+        // UserInfo.getUserId is not valid for Account, but doesn't matter for this test because we are only validating syntactical correctness.
+        UTIL_AbstractCommon_MOCK tab = new UTIL_AbstractCommon_MOCK(10, 'Id= \'' + UserInfo.getUserId() + '\' or Id=\'' + UserInfo.getUserId() + '\' or Id=\'' + UserInfo.getUserId() + '\'');
```
force2b: Why have a set of OR conditions that are all the same? Would this work with just the following?
`'Id = \' + UserInfo.getUserId() + '\'`

Alternatively you can use 
`'Id = \' + UTIL_UnitTestData_TEST.mockId(Account.SObjectType) + '\'`

jbrandolinisf: Originally it was to test with a more complex query...but it is not really necessary to cover the functionality.

I can cut it down and switch to the mockId...I wasn't aware of that method!

PR #4549, opened 2019-08-30
W-022118 TDTM exclude username case bug

Julian88Tex:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


Safiyab:
@Julian88Tex  Seems like this is working on local but failing on MetaCI. I'd suggest you add a "capture_page_screenshot" and "log_source" command  prior to the failed keyword to see if everything on MetaCI looks as it should be. Coz we cannot merge this until we get at least one successful run on MetaCI.

Julian88Tex:
@Safiyab great will do. I'm guessing you're talking about the robots keyword `Capture Page Screenshot`?

Julian88Tex:
@Safiyab @skristem it appears a reason Feature Test - Robot Unstable is failing is because it's running the `qa_org` flow rather than `dev_org` flow. QA flow has a different layout on trigger handler. Do you know if there is a primary reason why the `qa_org` flow is run vs. `dev_org`?

skristem:
Yes we always use qa_org flow. It has additional metadata and layouts required for QA purposes and also for automation. If you need something that is not present in qa_org flow you might need to talk to RE team member to get it added.

Julian88Tex:
@skristem Okay thanks good to know! I'll check with them about how I can modify a layout in the flow.

Julian88Tex:
@davidmreed Hey, was wondering what would be the best way to determine why a field is missing on a layout based on the CCI flow. Specifically trying to figure out why Usernames_to_Exclude__c would be on the layout for Tigger Handler records on `dev_org` flow but not on `qa_org` flow. Thanks for any help pointing me in the right direction!

robot/Cumulus/tests/draft/tdtm_username_exclude.robot
```diff
@@ -0,0 +1,53 @@
...
+Resource        robot/Cumulus/resources/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Update a Trigger Handler to Exclude a Username
+    [tags]  unstable
+
+    # Create a Contact so Trigger Handler records are created
```
Julian88Tex: Change comment to "Account" based on @skristem 's recommendation.

robot/Cumulus/tests/draft/tdtm_username_exclude.robot
```diff
@@ -0,0 +1,53 @@
...
+    # Navigate to Recurring Donation Trigger Handler Record and exclude Scratch User
+    @{triggerhandler} =          Salesforce Query             Trigger_Handler__c
+    ...                          select=Id
+    ...                          Class__c=RD_RecurringDonations_TDTM
+
+    @{scratchuser} =             Salesforce Query             User
+    ...                          select=Username
+    ...                          Name=User User
+
+    Go To Record Home            ${triggerhandler}[0][Id]
```
Julian88Tex: Use API instead of UI to update Trigger Handler record based on @skristem 's suggestion.

Julian88Tex:
@skristem @Safiyab made several updates to the tests based on suggestions and it should be ready for another review thanks!

Julian88Tex:
@Safiyab Thanks that makes sense... I put that in there since the Keyword will be added to NPSP.robot from another PR... I think it makes sense to make this branch merge into that one instead so I'll do that since it is dependent on it.

Julian88Tex:
@skristem Thanks for the advice! Made those changes.

Unfortunately, I'm facing an issue where all the recent merges are showing as file changes in my PR. Going to be asking RelEng about how to prevent this in the future since it's the second time it's happened to me, but please let me know if my changes to that one .robot file look good. Thanks!

PR #4559, opened 2019-09-04
Advanced Mapping: Prevent non-boolean-style picklists from being mapped to booleans

psadie-sfdo:
- Restrict picklist to boolean mappings. Allow only if the picklist options match one of the following sets: true, false OR yes, no OR 1, 0 OR checked, unchecked, null.

# Critical Changes

# Changes
- We updated Field Mapping in Advanced Mapping so an NPSP Data Import picklist must have boolean values if it’s mapping to a checkbox field. When a user selects a picklist as the NPSP Data Import field that maps to a checkbox on the target object, the picklist options must be True or False.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/bdiFieldMappingModal/bdiFieldMappingModal.js
+++ b/src/lwc/bdiFieldMappingModal/bdiFieldMappingModal.js
@@ -104,7 +104,6 @@ export default class bdiFieldMappingModal extends LightningElement {
         'Multipicklist': ['Multipicklist'],
         'Percent': ['Percent'],
         'Phone': ['Phone', 'String'],
-        // Todo: Add logic to only allow Picklist (True, False, Blank) => Boolean
         'Picklist': ['Picklist', 'Boolean'],
         'Reference': ['Reference', 'String'],
         'String': ['String', 'Picklist'],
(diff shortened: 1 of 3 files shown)


melissabarber:
**lurch: attach W-034860

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C8STQA0" target="_blank">W-034860</a>

src/lwc/bdiFieldMappingModal/bdiFieldMappingModal.js
```diff
@@ -212,23 +211,14 @@ export default class bdiFieldMappingModal extends LightningElement {
...
             this.targetObjectFieldDescribes = event.targetObjectFieldDescribes;
 
-            this.row = event.row;
-
-            if (this.row) {
-                // Edit row
+            if (event.row) {
+                // Edit field mapping
                 this.modalMode = 'edit';
-                this.selectedSourceFieldLabel = this.row.Source_Field_Label;
```
ptewson-sfdo: Just want to make sure I'm reading the intent here - instead of copying the field values over like was happening in this block of removed code, you are parsing the results into the fields? I think I like the switch, if that's the case!

PR #4560, opened 2019-09-04
Feature/robot  winter20fix

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


skristem:
@davidmreed  this is still work in progress

robot/Cumulus/resources/DataImportPageObject.py
```diff
@@ -1,6 +1,6 @@
 from cumulusci.robotframework.pageobjects import ListingPage
 from cumulusci.robotframework.pageobjects import pageobject
-from locators_46 import npsp_lex_locators
+from locators_47 import npsp_lex_locators
```
davisagli: Here and in some other places below, I think this should be `from NPSP import npsp_lex_locators` rather than referring to a specific version. The idea is that the NPSP module has a `npsp_lex_locators` dict which it takes care of updating to hold the correct set of locators when the NPSP library is initialized. Then other modules can import the dict from NPSP without needing to know what version it is.

robot/Cumulus/resources/NPSP.py
```diff
@@ -739,6 +747,8 @@ def click_button_with_value (self,title):
         """clicks on the button on the payments page"""      
         locator=npsp_lex_locators['button'].format(title)
         self.selenium.get_webelement(locator).click()
+        if title == "Create Payments":
+            self.selenium.wait_until_page_does_not_contain_element(locator, error="Clicking Button "+ title +" was not successful")
```
davisagli: You can just do `Wait Until Page Does Not Contain Element | title=Create Payments` after the line in the test suite that calls this keyword. I think it would be better to keep this generic keyword clean.

robot/Cumulus/resources/DataImportPageObject.py
skristem: made suggested change

robot/Cumulus/resources/NPSP.py
skristem: removed this and I am using Wait Until Page Does Not Contain Element with xpath

davisagli:
@skristem the "Create Relationships for contacts" test is consistently failing on the prerelease org when run in MetaCI but not when I run it locally. I'm trying to figure out why.

PR #4561, opened 2019-09-04
RD2: Auto-populate Recurring Donation Name

RadSFDeveloper:
# Critical Changes

# Changes
- The Recurring Donation Name will be auto-generated on the record creation when "Recurring Donations Settings" "Enable Automatic Naming"  field is checked. Automatic naming is disabled by default.
   - The format for automatically named Recurring Donations is is: `{!DonorName} {CurrencySymbol}{!Amount} - Recurring`. When naming is enabled for a multi-currency org, the name format is: `{!DonorName} {ISO Code} {!Amount} - Recurring`. Future releases will support customizing the name format.

# Issues Closed

# New Metadata

Apex Class:
- RD2_Donor
- RD2_Donor_TEST
- RD2_NamingService
- RD2_NamingService_TEST
- TEST_RecurringDonationBuilder
- TEST_SObjectGateway

Custom Label:
- RecurringDonationNameSuffix

Custom Field:
- npe03__Recurring_Donations_Settings__c.EnableAutomaticNaming__c

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-034367

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C7TAQA0" target="_blank">W-034367</a>

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -14,4 +14,14 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>EnableAutomaticNaming__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>Indicates if naming is automated. When multi-currency is enabled, the name format is: {!DonorName} {ISO Code} {!Amount} - Recurring. Otherwise, {!DonorName} {CurrencySymbol}{!Amount} - Recurring.</description>
```
RadSFDeveloper: @sam-knox , can you please advise on the "Enable Automatic Naming" Description and Help Text?

src/classes/RD_RecurringDonations_TDTM.cls
```diff
@@ -71,6 +72,8 @@ public class RD_RecurringDonations_TDTM extends TDTM_Runnable {
             newlist = oldlist;
         }
 
+        autogenerateNames(newlist, oldlist, triggerAction);
```
RadSFDeveloper: @force2b , the current version will have automated naming applied. When the new RD version is enabled, the enablement and RD2 infrastructure will be developed in W-034362, so we should ensure naming is applied there. Let me know if there is any objection.

RadSFDeveloper:
@sam-knox , since the field is already in the context of Recurring Donations custom settings, the repetition of "RecurringDonation" in the API name might not be necessary? 
 If we follow the label to name the field, it would be `npe03__Recurring_Donations_Settings__c.AutomaticRecurringDonationNaming__c`. What name otherwise could we use so it is concise but still descriptive and aligns to the label change?

Currently used field name: `npe03__Recurring_Donations_Settings__c.EnableAutomaticNaming__c`

sam-knox:
@RadSFDeveloper - Yes, go ahead and keep the API name the way it was, but please do change the label for it. Thanks!

src/classes/RD2_NamingService.cls
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        if (isMultiCurrencyEnabled) {
+            String code = (String) rd.get(CURRENCY_ISO_CODE_FIELD);
+
+            if (String.isNotBlank(code)) {
+                formattedAmount = code + ' ' + formattedAmount;
+            }
+        } else {
+            formattedAmount = userDefaultCurrencySymbol + formattedAmount;
```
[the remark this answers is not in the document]
RadSFDeveloper: Line s 181-185 will add the record currency which is applicable when multi-currency is enabled. When it is not enabled, there is no record currency and the amount is displayed in the user default currency.

src/classes/UTIL_Currency.cls
```diff
@@ -35,6 +35,8 @@
 */
 public without sharing class UTIL_Currency implements UTIL_Currency.Interface_x {
 
+    public static final String CURRENCY_ISO_CODE_FIELD = 'CurrencyIsoCode';
```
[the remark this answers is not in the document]
RadSFDeveloper: Instead of using text value throughout code, it is recommended to use a Constant. If there is a text value that is a misspelled, and there is no unit test, the error will be generated in run time. Using a constant removes run-time errors.

src/objects/npe03__Recurring_Donations_Settings__c.object
RadSFDeveloper: @sam-knox ^

sam-knox: Ok, how about something more simple like:

"If enabled, Salesforce automatically populates the Recurring Donation Name." 

I assume that the name format configuration will eventually be on the same NPSP Settings page as this setting.

RadSFDeveloper: The Description and Help Text have been updated.

src/classes/RD2_NamingService.cls
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }   
+
+    /***
+    * @description Constructs recurring donation name
+    * Name formatting: 
+    *   If multi-currency, then "{!DonorName} {ISO Code} {!Amount} - Recurring".
+    *   Otherwise, "{!DonorName} {User Default Currency Symbol}{!Amount} - Recurring".
+    */
+    private String constructName(npe03__Recurring_Donation__c rd,  Map<Id, RD2_Donor.Record> donorById) {
```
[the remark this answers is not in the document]
RadSFDeveloper: As discussed, we will work on configurable naming in another WI. The code is ready for review again. @sam-knox , @force2b

src/classes/UTIL_Currency.cls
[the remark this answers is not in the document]
force2b: Understood, noting that there are dozens of uses of `CurrencyIsoCode` as s string throughout the code base already.

RadSFDeveloper:
@Julian88Tex , automatic naming on RDs is ready for testing.

src/classes/ACCT_Accounts_TDTM.cls
```diff
@@ -61,12 +61,38 @@ public class ACCT_Accounts_TDTM extends TDTM_Runnable {
...
+
+        for (Account acc : (List<Account>) accounts) {
+            if (acc.npe01__SYSTEM_AccountType__c != CAO_Constants.HH_ACCOUNT_TYPE
+                && acc.Name != oldAccountById.get(acc.Id).get('Name')
+            ) {
+                accountIds.add(acc.Id);
+            }
+        }
+
+        new RD2_NamingService().updateOnAccountChange(accountIds);
```
force2b: There is a potential - though small - for this synchronous action to cause issues with governor limits such as SOQL, DML or CPU timeout as well as just general performance. I recommend implementing support for an asynchronous method in the naming service and use that for all naming changes that originate from the Account or Contact object.

src/classes/RD2_NamingService.cls
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        List<npe03__Recurring_Donation__c> rds = new List<npe03__Recurring_Donation__c>();
+
+        for (npe03__Recurring_Donation__c rd : rdGateway.getRecordsReferencingContacts(contactIds)) {
+            rd.Name = System.Label.npo02.NameReplacementText;
+            rds.add(rd);
+        }
+
+        if (!rds.isEmpty()) {
```
force2b: Is there a need to actually push the change to the Name through as DML? Couldn't this method just call the `autogenerateNames()` method with the list of RD's to rename? That would save on DML and processing time.

src/classes/ACCT_Accounts_TDTM.cls
RadSFDeveloper: Updating RD names in a queueable is causing many HH and Contact test classes to fail due to "Max stack depth  has been reached"  and "Too many queueable jobs have been added to the queue: 2".
Also, when a HH naming batch is run, it seems better to update RD name instead creating queueable job for every `execute()` in the batch.

RadSFDeveloper:
@force2b , auto-generation of RD name on the record creation PR is ready for the code review.

force2b:
@RadSFDeveloper Looks like there's a merge conflict on this PR.

RadSFDeveloper:
@force2b , merge conflict has been resolved.

src/classes/RD2_RecurringDonationsGateway.cls
```diff
@@ -0,0 +1,131 @@
...
+*
+*/
+public with sharing class RD2_RecurringDonationsGateway {
+
+    /***
+    * @description Retrieves Recurring Donation records for specified Account Ids
+    * @param accountIds Account Ids
+    * @return List<npe03__Recurring_Donation__c>
+    */
+    public List<npe03__Recurring_Donation__c> getRecordsReferencingAccounts(Set<Id> accountIds) {
```
force2b: I'm not necessarily seeing these method used outside of a unit test. Are these left overs from the renaming/merging logic that was removed?

Out of curiosity, could these two methods be merged into a single `getRecordsForAccountOrContact()` method that still accepts `Set<Id>` and filters for `npe03__Organization__c IN :parentIds OR npe03__Contact__c IN :parentIds`? Mainly I'm worried about both methods being called in the same execution context (i.e., two queries). Plus the methods are very similar.

src/classes/RD2_RecurringDonationsGateway.cls
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /***
+    * @description Retrieves Recurring Donation records for the specified Ids
+    * @param rdIds Ids of the recurring donations
+    * @return List<npe03__Recurring_Donation__c>
+    */
+    public List<npe03__Recurring_Donation__c> getRecords(Set<Id> rdIds) {
+        return [
+            SELECT
```
force2b: Are there other fields that should be in this query, or better yet should it query *all* fields on the object? It probably depends on usage, but I get the impression this is a general selector purpose method to retrieve RDs.

src/classes/RD2_RecurringDonationsGateway.cls
```diff
@@ -0,0 +1,131 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Recurring Donations Gateway to return data
+*
+*/
+public with sharing class RD2_RecurringDonationsGateway {
```
force2b: I'm not sure this method should be declared as `with sharing`. It's feasible this could be called via a trigger or batch job that should be executing at the system context. I recommend using `inherited sharing` instead.

src/classes/RD2_RecurringDonationsGateway_TEST.cls
```diff
@@ -0,0 +1,141 @@
...
+        Account acc = new Account(Name = TEST_NAME);
+        insert acc;
+
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        contact.FirstName = TEST_NAME;
+        insert contact;
+
+        insert new List<npe03__Recurring_Donation__c>{
+            new TEST_RecurringDonationBuilder()
+                .withName('foo')
```
force2b: Same comment here about not specifying a hardcoded non-unique RD name.

src/classes/TEST_RecurringDonationBuilder.cls
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /***
+    * @description Sets default values
+    * @return TEST_RecurringDonationBuilder Builder instance
+    */
+    public TEST_RecurringDonationBuilder withDefaults() {
+        return this
+            .withAmount(2)
+            .withInstallments(2)
```
force2b: `withInstallments` should not be in defaults because the definition of the field is that it only applies to `Fixed` length RD's. 

Overall this entire class has to be able to differentiate the data between RD1 and RD2 format recurring donations. There are a few ways to implement this. One might be new methods for `useRD1format` (the default) and `useRD2format`. Another is to check the RecurringDonationSettings.IsRd2Enabled value in the constructor and set a private var accordingly.  A third might be to use different methods such as `withDefaultsRD1()` and `withDefaultsRD2()`. My preference is the first option, but we'd need to make sure that all RD2 tests call `useRDformat()`.

In another comment I mentioned throwing an exception (could be `system.assert(false)` call too) if an RD1 only method (such as `withInstallments()` is called when in RD2 format mode. I think this would be very helpful in ensuring that RD test data is properly constructed for test methods.

src/classes/TEST_RecurringDonationBuilder.cls
```diff
@@ -0,0 +1,363 @@
...
+    public TEST_RecurringDonationBuilder withScheduleTypeMultiplyValue() {
+        valuesByFieldName.put('npe03__Schedule_Type__c', System.Label.npe03.RecurringDonationMultiplyValue);
+        return this;
+    }
+
+    /***
+    * @description Sets Open Ended Status to Open Status
+    * @return TEST_RecurringDonationBuilder Builder instance
+    */
+    public TEST_RecurringDonationBuilder withOpenEndedStatusOpen() {
```
force2b: Since `npe03__Open_Ended_Status__c` is an RD1 only field, there will need to be other methods for `withRecurringDonationType` and `withStatus` that update the RD2 specific fields. Plus, as I noted before if these methods are called for an RD2 format RD an exception should be thrown.

src/classes/RD2_Donor.cls
```diff
@@ -0,0 +1,118 @@
...
+* @group Recurring Donations
+* @description Recurring Donations Donor record and gateway classes
+*
+*/
+public with sharing class RD2_Donor {
+
+    /***
+    * @description Donor Gateway interface
+    */
+    public interface IGateway {
```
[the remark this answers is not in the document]
RadSFDeveloper: Cannot use Stub since the `RD2_Donor.Gateway` is an inner class. I can create it as another class and use the Stub, however we can create a test Mock by implementing interface and achieve the same result.

src/classes/TEST_RecurringDonationBuilder.cls
```diff
@@ -0,0 +1,363 @@
...
+        return this;
+    }    
+
+    /***
+    * @description Sets number of Installments
+    * @param numberOfInstallments Number of Installments
+    * @return TEST_RecurringDonationBuilder Builder instance
+    */
+    public TEST_RecurringDonationBuilder withInstallments(Integer numberOfInstallments) {
+        valuesByFieldName.put('npe03__Installments__c', numberOfInstallments);
```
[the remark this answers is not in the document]
RadSFDeveloper: - There are many moving parts, so the RD builder is creating the same fields on the record in tests as is currently on `master`. If we remove number of `npe03__Installments__c` value for specific tests where OpenEndedStatus is not Closed, they start failing. Should we create another WI/PR to address this issue in tests that currently exists in `master`?
- There is a new method `validateMode()` that ensures fields applicable to legacy RDs are not created when enhanced RDs are enabled.

src/classes/RD2_RecurringDonationsGateway.cls
RadSFDeveloper: These left overs from the renaming/merging logic that was removed. Since RD retrieval as of current WIs in progress is applied in unit tests, I've created `TEST_SObjectGateway` to be used to retrieve data in tests. (Sorry for multiple comment emails, some comments are not visible In-Review mode).

src/classes/RD2_RecurringDonationsGateway.cls
RadSFDeveloper: The class has been removed and replaced with test `TEST_SObjectGateway` in order to retrieve test records. Different fields based on the Enhanced RD enablement will be done in the child branch PR.

src/classes/TEST_RecurringDonationBuilder.cls
RadSFDeveloper: I have removed `withDefaults()` and moved specific RD field creation in specific test classes.

src/classes/RD2_Donor.cls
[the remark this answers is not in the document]
force2b: Agreed

src/classes/RD2_RecurringDonationsGateway.cls
force2b: Yes, that can work.

RadSFDeveloper:
@Julian88Tex , the PR is ready for testing. Thanks for your work!

src/classes/RD2_NamingService.cls
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        if (isMultiCurrencyEnabled) {
+            String code = (String) rd.get(CURRENCY_ISO_CODE_FIELD);
+
+            if (String.isNotBlank(code)) {
+                formattedAmount = code + ' ' + formattedAmount;
+            }
+        } else {
+            formattedAmount = userDefaultCurrencySymbol + formattedAmount;
```
force2b: @RadSFDeveloper: @Julian88Tex made a very good suggestion. We made this change to the OppNaming service to add a space between the _symbol_ and the amount when the symbol is more than 1 character in length. It very much helps with readability. @Julian88Tex recommended that we make the same change here as well. Here's an example of what it could generate with and without the space. The second bullet is much easier to read than the first.
- Without: ₨100
- With: ₨ 100

```suggestion
            if (!String.isBlank(userDefaultCurrencySymbol) && userDefaultCurrencySymbol.length() > 1) {
                formattedAmount = userDefaultCurrencySymbol + ' ' + formattedAmount;
            } else {
                formattedAmount = userDefaultCurrencySymbol + formattedAmount;
            }
```

RadSFDeveloper: @force2b , the change has been applied. Can you please review again (note package.xml is fixed by `cci` deploy).
@sam-knox , can you please review PR's Changes above and check if they need to be modified based on the latest requirement?

force2b:
Merging to take advantage of the weekend build quiet time

PR #4563, opened 2019-09-04
RD2: CCI Task to Deploy Unmanaged Metadata Field & Layout Changes

force2b:
CCI Task (`deploy_rd2_config`) to install the following metadata components into an org explicitly for orgs that enable Enhanced Recurring Donations. This will work for all orgs with both Unmanaged NPSP and the Managed NPSP Package. It will be the task executed by the MetaDeploy page once it is ready.

- Enhanced Recurring Donations Layout: Assigned as the Default layout for the Admin and StandardUser profiles. This new layout is specific to the RD2 pilot and allows Users to interact with the Recurring Donations record using the fields relevant to RD2. All deprecated fields have been removed from this layout.
- Help Text Changes: Change or Add InlineHelpText for the following fields to better reflect usage in RD2:
   - npsp__Day_of_Month__c
   - npe03__Installment_Period__c
   - npe03__Amount__c
   - npe03__Date_Established__c
   - npe03__Last_Payment_Date__c
   - npe03__Paid_Amount__c
   - npe03__Total_Paid_Installments__c
   - npe03__Next_Payment_Date__c

* Picklist Value Changes to add or inactivate picklist values as necessary for RD2:
   - npsp__Day_of_Month__c
   - npe03__Installment_Period__c

- Field Label Changes: Add **Deprecated** to the field label for all these fields that are no longer used with RD2:
   - npe03__Installment_Amount__c
   - npe03__Installments__c
   - npe03__Open_Ended_Status__c
   - npe03__Schedule_Type__c
   - npe03__Total__c
   - npsp__Always_Use_Last_Day_Of_Month__c

* FLS Changes: The following are all changed to *Read-Only* for the Admin and StandardUser profiles
   - npe03__Paid_Amount__c
   - npe03__Total_Paid_Installments__c
   - npe03__Installment_Amount__c
   - npe03__Installments__c
   - npe03__Open_Ended_Status__c
   - npe03__Schedule_Type__c
   - npe03__Total__c
   - npsp__Always_Use_Last_Day_Of_Month__c

* FLS Changes: The following new RD2 fields all changed to *Editable* for the Admin and StandardUser profiles
   - npsp__Day_of_Month__c
   - npsp__InstallmentFrequency__c
   - npsp__RecurringType__c
   - npsp__PaymentMethod__c
   - npsp__StartDate__c
   - npsp__Status__c

There are no changes at all to the packaged metadata in this pull request.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
**lurch: attach W-035252

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9K9QAK" target="_blank">W-035252</a>

scripts/validate_rd2_enabled.cls
```diff
@@ -0,0 +1,20 @@
+/**
+* @description Validates that Recurring Donations 2 has been enabled in an org. Throws an exception if it hasn't been.
+* Note that this needs to work when a full Managed NPSP package is installed in an org.
+* @author Michael Smith
+* @date 2019-09-17
+*/
+public class SettingsException extends Exception {}
+
+public static Boolean validateRd2Enabled() {
```
force2b: @davisagli I wasn't quite sure what the task should return to confirm that RD2 is properly enabled. For this draft I just throw an exception if it's not and return true if it is. Is that the best way to handle the validation?

cumulusci.yml
```diff
@@ -291,6 +291,20 @@ tasks:
...
                 STG_SettingsService.stgService.saveAll();
 
+    deploy_rd2_unmanaged:
+        description: Deploy unmanaged field and layout changes into the org that are needed after enabling RD2 in an org
+        class_path: cumulusci.tasks.salesforce.Deploy
+        options:
+            path: unpackaged/config/rd2_post_config
+            namespace_inject: $project_config.project__package__namespace
+
+    validate_rd2_enabled:
```
force2b: @davisagli  Now that I'm looking at this, I suspect that the actions in this task should be part of the `deploy_rd2_unmanaged` flow, so that it first executes the validation, and if that doesn't throw an exception, it then executes the deployment. Is that correct?

scripts/validate_rd2_enabled.cls
davisagli: Hmm I think I'll need to wrap it in a special Python task that turns exception-or-no-exception into False/True for the Python return value.

cumulusci.yml
davisagli: You can do that if you want for when you're testing using the CLI, but it'll end up structured a little differently when I configure it for MetaDeploy. Preflight checks get configured separately from flow steps because they run at a different time. i.e. the user clicks the button to run pre-validation checks, then has a chance to review the list of steps for the actual plan (which may have been adjusted based on pre-validation), then clicks the button to run the plan.

force2b:
@sam-knox  Thanks for the review. I'm not sure what is meant by these lines: "Line 208, change the to match the", etc.

The changes to match the base package cannot be made at this time because the beta/production sync issue I noted in Chatter. Those changes will have to be held until the RD package is cut for release next month.

force2b:
@sam-knox Also, if the line 28 comment refers to the `Description` attribute of the`Day_of_Month__c` field, that change cannot be made here. We'd have to add the period to the Description of that field in the Recurring Donations base package.

sam-knox:
@force2b  - Ah geez, I tried to include description and inlinehelptext with brackets, but git rendered them as if they were HTML so they didn't show up. Those lines were supposed to read: 

Line 208, change the description to match the help text.

But I understand about the difference in repos. I was just trying to cover all the bases around these text strings. So if the notes are related to the base package (which I think many of them are), you can ignore them here.

force2b:
Thanks @sam-knox. _Description_ is a field attribute that cannot be changed here because it's not Subscriber Editable. Also, we are changing the InlineHelpText here so that the field better reflects how the field is used in an RD2 context. If we change the Description on those fields in the RecurringDonations repository to match the RD2-only help text, then it would be out of sync for RD1 users. For the few fields that we're overriding the HelpText on for RD2, the Description and HelpText will need to be different.

sam-knox:
@force2b - Can we enable the NPSP Compact Layout as well in the new RD2 layout? It might need adjustments per the new fields too.

force2b:
@sam-knox Thanks for the feedback. Here are my responses.

1. Does Payment Method really need to be required? 
  -- Can there a "Prospective Recurring Donation"? I changed the field to not-required, but let's talk about that more when we get to the custom UI.
2.  I removed Donor Name on the layout. Good catch
3.  Next Donation Date has been moved.
4.  "Deprecated" is on the "Always Use Last Day of Month" field already. 
  -- You should see this in your org, though keep in mind that you'd have to view the field on a report, list view or layout to see the label. It won't show up with the label looking at the object or in Page Layout Editor. This is because we "overrode" the label using translations.
5.  There is another work item to handle the Account/Contact fields.
6.  The Page Layout editor will always show the packaged label even if that label has been overridden through translations (which is what we did to add "Deprecated" to them).
7. The Compact Layout is a good suggestion. I'll take a look. Keep in mind that this page layout is primarily meant for Pilot-1 only and may not actually live on past the pilot since we'll be building our own custom LWC view/edit page for Recurring Donation records.

I'll update your scratch org shortly with those changes.

sam-knox:
Thanks @force2b - and I see the Always Use Last Day of Month translation now. I had to enable English in Translation Settings to get it to appear. Prior to that I was seeing only the Master Label for that field.

force2b:
@sam-knox Thanks for the feedback. Here are my responses.

1. Does Payment Method really need to be required? 
  -- Can there a "Prospective Recurring Donation"? I changed the field to not-required, but let's talk about that more when we get to the custom UI.
2.  I removed Donor Name on the layout. Good catch
3.  Next Donation Date has been moved.
4.  "Deprecated" is on the "Always Use Last Day of Month" field already. 
  -- You should see this in your org, though keep in mind that you'd have to view the field on a report, list view or layout to see the label. It won't show up with the label looking at the object or in Page Layout Editor. This is because we "overrode" the label using translations.
5.  There is another work item to handle the Account/Contact fields.
6.  The Page Layout editor will always show the packaged label even if that label has been overridden through translations (which is what we did to add "Deprecated" to them).
7. The Compact Layout is a good suggestion. I'll take a look. Keep in mind that this page layout is primarily meant for Pilot-1 only and may not actually live on past the pilot since we'll be building our own custom LWC view/edit page for Recurring Donation records.

The Standard and Compact Layout changes should be in your scratch org.changes. I also added the new "Starting Date" field to the layout.

sam-knox:
@force2b - I think my scratch org has expired. It's been more than 7 days.

force2b:
@sam-knox Good catch. Changes made.

PR #4566, opened 2019-09-05
Feature/robot  axe

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


robot/Cumulus/tests/browser/axe_test/first_axe_test.robot
```diff
@@ -0,0 +1,23 @@
+*** Settings ***
+
+Resource        robot/Cumulus/resources/NPSP.robot
+Suite Setup     Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Test Cases ***
+
+Analyze account object home page
+    Inject Axe Core Library
```
spelak-salesforce: I see `Inject Axe Core Library` in a lot of places.  Is it possible to call this command in `Suite Setup` so `Inject Axe Core Library` is done once and the **Test Cases** don't need to worry about it?

spelak-salesforce: Similar comment, can the following commands be combined into a single command to make it a one-line to check for axe violations?

```robot
** combine these**
    &{acc_results}    Get Axe Analysis Results
    Log Summary Of Results    ${acc_results}
    Warn On Incomplete Rules    ${acc_results}
    Warn On Violations Rules    ${acc_results}

** so we can instead write something like **
    Expect No Axe Viloations ${element_to_test}
```

robot/Cumulus/resources/NPSP.robot
```diff
@@ -386,6 +387,14 @@ Click Field And Select Date
     [Arguments]    ${field}    ${date}
     Click Element With Locator    bge.field-input    ${field}    
     Click BGE Button    ${date}    
+    
+Run Accessibility Check
```
spelak-salesforce: I love combining the individual methods into a single call so it's a one-liner to run an accessibility check.  Is it possible to have this `Run Accessibility Check` function/method defined "globally" so each robot test doesn't have to re-create it.  e.g. some setup file defines this method, and all robot tests use the common setup file.

robot/Cumulus/resources/NPSP.py
```diff
@@ -1166,13 +1171,134 @@ def click_wrapper_related_list_button(self,heading,button_title):
...
+
+
+
+    def inject_axe_core_library(self):
+        path = os.path.dirname(os.path.abspath(__file__))
+        self.selenium.execute_javascript(path + "/axe.js")
+    
+    
+    def get_axe_analysis_results(self,cssSelector = None):    
+        context = "document"
```
spelak-salesforce: It's possible to call [axe.run(...)](https://www.deque.com/axe/axe-for-web/documentation/api-documentation/#api-name-axerun) on a specific HTML DOM Element.  I see here this method checks the whole `document`.   It's possible there could be axe violations in parts of the `document` that a developer cannot control, e.g. Salesforce header bar.

Is it possible to add a robot method where `axe.run` could be called for a specified element such as through a given CSS selector?

force2b:
Hi @skristem Is this branch and PR still needed? It looks like it goes back almost 2 years.

skristem:
closing this PR as the changes are not needed anymore

PR #4567, opened 2019-09-06
Adding ability to define one field for matching for additional objects

callen-sfdo:
For the BDI Additional objects functionality this allows a user to define one field mapping for use in matching per Object Group.  This also includes creating an overrideable class BDI_ObjectMatchingLogic, which contains the logic for matching and in the future will be able to be replaced by client's custom logic.

In addition, this moves the log error functionality into the object wrapper in order to streamline the code and enable errors to be attached from outside of the Additional Object Service.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


melissabarber:
**lurch: attach W-033233

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087uviQAA" target="_blank">W-033233</a>

src/labels/CustomLabels.labels
```diff
@@ -2760,6 +2760,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>status on BDI object when user selected opp in BGE interface for new payment</shortDescription>
         <value>Apply New Payment</value>
     </labels>
+    <labels>
+        <fullName>bdiObjMatchErrorMultipleMatch</fullName>
+        <categories>Advanced Mapping</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when there are multiple matches returned</shortDescription>
+        <value>Error: Multiple matching records found for value of matching field:</value>
```
```suggestion
        <value>Error: Multiple records match the value in the unique ID field:</value>
```

src/labels/CustomLabels.labels
```diff
@@ -2760,6 +2760,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+    </labels>
+    <labels>
+        <fullName>bdiObjMatchErrorTooManyMatchingFields</fullName>
+        <categories>Advanced Mapping</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when multiple matching fields are selected</shortDescription>
+        <value>Error: Only one field mapping can be used for matching per Object Group.</value>
```
```suggestion
        <value>Error: Only one unique ID field can be used for matching per Object Group.</value>
```

src/objects/Data_Import_Field_Mapping__mdt.object
```diff
@@ -17,14 +17,40 @@
...
+        <description>When checked, indicates this record was deleted from the Data Import Advanced Mapping user interface and will be ignored during NPSP Data Import processing.</description>
         <externalId>false</externalId>
         <fieldManageability>SubscriberControlled</fieldManageability>
         <inlineHelpText>When checked, indicates this record was deleted from the Data Import Advanced Mapping user interface and will be ignored during NPSP Data Import processing.</inlineHelpText>
         <label>Is Deleted</label>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>Match_On_This_Mapping__c</fullName>
+        <description>Indicates whether this field mapping should be used for matching.</description>
```
melissabarber:
```suggestion
        <description>Indicates if this field mapping should be used for matching a unique ID.</description>
```

src/objects/Data_Import_Field_Mapping__mdt.object
```diff
@@ -17,14 +17,40 @@
...
+        <externalId>false</externalId>
+        <fieldManageability>SubscriberControlled</fieldManageability>
+        <inlineHelpText>Indicates whether this field mapping should be used for matching.</inlineHelpText>
```
melissabarber:
```suggestion
        <inlineHelpText>Indicates if this field mapping should be used for matching a unique ID.</inlineHelpText>
```

src/objects/Data_Import_Object_Mapping__mdt.object
```diff
@@ -13,6 +13,18 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Custom_Matching_Logic_Class__c</fullName>
+        <description>If custom apex logic should be used to map the fields for this object, then this field should be populated with the name of a class (including namespace if needed) that extends the BDI_ObjectMatchingLogic class and overrides the matchObjects method.</description>
```
melissabarber: @callen-sfdo, should "map" at the beginning be "match" instead so it reads: 
```suggestion
        <description>If custom apex logic should be used to match the fields for this object, then this field should be populated with the name of a class (including namespace if needed) that extends the BDI_ObjectMatchingLogic class and overrides the matchObjects method.</description>
```

src/objects/Data_Import_Object_Mapping__mdt.object
```diff
@@ -13,6 +13,18 @@
...
+        <externalId>false</externalId>
+        <fieldManageability>SubscriberControlled</fieldManageability>
+        <inlineHelpText>If custom apex logic should be used to map the fields for this object, then this field should be populated with the name of a class (including namespace if needed) that extends the BDI_ObjectMatchingLogic class and overrides the matchObjects method.</inlineHelpText>
```
melissabarber:
```suggestion
        <inlineHelpText>If custom apex logic should be used to match the fields for this object, then this field should be populated with the name of a class (including namespace if needed) that extends the BDI_ObjectMatchingLogic class and overrides the matchObjects method.</inlineHelpText>
```

src/classes/BDI_ObjectMatchingLogic.cls
```diff
@@ -0,0 +1,156 @@
...
+*/
+
+public with sharing virtual class BDI_ObjectMatchingLogic {
+
+    public virtual BDI_ObjectWrapper[] matchObjects(BDI_ObjectWrapper[] objWraps, BDI_DataImportService dataImportService) {
+        if (objWraps != null && objWraps.size() > 0 && objWraps[0].fieldMappings != null) {
+
+            Boolean shouldMatch = false;
+
+            //Note it is assumed that all object wraps passed into this class will be using the same
```
ptewson-sfdo: Just a question, are there any scenarios where ware currently calling this that the wrappers might be using different field mappings?

src/classes/BDI_ObjectMatchingLogic.cls
```diff
@@ -0,0 +1,156 @@
...
+                for (BDI_ObjectWrapper objWrap : objWraps) {             
+                    objWrap.logError(dataImportService.statusDIFailed(),System.label.bdiObjMatchErrorTooManyMatchingFields);
+                }
+            } else if (matchSourceCount == 1) {
+
+                // If the source field for the match on the DI has a value then we will attempt to match.
+                // also add the source strings for matching to a list.
+                for (BDI_ObjectWrapper objWrap : objWraps) {
+
+                    String sourceValue = (String)objWrap.dataImport.get(matchSourceField);
```
ptewson-sfdo: I can't remember, are we restricting type of the matching field at all? Is there any reason that casting to string here would be a problem depending on the type of the field?

src/classes/BDI_DataImportService.cls
```diff
@@ -572,7 +572,7 @@ global with sharing class BDI_DataImportService {
             importCampaignMembers();
             pl.stop();
 
-            if (additionalObjectService != null && isDryRun == false) {
+            if (additionalObjectService != null) {
```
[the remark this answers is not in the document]
callen-sfdo: Yes it was intentional, I think it was a mistake to have it there in the first place.  Previously not much would have happened in Additional Object Service anyway during dry run (besides checking dependencies), but now that we are doing matching for additional objects during dry run it has to be able to reach the matching logic.

src/classes/BDI_AdditionalObjectService.cls
```diff
@@ -141,19 +144,30 @@ public with sharing class BDI_AdditionalObjectService {
 
                             //If the predecessor doesn't exist, then update the Data import object with an error 
                             //and continue with the next dataImport.
-                            if (!predecessorExists(objWrap,predecessor)) {
-                                continue;
+                            if (!dataImportService.isDryRun) {
```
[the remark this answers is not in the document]
callen-sfdo: The reason I put this isDryRun check in here is because the predecessor check will always fail during dry run since the predecessor objects are not created during dry run.

src/classes/BDI_ObjectMatchingLogic.cls
callen-sfdo: Not currently but I could just change the signature to require an object mapping to be passed, which would eliminate any confusion about which settings to use.  I'll go ahead and make the change unless you think we should take a different route.

src/classes/BDI_ObjectMatchingLogic.cls
callen-sfdo: On the UI side I was planning on limiting it to be datatypes that could be converted to string easily, and were not long text (ie Text, Lookup, Id).  We can expand or change the options as needed, but Ali was ok with these options for now.

src/classes/BDI_ObjectMatchingLogic.cls
ptewson-sfdo: I think that change makes sense - I like it being explicit and clear as much as possible

PR #4568, opened 2019-09-06
Prevent Deletion of Recurring Donations with Paid Payments

KietTruongSFDO:
# Critical Changes

# Changes
- We have made an update which prevents Recurring Donations from being deleted if there are any paid Payments associated. 
# Issues Closed

# New Metadata

# Deleted Metadata


markross-sfdo:
@KietTruongSFDO Is this linked to an issue/WI?

KietTruongSFDO:
**lurch: attach W-035613

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CA1QQAW" target="_blank">W-035613</a>

src/classes/RD_CascadeDeleteLookups_TDTM.cls
```diff
@@ -233,7 +257,8 @@ public with sharing class RD_CascadeDeleteLookups_TDTM extends CDL_CascadeDelete
                     (child instanceof Allocation__c && (
                         ((Allocation__c) child).Recurring_Donation__c == deletedRecurringDonationId ||
                         ((Allocation__c) child).Opportunity__r.npe03__Recurring_Donation__c == deletedRecurringDonationId
-                    ))
+                    ) ||
+                    (child instanceof  npe01__OppPayment__c && ((npe01__OppPayment__c) child).npe01__Paid__c))
```
vmartinov-sforg: Could we use isPaidPayment() here?

src/classes/RD_CascadeDeleteLookups_TDTM.cls
```diff
@@ -198,6 +208,11 @@ public with sharing class RD_CascadeDeleteLookups_TDTM extends CDL_CascadeDelete
                         Opportunity closedWonOpportunity = (Opportunity) child;
                         error.addChildError(Label.ClosedWonOpportunities, closedWonOpportunity.Id, closedWonOpportunity.Name);
                     }
+
+                    if (isPaidPayment(child)) {
+                        npe01__OppPayment__c paidPayment = (npe01__OppPayment__c)child;
+                        error.addChildError(Label.PaidPayment, paidPayment.Id, paidPayment.Name);
```
force2b: @KietTruongSFDO Would you mind changing this to `System.Label.PaidPayment`?

src/labels/CustomLabels.labels
```diff
@@ -1278,6 +1278,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>Processing</shortDescription>
         <value>Processing</value>
     </labels>
+    <labels>
+        <fullName>PaidPayment</fullName>
```
force2b: @KietTruongSFDO Since this is a label that is specific to Cascade Delete error message, I recommend a label name such as `CascadeDeletionError_PaidPayment` so it's clear that this label is only to be used in this one single context. I realize there may be other labels that are named like this, such as `ClosedWonOpportunities`, but it's too late to fix those. 

I'll have to check in with the translations team on these as well. Basically they have to translate the base `CascadeDeletionErrorLightning` label and then hope the translated term, such as "Paid Payment" makes sense within that other translation. That's not something we can fix here, but it's definitely something to keep in mind. It's also why that context is critical for the translation team.

KietTruongSFDO:
@force2b  I've updated the custom label to have the name "CascadeDeletionError_PaidPayment" and also updated to use "System.label" instead of just "label"

src/classes/RD_CascadeDeleteLookups_TDTM.cls
KietTruongSFDO: Updated

force2b:
Thank you for making those changes @KietTruongSFDO!

PR #4577, opened 2019-09-10
Attempt to fix null current Namespace bug

callen-sfdo:
An attempt to fix the intermittent bug that occurs when the namespaceWrapper is not defined in time for the getAlreadyMappedDIFields method in the bdiObjectMappingModal to use it.  A second PR will be created to remove the debug statements.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/bdiObjectMappingModal/bdiObjectMappingModal.js
+++ b/src/lwc/bdiObjectMappingModal/bdiObjectMappingModal.js
@@ -172,6 +172,7 @@ export default class bdiObjectMappingModal extends LightningElement {
             this.getDataImportFieldDescribes();
             this.getDataImportMappedFields();
         } catch(error) {
+            console.log('Error encountered in objectMappingModal init: ' + error);
             this.handleError(error);
         }
     }
@@ -258,6 +259,7 @@ export default class bdiObjectMappingModal extends LightningElement {
 
             this.isLoading = false;
         } catch(error) {
+            console.log('Error encountered in loadModalData: ' + error);
             this.handleError(error);
         }
     }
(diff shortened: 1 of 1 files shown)


src/lwc/bdiObjectMappingModal/bdiObjectMappingModal.js
```diff
@@ -526,9 +529,10 @@ export default class bdiObjectMappingModal extends LightningElement {
     refreshImportRecordFieldOptions() {
         this.diImportRecordFieldOptions = [];
         this.diImportRecordStatusFieldOptions = [];
-
+        console.log('in refreshImportRecordFieldOptions namespace Wrapper is: ' + this.namespaceWrapper);
         let fieldsToExclude = this.getAlreadyMappedDIFields(); 
-        if (fieldsToExclude && this.dataImportFieldData) {
+        
+        if (fieldsToExclude && this.dataImportFieldData && this.namespaceWrapper) {
```
ptewson-sfdo: Doesn't the call to this.getAlreadyMappedDIFields() on line 533 also reference "this.namespaceWrapper" and it's also outside of this if statement? There were a couple variations of the error message that showed up, and the one that reads "cannot read property 'currentNamespace' of undefined" I think might not be protected by the new conditional.

ptewson-sfdo:
**lurch: attach W-035852

src/lwc/bdiObjectMappingModal/bdiObjectMappingModal.js
```diff
@@ -172,6 +172,7 @@ export default class bdiObjectMappingModal extends LightningElement {
             this.getDataImportFieldDescribes();
             this.getDataImportMappedFields();
         } catch(error) {
+            console.log('Error encountered in objectMappingModal init: ' + error);
```
ptewson-sfdo: For outside observers: these logs are temporary and will be removed, we just need to be able to track down the nature of the issue if our first fix doesn't tackle it. These logs will minimize future iterative merges to master.

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAPrQAO" target="_blank">W-035852</a>

PR #4584, opened 2019-09-11
Fix Errors Uncovered During Unit Test Execution

jbrandolinisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


jbrandolinisf:
@force2b - This is ready when you get back!  I made a few minor additional updates to the unit test class for the OPP_OpportunityNaming_BATCH_TEST class to use the UTIL_UnitTestData_TEST calls instead of how it was doing it, just to close that out.

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -1,10 +1,8 @@
 /*
     Copyright (c) 2015, Salesforce.org
     All rights reserved.
-
```
force2b: These blank lines should probably be here to remain consistent with our copyright notice in all the other classes.

src/classes/OPP_OpportunityNaming_BATCH_TEST.cls
```diff
@@ -1,10 +1,8 @@
 /*
-    Copyright (c) 2019, Salesforce.org
+    Copyright (c) 2015, Salesforce.org
```
force2b: This class actually originated in 2019 so the copyright here and below should remain as 2019

src/classes/OPP_OpportunityNaming_BATCH_TEST.cls
```diff
@@ -67,152 +66,46 @@ private with sharing class OPP_OpportunityNaming_BATCH_TEST {
...
     }
 
     /**
-     * @description Confirms that the finish method does not chain the next batch
-     * if it fails to capture the last Opportunity Id processed
+     * @description Coverage of Schedulable Method
      */
-    @isTest
-    private static void shouldNotChainNextBatchIfLastOppIdProcessedIsNull() {
+    static testMethod void testSchedulable() {
```
force2b: This should use @IsTest to remain consistent with all other methods in this class

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -262,16 +262,16 @@ public abstract inherited sharing class UTIL_AbstractCommon_BATCH implements Dat
     @testVisible private String getQueryNonLDVWhereClause() {
         String fullWhereClause = '';
 
-        String whereClause = getWhereClause();
+        String whereClause = this.getWhereClause();
 
         Boolean mainWhere = String.isNotBlank(whereClause);
-        Boolean addtlWhere = String.isNotBlank(additionalWhereClause);
+        Boolean addtlWhere = String.isNotBlank(this.additionalWhereClause);
```
force2b: What would you think of renaming these vars as `hasMainWhereClause' and `hasAdditionalWhereClause`?

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -135,22 +135,18 @@ public class UTIL_AbstractCommon_BATCH_TEST {
...
+    @isTest
     public static void executeAddtlWhereClause() {
         List<Account> accs = UTIL_UnitTestData_TEST.createMultipleTestAccounts(TEST_ACCOUNT_CNT, null);
         insert accs;
 
         Test.startTest();
         UTIL_AbstractCommon_MOCK tab = new UTIL_AbstractCommon_MOCK(10, 'Id = \'' + accs[0].Id + '\'');
-        tab.setTestChunkSize(10);
-        tab.setTestMaxNonLDVSize(10);
+        tab.setTestChunkSize(UTIL_AbstractCommon_BATCH.DEFAULT_LDV_CHUNK_SIZE);
```
force2b: Rather than `tab` as the var name for the instance, how about `mockAbstractCommon`?

src/classes/UTIL_AbstractCommon_BATCH.cls
```diff
@@ -262,16 +262,16 @@ public abstract inherited sharing class UTIL_AbstractCommon_BATCH implements Dat
     @testVisible private String getQueryNonLDVWhereClause() {
```
force2b: I'm not sure why I didn't notice this before, but `UTIL_AbstractCommon_BATCH` doesn't really say anything about this class.  It's not a general common abstract class for every single batch job. Instead it's explicitly for unique situations where batch chunking is required. The class isn't global so the name can be changed. I think we should definitely considering renaming the class to something like `UTIL_AbstractChunking_BATCH`.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
```diff
@@ -135,22 +135,18 @@ public class UTIL_AbstractCommon_BATCH_TEST {
...
+        tab.setTestMaxNonLDVSize(UTIL_AbstractCommon_BATCH.DEFAULT_MAX_NON_LDV_SIZE);
         tab.setExpectedResult(1);
 
-        Database.executeBatch(tab, 1);
+        Database.executeBatch(tab, 10);
```
force2b: What does the change from 1 to 10 for the batch size do here if the condition is supposed to limit this to one record queried anyway. If the querying is working correctly, then a batch size of 1 should be fine because there can never be more than one record.

src/classes/OPP_OpportunityNaming_BATCH.cls
jbrandolinisf: Whoops will fix it.

src/classes/OPP_OpportunityNaming_BATCH_TEST.cls
jbrandolinisf: OK, I think I just reverted it to the prior one, but will change it.

src/classes/UTIL_AbstractCommon_BATCH.cls
jbrandolinisf: It could be, and does provide other functionality, including the ability to provide an additional where clause, validation, etc.  My thought was we might extend/add additional functionality as we find other things we wanted to do across batches.  Even if you aren't running in an LDV mode, there are things it will do, like let you specify a subquery, and have it automatically retrieve that within each batch (instead of the batch needing to explicitly do that).

Not opposed...again I was just thinking that a nice idea for our batches is to extend from this, and allow them to support running both ways (without additional work).  

We can talk more about it tomorrow.

src/classes/UTIL_AbstractCommon_BATCH_TEST.cls
jbrandolinisf: Mostly I did it to make sure the batch would run blow up in a weird way if the batch returned more than one record.  It would still error when checking the expected result.  I can put it back though-- it is a good callout.  I am also going to set the testChunkSize to 1, and see about adding an assertion to make sure it is not running in ldv mode unexpectedly in the ones it should not be.

src/classes/UTIL_AbstractCommon_BATCH.cls
jbrandolinisf: Also, this will mean refactoring in 2 additional PRs in Cumulus, plus replicating to Abacus.  If it is important/you feel really strongly I will do it (after we determine whether there is any value for non chunking circumstances), but you will need to twist my arm a bit on this one, as it is a lot of pain...and I really did intend this to be common for both ldv and non ldv concerns-- and that, it should run in non ldv mode unless they record size indicates it needs to run in ldv mode (so we could worry about that less in general) and build with this for new work so even if it eventually grows to that size we do not need to worry about it because it will already be handled.

src/classes/OPP_OpportunityNaming_BATCH.cls
```diff
@@ -35,56 +33,38 @@
...
-            new List<String>{ idToOffset, String.valueOf(queryLimit) }
-        );
+        super();
     }
 
-    /** @description Batch process start method.*/
-    public Database.QueryLocator start(Database.BatchableContext BC) {
-        return Database.getQueryLocator(query);
+    /** @description Constructor to allow specifying batch size.*/
+    public OPP_OpportunityNaming_BATCH(Integer batchSize) {
```
force2b: @jbrandolinisf As Discussed
Follow-up for later: Refactor the overloaded constructors into a single constructor and use set___() methods to pass additional values.

jbrandolinisf: @force2b - What else do I need to do to finish this up?  Rename the test methods?  Anything else from the live review I need to do?  Rename the class?  I just want to make sure I get it in one more pass.

force2b: Yes, we talked about renaming the test methods to reflect what was being tested. For example, "shouldProcessZeroRecords". The class rename was another, as well as [eventually] refactoring the constructors and methods simplify the implementation. Removing unused methods is optional, and though it would reduce the complexity of the class, it would require removing the related test methods as well.

src/classes/UTIL_AbstractChunkingLDV_BATCH.cls
```diff
@@ -262,16 +230,16 @@ public abstract inherited sharing class UTIL_AbstractCommon_BATCH implements Dat
     @testVisible private String getQueryNonLDVWhereClause() {
         String fullWhereClause = '';
```
force2b: It doesn't look like `fullWhereClause` is ever used.

jbrandolinisf:
**lurch: attach W-036512

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CBWLQA4" target="_blank">W-036512</a>

PR #4585, opened 2019-09-11
Advanced Mapping: Fix reciprocal relationship bug when BDI DI import fires relationship triggers

psadie-sfdo:
Fixes a bug where the system created reciprocal relationship record would have the contact and related contact fields point to the same contact when the first relationship record was created via the BDI import process.

# Critical Changes

# Changes
- Lowercase string when using UTIL_Describe to get field describe details
# Issues Closed

# New Metadata

# Deleted Metadata


psadie-sfdo:
**lurch: attach W-035298

PR #4592, opened 2019-09-16
Feature/bge bdi performance tests

prescod:
Performance tests, including:

  * Updated mapping.yml with more columns and irrelevant objects removed

 * A bunch of variants of a basic test

 * A bunch of support libraries and tasks

Including some tasks which are copies of tasks which will be updated in CCI itself on an asynchronous schedule.

The metadata should be merged to master in a separate PR soon so don't worry about reviewing that.


cumulusci.yml
```diff
@@ -128,22 +128,13 @@ tasks:
...
-        description: Runs Robot Framework performance tests
+    test_performance:
+        description: Runs Robot Framework performance tests for BDI Complex Objects
         class_path: cumulusci.tasks.robotframework.Robot
         group: Performance Tests
         options:
             suites: robot/Cumulus/perftests/
-            options:
-                outputdir: robot/Cumulus/results
+            outputdir: robot/Cumulus/results
```
davisagli: This was right before. outputdir is one of the options that gets passed into Robot Framework, not one of the options for the CCI task.

datasets/bdi_benchmark/mapping.yml
```diff
@@ -3,49 +3,58 @@ Accounts:
     table: accounts
     fields:
         Name: name
-    record_type: Organization
```
davisagli: Is Organization the default Account record type?

tasks/npsp_delete.py
```diff
@@ -0,0 +1,117 @@
+### Delete this file when CCI supports deletes with where (i.e. Nov 2019)
```
davisagli: This was included in cumulusci 2.5.8 (looks like we missed it in the release notes)

robot/Cumulus/perftests/BDI_Tests.robot
```diff
@@ -1,22 +1,138 @@
...
+    Output  Accounts imported    ${result}
+
+    ${result} =   Row Count     Contact  
+    ...           Name Like '%BDITEST%'
+
+    Output  Contacts imported    ${result}
 
-Import a data batch via the API
+Workaround Bug
+    Return From Keyword If      ${persistent_org}   # persistent orgs don't have this bug
```
davisagli: Can you add a bit more of a comment about what bug is being worked around?

robot/Cumulus/perftests/BDI_Tests.robot
```diff
@@ -1,22 +1,138 @@
...
+    ${result} =    Check Row Count    ${count}     DataImport__c       Status__c=Imported
+    Run Keyword Unless   "${result}"=="PASS"      Display BDI Failures
+    Should be Equal     ${result}      PASS
+
+Setup For Test
+    [Arguments]                 ${count}    ${bdi_mode}
+    Output      Preparing for ${TEST NAME}
+    Clear Generated Records
+    Setup BDI                   ${bdi_mode}
+    Generate Data               ${count}
```
davisagli: This is not a criticism, but I'm curious how you decided to write these keywords in Robot Framework syntax rather than in Python?

robot/Cumulus/perftests/BDI_Tests.robot
```diff
@@ -1,22 +1,138 @@
...
+    Output      Starting ${TEST NAME}
+
+*** Test Cases ***
+BGE/BDI Import - CO - 100 / 250 - 7.5 Objs/DI - 0.75 New Acc 0.25 Mtchd Acc 0.25 New Con 0.25 Mtchd Con 1CO1 0.5CO2 1CO3 1Payment 1Allocation 0.5ASC 1 Opp
+    [Setup]     Setup For Test    100    Data Import Field Mapping
+    [Teardown]     Validate Data      100
+    [Tags]    ultra-short     advanced-mapping
```
davisagli: I wonder if we should prefix the tags with `bdi-` or something, since RF tags are a single namespace.

tasks/salesforce_robot_library_base.py
```diff
@@ -0,0 +1,58 @@
...
+import pprint
+
+from cumulusci.core.config import TaskConfig
+from robot.libraries.BuiltIn import BuiltIn
+from cumulusci.tasks.apex.anon import AnonymousApexTask
+
+from tasks.npsp_batch import BatchApexWait
+
+
+class SalesforceRobotLibraryBase(object):
```
davisagli: This is not a task. Maybe it should go in robot/Cumulus/resources instead of tasks.

robot/Cumulus/resources/BDI_API.py
```diff
@@ -0,0 +1,22 @@
+from tasks.salesforce_robot_library_base import SalesforceRobotLibraryBase
+from tasks.set_BDI_mapping_mode import SetBDIMappingMode
+
+
+class BDI_API(SalesforceRobotLibraryBase):
+    def configure_BDI(self, mode):
+        return self._run_subtask(SetBDIMappingMode, mode=mode)
+
+    def _get_di_mode(self):
+        token = self.get_org_namespace_prefix()
```
davisagli: I think you want self.get_npsp_namespace_prefix(). In an un-namespaced org with managed NPSP installed, the NPSP objects still have a namespace prefix.

robot/Cumulus/resources/Data.py
```diff
@@ -0,0 +1,7 @@
+from tasks.npsp_delete import DeleteData
+from tasks.salesforce_robot_library_base import SalesforceRobotLibraryBase
+
+
+class Data(SalesforceRobotLibraryBase):
+    def delete(self, objects, *, where=None, hardDelete=False):
```
davisagli: I'd suggest calling this bulk_delete to help disambiguate from the salesforce_delete keyword in the Salesforce library, which deletes a single record.

tasks/bulkutils.py
```diff
@@ -0,0 +1,127 @@
+### Delete this file when CCI supports deletes with where (i.e. Nov 2019)
```
davisagli: This can be updated to use cumulusci.tasks.bulkdata.utils, yeah?

tasks/data_generation_base.py
```diff
@@ -0,0 +1,104 @@
+import os
```
davisagli: This module can be replaced by cumulusci.tasks.bulkdata.base_generate_data_task, yeah?

tasks/factory_utils.py
```diff
@@ -0,0 +1,108 @@
+from abc import ABC, abstractmethod
```
davisagli: This module can be replaced by the one we added to CumulusCI, yeah?

tasks/set_BDI_mapping_mode.py
```diff
@@ -0,0 +1,70 @@
+from cumulusci.tasks.apex.anon import AnonymousApexTask
+from cumulusci.core.exceptions import TaskOptionsError
+import time
+
+
+class SetBDIMappingMode(AnonymousApexTask):
```
davisagli: This is not actually configured as a task in cumulusci.yml; should the implementation move into BDI_API.py?

prescod: Its a bit more flexible as a task. Even if it doesn't have a short name, it can be used from other tasks. I don't feel very strongly one way or the other, but it does have that advantage as its own task. Perhaps some future demo or trial configuration would want the org to have Advanced Mapping on when its built. Doing it this way is slightly more future-proof and minimal cost.

davisagli: You convinced me. But why not go ahead and add it to cumulusci.yml so that it can be called on its own from the CLI?

tasks/factory_utils.py
prescod: In general, I have been conservative about assuming cci will be deployed before this stuff is. After I merge, its "deployed" to anyone who does a git pull. Not admittedly, I guess, not many people will be running my code, but MetaCI will be...

I felt like it is a good habit to allow CCI a month or so to get new versions deployed before my code depends on those new versions.
 
Or at the very least I need MetaCI to have the new CCI.

It's easy to delete files later, so I decided to err on the side of backwards compatibility instead of risking support requests.

davisagli: MetaCI already has cumulusci 2.5.8. For people running CCI locally, you can update the first line of Cumulus' cumulusci.yml to say:

    minimum_cumulusci_version: 2.5.8

and then it will exit on startup with a message telling the user to upgrade.

tasks/set_BDI_mapping_mode.py
prescod: Just trying not to make "cci task list" too busy until someone actually nominates a use-case for it...I don't feel strongly about that but I'll note that it isn't the only task that lacks a short-name. For example: DeleteData, AnonApex, ...

cumulusci.yml
prescod: Okay.

But if we're going to pick and choose which Robot options to support, we might as well just make them task options.

datasets/bdi_benchmark/mapping.yml
prescod: I actually don't know, but I'm not sure if it matters. What happens if they default to the wrong record type?

I'd expect NPSP to "prefer" Household anyhow.

robot/Cumulus/perftests/BDI_Tests.robot
prescod: https://github.com/SalesforceFoundation/Cumulus/pull/4592/files#diff-c514c9a06a353c336a1da458456bac85R64

tasks/data_generation_base.py
prescod: Deleted now.

tasks/factory_utils.py
prescod: I decided not to be that strict because only a small subset of users (perhaps just me and MetaCI) care about these classes.

cumulusci.yml
prescod: But you probably agree with that and we just haven't gotten around to it.

datasets/bdi_benchmark/mapping.yml
davisagli: NPSP (at least in the most common configuration) will automatically create a Household account for contacts created without an explicit account relation. Accounts can also be created with the Organization record type to represent organizations that are giving money. Since the contacts in this file are created without any lookup to these accounts, I assume these accounts are meant to be the latter. I doubt it makes a difference for performance but I'm pointing out a place where the data may not be realistic, because I'm not sure what the default record type will be.

robot/Cumulus/perftests/BDI_Tests.robot
davisagli: I missed that the test_performance task is also specifying a particular suite, so I thought it was only using the length-based tags to select the right tests. Okay, I'm fine with these tags.

tasks/npsp_delete.py
davisagli: This file is no longer used.

tasks/salesforce_robot_library_base.py
davisagli: Can't any folder be treated as a package in Python 3?

datasets/bdi_benchmark/mapping.yml
prescod: Would you suggest going back to record_type: Organization or Household?

tasks/salesforce_robot_library_base.py
prescod: It may be a problem specific to the way that Robot loads .py files.

ImportError: attempted relative import with no known parent package

I can clean this up by moving SalesforceBase to the CCI repro (which is my plan anyhow) but I'd rather do that asynchronously instead of waiting to merge this PR. Once its been deployed over there I can delete it here, similar to some of the other files.

datasets/bdi_benchmark/mapping.yml
davisagli: Organization. You'll have Households generated automatically for the contacts that are inserted, like I said.

robot/Cumulus/resources/BDI_API.py
prescod: https://github.com/SalesforceFoundation/Cumulus/pull/4592/files#diff-5a22133e151c83a83d77f9f305c4bfdbR10

robot/Cumulus/resources/Data.py
prescod: https://github.com/SalesforceFoundation/Cumulus/pull/4592/files#diff-cab735d807b27fbf170d4ed5b5685119R6

tasks/salesforce_robot_library_base.py
davisagli: I'm okay with that.

tasks/bulkutils.py
prescod: Deleted.

cumulusci.yml
davisagli: So in the meantime, unless I'm missing something, this still needs to adjusted if you don't want the results to end up in the default outputdir.

davisagli: I was missing something, which is that this snippet is outdated.

datasets/bdi_benchmark/mapping.yml
prescod: https://github.com/SalesforceFoundation/Cumulus/pull/4592/commits/fd9b6892b9f7da5954b00a074a72af3d9c677dca

I'm mostly only interested in realism from a performance perspective (I don't even use Faker for names) but its an easy fix.

PR #4628, opened 2019-09-23
Advanced Mapping: add field mappings cascade delete

psadie-sfdo:
- "Deleting" an object mapping will now "delete" all of its field mapping children.

# Critical Changes

# Changes

- We added cascading delete to Advanced Mapping for Field Mappings. Now, when a user deletes an Object Mapping, all of the Field Mappings within the Object Mapping are deleted.

# Issues Closed

# New Metadata

# Deleted Metadata


psadie-sfdo:
**lurch: attach W-035505

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9obQAC" target="_blank">W-035505</a>

psadie-sfdo:
**lurch: attach W-035505

src/classes/BDI_ManageAdvancedMappingCtrl.cls
```diff
@@ -380,6 +432,7 @@ public class BDI_ManageAdvancedMappingCtrl {
             for(String fieldName : sObjectCastedCustomMetadata.getPopulatedFieldsAsMap().keySet()) {
 
                 Boolean invalidField =
+                    fieldName.contains('__r') ||
```
ptewson-sfdo: What is exactly is causing us to block the relationship fields - was that not an issue before we started handling cascade delete, or is this a ride-along fix for a separate issue?

psadie-sfdo: Ah, neither actually. This is an artifact from when I was trying out another way of cascade deleting. Pushed a new commit reverting this. Thank you!

PR #4629, opened 2019-09-23
User without Delete on Contact/Account is able to use Contact Merge

meena-sfdc:
# Critical Changes

# Changes
If the user has no delete permission on the contact object and when they try to merge the contacts, they should receive an error message and should prevent in doing the contact merge.
If the user has delete permission on the contact object, however no sharing permission on the contact record, they should receive an error message and should prevent in doing the contact merge.
# Issues Closed

# New Metadata

Custom Labels:
conMergeErrorNoDeleteObjPermission
conMergeErrorNoDeleteRecAccess

# Deleted Metadata


meena-sfdc:
**lurch: attach W-035925

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAY0QAO" target="_blank">W-035925</a>

src/labels/CustomLabels.labels
```diff
@@ -3622,6 +3622,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <shortDescription>conMergeErrorNoDeleteObjPermission</shortDescription>
+        <value>You must have Delete permissions on Contacts in order to merge Contacts.</value>
+    </labels>
+     <labels>
+        <fullName>conMergeErrorNoDeleteRecAccess</fullName>
+        <categories>Contact Merge</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>conMergeErrorNoDeleteRecAccess</shortDescription>
+        <value>You don't have the record sharing permissions to merge these Contacts. Your system administrator may need to adjust organization-wide sharing settings, sharing rules, or user roles.</value>
```
force2b: Instead of using the apostrophe, you can escape this as `&apos;`?
ex: "You don&apos;t have the record sharing permissions ..."

src/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -35,39 +35,45 @@
...
     private final Integer SOSL_LIMIT = 100;
-    
+
     /*******************************************************************************************************
     * @description The contact merge step user is in. (The process has multiple steps.)
     */
     public Integer step {get; private set;}
-    
+
+    public String errorMsg {get; set;}
```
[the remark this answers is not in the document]
force2b: I do see where the page is checking for an empty error message to render a section. How about a boolean var for `canContinueWithMerge` that defaults to true, but is changed to false for any rendered permissions exception?

src/classes/CON_ContactMerge_TEST.cls
```diff
@@ -42,306 +42,427 @@ public class CON_ContactMerge_TEST {
...
+        Test.startTest();
+        // Verify that when logged in as readonlyuser, it returns error when merging contacts because of record level permissions
+        System.runAs(readOnlyUser) {
+            Test.setCurrentPageReference(new PageReference('Page.CON_ContactMerge'));
+            Con_ContactMerge_CTRL ctlr = new Con_ContactMerge_CTRL(new ApexPages.StandardSetController(new list<Contact>()));
+            System.assertEquals(null, ctlr.errorMsg);
+
+            ctlr.searchText = 'test';
+            //SOSL always returns nothing in tests, unless you use the setFixedSearchResults value
+            Id[] fixedSearchResults=new Id[2];
```
force2b: An alternative way to write these three lines is as:
`List<Id> fixedSearchResults = new List<Id>{ con.Id, con2.Id };`

src/classes/CON_ContactMerge_TEST.cls
```diff
@@ -42,306 +42,427 @@ public class CON_ContactMerge_TEST {
...
+            Name = 'PermissionTest',
+            Label = 'PermissionTest'
+
+        );
+
+        insert permission;
+        List<ObjectPermissions> objPermissions;
+
+
+        insert
```
force2b: This might be more readable as 
```
insert new ObjectPermissions(
    SObjectType = 'Contact',
    ...
);
```

src/labels/CustomLabels.labels
```diff
@@ -3622,6 +3622,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <protected>true</protected>
         <shortDescription>conMergeBtnLabel</shortDescription>
         <value>Merge</value>
+    </labels>
+     <labels>
+        <fullName>conMergeErrorNoDeleteObjPermission</fullName>
+        <categories>Contact Merge</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>conMergeErrorNoDeleteObjPermission</shortDescription>
```
force2b: I know that some of legacy labels have a short description that is the same as the label full name, but you're not locked into following that legacy format. The Description can be a description of what the error is, such as `Delete permission on the Contact object is required to for merging`. Though it's not directly customer facing (outside of Salesforce Setup), if the Description will contain text that isn't a duplicate of the label api name, Sam should do a review of it as well.

src/classes/CON_ContactMerge_TEST.cls
```diff
@@ -42,306 +42,427 @@ public class CON_ContactMerge_TEST {
...
+
+        // Create a permission set with CustomizeApplication set ensuring other flags
+        // on which CustomizeApplication settings depends are set as well
+        PermissionSet permission = new PermissionSet(
+            Name = 'PermissionTest',
+            Label = 'PermissionTest'
+
+        );
+
+        insert permission;
```
force2b: Is `objPermissions` used anywhere?

src/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -596,30 +614,51 @@ public with sharing class CON_ContactMerge_CTRL {
...
+    private boolean isUserHasDeleteAccess(set<Id> ContactIds) {
+        //This is the logic which checks if delete permission for the selected records exists or not
+        List <UserRecordAccess> userAccess = [
+             SELECT RecordId, HasEditAccess, HasDeleteAccess, MaxAccessLevel
+             FROM UserRecordAccess
+             WHERE UserId = :UserInfo.getUserId()
+             And RecordId IN :ContactIds
+        ];
+        for (UserRecordAccess user: userAccess) {
+            if (!user.HasDeleteAccess && user.MaxAccessLevel != 'None') {
```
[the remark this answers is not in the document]
meena-sfdc: There is a scenario in the code where it checks if any records gets deleted(deleted by background process or something) after selecting the contact for the contact merge and it provides different error message for that scenario. However, that is getting caught by this method and provides this error msg instead of the other message. The two test methods failed because of that. I noticed that if a record is deleted in the process, maxaccesslevel changes to 'none', all access like editaccess,readaccess etc becomes false.  I added this criteria to resolve the failures. However, there is  an alternative to do another query for the contact whether isdeleted is false or not. But it becomes more complicated in the functionality. I can talk to you and explain about this in detail if needed.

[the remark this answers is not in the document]
meena-sfdc: The methods in the test class are failNextStepWithDeletedRecordSelectedOne2One and failNextStepWithDeletedRecordSelected.

src/classes/CON_ContactMerge_TEST.cls
meena-sfdc: @force2b , I don't see objPermissions used anywhere in test or main class.

force2b: Was this something you had added initially and maybe didn't end up using? If it's not used anywhere though, I recommend removing it.

meena-sfdc: Yes @force2b , I will remove that List named objPermissions, may be added initially and not needed later.

meena-sfdc:
@force2b , The changes are all done and ready for review. I have included my comments for one change.

src/classes/CON_ContactMerge_CTRL.cls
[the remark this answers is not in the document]
force2b: Ah. I see the issue. If the Contact is deleted, we want the error message to reflect that the Contact has been deleted instead of that they don't have delete permissions.

I can't find documentation that explicitly states the behavior of the `MaxAccessLevel` field for deleted records. That makes me a bit nervous about relying on the value of that field for undocumented behavior. Were you able to find something that called this out?

I suppose an alternative approach is to move the call to canUserDeleteSelectedContacts() to after the query, towards the bottom of the method. This way the method can do all the logic that queries the records first. It might make the method more readable if the entire block of code that builds the `fieldRows` list (effectively from line 284 to 431) is abstracted out into a method (something like `buildFieldListForSelectedContacts()`.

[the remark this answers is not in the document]
meena-sfdc: @force2b , If MaxAccessLevel is none, then, the record don't have even read permission on it. Without that, the user can't even select the contact for merging it. That's why I selected that criteria to bypass the deleted contacts in the logic. I will check if any documentation is there too.
For your alternative approach, I am not sure if I can move the call after as you mentioned. Because, lines 284 to 347, it checks if any contacts got deleted in the process and shows that query error message for that scenario. From lines 348 to 431, it does prework for the contact merge process for step3. We don't want to do that if the contact don't have record level access.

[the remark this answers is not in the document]
meena-sfdc: @force2b , I have moved the method to line 342 for this check after the query check. I have removed the Maxaccesslevel check in the criteria.

meena-sfdc:
@force2b , The changes are done and ready for review.

rponti-sforg:

For further WI and testing details, see:
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001CAY0QAO/view
https://salesforce.quip.com/ugezAUEpOTef

Merging PR once Feature Test - Namespaced is finished running.

sam-knox:
@meena-sfdc - Be sure to link PRs to related Issues for bug fixes like this. I added the link to Issues Closed here already.

PR #4632, opened 2019-09-23
Make Error Handling functionality available to other packages via Callable API

jbrandolinisf:
Additional Functionality added to CallableAPI to expose certain Error Functionality for use in Abacus
# Critical Changes

# Changes

# Issues Closed

# New Metadata
- CallableDispatchService_ERR
- CallableDispatchService_ERR_TEST

# Deleted Metadata


jbrandolinisf:
@force2b - Tagging for review on the changes to expose certain aspects of the Error Functionality for use in Abacus.

jbrandolinisf:
**lurch: attach W-035629

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CA4yQAG" target="_blank">W-035629</a>

src/classes/CallableDispatchService_ERR.cls
```diff
@@ -0,0 +1,207 @@
...
+     * @description Adds an Error Message via the ERR_Handler class.
+     *              @see ERR_Handler.processError
+     * @param context The context the exception should be logged under
+     * @param errorMessage The Error to Log
+     * @return Boolean true
+     */
+    private static Boolean addError(String context, String errorMessage) {
+        ERR_Handler.processError(errorMessage, context);
+
+        return true;
```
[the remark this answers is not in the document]
jbrandolinisf: I was leaving myself open to a future scenario where I did need to return an Object back via the Dispatcher...so I thought they all should return something...I could restructure it though.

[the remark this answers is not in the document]
jbrandolinisf: I refactored these to be void, and not return a value.  The dispatch action does, however, return true at the end (similar to the Callable API).  I feel strongly it should, so that we don't have to update the Callable API to go from a method invocation to a return method invocation if needed.

jbrandolinisf:
@force2b - Ready for re-review

src/classes/CallableDispatchService_ERR.cls
```diff
@@ -0,0 +1,196 @@
...
+
+            } when else { // Invalid Action Type / Action Type Not implemented
+                throw new MalformedErrorDispatchMethodApiException(
+                    String.format(System.Label.CallableApiMethodNotImplemented, new List<String> { 'err.dispatchaction => ' + dispatchAction})
+                );
+            }
+        }
+
+        // if the action does not return a value then return success
+        return true;
```
force2b: Perfect. That's how the CallableApi does it as well. This way your individual methods can return when they need to, but this acts as a catch all if there is no explicit return.

src/classes/CallableDispatchService_ERR.cls
```diff
@@ -0,0 +1,196 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description Dispatches calls from the Callable API to the NPSP Error and RecordError APIs,
+*              depending on the ActionType that is set
+*/
+public with sharing class CallableDispatchService_ERR {
```
force2b: Sorry, I missed the `static` part before. Why not just make this an instance class? It can get instantiated from the CallableApi class. The only public method to instantiate would be `dispatch`. The constructor could take the parameters collection, though that would totally be up to you.

jbrandolinisf: @force2b - I don't mind making this an instance class, if you don't mind reviewing it one more time.  It should be a quick refactor.  If we are going to do that, I will probably use the withParameters(...).dispatch() strategy.

jbrandolinisf:
@force2b - One more quick review?

PR #4636, opened 2019-09-25
Feature/bdi support formulas as source fields

callen-sfdo:
Made a change in the controller to allow fields with isCalculated = true into the field describe results, then added an attribute to the fieldinfo class so they could be identified.  Also explicitly excluded formula fields from making it into the list for target field options, or any of the object group field selection.

# Critical Changes

# Changes

- In Advanced Mapping, we added the ability to use formula fields as the source field on the NPSP Data Import object. A formula source field can be mapped to a writable target field.

# Issues Closed

# New Metadata

# Deleted Metadata


callen-sfdo:
**lurch: attach w-035404

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9YQQA0" target="_blank">W-035404</a>

PR #4637, opened 2019-09-25
Fix for Payment Allocation percentage calculation

lparrott:
# Critical Changes

# Changes

- We corrected an issue that was causing Payment Allocation percentages to not match their parent Opportunity Allocations.

# Issues Closed

# New Metadata

# Deleted Metadata


lparrott:
**lurch: attach W-036125

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAuvQAG" target="_blank">W-036125</a>

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -643,7 +643,9 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
                     // At this point, only the amounts on each allocation is correct, not the percentage.
                     // Go through the allocation and adjust the percentage so it is correct.
                     for (Allocation__c consolidatedAllocation : consolidatedAllocationList) {
-                        consolidatedAllocation.Percent__c = consolidatedAllocation.Amount__c / amount * 100;
+                        if(consolidatedAllocation.Percent__c == null){
```
jbrandolinisf: Hmmm.  Seems we broke one thing to fix another.  If I remember right, @KietTruongSFDO added this to address a situation where there were more than one of the same allocation from the parent.  Basically, in that scenario, we have to recalculate the % because we are combining two or more allocations from the parent.  It is not setup to differentiate when it does/does not need to do that, though.

Did you play around at all with setting the scale on the percentage?  Possibly to a scale of 0 (since we are multiplying by 100) to get to a whole number.

lparrott: That's what I was afraid of, thanks for the context. I'll add a test case for that situation and update the logic to cover it.

lparrott: @jbrandolinisf I put in some additional logic that will add up the allocation percentages when the same GAU is being used, and only recalculate the percentage if none was provided. I think that should cover our use cases?

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -117,6 +117,77 @@ private with sharing class ALLO_Allocations_TEST {
         system.assertEquals(8, queryOpps[0].Amount, 'The opportunity amount should remain unchanged after the exception is caught.');
     }
 
+    /*******************************************************************************************************
+    * @description When Payments are created after the Opportunity and Opportunity Allocations exist,
+    * ensure that the payment allocation percentages match those from the Opportunity Allocations, and
+    * are not overwritten by additional logic in the trigger.
+    * Also check that allocation rounding works correctly and no amounts are allocated to the Default GAU
+    ********************************************************************************************************/
+    static testMethod void pmtAllocationPercentCheck() {
```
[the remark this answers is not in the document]
lparrott: I had my naming follow the other tests in this Class (oppAmountChange, pmtAmountChange, etc.), do you still think it needs to be updated?

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -620,6 +620,10 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
 
                     if (isPaymentAllocation == true && alloInMap != null) {
                         alloInMap.Amount__c += allo.Amount__c;
+                        if(alloInMap.Percent__c != null && allo.Percent__c != null) {
```
force2b:
```suggestion
                        if (alloInMap.Percent__c != null && allo.Percent__c != null) {
```

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -643,7 +647,10 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
                     // At this point, only the amounts on each allocation is correct, not the percentage.
                     // Go through the allocation and adjust the percentage so it is correct.
                     for (Allocation__c consolidatedAllocation : consolidatedAllocationList) {
-                        consolidatedAllocation.Percent__c = consolidatedAllocation.Amount__c / amount * 100;
+                        // Only recalculate if percent values were not set on the Allocations
+                        if(consolidatedAllocation.Percent__c == null){
```
force2b:
```suggestion
                        if (consolidatedAllocation.Percent__c == null) {
```

src/classes/ALLO_Allocations_TEST.cls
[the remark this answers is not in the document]
force2b: Yes please. We also try to only use this format for all new tests.
```
@isTest
private static void shouldTestSomething() {
 ....
}
```

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -117,6 +117,77 @@ private with sharing class ALLO_Allocations_TEST {
...
+        insert defaultGau;
+
+        setupSettings(new Allocations_Settings__c(
+            Payment_Allocations_Enabled__c = true,
+            Default_Allocations_Enabled__c = true,
+            Default__c = defaultGau.Id));
+
+        Account acc = new Account(Name='Account-pmtAlloPercentCheck');
+        insert acc;
+        Opportunity opp = new Opportunity(Name='Opp-pmtAlloPercentCheck', Amount = 100, AccountID=acc.id, CloseDate=system.today(), StageName=UTIL_UnitTestData_TEST.getClosedWonStage());
```
force2b:
```suggestion
        Opportunity opp = new Opportunity(
            Name = 'Opp-pmtAlloPercentCheck',
            Amount = 100, 
            AccountId = acc.Id, 
            CloseDate = System.Today(), 
            StageName = UTIL_UnitTestData_TEST.getClosedWonStage()
        );
```

force2b: There are too many for me to edit in this method, but field and classes names should always begin with a capital letter: `Id` instead of `id`, `System` instead of `system`, etc.

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -117,6 +117,77 @@ private with sharing class ALLO_Allocations_TEST {
...
+        gauForInsert.add(gau1);
+        General_Accounting_Unit__c gau2 = new General_Accounting_Unit__c(Name='GAU-2');
+        gauForInsert.add(gau2);
+        General_Accounting_Unit__c gau3 = new General_Accounting_Unit__c(Name='GAU-3');
+        gauForInsert.add(gau3);
+        insert gauForInsert;
+
+        List<Allocation__c> alloForInsert = new List<Allocation__c>();
+
+        Allocation__c percentAllo1 = new Allocation__c(Opportunity__c=opp.id, Percent__c = 10, General_Accounting_Unit__c = gau1.id);
```
force2b: Since you call this 4 times, I recommend abstracting out into a helper method. I suspect there's other code like this in this class too.

```
private static Allocation__c createAllocation(Id gauId, Id oppId, Double amt, Double percent) {
    return new Allocation__c(
            Opportunity__c = oppId, 
            General_Accounting_Unit__c = gauId,
            Percent__c = percent,
            Amount__c = amt
        );
}
```

The the lines in this method can just be:
```
         List<Allocation__c> alloForInsert = new List<Allocation__c>{
             createAllocation(gau1.Id, opp.Id, null, 10),
             createAllocation(gau1.Id, opp.Id, null, 24),
             createAllocation(gau2.Id, opp.Id, null, 33),
             createAllocation(gau1.Id, opp.Id, 33, null)
         };

File: src/classes/ALLO_Allocations_TDTM.cls [hunk 7]
```diff
@@ -643,7 +647,10 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
+                            consolidatedAllocation.Percent__c = consolidatedAllocation.Amount__c / amount * 100;
```
force2b: Since this is about fixing rounding issues, is there any value to implementing the `Divide()` method so that you can explicitly specify scale and the rounding method? The `PMT_PaymentWizard_CTRL` class uses the `Divide` method as well. For example:
```
Decimal consolidatedAmt = consolidatedAllocation.Amount__c;
Decimal consolidatedPercent = consolidatedAmt.divide(amount, 2, System.RoundingMode.HALF_UP);
consolidatedAllocation.Percent__c = consolidatedPercent;
```

https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_methods_system_decimal.htm

src/classes/ALLO_Allocations_TEST.cls
force2b: I also have a larger question about what this test is actually testing. The method that was changed in the `ALLO_Allocations_TDTM` class is `copyRecurringDonationCampaignAllocations()`. Unless I'm mis-reading the TDTM class, I'm fairly sure that method is only called copying default Allocations from a related RD or Campaign to an Opportunity and/or for adjusting the default allocation amounts. This test method doesn't use a Campaign or RD for default Allocations though. Am I misunderstanding how the logic in the trigger is actually working? And if, so does the description of that method name (and maybe the method name) need to be updated to reflect it's true behavior?

src/classes/ALLO_Allocations_TDTM.cls
lparrott: I like this idea, since it gives more control to the rounding, but I'm still not convinced we **want** to round these percents, given the possibility that a user could have entered a long decimal, for whatever reason.

src/classes/ALLO_Allocations_TEST.cls
lparrott: Totally agree here, I was just trying to make as few changes as possible! I'll update the rest of the class to use this format, it's much cleaner.

src/classes/ALLO_Allocations_TEST.cls
lparrott: It appears that the `copyRecurringDonationCampaignAllocations()` method has been repurposed to also copy Payment Allocations from Opportunities, without being renamed. The description includes this detail, but I can update the method name.

src/classes/ALLO_Allocations_TEST.cls
lparrott: That definitely looks nicer, I'll do a find and replace in this class to correct those cases.

lparrott:
Thanks for taking a look at this, @force2b, I just pushed your requested style changes and the updated method name, but I left the percentage calculation alone for now, since I think that warrants further discussion, and probably a separate work item.

src/classes/ALLO_Allocations_TDTM.cls
lparrott: Looking at this again, this code only gets called if the user didn't enter a Percent value for all of their Allocations, **but** rounding the percent value could technically make it less accurate than the amount value...

src/classes/ALLO_Allocations_TEST.cls
force2b: Thanks @lparrott. It's ok to refactor when it helps with readability, such as to fix gross-style errors like missing curly brackets, missing spaces, or lowercase class/field names. A bonus here is a reduction in the number of lines of code in the test class.

src/classes/ALLO_Allocations_TEST.cls
force2b: We probably don't want to take this on for this work item, but I'd say that it's worth having a conversation at some point about completely rebuilding that method so that it delegates individual responsibilities to much smaller methods that do _one_ thing. That would also significantly help with debugging, future updates, and writing unit tests. fyi @jbrandolinisf

src/classes/ALLO_Allocations_TDTM.cls
force2b: Understood. Floating point fields are always a bit challenging to understand for me. Even though we define a numeric field with 2 decimal points (scale), Salesforce actually stores the data with (what I think is) nearly unlimited number of values after the decimal point. From what I've seen, the scale is only used for rounding and displaying the value in the UI. So in this case I'm pretty sure we're actually storing the full result of the calculation as well as using it for the resulting calculations. And, that is probably better for this use case than forcing the rounding and having to deal with what could be a fractional remainder.

lparrott:
@jbrandolinisf @force2b I'm bringing this PR back from purgatory with a quick change: AFTER calculating the allocation amount, I think rounding the Payment Allocation percentage is a good idea, since it provides a better visual in the UI (50% vs 49.9999% for instance).

Let me know what you think, I also removed a section that was never being called, since percentages were being calculated earlier in the code.

force2b:
Thanks @lparrott.

PR #4645, opened 2019-09-27
Update BDI_DataImportFLSService to always consider Master-Detail fields update-able.

### LINKED ISSUE #4659: Mapping to a Master-Detail field in Advanced Mapping can cause Data Import to fail with permissions error
Mapping to a target master-detail field in Advanced Mapping can cause the Data Import process to fail with a "You do not have permission to modify" error when the field is not re-parent-able.

Steps to reproduce:
*Step 1: Create Data Import Fields:*

1. Engagement Plan 1 Imported (lookup)
2. Engagement Plan 1 Import Status (text)
3. Engagement Plan 1 Engagement Plan Template (lookup)

*Step 2: Create Object Group:*

* Group Name = Engagement Plan 1
* Object Name = Engagement Plan
* Is Child/Parent = Child
* Of This Mapping Group = Contact 1
* Through This Field = Contact
* Link to Field = Engagement Plan 1 Imported
* Import Status = Engagement Plan 1 Import Status

*Step 3: Set up Field Mappings*

1. Click the down arrow on the Engagement Plan 1 Object Group entry and choose View Field Mappings
2. Click the Create New Field Mapping button
1. Source Field Label : Engagement Plan 1 Engagement Plan Template
... maps to...
2. Target Field Label: Engagement Plan Template

Step 4: Process the Data Import record

The Data Import record fails with: “You do not have permissions to modify Engagement_Plan__c.Engagement_Plan_Template__c.”

kenlewis-sfdo:
Fix master-detail field mapped as target field in Advanced Mapping causing Data Import to fail with permissions error.

This set of changes enables the BDI_DataImportFLSService to always
consider Master-Detail fields as valid when checking their isUpdateable
property - in order to avoid the "You do not have permissions to modify field <xyz>..."
error when Master-Detail fields are mapped as target fields in Batch
Data Import Advanced Mapping.

# Critical Changes

# Changes

We fixed an issue that prevented users from using Advanced Mapping for Engagement Plans and different types of Soft Credits. We’ve added these recipes to the Advanced Mapping Recipe article: https://powerofus.force.com/s/article/NPSP-Advanced-Mapping-Recipes

# Issues Closed
Fixes #4659 

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-036222

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CB5yQAG" target="_blank">W-036222</a>

src/classes/UTIL_Permissions.cls
```diff
@@ -289,6 +289,16 @@ public with sharing class UTIL_Permissions {
         return true;
     }
 
+    /**
+     * @description Instance method used to determine whether the running user has
+     * update permissions on the given field.  Facilitates dependency injection in tests.
+     * @param field the SObjectField token of the field to check
+     * @return true if the field is updateable by the running User, otherwise false
+     */
+    public Boolean isUpdateable(SObjectField field) {
```
ptewson-sfdo: I think this method is (more or less) already implemented in this class - would it be possible to use canUpdate(field.getDescribe(), false);?

src/classes/BDI_DataImportFLSService.cls
```diff
@@ -137,7 +141,8 @@ public class BDI_DataImportFLSService {
                 && !field.getDescribe().isCreateable())
                 ||
                 (accessLevels.contains(AccessLevel.VALIDATE_UPDATE)
-                        && !field.getDescribe().isUpdateable())
+                        && !permissionsService.isUpdateable(field)
```
ptewson-sfdo: Would it work to use the previously implemented `permissionsService.canUpdate(field.getDescribe(), false)` here?

src/classes/UTIL_DescribeProvider.cls
```diff
@@ -0,0 +1,51 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Utilities
+* @description This class facilitates the use of dependency injection to mock calls to
+* instance methods in the UTIL_Describe class within tests.
+*/
+public with sharing class UTIL_DescribeProvider {
```
ptewson-sfdo: What's the benefit of having a separate provider class, as opposed to having utilDescribeInstance being a static member variable of UTIL_Describe itself and supporting these two methods under that class? I'm wondering if only because we haven't included separate providers in our repo as yet, and I don't want to introduce the pattern into our code base lightly if it's not one we at least want to move towards. That, and we have more than 1000 class files in our repo already!

Is there any downside to just moving the guts of this class into UTIL_Describe (and doing similar transfer for UTIL_Permissions) and basically allowing those classes to serve as providers for themselves?

kenlewis-sfdo: Yeah, that's a good catch - in this case it probably doesn't provide any specific value.  If the UTIL_Describe or UTIL_Permissions classes were implementations of interfaces, then having a separate provider class would enable complete switching of implementations down the road - but since they are not it makes sense to have the provider function inside the class itself.  I'll switch it up on a follow-up commit!

src/classes/BDI_DataImportFLSService.cls
kenlewis-sfdo: All of the "canUpdate" methods are static.  So we need an instance method to facilitate the test.  Re: naming and params of the instance method, I feel that the name "canUpdate" seems a bit like an unnecessary abstraction, and I prefer to pass the lighter-weight SObjectField when calling methods related to a field, rather than passing the describe object.  That being said, and instance method that follows the pattern you show here would be more in-line with the existing code.  Do you have a strong preference?  Happy to align it if you see more value that way.

ptewson-sfdo: There are actually a few instance (non-static) versions of "canUpdate" already - give it a look (I think Jenny added these a while back for similar reasons as you are interested, aka supporting good unit test practices). I don't think the exact signature you are looking for is an instance method yet, so I think it's worth adding but I'd like to keep the naming consistent and also probably to use the FieldDescribe for the argument instead of the SObjectField just to be consistent with what's already in there, yeah

ptewson-sfdo: (Check out line 350 of UTIL_Permissions! `public Boolean canUpdate(SObjectType sObjType, Set<SObjectField> sObjFields) {`)

ptewson-sfdo: Don't use that signature, I do think you're right we need a variant that supports the FieldDescribe argument, maybe along with the boolean for throwing the exception, so we still are aligned with adding the new method

kenlewis-sfdo: Ok i see now, the instance methods are below the other static methods at the bottom of the class!  Ok I'll update this also - thanks!

kenlewis-sfdo:
Ok, I created static instances in both the Permissions and Describe classes, and made a few updates to use the permissions instance for permissions calls in the isValid method.  I made a few methods on the Permissions class public and added one canCreate method to have consistency re: how those methods are called from the FLS validator class.

src/classes/BDI_DataImportFLSService.cls
```diff
@@ -133,14 +137,18 @@ public class BDI_DataImportFLSService {
     }
 
     private Boolean isValid(SObjectField field) {
+        Set<SObjectField> fieldSet = new Set<SObjectField>{
+                field
+        };
         if ((accessLevels.contains(AccessLevel.VALIDATE_CREATE)
-                && !field.getDescribe().isCreateable())
+                && !permissionsService.canCreate(fieldSet))
```
ptewson-sfdo: I honestly could go either way on this, but for the purpose of discourse - is it worth adding methods to UTIL_Permissions to test single SObjectField? Like `canCreate(SObjectField field)`?

ptewson-sfdo:
Approved, except then I noticed that something is going on with the tests? @kenlewis-sfdo

src/classes/BDI_DataImportFLSService.cls
kenlewis-sfdo: Yeah, it would look cleaner from the calling location in this instance, but you can obviously accomplish what is needed by passing in a collection, even of size 1, like here.  I could go either way also.

kenlewis-sfdo:
@ptewson-sfdo Looks like one change (to go back to calling the static getFieldDescribe method) didnt get into the last commit.  Fixed and pushed - tests re-running now.

PR #4651, opened 2019-09-30
Add all ISO to ICU currency symbols to existing map

### LINKED ISSUE #2931: Expand support of currency symbols in Opportunity Naming
[Thread.](https://powerofus.force.com/0D58000003xKB5n)

Our code currently only supports a narrow set of currency symbols in Opportunity Naming. We should allow for customization to expand the supported symbols for organizations that need it.

Julian88Tex:
# Critical Changes

# Changes
- We made all ICU currency symbols supported by [Salesforce](https://help.salesforce.com/articleView?id=admin_supported_currencies.htm&type=5) available in NPSP. The currency symbols display in automatically named Opportunities for non-multicurrency Salesforce organizations. Previously, only 8 ISO codes were mapped to ICU currency symbols.

# Issues Closed

- Fixes #2931 

# New Metadata

# Deleted Metadata


Julian88Tex:
@sam-knox sorry this is my first dev ticket so not sure about doc process so I thought I'd just tag you so you can review my "Changes" section above.

Julian88Tex:
**lurch: attach W-022486

Julian88Tex:
@force2b This PR is ready for a review. Also, would you mind telling me if I did something wrong with my lurch? It does not appear to be connecting. Thanks!

Julian88Tex:
> Very nice work @Julian88Tex, though there are some crypto-currencies in this list that should be removed: ETH, BTC, XBT, and LTC.
> 
> Overall this is an excellent start towards the symbols. However, I think that long term we need a more robust solution that can define not only the symbol but whether the symbol goes before or after the amount. That might be well outside of the scope of this work item though. @mpusto What do you think?

@force2b Thanks and good catch on crypto... didn't even notice there were in there. That being said there were indeed other ICU currency symbols besides crypto that Salesforce core does not support. I decided to keep them in there in case they were ever to expand so that we would future proof it to a certain degree, but I think there's also a case to be made to not confuse anyone looking at the code into thinking a certain symbol is supported by NPSP when in reality we are limited by core. I've gone ahead and removed all non-core supported symbols including crypto currencies. I would recommend in the long term we consider pulling from a more [official library](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/util/Currency.html) if that is possible.

I also agree that we should consider support for symbols being on the correct side. I was going to add that as a recommendation to our WI until I noticed we had one for it already: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x80000003xnr0AAA/view

I'd recommend this support be considered alongside any consideration to create a naming service. Probably in a potential "amount" method. This wiki will likely come in handy when that time comes: http://userguide.icu-project.org/formatparse/numbers#TOC-Formatting-for-a-Locale

src/classes/UTIL_Currency.cls
```diff
@@ -149,20 +149,154 @@ public without sharing class UTIL_Currency implements UTIL_Currency.Interface_x
...
+        'BZD' => 'BZ$',
+        'CAD' => '$',
+        'CDF' => 'FC',
+        'CHF' => 'CHF',
+        'CLP' => '$',
         'CNY' => '¥',
-        'JPY' => '¥',
+        'COP' => '$',
+        'CRC' => '₡',
+        'CSD' => 'CSD'
```
force2b: @Julian88Tex A comma is missing here. That's why the builds are failing for this branch.

sam-knox:
Thanks @Julian88Tex. For bugs, the process is to write something in Changes that explains what the PR is doing/changing and then ping me to review (which is exactly what you've done ;-))

For new functionality and features, it's the same process but do try to call out if there are new objects, fields, help text, UI elements, etc. that need review also.

Julian88Tex:
> Thanks @Julian88Tex. For bugs, the process is to write something in Changes that explains what the PR is doing/changing and then ping me to review (which is exactly what you've done ;-))
> 
> For new functionality and features, it's the same process but do try to call out if there are new objects, fields, help text, UI elements, etc. that need review also.

@sam-knox great thanks for info!

src/classes/UTIL_Currency.cls
Julian88Tex: @force2b oops I think I CMD+Z'd a one times too many before the commit. corrected that thanks!

Julian88Tex:
@meena-sfdc this is ready for QA whenever you're ready :-)

Julian88Tex:
> Great work @Julian88Tex

thanks! Mind running this command so we can connect to the WI: `**lurch: attach W-022486`

force2b:
**lurch: attach W-022486

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000009UcIgAAK" target="_blank">W-022486</a>

Julian88Tex:
@meena-sfdc has QAed this for me and confirmed it's working. Thanks!

PR #4653, opened 2019-10-01
Fix the error in opp naming when there is no mapped currency symbol

### LINKED ISSUE #4612: Error in Opportunity Naming when there is no mapped Currency Symbol for Opp Currency
Steps to reproduce:

1. Enable Multi-Currency in a Scratch Org
1. Create a new currency for the "Bolivian Boliviano" (BOB), leaving the default as USD
1. Configure an Opportunity Naming Pattern using NPSP Settings (`{!Account.Name} {!Amount} {!RecordType.Name} {!CloseDate}`)
1. Create a new Opportunity for a Household Contact with a currency of USD putting "Curr Test" in the Opp Name field. See that the Opp is "named" properly.
1. Create a new Opportunity with a currency of BOB putting "Curr Test" in the Opp Name field. See that the Opp Name field is left unchanged after saving.
1. Return to NPSP Settings and open SystemTools->Error Log. See that a new log record is created for a the OPP_OpportunityNaming class.

===========================================================
Failed to process batch for class 'npsp.OPP_OpportunityNaming_BATCH'

caused by: System.NullPointerException: Attempt to de-reference a null object

Class.npsp.OPP_OpportunityNaming.getFieldValue: line 443, column 1
Class.npsp.OPP_OpportunityNaming.replaceFieldNamesWithValues: line 395, column 1
Class.npsp.OPP_OpportunityNaming.refreshOppNames: line 90, column 1
Class.npsp.OPP_OpportunityNaming_BATCH.execute: line 96, column 1

meena-sfdc:
# Critical Changes

# Changes
Fix the error in Opportunity Naming when there is no mapped Currency Symbol for Opportunity's Currency.
# Issues Closed
Fixes #4612
# New Metadata

# Deleted Metadata


meena-sfdc:
**lurch: attach W-036136

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAvyQAG" target="_blank">W-036136</a>

meena-sfdc:
@force2b , This PR is done and ready for review.

src/classes/OPP_OpportunityNaming.cls
```diff
@@ -440,12 +440,13 @@ public with sharing class OPP_OpportunityNaming implements OPP_INaming {
                     currCode = (String) obj.get('CurrencyIsoCode');
                 }
                 String currencySymbol = UTIL_Currency.getInstance().getCurrencySymbol(currCode);
-                if (currencySymbol.length() == 3) {
+                if (!String.isBlank(currencySymbol) && currencySymbol.length() == 3) {
```
force2b: @meena-sfdc and @Julian88Tex Given that the updated list of symbols has a number of them that are 2, 3 and 3 even 4 characters, I think this should add a space whenever the symbol length is `> 1`. That would cover symbols like "KM", "$U", and even "Дин.". What do you think?

```suggestion
                if (!String.isBlank(currencySymbol) && currencySymbol.length() > 1) {
```

meena-sfdc: @force2b , In the code, it checks if the length is 3, then adds space in between the symbol and the amount. Are you sure if the space is needed when it is two characters or more?

force2b: Here are some examples:

- "$1000"
- "¥1000"
- "KM1000" or "KM 1000"
- "$U1000" or "$U 1000"
- "Дин.1000" or "Дин. 1000"
- "XYZ1000" or "XYZ 1000"

For the ones that are longer than 1 character, I feel they're easier to read with the extra space. Longer term we want to add logic to put the symbol either before or after the amount, depending on the locale, but that's for some point in the future. What do you think @Julian88Tex?

force2b:
@meena-sfdc Can you add an entry for the github issue this addresses to the "Fixes" section of the PR comment? You can find a link to the GH issue in the Work Item chatter feed. Thanks.

meena-sfdc:
Fixes #4612

force2b:
@meena-sfdc Sorry. The "Fixes #4612" part has to go in the PR description in the section for "Issues Closed".

meena-sfdc:
Yeah, @force2b , got it, I added in there too, sorry!

src/classes/OPP_OpportunityNaming.cls
meena-sfdc: I agree that it is more readable, @force2b . As all the currency symbols are added, not sure if it is expected or not.

Julian88Tex: @force2b yeah, I think that logically makes sense. I want to do a small amount of research first to just verify what the standard would be and will provide some feedback on that. I'll go ahead and take the PR back, for now, @meena-sfdc and reply back here after doing that research and/or making a change.

Julian88Tex: @force2b @meena-sfdc oops take back a bit of what I said... for some reason I got confused and thought this was my PR. I agree though that it's more readable and would be happy to check some ICU standards to just verify if that's what they would recommend.

Julian88Tex: @force2b @meena-sfdc hmm doesn't look like this section gives any guidance on spacing: http://userguide.icu-project.org/formatparse/numbers#TOC-Currency-Formatting

I think we'd have to plug in a few and/or read through their code to get an idea of how to format which is maybe not worth it. I'd say it's fair to say that just adding a space with anything larger than 1 character is reasonable for now until we do some more work correctly formatting currency.

meena-sfdc: @force2b & @Julian88Tex , Got it, thanks!  I will do the code change of checking the length is larger than 1 instead of equal to 3.

meena-sfdc:
@force2b , The changes are done and ready for review.

Julian88Tex:
@force2b do we want to implement this saming spacing with recurring autonaming too?

> Last2 ₨7 - Recurring

rponti-sforg:
QE testing is complete for this PR. Tests have passed.

https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001CAvyQAG/view
https://salesforce.quip.com/J1UVAcqgzQL3

Merging PR.

PR #4663, opened 2019-10-03
Define MetaDeploy plan for RD2 enablement with preflight check

davidmreed:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


scripts/validate_rd2_enabled.cls
```diff
@@ -1,19 +0,0 @@
-/**
```
force2b: Nice catch. I forgot that I added this.

tasks/check_rd2_enablement.py
```diff
@@ -0,0 +1,17 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+class is_rd2_enabled(BaseSalesforceApiTask):
+    def __call__(self):
+        self._init_task()
```
davisagli: Put the custom task code in the `_run_task` method instead of `__call__`; then you don't need to call `_init_task` yourself.

davidmreed: @davisagli To switch to `_run_task()`, I had it set `self.return_values` to a Boolean value instead of a dict. It looked to me on a review of the FlowRunner system like task results are type-agnostic; is that accurate?

davisagli: Ah yes, I guess that would be necessary too. I don't think anything currently makes assumptions about the type of return_values, and it's what preflight checks look at (I wanted them to be able to receive multiple values in some cases).

Issue #4591, opened 2019-09-16
Donation Record Type Name Field Missing From Batch Gift Entry


jesslopez-sf:
_**Found in: NPSP 3.161:**_
If the _Donation Record Type Name_ field on the NPSP Data Import Object does not explicitly have Help Text mapping to Opportunity.RecordTypeId, it will not appear as an option in Batch Gift Entry.

Update 9/17/2019: this issue might also manifest itself with an `Attempt to de-reference a null object` error when viewing older batch records that included the Donation Record Type Name field in its definition.

### **Workaround:**
1. Go to Setup --> Object Manager --> NPSP Data Import --> Fields and Relationships.
2. Click the Edit link for the Donation Record Type Name field.
3. Clear out any existing values in the Help Text field
4. Add the following to the Help Text field: Opportunity.RecordTypeId
5. Save.

![Screen Shot 2019-09-16 at 10 01 00 AM](https://user-images.githubusercontent.com/29731456/64964349-e879b700-d868-11e9-805f-55a75d7ece1c.png)

### **Root Cause:**
Back in February, we discovered that the **_Donation Record Type Name_** field did not appear as an option in Batch Gift Entry unless the field was explicitly mapped via Help Text. It worked fine in BDI, presumably because we had hard-coded this mapping in Apex. This may have been a field added  more recently, so trial orgs (and those with NPSP installed via our installer) did not have a help text mapping for this field.  We added this mapping in the Trial Source Org in February and also delivered a code fix in March.

That fix included the following code in BDI_DataImportService.cls, within the getTargetObject() method:
[https://github.com/SalesforceFoundation/Cumulus/pull/4081/files
](https://github.com/SalesforceFoundation/Cumulus/pull/4081/files
)
```
if (sourceFieldName == UTIL_Namespace.StrTokenNSPrefix('Donation_Record_Type_Name__c')) {
            return 'Opportunity';
}
```

We removed this reference in 3.161:
[https://github.com/SalesforceFoundation/Cumulus/commit/01f9e3ebcbc6603056dc6956de64e022c1a50cb6#diff-bfb0a307281cf9a4f894c781df109d1c
](https://github.com/SalesforceFoundation/Cumulus/commit/01f9e3ebcbc6603056dc6956de64e022c1a50cb6#diff-bfb0a307281cf9a4f894c781df109d1c
)

So, anyone who does not have this mapping will have to manually update that Help Text.

jesslopez-sf:
[...]

PR #4680, opened 2019-10-09
BGE Use Default Help Text Mapping for Donation Record Type When Not Mapped

kenlewis-sfdo:
# Critical Changes

# Changes

- We have added a default help text mapping from the NPSP Data Import Donation Record Type Name field to Opportunity Record Type that will be used when an Org is configured to use Help Text mapping and the Donation Record Type Name field on the NPSP Data Import object has no mapping present in its inline help text.

# Issues Closed

- Fixes [#4591](https://github.com/SalesforceFoundation/Cumulus/issues/4591)

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-036076

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAqZQAW" target="_blank">W-036076</a>

PR #4683, opened 2019-10-09
BDI Advanced Mapping: Allow reference fields on target object to be mapped from DI Text field

### LINKED ISSUE #4684: NPSP Data Import: Cannot map text field on DataImport__c to a Relationship field on Target Object
Currently, a DI text field cannot be mapped to a relationship field on the target object (although it can be the relationship field defined in the object mapping). There are valid scenarios where a DI text field should be mapped to a relationship field on a target object, such as an Account1 RecordType ID (which can only be stored as text on the DI, since you cannot create a relationship field to record type of a different object) but must be mapped to the RecordTypeId relationship on the Account. 

Current allowed target mappings for a DI text field are "String" and "Picklist", but this list should be expanded to include "Reference"  (validTargetTypesBySourceType in bdiFieldMappingModal.js).

Steps to reproduce:
Add text field to DataImport__c, "Account1 RecordTypeId"
Navigate to Advanced Mapping setup, Account1 Object group
New Field mapping, select "Account1 RecordTypeId" as Source Field
EXPECTED: RecordTypeId as option in target fields
ACTUAL: No RecordTypeId field

ptewson-sfdo:
This change allows text fields on the DataImport__c object to be mapped to Reference fields on the target object. This will be needed for a handful of use cases, but the first motivation here is to support record types on target objects. RecordType is a reference field on the target, but must be stored as text on the DataImport__c records as the platform does not allow lookup fields to record types.

# Critical Changes

# Changes

# Issues Closed

Fixes #4684 

# New Metadata

# Deleted Metadata


ptewson-sfdo:
**lurch: attach W-036683

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CBpDQAW" target="_blank">W-036683</a>

PR #4690, opened 2019-10-11
Edit RD2 MetaDeploy enablement check task

davidmreed:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


davidmreed:
This preflight check will be revised prior to final pilot version; it's currently pretty inelegant. This version has been tested and works with the current implementation.

PR #4692, opened 2019-10-11
Create Recurring Donation Schedule on Recurring Donation insert

npsp-reedestockton:
# Critical Changes

# Changes

# Issues Closed

# Enhanced Recurring Donations
- NPSP automatically creates Recurring Donation Schedules for new Recurring Donations.

# New Metadata
- Apex Class: RD2_ScheduleService
# Deleted Metadata


src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -102,4 +104,21 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
         }
     }
 
+    /***
+    * @description Creates schedule record after insert of recurring donation
+    * @param rds Trigger.new records
+    * @param triggerAction BeforeInsert, AfterInsert, etc.
+    */
+    private void createRecurringDonationSchedule(List<SObject> rds, TDTM_Runnable.Action triggerAction) {
```
force2b: You can specify the parameter as `List<npe03__Recurring_Donation__c>` to avoid the cast in the for loop

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
+*
+*/
+
+public with sharing class RD2_ScheduleService {
```
force2b: Given that that RDSchedule object is a protected (hidden) object, I recommend that this class use `without sharing`.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+
+    private npe03__Recurring_Donation__c parentRecurringDonation;
+    @TestVisible
```
force2b: Blank line needed here

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+    private List<RecurringDonationSchedule__c> recurringDonationSchedules = new List<RecurringDonationSchedule__c>();
+
+    private void init(npe03__Recurring_Donation__c rd, List<RecurringDonationSchedule__c> rdSchedules) {
+        parentRecurringDonation = rd;
+        if (rdSchedules != null) {
+            recurringDonationSchedules = rdSchedules;
+        }
+    }
+
+    public RD2_ScheduleService(npe03__Recurring_Donation__c rd) {
```
force2b: Rather than overloading constructors, I'm a fan of the builder pattern. For example:
```
public RD2_ScheduleService() { }

public withRecurringDonation(npe03__Recurring_Donation__c rd) {
    parentRecurringDonation = rd;
}
public withSchedules(List<RecurringDonationSchedule__c> rdSchedules) {
    recurringDonationSchedules = rdSchedules;
}
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+        init(rd, rdSchedules);
+    }
+
+    /*******************************************************************************************************
+    * @description Initializes RecurringDonationSchedule__c record(s) associated with a parent
+    * npe03__Recurring_Donation__c record (parentRecurringDonation)
+    * @return List<RecurringDonationSchedule__c>
+    */
+    public List<RecurringDonationSchedule__c> initializeSchedule() {
+        if (parentRecurringDonation != null && parentRecurringDonation.Id != null && parentRecurringDonation.Status__c != 'Closed') {
```
force2b: To reduce indentation and improve readability, how about
```suggestion
        if (parentRecurringDonation == null || parentRecurringDonation.Id == null || parentRecurringDonation.Status__c == 'Closed') {
            return recurringDonationSchedules;
        }
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+            recurringDonationSchedules.add(new RecurringDonationSchedule__c());
+            if (parentRecurringDonation.Day_of_Month__c == '1st and 15th' && parentRecurringDonation.npe03__Installment_Period__c == 'Monthly') {
```
force2b: I suggest moving "1st and 15th" to a constant at the top of the class.

force2b: I recommend abstracting out this condition to a helper method for `isFirstAndFifteenth(parentRecurringDonation)`

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
+*
+*/
+
+public with sharing class RD2_ScheduleService {
+
+    private npe03__Recurring_Donation__c parentRecurringDonation;
```
force2b: The var name for this could be something shorter. There really isn't a parent/child relationship in place here, so maybe something like `rd` or `recurringDonation`.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+
+    // TODO: Will return an inner class (TBD)
+    public List<Date> visualizeSchedule() {
+        return null;
+    }
+
+    private void generateRecurringDonationScheduleDetails() {
+        Integer recordCount = 1;
+        for (RecurringDonationSchedule__c rds : recurringDonationSchedules) {
+            rds.InstallmentAmount__c = parentRecurringDonation.npe03__Installment_Amount__c;
```
force2b: I recommend moving everything in this `if` block into a separate method so that there just a single line for something like `buildScheduleRecord(parentRecurringDonation);`. Plus that will allow the method to be used by other methods as we eventually build out the updateSchedules logic.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+            rds.InstallmentFrequency__c = parentRecurringDonation.InstallmentFrequency__c;
+            rds.InstallmentPeriod__c = parentRecurringDonation.npe03__Installment_Period__c;
+            rds.PaymentMethod__c = parentRecurringDonation.PaymentMethod__c;
+            rds.RecurringDonation__c = parentRecurringDonation.Id;
+            rds.Campaign__c = parentRecurringDonation.npe03__Recurring_Donation_Campaign__c;
+            Date refDate =
+                parentRecurringDonation.StartDate__c > parentRecurringDonation.npe03__Date_Established__c ?
+                parentRecurringDonation.StartDate__c :
+                parentRecurringDonation.npe03__Date_Established__c;
+            if (parentRecurringDonation.npe03__Installment_Period__c == 'Monthly') {
```
force2b: Same with the date logic here. Even though everything in the `if` is abstracted out, the logic here would also be separated out into a smaller method that handles the different date calculations.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,120 @@
...
+    /*******************************************************************************************************
+    * @description Initializes RecurringDonationSchedule__c record(s) associated with a parent
+    * npe03__Recurring_Donation__c record (parentRecurringDonation)
+    * @return List<RecurringDonationSchedule__c>
+    */
+    public List<RecurringDonationSchedule__c> initializeSchedule() {
+        if (parentRecurringDonation != null && parentRecurringDonation.Id != null && parentRecurringDonation.Status__c != 'Closed') {
+            recurringDonationSchedules.add(new RecurringDonationSchedule__c());
+            if (parentRecurringDonation.Day_of_Month__c == '1st and 15th' && parentRecurringDonation.npe03__Installment_Period__c == 'Monthly') {
+                recurringDonationSchedules.add(new RecurringDonationSchedule__c());
```
force2b: I think I see how you're building this for 1st and 15th, but I think it may be limiting. At the moment, we only know of a single situation where there will be two schedule records. However, that could change at some point. Rather than add a blank sobject to the collection and then call a method that populates the fields in that blank sobject. I would eliminate the `recurringDonationSchedules` public var and just have the generateRecurringDonationScheduleDetails() build and return the correct number of records.

npsp-reedestockton: All great suggestions.  Thanks!!!

npsp-reedestockton:
I'm all in on everything you suggested.  I wasn't sure about the 2 blank record approach.  The only reason I headed that way is that I thought it would be easy to reuse the method for modification, but it is seriously non-intuitive, will be hard to work with if alterations are ever needed, and that probably doesn't mean good things for durability.  Happy to rework it!

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: shorter is better, but rd to schedule is master-detail in my org.  is it supposed to be a lookup instead??

force2b: It is an MD field. I just wasn't thinking it as a parent, even if it's in the MD relationship. Plus it isn't necessary to represent the field/relationship type in the var name.

npsp-reedestockton: I'm fine with changing - no issues at all.  I was just confused by your parent/child comment.

npsp-reedestockton:
The approach is complete rewritten.  I still have testing to do, but want to get comments on the revision before proceeding any further.

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
+*
+*/
+
+public without sharing class RD2_Schedule_SVC {
+
+    public static final String FIFTEENTH = '15';
+    public static final String FIRST = '1';
+    public static final String FIRST_AND_FIFTEENTH = '1st and 15th';
+    public static final String LAST_DAY_OF_MONTH = 'Last Day';
```
force2b: I believe the correct api value for this is `Last_Day`.
https://github.com/SalesforceFoundation/Cumulus/blob/master/unpackaged/config/rd2_post_config/objects/npe03__Recurring_Donation__c.object#L181

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+        return retDate;
+    }
+
+    /***
+    * @description Determines whether '1st and 15th' is the schedule selected.
+    * @param rd Recurring Donation record
+    * @return Boolean
+    */
+    private Boolean isFirstAndFifteenth(npe03__Recurring_Donation__c rd) {
+        return rd.Day_of_Month__c == FIRST_AND_FIFTEENTH && rd.npe03__Installment_Period__c == MONTHLY;
```
force2b: The First & Fifteenth would just be `npe03__Installment_Period__c == '1st and 15th'`. Day of Month is always either a specific number (1-28) or the value `Last_Day`.
https://github.com/SalesforceFoundation/Cumulus/blob/master/unpackaged/config/rd2_post_config/objects/npe03__Recurring_Donation__c.object#L235

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+         */
+
+        rdsched.InstallmentAmount__c = rd.npe03__Installment_Amount__c;
+        rdsched.InstallmentFrequency__c = rd.InstallmentFrequency__c;
+        rdsched.InstallmentPeriod__c = rd.npe03__Installment_Period__c;
+        rdsched.PaymentMethod__c = rd.PaymentMethod__c;
+        rdsched.RecurringDonation__c = rd.Id;
+
+        if (rdsched.InstallmentPeriod__c == MONTHLY) {
+            rdsched.DayOfMonth__c = getDayOfMonth(rd);
```
force2b: I don't think this is necessary. It can be just 
```suggestion
            rdsched.DayOfMonth__c = rd.Day_of_Month__c;
```

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
+*
+*/
+
+public without sharing class RD2_Schedule_SVC {
```
force2b: For consistency with other RD2 classes, this should be
```suggestion
public without sharing class RD2_ScheduleService {
```

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+
+    public static final String FIFTEENTH = '15';
+    public static final String FIRST = '1';
+    public static final String FIRST_AND_FIFTEENTH = '1st and 15th';
+    public static final String LAST_DAY_OF_MONTH = 'Last Day';
+    public static final Integer LAST_DAY_THRESHHOLD = 29;
```
force2b: I don't believe this constant will be needed. See below.

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+    private Boolean isFirstAndFifteenth(npe03__Recurring_Donation__c rd) {
+        return rd.Day_of_Month__c == FIRST_AND_FIFTEENTH && rd.npe03__Installment_Period__c == MONTHLY;
+    }
+
+    /***
+    * @description Gets schedule day of month from recurring donation day of month
+    * @param rd Recurring Donation
+    * @return String
+    */
+    private String getDayOfMonth(npe03__Recurring_Donation__c rd) {
```
force2b: Based on usage from above, I don't think this method is necessary. The DayOfMonth value on the Schedule record should always be either an exact copy of the RD.Day_Of_Month__c field value or null (for Weekly and Yearly schedules).

src/classes/RD2_Schedule_SVC.cls
```diff
@@ -0,0 +1,185 @@
...
+                  Date.today();
+
+        if (rdSchedule.InstallmentPeriod__c == MONTHLY) {
+            Integer lastDay = Date.daysInMonth(retDate.year(), retDate.month());
+            retDate = Date.newInstance(retDate.year(), retDate.month(), rdSchedule.DayOfMonth__c == LAST_DAY_OF_MONTH ?
+                                                                        lastDay :
+                                                                        Integer.valueOf(rdSchedule.DayOfMonth__c));
+            retDate = retDate >= Date.today() ?
+                      retDate :
+                      retDate.addMonths(1);
```
force2b: I believe this assumes a Monthly schedule. It might be better to add logic to handle other schedule types such as Daily, Weekly and Yearly this way it adds the appropriate interval.

npsp-reedestockton: This section of code should check for either monthly or 1st and 15th in period, so I need to fix that instead of just testing for monthly, but other schedule types should already be handled earlier in the method.  Do you think there is adjustment needed in the logic?

force2b: Good point. I wonder if we need to eventually allow for a grace period on these. For example, what if the record was inserted into the BDI object on 10/21, but not processed - and thus not inserted into the RD object - until 10/22. In that case, the intent is for the record to have a start date of 10/21 even though it's not actually being processed until 10/22. More than likely, BDI would also be inserting/matching to the first installment.

Thinking about these use cases more, I would probably recommend that we don't push it out to the next month at all if the StartDate is before the current date. I realize this doesn't match the work item A/C's though. Let's talk through it during stand-up.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,166 @@
...
+                  Date.today();
+
+        if (rdSchedule.InstallmentPeriod__c == MONTHLY || rdSchedule.InstallmentPeriod__c == FIRST_AND_FIFTEENTH) {
+            Integer lastDay = Date.daysInMonth(retDate.year(), retDate.month());
+            retDate = Date.newInstance(retDate.year(), retDate.month(), rdSchedule.DayOfMonth__c == LAST_DAY_OF_MONTH ?
+                                                                        lastDay :
+                                                                        Integer.valueOf(rdSchedule.DayOfMonth__c));
+            retDate = retDate >= Date.today() ?
+                      retDate :
+                      retDate.addMonths(1);
```
force2b: As noted in the previous comment (that seems to be hidden now), I'm questioning whether we should recalculate the start date at all. The A/C's on the work item were updated with questions. We'll have an exact answer once Mary and I discuss.

npsp-reedestockton:
@sam-knox request review of PR change description

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -102,4 +106,46 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+        else if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            synchronizeSchedulesAfterRDUpdate(rds, oldRds);
+        }
+    }
+
+    /***
+    * @description Handles schedule creation after insert of recurring donation
+    * @param rds Trigger.new records
+    */
+    private void createNewSchedulesAfterRDInsert(List<npe03__Recurring_Donation__c> rds) {
```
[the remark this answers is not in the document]
npsp-reedestockton: Not sure I like buildSchedules.  There will be one method to handle creation and another to handle update.  So what we use should distinguish between the two.  How about buildNewSchedules?

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -162,6 +163,64 @@ private with sharing class RD2_RecurringDonations_TEST {
             'The triggerOld parameter should contain the data from the oldList parameter');
     }
 
+    /***
+    * @description Verifies schedule is created when recurring donation is inserted
+    */
+    @isTest
+    private static void shouldCreateNewRDSchedulesOnInsert() {
```
force2b: This is works as an integration test, but I think you should add a new test class for `RD2_ScheduleService_TEST` that has unit tests methods to validate the actual functionality more granularly. You could change this method to only insert one RD and just validate that the trigger runs (i.e., the integration), but then have individual methods in the service test class that tests each individual type to validate that the schedule is created correctly.

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,101 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
+*
+*/
+
+public without sharing class RD2_ScheduleService {
+
+    public static final String FIFTEENTH = '15';
```
force2b: @npsp-reedestockton The `RD2_Constants` class has been merged. You should be able to remove these constants now and replace them with references to the constants in that class.

PR #4696, opened 2019-10-15
Update GAU Allocation report description

judisohn:
Remove "(Classic Only)" from GAU Allocation & Payment report description now that joined reports are supported in Lightning.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


judisohn:
@davisagli Went to merge this and it's failing robot test because of org pref file? Don't think there's anything I can do here.

davisagli:
@judisohn that's just a warning; it's not why the build failed. There were 2 robot tests that failed (see Tests tab); probably ones that are failing intermittently, since I don't see how they could be related. You can go ahead and merge.

PR #4708, opened 2019-10-18
Support Custom Field Mappings between Engagement Plan Task and Task

jbrandolinisf:
# Critical Changes

# Changes
- Community Sprint Driven Idea to Map Matching Custom Fields between Engagement Plan Task And Task when creating the Engagement Plan from the Template
# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/EP_Task_UTIL.cls
+++ b/src/classes/EP_Task_UTIL.cls
@@ -1,10 +1,10 @@
 /*
     Copyright (c) 2016, Salesforce.org
     All rights reserved.
-    
+
     Redistribution and use in source and binary forms, with or without
     modification, are permitted provided that the following conditions are met:
-    
+
     * Redistributions of source code must retain the above copyright
       notice, this list of conditions and the following disclaimer.
     * Redistributions in binary form must reproduce the above copyright
@@ -15,16 +15,16 @@
       from this software without specific prior written permission.
 
     THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT 
-    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS 
-    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE 
-    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, 
-    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
-    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
-    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER 
-    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT 
-    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
-    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
     POSSIBILITY OF SUCH DAMAGE.
 */
 /**
@@ -51,7 +51,7 @@ public class EP_Task_UTIL {
                 }
             }
             return defaultTaskStatus;
-            
+
         }
         private set;
     }
@@ -120,6 +120,12 @@ public class EP_Task_UTIL {
     * @return void This method instantiates class properties.
     ********************************************************************************************************/
     public void initializeMaps(Set<Id> templateIds) {
+        Set<String> addlEPTaskFields = EP_Task_InterimFieldHelper_UTIL.determineAdditionalEPTaskFieldsToMap();
+        Set<String> epTaskFields = getPrimaryEPTaskFields();
+        // Determine what additional fields need to be added to the task query, if any
+        addAdditionalEPTaskFieldsIfNotPresent(epTaskFields, addlEPTaskFields);
+        System.debug('*** addlEPTaskFields: ' + addlEPTaskFields);
+        System.debug('*** epTaskFields: ' + epTaskFields);
 
         //query our custom fields and fieldset fields for when this template is accessed from the UI
         Set<String> templateFields = new Set<String>{'Id', 'Name', 'Skip_Weekends__c', 'Reschedule_To__c', 'Automatically_Update_Child_Task_Due_Date__c', 'Default_Assignee__c', 'Description__c'};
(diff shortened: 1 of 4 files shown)


jbrandolinisf:
@RadSFDeveloper - Hi Radmilla, I have not added unit tests yet for the UTIL_SimpleTestMapper class, but wanted to give you an opportunity to provide feedback while I am doing that.  Hence, I am tagging you for a preliminary round of feedback.

@force2b - FYI, I went a little further with it today, and generalized the part that determines what fields are matches into a general shared class we could reuse if we want/need to to do this with other objects.  It follows the <<class>>.getInstanceOf(...).with<<ConfigParam(...)...doSomething() that we are using in many of our newer code implementations.

I will work on the test class later this weekend, or some evening this week.

jbrandolinisf:
**lurch: attach W-017870

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000005FM6gAAG" target="_blank">W-017870</a>

src/classes/EP_Task_UTIL.cls
```diff
@@ -39,6 +39,9 @@
 */
 public class EP_Task_UTIL {
 
+    /** @description Additional Field Mappings found using UTIL_SimpleFieldMapper functionality. */
+    private Set<String> additionalFieldMappings;
```
RadSFDeveloper: Instead of using the `getAdditionalFieldMappings();` and calling it separately from code, the var can be called instead and utilize lazy load.

src/classes/EP_Task_UTIL.cls
```diff
@@ -113,35 +116,39 @@ public class EP_Task_UTIL {
...
     /*******************************************************************************************************
     * @description Queries and builds maps for templates and template tasks from a set of template Ids.
     * @param templateIds a set of Engagement Plan Template Ids to query and build maps for.
     * @return void This method instantiates class properties.
     ********************************************************************************************************/
     public void initializeMaps(Set<Id> templateIds) {
+        Set<String> addlEPTaskFields = this.getAdditionalFieldMappings();
+        Set<String> epTaskFields = getPrimaryEPTaskFields();
+        // Determine what additional fields need to be added to the task query, if any
+        addAdditionalEPTaskFieldsIfNotPresent(epTaskFields, addlEPTaskFields);
```
RadSFDeveloper: Since the `addlEPTaskFields`, ie `additionalFieldMappings` is not used in this method, then it can be called in `addAdditionalEPTaskFieldsIfNotPresent` without being passed as the input argument. So lines 125-128 can be replaced with
```
Set<String> epTaskFields = getPrimaryEPTaskFields();
epTaskFields.addAll(getAdditionalEPTaskFieldsIfNotPrimary().values());
```

and lines 174-183 can be replacced with 
```
taskEPTaskFieldMap.putAll(getAdditionalEPTaskFieldsIfNotPrimary());
```

where `addAdditionalEPTaskFieldsIfNotPresent` is replaced with the following method:
```
private Map<String, String> getAdditionalEPTaskFieldsIfNotPrimary() {
        Map<String, String> fields = new Map<String, String>();
        Set<String> primaryFields = toLowerCase(getPrimaryEPTaskFields());

        for (String fieldName : additionalEPTaskFields) {
            if (!primaryFields.contains(fieldName.toLowerCase())) {
                fields.put(fieldName, fieldName);
            }
        }

        return fields;
    }
```

src/classes/EP_Task_UTIL.cls
```diff
@@ -367,4 +385,90 @@ public class EP_Task_UTIL {
...
+
+    /*******************************************************************************************************
+    * @description Returns a Set containing the primary fields that need to be/are retrieved
+    *              as part of EP Task SOQL queries
+    * @return Set<String> Set of String API field names for the Enrollment_Plan_Task__c Object to
+    *                     be retrieved
+    ********************************************************************************************************/
+    private Set<String> getPrimaryEPTaskFields() {
+        return new Set<String> {'Id',
+                                'Name',
```
RadSFDeveloper: Indentation is off here and in the next method.

src/classes/UTIL_SimpleFieldMapper.cls
```diff
@@ -0,0 +1,282 @@
...
+public with sharing class UTIL_SimpleFieldMapper {
+
+    /** @description Exception Class for problems that occur during runtime use */
+    public class SimpleFieldMapperException extends Exception {}
+
+    /** @description SObjectType for the Source Object to determine overlapping fields from */
+    @testVisible private Schema.SObjectType sourceSObjectType;
+
+    /** @description Source fields map for the Source Object */
+    @testVisible private Map<String, Schema.SObjectField> sourceFields;
```
RadSFDeveloper: The `sourceFields` and `targetFields` can use lazy load and can be called from other methods instead `getSourceFields()` and `getTargetFields()` respectively.

src/classes/UTIL_SimpleFieldMapper.cls
```diff
@@ -0,0 +1,282 @@
...
+        'Id',
+        'Name',
+        'RecordType',
+        'CreatedById',
+        'CreatedDate',
+        'IsDeleted',
+        'LastModifiedById',
+        'LastModifiedDate'
+    };
+    // TODO: Confirm list of default fieldsToExclude values is sufficient
```
RadSFDeveloper: Let's clarify `TODO` statements and implement them where required.

src/classes/UTIL_SimpleFieldMapper.cls
```diff
@@ -0,0 +1,282 @@
...
+                candidateFields.add(candidateField);
+            }
+        }
+
+        // Check if candidate fields are compatible and then add to mappableFields Set
+        for (String candidateField : candidateFields) {
+            Schema.SObjectField sourceField = sourceFields.get(candidateField);
+            Schema.SObjectField targetField = targetFields.get(candidateField);
+
+            if (isCompatibleFieldMapping(sourceField, targetField)) {
```
RadSFDeveloper: Can we avoid this loop and use `if (isCompatibleFieldMapping(sourceField, targetField))` in the above loop (line 183)?

src/classes/UTIL_SimpleFieldMapper.cls
```diff
@@ -0,0 +1,282 @@
...
+
+    /*********************************************************************************************************
+    * @description Checks to see if the source and target fields are compatible
+    * @param sourceField Field from Source Object
+    * @param targetField Field from Target Object
+    * @return Boolean true, if the fields can be mapped.
+    */
+    @testVisible private Boolean isCompatibleFieldMapping(Schema.SObjectField sourceField, Schema.SObjectField targetField) {
+
+        Map<Schema.DisplayType, Set<Schema.DisplayType>> validFieldMappingCombinations = new Map<Schema.DisplayType, Set<Schema.DisplayType>> {
```
RadSFDeveloper: This mapping can be extracted as the class var.

src/classes/EP_Task_UTIL.cls
```diff
@@ -39,6 +39,9 @@
 */
 public class EP_Task_UTIL {
```
RadSFDeveloper: This class does not have its own test class but is (in)directly tested in the EP_* classes where it is being called. Do we need new or adjusted existing tests in the EP_*_TEST classes to verify new functionality?

jbrandolinisf: It would likely need regression flows that specifically create and map custom fields on the Plan Template and Task objects.  That felt out of scope for me, for the amount of time I had to work on this.

anuragoyar:
# Potential Risks and Suggestions:

1.  
src/classes/EP_Task_UTIL.cls  
Code Snippet: private Set<String> additionalFieldMappings;  
Comment: This new instance variable is initialized lazily but not marked with @TestVisible. Consider adding @TestVisible to ensure it can be properly tested in unit tests.

2.  
src/classes/EP_Task_UTIL.cls  
Code Snippet: Set<String> addlEPTaskFields = this.getAdditionalFieldMappings();  
Comment: The method getAdditionalFieldMappings() could return null if there's an issue with UTIL_SimpleFieldMapper. Consider adding null checking before using the returned value.

3.  
src/classes/UTIL_SimpleFieldMapper.cls  
Code Snippet: public class SimpleFieldMapperException extends Exception {}  
Comment: This exception class is defined but never used in the implementation. Either use it for error handling or consider removing it.

4.  
src/classes/UTIL_SimpleFieldMapper_TEST.cls  
Code Snippet: @isTest private static void determineMappableFieldsBetweenContactAndAccount()  
Comment: This test relies on standard objects (Account and Contact) which may have different fields in different orgs. Consider creating test-specific custom objects with known fields to make the test more deterministic.

5.  
src/classes/UTIL_SimpleFieldMapper_TEST.cls-meta.xml  
Code Snippet: <ApexClass xmlns="urn:metadata.tooling.soap.sforce.com" fqn="UTIL_String_TEST">  
Comment: The fqn attribute references "UTIL_String_TEST" instead of "UTIL_SimpleFieldMapper_TEST". Update this to match the actual class name.

# Summary:

This PR introduces a new field mapping utility that enhances the EP_Task_UTIL class to dynamically map fields between Engagement Plan Tasks and standard Salesforce Tasks. The changes include:

1. New Components:
   - UTIL_SimpleFieldMapper.cls - A utility class for identifying compatible fields between two SObject types
   - UTIL_SimpleFieldMapper_TEST.cls - Test class for the new utility

2. Modified Components:
   - EP_Task_UTIL.cls - Enhanced to use the new field mapper for dynamic field mapping
   - .forceignore - Minor whitespace change

The core functionality added is the ability to automatically map fields with identical names between Engagement Plan Tasks and standard Tasks, beyond just the predefined mappings. This makes the system more flexible when custom fields are added to both objects.

The implementation uses a "lazy loading" approach for field mappings and includes safeguards to exclude certain fields from automatic mapping. The code is well-structured with proper separation of concerns and includes comprehensive test coverage.

The changes appear to be backward compatible as they enhance existing functionality without breaking existing behavior. The new utility class is designed to be reusable across the codebase.

No metadata changes to objects, fields, or permissions are included in this PR, reducing deployment risk.

## Deployment Considerations:

1. The changes should be deployed together as a unit since EP_Task_UTIL.cls now depends on UTIL_SimpleFieldMapper.cls.
2. Test coverage appears adequate with dedicated test classes.
3. No data migration is needed as this is enhancing functionality without changing data structures.
4. The implementation is defensive with proper error handling and null checking.

This change is suitable for production deployment after addressing the minor issues noted above. The risk is relatively low as it's enhancing existing functionality rather than changing core behavior.

PR #4723, opened 2019-10-24
Create Recurring Donation Schedule on Recurring Donation insert

npsp-reedestockton:
# Critical Changes

# Changes

# Issues Closed

# Enhanced Recurring Donations

# New Metadata
- Apex Classes
RD2_ScheduleService
RD2_ScheduleServiceTest

# Deleted Metadata


src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,96 @@
...
+
+        /* TODO: For fixed length recurring donations we may need to calculate this
+        rdsched.EndDate__c = ??
+         */
+
+        rdsched.InstallmentAmount__c = rd.npe03__Installment_Amount__c;
+        rdsched.InstallmentFrequency__c = rd.InstallmentFrequency__c;
+        rdsched.InstallmentPeriod__c = rd.npe03__Installment_Period__c;
+        rdsched.PaymentMethod__c = rd.PaymentMethod__c;
+        rdsched.StartDate__c = rd.StartDate__c;
```
force2b: As a failsafe against possibly data inserted directly into the RD object, I suggest changing this to use the DateEstablished if the rd.StartDate field is empty.
```suggestion
        rdsched.StartDate__c = (rd.StartDate__c != null ? rd.StartDate__c : rd.Date_Established__c);
```

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+    /***
+    * @description Verifies details of schedule creation
+    */
+    @isTest
+    private static void shouldCreateNewRDScheduleOnInsert() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+
+        Contact contact = getContact();
+
+        List<npe03__Recurring_Donation__c> rds = new List<npe03__Recurring_Donation__c>{
```
force2b: I'd like to suggest an alternative approach to this unit test that does not actually insert any data. The ContactId can be mocked using `UTIL_UnitTestData_TEST.mockId(Contact.SObjectType);`. Plus, the TEST_RecurringDonationBuilder class can be called with the `mockId()` method to mock an RD Id.

In this class, which is testing the actual service as opposed to the trigger test which is in a different class, this method would simply call the `RD2_ScheduleService.createSchedule()` method directly with the mocked RD and then validate the results from the `getRDSchedulesToInsert()`. No DML necessary.

With this approach,  this class can have multiple tests to validate different schedule types:

- shouldCreateMonthlySchedule()
- shouldCreateFirstandFifteenthSchedule()
- shouldCreateYearlySchedule()
- shouldCreateWeeklySchedule()

The assert for these would be more specific than just validating the number of records, but actually verifying that the records are as expected - the correct frequency, period, start date, etc.

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+                new TEST_RecurringDonationBuilder()
```
force2b:
```suggestion
                TEST_RecurringDonationBuilder.constructEnhancedBuilder()
                    .withMockId()
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,96 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Class with methods to create, modify and visualize RecurringDonationSchedule__c records.
```
force2b:
```suggestion
* @description All services specific to the Recurring Donation Schedule object
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -102,4 +106,32 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+
+    /***
+    * @description Handles schedule creation after insert of recurring donation
+    * @param rds Trigger.new records
+    */
+    private void buildNewSchedules(List<npe03__Recurring_Donation__c> rds) {
+
+        for (npe03__Recurring_Donation__c rd: rds) {
+            dmlActions.objectsToInsert.addAll(
+                    new RD2_ScheduleService()
```
RadSFDeveloper: Since service methods need to be independent, we can call one method instead. Would the following work wrt future requirements and changes?
```
RD2_ScheduleService scheduleService = new RD2_ScheduleService();

    dmlActions.objectsToInsert.addAll(scheduleService.getNewSchedules(rd));  
}
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,96 @@
...
+    public List<RecurringDonationSchedule__c> getRDSchedulesToInsert() {
+        return schedulesToInsert;
+    }
+
+    /***
+    * @description Creates new recurring donation schedule(s) for recurring donation
+    * @param rd Recurring donation record
+    * @return RD2_Schedule_SVC
+    */
+    public RD2_ScheduleService createSchedule(npe03__Recurring_Donation__c rd) {
```
RadSFDeveloper: If my previous comment is valid, then following modifications can be made:
- `createSchedule()` can be renamed to `getNewSchedules()`
- `getRDSchedulesToInsert()` removed and 
- `private` class variable `schedulesToInsert` can be removed and replaced with local `newSchedules` variable?

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,96 @@
...
+
+        RecurringDonationSchedule__c rdsched = new RecurringDonationSchedule__c();
```
RadSFDeveloper: `rdsched` -> `schedule`

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,96 @@
...
+        rdsched.EndDate__c = ??
+         */
+
+        rdsched.InstallmentAmount__c = rd.npe03__Installment_Amount__c;
+        rdsched.InstallmentFrequency__c = rd.InstallmentFrequency__c;
+        rdsched.InstallmentPeriod__c = rd.npe03__Installment_Period__c;
+        rdsched.PaymentMethod__c = rd.PaymentMethod__c;
+        rdsched.StartDate__c = (rd.StartDate__c != null ? rd.StartDate__c : rd.Date_Established__c);
+
+        if (!isFirstAndFifteenth(rd)) {
```
RadSFDeveloper: We can write `if (isFirstAndFifteenth(rd))` block first and avoid check for its complement:

```
if (isFirstAndFifteenth(rd)) {
    schedule.DayOfMonth__c = RD2_Constants.DAY_OF_MONTH_FIRST;
    newSchedules.add(schedule);

    schedule = schedule.clone();
    schedule.DayOfMonth__c = RD2_Constants.DAY_OF_MONTH_FIFTEENTH;

} else {
    schedule.DayOfMonth__c = rd.Day_of_Month__c;
}
```

npsp-reedestockton: @force2b and I have discussed this previously and we have had it both ways at different points.  I think it comes down to which version is more readable: use a negative test with the most common block first, or use a positive test with the least common block first.  I'm happy to change it as you suggest, but if it is important enough to any of us to discuss further, we should probably have a short call.

npsp-reedestockton:
**lurch: attach W-035248

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9JpQAK" target="_blank">W-035248</a>

src/classes/RD2_ScheduleService.cls
```diff
@@ -0,0 +1,89 @@
...
+    /***
+    * @description Builds new recurring donation schedule(s) for recurring donation
+    * @param rd Recurring donation record
+    * @return List<RecurringDonationSchedule__c>
+    */
+    public List<RecurringDonationSchedule__c> getNewSchedules(npe03__Recurring_Donation__c rd) {
+
+        List<RecurringDonationSchedule__c> newSchedules = new List<RecurringDonationSchedule__c>();
+
+        RecurringDonationSchedule__c schedule = new RecurringDonationSchedule__c();
```
force2b: I recommend adding the following
```
if (rd.Status__c == RD2_Constants.STATUS_CLOSED) {
    return newSchedules;
}
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -116,4 +120,32 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+    /***
+    * @description Handles schedule creation after insert of recurring donation
+    * @param rds Trigger.new records
+    */
+    private void buildNewSchedules(List<npe03__Recurring_Donation__c> rds) {
+
+        RD2_ScheduleService scheduleService = new RD2_ScheduleService();
+
+        for (npe03__Recurring_Donation__c rd: rds) {
+            if (rd.Status__c != RD2_Constants.STATUS_CLOSED) {
```
force2b: I think this check should be in the actual `getNewSchedules()` method rather than here. I added a comment in that method as well.

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -162,6 +163,40 @@ private with sharing class RD2_RecurringDonations_TEST {
...
+    * @description Verifies schedule is created when recurring donation is inserted
+    */
+    @isTest
+    private static void shouldCreateNewRDScheduleOnInsert() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+
+        Contact contact = getContact();
+
+        List<npe03__Recurring_Donation__c> rds = new List<npe03__Recurring_Donation__c>{
+                new TEST_RecurringDonationBuilder()
```
force2b: This builder class supports both legacy and RD2 format RD's. To know exactly which format you're creating, I recommend using the following format.
```suggestion
                TEST_RecurringDonationBuilder.constructEnhancedBuilder()
```

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -0,0 +1,172 @@
...
+                    .withPaymentMethod('Check')
+                    .withInstallmentPeriodMonthly()
+                    .withInstallmentFrequency(1)
+                    .withDayOfMonth('20')
+                    .withDateEstablished(Date.newInstance(2019, 9, 15))
+                    .withStartDate(Date.newInstance(2019, 9, 15))
+                    .build()
+        };
+
+        System.debug('rds - ' + rds);
```
force2b: System.debug statements should be removed

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -0,0 +1,172 @@
...
+@isTest(IsParallel=true)
+private with sharing class RD2_ScheduleService_TEST {
+
+    /***
+    * @description Verifies details of monthly schedule creation
+    */
+    @isTest
+    private static void shouldCreateMonthlySchedule() {
+
+        List<npe03__Recurring_Donation__c> rds = new List<npe03__Recurring_Donation__c>{
```
force2b: Though this works - creating the list of RD's, it's not needed for any of these tests. This could just be a simple assignment to a single `rd` var. That would also eliminate the need for using `rds[0]` in the call to `getNewSchedules()`

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -0,0 +1,172 @@
...
+        System.assertEquals(schedules[0].InstallmentFrequency__c, 13, 'Installment Frequency should be 13');
+        System.assertEquals(schedules[0].InstallmentPeriod__c, RD2_Constants.FREQUENCY_WEEKLY, 'Installment Period should be Weekly');
+        System.assertEquals(schedules[0].PaymentMethod__c, 'Check', 'Payment Method should be check');
+        System.assertEquals(schedules[0].StartDate__c, Date.newInstance(2019, 11, 1), 'Start Date should be 11/1/2019');
+        System.assertEquals(schedules[0].DayOfMonth__c, null, 'Day of Month should be null');
+        System.assertEquals(schedules[0].EndDate__c, null, 'End Date should be null');
+
+    }
+
+}
```
force2b: I recommend one more test for `shouldNotCreateAScheduleWhenClose()`

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: Mild pushback on this one.  For monthly, the number of schedules created will always be 1 IF the code is functioning corrrectly.  However, there are situations where we produce more than 1 schedule (1st and 15th at this time).  The point of this test is to make sure that in this scenario we aren't inadvertently executing some code that creates >1 schedule.

force2b: True, though this declaration is for `List<npe03__Recurring_Donation__c>` as opposed to `List<RecurringDonationSchedule__c>`. I'm ok with it staying this way as well since it works. I was just noting that all the tests here create a single RD. It may be that future tests will mock multiple RD's for bulk testing in which case this is a good pattern to have in place.

src/objects/RecurringDonationSchedule__c.object
```diff
@@ -153,6 +153,18 @@
     <enableSearch>false</enableSearch>
     <enableSharing>true</enableSharing>
     <enableStreamingApi>true</enableStreamingApi>
+    <fields>
+        <fullName>Active__c</fullName>
+        <description>Establishes whether the Recurring Donation Schedule is active or inactive. Read only</description>
```
force2b: @Sam We decided to add this new formula field to centralize the logic for what an Active schedule record is. Would you mind reviewing the label and help text on these?

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: Well, I suppose it makes some sense to expect the point I made to actually apply to the requested change.  ;-) . I'm going to push the change initially requested.  If we change things later, we can go back to my forward-thinking pattern. <grin>

src/objects/RecurringDonationSchedule__c.object
npsp-reedestockton: @sam-knox Also add the new metadata to the PR info.

sam-knox: @force2b - Gotcha! That's true. So End Date on the RD is always set by the related Schedule record then. They're always in sync.

force2b: There is no end date on the RD record. The Period, Frequency, DayOfMonth and StartDate fields on the RD are what is used to create/update Schedule records. Schedule records aren't visible or editable, so they don't drive anything on the RD; other than the calculation for Next Donation Date and eventually CurrentYearExpectedValue.

sam-knox: Got it!

PR #4730, opened 2019-10-24
Add automation documentation

davidmreed:
This PR adds overview documentation summarizing the product automation for the NPSP.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


documentation/automation.md
```diff
@@ -0,0 +1,142 @@
...
+| Development              | `dev_org`                      | dev            |         |            |
+| Development (Namespaced) | `dev_org_namespaced`           | dev_namespaced |         | ✔          |
+| QA                       | `qa_org`                       | dev            |         |            |
+| QA (Namespaced)          | `qa_org_namespaced`            | dev            |         | ✔          |
+| Regression               | `regression_org`               | regression     | ✔       |            |
+| Trial (Clone of TSO)     | None (`cci org browser trial`) | trial          | ✔       |            |
+
+### Org Definitions
+
+NPSP includes many org definitions tailored for testing - in particular, beta testing - of the product in a variety of org shapes.
```
force2b: Most of these orgs are identical in configuration. Do you think it would help to list out that configuration here, maybe just specifying where they differ?

documentation/automation.md
```diff
@@ -0,0 +1,142 @@
...
+
+This data set includes most NPSP objects and is scaled based on 100 Contacts.
+
+| Deploy Task         | Capture Task | Delete Task        |
+| ------------------- | ------------ | ------------------ |
+| `test_data_dev_org` |              | `test_data_delete` |
+
+For managed orgs, the mapping must be manually set to `datasets/mapping-managed.yml`.
+
+### 1K Data Set
```
force2b: I'm not really sure this flow works. When I tried it again this morning, it fails with the following error:
`Error on row 600: DUPLICATES_DETECTED:Duplicate Alert:--` while attempting to upload Contacts.

davidmreed: It definitely doesn't work. (Or rather, it won't work on any org that has duplicate rules active).

I have a to-do list item to fix that by deactivating Duplicate Rules. Let me know if that needs to be higher priority.

documentation/automation.md
davidmreed: I think that's a great idea. Added.

documentation/automation.md
force2b: Ah. Is it worth trying to fix the data so that the dupe rule doesn't need to be deactivated? If I remember the CCI training at the sprint, it sounded like there is a theoretical process of:
1. Deploy the data into a scratch org (without dupe rules)
2. Fix the data
3. run the `extract_dataset` task to retrieve the updated data

Though I suspect it's a bit more complicated than that.

documentation/automation.md
force2b: Thank you!

documentation/automation.md
davidmreed: That's pretty much it, honestly. I think that process _should_ work, but given that this is a 1k scaled data set, it may be more labor than it's worth - especially if nobody is actively using this data set.

PR #4732, opened 2019-10-25
Gift Entry - Form Rendering

bdvorachek:
# Critical Changes

# Changes
* Gift Entry form will now render for most field types.
* Picklist values still forthcoming.
# Issues Closed

# New Metadata
* Apex Classes
  * FORM_RenderWrapper
  * GE_FormRendererService
* Lightning Components
  * geFormField
  * geFormRenderer
  * geFormSection
  * geFormService
* Aura Components
  * geFormRendererPlaceholder
# Deleted Metadata

# Testing Notes

In order to test this, Advanced Mapping must be turned on and a form template must exist in your org. See here for instructions - https://salesforce.quip.com/AxEnAN3xcSrM


src/lwc/geFormSection/geFormSection.js
```diff
@@ -0,0 +1,39 @@
...
+    get iconName() {
+        return this.expanded ? 'utility:chevrondown' : 'utility:chevronright';
+    }
+
+    /**
+     * Get the alternative text that represents the section expand/collapse button
+     * @returns {string} containing the section expand alternative text
+     */
+    get altTextLabel() {
+        return 'Toggle ' + this.section.label;
```
lparrott: This was me being lazy, but we'll want this to be a custom label.

src/lwc/geFormService/geFormService.js
```diff
@@ -0,0 +1,70 @@
+import getRenderWrapper from '@salesforce/apex/GE_TemplateBuilderCtrl.retrieveDefaultFormRenderWrapper';
+
+const inputTypeByDescribeType = {
```
lparrott: More of a QA item (and I know this is just a first pass!), but we're missing PERCENT here, which I guess could just be number for now?

src/lwc/geFormSection/geFormSection.html
```diff
@@ -0,0 +1,24 @@
...
+            {section.label}
+        </h2>
+    </button>
+    <lightning-layout multiple-rows='true'>
+        <template for:each={section.elements} for:item='element'>
+            <lightning-layout-item key={element.value} class={sectionClassName}
+                    padding='around-small' 
+                    size='6' 
+                    medium-device-size='3'>
+                    {element.label}
```
vmartinov-sforg: Remove this line, I think I added it while debugging

bdvorachek:
Helptext is behaving like this in our temporary lookup component - 
image

I think this is ok for now, we'll resolve when we migrate to a lightning-input component for the lookup.

src/lwc/geFormService/geFormService.js
bdvorachek: PERCENT should now render. I had it added as a valid lightning-input type, but it didn't have a mapping in this object. It does now, and will render as a number lightning-input.

PR #4738, opened 2019-10-27
RD2: Migrate Recurring Donation fields

RadSFDeveloper:
Migrate existing Recurring Donation data to the enhanced Recurring Donation format.

# Critical Changes

# Changes

# Issues Closed

# New Metadata
Custom Labels:
- RD2_DataMigrationInvalidInstallmentPeriod

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_Constants.cls
+++ b/src/classes/RD2_Constants.cls
@@ -37,7 +37,7 @@
 public inherited sharing class RD2_Constants {
 
     /**
-    * @description npe03_RecurringDonation.InstallmentPeriod__c picklist API values
+    * @description Recurring Donation "Installment Period" picklist API values
     */
     public static final String INSTALLMENT_PERIOD_YEARLY = 'Yearly';
     public static final String INSTALLMENT_PERIOD_MONTHLY = 'Monthly';
@@ -46,21 +46,21 @@ public inherited sharing class RD2_Constants {
     public static final String INSTALLMENT_PERIOD_DAILY = 'Daily';
 
     /**
-    * @description npe03_RecurringDonation.Day_Of_Month__c picklist API values
+    * @description Recurring Donation "Day of Month" picklist API values
     */
     public static final String DAY_OF_MONTH_FIRST = '1';
     public static final String DAY_OF_MONTH_FIFTEENTH = '15';
     public static final String DAY_OF_MONTH_LAST_DAY = 'Last_Day';
 
     /**
-    * @description npe03_RecurringDonation.Status__c picklist API values
+    * @description Recurring Donation "Status" picklist API values
     */
     public static final String STATUS_ACTIVE = 'Active';
     public static final String STATUS_LAPSED = 'Lapsed';
     public static final String STATUS_CLOSED = 'Closed';
 
     /**
-    * @description npe03__RecurringDonation.RecurringType__c picklist API values
+    * @description Recurring Donation "Recurring Type" picklist API values
     */
     public static final String RECURRING_TYPE_OPEN = 'Open';
 
--- a/src/classes/RD2_Constants.cls
+++ b/src/classes/RD2_Constants.cls
@@ -43,6 +43,7 @@ public inherited sharing class RD2_Constants {
     public static final String INSTALLMENT_PERIOD_MONTHLY = 'Monthly';
     public static final String INSTALLMENT_PERIOD_WEEKLY = 'Weekly';
     public static final String INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH = '1st and 15th';
+    public static final String INSTALLMENT_PERIOD_DAILY = 'Daily';
 
     /**
     * @description npe03_RecurringDonation.Day_Of_Month__c picklist API values
--- a/src/classes/RD2_Constants.cls
+++ b/src/classes/RD2_Constants.cls
@@ -37,29 +37,29 @@
 public inherited sharing class RD2_Constants {
 
     /**
-    * @description Api Values of the npe03_RecurringDonation.Installment_Frequency__c picklist
+    * @description npe03_RecurringDonation.InstallmentPeriod__c picklist API values
     */
-    public static final String FREQUENCY_MONTHLY = 'Monthly';
-    public static final String FREQUENCY_WEEKLY = 'Weekly';
-    public static final String FREQUENCY_YEARLY = 'Yearly';
-    public static final String FREQUENCY_FIRST_AND_FIFTEENTH = '1st and 15th';
+    public static final String INSTALLMENT_PERIOD_YEARLY = 'Yearly';
+    public static final String INSTALLMENT_PERIOD_MONTHLY = 'Monthly';
+    public static final String INSTALLMENT_PERIOD_WEEKLY = 'Weekly';
+    public static final String INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH = '1st and 15th';
 
     /**
-    * @description Api Values of the npe03_RecurringDonation.Day_Of_Month__c picklist
+    * @description npe03_RecurringDonation.Day_Of_Month__c picklist API values
     */
     public static final String DAY_OF_MONTH_FIRST = '1';
     public static final String DAY_OF_MONTH_FIFTEENTH = '15';
     public static final String DAY_OF_MONTH_LAST_DAY = 'Last_Day';
 
     /**
-    * @description Api Values of the npe03_RecurringDonation.Status__c picklist
+    * @description npe03_RecurringDonation.Status__c picklist API values
     */
     public static final String STATUS_ACTIVE = 'Active';
     public static final String STATUS_LAPSED = 'Lapsed';
     public static final String STATUS_CLOSED = 'Closed';
 
     /**
-    * @description Api Values of the npe03__RecurringDonation.RecurringType__c picklist
+    * @description npe03__RecurringDonation.RecurringType__c picklist API values
     */
     public static final String RECURRING_TYPE_OPEN = 'Open';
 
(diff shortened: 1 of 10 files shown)


RadSFDeveloper:
**lurch: attach W-035562

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9tRQAS" target="_blank">W-035562</a>

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -60,22 +60,127 @@ public class RD2_DataMigrationMapper {
...
+            System.Label.npe03.RecurringDonationInstallmentPeriodYearly,
+            System.Label.npe03.RecurringDonationInstallmentPeriodMonthly,
+            System.Label.npe03.RecurringDonationInstallmentPeriodWeekly
+        };
+
+        if (quarterlyInstallmentPeriods.contains(rd.npe03__Installment_Period__c)) {
+            rd.InstallmentFrequency__c = 3;
+            rd.npe03__Installment_Period__c = RD2_Constants.INSTALLMENT_PERIOD_MONTHLY;
+
+        } else if (frequencyOneInstallmentPeriods.contains(rd.npe03__Installment_Period__c) && rd.InstallmentFrequency__c == null) {
```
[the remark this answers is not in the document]
RadSFDeveloper: @force2b , conversion based on the custom installment settings has been applied.

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -60,22 +60,166 @@ public class RD2_DataMigrationMapper {
...
     }
 
     /**
-     * @description Returns true if the RD record is already in RD2 format
+     * @description Sets Installment Frequency and Installment Period for custom Installment Period
+     * @param rd Recurring Donation in the legacy format
+     * @return void
      */
-    @TestVisible
-    private Boolean isConverted() {
```
RadSFDeveloper: @force2b , I have removed `isConverted()` since its check is not valid. The reason is that `Status` will be populated for every new legacy Recurring Donation created after RD2 schema has been deployed. We can reintroduce `isConverted()` but we need additional/other checks. Currently when data migration batch is being executed *again* , converted RDs will be picked up and converted again.

RadSFDeveloper: `isConverted()` has been added back with additional check on "Installment Frequency". Should we check for any other field?

src/labels/CustomLabels.labels
```diff
@@ -1318,6 +1318,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>Processing</shortDescription>
         <value>Processing</value>
     </labels>
+    <labels>
+        <fullName>RD2_DataMigrationInvalidInstallmentPeriod</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Installment Period cannot be converted to enhanced Recurring Donation format</shortDescription>
+        <value>Recurring Donation Installment Period "{0}" cannot be converted to enhanced Recurring Donation format.</value>
```
RadSFDeveloper: @sam-knox , can you please advise on the message displayed when Installment Period cannot be converted?

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -55,27 +55,185 @@ public class RD2_DataMigrationMapper {
...
         return rd;
     }
 
+    /**	
+     * @description Returns true if the RD record is already in enhanced Recurring Donation format	
+     */	
+    public Boolean isConverted() {	
+        return String.isNotBlank(rd.Status__c) 
+            && rd.InstallmentFrequency__c != null;	
+            //frequency is null for "1st and 15th" installment period even when converted
```
force2b: There should _always_ be a value in the Frequency field for RD2 format RD's. Though we do not have any validation or logic set to do this. It's possible that an RD inserted through an integration might not set the InstallmentFrequency field. It's probably necessary to default to this field value to 1 on the object itself so that any new records will have a value in the field; which would mean that even RD1 format RD's (created after we change the default value on the field) would have a value.

For now, let's always set the field to a value in the `setInstallmentFields()` method.

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -55,27 +55,185 @@ public class RD2_DataMigrationMapper {
...
+            //check if the RD schedule exists //TODO
```
force2b: An RD schedule will not exist for a converted RD if the RD status is Closed, but the RD is still converted.

src/labels/CustomLabels.labels
sam-knox: @RadSFDeveloper - What would the advice be to a user in this scenario? Should they re-create manually somehow, or are we saying that certain custom installment periods simply won't work in RD2?

src/classes/RD2_DataMigrationMapper.cls
RadSFDeveloper: - What would Installment Frequency be when Installment Period is "1st and 15th"?
- If we set Installment Frequency to 1 as default, legacy RD records created after the schema deployment will be considered as converted but they are not.

src/labels/CustomLabels.labels
RadSFDeveloper: The custom installment needs to conform to "Days", "Weeks", "Months" and "Years". I assume users would need to update their RDs  to those installments. @force2b , thoughts?

src/classes/RD2_DataMigrationMapper.cls
force2b: - Installment Frequency will generally be 1 for 1st and 15th, but in theory it could be something else.
- Very good point on that. Status__c defaults to Active, so any RD records created after we deployed the Status__c field into orgs in the last release would have a value of "Active" in them. Let's leave Installment Frequency with a null default for now, but conversion can still set it to 1 for 1st and 15th.

Given the above, how about the following for Is Converted:
- Status__c == null
- OR (Status != 'Closed' AND {no attached Schedule records})

That would catch older RD records (the first condition) and newer RD records that haven't been converted (second condition)

src/labels/CustomLabels.labels
force2b: I'm not sure what you mean for this? Custom Installments have a Name field that the Admin adds to the Picklist. RD2 post installment instructions will have to include a note about asking them to remove those picklist values from the field, but the User shouldn't need to change the RD since migration will apply the converted schedule type to the record.
https://powerofus.force.com/s/article/NPSP-Installmt-Periods-Recur-Dons

sam-knox: @force2b - So does this mean we don't need an error message for this scenario?

force2b: An error is still possible if there is a Picklist value in the InstallmentPeriod field that does not map to something we can understand. That would be Weekly, Monthly, Yearly, Quarterly, 1st and 15th, and any custom names that are defined in the CustomInstallmentPeriod settings object. For example, if they create a picklist value called "Eventually" and there is no record in CustomInstallmentPeriod with a name of "Eventually" that would generate an error.

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -55,27 +55,185 @@ public class RD2_DataMigrationMapper {
...
+        }
+    }
+
+    /**
+     * @description Sets Installment Frequency and Installment Period for custom Installment Period
+     * @param rd Recurring Donation in the legacy format
+     * @return void
+     */
+    private void handleCustomInstallmentPeriod(npe03__Recurring_Donation__c rd) {
+        npe03__Custom_Installment_Settings__c settings = UTIL_ListCustomSettingsFacade.getMapCustomInstallmentSettings()
```
force2b: @RadSFDeveloper This looks correct. You're trying to match the installment period period picklist value to a custom installment period and if there is a match then you set the InstallmentPeriod field correctly and apply the installment frequency from the custom installment record. 

When there is no match then an exception is thrown (which will eventually be a log action, but exception can work for now).

I'm not necessarily sure there is anything to fix here.

RadSFDeveloper: Thanks @force2b for the verification, I have updated tests to set the legacy RD installment period to the custom installment *name* and verified tests pass.

src/labels/CustomLabels.labels
RadSFDeveloper: @sam-knox , the change to the text message has been applied.

RadSFDeveloper:
@rponti-sforg , RD field migration is ready for testing.

src/classes/RD2_Constants.cls
```diff
@@ -37,29 +37,30 @@
...
 
     /**
-    * @description Api Values of the npe03_RecurringDonation.Installment_Frequency__c picklist
+    * @description Recurring Donation "Installment Period" picklist API values
     */
-    public static final String FREQUENCY_MONTHLY = 'Monthly';
-    public static final String FREQUENCY_WEEKLY = 'Weekly';
-    public static final String FREQUENCY_YEARLY = 'Yearly';
-    public static final String FREQUENCY_FIRST_AND_FIFTEENTH = '1st and 15th';
+    public static final String INSTALLMENT_PERIOD_YEARLY = 'Yearly';
```
force2b: @RadSFDeveloper  Thanks for noticing this and fixing. I'm not sure why I used Frequency instead of Period.

rponti-sforg:

For further testing details and comments, see:
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001C9tRQAS/view
https://salesforce.quip.com/9656A2CGxgtx

Merged PR.  WI status set to Closed.

PR #4751, opened 2019-10-30
RD2: Create Schedule records during data migration

meena-sfdc:
# Critical Changes

# Changes
When Data Migration occurs, a schedule record will be created from the values of the Recurring Donation when status is active.

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_DataMigrationMapper.cls
+++ b/src/classes/RD2_DataMigrationMapper.cls
@@ -81,15 +81,6 @@ public class RD2_DataMigrationMapper {
     }
 
     /**
-     * @description Generate the RDSchedule Record(s) for the Recurring Donation
-     * @return List<RecurringDonationSchedule__c> Schedule records to insert
-     */
-    public List<RecurringDonationSchedule__c> getScheduleRecords() {
-        RD2_ScheduleService scheduleService = new RD2_ScheduleService();
-        return scheduleService.getNewSchedules(rd);
-    }
-
-    /**
      * @description Sets Installment Frequency and Installment Period fields where applicable
      * @param rd Recurring Donation in the legacy format
      * @return void
@@ -232,6 +223,15 @@ public class RD2_DataMigrationMapper {
      */
     private void setOtherFields(npe03__Recurring_Donation__c rd) {
         rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
+    }
+
+    /**
+     * @description Builds the Schedule Record(s) for the Recurring Donation
+     * @return List<RecurringDonationSchedule__c> Schedule records to insert
+     */
+    public List<RecurringDonationSchedule__c> getScheduleRecords() {
+        RD2_ScheduleService scheduleService = new RD2_ScheduleService();
+        return scheduleService.getNewSchedules(rd);
     }
 
     public class MigrationException extends Exception { }
--- a/src/classes/RD2_DataMigrationMapper.cls
+++ b/src/classes/RD2_DataMigrationMapper.cls
@@ -68,7 +68,8 @@ public class RD2_DataMigrationMapper {
      * @return List<RecurringDonationSchedule__c> Schedule records to insert
      */
     public List<RecurringDonationSchedule__c> getScheduleRecords() {
-        return new List<RecurringDonationSchedule__c>();
+        RD2_ScheduleService scheduleService = new RD2_ScheduleService();
+        return scheduleService.getNewSchedules(rd);
     }
 
     /**
--- a/src/classes/RD2_DataMigration_TEST.cls
+++ b/src/classes/RD2_DataMigration_TEST.cls
@@ -584,12 +584,13 @@ private class RD2_DataMigration_TEST {
     private static void shouldNotCreateScheduleForClosedEnhancedRDs() {
 
         List<RecurringDonationSchedule__c> rdSchedules  = new List<RecurringDonationSchedule__c>();
+        Id contactId = UTIL_UnitTestData_TEST.mockId(Contact.SObjectType);
 
-        npe03__Recurring_Donation__c rd = getLegacyRecurringDonationBuilder(UTIL_UnitTestData_TEST.mockId(Contact.SObjectType))
-            .withStatusClosed()
+        npe03__Recurring_Donation__c rd = getLegacyRecurringDonationBuilder(contactId)
+            .withOpenEndedStatusClosed()
             .build();
 
-        RD2_DataMigrationMapper mapper = new RD2_DataMigrationMapper(rds);
+        RD2_DataMigrationMapper mapper = new RD2_DataMigrationMapper(rd);
         mapper.convertToEnhancedRD();
 
         rdSchedules = mapper.getScheduleRecords();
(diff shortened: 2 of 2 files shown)


src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -131,6 +131,22 @@ private class RD2_DataMigration_TEST {
             'This RD should be in Enhanced RD format');
     }
 
+     /**
+     * @description Verifies that a schedule record is created for Enhanced Recurring Donation record.
+     */
+    @IsTest
+    private static void shouldCreateScheduleForEnhancedRDs() {
+        List<RecurringDonationSchedule__c> rdSchedules  = new List<RecurringDonationSchedule__c>();
```
RadSFDeveloper: - The first argument in `System.assert` is expected value and the second one is actual value.
- The schedule record can be built only for the enhanced RD, so we need to convert the record before creating the schedule.
- Do we need to check any field on the schedule?

Based on the above, we can rewrite the test as:
```
Id contactId = UTIL_UnitTestData_TEST.mockId(Contact.SObjectType);
npe03__Recurring_Donation__c rd = getNewRecurringDonationEnhancedFormat(contactId);

RD2_DataMigrationMapper mapper = new RD2_DataMigrationMapper(rd);
mapper.convertToEnhancedRD();

List<RecurringDonationSchedule__c> schedules = mapper.getScheduleRecords();

System.assertEquals(1, schedules.size(), 'Number of schedules should match');
```

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -68,7 +68,8 @@ public class RD2_DataMigrationMapper {
      * @return List<RecurringDonationSchedule__c> Schedule records to insert
      */
     public List<RecurringDonationSchedule__c> getScheduleRecords() {
-        return new List<RecurringDonationSchedule__c>();
+        RD2_ScheduleService scheduleService = new RD2_ScheduleService();
```
RadSFDeveloper: The schedule record can be built only for the enhanced RD, so we need a check for `isConverted()` before returning schedule(s).
If the call is made to get schedules, but the record is not converted, let's generate an exception.

meena-sfdc: @RadSFDeveloper, if the record is not converted, do you suggest to throw an custom exception here or error message?

RadSFDeveloper: Custom exception: new `public class MigrationException extends Exception { }` defined in the `RD2_DataMigrationMapper`.

force2b: I don't think the check is required here. This method will be called by the RD2_DataMigration_BATCH.handleConversion() method. In that method, it will only call this if the RD itself was converted. As a result, this method will never be called unless the RD was converted by the RD2_DataMigrationWrapper.

meena-sfdc: @force2b and @RadSFDeveloper , Can you confirm if this check is not needed? I was about to make these changes.

meena-sfdc: @force2b , I see it is handled in batch class, will not do these change then, thanks!

RadSFDeveloper: @meena-sfdc , the check is not needed, but let's make changes to the test method.

RadSFDeveloper:
**lurch: attach W-036727

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CBuNQAW" target="_blank">W-036727</a>

src/classes/RD2_DataMigrationMapper.cls
force2b: Good call out on the tests.

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -131,6 +131,54 @@ private class RD2_DataMigration_TEST {
...
+        Id contactId = UTIL_UnitTestData_TEST.mockId(Contact.SObjectType);
+        npe03__Recurring_Donation__c rd = getNewRecurringDonationEnhancedFormat(contactId);
+
+        RD2_DataMigrationMapper mapper = new RD2_DataMigrationMapper(rd);
+        mapper.convertToEnhancedRD();
+
+        rdSchedules = mapper.getScheduleRecords();
+
+        System.assertEquals(1, rdSchedules.size(), 'Number of RDs should be 1');
```
RadSFDeveloper: The message should indicate `Number of schedules should match`. Current message is indicating number of RDs is being verified.

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -131,6 +131,54 @@ private class RD2_DataMigration_TEST {
...
+
+    /**
+     * @description Verifies that a schedule record is not created for Closed Enhanced Recurring Donation record.
+     */
+    @IsTest
+    private static void shouldNotCreateScheduleForClosedEnhancedRDs() {
+
+        List<RecurringDonationSchedule__c> rdSchedules  = new List<RecurringDonationSchedule__c>();
+
+        npe03__Recurring_Donation__c rds =
```
RadSFDeveloper: `rds` -> `rd` since we have one Recurring Donation.

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -131,6 +131,54 @@ private class RD2_DataMigration_TEST {
...
+            TEST_RecurringDonationBuilder.constructEnhancedBuilder()
+                .withMockId()
+                .withContact(UTIL_UnitTestData_TEST.mockId(Contact.SObjectType))
+                .withAmount(400)
+                .withPaymentMethod('Check')
+                .withInstallmentPeriodWeekly()
+                .withInstallmentFrequency(1)
+                .withDateEstablished(Date.newInstance(2019, 9, 15))
```
RadSFDeveloper: Why these particular dates and numbers? Can we reuse the RD construction as done in the previous test but just set Open Ended Status to Closed before RD is being converted?

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -131,6 +131,54 @@ private class RD2_DataMigration_TEST {
...
+                .withStartDate(Date.newInstance(2019, 11, 1))
+                .withRecurringDonationStatusClosed()
+                .build();
+
+        RD2_DataMigrationMapper mapper = new RD2_DataMigrationMapper(rds);
+        mapper.convertToEnhancedRD();
+
+        rdSchedules = mapper.getScheduleRecords();
+
+        System.assertEquals(0, rdSchedules.size(), 'Number of RDs should be 0');
```
RadSFDeveloper: The assert message should indicate number of schedules is being verified.

src/classes/RD2_DataMigration_TEST.cls
meena-sfdc: I am using the new method .withRecurringDonationStatusClosed() in that class to mark the status as closed, Do you want to change it?

Julian88Tex:
@RadSFDeveloper @force2b Testing is complete and will merge as soon as builds complete successfully. Great job @meena-sfdc!

PR #4764, opened 2019-11-02
[RD] Synchronize Schedule object based on RD info

npsp-reedestockton:
In Enhanced Recurring Donations, NPSP will synchronize recurring donation schedules to reflect updates to recurring donations.
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


npsp-reedestockton:
@force2b @RadSFDeveloper @sam-knox Had to create new PR - naming convention error on the branch (one underscore instead of 2)

npsp-reedestockton:
**lurch: attach W-036113

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAtYQAW" target="_blank">W-036113</a>

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -173,31 +173,121 @@ private with sharing class RD2_RecurringDonations_TEST {
...
+        };
+
+        Test.startTest();
+        insert rds;
+        rds[0].InstallmentFrequency__c = 3;
+        update rds;
+        Test.stopTest();
+
+        Map<Id, npe03__Recurring_Donation__c> actualRdById = new Map<Id, npe03__Recurring_Donation__c>(rdGateway.getRecords(rds));
+        System.assertEquals(rds.size(), actualRdById.size(), 'Number of RDs should match');
```
force2b: If this is retrieving records by Id, is it even possible for this to ever return more than the single record Id that it was passed originally?

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -173,31 +173,121 @@ private with sharing class RD2_RecurringDonations_TEST {
...
+
+        List<RecurringDonationSchedule__c> rdSchedules = rdScheduleGateway.getRecords(rds);
+        System.assertEquals(rds.size()+1, rdSchedules.size(), 'Number of RD schedule records should be one greater than the number of RD records');
```
force2b: Would be good if this verified the schedule record changes in additional to fact that a new schedule record was created.

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -182,10 +182,593 @@ private with sharing class RD2_ScheduleService_TEST {
...
+        rd2.Status__c = RD2_Constants.STATUS_CLOSED;
+
+        RD2_ScheduleService schedule = new RD2_ScheduleService();
+        Boolean updateNeeded = schedule.recurringDonationScheduleUpdateNeeded(rd2, rd);
+
+        System.assertEquals(false, updateNeeded, 'recurringDonationScheduleUpdateNeeded should return false.');
+
+    }
+
+
```
force2b: If it's possible (given the really tight timeline), it would be great to have unit tests against the `getUpdatedSchedules` method. Maybe something like
- shouldCreateNewMonthlySchedule (when schedule changed)
- shouldInactivateCurrentSchedule (when the RD is changed to Closed)
- shouldChangeFromFirstAndFifteenthToWeekly

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -139,11 +176,71 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+            if (scheduleService.recurringDonationScheduleUpdateNeeded(rd, oldRd)) {
+                filteredRecurringDonations.put(rd.Id, rd);
+            }
+        }
+
+        Map<Id, List<RecurringDonationSchedule__c>> oldSchedules = new Map<Id, List<RecurringDonationSchedule__c>>();
+        for (RecurringDonationSchedule__c oldSchedule : [
+                SELECT RecurringDonation__c, EndDate__c FROM RecurringDonationSchedule__c
+                WHERE RecurringDonation__c IN :filteredRecurringDonations.keySet()
+                AND (EndDate__c = null OR EndDate__c >= TODAY)]) {
```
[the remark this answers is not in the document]
npsp-reedestockton: Yes, I think so.  That method was written before we decided to add IsActive.  I'll use IsActive.

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: I can complete these this evening.

src/classes/RD2_ScheduleService.cls
```diff
@@ -90,4 +263,33 @@ public without sharing class RD2_ScheduleService {
...
+    * @description Determines whether changes to recurring donation should trigger creation of new schedule(s)
+    * @param rd Trigger.new record
+    * @param oldRd Trigger.old record
+    * @return Boolean
+    */
+    public Boolean recurringDonationScheduleUpdateNeeded(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
+
+        Boolean updateNeeded = false;
+
+        if (rd.Status__c == RD2_Constants.STATUS_CLOSED || rd.Status__c == RD2_Constants.STATUS_LAPSED) {
```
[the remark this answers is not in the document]
npsp-reedestockton: so my take on your comment is that  I should also change getNewSchedules so that it does not create new schedules (leaving old schedule in place, but not create a new schedule).  Correct?

`    public List<RecurringDonationSchedule__c> getNewSchedules(npe03__Recurring_Donation__c rd) {

        List<RecurringDonationSchedule__c> newSchedules = new List<RecurringDonationSchedule__c>();

        if (rd.Status__c == RD2_Constants.STATUS_CLOSED || rd.Status__c == RD2_Constants.STATUS_LAPSED) {
            return newSchedules;
        }
`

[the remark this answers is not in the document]
force2b: If the RD Status is changed from Active to Closed, then yes - apply EndDate to the current open schedules and not create a new schedule.
If the RD Status is changed from Active to Lapsed, then do nothing.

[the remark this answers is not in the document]
npsp-reedestockton: shouldn't we also apply the end date if the change is from lapsed to closed?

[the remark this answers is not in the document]
force2b: Yes. Good point. I left out that use case. I guess the first item should have been "If the RD Status is changed from Active/Lapsed to Closed, then apply EndDate to the current open schedules".

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -139,11 +176,71 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+    }
+
+    /***
+    * @description Handles schedule modification/creation after modification of recurring donation
+    * @param rds Trigger.new records
+    * @param oldRds Trigger.old records
+    */
+    private void synchronizeSchedulesAfterRDUpdate(List<npe03__Recurring_Donation__c> rds, List<npe03__Recurring_Donation__c> oldRds) {
+
+        Map<Id, npe03__Recurring_Donation__c> mapOldRecurringDonationsByIds = new Map<Id, npe03__Recurring_Donation__c>();
```
force2b:
```suggestion
        Map<Id, npe03__Recurring_Donation__c> oldRdsById = new Map<Id, npe03__Recurring_Donation__c>();
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -139,11 +176,71 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+        mapOldRecurringDonationsByIds.putAll(oldRds);
```
force2b: I recommend moving this block of lines (215 to 225) into a new method getRDsNeedingScheduleUpdate() that returns a list<RD>. That would make this a bit cleaner to read through.

src/classes/RD2_ScheduleService.cls
```diff
@@ -81,6 +90,174 @@ public without sharing class RD2_ScheduleService {
         return newSchedules;
     }
 
+    /***
+    * @description Synchronizes the schedule records to match an updated recurring donation
+    * @param rd Recurring Donation record
+    * @param schedules Recurring Donation Schedule record
+    * @return TDTM_Runnable.DMLWrapper
+    */
+    public TDTM_Runnable.DMLWrapper getUpdatedSchedules(npe03__Recurring_Donation__c rd, List<RecurringDonationSchedule__c> schedules) {
```
force2b: I like that this single method returns back both the new and updated schedules, though I haven't really figured out if DMLWrapper is the right construct. It clearly works and it is the only data type we have in the package to meet this objective. And, it's certainly fine for now. Let's talk about it later this week or next though to see if there's an alternative approach that doesn't feel so much like it's tied to TDTM. There may not be, but it's probably worth the hypothetical discussion.

src/classes/RD2_ScheduleService.cls
```diff
@@ -277,8 +277,7 @@ public without sharing class RD2_ScheduleService {
 
         Boolean updateNeeded = false;
 
-        if (newRd.Status__c == RD2_Constants.STATUS_LAPSED ||
-                (oldRd.Status__c == RD2_Constants.STATUS_LAPSED && newRd.Status__c == RD2_Constants.STATUS_ACTIVE)) {
+        if (oldRd.Status__c == RD2_Constants.STATUS_LAPSED && newRd.Status__c == RD2_Constants.STATUS_ACTIVE) {
```
force2b: @npsp-reedestockton  What happens if the status was Active is is changed to Lapsed?

npsp-reedestockton: Yep.  That is an issue.  

I'm going to take the time now to rework this qualification logic.  It is getting way too spread out.  It is starting to remind me of the Ptolemaic explanations for planetary orbits.  Something wrong?  Add a new epicycle...

We need a single source of truth and one place to maintain the logic.

src/classes/RD2_ScheduleService.cls
```diff
@@ -121,6 +122,40 @@ public without sharing class RD2_ScheduleService {
         return scheduleChanges;
     }
 
+    /***
+    * @description Determines whether changes to recurring donation should trigger creation of new schedule(s)
+    * @param rd Trigger.new record
+    * @param oldRd Trigger.old record
+    * @return Boolean
+    */
+    public Boolean isScheduleOrNextDonationDateUpdateNeeded(npe03__Recurring_Donation__c newRd, npe03__Recurring_Donation__c oldRd) {
```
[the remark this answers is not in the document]
npsp-reedestockton: newRd will be null on afterUpdate and that is the point of the test.  this method is intended to be used on both before and after updates.

an update is not always needed, at least in the sense meant here, in an insert context.  for example, when a new rd is created, neither schedules nor next donation date should be updated, and those are the updates this methods controls.

we don't always want nextdonationdate updated.

Julian88Tex:
Testing has completed for PR has been merged for WI: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001CAtYQAW/view

PR #4767, opened 2019-11-04
RD2: Fix Unit Test Failures

force2b:
- Update the RD2 Unit Tests to further test RD2 trigger isolation and to leverage the CalculateNextDonation logic in the TestRecurringDonationsBuilder.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_EnablementService_TEST.cls
+++ b/src/classes/RD2_EnablementService_TEST.cls
@@ -48,10 +48,11 @@ public with sharing class RD2_EnablementService_TEST {
 
     /****
     * @description Verifies a new Opportunity is created on Recurring Donations insert
-    * when enhanced Recurring Donations are not enabled
+    * when Enhanced Recurring Donations are not enabled and that no RD Schedule Records are
+    * created which ensures that the RD2 trigger is not executing.
     */
     @isTest
-    private static void shouldCreateNewOppOnInsertWhenEnhancementIsNotEnabledWhenRD2TriggerIsDisabled() {
+    private static void shouldCreateNewOppOnInsertWhenEnhancementIsNotEnabled() {
         setRecurringDonations2Disabled();
 
         npe03__Recurring_Donation__c rd = new TEST_RecurringDonationBuilder()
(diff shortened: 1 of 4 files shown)


src/classes/RD2_EnablementService_TEST.cls
```diff
@@ -51,14 +51,16 @@ public with sharing class RD2_EnablementService_TEST {
     * when enhanced Recurring Donations are not enabled
     */
     @isTest
-    private static void shouldCreateNewOppOnInsertWhenEnhancementIsNotEnabled() {
+    private static void shouldCreateNewOppOnInsertWhenEnhancementIsNotEnabledWhenRD2TriggerIsDisabled() {
```
RadSFDeveloper: Is it valid to rename to `shouldCreateNewOppOnInsertWhenRD2TriggerIsDisabled()`?

src/classes/RD2_EnablementService_TEST.cls
```diff
@@ -69,17 +71,20 @@ public with sharing class RD2_EnablementService_TEST {
 
     /****
     * @description Verifies a new Opportunity is not created on Recurring Donations insert
-    * when enhanced Recurring Donations are enabled
+    * when enhanced Recurring Donations are enabled and the RD2 Trigger has been disabled.
+    * This verifies that RD2 functionality is isolated from RD1 functionality.
     */
     @isTest
-    private static void shouldNotCreateNewOppOnInsertWhenEnhancementIsEnabled() {
+    private static void shouldNotCreateNewOppOnInsertWhenEnhancementIsEnabledIfRD2TriggerIsDisabled() {
```
RadSFDeveloper: Since the AC that was valid before is not valid currently due to new requirements, we do not need this test, thus I would suggest we delete it. The test itself has contradiction in the name. Also, this is a test that tests code not what user expects or what requirements are.

src/classes/RD2_EnablementService_TEST.cls
force2b: I do agree that the method name should be changed back. 

As per our quip chat, the RD2_RecurringDonation_TDTM trigger will always be executed via TDTM - even if the `RecurringDonationSettings__c.IsRecurringDonations2Enabled__c` field is set to false. This method can validate the code isolation by making sure that there are no RecurringDonatinoSchedule records created by the RD2 trigger if RD2 is not enabled.

src/classes/RD2_EnablementService_TEST.cls
force2b: I agree that original test method is no longer valid.

However, like the previous test, this validates isolation by ensuring that when RD2 is enabled, the RD1 triggers are not running to create the installment opp. It does this by enabling RD2 and disabling the RD2 trigger. If code isolation is working then neither trigger will execute and no Installment Opp will be created. If code isolation was not working, then the RD1 trigger _might_ run and create the installment opp. 

What would you think about a method name like `shouldNotCreateNewOppOnInsertIfRD2TriggerIsDisabledWhenEnhancedRDIsEnabled`?

PR #4778, opened 2019-11-07
NPSP: BDI-AM fix bug caused by case mismatch on fieldname map key

### LINKED ISSUE #4744: NPSP - BDI - Donation matching fails when matching against Opportunity (and not payment)
**Issue:**

If a customer is using BDI or BGE to match an opportunity, the match will not find any matched records. This occurs only when matching against an open opportunity with no unpaid payments.

**Steps to repeat:**

Create Contact & Opportunity, $100, Closed Date = today Stage = Pledged
Remove any payments from the newly created Opportunity
Set up Batch, set Donation matching rule to match Donation on Amount and Date
Populate Data Import:
- Contact Imported (Contact from previous)
- Donation Donor = Contact1
- Donation Date = today
- Donation Amount = $100 
Process Batch with matching to Single match or create

**Expected Result:** 

Match to previously created donation

**Actual Result**

Donation created new

ptewson-sfdo:
The map of Data Import Field -> Target field generated from Advanced Mapping was using a mixed case field name for the map key, unlike the help text variant which was forcing map keys to lowercase. There was at least one bug attached to this behavior, and this work correctly lowercases the map keys for advanced mapping.

# Critical Changes

# Changes

# Issues Closed

Fixes #4744 

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_FieldMappingCustomMetadata.cls
+++ b/src/classes/BDI_FieldMappingCustomMetadata.cls
@@ -133,7 +133,7 @@ public with sharing class BDI_FieldMappingCustomMetadata implements BDI_FieldMap
         if (diFieldMappings != null) {
             for (Data_Import_Field_Mapping__mdt difm : diFieldMappings) {
 
-                targetFieldByDataImportField.put(difm.Source_Field_API_Name__c,difm.Target_Field_API_Name__c);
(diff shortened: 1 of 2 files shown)


ptewson-sfdo:
**lurch: attach W-037136

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HnwxQAC" target="_blank">W-037136</a>

PR #4784, opened 2019-11-12
[RD] Refactor Next Donation Date Method

npsp-reedestockton:
Simplified getNextDonationDateGreaterThanToday()
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


src/classes/RD2_ScheduleService.cls
```diff
@@ -207,72 +207,73 @@ public without sharing class RD2_ScheduleService {
...
+    /***
+    * @description Returns the date units between an input date and current date based on installment period
+    * @param firstValid Date
+    * @param period String
+    * @return Integer
+    */
+    private Integer getDateUnitsBetweenFirstValidAndCurrentDate(Date firstValid, String period) {
+        Integer unitsBetween = 0;
 
+        if (period == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY || period == RD2_Constants.INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH) {
```
force2b: If this method is only called from the `getNextDonationDateGreaterThanToday` method, and that method only takes an RDSchedule record, then the period will never be INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH. Does that impact the logic or the results?

src/classes/RD2_ScheduleService.cls
```diff
@@ -207,72 +207,73 @@ public without sharing class RD2_ScheduleService {
...
+    * @param period String
+    * @return Date
+    */
+    private Date addDateUnits(Date dateIn, Integer addUnits, String period) {
+        Date adjustedDate = dateIn;
 
-            unitsBetween = firstValid.daysBetween(currentDate);
-            remains = Math.mod(unitsBetween, frequency);
-            adjusted = firstValid.addDays((unitsBetween / (frequency)) * (frequency));
+        if (period == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY || period == RD2_Constants.INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH) {
```
force2b: I think the same is true here. This method never gets a Period with INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH.

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -241,6 +241,60 @@ private with sharing class RD2_ScheduleService_TEST {
 
     }
 
+    /***
```
force2b: What might be helpful in this test class is to create unit tests that explicitly validate the results of the `getNextDonationDate` method without going through the full service. That will result is smaller test methods that can be easier to manipulate and create multiple variations for. For example:

```
@IsTest
private static void shouldCalculateNddWhenSomeCondition() {
    RecurringDonationSchedule__c rds =  new RecurringDonationSchedule__c(
       InstallmentPeriod__c = RD2_Constants.PERIOD_MONTHLY,
       InstallmentFrequency__c = 2,
       StartDate = Date.newInstance(2019,6,1)
    );
    RD2_ScheduleService scheduleService = new RD2_ScheduleService();
    scheduleService.currentDate = Date.newInstance(2019,11,15);
    Date nextDonationDate = scheduleService.getNextDonationDate(new List<RecurringDonationSchedule__c> { rds  });
    System.assertEquals( ... );
}
```

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: There is a call that takes a recurring donation and ends up calling `getNextDonationDateGreaterThanToday`

`    public Date getNextDonationDate(npe03__Recurring_Donation__c rd) {
        if (rd.Status__c == RD2_Constants.STATUS_CLOSED) {
            return null;
        }

        return getNextDonationDate(getNewSchedules(rd));
    }
`

Also, schedules can have 1st and 15th period.

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: There are already a bunch of those.  Are there some you see should be added?

```
    @isTest
    private static void shouldCreateNextDonationDateFromWeeklyStartDate() {

        npe03__Recurring_Donation__c rd =
            TEST_RecurringDonationBuilder.constructEnhancedBuilder()
                .withMockId()
                .withContact(UTIL_UnitTestData_TEST.mockId(Contact.SObjectType))
                .withAmount(400)
                .withPaymentMethod('Check')
                .withInstallmentPeriodWeekly()
                .withInstallmentFrequency(1)
                .withDateEstablished(Date.newInstance(2019, 9, 15))
                .withStartDate(Date.newInstance(2018, 3, 17))
                .build();

        RD2_ScheduleService service = getScheduleServiceWithCurrentDateOverride(Date.newInstance(2018, 3, 17));

        System.assertEquals(Date.newInstance(2018, 3, 17),
            service.getNextDonationDate(rd),
            'Next Donation Date should equal 3/17/2018');

    }
```

npsp-reedestockton: Is your suggestion to avoid creating from RD then?

src/classes/RD2_ScheduleService.cls
force2b: Ah. I see what you mean. The `RecurringDonationSchedule` record retain the period of "`st and 15th", but the DayOfMonth field on the 2 records get a 1 and 15 respectively. I though that at one point these were actually both set to Monthly. This way works fine as well though. I just mis-remembered how the records were created.

src/classes/RD2_ScheduleService_TEST.cls
force2b: I see your point, however there are two notes with this:

- This is not actually testing building the NDD off the Schedule Records which is important. It's always recreating the Schedules from the RD record itself - which effectively ignores any existing Schedule records and any special scheduling logic they may implement (such as skips, etc.).
- You can use `withCalculateNextDonationDate()` in the factory to have it automatically calculate the NDD and apply to the NextPaymentDate field.

npsp-reedestockton: I see your point.  Sound good!  

I'll make adjustments to what is already there if that sounds okay to you, since as you said we really need to test the behavior with a collection of schedules as well.

force2b:
@npsp-reedestockton Did you mean to close this?

npsp-reedestockton:
@force2b I merged the code into the branch I created for visualization.  It was looking to me like we would never actually get to this as a separate work item.

PR #4785, opened 2019-11-12
Fix for DI Mapping fields being unavailable 

### LINKED ISSUE #4768: BDI Advanced Mapping: Mapped fields unavailable for Donation Matching rule without help text mapping
**Summary**

With Advanced Mapping enabled, mapped fields are not available for use with the donation matching configuration when processing Data Imports through BDI, unless they have help text mapping as well. This implies that when the process screen configuration is loaded, the help text mappings are being referenced when populating the donation matching rule options. There's no current indication that this has a broader effect on processing Data Imports or batches. 

**Workaround**

Specifying the equivalent mapping in help text on the Data Import field will make it available for use in the donation matching rule.

**Steps to Repeat**

1. In a fresh scratch org, enable BDI advanced mapping
2. Add a new text field on Opportunity "Test_Text__c"
3. Add a new text field on DataImport__c "Donation_Test_Text__c"
4. Add a new field mapping on the Opportunity object group "Donation_Test_Text__c" to "Test_Text__c" on Opportunity 
5. Create a new DataImport__c object (data unimportant)
6. Click "Start Data Import" 
7. Edit configuration options

EXPECTED: "Donation_Test_Text__c" field is available for donation matching 
ACTUAL: "Donation_Test_Text__c" is NOT available for matching

Secondarily, one can verify that including help text mapping "Opportunity.Test_Text__c" (or equivalent namespaced field) for the Data Import field will now allow the field to be available for inclusion in the donation matching config. 

PUB thread: https://powerofus.force.com/s/feed/0D51E000055eDGy

voduyemi:
The BDI_Setting UI was using the Helptext Field Mapping by default instead of reading what the default mapping of the user is based on the DI import Settings. This fix makes the BDI Settings UI determine field mapping based on the DI import settings defined
# Critical Changes

# Changes

# Issues Closed
#4768 

# New Metadata

# Deleted Metadata


src/classes/BDI_SettingsUI_CTRL.cls
```diff
@@ -94,7 +94,7 @@ public with sharing class BDI_SettingsUI_CTRL {
 
                 // this UI is specific to help text settings in current state
```
kenlewis-sfdo: Lets update this comment since this is no longer specific to help text mapping.

voduyemi: @kenlewis-sfdo I totally agree with writing some tests for it. And I'll update the comment.

src/classes/BDI_SettingsUI_CTRL.cls
```diff
@@ -94,7 +94,7 @@ public with sharing class BDI_SettingsUI_CTRL {
...
                 BDI_Donations donationImportHandler = 
-                    new BDI_Donations(new BDI_DataImportService(true, BDI_FieldMappingHelpText.getInstance()));
+                    new BDI_Donations(new BDI_DataImportService(true, BDI_DataImportService.getDefaultFieldMapping()));
```
ptewson-sfdo: This is a small comment, but can add a line break so that the display in github is a bit better? I think we have suggestions for how to break lines in the style guide somewhere

ptewson-sfdo:
**lurch: attach W-037351

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HoMRQA0" target="_blank">W-037351</a>

PR #4789, opened 2019-11-13
RD2: Nightly Batch Framework

force2b:
Framework for automated process whereby each Active recurring donation is evaluated to process past due installments, create the next installment if required, update the Next Donation Date, and other required automation.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

Apex Classes:
- RD2_OpportunityEvaluation_BATCH
- RD2_OpportunityEvaluation_TEST

Custom Labels:
- RD2_ErrorEnhancedRDBatchJobCannotBeRun
- RD_ErrorLegacyBatchJobCannotBeRun

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -1576,6 +1576,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>The job can only be executed when Enhanced RD is enabled in the org</shortDescription>
         <value>You must enable Enhanced Recurring Donations before you can run the Enhanced Recurring Donations Data Migration job.</value>
     </labels>
+    <labels>
```
force2b: @sam-knox Would you mind reviewing thees two labels? Ideally an Admin will never see these in their org. They're here in case someone makes an attempt to manually execute the wrong RD jobs based on whether RD2 is enabled or not.

force2b:
**lurch: attach W-035566

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9tlQAC" target="_blank">W-035566</a>

src/classes/RD2_EnablementDelegate_CTRL.cls
```diff
@@ -141,10 +141,12 @@ public with sharing class RD2_EnablementDelegate_CTRL extends STG_Panel {
     @AuraEnabled
     public static void enableEnhancement() {
         try {
-            STG_Panel.stgService.stgRD.isRecurringDonations2Enabled__c = true;
+            STG_Panel.stgService.stgRD.IsRecurringDonations2Enabled__c = true;
 
             if (!Test.isRunningTest()) {
                 upsert STG_Panel.stgService.stgRD;
+                UTIL_MasterSchedulableHelper.setScheduledJobs();
+                UTIL_OrgTelemetry_SVC.asyncProcessNpspSettingsTelemetry();
```
force2b: @RadSFDeveloper I added this second line because I realized we never connected enablement to the feature parameter for telemetry. This follows the same logic we have for Customizable Rollups where the FeatureParameter is updated immediately after the setting is changed. That allows the information to be sync'd into Wilma that evening instead of waiting for the next package push to get the state of RD2.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -100,21 +118,41 @@ public class RD2_OpportunityEvaluationService {
...
-        // TODO Recalculate NextDonationDate
-        // Date nextDonationDate = scheduleService.calculateNextDonationDate(rd);
+        if (disableRollupsBeforeDml) {
+            TDTM_ProcessControl.toggleTriggerState('Opportunity', 'CRLP_Rollup_TDTM', false);
+            TDTM_ProcessControl.toggleTriggerState('npe01__OppPayment__c', 'CRLP_Rollup_TDTM', false);
+            TDTM_ProcessControl.toggleTriggerState('Allocation__c', 'CRLP_Rollup_TDTM', false);
+        }
 
+        // Create the next installment opportunity where needed
+        // TODO Refactor this to return a set of Opps to be inserted as opposed to having it do the DML in the service
```
RadSFDeveloper: Do we have a WI that will handle this TODO?

src/classes/RD2_OpportunityEvaluation_TEST.cls
```diff
@@ -0,0 +1,119 @@
...
+@IsTest(IsParallel=true)
+private class RD2_OpportunityEvaluation_TEST {
+    private static final TEST_SObjectGateway.RecurringDonationGateway rdGateway = new TEST_SObjectGateway.RecurringDonationGateway();
+
+    /**
+     * @description Verify that when RD2 is enabled in an org, attempting to execute the older RD1 batch
+     * job throws an exception
+     */
+    @IsTest
+    private static void shouldThrowExceptionForRD1JobIfRD2IsEnabled() {
```
RadSFDeveloper: `RD_RecurringDonations_BATCH` verification belongs in `RD_RecurringDonations_TEST`.

src/classes/RD2_OpportunityEvaluation_TEST.cls
```diff
@@ -0,0 +1,119 @@
...
+        System.assert(actuaException.getMessage().containsIgnoreCase(System.Label.RD2_ErrorRD1BatchJobCannotBeRun),
+            'The exception should have indicated that the job cannot be executed because RD2 is enabled ' + actuaException.getMessage());
+    }
+
+    /**
+     * @description Verify that when RD2 is NOT enabled in an org, attempting to execute the new RD2 batch
+     * job throws an exception
+     */
+    @IsTest
+    private static void shouldThrowExceptionForRD2JobIfRD2IsNotEnabled() {
```
RadSFDeveloper: For readability and shared understanding, let's use legacy for RD1 and enhanced for RD2 in method/var names and descriptions.
For example, this test method can be named as `shouldThrowExceptionWhenEnhancedRDsAreNotEnabled`

src/classes/RD2_OpportunityEvaluation_TEST.cls
```diff
@@ -0,0 +1,119 @@
...
+        System.assert(actuaException.getMessage().containsIgnoreCase(System.Label.RD_ErrorRD2BatchJobCannotBeRun),
+            'The exception should have indicated that the job cannot be executed because RD2 is not enabled ' + actuaException.getMessage());
+    }
+
+    /**
+     * @description Verify that when RD2 is enabled in an org, executing the new RD2 batch job does not
+     * throw an exception
+     */
+    @IsTest
+    private static void shouldNotThrowExceptionForRD2JobIfRD2IsEnabled() {
```
RadSFDeveloper: `shouldNotThrowExceptionWhenEnhancedRDsAreEnabled`

src/classes/RD_RecurringDonations_BATCH.cls
```diff
@@ -93,16 +93,21 @@ Schedulable {
         // this must return a QueryLocator, or else uncaught exceptions in the
         // execute() method will not properly update the processed/failure
         // counts (see W-2443023).
-        return Database.getQueryLocator(query);      
+
+        if (RD2_EnablementService.isRecurringDonations2Enabled) {
```
RadSFDeveloper: If we are stopping legacy RD batch to execute when enhanced RDs are enabled, then code above
```
if (RD2_EnablementService.isRecurringDonations2Enabled) {
    query += ' LIMIT 0';
}
```
is not valid any longer.

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -531,68 +531,74 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
     /***
      * @description Verifies Recurring Donations are returned when enhanced Recurring Donations are not enabled
      */
-    @isTest 
-    private static void shouldReturnRDsWhenEnhancementIsNotEnabled() {  
-        List<npe03__Recurring_Donation__c> records = executeBatchStart();
+    @isTest
+    private static void shouldReturnRDsWhenEnhancementIsNotEnabled() {
+        List<npe03__Recurring_Donation__c> records = new List<npe03__Recurring_Donation__c>();
+        try {
```
RadSFDeveloper: The batch should execute since enhanced RDs are not enabled by default. Why do we need the `try/catch` block?

src/classes/RD_RecurringDonations_TEST2.cls
```diff
@@ -531,68 +531,74 @@ private with sharing class RD_RecurringDonations_TEST2 {
...
      * @description Verifies Recurring Donations are not returned when enhanced Recurring Donations are enabled
      */
-    @isTest 
-    private static void shouldNotReturnRDsWhenEnhancementIsEnabled() {  
+    @isTest
+    private static void shouldNotReturnRDsWhenEnhancementIsEnabled() {
         RD2_EnablementService_TEST.setRecurringDonations2Enabled();
 
-        List<npe03__Recurring_Donation__c> records = executeBatchStart();
+        List<npe03__Recurring_Donation__c> records = new List<npe03__Recurring_Donation__c>();
```
RadSFDeveloper: The test method name should verify the exception is thrown in this case now. We will not return zero records any more.

src/labels/CustomLabels.labels
```diff
@@ -1640,6 +1648,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>The total Amount of all Closed Won Opps is not equal to the RD Amount.</shortDescription>
         <value>The combined Amount of all related Closed Won and Pledged Opportunities must match the Amount of a Fixed Length Recurring Donation.</value>
     </labels>
+    <labels>
+        <fullName>RD_ErrorRD2BatchJobCannotBeRun</fullName>
```
RadSFDeveloper: Can we name custom labels to indicate requirement and/or replace RD1 with legacy and RD2 with enhance? For example, `_EnhancedBatchCannotBeRun` or `_EnhancedBatchRequiresEnablement`. Similarly, `_LegacyBatchCannotBeRun`.

Which prefix should these custom labels use? The enhanced RD batch custom label has `RD_` prefix while the legacy batch custom label has `RD2_` prefix.

src/classes/RD2_EnablementDelegate_CTRL.cls
```diff
@@ -141,10 +141,12 @@ public with sharing class RD2_EnablementDelegate_CTRL extends STG_Panel {
     @AuraEnabled
     public static void enableEnhancement() {
         try {
-            STG_Panel.stgService.stgRD.isRecurringDonations2Enabled__c = true;
+            STG_Panel.stgService.stgRD.IsRecurringDonations2Enabled__c = true;
 
             if (!Test.isRunningTest()) {
                 upsert STG_Panel.stgService.stgRD;
+                UTIL_MasterSchedulableHelper.setScheduledJobs();
```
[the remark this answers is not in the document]
force2b: Yes, but it's because the `RD2_EnablementDelegate_TEST` class runs in parallel. I could remove the `IsParallel=true` annotation from the test class to resolve the error, but there are already other test methods that validate the individual functionality of the scheduler and the telemetry call.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -34,10 +34,29 @@
...
+public inherited sharing class RD2_OpportunityEvaluationService {
 
     private npe03__Recurring_Donations_Settings__c rdSettings = UTIL_CustomSettingsFacade.getRecurringDonationsSettings();
 
+    /**
+    * @description Flag to allow customers to optionally disable Rollups before inserting/updating Opps
+    * to address issues with performance in larger organizatios
+    */
+    @TestVisible
+    private Boolean disableRollupsBeforeDml {
```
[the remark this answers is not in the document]
force2b: Good point. In a very near future WI (one that's actually not specific to RD2), we will introduce this new field on the RD Settings object. At that point, we could just address it in the code as `rdSettings.DisableRollupsInBatchJob__c`. That doesn't exist yet though, so I've separated into this getter method.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -100,21 +118,41 @@ public class RD2_OpportunityEvaluationService {
...
+
+            // TODO Schedule Visualizer
             // TODO Update the Future Installment Opportunity if needed
+            // TODO Handle Past Due Installment
+
         }
 
-        // TODO Recalculate NextDonationDate
-        // Date nextDonationDate = scheduleService.calculateNextDonationDate(rd);
+        if (disableRollupsBeforeDml) {
```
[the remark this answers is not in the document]
force2b: DML is supposed to be isolated to the end of the evaluate opportunities process, so there should not be a need to disable any earlier than the first DML.

src/classes/RD2_OpportunityEvaluationService.cls
force2b: It's forthcoming.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -162,7 +214,8 @@ public class RD2_OpportunityEvaluationService {
 
     /**
      * @description Populate the collection of Recurring Donation records, querying ALL fields,
-     * with the related Installment Opportunities and Active Schedule Records
+     * with the related Installment Opportunities and Active Schedule Records. Due to a KI (W-2503825)
```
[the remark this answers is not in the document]
force2b: This is a core bug.

src/classes/RD2_OpportunityEvaluation_BATCH.cls
```diff
@@ -0,0 +1,137 @@
...
+    * @description Build the query of all standard RD fields to use for the batch query
+    * @return SOQL
+    */
+    @TestVisible
+    private String getRDBatchQuery() {
+        Set<String> rdFields = new Set<String>{
+            'Id'
+        };
+
+        return new UTIL_Query()
```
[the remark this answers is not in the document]
force2b: All RD's are selected to ensure that an RD that might have been changed to Closed, but still has Open opps, is properly processed. That could happen in a bulk data load or bulk update scenario where triggers on the RD were disabled or unable to run due to bulk constraints (i.e., gov limits).

Sorting is not guaranteed in a batch query locator, so querying all Active Schedules, sorted by RecurringDonationId will not give us what we actually need to process the Opps. Not to mention, it will return more records than there are RD's, and the individual batches might cause schedule records for a single RD to be split across multiple batches.

src/classes/RD2_OpportunityEvaluation_BATCH.cls
```diff
@@ -0,0 +1,137 @@
...
+            .withFrom(npe03__Recurring_Donation__c.SObjectType)
+            .withSelectFields(rdFields)
+            .build();
+    }
+
+    /**
+    * @description Batch Finish Method
+    */
+    public void finish(Database.BatchableContext context) {
+        AsyncApexJob job = [SELECT CompletedDate FROM AsyncApexJob
```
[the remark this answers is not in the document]
force2b: Very good suggestion

src/classes/RD_RecurringDonations_BATCH.cls
force2b: Good point

src/labels/CustomLabels.labels
force2b: Good point. Label names updated
I do think the prefixes are correct. The RD2 prefixed error is used in the RD2 job, while the RD prefixed error is used in the RD job.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -148,6 +186,20 @@ public class RD2_OpportunityEvaluationService {
                 !rd.RecurringDonationSchedules__r.isEmpty());
     }
 
+    /**
+     * @description Determine if the specified Opportunity has Open Installment Opportunities
+     * @param rd
+     * @return True if there are open installments attached to the RD
+     */
+    private Boolean hasOpenInstallments(npe03__Recurring_Donation__c rd) {
```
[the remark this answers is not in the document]
RadSFDeveloper: Based on the [Handle Close RD Open Opportunities PR](https://github.com/SalesforceFoundation/Cumulus/blob/e44ddc5294638632d6dcb540cd0b993c50eb3eea/src/classes/RD2_OpportunityEvaluationService.cls#L141), this method  and lines 138-140 are not required. So we can remove the merge conflict in this PR or during the merge.

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
RadSFDeveloper: The var can be local variable in the `disableRollups()`.

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
RadSFDeveloper: The `if (disableRollupsBeforeDml) {` can be moved to the `disableRollups()` method.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -128,7 +164,7 @@ public class RD2_OpportunityEvaluationService {
      * Active Schedule records.
      */
     @TestVisible
-    private Boolean isCandidate(npe03__Recurring_Donation__c rd) {
+    private Boolean isEligibleForEval(npe03__Recurring_Donation__c rd) {
```
RadSFDeveloper: Based on [Handle Closed RD Open Opportunities PR](https://github.com/SalesforceFoundation/Cumulus/blob/e44ddc5294638632d6dcb540cd0b993c50eb3eea/src/classes/RD2_OpportunityEvaluationService.cls#L129), this method is removed, so due to renaming there might be merge conflict.

src/classes/RD_RecurringDonations_TEST2.cls
RadSFDeveloper: @force2b ^

src/package.xml
```diff
@@ -1476,13 +1478,15 @@
         <members>RD2_EnablementPrepTasks</members>
         <members>RD2_EnablementPrepTitle</members>
         <members>RD2_ErrorDataMigrationJobCannotBeRun</members>
+        <members>RD2_ErrorRD1BatchJobCannotBeRun</members>
```
RadSFDeveloper: Custom labels in package.xml are not updated.

src/classes/STG_PanelCustomizableRollup_CTRL.cls
```diff
@@ -105,7 +105,6 @@ public with sharing class STG_PanelCustomizableRollup_CTRL extends STG_Panel {
                 // there are existing rollups. just save the setting and reschedule jobs, don't deploy.
                 STG_Panel.stgService.stgCRLP.Customizable_Rollups_Enabled__c = true;
                 stgService.saveAll();
-                UTIL_MasterSchedulableHelper.setScheduledJobs();
```
RadSFDeveloper: Wondering why batch jobs scheduling is removed from NPSP Settings Customizable Rollup enablement page?

src/labels/CustomLabels.labels
```diff
@@ -1640,6 +1648,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
         <shortDescription>The total Amount of all Closed Won Opps is not equal to the RD Amount.</shortDescription>
         <value>The combined Amount of all related Closed Won and Pledged Opportunities must match the Amount of a Fixed Length Recurring Donation.</value>
     </labels>
+    <labels>
+        <fullName>RD_ErrorEnhancedRDBatchJobCannotBeRun</fullName>
```
RadSFDeveloper: A note about the prefix, the `RD_ErrorEnhancedRDBatchJobCannotBeRun` is called from `RD2_OpportunityEvaluation_BATCH`. Should prefix match?

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
force2b: True. I've been thinking about the concept of not adding placeholders for upcoming work. How about if I just remove this and add a simple `// TODO` comment to reflect that this will be needed in the near future?

src/classes/RD2_OpportunityEvaluationService.cls
force2b: Just to add to this, the current OpportunityService commits the new Opps in the service method instead of returning instances of the new Opps. Though it does work in its current state, I believe that if we want to be able to better manage DML (performance) in the batch (actually the whole EvaluateOpportunities logic), the service should act more like the schedule or NDD services that can operate without actually committing the records.  What do you think?

src/classes/RD2_OpportunityEvaluationService.cls
force2b: Great catch. Thank you!

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
RadSFDeveloper: This method is fine, I do not see any conflicts with future work. It would be great if we could localize changes, ie the `if` statement and the class variable related to the method pulled in, otherwise, I vote for this method and its call from `execute()` to stay. 
Other TODOs are ok as long as they will be addressed in future WIs.
The only merge conflict I see is work related to Close RD open Opportunities and the rename of `isCandidate()`.

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
force2b: You're right. I'll pull this out.

src/classes/STG_PanelCustomizableRollup_CTRL.cls
force2b: Oops. Thank you for catching this!

src/classes/RD_RecurringDonations_TEST2.cls
force2b: True. Not needed.

src/labels/CustomLabels.labels
force2b: Ugh. Yeah. I definitely got the labels/prefixed reversed.

src/classes/RD2_OpportunityEvaluationService.cls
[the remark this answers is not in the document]
force2b: I updated the method description to better clarify the change.

force2b:
Thanks for your review @RadSFDeveloper. All changes made.

rponti-sforg:

https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001C9tlQAC/view
https://salesforce.quip.com/2O0bARVu6IfV

Merging PR (once build is complete) and setting WI status to Closed.

PR #4792, opened 2019-11-13
W-036287: Gift Entry: Save and Dry Run Data Import Record in Batch Entry Mode

kenlewis-sfdo:
This is the first PR for the Batch Entry mode of the new form.  These changes allow the form to be loaded in "batch mode" and save submitted rows to a table below the form, comparable to the table in the current BGE wizard.  The data import record is saved and dry run is performed, with dry run results being returned and showing in the Status column.  

The Status and Donor columns are displayed by default, the other columns are generated from the fields in the sections in the template being used to generate the form.  

Error handling is displayed in the Errors column but could be relocated if necessary.

After deploying, testing can be done by accessing the placeholder Aura component at '<yourInstanceURL>/lightning/cmp/c__geFormRendererPlaceholder'.

**While testing, a Batch Id and an Account Id should be manually placed into the two temporary input fields at the top of the form.**  These simulate loading the form from a Batch, and either having lookups or loading from an Account record (or eventually Contact) in the future.  Without a Batch Id or Account Id the form currently stalls and will not work.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/ApexDocContent/RecurringDonations.htm
+++ b/ApexDocContent/RecurringDonations.htm
@@ -1,7 +1,7 @@
 <html>
     <body>
         <h2>Recurring Donations 1.0 Related Classes</h2>
-        <p>RD1 in NPSP supports two distinct types of Recurring Donations.</p>
+        <p>Legacy Recurring Donations in NPSP support two distinct types of Recurring Donations.</p>
         <ul>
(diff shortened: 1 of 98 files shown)


kenlewis-sfdo:
Going to replace this PR with a new one that has a few additional changes.

PR #4795, opened 2019-11-14
Custom setting to keep track of the Default SGE/BGE Template

KietTruongSFDO:
# Critical Changes

# Changes
We have added a new custom setting to keep track of the default SGE/BGE template. 

# Issues Closed

# New Metadata

# Deleted Metadata


src/objects/Gift_Entry_Settings__c.object
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ptewson-sfdo: Did we talk as team about whether we wanted a new settings object to store these settings? I don't object necessarily but we also have Data Import Settings already defining some related config for more of the BDI side of things - I just want to make sure we are being strategic, and intentional.

vmartinov-sforg: @ptewson-sfdo We did talk about creating a separate custom setting for GE, as we may add more settings in the future.

PR #4800, opened 2019-11-15
[RD] Fix Issues with StartDate When a Schedule is Modified

npsp-reedestockton:
When Enhanced Recurring Donations is enabled, ensure that Start Date of updated schedules defaults to current date.
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


npsp-reedestockton:
**Lurch: attach W-037358

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HoN0QAK" target="_blank">W-037358</a>

npsp-reedestockton:
@RadSFDeveloper @force2b Ready for review
@sam-knox Could you please review PR description

src/classes/RD2_ScheduleService.cls
```diff
@@ -187,14 +192,17 @@ public without sharing class RD2_ScheduleService {
         Date nextDate;
         Date previousDate;
 
-        for (RecurringDonationSchedule__c schedule : rdSchedules) {
-            nextDate = getNextDonationDateGreaterThanToday(schedule);
+        if (rdSchedules != null) {
```
force2b: For readability, I suggest:
```suggestion
        if (rdSchedules == null) {
            return nextDate;
        }
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,38 +260,101 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
-        Map<Id, npe03__Recurring_Donation__c> filteredRecurringDonations = getRecurringDonationsNeedingScheduleUpdates(rds, oldRds);
+    private Map<Id, List<RecurringDonationSchedule__c>> getExistingSchedules(List<npe03__Recurring_Donation__c> rds) {
 
         Map<Id, List<RecurringDonationSchedule__c>> oldSchedules = new Map<Id, List<RecurringDonationSchedule__c>>();
+
         for (RecurringDonationSchedule__c oldSchedule : [
-                SELECT RecurringDonation__c, EndDate__c FROM RecurringDonationSchedule__c
-                WHERE RecurringDonation__c IN :filteredRecurringDonations.keySet()
+                SELECT RecurringDonation__c, StartDate__c, EndDate__c, InstallmentFrequency__c, InstallmentPeriod__c, DayOfMonth__c FROM RecurringDonationSchedule__c
+                WHERE RecurringDonation__c IN :rds
```
force2b:
```suggestion
                for (RecurringDonationSchedule__c oldSchedule : [
                                         SELECT RecurringDonation__c, 
                                             StartDate__c, 
                                             EndDate__c, 
                                             InstallmentFrequency__c, 
                                             InstallmentPeriod__c, 
                                             DayOfMonth__c 
                                         FROM RecurringDonationSchedule__c 
                                         WHERE RecurringDonation__c IN :rds
                                         AND Active__c = True
                ]) {
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -193,7 +193,7 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
             buildNewSchedules(rds);
 
         } else if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
-            setNextDonationDateBeforeUpdate(rds);
+            setScheduleDependentRecurringDonationFields(rds, oldRds);
```
force2b:
```suggestion
            setScheduleDependentFields(rds, oldRds);
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,38 +260,101 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
             newRd.Status__c != oldRd.Status__c);
     }
+
     /***
     * @description Handles setting next donation date for updated recurring donations
     * @param rds Trigger.new records
     * @param oldRds Trigger.old records
     */
-    private void setNextDonationDateBeforeUpdate(List<npe03__Recurring_Donation__c> rds) {
+    private void setScheduleDependentRecurringDonationFields(List<npe03__Recurring_Donation__c> rds,List<npe03__Recurring_Donation__c> oldRds) {
```
force2b:
```suggestion
    private void setScheduleDependentFields(List<npe03__Recurring_Donation__c> rds, List<npe03__Recurring_Donation__c> oldRds) {
```

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,38 +260,101 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+    }
+
+    /***
+    * @description Adds recurring donation schedules to map by recurring donation Id
+    * @param sobj SObject
+    * @param newSchedulesByRdId Map<Id, List<RecurringDonationSchedule__c>>
+    */
+    private void addUpdatedSchedules(SObject sobj, Map<Id, List<RecurringDonationSchedule__c>> newSchedulesByRdId) {
+
+        if (sobj.getSObjectType() == RecurringDonationSchedule__c.SObjectType) {
```
force2b: Since this is a private method in this class, can this ever be any other type than a `RecurringDonationSchedule__c` record, especially if the first parameter is of `RecurringDonationSchedule__c` type instead of SObject?

I think you can do that by changing the calling methods to 
`addUpdatedSchedules((RecurringDonationSchedule__c)sobj, newSchedulesByRdId); `

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,38 +260,101 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
             newRd.Day_of_Month__c != oldRd.Day_of_Month__c ||
             newRd.Status__c != oldRd.Status__c);
     }
+
     /***
     * @description Handles setting next donation date for updated recurring donations
```
force2b: Just so that I'm sure, the intent of this method is the following:

- Perform a full analysis and rebuild of the Schedule records for a Recurring Donation
- Do not commit the schedule records to the database
- Use the active schedules for the RD to calculate the Next Donation Date and store onto the RD record (in a Before Update context, thus avoiding the extra DML against the object).


If correct, I suggest updating the method comment to provide a bit more background on the why for this method.

force2b: Though if that is accurate, shouldn't this logic be roughly the same as the `synchronizeSchedules` method, maybe less the DML operations?

npsp-reedestockton: I've refactored the code so that the commonalities between the two methods are abstracted out into separate methods that can be called from each - getExistingSchedules() and getRecurringDonationsNeedingScheduleUpdates().  There is a key difference, though.  The before update version unconditionally sets next donation date, so it needs to look at all updates, not just those that qualify for schedule updates.

src/classes/RD2_RecurringDonations_TDTM.cls
npsp-reedestockton: Right now it is true that the only type of sobject this can be is a `RecurringDonationSchedule__c` record.  I was a bit reluctant to lock that conclusion into code, but that degree of caution may be overkill.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: I'm in on this.  Old habits die hard.

src/classes/RD2_RecurringDonations_TDTM.cls
npsp-reedestockton: It _may_ be true that the next donation date can't ever change unless there is an update to the RD.  If so, it would be true that the methods should be identical.  I'm very reluctant to make that assumption, though.  If it isn't 100% true both now and in the future, that would be a very difficult issue to track down.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,38 +260,103 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
-    * @description Handles setting next donation date for updated recurring donations
+    * @description Performs a full analysis and rebuild of the Schedule records for a Recurring Donation,
+    * but does not commit the schedule records to the database.  Uses the active schedules for the RD
+    * to calculate the Next Donation Date and store onto the RD record (in a Before Update context,
+    * thus avoiding the extra DML against the object).
     * @param rds Trigger.new records
     * @param oldRds Trigger.old records
     */
-    private void setNextDonationDateBeforeUpdate(List<npe03__Recurring_Donation__c> rds) {
+    private void setScheduleDependentFields(List<npe03__Recurring_Donation__c> rds,List<npe03__Recurring_Donation__c> oldRds) {
```
force2b: What would you think about moving the setScheduleDependentFields, getExistingSchedules and addUpdatedSchedules methods into the ScheduleService class and just have a method in this class for `calculateNextDonationDateBeforeUpdate` that delegates to the service method? You could also consider slightly updating the existing `calculateNextDonationDates` method to support both insert & update. Either way is probably fine though, whatever you think makes the most sense.

Overall, I'm looking to keep the logic to handle the schedules and NDD in the service class and have the TDTM class delegate out as needed.

npsp-reedestockton: I debated about these.  getExistingSchedules() would become the only method in the service that does dml.  Are you okay with that?  If so, I'm fine to relocate them.

force2b: getExistingSchedules does a query, as opposed to DML, correct? Either way though I think it's ok.

npsp-reedestockton: Right you are.  Query only.  Okay, I'll move these 3 methods to the service class.  Thanks!

npsp-reedestockton: Looking at this a little more closely, I'm not sure I understand the logic of keeping synchronizeSchedulesAfterRDUpdate() in the trigger while moving setScheduleDependentFields().  Could you help me out with that?

force2b: The `synchronizeSchedulesAfterRDUpdate` method delegates **all** processing to the service class and only collects sobject instances to add to the DMLWrapper. setScheduleDependentFields feels like it does more than that, especially with the other two methods. However, if you think the method is solid enough and doesn't having anything too business-logic specific we can keep it in this class. 

I was thinking of looking something like:
```
private void calculateNextDonationDateBeforeUpdate() {
    Map<Id, List<RecurringDonationSchedule__c>> schedulesByRdId = scheduleService.getUpdatedSchedules(rds, oldRds)
        rd.npe03__Next_Payment_Date__c = scheduleService.getNextDonationDate(schedulesByRdId.get(rd.Id));
    }
```

Basically the service is class is called to calculate the schedule records and those schedule records are used to recalc the NDD. I'm not tied to the method names, but the concept is that logic to get the schedules is in the schedules service.

npsp-reedestockton: Okay.  I buy that.  Integrating both existing schedules and new schedules is more like business logic.  I'm in.  I'll finish up and commit.

npsp-reedestockton: Also moving getRecurringDonationsNeedingScheduleUpdates().  Don't want to call from schedule service to the trigger.

src/classes/RD2_ScheduleService.cls
```diff
@@ -319,4 +319,120 @@ public without sharing class RD2_ScheduleService {
...
+            newSchedulesByRdId.get(rd.RecurringDonation__c).add(rd);
+        }
+    }
+
+    /***
+    * @description Filters recurring donations to determine which need schedule updates
+    * @param rds Trigger.new records
+    * @param oldRds Trigger.old records
+    */
+    public Map<Id, npe03__Recurring_Donation__c> getRecurringDonationsNeedingScheduleUpdates(
```
force2b: This method is really just checking for changes to the fields, which is more related to the trigger. Is it feasible to have this method in the TDTM class or is there something unique to what it does that it belongs in the service class?

npsp-reedestockton: It gets called from the methods we just moved.  Do you really want the service to make a call to the trigger method?

npsp-reedestockton: Possibilities...  Keep this in service or move setScheduleDependentFields() back to trigger.

npsp-reedestockton: my vote is to move setScheduleDependentFields(), but leave addUpdatedSchedules() and getExistingSchedules() in the service.

npsp-reedestockton: I've pushed changes that reflect the way I think it should be.  There is no perfect solution here, but I think this comes the closed to what you are trying to achieve without a bunch of gymnastics.

force2b: I see. How about this?

In the TDTM class:
```
private void calculateNextDonationDatesBeforeUpdate(rds, oldRds) {
    Map<Id, npe03__Recurring_Donation__c> filteredRecurringDonations = getRecurringDonationsNeedingScheduleUpdates(rds, oldRds, TDTM_Runnable.Action.BeforeUpdate);

    Map<Id, List<RecurringDonationSchedule__c>> schedulesByRdId = scheduleService.getSchedules(filteredRecurringDonations)

        if (schedulesById.containsKey(rd.Id) {
            rd.npe03__Next_Payment_Date__c = getNextDonationDate(schedulesByRdId.get(rd.Id));
        }
    }

}
```

In the ScheduleService class:
```
// Mostly the method that was setScheduleDependentFields without the bits that abstracted into the above
public Map<Id, List<RecurringDonationSchedule__c>> buildSchedules(rds) {

}
```

npsp-reedestockton:
Major changes to the ordering of methods - moved several methods back to ScheduleService and left very few in the trigger.  

Used the basic structure of your suggestion, but needed to replace getUpdateSchedules with getAllRecurringDonationSchedules (renamed from setScheduleDependentFields).

```
    private void calculateNextDonationDateBeforeUpdate(List<SObject> rds, List<SObject> oldRds) {
        Map<Id, List<RecurringDonationSchedule__c>> schedulesByRdId = scheduleService.getAllRecurringDonationSchedules(rds, oldRds);
        for (npe03__Recurring_Donation__c rd : (List<npe03__Recurring_Donation__c>)rds) {
            rd.npe03__Next_Payment_Date__c = scheduleService.getNextDonationDate(schedulesByRdId.get(rd.Id));
        }
    }

```

npsp-reedestockton:
@force2b Ready for re-review.

Julian88Tex:
Testing has passed for WI: https://foundation.lightning.force.com/lightning/_classic/%2Fa2x1E000001HoN0QAK

PR #4802, opened 2019-11-15
RD2: Handle Installment Opportunities when an RD is closed

RadSFDeveloper:
Handle Open Opportunity attached to a Recurring Donation when it is closed so that users do not continue to expect donations from the donor.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-037452

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HoUqQAK" target="_blank">W-037452</a>

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -160,14 +177,54 @@ public class RD2_OpportunityEvaluationService {
...
+    private void handleClosedRecurringDonationOpps(List<Opportunity> opps) {
+        if (opps.isEmpty()) {
+            return;
+        }
+
+        if (rdSettings.npe03__Open_Opportunity_Behavior__c == CloseActions.Delete_Open_Opportunities.name()) {
+            delete opps;
+
+        } else if (rdSettings.npe03__Open_Opportunity_Behavior__c == CloseActions.Mark_Opportunities_Closed_Lost.name()) {
+            for (Opportunity opp : opps) {
```
RadSFDeveloper: @force2b , should we create new SObject `new Opportunity(Id = opp.Id, StageName = System.Label.npe03.RecurringDonationClosedLostOpportunityStage);` for the update and not use other fields in the update DML?

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -86,66 +95,74 @@ public class RD2_OpportunityEvaluationService {
...
-            }
 
             if (isNewRecurringDonation(rd)) {
-                rdsWithCloseDates.put(rd, getInstallmentOppCloseDates(rd) );
+                newRDsWithCloseDates.put(rd, getInstallmentOppCloseDates(rd) );
                 continue;
             }
-            // TODO Update the Future Installment Opportunity if needed
+            
+            if (rd.Status__c == RD2_Constants.STATUS_CLOSED) {
```
RadSFDeveloper: @force2b , when RD data is migrated and `Status__c` updated to Closed for specific RDs, the open Opportunities  will be deleted/closed/unchanged based on the RD settings. If our requirements are to leave Opportunities for migrated RD currently, should we set RD settings to no action during data migration?

RadSFDeveloper:
> Really nice (again). I'm impressed that you were able to refactor the comment blocks in the UTIL classes so quickly. As you suspected, I do think that the new CloseActions enum should only apply to RD2 code. Even though the enum is effectively a duplicate of what exists in the RD_RecurringDonations class, it's probably safer to leave the legacy code untouched. Plus having a separate enum would give us the ability to expand the options at some point (if ever needed) without impacting the legacy RD logic.

@force2b , changes for legacy RD code have been reverted for legacy code. However, there is common code in [NPSP Settings RD settings](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/STG_PanelRD_CTRL.cls#L121) and [custom settings facade](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/UTIL_CustomSettingsFacade.cls#L590). I left it to use legacy RD constants. If any change is to be made, we can apply it in this PR or resolve in [W-035815](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001CAKcQAO/view). Since values are the same for both legacy and enhanced RDs, should we still have usage of different constants based on if RD2 is enabled?

force2b:
Thanks @RadSFDeveloper. I'm advocating for different constants to allow us to expand behavior at some future point without requiring any changes to the legacy code. Still, as you've noted there is some shared code such as in NPSP Settings RD Controller and the CustomSettingsFacade class. Those are probably fine for now. If we do decide to change or expand the RD2 specific behavior there would only be these two places to modify change where the Enum values come from.

RadSFDeveloper:
@force2b , can you please review merge conflict resolution and clean up applied in commits today? Thank you!

RadSFDeveloper:
@Julian88Tex , the PR is ready for testing. Thank you!

Julian88Tex:
Testing has passed for WI: https://foundation.lightning.force.com/lightning/_classic/%2Fa2x1E000001HoUqQAK

PR #4803, opened 2019-11-18
[Robot] Changes to increase time waits to fix the failing robot tests

rjanjanam:
# Critical Changes
Changes to NPSP.Py
# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/NPSP.py
```diff
@@ -858,7 +858,7 @@ def wait_for_batch_to_complete(self, path, *args, **kwargs):
         locator = self.get_npsp_locator(path,*args, **kwargs)
         while True:
             i += 1
-            if i > 14:
+            if i > 20:
```
boakley: Make sure you update the keyword documentation to match the implementation. The documentation says it checks every 15 seconds. for 3.5 minutes.  The code now waits every 20 seconds for over 6.5 minutes. It can actually wait way longer than that, since it first asks seleniun to wait, and then it adds an extra 20 seconds on each loop.

robot/Cumulus/resources/NPSP.py
```diff
@@ -868,7 +868,7 @@ def wait_for_batch_to_complete(self, path, *args, **kwargs):
                     self.selenium.wait_until_element_is_visible(locator)
                     break
                 except Exception:
-                    time.sleep(15)    
+                    time.sleep(20)
```
boakley: This results in waiting for more than six minutes. Is that really what you intend to do?

rjanjanam:
No more needed

PR #4840, opened 2019-11-22
NPSP process HH naming as asyc when operating in bulk

### LINKED ISSUE #4535: Household Naming Causes Increased CPU Consumption During Contact Insertion
Inserting 200 Contacts in a clean scratch org with only NPSP installed consumes at least ~7500 milliseconds of CPU time. 

The debug log shows an extra Account update occurring, which is triggered from a previous Account update:
Screen Shot 2019-08-21 at 4 27 27 PM
Screen Shot 2019-08-21 at 4 27 49 PM

Household Naming is configured to run asynchronously when there is more than one Account within context and we do kick off an async job... but the Account update that occurs when we add the Primary Contact is setting off a synchronous Household Naming update as well, in ACCT_Accounts_TDTM... and all Trigger Handlers fire again. Removing the synchronous update on Account Update removed ~3000 milliseconds of CPU time. 
This also resulted in a cleaner log:
Screen Shot 2019-08-25 at 4 00 18 PM

### More Info:
ACCT_Accounts_TDTM calls HH_HouseholdNaming.handleHHNameUpdates(). Within that class, we have:

```
 if (!hhlist.isEmpty()){
    HH_HouseholdNaming hn = new HH_HouseholdNaming();
    hn.UpdateNames(hhlist);
}

```
This is a real-time update:

```
if (!hhupdatelist.isEmpty()) {
   // all or nothing, so errors not lost
   UTIL_DMLService.updateRecords(hhupdatelist);
}

```

Contrast this with ACCT_IndividualAccounts_TDTM, which determines sync vs. async processing via this code:

```
if (accountIds.size() == 1 || System.isFuture() || System.isBatch() || System.isQueueable()) {
    HH_HouseholdNaming hhName = new HH_HouseholdNaming();
    hhName.UpdateNames(accountIds);
} else {
    HH_HouseholdNaming.FutureUpdateNames(accountIds);
}
```
-------

More details from my Investigation will be linked to the Work Item.

ptewson-sfdo:
Updates the processing of HH naming when handling bulk cases to an asynchronous call. This addresses issues with performance on inserting bulk contacts that were potentially affected by attempting to synchronously update HH name.

It also updates a unit test to deploy Test.startTest() and Test.stopTest() in order to process the asynchronous update from above ahead of evaluating the test results. 

# Critical Changes

# Changes

# Issues Closed
Fixes #4535
# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/HH_HouseholdNaming.cls
+++ b/src/classes/HH_HouseholdNaming.cls
@@ -435,7 +435,7 @@ public without sharing class HH_HouseholdNaming {
                 if (hhlist.size() == 1 || System.isFuture() || System.isBatch() || System.isQueueable()) {
                     HH_HouseholdNaming hn = new HH_HouseholdNaming();
                     hn.UpdateNames(hhlist);
-                } else {
+                } else (!hhlist.isEmpty()) {
                     HH_HouseholdNaming.FutureUpdateNames(hhlist);
                 }
             }
--- a/src/classes/HH_HouseholdNaming.cls
+++ b/src/classes/HH_HouseholdNaming.cls
@@ -435,7 +435,7 @@ public without sharing class HH_HouseholdNaming {
                 if (hhlist.size() == 1 || System.isFuture() || System.isBatch() || System.isQueueable()) {
                     HH_HouseholdNaming hn = new HH_HouseholdNaming();
                     hn.UpdateNames(hhlist);
-                } else (!hhlist.isEmpty()) {
+                } else if (!hhlist.isEmpty()) {
                     HH_HouseholdNaming.FutureUpdateNames(hhlist);
                 }
             }
--- a/src/classes/HH_HouseholdNaming.cls
+++ b/src/classes/HH_HouseholdNaming.cls
@@ -368,7 +368,7 @@ public without sharing class HH_HouseholdNaming {
             
             if (hs != null && hs.npo02__Advanced_Household_Naming__c == true) {
                 // AfterUpdate data
-                list<id> hhlist = new list<id>();
+                list<id> hhList = new list<id>();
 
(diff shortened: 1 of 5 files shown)


ptewson-sfdo:
**lurch: attach W-037776

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hp9AQAS" target="_blank">W-037776</a>

melissabarber:
Release note description is ready to go!

src/classes/HH_HouseholdNaming.cls
```diff
@@ -432,13 +432,13 @@ public without sharing class HH_HouseholdNaming {
                             hhlist.add(h.id);
                     }
                 }
-                if (!hhlist.isEmpty()){
+                if (hhlist.size() == 1 || System.isFuture() || System.isBatch() || System.isQueueable()) {
                     HH_HouseholdNaming hn = new HH_HouseholdNaming();
                     hn.UpdateNames(hhlist);
+                } else if (!hhlist.isEmpty()) {
+                    HH_HouseholdNaming.FutureUpdateNames(hhlist);
```
kenlewis-sfdo: It looks like there is an existing static method in `ACCT_IndividualAccounts_TDTM` called `renameHHAccounts` that does this also: https://github.com/SalesforceFoundation/Cumulus/blob/c8702e2ae8c4a44304c5f692f95e168a7b5e3486/src/classes/ACCT_IndividualAccounts_TDTM.cls#L928-L939 

What do you think about moving that method to this class and calling it here to avoid duplicating the logic?  I think this class might be a better home for that `renameHHAccounts` method but curious to hear your thoughts!

PR #4849, opened 2019-11-26
Gift Entry: enforcing data/batch import object-level required fields

psadie-sfdo:
# Critical Changes

# Changes
- Checks that the Template Name field has a value.
- Checks field required-ness for DataImport__c and DataImportBatch__c objects in Template Builder.
- Change GE_Templates and GE_Template_Builder tab visibility for System Admins to 'Default On'.
# Issues Closed

# New Metadata
- Lightning Web Components
  - utilPageLevelMessage
# Deleted Metadata


src/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -92,6 +142,75 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
...
+    * additionally required fields and exluded fields. List is used by the sidebar
+    * for the lightning-input checkboxes.
+    */
+    collectBatchHeaderFields() {
+        this.batchFields = mutable(this.diBatchInfo.fields);
+
+        Object.getOwnPropertyNames(this.batchFields).forEach((key) => {
+            let field = this.batchFields[key];
+
+            if (EXCLUDED_BATCH_HEADER_FIELDS.includes(field.apiName.replace(this.namespace, ''))) {
```
kenlewis-sfdo: What do you think about importing the fields so that we don't need to rely on manually replacing the namespace here?

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
```diff
@@ -1,6 +1,51 @@
...
+    'Name'
+];
+Object.freeze(ADDITIONAL_REQUIRED_BATCH_HEADER_FIELDS);
+
+// We've opted to exclude the following batch fields related to
+// matching logic as we're removing the matching options page
+// from this flow.
+// We're considering putting matching options in a 'global
+// batch settings' area. Potentially in NPSP Settings.
+const EXCLUDED_BATCH_HEADER_FIELDS = [
```
kenlewis-sfdo: See comment above re: importing these so we can reference their `fieldApiName` properties and avoid explicit handling of namespaces.

src/lwc/geTemplateBuilder/geTemplateBuilder.js
psadie-sfdo: So we are actually importing the fields through `getObjectInfo` here. The reason I'm removing the namespace here is because we have the two hardcoded list of excluded and additionally required fields. Maybe a better way to handle this would be to handle namespacing on the hardcoded lists before they're used here?

src/lwc/utilPageLevelMessage/utilPageLevelMessage.js
```diff
@@ -0,0 +1,74 @@
...
+    @api subtitle;
+    @api errors;
+    @api iconName;
+    @api iconSize;
+    @api iconVariant;
+    @api iconDescription;
+    @api variant;
+
+    get titleText() {
+        return this.title ? this.title : 'Review the errors on this page.';
```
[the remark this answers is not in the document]
psadie-sfdo: Definitely! Am thinking about including this in the custom labels specific WI that's also in play this sprint though.

src/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -164,7 +165,7 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
                 field.isPicklist = false;
             }
 
-            if (ADDITIONAL_REQUIRED_BATCH_HEADER_FIELDS.includes(field.apiName.replace(this.namespace, ''))) {
+            if (ADDITIONAL_REQUIRED_BATCH_HEADER_FIELDS.includes(field.apiName)) {
```
kenlewis-sfdo: 👍🏽

src/lwc/utilPageLevelMessage/utilPageLevelMessage.js
```diff
@@ -10,7 +10,7 @@ const THEME_ERROR = 'slds-theme_error';
 const THEME_INFO = 'slds-theme_info';
 const TEXT_ERROR = 'slds-text-color_error';
 
-export default class utilPageMessage extends LightningElement {
+export default class utilPageLevelMessage extends LightningElement {
```
kenlewis-sfdo: 👍🏽

src/lwc/utilPageLevelMessage/utilPageLevelMessage.js-meta.xml
```diff
@@ -2,4 +2,5 @@
 <LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
     <apiVersion>47.0</apiVersion>
     <isExposed>false</isExposed>
+    <description>A page level message to provide feedback to a user following an action, such as after a record save and a required field is missing.</description>
```
kenlewis-sfdo: Looks great!

src/classes/BDI_FieldMapping.cls
```diff
@@ -122,7 +122,7 @@ public virtual with sharing class BDI_FieldMapping {
             }
         }
 
-        // TODO: Complete when Template-level validation work item is in play
-        this.Is_Required = (this.Required == 'Yes' || !sourceFieldDescribe.isNillable() || !targetFieldDescribe.isNillable()) ? true : false;
+        this.Is_Source_Field_Required = !sourceFieldDescribe.isNillable() ? true : false;
+        this.Is_Required = !sourceFieldDescribe.isNillable() ? true : false;
```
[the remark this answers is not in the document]
psadie-sfdo: Nope! I was thinking it may be useful in the future when this ticket was still in flux. Blasted it for now.

https://github.com/SalesforceFoundation/Cumulus/pull/4849/commits/2edf43a8d63e69bd6940dbd9b57ae8194c51d606

src/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -92,6 +143,75 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
...
+                field.isPicklist = true;
+            } else {
+                field.isPicklist = false;
+            }
+
+            if (ADDITIONAL_REQUIRED_BATCH_HEADER_FIELDS.includes(field.apiName)) {
+                field.required = true;
+            }
+
+            if (field.createable && field.updateable) {
```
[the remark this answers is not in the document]
psadie-sfdo: I moved the check up a bit further in the block so we immediately return if any of the following are true: field is explicitly excluded, a new template and field is not createable, or an existing template and field is not createable/editable.

https://github.com/SalesforceFoundation/Cumulus/pull/4849/commits/2edf43a8d63e69bd6940dbd9b57ae8194c51d606

src/classes/BDI_FieldMapping.cls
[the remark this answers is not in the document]
kenlewis-sfdo: 👍🏽

PR #4855, opened 2019-11-30
RD2: Data migration - void open Opps

RadSFDeveloper:
Remove the extra forecasted Opportunities during data migration in order to save on storage and see less visual clutter on Donor records.

# Critical Changes

# Changes

# Issues Closed

# New Metadata
Custom Labels:
- RD2_DataMigrationInvalidCloseAction

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-036853

RadSFDeveloper:
@force2b , it might be better to review PR after PR #4831  is merged since I needed to merge its changes before extending the data migration functionality.

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CC9XQAW" target="_blank">W-036853</a>

src/classes/RD2_DataMigration_BATCH.cls
```diff
@@ -114,40 +167,125 @@ public class RD2_DataMigration_BATCH implements Database.Batchable<SObject>, Sch
...
+            }
+        }
+
+        if (opps.isEmpty()) {
+            return;
+        }
+
+        List<Object> dmlResults = new List<Object>();
+        if (rdSettings.npe03__Open_Opportunity_Behavior__c == RD2_Constants.CloseActions.Delete_Open_Opportunities.name()) {
+            dmlResults = Database.delete(opps, false);
```
RadSFDeveloper: @force2b , if an Open Opp deletion or update to Closed Lost fails, Recurring Donation will still be migrated and an error log will be created for the failed Opp. Is this expected behaviour?
FYI @Julian88Tex

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -196,26 +199,28 @@ public inherited sharing class RD2_OpportunityEvaluationService {
...
      * @param rd Recurring donation containing list of all related Opportunities
-     * @return List<Opportunity> Open Opportunities
+     * @param openOppsToVoid Open Opportunities to void from other Recurring Donations
+     * @return void
      */
-    private List<Opportunity> extractOpenInstallmentOpps(npe03__Recurring_Donation__c rd) {
-        List<Opportunity> opps = new List<Opportunity>();
+    private void handleClosedRecurringDonation(npe03__Recurring_Donation__c rd, List<Opportunity> openOppsToVoid) {
+
+        if (rd.Status__c != RD2_Constants.STATUS_CLOSED || isCloseActionOverride) {
```
RadSFDeveloper: Even though behaviour is not directly related to this PR, wanted to verify functionality. When a user creates an open Opp for a Closed RD, it will be voided (deleted or set to Closed Lost if settings indicate so) by this service. Is this expected behaviour, ie user cannot create an open Opp for a Closed RD?

FYI @Julian88Tex

RadSFDeveloper:
@force2b , the PR is ready for the review.

src/classes/RD2_DataMigration_BATCH.cls
RadSFDeveloper: @force2b , as discussed, I will make the change to rollback the whole batch iteration if Opp deletion/update fails.

RadSFDeveloper: @force2b , batch iteration rollback change has been applied.

force2b: Thanks @RadSFDeveloper. Ideally once we move to a _successful-dry-run-required_ enablement process this won't matter as much.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -83,6 +83,13 @@ public inherited sharing class RD2_OpportunityEvaluationService {
         } set;
     }
 
+    /**
+     * @description Indicates if Recurring Donation Settings close action on Open Opportunities 
+     * is overriden when Recurring Donation is Closed
+     */
+    public static Boolean isCloseActionOverride = false;
```
force2b: What would you think of specifying what the override is for the var name? Maybe `isDoNotVoidOpenOpps` or `overrideVoidOfOpenOpps`?

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -264,5 +293,54 @@ public class RD2_DataMigrationMapper {
         rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
     }
 
+    /**
+     * @description Extracts open installment Opportunities with Close Date >= today and return all but the first one.
+     * These Opportunities will become void (either closed lost or deleted).
+     * @return List<Opportunity> Open Opportunities
+     */
+    private void validateOpenOpportunities() {
```
force2b: The method name implies that it's validating something about these open opportunities, but if I'm following the program flow correctly it's actually reviewing the open opportunities that are eligible to be _voided_ (deleted or closed) and then if there are validating that the global RD Settings is valid. So if anything, it's only validating that the setting is valid. I wonder if that check should just done once at the start of the batch job rather than every time since the setting doesn't change for each recurring donation.

RadSFDeveloper: The RD settings close action is applicable only for RDs with "Open Ended Status" set as "Open". If the RD is not Open or  it is Open, but the RD has no open Opp to void, then RD settings close action is not applicable. Thus, no error is generated if the close action is No_Action for an RD record that has no open Opp to void. For example: [shouldConvertRDWhenItHasOneOpenFutureOppAndCloseActionIsNoAction()](https://github.com/SalesforceFoundation/Cumulus/blob/38f783f0bee989ab52849e1fd7b1f47a516e7321/src/classes/RD2_DataMigration_TEST.cls#L1139).

So we can either validate settings on a need basis as currently done, or the requirement is to stop the whole data migration if the RD settings close action is No_Action?

RadSFDeveloper: @force2b , @Julian88Tex 
When data migration is started and RD custom settings close action is No_Action, an error will be displayed to the user and data migration will fail (note: the batch Extended Status will contain the error and no Error record will be created). 
Other suggested changes have been applied as well. The PR is ready for review.

RadSFDeveloper:
@Julian88Tex , data migration of open Opps on RDs is ready for testing.

Julian88Tex:
Manual tests have passed. Just waiting for builds to complete.

PR #4865, opened 2019-12-04
[RD] Schedule Visualization

npsp-reedestockton:
We added a Recurring Donations Schedule component for use on the Enhanced Recurring Donations record page.  The Recurring Donations Schedule component displays a configurable number of upcoming installment Opportunities for the Recurring Donation.

# Critical Changes

# Changes

# Issues Closed

# New Metadata
Apex Classes:
- RD2_VisualizeScheduleController
- RD2_VisualizeScheduleControllerTest

Custom Labels:
- RD2_ScheduleVisualizerColumnDate
- RD2_ScheduleVisualizerErrorEnhancedRDNot
- RD2_ScheduleVisualizerErrorInvalidUsage
- RD2_ScheduleVisualizerMessageNoSchedule
- RD2_ScheduleVisualizerTitle

Lightning Component Bundle:
- rdScheduleVisualizer

# Deleted Metadata


src/lwc/rdScheduleVisualizer/rdScheduleVisualizer.js
```diff
@@ -0,0 +1,143 @@
+// TODO Implement Custom Label as defined below
+// TODO (Post-MVP) Implement a user definable setting (in the LWC) to define the columns to show
```
force2b: I think these two TODO's can be removed now

src/lwc/rdScheduleVisualizer/rdScheduleVisualizer.js
```diff
@@ -0,0 +1,143 @@
...
+    @track fldAmount;
+
+    /*******************************************************************************
+     * @description Whenever the related RD record is updated, this is called to force
+     * a refresh of the table and component.
+     */
+    @wire(getRecord, { recordId: '$recordId',
+        fields: [FIELD_RD_AMOUNT, FIELD_RD_DAYOFMONTH, FIELD_RD_FREQUENCY, FIELD_RD_PERIOD, FIELD_RD_STARTDATE, FIELD_RD_PAYMENT_METHOD] })
+    wireRecordChange() {
+        console.log('The record has changed: ', this.recordId);
```
force2b: the `console.log` statements in this file can be removed now too.

npsp-reedestockton:
***lurch: attach W-037685*

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hp0SQAS" target="_blank">W-037685</a>

npsp-reedestockton:
@force2b @RadSFDeveloper Ready for review.  Thanks!
@sam-knox Please have a look at labels and the PR description.  Thanks!

npsp-reedestockton:
@sam-knox Label changes have been made.

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,128 @@ public without sharing class RD2_ScheduleService {
...
+        List<Installment> vizSchedules = new List<Installment>();
+        currentDate = visualizationStartDate;
+        Integer dateCount = 0;
+        List <RecurringDonationSchedule__c> filteredSchedules;
+
+        List <RecurringDonationSchedule__c> schedules = getExistingSchedules(rd);
+
+        if (schedules == null) {
+            return vizSchedules;
+        }
```
npsp-reedestockton: @force2b As currently built, getNextDonationsForVisualization() assumes we're dealing with a single RD.  The underlying implementation of getExistingSchedules uses a bulk pattern with an overload for a single RD, so it wouldn't be hard to bulkify getNextDonationsForVisualization().  We'd just need to build a map and iterate through it for the rest of the logic.  I think we should probably make that change.  What do you think?

npsp-reedestockton: If we make that change, we can use the same overloading pattern used for getExistingSchedules() to make it easy to use for either single RD or bulk and support a return type of map for the bulk pattern and list for the single RD.

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,128 @@ public without sharing class RD2_ScheduleService {
...
+    */
+    public class Installment {
+
+        public Id       campaignId {get; set;}
+        public String   campaignName {get; set;}
+        public String   dayOfMonth {get; set;}
+        public Date     endDate {get; set;}
+        public String   externalId {get; set;}
+        public Decimal  installmentAmount {get; set;}
+        public Integer  installmentFrequency {get; set;}
```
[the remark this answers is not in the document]
npsp-reedestockton: It was useful during debugging, but I frankly don't know of a use case from this point forward that will require them to be present.  I'll pull them out.

src/lwc/rdScheduleVisualizer/rdScheduleVisualizer.html
```diff
@@ -0,0 +1,25 @@
+<!-- Enhanced Recurring Donations Schedule Visualizer -->
+<!-- TODO Accessibility!! -->
```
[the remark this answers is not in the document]
npsp-reedestockton: I'll do both.  Would like to get comfortable with both Axe and screen reader.

npsp-reedestockton:
Accessibility check remains outstanding, but all other issues have been addressed.

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
 
+
+    /***
+    * @description Calculates the dates of the next n donations
+    * @param rds Recurring Donation records
+    * @param visualizationStartDate Date
+    * @param maxDates Integer
+    * @return Map<Id, List<Installment>>
+    */
+    public List<Installment> getNextDonationsForVisualization(npe03__Recurring_Donation__c rd, Date visualizationStartDate, Integer maxDates) {
```
force2b: What do you think about changing this to:
```suggestion
    public List<Installment> getVisualizedInstallments(npe03__Recurring_Donation__c rd, Date startDate, Integer numInstallments) {
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+
+        if (maxDates == null) {
+            maxDates = DEFAULT_VISUALIZATION_SCHEDULE_COUNT;
+        }
+
+        if (visualizationStartDate == null) {
+            visualizationStartDate = currentDate;
+        }
+
+        Map<Id, List<Installment>> installmentsByRDIds = new Map<Id, List<Installment>>();
```
force2b:
```suggestion
        Map<Id, List<Installment>> installmentsByRd = new Map<Id, List<Installment>>();
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+        Map<Id, List<RecurringDonationSchedule__c>> schedulesByRDIds = getExistingSchedules(rds);
```
force2b:
```suggestion
        Map<Id, List<RecurringDonationSchedule__c>> schedulesByRd = getExistingSchedules(rds);
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+
+        for (npe03__Recurring_Donation__c rd : rds) {
+
+            if (rd == null) {
+                continue;
+            }
+
+            installmentsByRDIds.put(rd.Id, new List<Installment>());
```
force2b: I think you can just eliminate this line and change line 271 from `installmentsByRDIds.get(rd.Id).addAll(vizSchedules);` to `installmentsByRDIds.put(rd.Id, vizSchedules);`

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+
+            Installment ins;
```
force2b: I don't think this declaration is needed either. Line 260 can just make the declaration in the loop `Installment ins = getNextDonationSchedule(filteredSchedules);`

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+            List<Installment> vizSchedules = new List<Installment>();
+            currentDate = visualizationStartDate;
+            Integer dateCount = 0;
+            List <RecurringDonationSchedule__c> filteredSchedules;
+
+            if (schedulesByRDIds.get(rd.Id) == null) {
+                continue;
+            }
+
+            while (dateCount < maxDates) {
```
force2b: Could this just be `while (visSchedules.size() < numInstallments) {` this way the `dateCount` var isn't needed at all?

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+            startDate =             rdSchedule.StartDate__c;
+        }
+    }
+
+    /***
+    * @description Calculates the date of the next donation
+    * @param rdSchedules Recurring Donation Schedule records
+    * @return Date
+    */
+    public Installment getNextDonationSchedule(List<RecurringDonationSchedule__c> rdSchedules) {
```
force2b: Rather than define `currentDate` in the calling method, I'm suggesting to pass that value as a parameter to this method and have this method "override" `currentDate`. This way the logic related to that is encapsulated in the method it's used from. Also, does this need to be a public method?

```suggestion
    public Installment getNextInstallment(List<RecurringDonationSchedule__c> rdSchedules, Date startingDate) {
         currentDate = startingDate;
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+
+        Date nextDonationDate;
+        Installment nextDonationSchedule;
```
force2b:
```suggestion
        Installment nextInstallment;
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+        Installment previousDonationSchedule;
```
force2b:
```suggestion
        Installment previousInstallment;
```

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,152 @@ public without sharing class RD2_ScheduleService {
...
+
+        if (rdSchedules == null) {
+            return nextDonationSchedule;
+        }
+
+        for (RecurringDonationSchedule__c schedule : rdSchedules) {
+
+            nextDonationDate = getNextDonationDateGreaterThanToday(schedule);
+            if (nextDonationDate != null &&
```
force2b: To make this loop a bit more readable, how about:
```
    if (nextDonationDate == null) {
        continue;
    }
    if ((schedule.EndDate__c == null || nextDonationDate <= schedule.EndDate__c) &&
            (previousDonationSchedule == null || previousDonationSchedule.nextDonationDate > nextDonationDate)) {
        ....
```

src/classes/RD2_VisualizeScheduleController.cls
```diff
@@ -0,0 +1,150 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Recurring Donations
+* @description Controller for the Schedule Visualization Component
+*
+*/
+public with sharing class RD2_VisualizeScheduleController {
```
force2b: I'm considering using this same controller class for the LWC that renders the active schedule records since there are a number of parts of this that are similar. I can see a couple of approaches, but I'm curious what you think would work well here.

- Rename this class to `RD2_ScheduleController` and have methods for both LWC components. That's likely `getInstallments()` (renamed from `getSchedule`) and `getActiveSchedule()`.
- Create a base abstract class for `RD2_ScheduleController` that has the common methods and use one controller class per LWC that extends from the base: `RD2_VisualizeScheduleController` and `RD2_ShowActiveScheduleController`

Or the third option is to just have two different classes without any overlap.

src/classes/RD2_VisualizeScheduleControllerTest.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>47.0</apiVersion>
```
force2b: Should be api v46

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -949,9 +1016,281 @@ private with sharing class RD2_ScheduleService_TEST {
...
+        Test.startTest();
+        insert rd;
+        List<RD2_ScheduleService.Installment> installments = service.getNextDonationsForVisualization(rd, Date.newInstance(2019, 11, 1), 480);
+        Test.stopTest();
+
+        System.assertEquals(480, installments.size(), 'There should be 480 installments.');
+        System.assertEquals(Date.newInstance(2059, 10, 1), installments[479].nextDonationDate, 'Final installment date should be 10/1/2059.');
+
+    }
+
```
force2b: What would you think about a test that returns a limited number of installments, based on an EndDate in the schedule? For example, if the Schedule record has a StartDate of 1/1/2020, and EndDate to 10/30/2020 and the method is called to retrieve 12 schedules, the unit test would verify that only 10 are returned.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: Don't think so.  We'd want to return an empty schedule if, for whatever reason, there were no valid upcoming dates.  In that case, we'd need an empty schedule list.

src/classes/RD2_VisualizeScheduleControllerTest.cls-meta.xml
npsp-reedestockton: Interesting.  I thought our approach was to always go with latest.  What drives our choice?

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: Sounds good.  If we do that, can't we say we already support fixed length recurring donations?

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: I'll use startDate instead of startingDate to be consistent with what we have done elsewhere.

npsp-reedestockton: Actually, having already worked through this once and watched how it borked everything, I'm going to push back a bit on this one.  While I agree that encapsulating changes is generally a good practice, it doesn't really buy us anything in this case.  Now that we have bulkified the caller, we have to reset to startDate with every iteration of the schedule analysis for each RD passed in.  We also need to reset to current date at the end of the entire process to ensure that any caller holding a reference to RD2_ScheduleService will end up with a status that is what they rightfully expect.  To me, that means that encapsulating startDate as you suggest is actually a negative, not a positive.  It doesn't get us anything of value and makes it more cumbersome to do what we require.

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: Answering my own question...  Nope.  We'd need End Date on RD in order to make that claim.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: Alternatively....  is it worth considering a more radical set of changes to accept date inputs pretty much everywhere?  The fact that we are relying on a class variable to handle currentDate is useful for testing, but is resulting in some awkward patterns like what we are encountering here.  Worth some refactoring perhaps?

src/classes/RD2_VisualizeScheduleController.cls
npsp-reedestockton: I think it might be pretty straightforward to have one controller for both components, but I guess it depends at least partially on how many additional items this might be used for.  If you think it will expand further, it might make some sense to go the abstract base route for the stuff related to RD and currency code.  Or, if it offends our standard approach to not have a one apex controller to one javascript controller model, I suppose separate would be okay too.  But my vote is for one.

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +193,153 @@ public without sharing class RD2_ScheduleService {
...
+                    if (schedule.EndDate__c == null || schedule.EndDate__c >= currentDate) {
+                        filteredSchedules.add(schedule);
+                    }
+                }
+
+                Installment ins = getNextInstallment(filteredSchedules);
+                if (ins == null) {
+                    currentDate = date.today();
+                    break;
+                }
```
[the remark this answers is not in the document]
npsp-reedestockton: Yes, that looks like a good approach to me.  That accomplishes the sort of date manipulation encapsulation you were looking for.

src/classes/RD2_ScheduleService.cls
```diff
@@ -364,6 +529,53 @@ public without sharing class RD2_ScheduleService {
         return schedulesByRdId;
     }
 
+    /***
+    * @description Retrieves recurring donation schedules associated with a single recurring donation
+    * @param rd npe03__Recurring_Donation__c
+    */
+    public List<RecurringDonationSchedule__c> getExistingSchedules(npe03__Recurring_Donation__c rd) {
```
force2b: Do these methods need to be public?

src/classes/RD2_ScheduleService.cls
```diff
@@ -364,6 +529,53 @@ public without sharing class RD2_ScheduleService {
...
+        }
+
+        return getExistingSchedules(new List<npe03__Recurring_Donation__c> {rd}).get(rd.Id);
+    }
+
+    /***
+    * @description Retrieves recurring donation schedules associated with recurring donation list
+    * @param rds List<npe03__Recurring_Donation__c>
+    */
+    public Map<Id, List<RecurringDonationSchedule__c>> getExistingSchedules(List<npe03__Recurring_Donation__c> rds) {
```
force2b: I have a suggestion for this. When the Schedule Service is called from the OpportunityEvaluationService class for the nightly job and queuable, the RD record will already have a sub-query on active RecurringDonationSchedule records (see [this method](https://github.com/SalesforceFoundation/NPSP/blob/feature/RD2/src/classes/RD2_OpportunityEvaluationService.cls#L254)). So, for this method it might be helpful if it first checked to see if there's an subquery on the RD (`RecurringDonationSchedules__r != null` possibly) to avoid an extra query when the visualization service is called from the OppEvalService. You can modify that query to add the foreign relationships if needed.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: The overload for the single RD is actually not used at all.  The list version needs to be public.  I'm removing the unused overload.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: I can do this, but doesn't that create a pretty tight linkage between these two pieces of code?  Let's discuss Monday.

force2b: It doesn't have create a linkage. What I'm thinking is to use the sub-query if it's there, otherwise initiate a query. 

Maybe something like the following:
```
Set<Id> rdsToQuery = new Set<Id>();
    // might need to apply a namespace to the field in the keyset. Unsure.
    if (rd.getPopulatedFieldsAsMap().keySet().containsKey('RecurringDonationSchedules__r') {
        oldSchedules.put(rd.Id, rd.RecurringDonationSchedules__r);
    } else {
        rdsToQuery.add(rd.Id);
    }
}

if (rdsToQuery.isEmpty()) {
    return oldSchedules;
} 

// the rest of the logic to query and loop goes here
```

npsp-reedestockton: I'd be happier with providing an override to the visualization method that allows the user to pass in a map of RD Ids, to schedules.  That makes it an explicit request on the part of the user to use the passed in schedules to determine the installments.  The aspect of your suggestion that makes me uncomfortable is the implicit rather than explicit nature of the request.

npsp-reedestockton:
@force2b Code review changes complete.  Unless you have more, we should be good to go.

src/classes/RD2_ScheduleService.cls
```diff
@@ -185,6 +194,167 @@ public without sharing class RD2_ScheduleService {
...
+
+            installments.add(ins);
+            currentDate = ins.nextDonationDate.addDays(1);
+        }
+
+        return installments;
+    }
+
+    /***
+    * @description Wrapper class for Recurring Donation schedule data
```
force2b: The description here might be better as "Wrapper class for Recurring Donation Installment data"

npsp-reedestockton:
@force2b I agree about the builder pattern refactor.  When you mentioned it earlier this morning I briefly considered implementing now, but it impacts too many other things to implement quickly at the end of our code review cycle.

npsp-reedestockton:
@ravensisco This one is ready for testing.  Thanks!

PR #4879, opened 2019-12-11
CCI Task to cleanup translation metadata after using retrieve_unpackaged task

force2b:
* Add a new CCI task for `cleanup_translation_metadata` that can be used after new translations have been received from the translations team. This task should only be used after running the `retrieve_unpackaged` against a scratch org that does not include scratch org SampleData. The purpose of this is to strip out translated components in the metadata that should not be included in our managed package. Some examples are translations from underlying packages and translations from unmanaged metadata.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


force2b:
@prescod Should the `modify.py` file be in the root CCI repo vs. Cumulus?

davisagli:
@force2b Paul is in the process of submitting these improvements to the task that already exists in CCI: https://github.com/SFDO-Tooling/CumulusCI/pull/1389

force2b:
Thanks @davisagli I'll revert that file from this PR and flag this for holding until SFDO-Tooling/CumulusCI#1389 is merged.

prescod:
Yes that plan sounds good. It should just take a week or at most 2 for it to come out as part of CCI.

force2b:
@prescod and @davidmreed I think this PR is good to go now. If you agree, would you mind approving?

force2b:
**lurch: attach W-035813

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAKSQA4" target="_blank">W-035813</a>

cumulusci.yml
```diff
@@ -341,6 +341,164 @@ tasks:
                 - "connection was cancelled here"
             retry_always: True
 
+    cleanup_translation_metadata:
+        description: Delete extraneous and translation-excluded metadata from translation files after using retrieve_unpackaged to pull down cleaned translations from a scratch org.
+        class_path: tasks.modify.RemoveElementsXPath
```
davidmreed: @force2b  Does this need to be changed back to the core CCI class?

force2b: Good question. I thought that was the core class.

prescod: The core CCI class is:

cumulusci.tasks.metadata.modify.RemoveElementsXPath

I "forked" it for you so you wouldn't need to wait until CCI included a version with the delete feature.

force2b: Ah. Thanks. I'll fix here and in all the underlying repo's as well.

PR #4881, opened 2019-12-12
Create a Script to Easily Rebuild a QA Org on NPSP with Latest Changes

Julian88Tex:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/scripts/qa_pull_and_rebuild.sh
+++ b/scripts/qa_pull_and_rebuild.sh
@@ -11,31 +11,31 @@ if [[ $REPLY =~ ^[Yy]$ ]]
 then
 
 echo What branch would you like to pull most recent changes from?? [example: feature/julian-qa-rebuild-script]
-read 1
+read branch_name
 
 echo What would you like to name your org? [example: W-123456]
-read 2
+read org_name
 
 
 # Remove any existing modified files so a pull can take place
 git checkout -- .
 
 # Switch to desired branch
-git checkout $1
+git checkout branch_name
 
 # Pull most recent changes
 git pull
 
 # Delete exisiting scratch org if one exists with provided name
-cci org scratch_delete $2
+cci org scratch_delete org_name
 
 # Rebuild org with qa flow
-cci flow run qa_org --org $2
+cci flow run qa_org --org org_name
 
 # Mark org as default
-cci org default $2
+cci org default org_name
 
 # Open newly created org
-cci org browser $2
+cci org browser org_name
 
 fi
--- a/scripts/qa_pull_and_rebuild.sh
+++ b/scripts/qa_pull_and_rebuild.sh
@@ -1,21 +1,27 @@
 #!/usr/bin/env bash
 
 # Script to be used by qa to rebuild a scratch environment.
-
+echo 
+echo ***WARNING******WARNING******WARNING******WARNING******WARNING******WARNING******WARNING******WARNING******WARNING******WARNING******WARNING***
+echo 
 echo This script is to be used by qa to rebuild a scratch environment.
-echo ***WARNING*** This script will rollback any modifications to files and delete a scratch org with the same name given. Please commit any file changes before running this script. ***WARNING***
+echo This script will rollback any modifications to files and delete a scratch org with the same name given.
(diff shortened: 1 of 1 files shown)


Julian88Tex:
**lurch: attach W-038159

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HprXQAS" target="_blank">W-038159</a>

scripts/qa_pull_and_rebuild.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Script to be used by qa to rebuild a scratch environment.
+# To run paste into command line "bash scripts/qa_rebuild_script.sh <orgname> <branch>"
+# For example "bash scripts/qa_rebuild_script.sh W-037818 feature/RD2__rad-data-migration-display-errors"
+
+# ***IMPORTANT*** This script will rollback any modifications to files and delete a scratch org with the same name.
+
+# Remove any existing modified files so a pull can take place
+git checkout -- .
```
davidmreed: @Julian88Tex  I understand why you're doing this, but it makes me nervous - especially doing this without a prompt. 

How the user manages their checkouts and branch switches (and default orgs, which some users do not use at all) can vary a lot and making this assumption silently concerns me as in some sense made official by incorporating this script into CumulusCI.

I feel more comfortable keeping this as a user script for now, but understand the highlighting of a pain point in CCI vis-a-vis rapidly getting access to an org on a new branch.

davidmreed: We'd be interested (without making any specific promises!) in better understanding the use cases around rebuilding orgs, which could yield some improvements to CCI's `org` command tree.

Julian88Tex: @davidmreed Makes sense. What if I add this big scary prompt? Does it feel safe enough?

Julian88Tex: @davidmreed In terms of use case, would it be most helpful for me to add that to a RelEng ticket? Or a CumulusCI Issue? Or somewhere else?

The gist of the use case is often I will create a scratch org, a developer will make a change, then I'll need to recreate a scratch org. I'm guessing the first question might be "why don't you just repush the changes instead of rebuilding". That may work in some scenarios, but fairly often we are testing one-time enablement/feature setup items that can't really easily be restarted.... and if they can be restarted I often wouldn't call a manual restart a valid test of enablement.... especially if it's UI-related.

davidmreed: I strongly agree that users should build new orgs, not attempt to re-push new changes to the same org.

I'd like to propose two things:

 - Distribute this script via (e.g.) https://salesforce.quip.com/L9tmAVVA5Hu2, which @jstvz  just put together to help share efficient user scripts and aliases.
 - If you'd be willing, start a Quip doc discussing this and other org-management workflows that represent inefficiencies or pain points for the QE team so we can have a conversation about how we could consider evolving the CCI `org` command set.

I don't think that the NPSP repo is the best distribution vehicle.

Julian88Tex: @davidmreed Thank you for this feedback! Just added to @jstvz's doc and will think about how/where to start the discussion. I'll be demoing the local script at our review to get some more feedback on how useful it is. Based on that feedback I'll probably find a place to have that discussion and see what next steps are.

Are you opposed to me instead adding to the .gitignore file? Or, is there an existing .gitignore naming convention that is recommended I use to format any local files to?

davidmreed: @Julian88Tex  You can add it to your global `.gitignore`, or to a repository-local exclude file (`.git/info/exclude`): https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

Julian88Tex: @davidmreed Cool thanks didn't even realize I could have a .gitignore outside of repo folder but that makes sense. Made the change!

PR #4901, opened 2019-12-19
Emergency Fix for Duplicate Field in CRLP Rollup Queries

### LINKED ISSUE #4899: Customizable Rollups: Duplicate field selected error in Soft Credit rollup job
Issue emerging with NPSP 3.166 release. 

With Customizable Rollups enabled and no changes made to out of the box rollups, Soft Credit Rollup job is failing with an error that is visible in the NPSP Error Log:

```
npsp__Largest_Soft_Credit_Amount__c,npsp__Largest_Soft_Credit_Date__c,npo02__LastOppAmount__c

^

ERROR at Row:1:Column:259

duplicate field selected: Amount

Rollups State:
```

This error does not occur with legacy rollups.

force2b:
Assuming this is the root area of the issue, and that for whatever reason the `Set<DescribeResult>` is no longer ensuring uniqueness, this puts the field name strings into a `Set<String>` that we're confident **is** unique and then builds the additional query string from that set. 

# Critical Changes

- Fixes #4899 

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_Query_SEL.cls
+++ b/src/classes/CRLP_Query_SEL.cls
@@ -197,8 +197,6 @@ public inherited sharing class CRLP_Query_SEL {
      * @return part of a SOQL string
      */
     private static String appendParentQuery(SObjectType parentObject, String parentKeyField) {
-        String addlQuery = '';
-
         Set<Schema.DescribeFieldResult> allFieldsToQuery = new Set<DescribeFieldResult>();
         allFieldsToQuery.addAll(CMT_FilterRule_SEL.getReferencedFieldsByObject(parentObject));
         allFieldsToQuery.addAll(CRLP_Rollup_SEL.getReferencedFieldsByObject(parentObject));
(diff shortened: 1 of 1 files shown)


force2b:
**lurch: attach W-038326

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hq8TQAS" target="_blank">W-038326</a>

src/classes/CRLP_Query_SEL.cls
```diff
@@ -210,20 +208,27 @@ public inherited sharing class CRLP_Query_SEL {
             allFieldsToQuery.add(SObjectType.Opportunity.fields.Amount.getSobjectField().getDescribe());
         }
 
+        Set<String> queryFields = new Set<String>();
```
ptewson-sfdo: I think we talked through this to mutual satisfaction, but for a hypothetical future refactor it might be nice to pass the Set<String> of fields through all the methods that add to this query just as the most robust way to ensure absolutely uniqueness of every field in the list, regardless of the source (CC: @lmeerkatz) However I think that's beyond the scope for this fix.

PR #4905, opened 2019-12-20
Gift Entry: Update template list view and add batch list view

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
- Lightning Components
  - geListView
  - utilCumulusStaticResources
  - utilDualListbox
- Aura Components
  - GE_GiftEntry
- Custom Setting
  - Custom_Column_Header__c
- Custom Labels
  - commonListView
  - commonNoItems
  - commonTemplates
  - commonViewAll
  - commonViewMore
  - geAssistiveShowMenu
  - geBodyTemplatesTabDescription
  - geButtonEnterGifts
  - geButtonTemplatesTabCreateTemplate
  - geHeaderCustomTableHeaders
  - geHeaderGiftEntry
  - geLabelCustomTableSelectedFields
  - geLabelCustomTableSourceFields
  - geTextListViewItemCount
  - geTextListViewItemsCount
  - geTextListViewSortedBy
  - geTextListViewUpdatedAgo
  - geToastTemplateDeleteSuccess
# Deleted Metadata


psadie-sfdo:
**lurch: attach W-037896

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HpRYQA0" target="_blank">W-037896</a>

psadie-sfdo:
**lurch: attach W-037509

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HofpQAC" target="_blank">W-037509</a>

src/labels/CustomLabels.labels
```diff
@@ -4085,6 +4085,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Field Label</shortDescription>
         <value>Field Label</value>
     </labels>
+    <labels>
+        <fullName>commonListView</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>List View</shortDescription>
```
```suggestion
        <shortDescription>Button to navigate to a list view page</shortDescription>
```

src/labels/CustomLabels.labels
```diff
@@ -4085,6 +4085,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <value>List View</value>
```
```suggestion
        <value>Edit List View</value>
```

src/labels/CustomLabels.labels
```diff
@@ -4539,6 +4587,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Warning message for deleting a section without required fields</shortDescription>
         <value>Deleting this section also removes any fields in it.</value>
     </labels>
+    <labels>
+        <fullName>geBodyFormFieldsRightCol</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Form Fields right column text block</shortDescription>
```
```suggestion
        <shortDescription>Form Fields sidebar text block</shortDescription>
```

src/labels/CustomLabels.labels
```diff
@@ -4539,6 +4587,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
+        <value>Click a section, then select a field in the sidebar to add it to the section in the main screen. For each field, you can edit the field label, mark it as required, assign a default value or remove it from the section. The gift entry form layout can be managed by creating sections and reordering the sections or fields.</value>
```
```suggestion
        <value>Click a section, then select a field in the sidebar to add it to the section in the main screen. For each field, you can edit the field label, mark it as required, assign a default value or remove it from the section. Adjust the layout of the gift entry form by creating sections and reordering the sections or fields.</value>
```

src/labels/CustomLabels.labels
```diff
@@ -4865,13 +4945,38 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>Template Info</value>
     </labels>
     <labels>
-        <fullName>geToastTemplateCreateSuccess</fullName>
```
melissabarber: @psadie-sfdo If I'm reading this correctly, it looks like it changed to the geTextListViewItemCount custom label. Did I get that right? Might be good to get a little demo of this change. I added a comment in the custom labels spreadsheet too.

psadie-sfdo: @melissabarber  - super late response! No change in this label actually. This diff just looks a little wonky. `geToastTemplateCreateSuccess` just got moved down a bit farther to line 4998 because the order changed as I committed additional labels in this case `geTextListViewItemCount`.

src/lwc/geListView/geListView.js
```diff
@@ -0,0 +1,714 @@
...
+                typeAttributes: { rowActions: this.actions },
+            }];
+        }
+    }
+
+    /*******************************************************************************
+    * @description Method builds a SOQL query based on the currently selected list
+    * view's describe info.
+    */
+    buildSoqlQuery(selectedFieldApiNames) {
```
ptewson-sfdo: I have a few questions here:
1) first of all, should we building the SOQL query in JS as opposed to sending the needed info into the call to the Apex method and building the query using the tools available there already?
2) Is there a potential security risk for building the SOQL query in JS? 
3) If we DO want to build the SOQL in javascript, should this be moved to a more commonly accessible utility not just built into the list view component?

src/lwc/geTemplates/geTemplates.js
```diff
@@ -1,44 +1,107 @@
...
+const IS_LOADING = 'isLoading';
+const EVENT_TOGGLE_MODAL = 'togglemodal';
+const SAVE = 'save';
+
+const TEMPLATES_LIST_VIEW_NAME = 'Templates';
+const TEMPLATES_LIST_VIEW_SORT_DIRECTION = 'desc';
+const TEMPLATES_LIST_DEFAULT_LIMIT = 10;
+const TEMPLATES_LIST_VIEW_ICON = 'standard:visit_templates';
+const TEMPLATES_TABLE_ACTIONS = [
     { label: 'Edit', name: 'edit' },
```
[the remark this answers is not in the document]
psadie-sfdo: Yep, definitely! Turning the labels into custom labels and the names to constants.

src/lwc/geTemplates/geTemplates.js
```diff
@@ -1,44 +1,107 @@
...
-    @track columns = columns;
-    @track isLoading = true;
-    currentNamespace;
+    @track templatesTableActions = TEMPLATES_TABLE_ACTIONS;
     @track isAccessible = true;
+    @track isLoading = true;
 
     get templateBuilderCustomTabApiName() {
-        return this.currentNamespace ? `${this.currentNamespace}__GE_Template_Builder` : 'GE_Template_Builder';
+        return TemplateBuilderService.alignSchemaNSWithEnvironment(TEMPLATE_BUILDER_TAB_NAME);
```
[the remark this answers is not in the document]
psadie-sfdo: Unfortunately yes for this. The main issue is that we can't get the tab describe info through the schema like we do for objects/fields and hence the TEMPLATE_BUILDER_TAB_NAME constant there.

We then need to align that string with the proper namespace because for example if we're in a 'namespaced' org and navigate to the tab _without_ that namespace aligned URL, all of our imports from the schema in the web component will then not include a namespace breaking all references relying on those schema imports.

Additional details on that finding in this quip doc https://salesforce.quip.com/BNCFAalNyIqy#QFFACAw12r8

src/lwc/geListView/geListView.js
psadie-sfdo: Yeah, this was also the most if-y section of this work in my mind. This bit was left over from when I was relying on the metadata from the uiListApi and decided to make the most minimal amount of change and still have it work.

The only input into this query builder that's end-user accessible is for the query's select clause via their selection of column headers in that dual-listbox modal. I _want_ to say it's relatively safe as we're not pulling strings/values that's directly written by the user.

Regardless I totally get the concern and am down to move this logic into the apex method if that's what we want 👍

src/lwc/geListView/geListView.js
```diff
@@ -1,7 +1,7 @@
 import { LightningElement, api, track, wire } from 'lwc';
 import { getObjectInfo } from 'lightning/uiObjectInfoApi';
 import { dispatch, handleError, format, generateId, sort, deepClone, showToast } from 'c/utilTemplateBuilder';
-import CumulusStaticResources from 'c/utilCumulusStaticResources';
+import LibsMoment from 'c/libsMoment';
```
psadie-sfdo: @ptewson-sfdo - Took a closer look at this and ran into some issues with the resulting references from the loadScript method With that said, I'd actually like to make this component _more_ specific and make it responsible for only importing the moment.js library. I've renamed it to reflect that here and I think makes what this component does more clear. Imo we should either do single responsibility library import components like this or have one that imports all the resources that we ship in cumulus static resources. Maybe we can take a look at creating a more generic one at a later time?

ptewson-sfdo: I think this all makes sense, and thanks for looking into that - I think given those findings, updating this to be more specific is the right way to go!

src/lwc/geListView/geListView.js
ptewson-sfdo: @kenlewis-sfdo This was what I wanted to get a second opinion on - This is replicating some SOQL query-building function that already exists in apex within the Javascript controller, and I'm wondering both if this is a good idea, and if we would want to bundle this as its own service component if we did like it within the front end. Or any other thoughts or concerns!

src/lwc/geTemplates/geTemplates.js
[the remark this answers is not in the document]
ptewson-sfdo: That's a bummer, but happy to see all the collective wisdom and diligence we are gathering to inform our approach. nice!

src/lwc/geListView/geListView.js
kenlewis-sfdo: If we move ahead with keeping it, a service component would be a good idea.  But my opinion would be to lean away from replicating the SOQL query-building functionality in JS.  Since we already have it in Apex and this component already imports it (`FORM_ServiceGiftEntry.retrieveRecords`), I wouldn't keep it unless we verify that the performance gain outweighs the added maintenance having query-building logic in multiple places.  I would pass the fields as an array to the Apex method and maybe add a `withOffset` method to the `UTIL_Query` class to implement infinite scroll on the `lightning-datatable`.

ptewson-sfdo: I think that's how I'm leaning as well (or at least how I'm leaning now that I know that's how Ken is leaning!) ... but it feels better to only build that code once and use it in Apex where it's already developed.

PR #4909, opened 2019-12-24
fixing a namespace issue in error message validation

skristem:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


skristem:
@sgovindankutty Configure Advanced Mapping is a button on Data Imports Advanced Mapping page. Yes Advanced Mapping is being enabled at org level. Org should come with it being disabled and then a user should enable it and test stuff

PR #4919, opened 2019-12-31
Gift Entry: enforce unique form template names in builder

psadie-sfdo:
# Critical Changes

# Changes
- Add a maximum length of 80 characters to the Form Template record Name input field
- Add a check for Form Template Name uniqueness

# Issues Closed

# New Metadata

# Deleted Metadata


psadie-sfdo:
**lurch: attach W-037361

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HoNFQA0" target="_blank">W-037361</a>

src/classes/FORM_ServiceGiftEntry_TEST.cls
```diff
@@ -126,6 +126,31 @@ public with sharing class FORM_ServiceGiftEntryTest {
...
+        String templateJSON = JSON.serialize(template);
+        fs.storeFormTemplate(null,template.name, template.description, template.version, templateJSON);
+
+        Form_Template__c[] formTemplates = [SELECT Id, Name, Template_JSON__c FROM Form_Template__c];
+
+        Boolean isExistingName = FORM_ServiceGiftEntry.checkNameUniqueness(
+            formTemplates[0].Name,
+            null);
+
+        Boolean someUniqueName = FORM_ServiceGiftEntry.checkNameUniqueness(
```
kenlewis-sfdo: Since this is a boolean, maybe rename to `isUniqueName`?  That would make the assert more readable below also.

src/classes/FORM_ServiceGiftEntry_TEST.cls
```diff
@@ -126,6 +126,31 @@ public with sharing class FORM_ServiceGiftEntryTest {
...
+            'New Form Template Sample Name',
+            null);
+
+        System.assertEquals(isExistingName, false);
+        System.assertEquals(someUniqueName, true);
```
kenlewis-sfdo: The expected value is the first argument for the `assertEquals` method, so maybe switch the order of the parameters here?  And might be helpful to have an error message specified also as the third param (_"Existing template names should return false from checkNameUniqueness..."_ or something similar).

melissabarber:
Reviewed with @psadie-sfdo. Looks good for docs.

PR #4920, opened 2019-12-31
Feature/robot  affiliations pom

skristem:
Converting Affiliation test cases to Page Object Model and also making tests more stable
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/AffiliationPageObject.py
```diff
@@ -0,0 +1,24 @@
...
+    
+@pageobject("Details", "Affiliation")
+class AffiliationDetailPage(BaseNPSPPage,DetailPage ):
+    object_name = "npe5__Affiliation__c"
+
+    def _is_current_page(self):
+        """ Verify we are on the Account detail page
+            by verifying that the url contains '/view'
+        """
+        self.selenium.wait_until_location_contains("/view", timeout=60, message="Record view did not open in 1 min")
```
boakley: 👍 I like the use of custom error messages here.

robot/Cumulus/resources/NPSP.py
```diff
@@ -413,15 +413,21 @@ def click_managehh_link(self,title):
...
 
         for i in locators:
             locator = i.format(title)
             if self.check_if_element_exists(locator):
-                self.selenium.get_webelement(locator).click()
+                checkbox=self.selenium.get_webelement(locator)
+                if (status == 'checked' and checkbox.is_selected() == False) or (status == 'unchecked' and checkbox.is_selected() == True):
+                    self.selenium.click_element(checkbox)
+                else:
+                    self.builtin.log("This checkbox is already in the expected status", "WARN")
```
boakley: "WARN" doesn't seem like the correct level here. It seems to me like "INFO" would be better. Ultimately it's your decision, but warnings generally should be for something unexpected.

robot/Cumulus/resources/NPSP.py
```diff
@@ -1279,4 +1286,50 @@ def verify_toast_message(self,value):
                 found=True
                 print("Toast message verified")
                 break
-        assert found, "Expected Toast message not found on page"
+        assert found, "Expected Toast message {} not found on page".format(value)
+
+    def edit_record_field_value(self,field,value):
```
boakley: This is a good example of why I like robot - sometimes the simplest user action requires lots of minor details, like scrolling an element into view. This is a cool keyword because it hides a bunch of complexity behind an easy-to-understand keyword.

robot/Cumulus/resources/NPSP.py
```diff
@@ -1279,4 +1286,50 @@ def verify_toast_message(self,value):
...
+        footer=npsp_lex_locators["record"]["footer"]
+        self.selenium.wait_until_page_contains_element(footer)
+        self.set_checkbutton_to(field,status)    
+        
+    def save_record(self): 
+        """Saves record by clicking on footer button 'Save'"""
+        footer=npsp_lex_locators["record"]["footer"]
+        self.click_record_button("Save")
+        self.selenium.wait_until_page_does_not_contain_element(footer)
+        self.selenium.execute_javascript("window.scrollTo(0,0)")
```
boakley: It might be good to add a comment as to why you're scrolling the window. 

Other than that, this is another good, robust keyword that does some work (clicking a button) and then waiting before returning. 👍

robot/Cumulus/tests/browser/affiliations/change_secondary_as_primary_affiliation.robot
```diff
@@ -1,29 +1,39 @@
...
+...             robot/Cumulus/resources/AccountPageObject.py
+...             robot/Cumulus/resources/AffiliationPageObject.py
+...             robot/Cumulus/resources/NPSP.py
+Suite Setup     Run keywords
+...             Open Test Browser
+...             Setup Test Data
 Suite Teardown  Delete Records and Close Browser
 
+***Keywords***
+Setup Test Data
```
boakley: This isn't a comment, but a question. I've seen code like this in several tests that I've reviewed lately. I'd like to find a way to make this less tedious to write.  It's cumbersome to call a keyword, save the value to a variable, then call `set suite variable` on that variable, and do that several times before calling `Store Session Record`.

It would be nice if we could do that all on one line. If you have a chance, I'd love to hear your opinions on how we could make creating test data easier.  

For example, maybe we could write our own keyword that works like `Set suite variable` but which takes a keyword instead of regular data, so you could do something like this: 

```
Create suite data  &{account}  API Create Organization Account
Create suite data  &{contact}  API Create Contact  email=automation@example.com
```

We could even make it so that you can create multiple variables at once, sort of like how `Run keywords` works

```
Create suite data
...       &{account}  API Create Organization Account
...  AND  &{contact}  API Create Contact  email=automation@example.com
```
I'm just brainstorming here -- if you have ideas for how would could make the creation of test data easier, I'd like to hear them.

robot/Cumulus/resources/locators_47.py
```diff
@@ -16,6 +16,7 @@
     },
     "record": {
         'button': "//div[@class='actionsContainer']/button[@title='{}']",
+        'footer':"//div[@class='footer active' or contains(@class,'footer-visible')]",
```
boakley: is it really "footer<space>active" but "footer<dash>visible? It's a shame they aren't consistent "footer active" and "footer visible" or "footer-active" and "footer-visible".

robot/Cumulus/resources/NPSP.py
skristem: the scenarios I have used this keyword (so far)should not have this behavior if the org is set up correctly so I wanted to add it as a warning for a manual tester as a heads up in case those are not expected.

robot/Cumulus/resources/locators_47.py
skristem: unfortunately yes there is no consistency :(

robot/Cumulus/resources/NPSP.py
```diff
@@ -329,13 +314,13 @@ def verify_header(self,value):
...
         
-    def Verify_affiliated_contact(self,list_name,first_name,last_name, y):   
-        """Validates if the affiliated contacts have the added contact details enter Y for positive case and N for negative case"""
-        name = first_name + ' ' + last_name
+    def verify_related_list(self,list_name,status,name):   
+        """If status is 'contains' then the specified related list should contain name
+                        'does not contain' then the specified related list should not contain name"""
         locator = self.salesforce.get_locator('record.related.link', list_name, name)
-        if y.upper()=="Y":
+        if status=="contains":
```
rjanjanam: status.lower() == "contains", maybe a good idea to avoid confusion

robot/Cumulus/resources/NPSP.py
```diff
@@ -1288,4 +1280,57 @@ def verify_toast_message(self,value):
...
+    def edit_record_checkbox(self,field,status):
+        """Scrolls just a little below the field
+           Clicks on Edit icon next to field
+           checks if status is 'checked' 
+           unchecks if status in 'unchecked'"""
+        scroll_loc=npsp_lex_locators["span_button"].format(field)
+        # To make sure the field we want to edit has rendered
+        # and is not obscured by the footer, scroll down a little below the element
+        self.selenium.scroll_element_into_view(scroll_loc)
+        self.selenium.execute_javascript("window.scrollBy(0,50)")
```
rjanjanam: Why do you have to scroll and execute javascript? Is the view of the element blocked by something ?

robot/Cumulus/resources/NPSP.py
skristem: added comment

robot/Cumulus/resources/NPSP.py
skristem: yes we have to scroll just a little below the actual element we want to click otherwise click is obstructed by footer

robot/Cumulus/resources/NPSP.py
skristem: thats why i specified in documentation to pass either 'contains' or 'does not contain'

rjanjanam: Yes true, just a safe practice to avoid human errors. Not a must. Just a suggestion.

PR #4921, opened 2020-01-02
NPSP DevOrg Test Data Tweaks

force2b:
Data Changes:
* Fix the sample RecurringDonation record to correct the InstallmentPeriod and OpenEndedStatus values. Also changed the starting date of the RD to December 2019 so the data is more current
* Add 2 years to all Opportunity CloseDates so they range from 2016-2019.
* Extract some additional fields on the Payment, Allocation, and RecurringDonation objects

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


datasets/dev_org/test_data.sql
```diff
@@ -687,160 +689,174 @@ CREATE TABLE "Contact" (
...
+	"Description__c" VARCHAR(255), 
+	"Level_Field__c" VARCHAR(255), 
+	"Maximum_Amount__c" VARCHAR(255), 
+	"Minimum_Amount__c" VARCHAR(255), 
+	"Previous_Level_Field__c" VARCHAR(255), 
+	"Source_Field__c" VARCHAR(255), 
+	"Target__c" VARCHAR(255), 
+	PRIMARY KEY (id)
+);
+INSERT INTO "Level__c" VALUES(1,'Gold Donor','true','','Current_Level__c','','10000.0','Previous_Level__c','npo02__TotalOppAmount__c','Contact');
```
davidmreed: We aren't mapping the field `Current_Level__c`, and I'm not sure where it's defined (same for `Previous_Level__c`). Is this a concern?

force2b: I don't think so. Those are unmanaged fields that get deployed into the org anyway. I think of these like Rollup fields, which we don't map either. In terms of performance, what would be really cool (and would also address this), is to disable rollups before inserting/updating Opps, and then run the rollup and levels job after the data load has completed. I think that's possible with some apex-anon that runs as part of the task and/or flow.

davidmreed: It looks like this field is defined in `config/qa`, which means this dataset comes with a dependency on that unpackaged metadata set, and that set is not currently deployed in `dev_org`. Am I missing another source for it?

force2b: Ah. That is a really good point. The data load won't fail if the fields don't exist because these are just text fields with those field names, however the Levels job would crash when run if the fields aren't there. 

Darn. I was hoping to include a Level in the data to provide more realistic test data. I will remove the Level__c object from the mapping/sql.

PR #4933, opened 2020-01-05
RD2: Unmanaged Metadata Updates

force2b:
* Updates to the Unmanaged metadata deployed when Enhanced Recurring Donations is enabled:
   - Update the HelpText for the StartDate__c field to match recently released change
   - Update the FlexiPage to include the ScheduleVisualizer component
   - Add permissions to apex classes needed for LWC to the Standard User Profile

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata


unpackaged/config/rd2_post_config/flexipages/NPSP_Recurring_Donation.flexipage
```diff
@@ -0,0 +1,113 @@
...
+        <name>main</name>
+        <type>Region</type>
+    </flexiPageRegions>
+    <flexiPageRegions>
+        <componentInstances>
+            <componentInstanceProperties>
+                <name>displayNum</name>
+                <value>12</value>
+            </componentInstanceProperties>
+            <componentName>%%%NAMESPACE%%%rdScheduleVisualizer</componentName>
```
force2b: @davidmreed I was wondering if there is a chance that CCI would be able to use the `:` separator instead of the `__` seperator for flexipages when converting `%%%NAMESPACE%%%` tokens?

davidmreed: @force2b You can use the special `%%%NAMESPACE_OR_C%%%` token for those component references.

force2b: Thank you!

unpackaged/config/rd2_post_config/flexipages/NPSP_Recurring_Donation.flexipage
```diff
@@ -0,0 +1,113 @@
...
+            <componentInstanceProperties>
+                <name>tabs</name>
+                <value>facet-7003451d-d97d-44c2-b74f-b56f89537a36</value>
+            </componentInstanceProperties>
+            <componentName>flexipage:tabset</componentName>
+        </componentInstances>
+        <name>main</name>
+        <type>Region</type>
+    </flexiPageRegions>
+    <flexiPageRegions>
```
npsp-reedestockton: In my test, the Schedule Visualizer is not on the Recurring Donations page until I add it and set the page as org default.  Is that the expected behavior?

unpackaged/config/rd2_post_config/flexipages/NPSP_Recurring_Donation.flexipage
```diff
@@ -0,0 +1,113 @@
...
+        <componentInstances>
+            <componentInstanceProperties>
+                <name>displayNum</name>
+                <value>12</value>
+            </componentInstanceProperties>
+            <componentName>%%%NAMESPACE_OR_C%%%rdScheduleVisualizer</componentName>
```
force2b: @davidmreed In a non-namespaced org, the above token is getting replaced a value of "c:c". The error I get in deployment is `Update of FlexiPage NPSP_Recurring_Donation: Error: We couldn't retrieve the design time component information for component c:crdScheduleVisualizer`. Seems strange to be honest.

unpackaged/config/rd2_post_config/flexipages/NPSP_Recurring_Donation.flexipage
force2b: It should be there. One thing to check is that "page" you're using is the "NPSP Recurring Donations" page. There's a picklist for the page at the top left of the Page editor. If there was a Lighting Page already for that object, I think this one doesn't replace the pre-existing one.

unpackaged/config/rd2_post_config/flexipages/NPSP_Recurring_Donation.flexipage
force2b: Though, since Reede didn't see that error when he ran this task, maybe it only happens when the task is run after the flexipage is deployed the first time. Now that would be weird.

npsp-reedestockton: I am seeing that error now.

force2b: Interesting. Clearly the second time is the charm, which really makes no sense. Any thoughts on what might be happening here @davidmreed?

npsp-reedestockton: I tried the install a 2nd time on a fresh org to see if it was intermittent, and it showed up again.  I think it is likely that I just missed the error on my initial test.

davidmreed: I'll dig into this as soon as we get the release out.

davidmreed:
```suggestion
            <componentName>%%%NAMESPACE_OR_C%%%:rdScheduleVisualizer</componentName>
```

It turned out to be a _very_ confusing error message. The token doesn't insert the `:`, but the error message _adds_ a `c:`. Hence, we tried to deploy `crdScheduleVisualizer` and the system told us `c:crdScheduleVisualizer` didn't exist.

force2b: Thanks @davidmreed. Follow-up question. How do I test this against a beta org? I'm trying the following to force the namespace injection, but nothing works. Tried using `-o namespaced_org true` as well.
`cci task run deploy_rd2_config --org rd2beta -o namespace_inject npsp`

davidmreed: @force2b  You can safely omit `namespace_inject`, because it's specced in the task definition. For beta, use `-o unmanaged False`.

force2b: That did it. Thank you!! I confirm that adding the `:` worked.

npsp-reedestockton: Works great!!!

force2b:
@davidmreed  I change this PR so it points to the `feature/RD2` branch this way we can merge into that branch and only have a single merge into master. If you're fine with that I'll merge.

PR #4938, opened 2020-01-09
Gift Entry: Added validation on template save

voduyemi:
# Critical Changes

# Changes

- Made changes to the template save validation logic to enforce the Donation Donor and one of the follow fields: Account1 Imported, Account1 Name, Contact1 Imported, Contact1 Last Name on template save.

# Issues Closed

# New Metadata

- Custom Label : `geErrorPageLevelMissingRequiredGroupFields`

# Deleted Metadata

### DIFF
--- a/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
+++ b/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
@@ -1,5 +1,4 @@
 import { LightningElement, track, api, wire } from 'lwc';
-import { findIndexByProperty, mutable, generateId, dispatch, showToast } from 'c/utilTemplateBuilder';
 import { generateId, dispatch, showToast } from 'c/utilTemplateBuilder';
 import { mutable, findIndexByProperty } from 'c/utilCommon';
 import TemplateBuilderService from 'c/geTemplateBuilderService';
--- a/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
+++ b/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
@@ -86,8 +86,7 @@ export default class geTemplateBuilderFormFields extends LightningElement {
             if (REQUIRED_FORM_FIELDS.includes(el.getAttribute('data-source-api-name')) && el.checked) {
               countOfConditionallyRequiredFormFields++;
             }
-            if (!el.checked && (el.required ||
-                el.getAttribute('data-source-api-name') === DONATION_DONOR_INFO.fieldApiName)) {
+            if (el.required && !el.checked) {
                 objectMappingsWithMissingRequiredFields.add(el.getAttribute('data-object-mapping'));
                 let objectMappingLabel = el.getAttribute('data-object-mapping-label');
                 missingRequiredFieldMappings.push(`${objectMappingLabel}: ${el.label}`);
@@ -371,6 +370,8 @@ export default class geTemplateBuilderFormFields extends LightningElement {
             this.handleRemoveFormElement(name);
         }
 
+        fieldMapping.Is_Required =
+            fieldMapping.Target_Field_Label === DONATION_DONOR_LABEL ? true : fieldMapping.Is_Required;
         if (fieldMapping.Is_Required) {
             this.validate();
         }
@@ -655,7 +656,7 @@ export default class geTemplateBuilderFormFields extends LightningElement {
             let character;
             if (i < REQUIRED_FORM_FIELDS.length - 1) {
                 character = COMMA_CHARACTER;
-            }else if ( i === REQUIRED_FORM_FIELDS.length - 1) {
+            } else if ( i === REQUIRED_FORM_FIELDS.length - 1) {
                 character = PERIOD_CHARACTER;
             }
             REQUIRED_FORM_FIELDS_MESSAGE += objectData.fields[REQUIRED_FORM_FIELDS[i]].label + character;
--- a/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
+++ b/src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
@@ -150,8 +150,8 @@ export default class geTemplateBuilderFormFields extends LightningElement {
                 this.objectMappingNames = [...this.objectMappingNames, objMappingDevName];
 
                 let fieldMappings = TemplateBuilderService.fieldMappingsByObjMappingDevName[objMappingDevName];
-                fieldMappings.forEach(fieldMapping => {
-                    if(fieldMapping.Target_Field_Label === DONATION_DONOR_LABEL){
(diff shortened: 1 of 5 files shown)


src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -16,8 +16,13 @@ import PAYMENT_METHOD_INFO from '@salesforce/schema/DataImport__c.Payment_Method
 import ACCOUNT1_IMPORTED_INFO from '@salesforce/schema/DataImport__c.Account1Imported__c';
 import CONTACT1_IMPORTED_INFO from '@salesforce/schema/DataImport__c.Contact1Imported__c';
 import DONATION_DONOR_INFO from '@salesforce/schema/DataImport__c.Donation_Donor__c';
+import CONTACT1_LASTNAME_INFO from '@salesforce/schema/DataImport__c.Contact1_Lastname__c';
+import ACCOUNT1_NAME_INFO from '@salesforce/schema/DataImport__c.Account1_Name__c';
 
 const WARNING = 'warning';
+const DONATION_DONOR_LABEL = 'Donation Donor';
+const REQUIRED_FORM_FIELDS_MESSAGE = 'Account1 Imported, Account1 Name, Contact1 Imported, Contact1 Last Name';
```
ptewson-sfdo: Can we grab these field labels from the imported schema information (ACCOUNT1_IMPORTED_INFO, etc) instead of hard-coding the strings? Same with "Donation Donor" above?

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -58,15 +71,27 @@ export default class geTemplateBuilderFormFields extends LightningElement {
         let missingRequiredFieldMappings = [];
 
         const elements = this.template.querySelectorAll('lightning-input');
+        let counter = 0;
```
ptewson-sfdo: Maybe name this variable something more descriptive (countOfConditionallyRequiredFormFields perhaps)?

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -58,15 +71,27 @@ export default class geTemplateBuilderFormFields extends LightningElement {
...
+
         elements.forEach(el => {
-            if (el.required && !el.checked) {
+            if (REQUIRED_FORM_FIELDS.includes(el.getAttribute('data-source-api-name')) && el.checked) {
+              counter++;
+            }
+            if (el.required && !el.checked ||
+                el.getAttribute('data-source-api-name') === DONATION_DONOR_INFO.fieldApiName && !el.checked) {
```
ptewson-sfdo: Could this be simplified a bit as "!el.checked && (el.required  || el.getAttribute('data-source-api-name') === DONATION_DONOR_INFO.fieldApiName)"?

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
voduyemi: It doesn't seem like we can retrieve the field label through the schema, the variable returned from the schema call only carries the objectApiName and fieldApiName. Except if we want to do some string manipulation on the fieldApiName, although we won't be able to guarantee that they will be the same(fieldApiName and field label).

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
voduyemi: Absolutely!

voduyemi:
**lurch: attach W-036283

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CBAtQAO" target="_blank">W-036283</a>

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
voduyemi: I take back my comment, found a way around it. Thanks for calling it out!

melissabarber:
@ptewson-sfdo and @voduyemi I reviewed in a demo with Victor, and it looks good to me!

ptewson-sfdo:
This looks great @voduyemi - I really like the handling of the field labels to avoid hard-coding those. I did manage to get a doubled list of fields at one point, but now I can't reproduce it - any idea what might be going on? 

image

ptewson-sfdo:
@voduyemi I think you may need to run the "cci task run update_package_xml" task and then check in an updated package.xml as well - when I first pulled this it gave me an error on deploy that some resource wasn't included in the manifest. Thanks!

voduyemi:
@ptewson-sfdo I'm not sure why you got the double list of fields, but I haven't been able to reproduce it, weird. I'll run the update package.xml cci task to update my package.xml and check-in the updated one

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -604,4 +644,21 @@ export default class geTemplateBuilderFormFields extends LightningElement {
...
+     * @description Method generates the required fields text attached to the validation message.
+     *
+     * @param {object} objectData: Returned Object Schema
+     */
+    buildRequiredFieldsMessage(objectData){
+        for (let i= 0; i< REQUIRED_FORM_FIELDS.length; i++) {
+            let character;
+            if (i < REQUIRED_FORM_FIELDS.length - 1) {
+                character = COMMA_CHARACTER;
+            }else if ( i === REQUIRED_FORM_FIELDS.length - 1) {
```
ptewson-sfdo:
```suggestion
            } else if ( i === REQUIRED_FORM_FIELDS.length - 1) {
```

src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -604,4 +644,21 @@ export default class geTemplateBuilderFormFields extends LightningElement {
...
+                character = PERIOD_CHARACTER;
+            }
+            REQUIRED_FORM_FIELDS_MESSAGE += objectData.fields[REQUIRED_FORM_FIELDS[i]].label + character;
```
ptewson-sfdo: I'm still not sure what I'm doing to reproduce the issue, but I'm occasionally getting repeated fields in the error message (see screenshot). Do we need to clear this array out at the beginning of this method, since the variable is in scope and continually appended to? 

image

voduyemi: To handle it now, I'm clearing out the string variable just before I start building the message, I still haven't been able to reproduce it though

ptewson-sfdo: I don't think I can reproduce it with those changes now either! Thanks!

PR #4939, opened 2020-01-09
Remove Custom Settings Field References from Visualforce Pages

### LINKED ISSUE #4934: Payment Scheduler: CRUC causes Non-System Admin to see NPSP Settings Button at Top of Page
When a non-System Admin clicks on the "Schedule Payments" button and views the Payment Wizard VF Page (npsp__PMT_PaymentWizard), an NPSP Settings button displays at the top and they do not have a section to create a payment schedule.

This is happening because their Profile does not have access to the npe01__Contacts_And_Orgs_Settings__c Custom Setting.

We display an NPSP Settings button if the following criteria are true:
https://github.com/SalesforceFoundation/NPSP/blob/master/src/pages/PMT_PaymentWizard.page#L55-L63
- the Payment object .isCreateable(), isDeletable(), and  isAccessible().
- the Contact and Org Custom Setting's Automatic Payment Creation field is unchecked (npe01__Payments_Enabled__c)
paymentsSettings.npe01__Payments_Enabled__c == false

The "Require Customize Application permission for direct read access to custom settings" Critical Update was just auto-activated in orgs. If a profile doesn't have explicit permission to that Custom Setting, it thinks npe01__Payments_Enabled__c == false even if it's checked off. As a result, it displays the NPSP Settings button.

The overall fix for this is to grant Profiles access to that Custom Setting. However, there may need to be some additional code written to see that the User actually has permission to a Custom Setting so we don't display this button incorrectly.

Screen Shot 2020-01-06 at 3 31 53 PM

meena-sfdc:
# Critical Changes

# Changes
As a user I want to be able to access Visualforce pages even if I do not have access to Custom Settings, so that I can get my work done without having admin permissions.

# Issues Closed
- Fixes #4934

# New Metadata

# Deleted Metadata


src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -154,43 +154,55 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
+
+    /*******************************************************************************************************
+    * @description Returns true if Payments enabled is true on paymentsSettings
+    * @return Boolean
+    */
+    public Boolean getIsPaymentsEnabled() {
+        if (getPaymentsSettings() == null) {
+            return false;
+        }
+        return getPaymentsSettings().npe01__Payments_Enabled__c;
```
force2b: Normally I'd say this is fine, however Checkbox fields on a Custom Settings object is the one exception where the field value can actually be null (normally checkbox fields are _always_ either true or false). So, to avoid an error evaluating something like `!isPaymentsEnabled` in the VF page, I recommend modifying this statement as follows. This way the method explicitly returns true only if the checkbox value is true and returns false otherwise.
```suggestion
        return getPaymentsSettings().npe01__Payments_Enabled__c == true;
```

meena-sfdc:
**lurch: attach - W-038485

src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -154,43 +154,55 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
+
         return paymentsSettings;
     }
-  
+
+    /*******************************************************************************************************
+    * @description Returns true if Payments enabled is true on paymentsSettings
+    * @return Boolean
+    */
+    public Boolean getIsPaymentsEnabled() {
```
force2b: This will definitely work, but I think you should change the `paymentsSettings` var to private since it no longer needs to be exposed on the VF page.

src/classes/PMT_PaymentWizard_TEST.cls
```diff
@@ -669,6 +669,64 @@ private class PMT_PaymentWizard_TEST {
         System.assert(redirect);
     }
 
+    /*********************************************************************************************************
+    * @description Verify if Warning message is displayed in Schedule Payment when Automatic Payment is disabled
+    * in Payment custom settings
+    */
+    @isTest
+    private static void shouldDisplayWarningInSchedulePaymentWhenAutoPaymentDisabled() {
+        npe01__Contacts_And_Orgs_Settings__c PaymentsSettingsForTests = UTIL_CustomSettingsFacade.getContactsSettingsForTests(
```
force2b: There's actually no need to assign this to a var since the method isn't using the `PaymentsSettingsForTests` var anyway. Just call the method without the assignment. Same for the other new test methods.

meena-sfdc: I didn't notice it, @force2b , will take out the assignment.

src/classes/PMT_PaymentWizard_CTRL.cls
meena-sfdc: @force2b , When I change this var to private, then when logging as standard user, I get the method not visible error when I navigate to schedule payment page.

meena-sfdc: Screen Shot 2020-01-10 at 4 08 32 PM

force2b: Did you change this method to private? If so, what I was recommending above was to change the `paymentSettings` var to private.

meena-sfdc: > Did you change this method to private? If so, what I was recommending above was to change the `paymentSettings` var to private.

Yes, I got it wrong, sorry, will change paymentsSettings, @force2b

RadSFDeveloper:
**lurch: attach W-038477

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HqO2QAK" target="_blank">W-038477</a>

RadSFDeveloper:
**lurch: attach W-038485

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HqSiQAK" target="_blank">W-038485</a>

meena-sfdc:
@jesslopez-sf , This Item is ready for QA!

rponti-sforg:
QE testing is complete for this PR and associated WI's.

https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001HqSiQAK/view
https://salesforce.quip.com/6a1bAnz8dvm3

PR is merged.  Will set associated WI's to Close after post builds pass (success)

PR #4952, opened 2020-01-13
Clean up extra functions from feature/gift-entry__defaultValue merge

bdvorachek:
# Critical Changes

# Changes
- Update deepClone to be in-line with latest version from utilTemplateBuilder
- Add isPrimative from utilTemplateBuilder
- Clean up duplicate functions from earlier merge
# Issues Closed

# New Metadata

# Deleted Metadata


src/lwc/utilCommon/utilCommon.js
```diff
@@ -36,6 +36,10 @@ const debouncify = (anyFunction, wait) => {
 const deepClone = (src) => {
```
psadie-sfdo: @bdvorachek - could we move this method below all the `is...` methods it's relying on? My IDE is complaining of usage before defining.

PR #4957, opened 2020-01-15
RD2: Consider future Closed Won/Lost Opps when matching projected installments

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_OpportunityService.cls
+++ b/src/classes/RD2_OpportunityService.cls
@@ -273,13 +273,9 @@ public with sharing class RD2_OpportunityService {
     */
     private List<Opportunity> getCurrentOpps(npe03__Recurring_Donation__c rd) {
         List<Opportunity> opps = new List<Opportunity>();
-
-        if (rd.npe03__Donations__r.isEmpty()) {
-            return opps;
-        }
         
         for (Opportunity opp : rd.npe03__Donations__r) {
-            if ((opp.IsClosed == false && opp.CloseDate > currentDate) || opp.CloseDate == currentDate) {
+            if (opp.CloseDate >= currentDate) {
                 opps.add(opp);
             }
         }
--- a/src/classes/RD2_OpportunityEvaluationService_TEST.cls
+++ b/src/classes/RD2_OpportunityEvaluationService_TEST.cls
@@ -413,6 +413,61 @@ private class RD2_OpportunityEvaluationService_TEST {
     }
 
     /**
+     * @description Verifies future open Opp is updated when an active RD is updated. 
+     * The future Closed Won Opp is not updated.
+     */
+    @IsTest
+    private static void shouldNotUpdateFutureOpenOppCloseDateWhenRDNextDonationDateIsNotUpdated() {
+        //set current date override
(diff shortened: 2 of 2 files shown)


RadSFDeveloper:
**lurch: attach W-038666

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hqm9QAC" target="_blank">W-038666</a>

src/classes/RD2_OpportunityService.cls
```diff
@@ -273,13 +273,9 @@ public with sharing class RD2_OpportunityService {
...
     private List<Opportunity> getCurrentOpps(npe03__Recurring_Donation__c rd) {
         List<Opportunity> opps = new List<Opportunity>();
-
-        if (rd.npe03__Donations__r.isEmpty()) {
-            return opps;
-        }
         
         for (Opportunity opp : rd.npe03__Donations__r) {
-            if ((opp.IsClosed == false && opp.CloseDate > currentDate) || opp.CloseDate == currentDate) {
+            if (opp.CloseDate >= currentDate) {
```
force2b: The `OpportunityEvaluationService.getRecurringDonations()` method that builds the subquery on Opps will still filter for `IsClosed=False OR CloseDate >= :currentDate`. Does that impact the change made to this condition?
https://github.com/SalesforceFoundation/NPSP/blob/feature/RD2__rad-update-matching-opp-on-rd-update/src/classes/RD2_OpportunityEvaluationService.cls#L348

If I'm following it correct, the query retrieves anything that is either open or in the `future` (as defined by CurrentDate). This condition would then just return anything in the `future` (i.e., all the record given that the `isClosed=True &&` is removed). That would include Open, ClosedWon and ClosedLost Opps - basically with a Close Date `in the future`. I think that makes sense based on the objective. 

My hope is to turn this into a part of an InstallmentOpportunityMatchService that takes a list of installments from the `getVisualizedInstallments()` method and a list of future opps, and reconciles the two collections returning what _should_ be the list of future installments - whether they exist as open or closed, or if they don't exist as opps yet. That returned list would be what we use for CurrentYear and NextYear calculations as well as for determining whether a future installment Opp needs to be created or updated.

RadSFDeveloper: The Opp eval service query returns all Open Opps regardless of the Close Date and all Opps with Close Date >= current date regardless if Open or Closed Won/Lost. The `getCurrentOpps` extracts all Opps with Close Date >= current date regardless if Open or Closed, so filtering is required.

In future PRs we will have centralized method/class that matches and filters out Opps, but it would be too much refactoring for the current OR. However, it is work in progress in the Current Year Value PR #4969 .

RadSFDeveloper:
@Julian88Tex , the PR is ready for testing.

Julian88Tex:
Testing is complete for this WI: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001Hqm9QAC/view

PR #4970, opened 2020-01-17
Gift Entry: use custom lookup component in template builder and batch wizard

psadie-sfdo:
# Critical Changes

# Changes
- Use `geFormFieldLookup` in Template Builder and Batch Wizard
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/utilInput/utilInput.js
+++ b/src/lwc/utilInput/utilInput.js
@@ -16,11 +16,15 @@ const DATETIME = 'datetime-local';
 const YES = 'Yes';
 const TEXT = 'text';
 const TRUE = 'true';
+const RICH_TEXT_FORMATS = [
+    'font', 'size', 'bold', 'italic', 'underline', 'strike', 'list', 'indent', 'align', 'link', 'clean', 'table', 'header'
+];
 
 export default class utilInput extends LightningElement {
 
     // expose custom labels to template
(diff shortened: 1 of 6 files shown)


psadie-sfdo:
**lurch: attach W-038443

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HqKKQA0" target="_blank">W-038443</a>

src/lwc/utilInput/utilInput.js
```diff
@@ -16,11 +16,15 @@ const DATETIME = 'datetime-local';
 const YES = 'Yes';
 const TEXT = 'text';
 const TRUE = 'true';
+const RICH_TEXT_FORMATS = [
```
ptewson-sfdo: What is this being used for? It doesn't seem like this + richTextFormats is actually being used, unless I'm missing something?

psadie-sfdo: Ah, yes this is the whitelist of formats being set in the `lightning-input-rich-text` component's `formats` property. That property controls some of the buttons and what formatted content can be pasted into that rich input.

I straight up lifted this list from the one being used in the renderer so we have alignment across Gift Entry with regards to rich text inputs.

PR #4972, opened 2020-01-17
Use custom labels always, fix SGE loading

bdvorachek:
# Critical Changes

# Changes
- Always display the label defined in the template when rendering fields
- Fixed an issue where the form did not load in single gift entry mode
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

**lurch: attach W-038207

### DIFF
--- a/src/lwc/geFormRenderer/geFormRenderer.js
+++ b/src/lwc/geFormRenderer/geFormRenderer.js
@@ -6,7 +6,7 @@ import geSave from '@salesforce/label/c.labelGeSave';
 import geCancel from '@salesforce/label/c.labelGeCancel';
 import geUpdate from '@salesforce/label/c.labelGeUpdate';
 import { showToast, handleError, getRecordFieldNames, setRecordValuesOnTemplate } from 'c/utilTemplateBuilder';
-import { getQueryParameters, isNotEmpty } from 'c/utilCommon';
+import {getQueryParameters, isEmpty, isNotEmpty} from 'c/utilCommon';
 import { getRecord } from 'lightning/uiRecordApi';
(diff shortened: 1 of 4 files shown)


LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HpwXQAS" target="_blank">W-038207</a>

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -69,6 +69,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
 
                 // get the target field names to be used by getRecord
                 this.fieldNames = getRecordFieldNames(this.formTemplate, this.fieldMappings);
+                this.initializeForm(this.formTemplate);
```
vmartinov-sforg: @bdvorachek Did you check if the contact / account field values still display after adding this line?

bdvorachek: I've updated this to handle the case where we do have a donor record id present and need to load existing data. If there is one present, the initializeForm call should happen during the wiredGetRecord for the donor.

vmartinov-sforg: I'm getting an exception, it looks like isEmpty() is being used in utilTemplateBuilder, but it was moved to utilCommon. It needs to be imported.

Uncaught (in promise) ReferenceError: LWC component's @wire target property or method threw an error during value provisioning. Original error:
[isEmpty is not defined]
    at Object.e.setRecordValuesOnTemplate (/lightning/n/modules/c/utilTemplateBuilder.js:4)

PR #4969, opened 2020-01-17
RD2: Calculate calendar Current and Next Year Values

RadSFDeveloper:
- Calculate expected current year value of a Recurring Donation so users can create projections of recurring donation revenue.
- Determine how much is expected from a recurring donor next year so that users could forecast the revenue they expect from the recurring giving program.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Apex Class:
- RD2_OpportunityMatcher
- RD2_OpportunityMatcher_TEST
- RD2_RecurringDonationsOpp_TDTM
- RD2_RecurringDonationsOpp_TEST

Custom Field:
- npe03__Recurring_Donation__c.CurrentYearValue__c
- npe03__Recurring_Donation__c.NextYearValue__c

# Deleted Metadata
Custom Label:
- RD2_ScheduleInstallmentIsRequired

PR #4973, opened 2020-01-17
RD2: Do Not Reset Effective Date to Current Date

npsp-reedestockton:
Modified the behavior or changes to Recurring Donation schedule so that Effective Date (StartDate__c) is always respected.  Previously, Effective Date was changed to current date if there was no change to Effective Date itself.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
***lurch: attach W-038591*

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HqeKQAS" target="_blank">W-038591</a>

npsp-reedestockton:
@force2b @RadSFDeveloper Ready for review.

npsp-reedestockton:
@sam-knox Don't believe changes need to be published. Unpublished change description is the only PR text.

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -332,34 +334,66 @@ private with sharing class RD2_RecurringDonations_TEST {
...
-        rd.npe03__Amount__c = 100;
+        rd.npe03__Amount__c = 250;
         rd.InstallmentFrequency__c = 2;
         update rd;
         Test.stopTest();
 
+        List<RecurringDonationSchedule__c> rdSchedules = rdScheduleGateway.getRecords(rd);
+        System.assertEquals(2, rdSchedules.size(), 'Should be 2 schedules.');
+
+        System.assertEquals(startDate, rdSchedules[0].StartDate__c, 'Original start date should be 10/5/2019.');
```
RadSFDeveloper: Since we would like to address and document expected behaviour in tests, mentioning dates in assert statements can bring extra detail that might be redundant. Based on the assert statements, it seems that expected behaviour is to neutralize the existing original schedule and create a new schedule valid from the original schedule Start Date? 

If so, then we can write tests to document behaviour:
```
System.assertEquals(startDate, rdSchedules[0].StartDate__c, 'Original schedule Start Date should be unchanged');
System.assertEquals(startDate.addDays(-1), rdSchedules[0].EndDate__c, 'Original schedule End Date should be changed so the schedule is invalid');
System.assertEquals(startDate, rdSchedules[1].StartDate__c, 'New schedule Start Date should be original schedule Start Date ');
System.assertEquals(null, rdSchedules[1].EndDate__c, 'New schedule End Date should be undefined');
```

RadSFDeveloper: @force2b ^, is the expected behaviour for the old schedule to be set as invalid when Amount changes and Effective Date (which is in past) is unchanged?

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -332,34 +334,66 @@ private with sharing class RD2_RecurringDonations_TEST {
...
+        insert rd;
+        Test.startTest();
+        rd.npe03__Amount__c = 250;
+        rd.StartDate__c = null;
+        update rd;
+        Test.stopTest();
+
         npe03__Recurring_Donation__c rd2 = rdGateway.getRecord(rd.Id);
-        System.assertEquals(Date.today(), rd2.StartDate__c, 'Start date should be today()');
+        System.assertEquals(Date.today(), rd2.StartDate__c, 'Next donation date should be today()');
```
RadSFDeveloper: - No need to create a new var for the RD, the existing one can be reused. 
- The assert statement should be mentioning RD Effective Date.
```
rd = rdGateway.getRecord(rd.Id);
System.assertEquals(Date.today(), rd.StartDate__c, 'RD Effective Date should be set to today');
```

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -332,34 +334,66 @@ private with sharing class RD2_RecurringDonations_TEST {
...
+
+        List<RecurringDonationSchedule__c> rdSchedules = rdScheduleGateway.getRecords(rd);
+        System.assertEquals(2, rdSchedules.size(), 'Should be 2 schedules.');
+
+        System.assertEquals(startDate, rdSchedules[0].StartDate__c, 'Original start date should be 11/1/2019.');
```
RadSFDeveloper: Similarly here, the assert statements should be documenting expected behaviour without utilizing details such as specific date values.

src/lwc/rdActiveSchedule/rdActiveSchedule.html
```diff
@@ -30,68 +30,68 @@ <h3 class="slds-tile__title slds-truncate" if:true={schedule.isCurrent} title={l
                             <div class="slds-tile__detail slds-m-around--xx-small">
                                 <dl class="slds-list_horizontal slds-wrap slds-grid">
 
-                                    <dt class="slds-dl_horizontal__label slds-text-color_weak slds-medium-size_5-of-12" title={labels.lblAmount}>
+                                    <dt class="slds-dl_horizontal__label slds-text-color_weak slds-medium-size_5-of-12 slds-border_bottom slds-border_top slds-p-bottom--xxx-small slds-p-top--xxx-small" title={labels.lblAmount}>
```
RadSFDeveloper: @Julian88Tex , the active schedule component needs to be tested in this PR.

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -1017,6 +1017,56 @@ private with sharing class RD2_ScheduleService_TEST {
         System.assertEquals(0, schedules.size(), 'There should be zero schedules.');
     }
 
+    /****
+    * @description Validates start date on RD with unmodified past start date
+    */
+    @IsTest
+    private static void shouldApplyRDChangesIfPastStartDateUnchanged() {
```
RadSFDeveloper: The test should be verifying the RD submitted to the update trigger is marked for update and Effective Date is unchanged? If so, the ApexDoc description and test can be written as:
```
    /****
    * @description Validates the RD past Effective Date is unchanged when Amount is changed 
    * and schedule is marked for update.
    */
    @IsTest
    private static void shouldMarkRDForScheduleUpdateWhenStartDateIsInPast() {
        Date startDate = Date.newInstance(2019, 11, 1);
        RD2_EnablementService_TEST.setRecurringDonations2Enabled();

        List<npe03__Recurring_Donation__c> oldRds = new List<npe03__Recurring_Donation__c>{
            getRecurringDonationMonthlyBuilder()
                .withStartDate(startDate)
                .withMockId()
                .build()
        };

        Decimal newAmount = 250;
        List<npe03__Recurring_Donation__c> updatedRds = new List<npe03__Recurring_Donation__c>{
            oldRds[0].clone(true)
        };
        updatedRds[0].npe03__Amount__c = newAmount;
        
        RD2_ScheduleService service = new RD2_ScheduleService();
        Map<Id, npe03__Recurring_Donation__c> rdsNeedUpdate = service.getRecurringDonationsNeedingScheduleUpdates(
            updatedRds, oldRds, TDTM_Runnable.Action.BeforeUpdate
        );

        System.assertEquals(1, rdsNeedUpdate.size(), 'Recurring Donation should be marked for update');
        System.assertEquals(newAmount, rdsNeedUpdate.get(rd.Id).npe03__Amount__c, 'Updated amount should match');
        System.assertEquals(startDate, rdsNeedUpdate.get(rd.Id).StartDate__c, 'RD Effective Date should be unchanged');
    }
```

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -1017,6 +1017,56 @@ private with sharing class RD2_ScheduleService_TEST {
...
+        System.assertEquals(1, rdsNeedUpdate.size(), 'There should be 1 Recurring Donation which needs to be updated.');
+        System.assertEquals(250, rdsNeedUpdate.get(rd.Id).npe03__Amount__c, 'Updated amount should equal 250.');
+        System.assertEquals(startDate, rdsNeedUpdate.get(rd.Id).StartDate__c, 'Start date should be 11/1/2019.');
+    }
+
+    /****
+    * @description Validates start date on RD with unmodified future start date
+    */
+    @IsTest
+    private static void shouldApplyRDChangesIfFutureStartDateUnchanged() {
```
RadSFDeveloper: Let's make changes for this test as recommended for the previous one.

src/classes/RD2_RecurringDonations_TEST.cls
force2b: I believe that we previously determined any changes to the contents of the schedule (dates, amounts, payment methods, etc.) will cause the existing Schedule record to be set as inactive (gets an end date) and a new schedule record is created. However, I think what you're asking is if we have a Schedule record that hasn't really started yet and a change is made to a field on the RD should the  Schedule record just be updated rather than create a new one.

I suppose that technically either are valid in this scenario, however I would choose the path of least resistance in terms of code/logic changes. If the existing behavior works fine - where it closes the previous schedule and creates a new one - then I'm not sure there's a lot of value in refactoring it to do something different in a unique case such as this.

src/classes/RD2_ScheduleService.cls
```diff
@@ -698,12 +695,11 @@ public without sharing class RD2_ScheduleService {
 
             if (isScheduleUpdateNeeded(rd, oldRd)) {
                 if (triggerAction == TDTM_Runnable.Action.BeforeUpdate &&
-                    rd.StartDate__c == oldRd.StartDate__c &&
-                    rd.Status__c != RD2_Constants.STATUS_CLOSED
-                ) {
+                        rd.StartDate__c == null &&
```
force2b: Is there ever a scenario where rd.StartDate can be null? I believe this is required field not just in the UI, but also in our validation triggers.

src/lwc/rdActiveSchedule/rdActiveSchedule.html
```diff
@@ -61,7 +61,7 @@ <h3 class="slds-tile__title slds-truncate" if:true={schedule.isCurrent} title={l
                                         <lightning-formatted-text value={labels.lblStartDate}></lightning-formatted-text>
                                     </dt>
                                     <dd class="slds-dl_horizontal__detail slds-medium-size_7-of-12">
-                                        <lightning-formatted-date-time value={schedule.startDate} year="2-digit" month="2-digit" day="2-digit" time-zone="UTC"></lightning-formatted-date-time>
+                                        <lightning-formatted-date-time value={schedule.startDate} year="numeric" month="2-digit" day="2-digit" time-zone="UTC"></lightning-formatted-date-time>
```
force2b: Nice. Just learned something new here.

src/classes/RD2_RecurringDonations_TEST.cls
```diff
@@ -332,34 +334,66 @@ private with sharing class RD2_RecurringDonations_TEST {
...
-            .withAmount(50)
-            .withInstallmentPeriod(RD2_Constants.INSTALLMENT_PERIOD_MONTHLY)
-            .withDayOfMonth('10')
-            .withInstallmentFrequency(1)
-            .withStatus(RD2_Constants.STATUS_ACTIVE)
-            .withDateEstablished(Date.newInstance(2019, 6, 5))
-            .withStartDate(Date.newInstance(2019, 6, 5))
+    private static void shouldNotOverrideStartDateWithCurrentDate() {
+        Date startDate = Date.newInstance(2019, 10, 5);
+        npe03__Recurring_Donation__c rd = getRecurringDonationMonthlyBuilder()
```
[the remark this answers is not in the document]
npsp-reedestockton: I'd like to understand the best practice we are targeting.  We have refactored other tests significantly, so I'd like to understand what drives the distinction between the two approaches.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: Without this code, the following code results in a null pointer exception:

`npe03__Recurring_Donation__c rd = new npe03__Recurring_Donation__c(Id='a09J000000O12Q5IAJ', npe03__Amount__c = 400, StartDate__c = null);
update rd;
`

src/classes/RD2_RecurringDonations_TEST.cls
npsp-reedestockton: @RadSFDeveloper @force2b Our logic so far is that every schedule change puts an end date on any old schedule and creates a new schedule with an appropriate start date (sometimes the end date on a schedule is before the start date if we consider the schedule to be permanently disabled).  If schedules were reportable user records we might not approach it this way, but given the fact that schedules are permanently and irrevocably inaccessible to our users, the only thing we should value is the clarity of what a schedule is intended to do, and this is to drive the creation of future opportunities.  For that, it doesn't matter at all whether we update or insert a new schedule, so I don't see any value to the increased complexity that results from doing thing in two different ways.

RadSFDeveloper: Thank you for the clarification!

src/classes/RD2_RecurringDonations_TEST.cls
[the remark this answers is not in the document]
RadSFDeveloper: New helper methods are applied to changed tests only although they might be used in more tests. Most importantly, since this is a critical change, less moving parts makes it easier to determine existing tests work as before for the same RD condition. I am additionally making changes to existing tests in PR #4969  and there might be duplication of the effort.

npsp-reedestockton:
@force2b @RadSFDeveloper Please review again.  Thanks!

src/classes/RD2_ScheduleService.cls
force2b: I see your point. I think we only require on the page layout, but not in our apex validation.

Julian88Tex:
@npsp-reedestockton took me going against the QA bones in my body but will merge with some robot tests failing since they are failing on every branch and kinds of tests failing are unlikely related to this change in any way so very unlikely they are related to this PR

PR #4982, opened 2020-01-21
Gift Entry: one gift entry tab.

psadie-sfdo:
# Critical Changes

# Changes
- Move all Gift Entry components under one Custom Tab (except Batch Gift Entry)
- Update `Gift_Entry` web links for Account and Contact objects
- Fix missing form template name in toast for template delete action

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Lightning Web Components
  - geHome
# Deleted Metadata


src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -140,7 +143,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
         this.reset();
 
         // go back to the donor record page
-        if(isNotEmpty(this.donorRecordId)) {
```
vmartinov-sforg: I think we need to add implementation for Cancel when the form gets loaded from the 'New Single Gift' button on the landing page. We can do that in a separate ticket.

PR #4983, opened 2020-01-21
Gift Entry: Feature Gating GE

voduyemi:
# Critical Changes

# Changes
- Page Access on the Template Builder, Gift Entry Landing Page and Single Gift Entry Form now checks for Advanced Mapping and the Gift Entry Feature.
- Updated Custom Labels

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom Setting Field : `Enable_Gift_Entry__c`

# Deleted Metadata


voduyemi:
**lurch: attach W-037632

**lurch: attach W-038563

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HouAQAS" target="_blank">W-037632</a>

src/labels/CustomLabels.labels
```diff
@@ -4824,15 +4824,15 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
...
         <shortDescription>Page-level error message body when Advanced Mapping off</shortDescription>
-        <value>This feature requires Advanced Mapping. Please enable it in NPSP Settings.</value>
+        <value>This feature requires both Advanced Mapping and Gift Entry. Please enable them in NPSP Settings.</value>
     </labels>
     <labels>
         <fullName>geErrorPageLevelAdvancedMappingHeader</fullName>
         <categories>Gift Entry</categories>
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Page-level error message header when Advanced Mapping off</shortDescription>
```
```suggestion
        <shortDescription>Page-level error message header when Advanced Mapping and Gift Entry are off</shortDescription>
```

melissabarber:
@voduyemi @ptewson-sfdo @psadie-sfdo I added a suggestion for a custom label shortDescription edit based on a review of the geErrorPageLevelAdvancedMappingHeader row Mark and I did recently. Is there anything else other than the custom labels you'd like me to review?

src/labels/CustomLabels.labels
voduyemi: Thanks! @melissabarber

src/objects/Gift_Entry_Feature_Gate_Settings__c.object
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <customSettingsType>List</customSettingsType>
```
ptewson-sfdo: What was the reasoning behind using a list custom setting? I think this may be better handled with a hierarchical setting (similar to the the Gift Entry Settings). And actually - should we just add this field to the Gift Entry settings we already introduced as part of this build? @psadie-sfdo

voduyemi:
@ptewson-sfdo For some reason, I can't find your comment, I used a list custom setting because the gating feature was temporary and we'd be removing it soon, but it does makes sense to add a field to `Gift Entry Settings`. I'll make that change.

ptewson-sfdo:
Sorry Victor - I meant to start a review but accidentally added the single comment, but then deleted it to re-add to a review - but thanks for jumping on it so quickly and I will update my branch with this!

melissabarber:
@voduyemi new field, you say...could I have you add that to the Customer-Visible Metadata tab in the Custom Labels sheet for me? @markross-sfdo and I will likely need to refine the description and help text for it. We don't need to refine those before this PR is merged; we can catch it in our final sweep of labels. Thanks! FYI @ptewson-sfdo

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
```diff
@@ -53,10 +54,16 @@ import ACCOUNT_INFO from '@salesforce/schema/Account';
 import commonError from '@salesforce/label/c.commonError';
 import commonUnknownError from '@salesforce/label/c.commonUnknownError';
 
+import getDataImportSettings from '@salesforce/apex/UTIL_CustomSettingsFacade.getDataImportSettings';
+import getGiftEntryFeatureGateSettings from
```
ptewson-sfdo: UTIL_CustomSettingsFacade.getGiftEntryFeatureGateSettings was included in previous commits, but I think is missing currently from your build

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
```diff
@@ -385,6 +392,20 @@ const setRecordValuesOnTemplate = (templateSections, fieldMappings, record) => {
...
+ * @description Method checks for page level access. Currently checks
+ * if Advanced Mapping is on from the Data Import Custom Settings and
+ * if the Gift Entry Feature Gate is turned on.
+ */
+const getPageAccess = async () => {
+    const dataImportSettings = await getDataImportSettings();
+    const giftEntryGateSettings = await getGiftEntrySettings();
+    const isAdvancedMappingOn =
+        dataImportSettings[FIELD_MAPPING_METHOD_FIELD_INFO.fieldApiName] === ADVANCED_MAPPING;
+    const isGiftEntryEnabled = giftEntryGateSettings.Enable_Gift_Entry__c;
```
ptewson-sfdo: This looks like isGiftEntryEnabled is returning "undefined" in my namespaced scratch org, even though I have the setting enabled - is it possible that something is missing here (maybe namespace related? @psadie-sfdo)

psadie-sfdo: Yeah, looks like we may want to import the field info for `Enable_Gift_Entry__c` here as we're doing with the `FIELD_MAPPING_METHOD_FIELD_INFO` for the Data Import Settings.

voduyemi: @ptewson-sfdo @psadie-sfdo Oh, I see. Thanks guys. Will add this change

src/lwc/geFormRenderer/geFormRenderer.html
```diff
@@ -1,4 +1,5 @@
 <template>
+    <template if:true={isAccessible}>
```
ptewson-sfdo: I think we need to adjust the indents on this file and move the contents of this <template> tag one tab in (so that the lightning-layout right  below here is not on the same level)

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -8,7 +8,11 @@ import geUpdate from '@salesforce/label/c.labelGeUpdate';
 import geLabelService from 'c/geLabelService';
 import { CONTACT1, ACCOUNT1,
         DONATION_DONOR_FIELDS, DONATION_DONOR,
-        showToast, handleError, getRecordFieldNames, setRecordValuesOnTemplate } from 'c/utilTemplateBuilder';
+        showToast,
+         handleError,
```
ptewson-sfdo: Can we fix this one-space shift on these arguments to import? (But thanks for moving these each onto their own line!)

src/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -132,14 +133,9 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
...
-            const fieldMappingApiName = FIELD_MAPPING_METHOD_FIELD_INFO.fieldApiName;
-
-            if (dataImportSettings[fieldMappingApiName] !== ADVANCED_MAPPING) {
-                this.isAccessible = false;
-                this.isLoading = false;
-
-            } else if (dataImportSettings[fieldMappingApiName] === ADVANCED_MAPPING) {
+            this.isAccessible = await getPageAccess();
+            this.isLoading = false;
+         if (this.isAccessible) {
```
ptewson-sfdo: I think this needs to be indented one additional tab

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
```diff
@@ -385,6 +393,20 @@ const setRecordValuesOnTemplate = (templateSections, fieldMappings, record) => {
     return sections;
 };
 
+/*******************************************************************************
+ * @description Method checks for page level access. Currently checks
+ * if Advanced Mapping is on from the Data Import Custom Settings and
+ * if the Gift Entry Feature Gate is turned on.
+ */
+const getPageAccess = async () => {
+    const dataImportSettings = await getDataImportSettings();
```
ptewson-sfdo: I am not sure if I want to block this PR over this, but I wonder if in the future the logic here should be handled in an apex controller method - I'm only thinking of minimizing the number of calls we need to make to the server before we can render the page.

voduyemi: I thought about it earlier on but we didn't really have a unified entry point for Gift Entry, now that we have that entry point `geHome` we can migrate it to be handled from an apex controller and check for access before while we're deciding which view to render. But also i'm not sure what plans we have for the views later on, will they ever become tabs of their own?

ptewson-sfdo: I think they will all be under one tab for the forseeable future, but I think the current implementation is ok for this story - just something to think about for potential future refactor

psadie-sfdo: Just chime in, the first server trip we make in the Gift Entry tab happens in `TemplateBuilderService` it might be worth revisiting that component (rename it to `GiftEntryService`?) and include this call there. I think we might not even want to bother making the trip to `BDI_MappingServiceAdvanced` to collect our maps if these settings aren't on.

ptewson-sfdo: @voduyemi - just chatted with @psadie-sfdo and we agreed that the broader refactor is out of scope for this story - in which case I would just leave this "getPageAccess" method as it currently is and we can approach later, unless you already have work in progress. Thanks!!

src/lwc/geFormRenderer/geFormRenderer.html
```diff
@@ -1,25 +1,26 @@
...
+                        </ul>
+                    </c-util-page-level-message>
+                </div>
+                <template for:each={sections} for:item='section'>
+                    <c-ge-form-section key={section.id} section={section}></c-ge-form-section>
+                </template>
+            </lightning-layout-item>
+        </lightning-layout>
 
     <div if:true={showSpinner} class="spinnerContainer">
```
ptewson-sfdo: I think these next two divs need to be shifted in one additional tab as well

PR #4995, opened 2020-01-24
GAU Allocations Widget

bdvorachek:
# Critical Changes

# Changes
- Added the GAU Widget
# Issues Closed

# Community Ideas Delivered

# New Metadata
- LWC
  - c-collapsible-section
  - ge-form-widget-row-allocation
  - ge-widget-service
- Apex Classes
- Labels
  - geErrorAmountDoesNotMatch
# Deleted Metadata


bdvorachek:
**lurch: attach W-037773

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hp8vQAC" target="_blank">W-037773</a>

src/classes/GE_FormRendererService.cls
```diff
@@ -88,8 +114,7 @@ public with sharing class GE_FormRendererService {
     public static Id saveAndProcessSingleGift(DataImport__c diRecord, String widgetData, Boolean updateGift) {
 
         // Check for widget data and convert it to the correct format for processing
-        String additionalObjectJSON = convertWidgetDataToObjectJSON(widgetData);
-        diRecord.Additional_Object_JSON__c = additionalObjectJSON;
+        convertWidgetDataToObjectJSON(diRecord, widgetData);
```
ptewson-sfdo: It looks like this is only happening in the single gift path, which explains the difficulty of getting the GAUs to save through Batch; it looks like the save paths may be more out of sync between Single and Batch than I had hoped, but one way or another this needs to work for both!

src/lwc/geFormWidgetRowAllocation/geFormWidgetRowAllocation.js
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Calculate what amount is X percent of the total donation amount, rounded down to two decimal places.
+     * @param percentValue
+     * @param total
+     * @return {Number}
+     */
+    calculateAmount(percentValue, total) {
+        const factor = +percentValue / 100;
+        const amountInCents = Math.round((total * factor) * 100);
```
vmartinov-sforg: Are we dividing by 100 and multiplying by 100  in this line? Why do we need the factor variable.

src/lwc/collapsibleSection/collapsibleSection.html
```diff
@@ -0,0 +1,16 @@
+<template>
```
psadie-sfdo: @bdvorachek - would it be worth consolidating this component and `utilExpandableSection` into one collapsible/expandable component? :\

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -6,6 +6,8 @@ import messageLoading from '@salesforce/label/c.labelMessageLoading';
 import geSave from '@salesforce/label/c.labelGeSave';
 import geCancel from '@salesforce/label/c.labelGeCancel';
 import geUpdate from '@salesforce/label/c.commonUpdate';
+import { registerListener } from 'c/pubsubNoPageRef';
+import geLabelService from 'c/geLabelService';
```
voduyemi: @bdvorachek I think we already have the `GeLabelService` imported above, so we won't need this import.

src/lwc/geFormField/geFormField.js
```diff
@@ -28,14 +31,19 @@ export default class GeFormField extends LightningElement {
         this.value = this.getValueFromChangeEvent(event);
         // TODO: The custom event below isn't carrying up any of the assigned properties (field and value)
```
psadie-sfdo: Looks like we can clean this `TODO` comment out now.

src/lwc/geFormWidget/geFormWidget.html
```diff
@@ -7,7 +7,7 @@
         <c-ge-form-field data-id='widgetComponent' element={element}></c-ge-form-field>
     </template>
     <template if:true={isAllocation}>
-        <c-ge-form-widget-allocation data-id='widgetComponent' element={element}></c-ge-form-widget-allocation>
+        <c-ge-form-widget-allocation data-id='widgetComponent' element={element} total-amount={totalAmount}></c-ge-form-widget-allocation>
     </template>
     <!-- Used for testing validation -->
     <!-- <lightning-button label="Check if Valid" onclick={checkValid}></lightning-button> -->
```
psadie-sfdo: Should we clear these commented lines out?

src/classes/GE_FormRendererService.cls
```diff
@@ -121,49 +147,74 @@ public with sharing class GE_FormRendererService {
...
             return null;
         }
 
-        Map<String, Object> objNameToObjList;
-        List<SObject> thisList;
+        BDI_MappingServiceAdvanced mappingService = BDI_MappingServiceAdvanced.getInstance();
+        Map<String,BDI_DynamicSource> dynamicSourceByObjMappingDevName = new Map<String,BDI_DynamicSource>();
+        BDI_DynamicSourceGroup dynSourceGroup = new BDI_DynamicSourceGroup();
+
+        dynSourceGroup.name = 'Additional_Object_JSON__c';
```
ptewson-sfdo:
```suggestion
        dynSourceGroup.name = DataImport__c.Additional_Object_JSON__c.getDescribe().getName();;
```

src/lwc/geFormField/geFormField.js
```diff
@@ -45,11 +53,16 @@ export default class GeFormField extends LightningElement {
         if(this.isRichText) {
             this.checkRichTextValidity();
         }
+
+        if(this.sourceFieldAPIName === DI_DONATION_AMOUNT.fieldApiName) {
+            // fire event for reactive widget component containing the Data Import field API name and Value
+            // currently only used for the Donation Amount.
+            fireEvent(null, 'widgetData', { Donation_Amount__c: this.value });
```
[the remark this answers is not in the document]
bdvorachek: Yeah, that would make sense here. We can call this one whatever we want to though it never gets converted into or from a proper sObject. It's a variable name on a global object that a widget can use to figure out what the entered amount is on the donation.

src/lwc/geFormWidgetRowAllocation/geFormWidgetRowAllocation.js
bdvorachek: Updated this with comments.

src/lwc/collapsibleSection/collapsibleSection.html
bdvorachek: I think we can re-name collapsibleSection to utilExpandableSubSection, as the styling between these two is different. Maybe further down the line we combine the two and allow a "variant" parameter?

bdvorachek: This has been merged into utilExpandableSection, and a few updates were made to that component now.

src/lwc/geFormField/geFormField.js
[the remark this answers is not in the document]
bdvorachek: This line of code isn't affected by namespaces and I think changing it to  `{ donationAmount: this.value }` would be good since Donation_Amount__c is only a variable name/key. Having it named the same as a SFDC Custom Field API Name is kinda misleading.

ptewson-sfdo:
@melissabarber @markross-sfdo - there's probably some docs catch up to do here, the markup for the component has some static text that should be brought in as Labels. For the sake of starting QA on this chunk of work, I merged this in ahead of that, but I'm happy to assist in bringing in those labels when we can!

PR #5000, opened 2020-01-25
Multi-Currency Causing Incorrect GAU Allocations

npsp-reedestockton:
Corrected an issue with multicurrency allocations that caused percentage-based allocations to be calculated incorrectly.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
**lurch: attach W-037009

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HnlPQAS" target="_blank">W-037009</a>

npsp-reedestockton:
@force2b @RadSFDeveloper This is ready for review.  Thanks!

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -554,6 +555,7 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
                 //get wrapper for parent object, preferring recurring donation over campaign allocations if they exist
                 else if (sObj.get('npe03__Recurring_Donation__c') != null && mapWrapper.containsKey((Id)sObj.get('npe03__Recurring_Donation__c')))
                     wrap = mapWrapper.get((Id)sObj.get('npe03__Recurring_Donation__c'));
+
```
force2b: Probably worth adding curly brackets here too

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -538,8 +538,9 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
         for (SObject sObj : listSObjs) {
 
             //if we have recurring donation or campaign allocations
-            if ((sObj instanceof npe01__OppPayment__c && mapWrapper.containsKey((Id)sObj.get('npe01__Opportunity__c'))) || !(sObj instanceof npe01__OppPayment__c) && ((sObj.get('npe03__Recurring_Donation__c') != null && mapWrapper.containsKey((Id)sObj.get('npe03__Recurring_Donation__c'))) ||
-                (sObj.get('CampaignId') != null && mapWrapper.containsKey((Id)sObj.get('CampaignId'))))) {
+            if ((sObj instanceof npe01__OppPayment__c && mapWrapper.containsKey((Id)sObj.get('npe01__Opportunity__c'))) ||
```
force2b: Wow, this method is really hard to read and decipher. In addition to the bug fix, it might be helpful to slightly refactor:

```
    for (SObject sObj : listSObjs) {
        Boolean isPaymentRecord = sObj instanceof npe01__OppPayment__c;
        Boolean isOpportunityRecord = sObj instanceof Opportunity;

        if (isPaymentRecord && mapWrapper.containsKey((Id)sObj.get('npe01__Opportunity__c'))) {
            copyRecurringDonationCampaignAndPaymentAllocations(sObj);

        } else if (isOpportunityRecord && sObj.get('npe03__Recurring_Donation__c') != null && mapWrapper.containsKey((Id)sObj.get('npe03__Recurring_Donation__c')) {
            copyRecurringDonationCampaignAndPaymentAllocations(sObj);

        } else if (!isOpportunityRecord && sObj.get('CampaignId') != null && mapWrapper.containsKey((Id)sObj.get('CampaignId')) {
            copyRecurringDonationCampaignAndPaymentAllocations(sObj);

        } else if (settings.Default_Allocations_Enabled__c) {
            dmlWrapper.objectsToInsert.add(makeDefaultAllocation(sObj)); 
        }
    }

.....

private void  copyRecurringDonationCampaignAndPaymentAllocations(SObject) {
    // copy everything in that if block from line 545 - 680
}
```

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -578,6 +581,7 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
                 //reparent the cloned allocation to the opportunity
                 for (Allocation__c allo : listAlloForInsert) {
                     Allocation__c alloInMap;
+                    Boolean isPercentOnly = false;
```
force2b: Could this just be 
```suggestion
                    Boolean isPercentOnly = allo.Percent__c !=null;
```
and eliminate line 602?

Also, since the var is isPercent**Only**, does this need to check for `allo.Amount__c == null` too?

force2b:
@npsp-reedestockton  Congratulations on having PR #5000!!

src/classes/ALLO_Allocations_TDTM.cls
npsp-reedestockton: I can add that as a fail safe, but you cannot enter both amount and percent from the UI.  Better to be safe, so I will add.

src/classes/ALLO_Allocations_TDTM.cls
npsp-reedestockton: I get where you are headed and will do that, but I believe it needs a little tweaking...

npsp-reedestockton: no tweaking needed .  bad first reading on my part.  Nice!

npsp-reedestockton:
@force2b pretty sure that got all of the suggestions.  Could you take another peek please?

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -536,149 +536,161 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
-                else
-                    wrap = mapWrapper.get((Id)sObj.get('CampaignId'));
+            } else if (settings.Default_Allocations_Enabled__c) {
+                dmlWrapper.objectsToInsert.add(makeDefaultAllocation(sObj));
+            }
+        }
+    }
 
-                List<Allocation__c> listAlloForInsert = new List<Allocation__c>();
+    private void copyRecurringDonationCampaignAndPaymentAllocations(SObject sObj) {
```
force2b: Just add an apexdoc comment here

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -536,149 +536,161 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
-            //if we have recurring donation or campaign allocations
-            if ((sObj instanceof npe01__OppPayment__c && mapWrapper.containsKey((Id)sObj.get('npe01__Opportunity__c'))) || !(sObj instanceof npe01__OppPayment__c) && ((sObj.get('npe03__Recurring_Donation__c') != null && mapWrapper.containsKey((Id)sObj.get('npe03__Recurring_Donation__c'))) ||
-                (sObj.get('CampaignId') != null && mapWrapper.containsKey((Id)sObj.get('CampaignId'))))) {
+            if (isPaymentRecord && mapWrapper.containsKey((Id)sObj.get('npe01__Opportunity__c'))) {
+                copyRecurringDonationCampaignAndPaymentAllocations(sObj);
 
-                alloWrapper wrap;
-                Map<Id, Allocation__c> GAUtoAlloMapping = new Map<Id, Allocation__c>();
-                Boolean isPaymentAllocation = false;
+            } else if (sObj.get('npe03__Recurring_Donation__c') != null && mapWrapper.containsKey((Id)sObj.get('npe03__Recurring_Donation__c'))) {
```
force2b: I think you need to keep the `!isPaymentRecord &&` here and the next if. For example, if this is a Payment Record (and it did not get caught in the first if), then this line will throw an exception because there is no `npe03__Recurring_Donation__c` field no the Payment object.

npsp-reedestockton: Not sure about that.  mapWrapper is Id to alloWrapper and alloWrapper is this...

```
        boolean isPercentOnly = true;                                        //tracks whether all allocations are percentage based
        Decimal parentAmount = null;                                         //amount of this allocation's parent
        Decimal totalAmount = 0;                                             //holds total allocation amount
        Decimal totalPercent = 0;                                            //holds total percentage allocations
        List<Allocation__c> triggerList = new List<Allocation__c>();         //all allocations for this opportunity in the current trigger set
        List<Allocation__c> listAllo = new List<Allocation__c>();            //all non-default allocations for this parent object
        Map<Id, Allocation__c> defaultDupesById = new Map<Id, Allocation__c>();//duplicate default allocations, if any
        Allocation__c defaultAllo = null;                                    //the default allocation for this parent object
        boolean defaultInTrigger = false;                                    //is the default allocation in the current trigger set?

```

Don't see how that ever throw an exception.  fillWrapper contains the same fields for any object, so if we have sobject, we pretty much have to have those fields don't we?

npsp-reedestockton: sorry, you're right.  will fix.

src/classes/ALLO_Allocations_TDTM.cls
force2b: @npsp-reedestockton  I think this is the only remaining change needed.

npsp-reedestockton:
@force2b I think that's it.

npsp-reedestockton:
@heidmary Ready for final testing.

heidmary:
All tests pass.

sam-knox:
@npsp-reedestockton - Is there a Git Issue we can attach this PR to?

heidmary:
@npsp-reedestockton - I think these are the two Git Issues - mind just confirming for @sam-knox ?
https://github.com/SalesforceFoundation/NPSP/issues/4714
https://github.com/SalesforceFoundation/NPSP/issues/4702

And it closed this other WI: [W-036915](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001CCFGQA4/view).

npsp-reedestockton:
@sam-knox @heidmary confirmed.  Thanks, @heidmary !

force2b:
@npsp-reedestockton and @heidmary I added the two fix git issues to the PR comment so they'll be pulled into the Release Notes

PR #5006, opened 2020-01-27
Create a default template if there isn't an existing default template

KietTruongSFDO:
# Critical Changes

# Changes
A default template will be created if there is no matching template in the Gift Entry Custom Setting. Additionally, the Gift Entry Custom Setting will use a record ID as opposed to the record name. 
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


KietTruongSFDO:
**lurch: attach W-037286

KietTruongSFDO:
**lurch: attach W-038656

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HoE8QAK" target="_blank">W-037286</a>

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    ':"RequireTotalMatch__c","dataType":"Boolean","isRequiredFieldDisabled":true,"label":"Require Expected Totals Match","required"' + 
+                                                    ':true}],"layout":{"fieldMappingSetDevName":"Migrated_Custom_Field_Mapping_Set","sections":[{"defaultDisplayMode":"expanded"' + 
+                                                    ',"displayRule":"displayRule","displayType":"accordion","elements":[{"customLabel":"Donor Type","dataImportFieldMappingDevNames"' + 
+                                                    ':["Donation_Donor__c"],"dataType":"PICKLIST","elementType":"field","id":"8ffa6e93-fd7b-2761-2c38-c466f1db98ec","label":' + 
+                                                    '"Data Import: Donation Donor","required":true,"sectionId":"6bc77bf4-9949-5b8b-9f2c-efe5663ef052"},{"customLabel":' + 
+                                                    '"Organization Account (Existing)","dataImportFieldMappingDevNames":["Account1Imported__c"],"dataType":"REFERENCE",' +
```
melissabarber:
```suggestion
                                                    '"Donor Organization Account (Existing)","dataImportFieldMappingDevNames":["Account1Imported__c"],"dataType":"REFERENCE",' + 
```

melissabarber:
```suggestion
                                                    '"Existing Donor Organization Account","dataImportFieldMappingDevNames":["Account1Imported__c"],"dataType":"REFERENCE",' + 
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    ':"Boolean","isRequiredFieldDisabled":true,"label":"Process Using Scheduled Job","required":true},{"allowDefaultValue":true,"apiName"' + 
+                                                    ':"RequireTotalMatch__c","dataType":"Boolean","isRequiredFieldDisabled":true,"label":"Require Expected Totals Match","required"' + 
+                                                    ':true}],"layout":{"fieldMappingSetDevName":"Migrated_Custom_Field_Mapping_Set","sections":[{"defaultDisplayMode":"expanded"' + 
+                                                    ',"displayRule":"displayRule","displayType":"accordion","elements":[{"customLabel":"Donor Type","dataImportFieldMappingDevNames"' + 
+                                                    ':["Donation_Donor__c"],"dataType":"PICKLIST","elementType":"field","id":"8ffa6e93-fd7b-2761-2c38-c466f1db98ec","label":' + 
+                                                    '"Data Import: Donation Donor","required":true,"sectionId":"6bc77bf4-9949-5b8b-9f2c-efe5663ef052"},{"customLabel":' +
```
melissabarber: @KietTruongSFDO Is there a label override missing here? We'd like this field to read Donor Type on form render.

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    ',"displayRule":"displayRule","displayType":"accordion","elements":[{"customLabel":"Donor Type","dataImportFieldMappingDevNames"' + 
+                                                    ':["Donation_Donor__c"],"dataType":"PICKLIST","elementType":"field","id":"8ffa6e93-fd7b-2761-2c38-c466f1db98ec","label":' + 
+                                                    '"Data Import: Donation Donor","required":true,"sectionId":"6bc77bf4-9949-5b8b-9f2c-efe5663ef052"},{"customLabel":' + 
+                                                    '"Organization Account (Existing)","dataImportFieldMappingDevNames":["Account1Imported__c"],"dataType":"REFERENCE",' + 
+                                                    '"elementType":"field","id":"5ed985fd-b631-852f-7d4e-c729db7d151d","label":"Data Import: Account1 Imported","required":' + 
+                                                    'false,"sectionId":"6bc77bf4-9949-5b8b-9f2c-efe5663ef052"},{"customLabel":"Contact (Existing)","dataImportFieldMappingDevNames"' +
```
melissabarber:
```suggestion
                                                    'false,"sectionId":"6bc77bf4-9949-5b8b-9f2c-efe5663ef052"},{"customLabel":"Existing Donor Contact","dataImportFieldMappingDevNames"' + 
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    '["Account1_Zip_Postal_Code_4aea9a02f"],"dataType":"STRING","elementType":"field","id":"56f7dbf7-99dc-e9be-d5d8-21df79ba1621","label"' + 
+                                                    ':"Account 1: Billing Zip/Postal Code","required":false,"sectionId":"94d2bb2b-8cc4-3644-6ac6-d9180269c95c"},{"customLabel":' + 
+                                                    '"Organization Account Country","dataImportFieldMappingDevNames":["Account1_Country_1359e000c"],"dataType":"STRING","elementType":' + 
+                                                    '"field","id":"e8a74c72-6a9a-66da-3129-f48cca998fdc","label":"Account 1: Billing Country","required":false,"sectionId":"94d2bb2b-8cc4-3644-6ac6-d9180269c95c"}],' + 
+                                                    '"id":"94d2bb2b-8cc4-3644-6ac6-d9180269c95c","label":"Add or Edit Organization Account"},{"id":"1cb17425-5ced-ebd6-cdaf-44556f648e3f",' + 
+                                                    '"displayType":"accordion","defaultDisplayMode":"expanded","displayRule":"displayRule","label":"Add or Edit Primary Contact","elements":' +
```
melissabarber:
```suggestion
                                                    '"displayType":"accordion","defaultDisplayMode":"expanded","displayRule":"displayRule","label":"Add or Edit Contact","elements":' + 
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    '[{"id":"05724374-e3fa-b94b-fc6d-00dfbf358abb","label":"Contact 1: First Name","customLabel":"Contact First Name","required":false,"sectionId"' + 
+                                                    ':"1cb17425-5ced-ebd6-cdaf-44556f648e3f","defaultValue":null,"dataType":"STRING","dataImportFieldMappingDevNames":["Contact1_First_Name_73301c3c5"]' + 
+                                                    ',"elementType":"field"},{"id":"d89dbbb6-1a87-41e6-6e14-dd34a13a9274","label":"Contact 1: Last Name","customLabel":"Contact Last Name","required":false,' + 
+                                                    '"sectionId":"1cb17425-5ced-ebd6-cdaf-44556f648e3f","defaultValue":null,"dataType":"STRING","dataImportFieldMappingDevNames":["Contact1_Last_Name_563ada236"]' + 
+                                                    ',"elementType":"field"},{"id":"39051aa1-b149-6df7-6ced-02639bdacd82","label":"Address: Mailing Street","customLabel":"Contact Street",' +
```
melissabarber: @KietTruongSFDO @ptewson-sfdo @vmartinov-sforg We need clarity from PMs on how to handle the Contact address fields on the default template. These fields map to the NPSP Address object, and GEM users use the EDA Address object.

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    ':["Payment_Method_449599d68"],"dataType":"PICKLIST","elementType":"field","id":"149fb6d3-4a3a-c387-6d2c-1a77dcd14f7f","label":"Payment: Payment Method",' + 
+                                                    '"required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"},{"customLabel":"Check/Reference Number","dataImportFieldMappingDevNames":' + 
+                                                    '["Payment_Check_Reference_Number_2947b94f7"],"dataType":"STRING","elementType":"field","id":"cc94d1ce-d93b-dc47-7703-cd89c27843ce","label":' + 
+                                                    '"Payment: Check/Reference Number","required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"},{"componentName":"geFormWidgetAllocation"' + 
+                                                    ',"elementType":"widget","id":"7019830b-8026-5eac-b65e-0f74a3e5382f","label":"Allocations","required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"}]' +
```
melissabarber:
```suggestion
                                                    ',"elementType":"widget","id":"7019830b-8026-5eac-b65e-0f74a3e5382f","label":"GAU Allocations","required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"}]' + 
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                                                    ':["Payment_Method_449599d68"],"dataType":"PICKLIST","elementType":"field","id":"149fb6d3-4a3a-c387-6d2c-1a77dcd14f7f","label":"Payment: Payment Method",' + 
+                                                    '"required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"},{"customLabel":"Check/Reference Number","dataImportFieldMappingDevNames":' + 
+                                                    '["Payment_Check_Reference_Number_2947b94f7"],"dataType":"STRING","elementType":"field","id":"cc94d1ce-d93b-dc47-7703-cd89c27843ce","label":' + 
+                                                    '"Payment: Check/Reference Number","required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"},{"componentName":"geFormWidgetAllocation"' + 
+                                                    ',"elementType":"widget","id":"7019830b-8026-5eac-b65e-0f74a3e5382f","label":"Allocations","required":false,"sectionId":"3c42b004-03d7-90d5-98ed-e04f25af89e3"}]' + 
+                                                    ',"id":"3c42b004-03d7-90d5-98ed-e04f25af89e3","label":"Donation Information"}]},"name":"Default Gift Entry Form"}';
```
melissabarber: @KietTruongSFDO I didn't find anything for Primary Campaign Source in this. Does it need to be added, or did I just miss it?

KietTruongSFDO: @melissabarber The Primary Campaign Source isn't there. It wasn't on the screenshots, should it be there?

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,234 @@
...
+                    // that is on the DI itself and not a mapped field.
+                    // e.g. 'Donation_Donor__c', 'Account1Imported__c', etc
+                    if (updatedDevNameMap.containsKey(partialDevName)) {
+                        element.dataImportFieldMappingDevNames.set(0, updatedDevNameMap.get(partialDevName));
+                    }
+                }
+            }
+        }
+
+        newDefaultTemplate.Name = 'Default Template';
```
melissabarber:
```suggestion
        newDefaultTemplate.Name = 'Default Gift Entry Template';
```

src/classes/GE_DefaultTemplate.cls
ptewson-sfdo: Thanks @melissabarber - keep us posted, but in the interest of keeping this story moving, I think we need to deliver it as it was spelled out in the requirements while that conversation occurs.

melissabarber: We have a meeting tomorrow to discuss. I suspect we're going with keeping the fields in for MVP and hopefully somewhat soon we're able to get farther into Mapping Sets that could provide a solution for GEM customers.

src/aura/GE_GiftEntryForm/GE_GiftEntryFormController.js
```diff
@@ -1,4 +1,33 @@
...
+            }
+            else if (state === "INCOMPLETE") {
+                // do something
+            }
+            else if (state === "ERROR") {
+                var errors = response.getError();
+                
+                if (errors) {
+                    if (errors[0] && errors[0].message) {
+                        console.log("Error message: " +
```
ptewson-sfdo: We may need to think of a way to display this error, and we should remove the console.log

src/aura/GE_GiftEntryForm/GE_GiftEntryFormController.js
```diff
@@ -1,4 +1,33 @@
...
+            else if (state === "ERROR") {
+                var errors = response.getError();
+                
+                if (errors) {
+                    if (errors[0] && errors[0].message) {
+                        console.log("Error message: " + 
+                                 errors[0].message);
+                    }
+                } else {
+                    console.log("Unknown error");
```
ptewson-sfdo: ... and also remove this console.log

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+
+public with sharing class GE_DefaultTemplate {
+    static final String ELEMENT_TYPE_FIELD = 'field';
+
+    public GE_DefaultTemplate() {
+
+    }
+
+    /**
+     * @description This method determines if there is a default template, if not, it will create a new template and update the custom setting with
```
ptewson-sfdo: Can you fix the line breaks in this comment so that the lines don't wrap (I think its 110 chars?)

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @return Returns whether or not there is an existing default template. 
+     */
+    private static boolean hasDefaultTemplate() {
+            
+        // Get the gift entry custom setting to check for the default template.             
+        Gift_Entry_Settings__c giftEntryCustomSetting = UTIL_CustomSettingsFacade.getGiftEntrySettings();
+        Boolean hasDefaultTemplate = false;
+
+        if (giftEntryCustomSetting.Default_Single_Gift_Entry_Template__c != null && giftEntryCustomSetting.Default_Single_Gift_Entry_Template__c != '') {
```
ptewson-sfdo: Add line breaks to keep the line from wrapping

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+            
+            // If there is an ID already stored, try to get this particular template. 
+            List<Form_Template__c> templateList = [SELECT Id, Name 
+                                                    FROM Form_Template__c 
+                                                    WHERE Id =: giftEntryCustomSetting.Default_Single_Gift_Entry_Template__c];
```
ptewson-sfdo: Add line breaks to keep the line from wrapping

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+            if (!templateList.isEmpty()){ 
+                hasDefaultTemplate = true;
+            }
+        }
+
+        return hasDefaultTemplate;
+    }
+
+    /**
+     * @description This method builds a new default template using the base JSON string. It will update the base JSON string with the updated field mapping dev names.
```
ptewson-sfdo: Add line breaks to keep the line from wrapping

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+     *   
+     *
+     * @return A new Form template that is ready to be saved into the DB and used. 
+     */
+    public static Form_Template__c buildDefaultTemplate() {
+        FORM_Template__c newDefaultTemplate = new FORM_Template__c();
+        List<StaticResource> defaultTemplateSRList = [Select Id, Body FROM StaticResource Where Name = 'DefaultGiftEntryTemplate' limit 1];
```
ptewson-sfdo: Add line breaks to keep the line from wrapping

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+
+        newDefaultTemplate.Name = defaultTemplate.name;
+        newDefaultTemplate.Description__c = defaultTemplate.description;
+        newDefaultTemplate.Template_JSON__c = JSON.serialize(defaultTemplate);
+        newDefaultTemplate.Format_Version__c = '1.0';
+        return newDefaultTemplate;
+    }
+
+    /**
+     * @description This method builds a mapping of partial dev name to the full dev name using the org's field mapping custom metadata records.
```
ptewson-sfdo: Add line breaks to keep the line from wrapping

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * @description This method builds a mapping of partial dev name to the full dev name using the org's field mapping custom metadata records. 
+     *
+     * @return A mapping of partial dev name to full dev name. e.g. Account1_Name -> Account1_Name_ee63cf9df
+     */
+    public static Map<String, String> getCurrentDevNameMapping() {
+        // Get the org's migrated field mappings. 
+        List<Data_Import_Field_Mapping__mdt> fieldMappingCMTList = [Select Id, MasterLabel, DeveloperName, Target_Field_API_Name__c, Source_Field_API_Name__c, Target_Object_Mapping__r.Object_API_Name__c
```
ptewson-sfdo: Can you adjust the format of this SOQL query (capitalize SELECT and WHERE and break each field onto its own line)? There should be good examples to draw from elsewhere

src/classes/GE_DefaultTemplate_TEST.cls
```diff
@@ -0,0 +1,51 @@
...
+    /*******************************************************************************************************
+     * @description Confirm that a default template is able to be deserialized to the current FORM_Layout structure.
+     * This is the best we can do since there will not be a migrated field set available for testing.  
+     */
+    @isTest
+    private static void testDeserializeBaseJSONToCurrentFORMLayout() {
+        try {
+            GE_DefaultTemplate.buildDefaultTemplate();
+        } catch (Exception e) {
+            System.Assert(false, 'There was an error deserializing the Base JSON to the current FORM_Layout structure');
```
ptewson-sfdo: Add line break to avoid wrapping the line

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+* @date 2019
+* @group BGE
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Class to create a default template when needed
+*/
+
+public with sharing class GE_DefaultTemplate {
+    static final String ELEMENT_TYPE_FIELD = 'field';
+
+    public GE_DefaultTemplate() {
```
ptewson-sfdo: Remove empty constructor

src/classes/GE_DefaultTemplate.cls
ptewson-sfdo: And adjust SOQL format

src/aura/GE_GiftEntry/GE_GiftEntryController.js
```diff
@@ -1,4 +1,35 @@
 ({
+    
+    doInit: function(cmp) {
+        var action = cmp.get("c.createDefaultTemplate");
```
ptewson-sfdo: This is probably late in the game to share this thought, but what if we moved this call into the error handler for where we try to load the default template (on SGE). That way we could avoid the querying etc until we know we actually need it. (This probably contradicts my earlier thoughts on how we deliver this, so maybe this isn't a critical change at this stage)

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+
+            if (batchHeaderField.apiName.endsWith('__c')) {
+                batchHeaderField.apiName = UTIL_Namespace.StrTokenNSPrefix(batchHeaderField.apiName);
+            }
+        }
+
+        newDefaultTemplate.Name = defaultTemplate.name;
+        newDefaultTemplate.Description__c = defaultTemplate.description;
+        newDefaultTemplate.Template_JSON__c = JSON.serialize(defaultTemplate);
+        newDefaultTemplate.Format_Version__c = '1.0';
```
vmartinov-sforg: Let's define a constant for '1.0' in case we'll use a different value in the future.

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+/**
+* @author Salesforce.org
+* @date 2019
+* @group BGE
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Class to create a default template when needed
+*/
+
+public with sharing class GE_DefaultTemplate {
+    static final String ELEMENT_TYPE_FIELD = 'field';
```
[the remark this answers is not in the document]
KietTruongSFDO: I added it into the FORM_PermissionValidator to fix a bug. I could use that one, but it doesn't quite seem like the right place to store a 'universal' element type variable.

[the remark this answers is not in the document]
vmartinov-sforg: You can leave it alone then, maybe we should start a UTIL_GiftEntry.cls.

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,191 @@
...
+    static final String TEMPLATE_VERSION = '1.0';
+    
+    /**
+     * @description This method determines if there is a default template, 
+     * if not, it will create a new template and update the custom setting with 
+     * the newly created template's ID as the default template. 
+     *
+     */
+    @auraEnabled
```
kenlewis-sfdo:
```suggestion
    @AuraEnabled
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,191 @@
...
+    public static void createDefaultTemplate() {
+        if (!hasDefaultTemplate()) {
```
kenlewis-sfdo: Code inside an `@AuraEnabled` method should be wrapped in a try/catch block and return errors in an `AuraHandledException`.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -11,6 +12,10 @@ export default class GeGiftEntryFormApp extends LightningElement {
     @track errorTitle;
     @track errorMessage;
 
+    async connectedCallback() {
+        createDefaultTemplate();
+    }
```
kenlewis-sfdo: Along the lines of @ptewson-sfdo's comment about timing for calling this method, instead of calling it imperatively every time the component is loaded, I wonder if it would make sense to put this inside the `GE_FormRendererService.getDefaultSGERenderWrapper` method.  If no default template is returned from `formService.retrieveDefaultSGEFormTemplate()`, then the `GE_DefaultTemplate.createDefaultTemplate` method could be called there, which would mean a default template would always be returned from `GE_FormRendererService.getDefaultSGERenderWrapper`, and there wouldn't be any additional Apex calls.

src/lwc/geHome/geHome.js
```diff
@@ -36,6 +37,7 @@ export default class geHome extends LightningElement {
         this.isAccessible = await getPageAccess();
         if (this.isAccessible) {
             this.isLoading = true;
+            createDefaultTemplate();
```
kenlewis-sfdo: Should this be called from both geHome _and_ geGiftEntryFormApp?  Since the geGiftEntryFormApp is always used to render the form, it should only be needed there - but is there a use case for having it in both that I am missing?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
ptewson-sfdo: I keep waffling about this, but I THINK this actually makes sense. The only other thought is that we might want a second catch for it - if you are loading your template list and get 0 results? Otherwise I think we might miss a path where you are only using Batch mode and may not have any templates at all

src/classes/FORM_PermissionsValidator_TEST.cls
```diff
@@ -90,10 +90,11 @@ private with sharing class FORM_PermissionsValidator_TEST {
         System.assert(updatedTemplate.permissionErrors.contains('Account'), 'The '
             + 'permissions error should contain the Account object, which the user does not '
             + 'have access to.'); 
-
+/*      // Should we be checking the widget permissions right now??
```
KietTruongSFDO: @vmartinov-sforg @dfuller114 Should we be checking the widget permissions here and below? It doesn't look like it is currently set up properly yet. 

cc: @ptewson-sfdo

src/classes/FORM_ServiceGiftEntry_TEST.cls
```diff
@@ -89,12 +89,11 @@ public with sharing class FORM_ServiceGiftEntry_TEST {
 
         FORM_Template template = UTIL_UnitTestData_TEST.createSampleTemplate();
         Gift_Entry_Settings__c giftEntryCustomSetting = new Gift_Entry_Settings__c();
-        giftEntryCustomSetting.Default_Single_Gift_Entry_Template__c = template.name;
-
-        UTIL_CustomSettingsFacade.getGiftEntrySettingsForTests(giftEntryCustomSetting);
 
         String templateJSON = JSON.serialize(template);
-        fs.storeFormTemplate(null, template.name, template.description, template.version, templateJSON);
+        String storedFormTemplateId = FORM_ServiceGiftEntry.storeFormTemplate(null, template.name, template.description, template.version, templateJSON);
```
KietTruongSFDO: @dfuller114 I had to update your test class to work with the new changes.

cc: @vmartinov-sforg

src/lwc/geHome/geHome.js
ptewson-sfdo: @kenlewis-sfdo I think that if you have no templates created but go straight to BGE, you won't have any templates (and we would like you to have the default template as an option no matter what). However, we COULD try and only create that default if you attempt to pull your template list and get 0 results (or a better way of accomplishing this?)

KietTruongSFDO: I believe it should be called from both because there's situations where you can get to geGiftEntryFormApp without going through the geHome.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
KietTruongSFDO: I think adding an extra call at `retrieveDefaultSGEFormTemplate` makes sense, but I don't think that would replace having the call here. I don't think batch situations are covered as mentioned by Patrick. Maybe it makes sense to do the whole create default template process at the beginning of 'retrieveDefaultSGEFormTemplate' so that there will always be a default template, in addition to keeping this call here?

src/classes/FORM_PermissionsValidator_TEST.cls
ptewson-sfdo: I think our work in progress on the widget front will actually take care of this, or at least set us up to do so. For now, let's not worry about it until a widget has arrived

force2b: Please be careful with looking for specific english terms in error messages since the word "Account" would likely be something different in other languages. I think alternatively the code could use something like `Account.SObjectType.getDescribe().getLabel()` to retrieve the localized label name for the Account object so the assert works in any language.

ptewson-sfdo: @KietTruongSFDO Mike speaks the truth - we will need that change else this test will fail in translated orgs! I will suggest a change on that line.

src/classes/FORM_PermissionsValidator_TEST.cls
```diff
@@ -90,10 +90,11 @@ private with sharing class FORM_PermissionsValidator_TEST {
         System.assert(updatedTemplate.permissionErrors.contains('Account'), 'The '
```
ptewson-sfdo:
```suggestion
        System.assert(updatedTemplate.permissionErrors.contains(Account.SObjectType.getDescribe().getLabel()),
         'The '
```

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
+/*
+    Copyright (c) 2019, Salesforce.org
```
kenlewis-sfdo: Should this be bumped to 2020?

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
```
kenlewis-sfdo: Same here should this be bumped?

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * @description This method builds a mapping of partial dev name to 
+     * the full dev name using the org's field mapping custom metadata records. 
+     *
+     * @return A mapping of partial dev name to full dev name. 
+     * e.g. Account1_Name -> Account1_Name_ee63cf9df
+     */
+    public static Map<String, String> getCurrentDevNameMapping() {
+        // Get the org's migrated field mappings. 
+        List<Data_Import_Field_Mapping__mdt> fieldMappingCMTList = [SELECT Id, MasterLabel, DeveloperName, Target_Field_API_Name__c, Source_Field_API_Name__c, Target_Object_Mapping__r.Object_API_Name__c
```
kenlewis-sfdo: Might be good to format this query to wrap it so the lines aren't so wide.

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -928,7 +927,7 @@ public without sharing class UTIL_CustomSettingsFacade {
             giftEntrySettings = new Gift_Entry_SEttings__c();
```
kenlewis-sfdo:
```suggestion
            giftEntrySettings = new Gift_Entry_Settings__c();
```

src/classes/FORM_PermissionsValidator_TEST.cls
KietTruongSFDO: I've updated this line and a couple others to use the object/field describe instead of a hardcoded string.

src/classes/GE_DefaultTemplate.cls
```diff
@@ -0,0 +1,190 @@
...
+                                                                    FROM Data_Import_Field_Mapping__mdt 
+                                                                    WHERE Data_Import_Field_Mapping_Set__r.DeveloperName = 'Migrated_Custom_Field_Mapping_Set' Limit 1000];
```
ptewson-sfdo: Can we use "Migrated_Custom_Field_Mapping_Set" as a defined constant? I feel like this already exists somewhere

src/classes/FORM_PermissionsValidator_TEST.cls
ptewson-sfdo: Thanks @KietTruongSFDO !

PR #5017, opened 2020-01-29
Handle widgets when building query string for batch mode UI

kenlewis-sfdo:
Widgets might reference multiple field mappings.  This change handles
that scenario so that all source fields used by the widget are included
in the query used in the form UI.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BGE_BatchGiftEntry_UTIL_TEST.cls
+++ b/src/classes/BGE_BatchGiftEntry_UTIL_TEST.cls
@@ -86,4 +86,55 @@ private with sharing class BGE_BatchGiftEntry_UTIL_TEST {
         }
     }
 
+    @isTest
+    static void shouldIncludeWidgetFields() {
+        FORM_Element widgetElement =
+                new FORM_Element(
+                        'widget', '', '', '',
+                        new String[]{
+                                UTIL_Namespace.alignSchemaNSWithEnvironment(
+                                        'Account_2_Country'),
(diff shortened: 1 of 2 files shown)


src/classes/BGE_BatchGiftEntry_UTIL_TEST.cls
```diff
@@ -86,4 +86,55 @@ private with sharing class BGE_BatchGiftEntry_UTIL_TEST {
...
     }
 
+    @isTest
+    static void shouldIncludeWidgetFields() {
+        FORM_Element widgetElement =
+                new FORM_Element(
+                        'widget', '', '', '',
+                        new String[]{
+                                UTIL_Namespace.alignSchemaNSWithEnvironment(
+                                        'Account_2_Country'),
```
kenlewis-sfdo: @psadie-sfdo Hey Prince, let me know if this looks right to you.  I can't remember if the field mappings get namespace pre-pended in namespaced scratches, but if so, let me know if you think this looks correct.

psadie-sfdo: @kenlewis-sfdo - No namespacing needed for the field mapping developer names. The metadata types and their fields themselves get namespaced, but not the value inside fields like `DeveloperName` in this case.

kenlewis-sfdo: Great - thank you!!

PR #5026, opened 2020-01-29
Gift Entry: Display validation rules in SGE and BGE

grivero:
[**W-037199:**](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001Ho5kQAC/view)
- Display validation rules in Gift Entry (SGE and BGE)

# Critical Changes

# Changes
- Display validation rules in Gift Entry (SGE and BGE)
  - move saveAndDryRunRow from BGE to GE class
  - change in exception handling for `GE_BatchGiftEntryController.saveAndDryRunRow` 
  - `handleSave()` method refactor for `geFormRenderer` cmp 
- Create unit test class for `GE_BatchGiftEntryController`

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Apex Class
  - GE_BatchGiftEntryController_TEST

# Deleted Metadata


grivero:
**lurch: attach W-037199

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Ho5kQAC" target="_blank">W-037199</a>

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -207,125 +207,159 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+            toggle();
+            handleCatchError(error);
+        });
+
+    }
+
+    handleSaveBatchGiftEntry(sectionsList,enableSave,toggle) {
+
+        // reset function for callback
+        const reset = () => this.reset();
```
ptewson-sfdo: Previously it looks like we were resetting the form for both Single and Batch gift entry - with this change it's only happening for batch. Is that intentional?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -27,8 +26,7 @@ export default class GeGiftEntryFormApp extends LightningElement {
                 }
             )
             .catch(error => {
-                handleError(error);
```
ptewson-sfdo: What's the consequence of not calling the utilTemplateBuilder handle error here? Just want to make sure we don't have a broader, unintentional change to error handling behavior

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -329,15 +329,25 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
             } else {
                 Database.insert(dataImport);
             }
+
             return runSingleDryRun(dataImport.Id, batchId);
+
         } catch (Exception ex) {
-            throw new AuraHandledException(ex.getMessage());
+
+            // create custom lwc exception and throw it
```
ptewson-sfdo: @kenlewis-sfdo - if you get a chance, maybe just be a second set of eyes here to make sure we aren't causing negative consequence to our legacy BGE product with this update?

src/lwc/geFormRenderer/geFormRenderer.js
grivero: yes, previously It was only being used for BGE mode

src/classes/BGE_DataImportBatchEntry_CTRL.cls
grivero: This was needed if we wanted to manage the exception in the same way as SGE, excellent catch Patrick. Hopefully didn't broke existent functionality

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
grivero: That method was causing an issue since the error was not in the expected format

grivero: Worth to mention that the issue existed before I changed the exception handling on BGE_DataImportBatchEntry_CTRL class

src/classes/BGE_DataImportBatchEntry_CTRL.cls
kenlewis-sfdo: This should be ok because this is a new method that is only used for the new GE.  This method should probably be located in `GE_BatchGiftEntryController`.  If you guys agree and it isn't too much work maybe we could move it there as part of this PR.

ptewson-sfdo: I think that makes sense to move this method to GE_BatchGiftEntryController - that's been throwing me because I assume everything in this class is legacy! @grivero do you think you can move this, and then we should be good to go!

grivero: Thanks for reviewing @kenlewis-sfdo. 
Regarding moving method: If that's the way to go I'll move that and any related Unit Tests to `GE_BatchGiftEntryController` right after our Sprint Review meeting

vmartinov-sforg:
The refactoring looks good from what I can tell, but I found this functional issue: create a new template with the default fields, go to the landing page, click new single gift, and try and save without entering any data. You get this:
Screen Shot 2020-01-30 at 6 26 02 PM

src/classes/GE_BatchGiftEntryController_TEST.cls
```diff
@@ -0,0 +1,70 @@
+@isTest
+private with sharing class GE_BatchGiftEntryController_TEST {
+
+    /*********************************************************************************************************
+    * @description Sets up a new NPSP_Data_Import_Batch__c Record
+    * @return Id of the new NPSP_Data_Import_Batch__c
+    */
+    private static Id createDataImportBatch( String batchName ) {
```
grivero: I saw similar methods being created in other Tests, but I didn't want to introduce circular dependencies so I created the ones I needed for my tests.
 
I think we should have a single unit test helper in charge of creating ALL the GiftEntry assets (di, batches, etc) or at least to act as a hub and only call that class from our unit tests. But that's require a lot of change at this moment

src/classes/GE_BatchGiftEntryController_TEST.cls
```diff
@@ -0,0 +1,70 @@
...
+                NPSP_Data_Import_Batch__c = batchId
+        );
+    }
+
+    /*******************************************************************************************************
+     * @description Saves the data import record, dry runs the record, and returns updated
+     * totals with the new row.
+     */
+    @isTest
+    private static void saveAndDryRunRow_test() {
```
grivero: I didn't create a method to get the exception that this method might return (when catched) -- hopefully is already being tested somewhere else as Ken mentioned. IF we need additional tests please let me know

src/classes/GE_BatchGiftEntryController_TEST.cls
ptewson-sfdo: Yes, good points - I generally agree, although we should also think about when and if we should build our tests in a way that is less dependent on data. It's less intuitive but having been around this product for a year I can really see the advantages to test stability and execution time. I think this is probably good for this delivery!

src/classes/GE_BatchGiftEntryController_TEST.cls
```diff
@@ -0,0 +1,70 @@
...
+        Test.startTest();
+        GE_BatchGiftEntryController.saveAndDryRunRow(batchId,di);
+        Test.stopTest();
+
+        List<DataImportBatch__c> result = [SELECT Id FROM DataImportBatch__c];
+        System.assertEquals( result.size(), 1 );
+
+    }
+
+    /*******************************************************************************************************
```
ptewson-sfdo: I think we can do without this last unit test in its current form - I vote for removing alignClassNSWithEnvironment_test() until we get something more complete (although a good unit test of this method would help reduce ongoing confusion about exactly what it does, which persists).

src/classes/GE_BatchGiftEntryController_TEST.cls
ptewson-sfdo:
```suggestion
    private static void testShouldSaveAndDryRunRow() {
```

src/classes/GE_BatchGiftEntryController_TEST.cls
grivero: gotcha going to remove test

vmartinov-sforg:
If @kenlewis-sfdo agrees, I say we merge it.

kenlewis-sfdo:
Thanks for this work, Gus!

PR #5067, opened 2020-02-06
RD2: Keep Opportunity Currency in sync

RadSFDeveloper:
Ensure Opportunities attached to a Recurring Donation have the same currency as the Recurring Donation.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Label:
- RD2_CurrencyChangeIsRestrictedOnOpp
- RD2_CurrencyChangeIsRestrictedOnRD

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-024076

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x8000000AVZ8EAAX" target="_blank">W-024076</a>

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -265,22 +265,26 @@ public inherited sharing class RD2_OpportunityEvaluationService {
             if (rdRecord.isClosed()) {
                 closedRDs.add(rd);
 
-            } else if (isNewInstallmentCandidate(rdRecord)) {
-                newOppRDs.add(rd);
-
-            } else if (isUpdateContext(rdRecord)) {
-                updateOppRDs.add(rd);
+            } else {
+                if (isNewOpportunityCandidate(rdRecord)) {
```
RadSFDeveloper: If there is an RD that does not have a future Opp but it has a past Opp, a new Opp will be created (if all conditions align for a new record) and the past Opp will be verified to ensure currency is matching, so there can be a scenario where both Opp insert and update will be done for the same RD.

src/classes/RD2_OpportunityService.cls
```diff
@@ -185,35 +203,35 @@ public with sharing class RD2_OpportunityService {
...
         List<Opportunity> oppsToUpdate = new List<Opportunity>();
         Map<Id, RD2_ScheduleService.Installment> installmentByOppId = new Map<Id, RD2_ScheduleService.Installment>();
 
         //Synchronize Opps with matching Schedule installments
         for (npe03__Recurring_Donation__c rd : rds) {
             RD2_OpportunityMatcher matcher = new RD2_OpportunityMatcher(currentDate)
-                .filterCurrent(rd.npe03__Donations__r);
-
-            List<RD2_ScheduleService.Installment> installments = scheduleService.getVisualizedInstallments(
+                .filterCurrentOrOpen(rd.npe03__Donations__r);
```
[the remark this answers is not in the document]
force2b: When you say "the past opps will be updated ...", I assume this means only past open opportunities? Anything Closed Won or Closed Lost is never modified.

src/classes/RD2_OpportunityService.cls
```diff
@@ -222,53 +240,102 @@ public with sharing class RD2_OpportunityService {
...
+    private Boolean syncOppCurrency(Opportunity opp, npe03__Recurring_Donation__c rd) {
+        if (!isMultiCurrencyEnabled) {
+            return false;
         }
 
-        if (isChanged) {
-            opp.name = buildOpportunityName(rd.npe03__Donor_Name__c, opp);
+        String rdCurrency = (String) rd.get(CURRENCY_ISO_CODE_FIELD);
+        if ((String) opp.get(CURRENCY_ISO_CODE_FIELD) != rdCurrency) {
+            opp.put(CURRENCY_ISO_CODE_FIELD, rdCurrency);
```
force2b: If I'm not mistaken, there's an issue with how Salesforce handles this kind of update. If the CurrencyCode is changed it will convert the Amount (i.e. change it). The way we handled this on the Allocation and Payment objects was to always update the CurrencyCode and the Amount together. That doesn't have to happen in this method, but it might be necessary to ensure that the Amount is updated on the Opp even if technically the Amount value wasn't changed on the RD. Not sure I'll have time to test this in the UI today to see if happens here though.

src/classes/RD2_RecurringDonationsOpp_TDTM.cls
```diff
@@ -75,11 +75,16 @@ public class RD2_RecurringDonationsOpp_TDTM extends TDTM_Runnable {
         }
 
         Set<Id> rdIds = new Set<Id>();
```
force2b: Since this trigger will get every single new Opportunity record, what would you think about first checking the trigger.new collection to see if there are any records with a value in the `npe03__Recurring_Donation__c` field before calling any of the other methods? I'm mostly thinking about reducing any additional processing in an already crowded trigger framework for Opps, where generally only a small fraction of the opps in an org will actually be linked to an RD record.

src/classes/RD2_OpportunityMatcher.cls
```diff
@@ -132,6 +132,22 @@ public inherited sharing class RD2_OpportunityMatcher {
...
     }
 
+    /***
+     * @description Filters open Opportunities regardless of the Close Date and
+     * open/closed Opportunities with Close Date greater or equal to the current date.
+     * Opportunities marked as Closed Lost in data migration are ignored.
+     * @param opps Recurring Donation Opportunities
+     * @return RD2_OpportunityMatcher This Opportunity matcher instance
+     */
+    public RD2_OpportunityMatcher filterCurrentOrOpen(List<Opportunity> opps) {
```
force2b: I'm not particularly clear on the use of the word "filter" through this code. I generally take that to mean "filter out" (exclude) records. However, I think that the intent of this is to choose which recors to include. If that's the case, I think it might be a good idea to replace all uses of `filter` to `include` or `exclude` depending on the intent of the method.

src/classes/PMT_Payment_TDTM.cls
```diff
@@ -150,25 +150,39 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
             'WHERE npe01__Opportunity__c IN :oppIds ' +
             'ORDER BY npe01__Opportunity__c';
+
         for (npe01__OppPayment__c pmt : database.query(soql)) {
             Opportunity opp = (Opportunity)pmt.getSObject('npe01__Opportunity__r');
-            pmt.put('CurrencyIsoCode', opp.get('CurrencyIsoCode'));
-            pmt.put('npe01__Payment_Amount__c', pmt.npe01__Payment_Amount__c);
-            dmlWrapper.objectsToUpdate.add(pmt);
+
+            if (dmlWrapperSObjectById.containsKey(pmt.Id)) {
```
RadSFDeveloper: When both currency and Amount are changed on RD, then "duplicate id in list" error is generated in the `PMT_Payment_TDTM` trigger handler. Thus, this change is verifying if `dmlWrapper` has the payment already. The original change for this part of code was done in PR #2710 (see [comment](https://github.com/SalesforceFoundation/NPSP/pull/2710#discussion_r135633410)).

If the payment is already in the `dmlWrapper`, then the payment currency is updated only. If the Amount is updated from the payment as it is currently in DB, then the Amount is wrong (`RD2_OpportunityService_TEST.shouldAllowCurrencyChangeWhenRDHasNoClosedOpp()` test would fail).

src/classes/RD2_OpportunityService.cls
RadSFDeveloper: I cannot reproduce the issue via unit tests or UI. Can you please let me know details.

RadSFDeveloper: @Julian88Tex , @ravensisco  ^

src/classes/PMT_Payment_TDTM.cls
RadSFDeveloper: @Julian88Tex , @ravensisco  ^

src/labels/CustomLabels.labels
```diff
@@ -1494,45 +1494,37 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
...
-        <fullName>RD2_ScheduleLWCTitle</fullName>
-        <categories>Recurring-Donations, Component</categories>
-        <language>en_US</language>
-        <protected>true</protected>
-        <shortDescription>Title of the Active Schedules Component</shortDescription>
-        <value>Active Schedules</value>
-    </labels>
-     <labels>
-        <fullName>RD2_AmountMustBeValid</fullName>
+        <fullName>RD2_CurrencyChangeIsRestrictedOnOpp</fullName>
```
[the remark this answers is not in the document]
RadSFDeveloper: @sam-knox , the same message is displayed with user creates an Opp manually and does not set the currency to be the same as the RD currency. Is "You can't change the Opportunity currency to a different currency than the related Recurring Donation." sufficient?

src/classes/PMT_Payment_TDTM.cls
force2b: FYI @jbrandolinisf @lparrott

src/classes/RD2_OpportunityMatcher.cls
force2b: Thanks for the clarification. Personally I do find it confusing. For example, the test method `shouldFilterAllOppsWhenCloseDateYearMatchesCurrentYear` reads to me as "it should filter out all close date matches the current year". Same with this method. I read that as "filter out current or open opps". I think of a filter as something that removes items, so anything listed after the word filter (to me) is filtered out. 

For this method, I think that either `includeCurrentOrOpenOpps` or `filterToCurrentOrOpen` would work better (the latter adding "to" so it's explicitly stating the inclusion vs. the default exclusion). It's minor, but I think long term it'll be easier to read back in 6 months and know exactly what's happening.

src/classes/RD2_RecurringDonationsOpp_TDTM.cls
force2b: Thank you

src/classes/RD2_OpportunityMatcher.cls
RadSFDeveloper: The change has been done, thanks @force2b  for the feedback!

RadSFDeveloper:
@ravensisco , @Julian88Tex : the PR is ready for testing. There might be a pending change to a  Custom label check although I did make changes per Sam's feedback. However, the main functionality can be tested and label as is now if fine.

src/classes/UTIL_SObject.cls
```diff
@@ -68,7 +68,9 @@ public with sharing class UTIL_SObject {
     * @return Set<Id>
     */
     public static Set<Id> extractId(List<SObject> records, Schema.SObjectField field) {
-        return new Set<Id>((List<Id>) extractField(List<Id>.Class, records, field.getDescribe().getName()));
+        Set<Id> values = new Set<Id>((List<Id>) extractField(List<Id>.Class, records, field.getDescribe().getName()));
```
RadSFDeveloper: @force2b , can you please review this change so utility does not return null Ids?
@ravensisco , affected is currency on Opp verification: AC2. There is no change in behaviour, but when Opps are not assigned to a RD, there is no redundant DB query to get RD Ids.

force2b: The change makes sense. I can see a null value in this set causing issues.

PR #5068, opened 2020-02-06
Gift Entry: Fix matching modal selection bug

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geReviewDonations/geReviewDonations.js
```diff
@@ -177,4 +176,9 @@ export default class geReviewDonations extends NavigationMixin(LightningElement)
                 handleError(error);
             });
     }
+
+    @api
+    setProperty(property, value) {
```
bdvorachek: This gives me pause. Do we need to be able to modify every property on this component?

psadie-sfdo: updated!

src/lwc/geFormField/geFormField.js
```diff
@@ -304,12 +304,13 @@ export default class GeFormField extends LightningElement {
...
      */
     @api
-    setValue(value) {
+    setValue(value, displayValue) {
         if (this.isLookup) {
             const lookup = this.template.querySelector('c-ge-form-field-lookup');
             if (value) {
-                const displayValue =
-                    data[this.sourceFieldAPIName.replace('__c', '__r')].Name;
+                // TODO: Where does data below come from?
```
ptewson-sfdo: I may be missing something, but in talking with @kenlewis-sfdo yesterday, there are instances where events have the field value in the payload directly, and some where the value is an object that contains the value in a displayValue property of that object.

From Ken's comments in his branch:
            // dataImport should be object with keys = source field api name, and value =
            // an object with at least one property "value" - but could have more, for
            // instance "displayValue" for lookup fields, etc.
            // See response from getRecord for data structure guidance.

https://github.com/SalesforceFoundation/NPSP/compare/feature/gift-entry...ken-populate-selected-lookup-record-sibling-fields?expand=1

psadie-sfdo: Oh interesting! Looks like we're consolidating `setValue` and `load` into one method. Should I preemptively update this to be inline with Ken's branch?

ptewson-sfdo: Maybe - does that feel like a smart improvement? My level of detail at this point is that it feels like you are both thinking about this and if there's a smart way to align then that's great, but I can dive deeper if need be!

PR #5070, opened 2020-02-06
Allocations Widget error label

bdvorachek:
# Critical Changes

# Changes
- Use customLabel from Donation Amount field when displaying Allocations error message that amounts do not match.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geFormService/geFormService.js
```diff
@@ -111,6 +115,30 @@ class GeFormService {
         return this.objectMappings[objectDevName];
     }
 
+    /**
+     * Gets the Custom Label that was used on the form template for the Donation Amount field
+     * @return {string}
+     */
+    getDonationAmountCustomLabel() {
```
ptewson-sfdo: I have a couple comments about this - the first is minor, but "custom label" makes me think of a platform custom label and just to make that distinction clear, maybe we can call this something else - "label override" or "display label" or something like that

ptewson-sfdo: The second comment is just wondering it this is the best way to do this - the basic idea here is that we need to communicate the donation amount between the allocations widget and the corresponding field on the form, and I think we've been doing that through sending up events when key fields change (Is that basically correct @psadie-sfdo and @kenlewis-sfdo ?) - basically I want to make sure that this is coherent with the method we are using to notify the match donations modal that the donor record is changing.

I think we all share a perspective that this could use an overhaul at some point in the (hopefully) not-too-distant future as we understand these concepts a bit more clearly, so maybe we just forge ahead on this path. But I just want to make sure we have a chance to review this approach and make sure its as consistent as it can be!

ptewson-sfdo: I'm looking primarily at setReviewDonationsDonorProperties() in formRendererService for comparison. I think my primary reservation here is a) making sure our approaches are consistent and b) This is a LOT of iteration to go through on a regular basis, and if I read this correctly this is called every time we validate the GAUs - which is called every time there's a relevant value change. 

I wonder if there's a way to capture these key fields - donation amount for this, donor identifiers for the matching modal - while we are rendering the form initially, where we already have to iterate through all these elements. The identifying label for the total amount field shouldn't change after that fact.

src/lwc/geFormService/geFormService.js
```diff
@@ -111,6 +115,30 @@ class GeFormService {
...
+        // find field that is mapped to Opportunity Amount
+        const mapping = this.getFieldMappingWrapperFromTarget(`${OPPORTUNITY_OBJECT.objectApiName}.${OPPORTUNITY_AMOUNT.fieldApiName}`);
+        const mappingDevName = mapping.DeveloperName;
+        // get developer name of mapping cmt
+        let fieldElement;
+        for(const section of this.formTemplate.layout.sections) {
```
psadie-sfdo: Just a note that I've found myself doing more and more of this section and fields looping also. Feels like a spot we could create a utility method for in the future.

bdvorachek: A way to flatten a full layout down to an array of elements, maybe? Or a class that accepts a form layout as a constructor and has methods that allow you to retrieve sections, elements etc? I agree but not sure which fits better.

src/lwc/geFormService/geFormService.js
bdvorachek: @ptewson-sfdo Would caching them on the GeFormService instance make more sense? Any time the form template is loaded, set some private variables that have those key strings and allow other components to access them via a getter or public method?

ptewson-sfdo: I think so, but I'm hesitant to wade into that level of refactor right at this juncture unless it seems like a really clear and straightforward change. I still want a chance to deploy this branch and test but maybe let's move this conversation out of scope for this story and I will try and approve this as-is. 

General observations, here's my understanding of field-changes that require some kind of event or notification for other elements on the form: 
* Account/contact donor lookups to notify review donations modal of donor change
* Donor Type picklist to notify review donations modal of donor change 
* Donation Amount for Allocations widget 
* ANY LOOKUP for Ken's work to populate sibling fields when you set a lookup

ptewson-sfdo: @bdvorachek I wonder if there's a short term improvement here - if you store the value of donationAmountCustomLabel() as a local variable, you could only call GeFormService.getDonationAmountCustomLabel() if the value hadn't been set? Essentially caching the result of that call locally - there's no reason the donation amount label changes while you are looking at the form, right?

ptewson-sfdo: (Oh right, you had that idea 2-3 comments ago! But maybe let's not worry about resolving any of the others for this work, just maybe avoid duplicate calls to GeFormService.getDonationAmountCustomLabel())

PR #5071, opened 2020-02-06
Gift Entry:  Exclude OCR groups

psadie-sfdo:
# Critical Changes

# Changes
- Exclude OCR Object Mappings from available form field selections in Template Builder
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geTemplateBuilderFormFields/geTemplateBuilderFormFields.js
```diff
@@ -27,6 +27,11 @@ const REQUIRED_FORM_FIELDS = [
     CONTACT1_LASTNAME_INFO.fieldApiName
 ];
 
+const EXCLUDED_OBJECT_MAPPINGS = [
+    'Opportunity_Contact_Role_1_',
```
ptewson-sfdo: This is a minor detail, but could you leave off the trailing underscore here and change the math on the substring below to be 'length - 10'? It both looks nicer, and would match to one of our default, protected mappings if for whatever reason we ever had those in play

annejackson:
Do we need to lurch the WI for this @psadie-sfdo ?

psadie-sfdo:
**lurch: attach W-038983

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HrITQA0" target="_blank">W-038983</a>

PR #5074, opened 2020-02-06
Allocations Widget: 4 column layout, hide error if no rows present

bdvorachek:
# Critical Changes

# Changes
- Updated Allocations widget to use 4 column layout
- Error message when allocations do not align with total amount donated should only appear when at least one allocation row is present
- Only show remaining amount when at least one allocation row is present
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


annejackson:
This will be a good one to get in soon @ptewson-sfdo @vmartinov-sforg

PR #5076, opened 2020-02-06
Gift Entry: fix missing name from SGE header

psadie-sfdo:
# Critical Changes

# Changes
- Fixes missing donor name from SGE header when coming from an account/contact page
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormRenderer/geFormRenderer.js
+++ b/src/lwc/geFormRenderer/geFormRenderer.js
@@ -27,6 +27,9 @@ import DATA_IMPORT_PAYMENT_IMPORTED_FIELD from '@salesforce/schema/DataImport__c
 import DATA_IMPORT_DONATION_IMPORT_STATUS_FIELD from '@salesforce/schema/DataImport__c.DonationImportStatus__c';
 import DATA_IMPORT_PAYMENT_IMPORT_STATUS_FIELD from '@salesforce/schema/DataImport__c.PaymentImportStatus__c';
 
+import ACCOUNT_NAME_FIELD from '@salesforce/schema/Account.Name';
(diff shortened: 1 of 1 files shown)


psadie-sfdo:
**lurch: attach W-039200

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HroBQAS" target="_blank">W-039200</a>

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -43,7 +46,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     @api donorApiName;
     @api donorRecord;
 
-    fieldNames = [];
+    fieldNames = [ ACCOUNT_NAME_FIELD, CONTACT_NAME_FIELD ];
```
vmartinov-sforg: In getRecordFieldNames in utilTemplateBuilder.js we initialize the array of fields to 
let fieldNames = [`${apiName}.Id`]; 
to handle the case when the template does not contain account or contact field names and the form was not rendering. Can we get rid of that line now that we initialize the array with Contact.Name and Account.Name?

vmartinov-sforg: I'm working in another branch and the default template loads fine for me from contact/account. This bug was already fixed once.

vmartinov-sforg: This bug was already fixed once, and I don't see it happening in my branch. Did we regress maybe at some point?

PR #5102, opened 2020-02-11
RD2: Default Installment Frequency

andrewyu-salesforce:
Set default Installment Frequency when null on a Recurring Donation insert. Ensure entered Installment Frequency is within allowed range.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom Label:
-- RD2_InstallmentFrequencyMustBeValid

# Deleted Metadata


andrewyu-salesforce:
**lurch: W-039192

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HrnFQAS" target="_blank">W-039192</a>

src/labels/CustomLabels.labels
```diff
@@ -1895,6 +1895,14 @@ If you&apos;re not sure how to resolve these errors, post a message in the Nonpr
...
         <shortDescription>The RD2 Batch job can only be executed when Enhanced RD is enabled</shortDescription>
         <value>You must enable Enhanced Recurring Donations before you can run the Enhanced Recurring Donations batch job.</value>
+    </labels>
+    <labels>
+        <fullName>RD2_InstallmentFrequencyMustBeValid</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Validate Recurring Donations data if it has a valid installment frequency</shortDescription>
+        <value>Installment Frequency is required. Enter a value between 1 and 20.</value>
```
andrewyu-salesforce: @sam-knox Please review the new label

andrewyu-salesforce:
@RadSFDeveloper The code is ready to be reviewed. I had a question on WI on the Installment Frequency range, but the change is minimum (condition update).

src/labels/CustomLabels.labels
sam-knox: @andrewyu-salesforce - Label looks good to me. Not sure about your question regarding a value of 0. I assumed that 0 was not a valid entry.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -81,30 +86,40 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
-        autogenerateNames(rds, oldRds, triggerAction);
-        validate(rds, oldRds, triggerAction);
-        handleRDSchedules(rds, oldRds, triggerAction);
+        autogenerateNames(rds, oldRds);
+
+        assignDefaultInstallmentFrequency(rds);
+
+        Boolean dataIsValid = validate(rds, oldRds);
+
+        if (!dataIsValid) {
```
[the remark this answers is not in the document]
andrewyu-salesforce: I agree the the class level variable will not work and have to be fixed. The problem right now is at RD2_ScheduleService line 422 `Integer.valueOf(schedule.InstallmentFrequency__c)` If the RD is updated to null, the code will throw an error at this line because it's an null pointer exception error. I will move the class level variable back to per record and add an extra null check in schedule class to prohibit any RD is entered with Installment Frequency being null.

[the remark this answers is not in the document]
RadSFDeveloper: There can be calls to the schedule service outside of the RD2 trigger. See if the schedule service needs to be updated with minimal changes to the RD2 trigger or data migration etc.

andrewyu-salesforce:
@RadSFDeveloper Thanks for the first code review. The code is ready to be reviewed again.

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -525,7 +525,92 @@ private with sharing class RD2_ValidationService_TEST {
...
+
+        System.assert(errMessage.contains(System.Label.RD2_InstallmentFrequencyMustBeValid),
+            'RD update should fail due to invalid installment frequency with the message: ' + System.Label.RD2_InstallmentFrequencyMustBeValid);
+    }
+
+    /***
+    * @description Verifies DML should fail when installment frequency is updated to a invalid value (greater than twenty) 
+    */
+    @isTest
+    private static void shouldBlockRdUpdateWhenInstallmentFrequencyChangesIsInvalid() {
```
RadSFDeveloper: `shouldBlockRDUpdateWhenInstallmentFrequencyIsOutOfRange`

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -525,7 +525,92 @@ private with sharing class RD2_ValidationService_TEST {
...
+                'RD update should fail due to invalid installment frequency with the message: ' + System.Label.RD2_InstallmentFrequencyMustBeValid
+            );
+        }
+    }
+
+    /***
+    * @description Verifies DML update should succeed when installment frequency is valid
+    */
+    @isTest
+    private static void shouldRdUpdateWhenInstallmentFrequencyIsValidChange() {
```
RadSFDeveloper: `shouldUpdateRDWhenInstallmentFrequencyIsWithinRange()`

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -525,7 +525,92 @@ private with sharing class RD2_ValidationService_TEST {
...
+    * @description Verifies DML should fail when installment frequency is updated to a invalid value (greater than twenty) 
+    */
+    @isTest
+    private static void shouldBlockRdUpdateWhenInstallmentFrequencyChangesIsInvalid() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+
+        List<npe03__Recurring_Donation__c> rds = getRecurringDonationBuilder()
+            .withContact(getContact().Id)
+            .withInstallmentFrequency(10)
+            .build(2);
```
RadSFDeveloper: Can we `build(3)` and ensure the valid RD is updated with other RDs in the bulk are not valid?

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -525,7 +525,92 @@ private with sharing class RD2_ValidationService_TEST {
...
+        String errMessage = '';
+
+        Test.startTest();
+        List<Database.SaveResult> results = Database.update(rds, false);
+        Test.stopTest();
 
+        for (Database.SaveResult result : results) {
+            System.assert(
+                result.getErrors()[0].getMessage().contains(System.Label.RD2_InstallmentFrequencyMustBeValid),
+                'RD update should fail due to invalid installment frequency with the message: ' + System.Label.RD2_InstallmentFrequencyMustBeValid
```
RadSFDeveloper: - We should also check that `!result.isSuccess()` for failed records and the `result.isSuccess()` for the valid update.
- On the case of error we should display actual error. So it can be `'RD update should fail due to invalid Installment Frequency. Actual message: ' + result.getErrors()[0].getMessage());`

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -525,7 +525,92 @@ private with sharing class RD2_ValidationService_TEST {
...
+            .withContact(getContact().Id)
+            .withInstallmentFrequency(10)
+            .build(2);
+            
+        insert rds;
+
+        rds[0].InstallmentFrequency__c = 21;
+        rds[1].InstallmentFrequency__c = 0;
+
+        String errMessage = '';
```
RadSFDeveloper: `errMessage` is not used in this test.

src/classes/RD2_ScheduleService.cls
```diff
@@ -172,6 +172,10 @@ public without sharing class RD2_ScheduleService {
         } else if (oldRd == null) {
             isNeeded = true;
 
+        // Invalid Installment Frequency should never result in a schedule update. This is a fail safe measure only
+        } else if (newRd.InstallmentFrequency__c == null || newRd.InstallmentFrequency__c < 1 || newRd.InstallmentFrequency__c > 20) {
```
RadSFDeveloper: I have reviewed code but need to test the change wrt schedule calculations and Active Schedule and Installments components.

andrewyu-salesforce: I have tested the Installment component and the future installments are calculated correctly.

andrewyu-salesforce:
@RadSFDeveloper Thanks for the review. The change is committed and is ready to be reviewed

andrewyu-salesforce:
@RadSFDeveloper I add an extra validation in the method to prevent the exception. This check prevent Integer.valueOf()'s null pointer exception and math divided by 0 exception. The return null will be handled by the parent function and not generate any installment.

src/classes/RD2_ScheduleService.cls
```diff
@@ -412,6 +417,10 @@ public without sharing class RD2_ScheduleService {
     * @return Date The Next Donation Date
     */
     private Date getNextDonationDateGreaterThan(RecurringDonationSchedule__c schedule, Date referenceDate) {
+        if (schedule.InstallmentFrequency__c == null || schedule.InstallmentFrequency__c < 1) {
```
RadSFDeveloper: Instead of checking the Installment Frequency here, can we check it in the `isActive()` method?
@npsp-reedestockton , do you see any issue with checking if the valid schedule is a schedule having Installment Frequency defined?

andrewyu-salesforce: From reading the code, it seem like moving the check to isActive() will be a better change.

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: I don't understand the test for > 20 here.  Can you explain?

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: I don't really see anything wrong with that, but shouldn't validation be happening in the trigger?

src/classes/RD2_ScheduleService.cls
andrewyu-salesforce: It's the new validation that only allows installment between 1 and 20, everything else should be invalid.

src/classes/RD2_ScheduleService.cls
RadSFDeveloper: It does, this is the case when the client org has a trigger setting the Installment Frequency to null. Since trigger order is not guaranteed, the result is not deterministic so there can be a record having Installment Frequency being null. If so, creating schedules and Installments component fail.

andrewyu-salesforce: Also, after adding the error to the record, the trigger will still run from start to finish and end up throwing an exception due to bad data. The validation in ScheduleServices class further prevent the null pointer exception and divided by 0 exception

andrewyu-salesforce:
@RadSFDeveloper Just in case you did not see the new change. The code is ready to review

src/classes/RD2_ScheduleService.cls
npsp-reedestockton: Thanks for the explanation!

PR #5131, opened 2020-02-18
Gift Entry: Set Batch Defaults on BGE Form

voduyemi:
# Critical Changes

# Changes
- Batch Defaults are now set on the Batch Gift Entry Form taking precedence over the Template level defaults that have been set. The values set on the Batch will always override the corresponding template field default value.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormRenderer/geFormRenderer.js
+++ b/src/lwc/geFormRenderer/geFormRenderer.js
@@ -77,7 +77,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     @track mappingSet = '';
     @track version = '';
     @track formTemplateId;
-    @track batchDefaults;
+    _batchDefaults;
 
     erroredFields = [];
     CUSTOM_LABELS = { ...GeLabelService.CUSTOM_LABELS, messageLoading };
@@ -209,7 +209,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     wiredBatch({data, error}) {
         if (data) {
             this.formTemplateId = data.fields[FORM_TEMPLATE_FIELD.fieldApiName].value;
-            this.batchDefaults = data.fields[BATCH_DEFAULTS_FIELD.fieldApiName].value;
+            this._batchDefaults = data.fields[BATCH_DEFAULTS_FIELD.fieldApiName].value;
             GeFormService.getFormTemplateById(this.formTemplateId)
                 .then(template => {
                     let errorObject = checkPermissionErrors(template);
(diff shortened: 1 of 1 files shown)


voduyemi:
**lurch: attach W-038958

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HrEMQA0" target="_blank">W-038958</a>

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -166,17 +178,17 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
 
         if (typeof formTemplate.layout !== 'undefined'
             && Array.isArray(formTemplate.layout.sections)) {
-
             // add record data to the template fields
-            if (isNotEmpty(fieldMappings) && isNotEmpty(this.donorRecord)) {
-                let sectionsWithValues = setRecordValuesOnTemplate(formTemplate.layout.sections,
+
+             if (isNotEmpty(fieldMappings) && isNotEmpty(this.donorRecord)) {
```
ptewson-sfdo: This line is just *one* space too far indented, I think!

voduyemi: You're right! Missed it.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -66,6 +77,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     @track mappingSet = '';
     @track version = '';
     @track formTemplateId;
+    @track batchDefaults;
```
kenlewis-sfdo: If this field doesn't need to be explicitly reactive, it could be an underscored "private" var `_batchDefaults` (even though all fields will be reactive from Spring '20 forward...)

voduyemi: Thanks Ken, you've got a point. We're now fully on Spring '20 and the it's not meant to be reactive. I just updated it.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -877,4 +890,33 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+            sections.forEach(section => {
+                const elements = section.elements;
+                elements.forEach(element => {
+                    for (let key in batchDefaultsObject) {
+                        if (batchDefaultsObject.hasOwnProperty(key)) {
+                            const batchDefault = batchDefaultsObject[key];
+                            if (batchDefault.objectApiName === element.objectApiName &&
+                                batchDefault.fieldApiName === element.fieldApiName) {
+                                if (!isUndefined(batchDefault.value)) {
+                                    element.defaultValue = batchDefault.value;
```
ptewson-sfdo: Do we need to worry about setting the display value as a default for lookups in addition to the field value? Or does that come along with the lookup behavior already?

PR #5132, opened 2020-02-18
Add a file with labels from the NPSP MetaDeploy plans, for translation

davisagli:
This adds a JSON file listing the strings that are used in NPSP's MetaDeploy installation plans, so that they can be submitted for translation.

@codexmarcel I'm open to suggestions if you think there is a better place to put this in the repository's directory layout.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
@davisagli Thanks. How does this need to be delivered by the translation team? Are they creating one new file per language or do all translations go in this file? Any other formatting rules they need to know about?

davisagli:
@force2b I'm flexible -- I figured I would wait to see what they deliver and then write the code to load the translations into MetaDeploy accordingly -- but my suggestion would be one file per language with names like labels_es.json (for Spanish). The top-level keys like "product title" are categories to provide context for the translators and do not need to be translated. Within each category, the keys should remain in English to identify the label, and the empty string value should be filled in with the translation.

davisagli:
@force2b I just received translations from the localization work that has been done on the non-product-specific parts of MetaDeploy, and it actually includes some of the NPSP-specific text. So I will update this PR to include those translations, and that will provide an example of what I mean.

davisagli:
@force2b @codexmarcel Okay, I've updated this to include one file for each language (with a revised format), including the translations that we've already received. If you look at the non-English files, you'll see that there are some strings which were already translated and some which still need to be translated.

I also added some strings to the metadeploy plan configuration in cumulusci.yml which are currently configured only via the MetaDeploy UI (preflight message, post-install message, and error message). Soon I'll make sure they get published to MetaDeploy when a new release is published via the metadeploy_publish task; for now it's just to facilitate my script which extracts the labels.

force2b:
Thanks @davisagli. Can you confirm that the `cumulusci.yml` file does not need translation?

davisagli:
@force2b correct -- we're just submitting the files in the metadeploy directory, which I generated from what's in cumulusci.yml using a script.

davisagli:
@force2b please wait to merge -- I have one more adjustment to the format coming, based on communication with Glory.

force2b:
@davisagli No problem. I actually thought you would merge. I can do it too. Just let let me know when it's ready. I assume we're waiting for the merge freeze to be lifted later today anyway.

davisagli:
@force2b sounds like the localization team would like me to submit these directly to their repository, so I'll do that, close this for now, and resubmit once we have the finished translations back.

PR #5135, opened 2020-02-18
RD2: Enforce migration validation when outdated

RadSFDeveloper:
# Critical Changes

# Changess

## Enhanced Recurring Donations Changes

- When data validation has completed more than 7 days ago, enforce validation again before users enable enhanced Recurring Donations.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-038102

src/labels/CustomLabels.labels
```diff
@@ -300,6 +300,14 @@
         <shortDescription>AssistiveTextError</shortDescription>
         <value>Expired</value>
     </labels>
+    <labels>
```
RadSFDeveloper: @force2b , @sam-knox : even though this PR is work in progress, can you please review three new custom labels so @meena-sfdc  can add them into the label freeze PR?

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HpkRQAS" target="_blank">W-038102</a>

src/labels/CustomLabels.labels
```diff
@@ -300,6 +300,14 @@
...
+        <fullName>AssistiveTextInfo</fullName>
+        <categories>AlternativeText</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>AssistiveTextInfo</shortDescription>
```
force2b:
```suggestion
        <shortDescription>AssistiveText for &quot;Info&quot;</shortDescription>
```

src/classes/RD2_EnablementDelegate_CTRL.cls
```diff
@@ -128,15 +157,48 @@ public with sharing class RD2_EnablementDelegate_CTRL extends STG_Panel {
...
+    /***
+    * @description Retrieves number of days that dry run is valid after completion
+    * @param state Enablement state
+    * @return Integer
+    */
+    private static Integer getDryRunLimit(EnablementState state) {
+        Integer dryRunLimit = DRY_RUN_VALID_DAYS;//default
+
+        //if this is a scratch org, enable override for testing purposes
+        if (UTIL_Namespace.isScratchOrg()) {
```
RadSFDeveloper: @force2b , the dry run days valid value is not configurable, but to allow testing without waiting 8 days, the 'dryRunLimit` can be added/overwritten in the RD Settings "Recurring Donations Enablement State" and that will be taken into consideration when RD2 enablement page is loaded on a scratch org. The number needs to be > 0 in order to be accepted. Overwriting is not applicable on an org that is not a scratch org. 

Days between is number of the days between validation batch CompletedDate and today. The validation is outdated when days between is > 7 (or any overwritten value specified in the RD settings enablement state).

@ravensisco , @heidmary  ^

src/labels/CustomLabels.labels
```diff
@@ -316,22 +316,6 @@
         <shortDescription>AssistiveTextNewWindow</shortDescription>
         <value>Link opens in a new window</value>
     </labels>
-    <labels>
-        <fullName>RD2_EnablementDryRunOutdatedError</fullName>
```
RadSFDeveloper: @force2b , the custom labels were not sorted so they are moved to the correct location in the Custom Labels. Let me know if there is any issue.

@meena-sfdc  ^

src/classes/RD2_EnablementDelegate_CTRL.cls
force2b: Trying to test this out and I found a way to test without waiting a full day. If you remove the `&& state.dryRunLimit > 0` condition in the method below then it's possible to set the dryRunLimit to -1 in Custom Settings. That forces a run of today to be out of date. I don't believe this is an issue to change in the production code as it's a value that even if a customer were to set would still force them to re-run the validation. I think this would greatly help in testing/qa.

`{"isReady":true,"isMigrationEnabled":false,"isMetaLaunched":false,"isMetaConfirmed":false,"isEnabled":false,"isDryRun2":false,"isConfirmed":false,"dryRunLimit":-1}`

force2b: Ah. That downside of the above is that the dryRunLimit is never reset back to 7. It stays as -1 so even after re-running the validation it still shows as out of date. For my testing, I just edited the custom setting to change -1 to 1.

RadSFDeveloper: That is correct, settings it to -1 will outdate every dry run.

src/classes/RD2_EnablementDelegate_TEST.cls
```diff
@@ -351,6 +348,163 @@ public with sharing class RD2_EnablementDelegate_TEST {
...
+        );
+
+        setLatestDryRunAsValid();
+
+        RD2_EnablementDelegate_CTRL.EnablementState loadState = (RD2_EnablementDelegate_CTRL.EnablementState) JSON.deserialize(
+            RD2_EnablementDelegate_CTRL.loadState(), RD2_EnablementDelegate_CTRL.EnablementState.class
+        );
+
+        System.assertEquals(true, loadState.isConfirmed,
+            'Enablement confirmation should stays checked when dry run validation is valid');
```
force2b:
```suggestion
            'Enablement confirmation should remain checked when dry run validation is valid');
```

src/classes/RD2_EnablementDelegate_TEST.cls
```diff
@@ -351,6 +348,163 @@ public with sharing class RD2_EnablementDelegate_TEST {
...
+            RD2_EnablementDelegate_CTRL.enableEnhancement();
+        } catch (AuraHandledException e) {
+            actualException = e;
+        }
+
+        System.assertNotEquals(null, actualException, 'The exception should be thrown');
+        System.assertEquals(System.Label.RD2_EnablementDryRunOutdatedError,
+            actualException.getMessage(), 'The exception message should match');
+
+        System.assertEquals(false, STG_Panel.stgService.stgRD.isRecurringDonations2Enabled__c, 'Custom Settings field should not be set');
```
force2b:
```suggestion
        System.assertEquals(false, STG_Panel.stgService.stgRD.isRecurringDonations2Enabled__c, 'IsRecurringDonations2Enabled field should not be checked');
```

PR #5153, opened 2020-02-21
Gift Entry: batch wizard validations

psadie-sfdo:
- Add required field validations to the batch wizard
- Fix rich text input types in utilInput
- Custom Labels
  - commonMissingRequiredFields

# Critical Changes

# Changes
## Gift Entry Enhancements
- We added field validations for batch header fields. Now, when you set batch header fields as required in Template Builder, users will be required to enter information in those fields when they create a new batch.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


psadie-sfdo:
**lurch: attach W-039300

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hs1UQAS" target="_blank">W-039300</a>

src/labels/CustomLabels.labels
```diff
@@ -4551,6 +4551,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Field Label</shortDescription>
         <value>Field Label</value>
     </labels>
+    <labels>
+        <fullName>commonMissingRequiredFields</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>These required fields must be completed: {0}</shortDescription>
```
```suggestion
        <shortDescription>Page-level error message about missing required fields</shortDescription>
```

melissabarber:
@ptewson-sfdo @psadie-sfdo I added one suggestion to the custom label and updated the PR description. Let me know if you have any questions. Thanks!

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -248,11 +254,43 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
+        this.resetValidations();
+    }
+
+    /*******************************************************************************
+    * @description Method collects all batch header fields and checks for their
+    * validity.
+    */
+    validateBatchHeaderFields() {
+        this.resetValidations();
+        this.hasInvalidBatchFields = false;
```
ptewson-sfdo: I think this line is an unneeded duplicate if you are calling resetValidations() above!

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -248,11 +254,43 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
+        this.missingBatchHeaderFieldLabels = [];
```
ptewson-sfdo: Also shouldn't need this line, duplicates resetValidations()

src/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
```diff
@@ -54,6 +54,16 @@ export default class geTemplateBuilderFormField extends LightningElement {
         return null;
     }
 
+    get fieldApiName() {
```
ptewson-sfdo: Just curious how we got here with field having either "apiName" or "fieldApiName" property?

src/lwc/utilInput/utilInput.js
```diff
@@ -196,13 +214,75 @@ export default class utilInput extends LightningElement {
             this.value = event.target.value;
         }
 
+        let isValid = true;
+        if (this.isLightningRichText) {
```
ptewson-sfdo: I'm wondering if we need this condition - the default check (this.isValid()) calls checkFieldValidity() and if the type is rich text, that calls this.checkRichTextValidity() ... am I right in thinking that isValid() checks rich text validity as well, and that you can just reduce this block to: 

isValid = this.isValid()

src/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
psadie-sfdo: Yeah, this bothered me as well. This is because when we get the describe info from `getObjectInfo` uiObjectInfoApi we get back the entire object describe along with all the fields. The field api name key from that is `apiName`. When we do `import ... from '@salesforce/schema/...` the field api name key is `fieldApiName`.

src/lwc/utilInput/utilInput.js
psadie-sfdo: That's totally right! Updated!

src/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
ptewson-sfdo: Ugh, bummer. But that makes sense why we would need to handle both cases!

PR #5183, opened 2020-02-27
Update Address__c.object

searheart:
# Critical Changes

# Changes
Added four fields to the address object in support of Insights Address Services release

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


lmeerkatz:
@searheart Can you clone the NPSP repo and branch from there? I believe that will give you automatic merges from master, which will make development easier.

force2b:
@searheart It would be helpful if you could do that instead of forking the repo. Also, feel free to assign me as the reviewer/assignee for the PR.

Other questions:

- Should one or more of these new fields be on the default Address page layout for all new installs? If so, can you either include an updated layout file or maybe provide a screen capture of the address record showing where you want the new fields.?
- You had mentioned apex changes that went along with the new fields. Are they coming in a separate PR?

searheart:
@force2b The apex work and the page layout will be in a different PR.  I need to get UX to look at the address page layout to approve my suggested changes.

lmeerkatz:
@searheart Let's get this moved into a new branch now that you have repo access. I would want to hold these changes until the Apex work is ready if we can, as anything merged into master is eligible for release and I wouldn't want to release the fields until the supporting automation is ready. /cc @force2b

src/objects/Address__c-Address Layout.layout
```diff
@@ -0,0 +1,196 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: Thanks, though this file needs to be in the src/layouts folder. You'll see there is already a layout file there. That file just needs to be updated to include the new fields.

src/objects/Address__c.object
```diff
@@ -4,40 +4,140 @@
         <actionName>Accept</actionName>
         <type>Default</type>
     </actionOverrides>
+    <actionOverrides>
```
force2b: Nothing should be changed in this file other than the new fields.

src/objects/package.xml
```diff
@@ -0,0 +1,3596 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b: As with the layout file, there is already an existing package.xml in the src folder. That file should only have 4 changed lines for the 4 new fields.

force2b:
@searheart  Also, now that you have access to create a branch in our repo can you close this PR, clone the npsp repo locally, create a new branch and then a new PR with the changes? This is still a forked PR so it cannot be merged.

PR #5198, opened 2020-03-03
Feature/searheart/insights address fields

searheart:
# Critical Changes

# Changes
Added fields for AddressVerification and NCOA per https://salesforce.quip.com/3HA7AkIV1YI9
Updated Address__c standard layout to include Address Verification section with related fields moved or added in this section
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/package.xml
+++ b/src/package.xml
@@ -400,7 +400,6 @@
         <members>RD2_VisualizeScheduleController</members>
         <members>RD2_VisualizeScheduleController_TEST</members>
         <members>RD_AddDonationsBTN_CTRL</members>
-        <members>RD_AddDonationsBTN_TEST</members>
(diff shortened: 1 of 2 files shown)


src/package.xml
```diff
@@ -845,6 +844,7 @@
         <members>Address__c.Congressional_District__c</members>
         <members>Address__c.County_Name__c</members>
         <members>Address__c.Default_Address__c</members>
```
force2b: @searheart The only changes that should be in this file are the 4 new Address__c fields.

force2b:
@searheart Also, in addition to the reduced `package.xml` changes, this PR should include the 4 new fields in the address object file.

PR #5207, opened 2020-03-05
Gift Entry: add elevate iframe

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Visualforce Pages
  - GE_TokenizeCard
- Apex Classes
  - GE_PaymentServices
  - GE_PaymentServices_TEST
# Deleted Metadata


psadie-sfdo:
**lurch: attach W-039740

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsmPQAS" target="_blank">W-039740</a>

src/classes/GE_GiftEntryController.cls
```diff
@@ -37,6 +37,21 @@
 */
 public with sharing class GE_GiftEntryController {
 
+    @AuraEnabled
```
bdvorachek: Let's add (cacheable=true) to this one.

src/package.xml
```diff
@@ -642,6 +642,7 @@
         <members>CRLP_RollupRD_BTN</members>
         <members>CRLP_RollupSetup</members>
         <members>EP_ManageEPTemplate</members>
+        <members>GE_TokenizeCard</members>
```
bdvorachek: Don't forget to add the GE_PaymentServices apex class to package.xml too

src/classes/GE_GiftEntryController.cls
```diff
@@ -37,6 +37,31 @@
...
+    }
+
+    @AuraEnabled
+    public static String makePurchaseCall(String token) {
+        GE_PaymentServices paymentServices = new GE_PaymentServices(token);
+        HttpResponse response = paymentServices.makePurchaseCall();
+
+        // Parse the JSON response...
+        if (response.getStatusCode() != 201) {
+            System.debug('The status code returned was not expected: ' +
```
kenlewis-sfdo: Are we going to keep these debug statements for continued dev?  Or should we not commit them... ?

src/classes/GE_PaymentServices.cls
```diff
@@ -0,0 +1,144 @@
...
+        }
+        return null;
+    }
+
+    //
+    // Methods for retrieving request headers
+    //
+
+    private String getEndpointRoot() {
+        return 'https://7jpea5zk79.execute-api.us-west-2.amazonaws.com';
```
kenlewis-sfdo: Have we looked at/ruled out using named credentials to store this endpoint info?

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -1,18 +1,114 @@
...
+        }
+    }
+
+    /*******************************************************************************
+    * @description Method sends a message to the visualforce page iframe requesting
+    * a token. Response for this request is found and handled in
+    * registerPostMessageListener.
+    */
+    requestToken() {
+        const iframe = this.template.querySelector('iframe');
```
kenlewis-sfdo: Should we/can we add a "data-" attribute and use that as a selector?  (although the likelihood of having another iframe in the template is low.... so maybe not necessary)

src/pages/GE_TokenizeCard.page-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexPage xmlns="http://soap.sforce.com/2006/04/metadata"> 
+    <apiVersion>47.0</apiVersion>
```
kenlewis-sfdo: Is 47 correct or should these be bumped to 48?

vmartinov-sforg:
@psadie-sfdo  I plan to take a look this afternoon, sorry it takes so long.

src/pages/GE_TokenizeCard.page
```diff
@@ -0,0 +1,94 @@
...
+        const hostname = window.location.hostname;
+        const arr = hostname.split(".");
+        let domain = arr[0].replace('--npsp', '');
+        domain = domain.replace('--c', '');
+        const lightningOrigin = `https://${domain}.lightning.force.com`;
+
+        const sfdo = new sfdoPaymentsJsSdk();
+        sfdo.mount({
+            id: 'payments',
+            clientId: '00D6g0000000000EAA',
```
vmartinov-sforg: Can we add a reminder comment here to use {!$Organization.Id} when that is going to be ready?

src/classes/GE_GiftEntryController.cls
psadie-sfdo: I'm gonna go ahead and sweep these out along with console logs/alerts. I was thinking they might be useful as a lot of these methods are just starting points, but I believe our current guidelines are to remove them.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
psadie-sfdo: Adding that in. This should be the only iframe element that's visible to the component, but I think better practice to use a specific identifier regardless 🙏

src/classes/GE_PaymentServices.cls
psadie-sfdo: I don't know enough about packaging named credentials (reading up on it), but ideally I think we would like to use a named credential for this endpoint which would also solve the need to add this in remote site settings.

@dfuller114 - tagging you as I think you did the research on storing protected info that may have touched on named credentials. What do you think about storing this endpoint as a named credential?

src/classes/GE_FormRendererService.cls
```diff
@@ -36,6 +36,41 @@
 */
 public with sharing class GE_FormRendererService {
 
+    // TODO: Maybe create a domain util class? Find a better place for this?
```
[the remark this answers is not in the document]
psadie-sfdo: I think I'm actually leaning towards not creating that new domain utility class until we have maybe another method to go in there. This one's pretty light and it feels a bit much to create an entire class for one method. Please let me know if we feel strongly that we should though. Thanks!

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -1,18 +1,111 @@
...
+    async handleMessage(message) {
+        if (message.error) {
+            // Error with tokenization
+            let error = JSON.stringify(message.error);
+            handleError(error);
+        } else if (message.token) {
+            // TODO: Start - Remove later
+            // Make purchase call... for dev only
+            try {
+                let purchaseCallResponse = await makePurchaseCall({ token: message.token });
```
kenlewis-sfdo: Should the purchase callout be made inside this "Tokenize" component?  Maybe this component should be encapsulated and solely responsible for tokenizing and returning a value (the token) and then the FormRenderer should make the callout within (or before) its "save" flow...  Thoughts?

bdvorachek: I think this is ok for SGE, but we need to make sure that `makePurchaseCall` calls something that BGE can use later on.

Alternatively, If we want the DI record to be populated and ready for import at the moment we make the purchase call, we should return the token to the formRenderer and let it take over from there.

I'm leaning slightly towards the second option, where the formRenderer receives the token and makes the purchase call.

voduyemi: I am leaning towards Ken's comment here, based on the component hierarchy we currently have, it might be better if we let FormRenderer be that one point for communicating with the back-end, It also just makes for an easily maintainable codebase.

psadie-sfdo: Would definitely not want to keep this here. Basically just left in here as an example to be removed later (which could be now?). I was thinking this would basically just receive the token from the iframe and saved locally in the component for the form renderer to later retrieve. Something along those lines.

src/classes/GE_PaymentServices.cls
dfuller114: A named credential normally might be a good place for that. however, we're already going to likely use a protected custom object to store other information from payment services that they update when they register an org. it might make sense to just store the URL there also. Named Credential also makes the most sense when you want the named credential to handle the authentication. to just store the URL there without using the auth protocols built into it is underutilizing that feature a bit I think.

Storing it in the custom object will also make it easier to know where all the payment related protect info is stored. It would also make it easier for PS to update the URL if it changes instead of that responsibility being on us since they will already have code to update that object.

src/classes/GE_PaymentServices.cls
```diff
@@ -0,0 +1,144 @@
...
+        Http http = new Http();
+        HttpRequest request = new HttpRequest();
+        request.setEndpoint(getEndpointRoot() + getEndpoint(ElevateEndpoints.PURCHASE));
+        request.setMethod('POST');
+        request.setHeader('Content-Type', 'application/json;charset=UTF-8');
+        request.setHeader('jwttoken', getJwt());
+        request.setHeader('sfdo.merchantids', getMerchantIds());
+        request.setHeader('sfdo.username', getUsername());
+
+        // Set the body as a JSON object...
```
voduyemi: @psadie-sfdo I think it might be beneficial if we create a class(maybe an inner class) for the body of the request and maybe call the `JSON.serialize(object)` on the object to help with building the JSON body string instead of building manually.

psadie-sfdo: That's a great idea! I wonder if we could make that update in the work item I have now that's focused on making this purchase call. If only to get this merged in and kind of level set in our feature branch. Happy to make that update here if we feel strongly about it now though.

voduyemi: That sounds good. I don't think this has to be done in this branch if you can implement in the your current WI. But we can open up this decision to the team. Thoughts? @vmartinov-sforg   @ptewson-sfdo @kenlewis-sfdo

PR #5226, opened 2020-03-09
Bug Bounty: Any User could edit the Level.PreviousLevel field value

force2b:
# Critical Changes

# Changes

- When editing a Level record, disable edit access to the Source, Target, Level, and PreviousLevel fields if the current User does not have Edit access to all those fields.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
**lurch: attach W-039488

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsLJQA0" target="_blank">W-039488</a>

src/pages/LVL_LevelEdit.page
```diff
@@ -13,125 +13,131 @@
...
-                                    <apex:selectOptions value="{!listSOSourceFields}"  />
-                                </apex:selectList>
+                            <div class="slds-form-element__row">
+                                <div class="slds-form-element slds-size_1-of-2">
+                                    <apex:outputLabel for="fldSourceField" styleClass="slds-form-element__label uiLabel-left form-element__label uiLabel" >
+                                        <span class="required slds-required">*</span>{!$ObjectType.Level__c.Fields.Source_Field__c.Label}
+                                    </apex:outputLabel>
+                                    <apex:selectList id="fldSourceField" value="{!Level__c.Source_Field__c}"
+                                                     required="false" size="1" styleClass="slds-select"
+                                                     disabled="{!NOT($ObjectType.Level__c.fields.Source_Field__c.updateable)}">
```
npsp-reedestockton: As you noted in the WI, Source_Field__c and Level_Field__c are required fields, so it isn't possible for disabled to ever evaluate to true.  I'm wondering if it might make sense here to test whether Previous_Field__c is updateable.  I know that isn't what the bug bounty description requires, but it makes no sense to me that both Source_Field__c and Level_Field__c would be updateable while Previous_Field__c would not be.  Would it make the page more secure in a way that makes sense to use Previous_Field__c as a proxy for all 3?  If Previous_Field__c is not updateable, disable all 3.

npsp-reedestockton: This really should change I think.  As the code currently stands, it is possible to leave a level__c record in an inconsistent state by changing source and level.  That leaves previous level potentially pointing to the wrong sobject.

force2b: Do you mean something like the following:
- An admin creates a Level pointing to the Account object and sets the Level and Previous Level fields pointing to fields on the Account object
- A non-admin user (without edit rights on the Previous Level field) edits the Level to change the target object to Contact and selects a new target field for the Level field, leaving the PreviousLevel field with it's previous value.

Even if unlikely, it's not impossible. I think you're suggesting that if the User cannot edit PreviousLevel field that they should not be able to edit the SourceField field either? Effectively, the User would need edit access on all 3 fields to edit any of those 3 fields. I like that. Great suggestion. Change forthcoming (after I reboot).

force2b:
Changes made @npsp-reedestockton

src/pages/LVL_LevelEdit.page
npsp-reedestockton: Yep.  That's the scenario I had in mind.  Definitely an obscure possibility, but I think your changes make the entire page more robust and secure.  Nice!

sam-knox:
@force2b - Can you confirm that the Change note is accurate the way I've written it?

force2b:
Perfect. Thanks @sam-knox

PR #5231, opened 2020-03-10
Gift Entry: Fabricate CardHolder Name

voduyemi:
# Critical Changes

# Changes
- Added functionality to fabricate the card holder name for the payment(credit card) to be used for tokenization and payments from the existing donor fields and name fields (contact and account) on a single gift entry form

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


voduyemi:
**lurch: attach W-039742

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsmZQAS" target="_blank">W-039742</a>

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -10,7 +10,9 @@ import {
     handleError,
     getRecordFieldNames,
     setRecordValuesOnTemplate,
-    checkPermissionErrors
+    checkPermissionErrors,
+    CONTACT_FIRST_NAME_INFO,
+    CONTACT_LAST_NAME_INFO, DI_CONTACT1_IMPORTED_INFO,
```
psadie-sfdo: Minor comment, new line for `DI_CONTACT1_IMPORTED_INFO`.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -21,6 +23,7 @@ import {
     isUndefined,
     checkNestedProperty,
     arraysMatch,
+    getValueFromDotNotationString,
```
psadie-sfdo: Looks like this isn't used anywhere any longer also.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -15,6 +15,7 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
     @track domain;
     @track visualforceOrigin;
     @track isLoading = true;
+    @api cardHolderName;
```
psadie-sfdo: Requesting that `cardHolderName` actually be an object with `firstName`, `lastName`, and `accountName` as properties instead of a string. Tokenization requires just one string to tokenize (`nameOnCard`), but the purchase call requires both `firstName` and `lastName`.

Not sure how we want to handle this situation when making a purchase call for an Account. We _could_ populate either of those name fields with the Account Name and the other with any string in this scenario, but it seems wrong to do that if the request is requiring a first and last name.

@ptewson-sfdo @vmartinov-sforg @voduyemi

voduyemi: @psadie-sfdo I have made updates to return an object with the `firstName`, `lastName` and `accountName`. Let me know if this works for you.

psadie-sfdo: @voduyemi - Thanks for making this change! We're good from my perspective regarding this fabricated cardholder name for tokenization. Let's chat about the purchase call end of things for organization accounts with the broader team.

PR #5244, opened 2020-03-12
Load GAU widget when opening a donation from Review Donations

dfuller114:
GAU allocations will now load into the batch gift entry form when a donation is selected from within the review donations modal.

# Critical Changes

# Changes
## Gift Entry Enhancements
- Any existing GAU Allocations will be loaded into your Gift Entry form when you select an existing Opportunity with the Review Donations feature.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_FormRendererService.cls
```diff
@@ -218,12 +218,26 @@ public with sharing class GE_FormRendererService {
         return null;
     }
 
+    @AuraEnabled(Cacheable=true)
```
bdvorachek: Small nitpick, lets lower case the `cacheable`

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
```diff
@@ -153,79 +161,183 @@ export default class GeFormWidgetAllocation extends LightningElement {
...
         let element = {};
-        element.key = this.rowList.length;
         const record = { apiName: ALLOCATION_OBJECT.objectApiName, ...properties};
         let row = {};
         if(isDefaultGAU === true) {
             // default GAU should be locked.
             element.disabled = true;
             row.isDefaultGAU = true;
             record[GENERAL_ACCOUNT_UNIT] = this.allocationSettings[ALLOC_SETTINGS_DEFAULT];
+            // this.rowList.push(row);
```
bdvorachek: Little clean-up here, lets delete this.

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
```diff
@@ -153,79 +161,183 @@ export default class GeFormWidgetAllocation extends LightningElement {
...
 
     /**
-     * Add a new record to the list
+     * @description Build a new row object using both record and element properties
      * @param isDefaultGAU {boolean} When initializing the first row, this should be true.
+     * @param properties {array} Array of record properties to add to the record object
+     * @return Object representing a widget row
      */
-    addRow(isDefaultGAU, properties) {
+    getRow(isDefaultGAU, properties) {
```
bdvorachek: `getRow` implies a row already exists to me, how about `buildRow` or `generateRow`?

vmartinov-sforg:
@dfuller114 Can you please lurch the W-039310.

dfuller114:
**lurch attach W-039310

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hs2hQAC" target="_blank">W-039310</a>

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
```diff
@@ -153,79 +161,183 @@ export default class GeFormWidgetAllocation extends LightningElement {
         this.init();
     }
 
+    /**
+     * @description Contains the load behavior for the widget. The load behavior depends on the format of the data
```
kenlewis-sfdo: Also a little nitpicky :) but this comment and the other new comments/descriptions below could be formatted to be a little narrower ( I think we try to break at 120 char wide?)

src/classes/GE_FormRendererService.cls
```diff
@@ -218,12 +218,26 @@ public with sharing class GE_FormRendererService {
...
+    public static List<Allocation__c> getDonationAllocations(Id donationId) {
+        if(donationId == null) {
+            return null;
```
vmartinov-sforg: I don't think this causes any issues, but is this a common practice for a method that returns a list to return null? Should we return an empty list instead?

dfuller114:
posting a comment from @kenlewis-sfdo:

"Hey DanielI'm testing out your branch an bit and I saw that Vlad had a comment about an error when updating an Pmt. I'm also seeing a little odd behavior related to Pmts. I might have been running an uncommon scenario, but I created an opp with a few pmts and a few allocations - then set the default gau to one of the gaus used on the opp - but when I select one of those pmts in review donations it looks overallocated. I think maybe it is because the Donation Amount gets overwritten with the Payment amount when a Pmt is selected in Review Donations."
image

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
dfuller114: fixed

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
dfuller114: I think this is actually under 120 chars actually. this is a screenshot from Intellij and that gray line is at 120 chars in the column.
Screen Shot 2020-03-16 at 7 04 53 PM

src/classes/GE_FormRendererService.cls
dfuller114: fixed

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
dfuller114: fixed

src/classes/GE_FormRendererService.cls
dfuller114: fixed

dfuller114:
> posting a comment from @kenlewis-sfdo:
> 
> "Hey DanielI'm testing out your branch an bit and I saw that Vlad had a comment about an error when updating an Pmt. I'm also seeing a little odd behavior related to Pmts. I might have been running an uncommon scenario, but I created an opp with a few pmts and a few allocations - then set the default gau to one of the gaus used on the opp - but when I select one of those pmts in review donations it looks overallocated. I think maybe it is because the Donation Amount gets overwritten with the Payment amount when a Pmt is selected in Review Donations."
> 
> image

@ptewson-sfdo see @ameyersalesforce's comment regarding the above item.
Screen Shot 2020-03-16 at 7 13 19 PM

melissabarber:
@dfuller114 I updated the PR description some for the release notes. Is what I have still accurate and complete?

dfuller114:
> @dfuller114 I updated the PR description some for the release notes. Is what I have still accurate and complete?

@melissabarber One thing we might want to add is that those GAU allocations will be hidden if "Update Payment" is selected since we don't support allocations for payments currently. I've come to learn that "donation" can mean an opportunity or a payment so we may just want to clarify on that.

melissabarber:
Thanks, @dfuller114. I think I'll run that addition past CCE to see if we want to call out Payment Allocation specifically. I have the release note specifically talking about Opportunities instead of donations, but perhaps that's not enough.

dfuller114:
> Oh, one more question @dfuller114. When you say the GAU Allocation field bundle is hidden when a Payment was selected, do you mean it's completely not visible on the page or is it greyed out. Have you run this UI behavior by @gjang-sfdo?

The widget is hidden completely for updating payments. It's looking like we will have to do it that way at least for now because the calculation doesn't work correctly with what's visible since we are using the opportunity donation amount to recalculate totals in the widget so when payment is loaded instead it will always show an over allocation error.

melissabarber:
Darn. Looks like @gjang-sfdo is out of office until Wednesday. @dfuller114 @vmartinov-sforg and @ptewson-sfdo please follow up with Grace when she's back to see if we need a follow-on WI for UI work if she has different recommendations.

force2b:
@dfuller114 I noticed this PR has been around for around 9 months and has a bunch of merge conflicts. Do you know if it's still needed?

dfuller114:
> @dfuller114 I noticed this PR has been around for around 9 months and has a bunch of merge conflicts. Do you know if it's still needed?

@force2b going to close this one. We never ended up adding it as a fix.

force2b:
Thanks @dfuller114

force2b:
@dfuller114 It does look like the branch still exists for this, which seems odd.

PR #5251, opened 2020-03-16
Gift Entry: add global method to set secrets 

vmartinov-sforg:
# Critical Changes

# Changes
* Added stub for Set Configuration method to be called from PS Integration package

# Issues Closed

# Community Ideas Delivered

# New Metadata
* PS_IntegrationService.cls
* PS_IntegrationService_TEST.cls

# Deleted Metadata


ptewson-sfdo:
@vmartinov-sforg this should probably point to the feature branch - feature/single-gift-payments-phase1

vmartinov-sforg:
@ptewson-sfdo  I can certainly do that, but I thought we want to deliver this sooner so the integrations team is 'unblocked'.

src/classes/PS_IntegrationServices.cls
```diff
@@ -0,0 +1,48 @@
...
+global class PS_IntegrationServices {
+    
+    /*******************************************************************************************************
+    * @description Provides an endpoint for the Payment Services Integrations package
+    * to call when a Customer registers, and pass parameters containing the secrets
+    * needed for calls to Payment Services endpoints(Tokenize, Purchase) and creating JWTs
+    * @param secrets contains secrets, such as the api key for creating a JWT
+    */
+    global static void setSecrets(Map<String, List<Map<String, String>>> secrets) {
+
```
ptewson-sfdo: Maybe add in a TO-DO here to make it clear that this will be implemented in the future? Especially if we are going to merge it into master, let's be clear about our intent there

src/classes/PS_IntegrationServices_TEST.cls
```diff
@@ -0,0 +1,49 @@
...
+@isTest
+private class PS_IntegrationServices_TEST {
+    
+    /*******************************************************************************************************
+     * @description tests storing of the secrets we receive from the Payment Services
+     * Integrations package
+     */
+    @isTest
+    private static void setSecretsTest() {
+
```
ptewson-sfdo: Again, let's add a clear TO-DO comment here to explain why this isn't being implemented yet

src/classes/PS_IntegrationServices.cls
ptewson-sfdo: Also - maybe want to re-think the name of this. Not everything is a secret, and it's more like "setOnboardingData" or "setElevateConfig" or something like that

PR #5256, opened 2020-03-17
Feature/robot  settings pom

skristem:
Converting tests in NPSP settings test suite to POM
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/CustomSettingsPageObject.py
```diff
@@ -0,0 +1,37 @@
...
+           identified by value for the given setting and unselects frame"""
+        self.npsp.choose_frame("Custom Settings")
+        locator = npsp_lex_locators['custom_settings']['link'].format(setting_name,value)
+        self.selenium.wait_until_element_is_visible(locator, timeout=60)
+        self.selenium.scroll_element_into_view(locator)
+        self.selenium.click_element(locator)
+        self.selenium.unselect_frame()
+        self.salesforce.wait_until_loading_is_complete()
+           
+    def verify_current_page(self,page_name):
```
rjanjanam: The keyword name is not matching the description mentioned:" verify page contains page_name and select frame identified with page_name". Please change it to be something like "verify_and_select_frame"

robot/Cumulus/resources/NPSPSettingsPageObject.py
```diff
@@ -26,7 +26,53 @@ def open_main_menu(self,title):
...
+        self.selenium.set_focus_to_element(locator)       
+        self.selenium.select_from_list_by_label(loc,value)
+    
+    def edit_selection(self,list_name,value):
+        """waits for edit mode and selects value from the list"""
+        save_button=npsp_lex_locators['npsp_settings']['batch-button'].format('idPanelCon','Save')
+        self.selenium.wait_until_page_contains_element(save_button,
+                                                       error=f"Edit mode is not active")
+        self.select_value_from_list(list_name,value)
+        time.sleep(2) # waiting for 2 seconds in case there is slowness
```
rjanjanam: Can this time. sleep be replaced with wait_for_element or wait_until_element_to_be  enabled disabled ?

skristem: there is no change in state of any elements on the page until I hit save but I noticed in some cases in metaci slect value from list takes a second or two to select the value so have to add hardcoded sleep to avoid flakiness

robot/Cumulus/resources/NPSPSettingsPageObject.py
```diff
@@ -26,7 +26,53 @@ def open_main_menu(self,title):
         self.selenium.wait_until_page_contains(title, 
                                                error=f"{title} link was not found on the page")  
         self.npsp.click_link_with_text(title)
+        locator=npsp_lex_locators["npsp_settings"]["main_menu"].format(title)
+        self.selenium.wait_until_page_contains_element(locator, 
+                                               error=f"click on {title} link was not successful even after 30 seconds")
         
+    def open_sub_link(self,title):  
+        """clicks on the button on the payments page"""
```
sgovindankutty: Is this a link or button. If it is a link can you please update the documentation accordingly

robot/Cumulus/resources/CustomSettingsPageObject.py
skristem: since the operation of selecting a frame is something that should be invisible, I named the keyword as verify current page but gave the details in the description so that anyone who wants to reuse this keyword can read and understand what it can do. Do you think this will cause confusion?

robot/Cumulus/resources/NPSPSettingsPageObject.py
skristem: I repurposed a keyword and forgot to update the documentation. Will update it

robot/Cumulus/resources/CustomSettingsPageObject.py
rjanjanam: As an end user it was not quite evident when I wanted to use the keyword. Feel changing the keyword would help.

skristem: ok will change to verify page and select frame

PR #5258, opened 2020-03-17
Expose Payment Allocations Functionality to Abacus

jbrandolinisf:
# Critical Changes

# Changes
- Replicates Trigger logic for Opportunity and Payment Allocations in a Utility class called ALLO_AllocationsUtil
- Exposes Payment Allocations Functionality to Abacus Batch via Callable API

NOTE: A future work item will begin to share this code with the existing Opportunity and Payment Allocation Triggers, where appropriate

# Issues Closed

# Community Ideas Delivered

# New Metadata
Classes
- ALLO_AllocationsDMLService
- ALLO_AllocationsDMLService_TEST
- ALLO_AllocationsRecalculateService
- ALLO_AllocationsRecalculateService_TEST
- ALLO_AllocationsRetrievalService
- ALLO_AllocationsRetrievalService_TEST
- ALLO_AllocationsReviewService
- ALLO_AllocationsReviewService_TEST
- ALLO_AllocationsService
- ALLO_AllocationsService_TEST
- ALLO_AllocationsSettings
- ALLO_AllocationsSettings_TEST
- ALLO_AllocationsUTIL
- ALLO_AllocationsUTIL_TEST
- ALLO_AllocationsWrapper
- ALLO_UnitTestHelper_TEST

Custom Labels
- alloDefaultGAUNotActive
- alloNullZeroInvalidStateError
- alloOverallocatedError
- alloPaymentAllocationBatchRecordError
- alloPaymentAllocationBatchRetriesExceeded
- alloSuccessfulRecordRetry
# Deleted Metadata


jbrandolinisf:
@lparrott - Took a little longer to cleanup than I originally expected but the PR is created and review requested.

jbrandolinisf:
**lurch: attach W-039291

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hs0WQAS" target="_blank">W-039291</a>

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Allocations
+* @group-content ../../ApexDocContent/Allocations.htm
+* @description Utility class to support shared functionality for Allocations
+*/
+public class ALLO_AllocationsUtil {
```
force2b: The class to specify sharing - with, without or inherited

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+
+    /** @description Caches call to UserInfo.IsMultiCurrencyOrganization */
+    private static final Boolean IS_MULTICURRENCY_ORG = UserInfo.isMultiCurrencyOrganization();
+
+    // Lazy initialization on use to reduce describes for Object Labels
+    /** @description Caches The Allocation Object's Label */
+    @testVisible private static String ALLOC_OBJ_LABEL {
+        get {
+            if (ALLOC_OBJ_LABEL == null) {
+                ALLOC_OBJ_LABEL = UTIL_Describe.getObjectLabel('Allocation__c');
```
force2b: I believe it's necessary to apply the namespace to this. See the `ALLO_ManageAllocations_CTRL` class for an example.

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+        PMT_ALLOC,
+        PMT_ALLOC_RECALC,
+        PMT_ALLOC_INITIALIZE,
+        PMT_ALLOC_SYNC_TO_OPP_ALLOC,
+        OPP_ALLOC_SYNC_FINALIZED,
+        OPP_ALLOC_SYNC_NONFINALIZED
+    }
+
+    /** @description Allocations settings. */
+    public static Allocations_Settings__c ALLOCATION_SETTINGS = UTIL_CustomSettingsFacade.getAllocationsSettings();
```
force2b: We generally don't capitalize non-constants like settings object instances. I recommend `allocSettings` instead.

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+* @group-content ../../ApexDocContent/Allocations.htm
+* @description Utility class to support shared functionality for Allocations
+*/
+public class ALLO_AllocationsUtil {
+    /** @description The number of times it will retry the set of Database operations, after first removing any with errors. */
+      private static final Integer NUM_DATABASE_RETRIES = 1;
+
+    /** @description Used as a way to identify parts of Allocation Sync that may or may not be allowed to
+      *execute depending on earlier execution statuses */
+    public enum PROCESS_DEF {
```
force2b:
```suggestion
    public enum processDefintionType {
```

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+            }
+            return EXCLUDED_OPPORTUNITY_TYPES;
+        }
+        private set;
+    }
+
+    /*******************************************************************************************************
+     * @description Inner class to help Manages Record Exclusions for downstream processes, so it can
+     * process as much as possible while skipping what it has to
+     ********************************************************************************************************/
```
force2b: NPSP Generally doesn't use full rows of asterisks in ApexDoc header comments. This might be better as
```suggestion
    /**
     * @description Inner class to help Manages Record Exclusions for downstream processes, so it can
     * process as much as possible while skipping what it has to
     */
```

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
+/*
```
force2b: This class is way too big. It needs be split into discreet classes that perform one and only function for that class, even that turns into 4 different classes, each with tis own test class. There's definitely no benefit in putting all in a single class. If anything that will significantly complicate future maintenance of this. 

Plus, there are a lot of large static methods in the class that each have a lot of parameters. I'd prefer to see a single service class with a builder pattern that that has properties such as withAllocations(), withExclusions(), etc. and then individual instance methods to handle the processing of those records - probably with a lot more abstraction so that methods are much shorter and easier to follow.

src/classes/ALLO_AllocationsUtil.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>46.0</apiVersion>
```
force2b: Api version should be 48

src/classes/ALLO_AllocationsUtil_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>46.0</apiVersion>
```
force2b: Api version should be 48

src/classes/Callable_API.cls
```diff
@@ -94,7 +94,20 @@ global with sharing class Callable_API implements System.Callable {
                 // Since there are multiple actions that can be invoked, the ERR_Dispatch_API will get the remaining params out of the
                 // parameter service depending on which action was invoked
                 return CallableDispatchService_ERR.getInstanceOf(paramService).dispatch();
-
+            } when 'allo.processoppsandpmts' {
```
force2b: Can you add blank lines before the`when` on both of these?

src/classes/UTIL_Query.cls
```diff
@@ -44,20 +44,23 @@ public with sharing class UTIL_Query {
     /** @description Error message when FROM value is not provided */
     private static final String FROM_ENTITY_REQUIRED = 'The FROM value is required.';
 
+    /** @description SELECT field to be returned if requested and the Org is a Multi Currency Enabled Org */
+    public static final String CURRENCYISOCODE_FIELD = 'CurrencyIsoCode';
```
force2b: You can use `UTIL_Currency.CURRENCY_ISO_CODE_FIELD` for the method below rather than defining a new constant

src/classes/UTIL_Query_TEST.cls
```diff
@@ -350,10 +350,36 @@ private class UTIL_Query_TEST {
...
+        };
+
+        String soql = new UTIL_Query()
+            .withSelectFields(selectFields)
+            .withMultiCurrencyField()
+            .withFrom('Opportunity')
+            .build();
+
+        if (UserInfo.isMultiCurrencyOrganization()) {
+            System.assert(soql.contains('CurrencyIsoCode'), 'Multi Currency Org: CurrencyIsoCode should be present. SOQL: ' + soql);
```
force2b: Probably best to use the constant since it's defined and being used for the method

src/labels/CustomLabels.labels
```diff
@@ -2661,6 +2677,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>alloObjectUnsupported</shortDescription>
         <value>The Manage Allocations button is only supported from a Campaign, Opportunity, Payment or Recurring Donation record.</value>
     </labels>
+    <labels>
+        <fullName>alloOverallocatedError</fullName>
+        <categories>Allocation, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when Allocations are cleared because amounts can't convert to percentages</shortDescription>
+        <value>The amount allocated exceeds the {obj_type} amount. Update Allocation amounts so that the total equals the {obj_type} amount.</value>
```
force2b: Is there a reason these use `{obj_type}` rather than the typical pattern of `{0}` used with the `format()` method?

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
...
+
+        // Process Payment Allocations Results
+        ALLO_AllocationsUtil.processAllocationResults(oppIds, ALLO_AllocationsUtil.PROCESS_DEF.PMT_ALLOC, allocResults, recs, recExcl);
+
+        // Determine successfully processed errors
+        Set<Id> successfullyProcessed = ALLO_AllocationsUtil.determineSucessfullyProcessedRecords(oppIds, recExcl);
+
+        if (commitAndClearRecordErrors) {
+            // Clear Retry Pending and add Retry Successful for Records that were successful that have existing errors flagged for Retry
+            recs.recErrs.clearErrors(successfullyProcessed, Label.errErrorRecordRetrySuccessful);
```
force2b:
```suggestion
            recs.recErrs.clearErrors(successfullyProcessed, System.Label.errErrorRecordRetrySuccessful);
```

Also, it seems odd that a label to display a success message would start with `errError`. If this is unique to Allocations, maybe this label is something like `alloSuccesfulRecordRetry`

src/classes/ALLO_AllocationsUtil_TEST.cls
```diff
@@ -0,0 +1,2828 @@
+/*
```
force2b: This also needs to be split up to match the abstractions in the main `ALLO_AllocationsUtil` class when that is broken up.

src/classes/ALLO_AllocationsUtil.cls
```diff
@@ -0,0 +1,1939 @@
+/*
+    Copyright (c) 2019, Salesforce.org
```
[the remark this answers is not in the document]
force2b: The copyright year still needs to be updated.

src/classes/ALLO_AllocationsSettings.cls
```diff
@@ -0,0 +1,123 @@
...
+            problems.add(System.Label.alloPaymentNotEnabled);
+        }
+        if (!isDefaultAllocationsEnabled()) {
+            problems.add(System.Label.alloDefaultNotEnabled);
+        }
+
+        if (allocationsSettings.Default__c == null) {
+            problems.add(System.Label.alloDefaultGAUMissing);
+        } else {
+            List<General_Accounting_Unit__c> defaultGAU = [select Id, Active__c, Name from General_Accounting_Unit__c where Id = :allocationsSettings.Default__c];
```
force2b, line 113:
```suggestion
            List<General_Accounting_Unit__c> defaultGAU = [SELECT Id, Active__c, Name FROM General_Accounting_Unit__c WHER Id = :allocationsSettings.Default__c];
```

src/classes/ALLO_AllocationsService_TEST.cls
```diff
@@ -0,0 +1,853 @@
...
+public class ALLO_AllocationsService_TEST {
+    /**
+     * @description Executes various Opportunity and Payment Allocation Scenarios when
+     *              Payment and Default Allocations are enabled
+     *              [ALLO_AllocationsService.processRecords]
+     *
+     */
+    @isTest
+    private static void confirmProcessRecordsUpdatesAllocationsCorrectlyWhenPaymentAllocationsAreEnabled() {
+        ALLO_UnitTestHelper_TEST.AllocationSetupTest alloSetupTest = new ALLO_UnitTestHelper_TEST.AllocationSetupTest()
```
force2b, line 47: This is a 280 line test method. Ideally it should test one concept at a time. I'm not necessarily able to figure out if it does or not. Is there a way to break this test into more logical chunks that test individual blocks of functionality?

src/classes/ALLO_AllocationsService_TEST.cls
```diff
@@ -0,0 +1,853 @@
...
+        }
+    }
+
+    /**
+     * @description Executes various Opportunity and Payment Allocation Scenarios when
+     *              Payment and Default Allocations are not
+     *              [ALLO_AllocationsService.processRecords]
+     */
+    @isTest
+    private static void confirmAllocationsServiceUpdatesAllocationsCorrectlyWhenDefaultAllocationsAreNotEnabled() {
```
force2b, line 319: This method is another huge individual test method that looks a lot like the large method above. Can these be further abstracted to use common helper methods and/or (better) into smaller tests that test individual functionality?

src/classes/ALLO_AllocationsRecalculateService_TEST.cls
```diff
@@ -0,0 +1,736 @@
...
+public class ALLO_AllocationsRecalculateService_TEST {
+
+     /**
+     * @description Confirm Recalculate Service creates correct DML changes when
+     *              Default Allocations Enabled
+     *              Payment Allocations Disabled
+     *              [ALLO_AllocationsRecalculateService.processOpportunityAllocations]
+     */
+    @isTest
+    private static void confirmProcessOpportunityAllocationsCreatesCorrectPendingDMLChangesWhenDefaultAllocationsEnabledPaymentAllocationsDisabled() {
```
force2b, line 47: Like another test class further down, the test methods in this class are extremely long, likely testing multiple concepts. Can these be split into smaller individual tests that validation individual pieces of functionality? Overall my concern is that these are extremely difficult to follow and understand exactly what's being tested/asserted.

src/classes/ALLO_AllocationsRecalculateService.cls
```diff
@@ -0,0 +1,604 @@
...
+                    allocation.Amount__c = (amount * allocationPercent * .01).setScale(2);
+
+                    // Round the visible percentage value to 2 decimal places after the amount is calculated
+                    allocation.Percent__c = allocationPercent.setScale(2, System.RoundingMode.HALF_UP);
+                }
+                remainder -= (allocation.Amount__c != null) ? allocation.Amount__c : 0;
+            }
+
+            //if our remainder is not the same sign as the amount, only create percentage based Allocations
+            if ((amount >= 0 && remainder < 0) || (amount < 0 && remainder > 0)) {
```
force2b, line 241: The amount/remainder comparisons throughout this rather large if-block make this hard to read/understand. Can these be assigned into vars and then compared in the if? For example:
```
Boolean isRemainderSameSign = (amount >= 0 && remainder < 0) || (amount < 0 && remainder > 0);

if (isRemainderSameSign) {
...
}
```

And so on with the subsequent ifs.

src/classes/ALLO_AllocationsDMLService.cls
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    /**
+     * @description Processes Allocation deletions, recording DML errors as Record Errors and adding
+     * Record Exclusions for ones that errored.  The ALLO_AllocationsWrapper Allocations Maps are updated removing
+     * Allocations from the Maps for successful DML.
+     * @return Boolean, true if there were any DML errors
+     */
+    @testVisible private Boolean processAllocationDeleteResults() {
+        List<Database.DeleteResult> deletes = null;
```
force2b, line 224: How about defining this as `new List<Database.DeleteResult>()` and then you don't need line 233 to compare for not null?

src/classes/ALLO_AllocationsDMLService.cls
```diff
@@ -0,0 +1,413 @@
...
+            // Process Deletes
+            dmlErrors = (dmlErrors || processAllocationDeleteResults());
+
+            // Process Inserts
+            dmlErrors = (dmlErrors || processAllocationInsertResults());
+
+            // Process Updates
+            dmlErrors = (dmlErrors || processAllocationUpdateResults());
+
+            if (dmlErrors) {
```
force2b, line 181: Could this be
```
if (!dmlErrors) {
   continue;
}
````

File: src/classes/ALLO_AllocationsDMLService.cls [hunk 23]
```diff
@@ -0,0 +1,413 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Allocations
+* @description Utility class to handle DML for Allocations
+*/
+public inherited sharing class ALLO_AllocationsDMLService {
+    /** @description The number of times it will retry the set of Database operations, after first removing any with errors. */
+    @testVisible private static Integer numberOfDatabaseRetries = 1;
```
force2b, line 38: Is this a constant that is eventually intended to be exposed? If so, you could consider moving the constant into a getter in the `ALLO_AllocationSettings` class so it's considered a setting from day-1 even if it's not yet enabled as a setting in the system.

src/classes/ALLO_AllocationsReviewService.cls
```diff
@@ -0,0 +1,232 @@
...
+
+        List<Allocation__c> parentAllocs = (!campaignAllocs.isEmpty()) ? campaignAllocs : recurringDonationAllocs;
+
+        if (isRequiresReview(opp, oppAllocs, parentAllocs)) {
+            // As soon as we find at least one reason to review we stop looking
+            return true;
+        }
+        // Payment Allocations are reviewed if Payment Allocations is enabled
+        if (ALLO_AllocationsSettings.getSettings().isPaymentAllocationsEnabled()) {
+            if (opp.npe01__OppPayment__r != null) {
```
force2b, line 114:
```suggestion
        if (!ALLO_AllocationsSettings.getSettings().isPaymentAllocationsEnabled() || opp.npe01__OppPayment__r != null) {
            return false;
        }
```

jbrandolinisf:
@force2b - There is a lot of history behind why this one became what it is.  I am aware we don't like items of this size...I don't like them either.  it was originally broken up into several work items, but even then it could not get reviewed (the reviewer did not have bandwidth-- for valid reasons).  This caused it all to back up.  And become a ginormous PR.  I pared out what I could for this reason, but it was still really, really large.

In terms of renaming them to PMTAlloc, I would prefer not to because the intent of much of the code is to be shared with Opportunity Allocations as well long term (future WI).  We can talk this through further.

src/labels/CustomLabels.labels
```diff
@@ -2783,6 +2831,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Each Allocation must have a single parent object: Pmt, Opp, Rec Donation, Camp.</shortDescription>
         <value>Ensure that this Allocation has only one parent object: Payment, Opportunity, Recurring Donation, or Campaign.</value>
     </labels>
+    <labels>
+        <fullName>alloSuccessfulRecordRetry</fullName>
+        <categories>Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>alloSuccessfulRecordRetry</shortDescription>
+        <value>Allocations reprocessing completed successfully. </value>
```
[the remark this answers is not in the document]
jbrandolinisf: I will ask Mark to review it.  It has not been re-reviewed yet.

src/classes/Callable_API.cls
```diff
@@ -95,6 +95,27 @@ global with sharing class Callable_API implements System.Callable {
                 // parameter service depending on which action was invoked
                 return CallableDispatchService_ERR.getInstanceOf(paramService).dispatch();
 
+            } when 'allo.processoppsandpmts' {
+                // This method is to be used only by Salesforce.org apps and is subject to change at any time. Do NOT call this method.
+                // If you do so, your functionality can break without any warning or liability from Salesforce.
+                // If you have any questions, please inquire on the Power of Us Hub.
+                Set<Id> opportunityIds = paramService.getSetId(CallableApiParameters.PARAM_OPPORTUNITY_IDS);
```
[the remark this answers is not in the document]
jbrandolinisf: Yes.  They are in ALLO_AllocationsService_TEST I believe

src/classes/ALLO_UnitTestHelper_TEST.cls
```diff
@@ -0,0 +1,822 @@
...
+        }
+
+        return allocsByPmtId;
+    }
+
+    /**
+     * @description Helper method to retrieve all error records into a Map by Salesforce Id
+     * @return Map of Error__c records by Related Record ID
+     */
+    public static Map<Id, Error__c> getErrorsByRecordId() {
```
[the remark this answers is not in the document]
jbrandolinisf: I will look at this to see if it will work for this purpose.  It needs to be keyed not by Errror__c.Id but Error__c.Record_Id__c

src/classes/ALLO_AllocationsService_TEST.cls
jbrandolinisf: So...this test, and the others like it are meant to confirm that operations being done in a batch iteration, or multiple records being updated in a trigger context doing different things all come together and properly render the correct allocations.  That is why it is so long.  Testing serially each of the allocations scenarios would be very easy to write, but doesn't really cover a bulkified scenario.  I am already testing each of the individual services that this calls.  It is intended to make sure that the data sent in comes back out properly, depending on the configuration.

src/classes/ALLO_AllocationsService_TEST.cls
jbrandolinisf: See above

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: Let me see why I did that.  I don't remember if there was a good (or even a not good reason).

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: No.  It needs to go through all the dml types before it rollback and filters the results.  If I had an unexpected error in the insert step, and skipped the remainder, and filtered, on the second try I could still hit errors on the update or delete steps.  I have to do all three and then rollback, filter, and retry.

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: I did not intend it to be, no, but I could still move it if you feel strongly about it.

src/classes/ALLO_AllocationsReviewService.cls
jbrandolinisf: Need to look at the block to make sure that does not alter the flow, but it doesn't I will make the change

force2b:
Thanks for the response (and I thought I was up late 😃). It all makes sense, though I'd say that if the reviewer didn't have time to review (and sometimes that applies to me too, sorry), it shouldn't be merged (or new work started) until someone can review/test. We've talked prioritizing code reviews ourselves, and I've been in this spot too. It's all in the past at this point so no worries. If you find yourself in this situation in the future though definitely push back and work with your scrum master, pm and/or lead to make sure that time is allocated for review. And, of course reach out to the reviewer to remind them (like you did with me). Quality (trust) is our number one value, so that always takes precedence over getting something merged.

Anyway, I'm fine with not renaming the classes if you feel these will eventually become part of the larger Allocations logic.

I'll look though your uploaded changes a little later. Question: What's the plan for full regression on Allocations & Payment Allocations before this is merged?

jbrandolinisf:
@force2b - I need to talk with Brian about it.  Unfortunately, this would be easier to test once merged (we have an Abacus PR on top of the Callable Layer from our batch ready to go), but I can take him through what he will have to do to test it without that.

src/classes/ALLO_UnitTestHelper_TEST.cls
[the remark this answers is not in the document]
jbrandolinisf: @force2b - I can either use the ErrorGateway class to get the records, and then still have my local class method to build the map the way I need it, or i can update the Error Gateway and add a getRecordsByRecordId method that uses the getRecords Method to retrieve the data and builds the map.  I am good with either approach.

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: I solved this by combining the two blocks, which eliminated the need for the null check.

src/classes/ALLO_AllocationsReviewService.cls
jbrandolinisf: @force2b - I keep looking at this one, and the change does not appear to do what I need it to.  I need to enter the block and do additional Payment Allocations checks for each payment, if Payment Allocations is Enabled, and if the Opportunity has Payments. If Payment Allocs are required, and the opp has allocs, i have to then review each payments allocations for changes.

jbrandolinisf: Disregrard.  It took me a bit, but I think I figured out what you meant.

jbrandolinisf:
@force2b - Up to date, based on feedback, other than having a doc reviewer check the label.  Let me know if there are any additional changes.  Before we move this to a Ready for QA state (other than doc review)

src/classes/ALLO_AllocationsDMLService.cls
force2b: If it's unique to this class and won't be exposed (other than likely to unit tests for manipulation, I assume), it's fine to stay here.

src/classes/ALLO_AllocationsDMLService.cls
force2b: Yes. Agree that it does need to go through the 3 above, but the large if block below (line 181 through 196) that starts with `if (dmlErrors) {` could be outdented if no errors. Just helps with readability.

src/classes/ALLO_AllocationsDMLService.cls
```diff
@@ -0,0 +1,407 @@
...
+     * Record Exclusions for ones that errored.  The ALLO_AllocationsWrapper Allocations Maps are updated removing
+     * Allocations from the Maps for successful DML.
+     * @return Boolean, true if there were any DML errors
+     */
+    @testVisible private Boolean processAllocationDeleteResults() {
+        List<Database.DeleteResult> deletes = null;
+        boolean dmlErrorsPresent = false;
+
+        // Process Deletions
+        if (allocationsWrapper.allocationsForDelete != null && allocationsWrapper.allocationsForDelete.size() > 0) {
```
force2b, line 228: Since this if block extends throughout the entire method, it could be simplified (and the code below outdented) by changing to:
```suggestion
        if (allocationsWrapper.allocationsForDelete |= null || allocationsWrapper.allocationsForDelete.isEmpty()) {
            return;
        }
```

src/classes/ALLO_AllocationsDMLService.cls
```diff
@@ -0,0 +1,407 @@
...
+     * Record Exclusions for ones that errored.  The ALLO_AllocationsWrapper Allocations Maps are updated adding all
+     * successful DML.
+     * @return Boolean true, if there were any dml errors
+     */
+    @testVisible private Boolean processAllocationInsertResults() {
+        List<Database.SaveResult> inserts = null;
+        Boolean dmlErrorsPresent = false;
+
+        // Process Inserts
+        if (allocationsWrapper.allocationsForInsert != null && allocationsWrapper.allocationsForInsert.size() > 0) {
```
force2b, line 262: Same here with the check for a return followed an outdent.

src/classes/ALLO_UnitTestHelper_TEST.cls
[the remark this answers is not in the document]
force2b: I like the idea of centralizing the query itself in the SObjectGateway and then restructuring into a map in this class.

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: I will make this change in the 3 methods, and then tag you.

jbrandolinisf: Done

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: Done

src/classes/ALLO_AllocationsDMLService.cls
jbrandolinisf: @force2b - I pushed the items in the if block into their own method.  That should help readability

force2b: Even better.

jbrandolinisf:
@force2b - fyi, I am done with all updates, including for doc.  Anything after this will be a result of QE testing, which Brian Kayser is starting work on.

PR #5268, opened 2020-03-20
Non-Selective Query Error for Opportunity Cascade Delete

npsp-reedestockton:
Refactored the query identified as the source of non-selective query error to remove unnecessary OR clause unless specifically required for payment allocations.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
***lurch: attach W-037942*

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HpWTQA0" target="_blank">W-037942</a>

npsp-reedestockton:
@sam-knox Description is the only thing here for you to look at, Sam.  Thanks!

src/classes/OPP_CascadeDeleteLookups_TDTM.cls
```diff
@@ -62,11 +62,12 @@ public with sharing class OPP_CascadeDeleteLookups_TDTM extends CDL_CascadeDelet
         * @return List<Allocation__c> Deleted Opportunities' Allocations.
         ********************************************************************************************************/
         public List<Allocation__c> load(Set<Id> opportunityIds) {
-            return [
-                SELECT Opportunity__c
-                FROM Allocation__c
-                WHERE Opportunity__c IN :opportunityIds OR Payment__r.npe01__Opportunity__c IN :opportunityIds
-            ];
+
```
force2b: @npsp-reedestockton I talked to @jbrandolinisf about this. He had a great suggestion to ensure that the query/logic is both scalable and handles a possible scenario where PaymentAllocations is enabled in an org and then disabled at a later time.

The change is effectively to:
- Query all Payment records for the set of OpporutnityIds to build a set of PaymentIds
- Modify the Allocation to use:
   - `WHERE Opportunity__c  IN :opportunityIds OR Payment__c IN :paymentIds`

The downside of this is that adds an additional query to the logic. The upside is that it is guaranteed to get any PaymentAllocation records in the event this functionality was enabled and then disabled. Plus, Jon pointed out that the query against `Payment__r.npe01__Opportunity__c` could actually be the root cause of the non-selective query error because it might be causing a table scan. I do feel this is mostly a guessing game for us though. In the mean time I like the idea of getting the paymentIds first because it addresses the possible scenario of PaymentAllocations being disabled in an org after it was enabled, and it it _may_ be more selective. Of course, you may be able to find some additional data in your query tests in the LDV-XXL org once it's online.

npsp-reedestockton: I am MUCH happier with the two query approach.  We are unlikely to be pushing the limits in this context, so while it will be a little less efficient, it feels a lot less likely to cause issues.

npsp-reedestockton: Restored to the version @lparrott originally wrote.  I'm happy with this outcome.  Given the location of the non-selective query error, I think there is a fair chance this was the cause of the problem.

force2b:
@npsp-reedestockton One question on this. I see the trigger doesn't delete records until the After Delete context. Since we're querying Payment records, which are deleted automatically when the Opp is deleted, do you think it's possible that that query will not actually return any records. Worth double-checking.

npsp-reedestockton:
@force2b On your question about the query possibly not returning records...  From my reading of the code, load() gets called in the Before Delete context and that's where this query runs.  The deletes don't happen until the After Delete context.  If I'm reading this incorrectly, please let me know.

force2b:
@npsp-reedestockton. It's hard to tell to be honest, but what I do see is a second query in that class that might need to be updated as well. It has the same OR condition, but adds an `IsDeleted=True` clause. https://github.com/SalesforceFoundation/NPSP/pull/5268/files#diff-866f24596bfe4c4d925ff93db23bb32cL83

npsp-reedestockton:
@force2b Good catch!  That second query is the undelete context, but it should be changed as well.  Have pushed that change.

npsp-reedestockton:
@force2b Could you please take a critical look at this? 

```
            List<npe01__OppPayment__c> relatedPayments = [
                    SELECT Id
                    FROM npe01__OppPayment__c
                    WHERE npe01__Opportunity__c IN :opportunityIds
            ];
```

This is the first of the two query set in the undelete context.  I restored the previous version, but shouldn't this query also include the IsDeleted=True clause?  All of those payment records should have been deleted I think.

force2b:
@npsp-reedestockton I was wondering the same thing, however I'm not sure that it's needed if the sequence of events is like follows. 

1. Undelete the Opportunity, which automatically undeletes child records in an MD-relationship: OCR's, Payments, PartialSoftCredit, and AccountSoftCredit.
2. After Trigger runs to undelete Allocations (both types)

The assumption is that this is all happening in the After trigger. I believe the Unit Tests in [PMT_CascadeDeleteLookups_TEST](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/PMT_CascadeDeleteLookups_TEST.cls) should be testing all of this, but the absolute test is to actually try it in a scratch org.

npsp-reedestockton:
@force2b If your logic is correct, then the first of this two query set is unneeded.  Seems to me that it is either needed and should use IsDeleted=True, or it is superfluous.  In any case, querying for npe01__OppPayment__c records that have been deleted along with the opportunity doesn't make sense.

npsp-reedestockton:
@force2b Testing confirms that I am 100% wrong.  The query needs to stay is is.  Tests validate that it works.

PR #5279, opened 2020-03-25
Gift Entry: Purchase call

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_GiftEntryController.cls
```diff
@@ -114,11 +116,11 @@ public with sharing class GE_GiftEntryController {
     * @return The Id of the new Opportunity record, created from the Data Import
     */
     @AuraEnabled
-    public static Id saveAndProcessDataImport(DataImport__c diRecord, String widgetData, Boolean updateGift) {
+    public static Id saveAndProcessDataImport(DataImport__c diRecord, Boolean updateGift) {
 
         // Check for widget data and convert it to the correct format for processing
-        String additionalObjectJSON = GE_FormRendererService.convertWidgetDataToObjectJSON(widgetData);
-        diRecord.Additional_Object_JSON__c = additionalObjectJSON;
+        diRecord.Additional_Object_JSON__c =
```
bdvorachek: This works, but we're sending the Additional_Object_JSON__c field to the back-end in a different format from what it is ultimately stored in. I'm not sure what the better way to go about this is, maybe another DI field that accepts the pre-transformation widgetData?

psadie-sfdo: Hmm good point. I was trying to reduce the number of things we're passing into apex, but that might preferable to adding an additional field to the DI object.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,95 @@
...
-                    Object.assign(dataImportModel.dataImportRows[0],
-                        dataImportModel.dataImportRows[0].record);
-                    table.upsertData(dataImportModel.dataImportRows[0], 'Id');
-                    table.setTotalCount(dataImportModel.totalCountOfRows);
-                    table.setTotalAmount(dataImportModel.totalAmountOfRows);
-                    event.detail.success(); //Re-enable the Save button
-                }
-            )
+        const dataImportRecord = event.detail.dataImportRecord;
+        GeFormService.saveAndDryRun(this.recordId, dataImportRecord)
```
kenlewis-sfdo: The three methods called in GeFormService (saveAndDryRun, handlePaymentProcessing, and handleProcessDataImport) are all only called from this component.  Could they each be imported by geGiftEntryFormApp directly so these call sites would reference the imported methods as opposed to calling methods on the GeFormService component's controller?

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -1,9 +1,8 @@
 import { LightningElement, track, api } from 'lwc';
 import GeLabelService from 'c/geLabelService';
+import { registerListener, unregisterListener, fireEvent } from 'c/pubsubNoPageRef';
```
kenlewis-sfdo: It looks like maybe registerListener and unregisterListener are unused in this component?  If that is correct can they be removed?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1213,6 +1293,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
             sectionsList.forEach(section => {
                 if (section.isCreditCardWidgetAvailable) {
                     section.setCardHolderName(this.fabricateCardHolderName(fieldList));
+                    GeFormService.setFabricatedCardholderNames(this.fabricateCardHolderName(fieldList));
```
kenlewis-sfdo: If those other methods from GeFormService are moved into the App, then this call might need to be reworked also as it might not make sense to set the names on the Service component any longer after that.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -761,19 +833,27 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+    * @description Builds a full DataImport__c record from the provided form sections
+    * and potential donor data selected from the review donations modal.
+    *
+    * @param {list} sections: List of all form sections
+    * @param {object} dataImportWithDonorData: Object holding data import values from
+    * the 'Review Donations' modal.
+    */
+    buildDataImportFromSections(sections, dataImportWithDonorData) {
+        let dataImportRecord =
+            GeFormService.buildDataImportRecord(sections, dataImportWithDonorData);
```
kenlewis-sfdo: Can buildDataImportRecord be imported here directly rather than imported by and referenced in the GeFormService cmp?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -265,55 +268,91 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+    /*******************************************************************************
+    * @description Handles upserting of a Data Import record and firing a
+    * 'singlesubmit' event.
+    *
+    * @param {object} dataImportRecord: A DataImport__c record
+    * @param {object} formControls: An object holding methods that control
+    * the form save button enablement and lightning spinner toggler.
+    */
+    handleSaveSingleGiftEntry = async (dataImportRecord, formControls) => {
+        this.dataImport = await GeFormService.saveDataImport(dataImportRecord)
```
kenlewis-sfdo: Can saveDataImport be imported directly here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.html
```diff
@@ -11,10 +11,12 @@
     
     <c-ge-form-renderer
             batch-id={recordId}
-            onsubmit={handleSubmit}
+            onbatchsubmit={handleBatchSubmit}
+            onsinglesubmit={handleSingleSubmit}
```
kenlewis-sfdo: I don't know if it is doable in this PR, but I'd love to see this be a single handler eventually for `onsubmit` since the App should already know whether it is in batch or single mode and what to do next in each case.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,95 @@
...
+    * callback for handling and displaying errors in the form.
+    */
+    handleSingleSubmit = async (event) => {
+        try {
+            let { dataImportRecord, hasUserSelectedDonation, errorCallback } = event.detail;
+
+            const hasPaymentToProcess = dataImportRecord[PAYMENT_AUTHORIZE_TOKEN__C];
+            if (hasPaymentToProcess) {
+                this.loadingText = 'Charging card...';
+                dataImportRecord = await GeFormService.handlePaymentProcessing(
```
kenlewis-sfdo: Sometimes "handle" signals to me that an event is being handled after an action happened.  Might this be better named "doPurchase" or "makePurchaseCall" or something to indicate that this is initiating the payment process?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
psadie-sfdo: Agree! Importing these directly into `geGiftEntryFormApp` now.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
psadie-sfdo: Removed!

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Updated this to set a public property in `geFormRenderer` that `geGiftEntryApp` can retrieve.

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Yup! Moved directly into `geFormRenderer` since it was referenced in this one place anyway.

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Yes, updated!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.html
psadie-sfdo: Rolled this back into a single `onsubmit`. Checking the `isBatchMode` getter within the submit handler to know which path to go down.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
psadie-sfdo: Agree! Moved the `handlePaymentProcessing` method out of `geFormService` and importing it directly in `geGiftEntryFormApp`. Also renamed it to `makePurchaseCall` and renamed the apex method it then calls to `createTransaction`.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,321 @@
...
+    REFUNDISSUED: 'REFUNDISSUED'
+});
 
-export default class GeGiftEntryFormApp extends LightningElement {
+export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement) {
     @api recordId;
     @api sObjectName;
 
     @track isPermissionError;
+    @track loadingText = 'Saving data import...';
```
vmartinov-sforg, line 42: Could we use a custom label here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,321 @@
...
+    * an Elevate token for a purchase call if needed.
+    *
+    * @param {object} inMemoryDataImport: DataImport__c object built from the form
+    * fields.
+    *
+    * @return {object} dataImportRecord: A DataImport__c record
+    */
+    saveDataImport = async (inMemoryDataImport) => {
+        if (this.previouslySavedDataImport.Id) {
+            this.loadingText = 'Re-saving Data Import record...';
```
vmartinov-sforg, line 162: Could we use a custom label here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,321 @@
...
+
+            inMemoryDataImport.Id = this.previouslySavedDataImport.Id;
+            inMemoryDataImport[PAYMENT_METHOD__C] = this.previouslySavedDataImport[PAYMENT_METHOD__C];
+            inMemoryDataImport[PAYMENT_STATUS__C] = this.previouslySavedDataImport[PAYMENT_STATUS__C];
+            inMemoryDataImport[PAYMENT_DECLINED_REASON__C] =
+                this.previouslySavedDataImport[PAYMENT_DECLINED_REASON__C];
+        } else {
+            this.loadingText = 'Saving Data Import record...';
```
vmartinov-sforg, line 170: Could we use a custom label here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,321 @@
...
+    * Services iframe (GE_TokenizeCard).
+    *
+    * @param {object} dataImportRecord: A DataImport__c record
+    * @param {function} errorCallback: An error handler function. Toggles
+    *   the form save button, lightning spinner, and catches aura exceptions.
+    *
+    * @return {object} dataImportRecord: A DataImport__c record
+    */
+    startPaymentProcessing = async (dataImportRecord, errorCallback) => {
+        this.loadingText = 'Charging card...';
```
vmartinov-sforg, line 192: Could we use a custom label here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,321 @@
...
+    *
+    * @param {object} dataImportRecord: A DataImport__c record
+    * @param {boolean} hasUserSelectedDonation: True if a selection had been made in
+    * the 'Review Donations' modal.
+    * Determines BDI matching criteria.
+    *   true = "single match or create" and means we are updating
+    *   false = "do not match"
+    */
+    startDataImportProcessing = (dataImportRecord, hasUserSelectedDonation, errorCallback) => {
+        this.loadingText = 'Processing data import record...';
```
vmartinov-sforg, line 302: Could we use a custom label here?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
psadie-sfdo: @vmartinov-sforg - updated with custom labels 👍

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,396 @@
...
+
+                    let errors = this.getFailedPurchaseMessage(purchaseResponse);
+                    this.errorCallback(errors);
+                    return;
                 }
-            )
+            }
+        }
+
+        this.processDataImport();
```
ptewson-sfdo, line 230: Should this call to processDataImport() get moved outside of the purchase call, or do they need to be bundled for some reason? It maybe just reads a little cleaner that way, whereas the current if/else on hasPaymentToProcess makes it look like we are *either* processing payment or processing data import, when in fact we process the DI in both cases. BUT, if it complicates the flow to move the processDataImport() call out of this method, let's keep it here!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,396 @@
...
+        submitDataImportToBDI({ diRecord: this.dataImportRecord, updateGift: this.hasUserSelectedDonation })
+            .then(opportunityId => {
+                this.loadingText = this.CUSTOM_LABELS.geTextNavigateToOpportunity;
+                this.navigateToRecordPage(opportunityId);
+            })
             .catch(error => {
-                event.detail.error(error);
+                // TODO: Placeholder method for handling the scenario in which
+                // we've charged a card successfully, but BDI processing failed.
+                this.doSomethingLoud(error);
```
ptewson-sfdo, line 377: This placeholder is great!!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -1,37 +1,396 @@
...
             });
     }
 
+    // TODO: Placeholder method, remove/replace in work item to better handle purchase call errors
+    doSomethingLoud(error) {
+        this.errorCallback(error);
+        // TODO: Potentially have to check for other types of status that MAY
+        // indicate a charge could still occur (pending, authorized, etc)
+        if (this.dataImportRecord[PAYMENT_STATUS__C] === PAYMENT_TRANSACTION_STATUS_ENUM.CAPTURED) {
+            showToast('HEY!',
```
ptewson-sfdo, line 387: @melissabarber - this exact behavior is being defined still, but FYI we will need to figure out the exact messaging for this scenario at some point. I'd love to start the conversation but I think we'll have to resolve it in a future sprint but just wanted to keep you in the loop!

src/lwc/pubsubNoPageRef/pubsubNoPageRef.js
```diff
@@ -58,7 +58,7 @@ const unregisterAllListeners = thisArg => {
  * Fires an event to listeners.
  * @param {object} pageRef - Reference of the page that represents the event scope.
  * @param {string} eventName - Name of the event to fire.
- * @param {*} payload - Payload of the event to fire.
+ * @param {*=} payload - Payload of the event to fire.
```
ptewson-sfdo, line 61: Was this change intentional?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -761,19 +828,69 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+                    diRecord[key] = value === null || value.value === null ?
+                        null : value.value || value;
+                }
+            }
         }
 
-        if (this.dataImport) {
-            diRecord.Id = this.dataImport.Id;
+        // Set Data Import's additional object json field if,
+        // we values from any form widgets
```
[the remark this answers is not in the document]
psadie-sfdo: 🤣 That's a goofy one that slipped past me.  I'll clear this up. This should read something like, 'Set Data Import's additional object json field if we have any useable widget values.'

src/lwc/pubsubNoPageRef/pubsubNoPageRef.js
psadie-sfdo: Nope!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
ptewson-sfdo: @psadie-sfdo should the 'Saving data import...' text still be here if we are using custom labels?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -70,6 +73,9 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     @api donorRecordId;
     @api donorApiName;
     @api donorRecord;
+    @api fabricatedCardholderNames;
+
+    @api loadingText;
```
ptewson-sfdo, line 78: Does loadingText need to be in geFormRenderer? It looks like it's now being used in the App....

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
psadie-sfdo: It sure should be. Updating for the next commit.

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Yep! Only the App is aware of where we are in the save flow and so we have to ferry that info into the renderer for the UI through this public `loadingText` property.

psadie-sfdo: @ptewson-sfdo

psadie-sfdo:
**lurch: attach W-039749

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hsn8QAC" target="_blank">W-039749</a>

src/objects/DataImport__c.object
```diff
@@ -1263,6 +1263,18 @@
         <trackTrending>false</trackTrending>
         <type>Lookup</type>
     </fields>
+    <fields>
+        <fullName>Payment_Authorization_Token__c</fullName>
```
Julian88Tex, line 1267: @psadie-sfdo @ameyersalesforce curious if these layout changes were intentional

PR #5283, opened 2020-03-25
RD2: Copy Installment Amount into fixed-length RD Amount when Schedule Type is null

RadSFDeveloper:
Amount on RD should be the same as the Installment Amount when Schedule Type is null for a fixed-length RD after RD2 enablement migration and RD2 nightly batch jobs are executed.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-040258

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Htn6QAC" target="_blank">W-040258</a>

PR #5284, opened 2020-03-26
Gift Entry: Make Picklists Support Record Types

### LINKED ISSUE #4912: Advanced Mapping Component Error in Sandbox
**Issue**
In sandbox org, clicking on _NPSP Settings > System Tools > Data Import Advanced Mapping_ (once it has been enabled), turn on _Setup > Develop > Lightning Components > Debug Mode_. When attempting to access Data Import Advanced Mapping again the pictured error is displayed:

image

**Steps to Replicate**
- Create Sandbox Org
- In NPSP Settings > System Tools > enable the Data Import Advanced Mapping
- In Setup > Develop > Lightning Components > enable Debug Mode for your user
- Try to access Data Import Advanced Mapping again and see the pictured error above

**Workaround**
- Disable Debug Mode for Lightning Components and access to the Data Import Advanced Mapping will be restored.

### LINKED ISSUE #5250: Gift Entry: Picklist field values are not getting populated when selecting a record in Gift entry
When we select any object in the Gift entry form, all the related values should get populated. However, We do not see picklist fields resonate with the same behavior. This is also true for template level/batch level default field values.

Screen Shot 2020-03-06 at 3 45 06 PM

**Steps to Reproduce**
1. Add Opportunity Record Type ID onto a Gift Entry template
2. When filling out the form, choose an Opportunity Record Type from the picklist in the field Opportunity: Record Type ID
3. Save the entry
4. See that the field on the Dry Run record on the batch shows the Opportunity: Record Type ID, not the text string
5. The record you saved on the batch displays under Status: "Dry Run - Error" and under Error: "Invalid Donation Record Type Name"

image

**Expected Behaviour**
The submitted record should properly translate the selected Record Type to the string the NPSP Data Importer expects, instead of converting it to the ID.

**Workarounds**
Manually updating the field on the NPSP Data Import records, and then returning to the Gift Entry batch and running "Batch Dry Run" again will update the Status to "Dry Run - Validated" and allow you to proceed with the import as expected.

kenlewis-sfdo:
This PR:
- manages the value for the DataImport Donation Record Type Name field so that the RecordType Name is saved to the record and displayed in the UI, and the RecordType Id is used to set the recordTypeId property on sibling picklist fields, and
- makes Gift Entry Picklists set their options in response to being assigned a RecordType Id value

Additionally:
- fixes an issue where the Review Donations modal would throw an exception when closed with LWC debug mode enabled
- enables the Review Donations modal link to appear on form load when Donation Donor and Account or Contact donor fields have default values
- enables related fields to be populated on form load when lookup fields have default values

# Critical Changes

# Changes
- In addition to the Issues fixed with this release, Gift Entry now updates picklist values to match the options assigned to different record types.

# Issues Closed

- Fixes #5250 
- Fixes #4912 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
**lurch: attach W-040105

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HtSEQA0" target="_blank">W-040105</a>

kenlewis-sfdo:
**lurch: attach W-040279

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HtosQAC" target="_blank">W-040279</a>

src/aura/GE_GiftEntryForm/GE_GiftEntryFormController.js
```diff
@@ -46,9 +46,7 @@
                         header: payload.modalProperties.header || '',
                         showCloseButton: payload.modalProperties.showCloseButton || true,
                         cssClass: component.getName() + ' custom-modal ' + payload.modalProperties.cssClass,
-                        closeCallback: payload.modalProperties.closeCallback || function () {
-                            component.set('v.isLoading', false);
-                        },
+                        closeCallback: payload.modalProperties.closeCallback,
```
kenlewis-sfdo: @psadie-sfdo I removed the anonymous function that was attempting to set isLoading back to false as it was causing an exception every time the review donations modal was closed in debug mode and isLoading was always undefined at that point anyway.  Let me know if you see any issues with this change!

src/lwc/geFormField/geFormField.js
```diff
@@ -384,19 +405,70 @@ export default class GeFormField extends LightningElement {
 
         }
 
-        lookup.setSelected({value, displayValue});
-
+        if (this.sourceFieldAPIName === DONATION_RECORD_TYPE_NAME.fieldApiName) {
```
ptewson-sfdo, line 408: I know you have a similar comment up above, but maybe just include a similar comment here about handling the Donation Record Type Name as a special case

src/lwc/geFormField/geFormField.js
```diff
@@ -384,19 +405,70 @@ export default class GeFormField extends LightningElement {
...
     reset() {
         this.value = this._defaultValue;
 
         if (this.isLookup) {
             const lookup = this.template.querySelector('c-ge-form-field-lookup');
             lookup.reset();
+            if (this.fieldApiName === 'RecordTypeId') {
+                // Using setTimeout here ensures that this recordTypeId
+                // will be set on sibling fields after they are reset by queueing the event.
+                setTimeout(() => {
```
ptewson-sfdo, line 425: This is interesting ... I kind of reflexively want to avoid setTimeout() but this looks like an interesting technique to address the potential order of execution issue

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -684,8 +687,54 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+        const recordTypeNameValue = dataImport[DONATION_RECORD_TYPE_NAME.fieldApiName];
+        if (!recordTypeNameValue && recordTypeNameValue !== null) {
+            return false;
+        }
+
+        // If value is the RecordType Name, change it to the RecordType Id since child
+        // fields are expecting the Id and not the Name.
+        if (recordTypeNameValue &&
+            !recordTypeNameValue.value &&
+            !recordTypeNameValue.startsWith('012')) {
```
ptewson-sfdo, line 710: I'm not sure there's a better way to handle this, but if I get this right you are identifying cases where this field already holds a record type ID (vs Name) by whether it starts with the record type prefix of 012? it seems just a *little* touchy to me, and would fail in at least one extremely silly case (a record type name that starts with "012") ... but like I said not sure if I have a better way to do it. 

If it's not a big lift it might be nice to avoid hardcoding the prefix string here - defining as constant maybe or checking to see if there's an equivalent in LWC to the "findObjNameFromRecordIdPrefix" call from apex?

kenlewis-sfdo: Yep, good call!  I changed it to check if the value is an existing Opportunity RecordType Name instead of the id prefix check.  Let me know what you think of this!

src/lwc/geFormField/geFormField.js
kenlewis-sfdo: Yeah, it feels a bit strange but seems to be an acceptable practice to resolve race conditions.  In this case, it is the only way I could manage for picklists to get set back to the correct recordTypeId after the reset() flow independent of the order of the fields on the form.

src/lwc/geFormField/geFormField.js
kenlewis-sfdo: Good call - I added a comment to explain the special case for RecordTypeId fields, and also made a change to get the RecordType Name from the stored objectInfo property in this component which should improve performance by preventing the geFormFieldLookup component from having to query for the name.

melissabarber:
@kenlewis-sfdo & @ptewson-sfdo I updated this PRs description some to highlight the new picklist feature and include the other issue that was fixed. Still look complete and accurate?

src/aura/GE_GiftEntryForm/GE_GiftEntryFormController.js
psadie-sfdo: This change looks good to me! Thank you!

kenlewis-sfdo:
Thank you @melissabarber! Looks great.  I added two bullets in the top section to reflect additional functionality that is enabled by this PR - just wanted to track those in the PR also.

src/aura/GE_GiftEntryForm/GE_GiftEntryFormController.js
kenlewis-sfdo: Great thanks Prince!

melissabarber:
@kenlewis-sfdo Thanks! Is there anything that you added in the top section I should be sure to pull into the aggregated sections Changes or Issues Closed?

kenlewis-sfdo:
@melissabarber these are the two WI that are also closed: 
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001HtSJQA0/view
https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001HsBPQA0/view

And these are the two bullets I added to the top section.  Not sure if these warrant additional description in the "changes" section.  What do you think?

- enables the Review Donations modal link to appear on form load when Donation Donor and Account or Contact donor fields have default values

- enables related fields to be populated on form load when lookup fields have default values

PR #5293, opened 2020-03-31
Google International Address Verification Format Fix

### LINKED ISSUE #2499: Google geocoder Street Address not formatted properly for all non-US addresses
## Problem:
Not all countries format street addresses in the same way.  Here's a long list of [official postal service formats, by country](http://www.bitboost.com/ref/international-address-formats.html#Formats).

However, ADDR_GoogleGeoAPI_Validator sets MailingStreet__c to an American-style address (lines 196-215).

This makes international addresses wrong, and difficult to read, use, or even update (since the geocoder may fail if the order is wrong)

This is a similar problem in Address__c.Formula_MailingAddress__c.

## Examples:

> Piedras 623
> C1070AAM CABA
> Argentina

Piedras is the street_name, 623 is street_number, C1070AAM is the zip plus suffix, and CABA is the "state" short_name.

Formula_MailingAddress__c may even cause problems in Queens, NY:

> 35-01 Union St
> Flushing, NY 11354
> USA

Note that Flushing is the "neighborhood", Queens is the sublocality, and there is no locality (though other points in Queens have a locality of New York).

## Suggested Solution:

I would suggest taking whatever formatted_address Google gives you.  If you want to break out a separate street address in order to set BillingAddress better, you can perhaps delimit based on commas and take the first row, but it can be quite complicated to figure out if there's a MailingStreet2 (although it seems you don't set that anyway).  Honestly, you are handicapped by how Salesforce treats addresses anyway (c.f., [the Success idea to fix Addresses](https://success.salesforce.com/ideaView?id=08730000000BrVWAA0))

force2b:
# Critical Changes

# Changes

# Issues Closed

- Fixes #2499

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
**lurch: attach W-017823

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000059swgAAA" target="_blank">W-017823</a>

src/classes/ADDR_GoogleGeoAPI_Validator.cls
```diff
@@ -310,29 +130,368 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
...
+            if (!calloutToAddressService()) {
+                return this;
+            }
+
+            if (googleResponse.status != ADDR_Validator_UTIL.RESPONSE_OK || googleResponse.results == null) {
+                handleOtherResponses();
+                return this;
+            }
+
+            Boolean ignoreAmbiguous = settings.Reject_Ambiguous_Addresses__c == true;
```
RadSFDeveloper, line 217: It took me a while to understand this part of the code due to `Reject_Ambiguous_Addresses__c` (mis-aligned field API name and Help Text)... if it is set and the address is the partial match, it should mark the address as `Ambiguous__c` and 'Verified__c`. 

I tested few address verifications where the `Reject_Ambiguous_Addresses__c`  is unchecked. 
- One test was verification of an address with Zip Code, State and Country (US). The address verifier response contains `"partial_match" : true`. The address was marked as not `Ambiguous__c` and set as 'Verified__c` after the verification.
- The other test was an address containing format something like "100-1230 Water St, ..., Canada" then the address verification returns address as "100 Water St,...", ie the street number is removed while the apt number is taken as the street number. This does not seem correct, but I have not verified this against `master` to see if it is an existing issue.
- The the address format is something like  "1230 Water St, Apt 100, .., Canada." , the `"partial_match" : true` is returned, I thought it should be false.

If the `Reject_Ambiguous_Addresses__c` is checked, then addresses containing the second and third format above would be marked as ambiguous because the `partial_match` is true.  Is this expected?

src/classes/ADDR_GoogleGeoAPI_Validator.cls
```diff
@@ -35,264 +35,84 @@
...
-        else       
-            this.END_POINT = getDefaultURL();
+        String endPoint = settings.Address_Verification_Endpoint__c != null ?
+            settings.Address_Verification_Endpoint__c : getDefaultURL();
 
         if (!addresses.isEmpty()){
-            for (Address__c a : addresses){
-                checked_addresses.add(verifyAddress(a));
+            for (Address__c a : addresses) {
+                AddressVerifier addrVerify = new AddressVerifier(a)
```
RadSFDeveloper, line 87: `addrVerify` -> `verifier`

src/classes/ADDR_GoogleGeoAPI_Validator.cls
```diff
@@ -310,29 +130,368 @@ public with sharing class ADDR_GoogleGeoAPI_Validator implements ADDR_IValidator
...
+                address.Verified__c = false;
+                address.API_Response__c = 'MISSING AUTHID OR ENDPOINT.  Please see the Address Verification menu in the NPSP Settings tab for more information';
+                return this;
+            }
+
+            if (!calloutToAddressService()) {
+                return this;
+            }
+
+            if (googleResponse.status != ADDR_Validator_UTIL.RESPONSE_OK || googleResponse.results == null) {
```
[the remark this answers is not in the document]
force2b: I had left it in this method so that the calloutToAddressService method only handled the callout and casting the response. Any processing of a valid response (even "other" responses are not total failures) was handled by verifyAddress.

src/classes/ADDR_Validator.cls
```diff
@@ -55,72 +58,83 @@ public with sharing class ADDR_Validator {
...
+        if (classType == null) {
+            ERR_Handler.processError(new ValidationException(System.Label.Addr_Verification_Wrong_Class), ERR_Handler_API.Context.ADDR);
+            return;
         }
+
+        Object classInstance = classType.newInstance();
+        Boolean isInstanceOfAddress = (classInstance instanceof ADDR_IValidator);
+        if (!isInstanceOfAddress) {
+            // The objects we create dynamically need to implement the ADDR_IValidator interface
+            ERR_Handler.processError(new ValidationException(System.Label.Addr_Verification_Wrong_Interface), ERR_Handler_API.Context.ADDR);
```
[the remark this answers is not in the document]
force2b: Great catch!

src/classes/ADDR_GoogleGeoAPI_Validator.cls
force2b: I went through the original code and I cannot find where the new logic is any different than the previous logic. Do you still have an org built from master configured for address verification? If not, I'll build one to compare results just to make sure this is unchanged.

RadSFDeveloper: The code with regards to setting the Ambigous__c is fine. The issue was reading it and understanding the result. Also, the mentioned address format examples (Canadian address) return different API response in the branch org and the master org. I expected the `partial_match` to be false when the address is exact. However, the same address format in the `master` org returns "bad request" response so cannot compare.
In summary, the question boils down to different API response for the same address format when running the verification on the branch org compared to the `master` org. The WI contains screenshots and more detail.

PR #5297, opened 2020-04-02
NPSP Settings: Store API values when configuration is updated

### LINKED ISSUE #3021: Translated Picklist Labels for OCR Roles can affect Rollups
I discovered the following regarding the use of Picklist Labels and API Names in the code:
* The Settings page will display the label value (including the translated labels) in the settings UI.
* The Settings page also stores the label value into the Custom Settings field instead of the API Name. For example, if the current User has a language other than English and the "Donor" picklist value has been translated into something different (ex: "La Donor"), the value stored in the NPSP Custom Settings field will be "La Donor".
* The actual data in the OpportunityContactRole.Role field is always the picklist api name regardless of the value of the label and regardless of any translations. The beauty of this implementation is that picklist values are translation and override agnostic, allowing system integrations and custom apex to "always" work regardless of what an admin might have translated a picklist value label to.
* The NPSP code base does not support this implementation because it always uses the picklist label. In the example above where the "Soft Credit" OCR Role was translated to "La Soft Credit" and an Admin with a default language of "French" modified the NPSP Settings for this Role, the legacy Rollup jobs will no longer work because the code will be looking for OCR records with a Role="La Soft Credit". 

https://releasenotes.docs.salesforce.com/en-us/spring17/release-notes/rn_forcecom_picklists_ui_and_api_labels.htm

**Steps to reproduce:**
1. Setup | Translation Settings
2. Enable the French language
3. In Setup | Translate, select "French"; "Picklist Value"; "Opportunity" and expand the section for "Opportunity Contact Role"
4. On the "Soft Credit" role row, double-click in the second column to enable entry. Enter a value such as "La Soft Credit".
5. Click Save
6. In NPSP Settings (while the current user's language is English) verify the current Contact Role settings (Primary and Soft Credit Donor roles) and then execute the Donor Rollup Batch job.
7. Select a Contact that has Soft Credit rollups on it.
8. Make a note of the Soft Credit rollup values.
9. Change the current User's default language to "French"
10. Open the NPSP Settings Page. Select "Rôles des contacts" under "Dons" and click the [Modifier] button.
11. In the section for "PARAMÈTRES DE DONS DE GROUPE", change the selected Role to "La Soft Credit". 
12. Click Save
13. Change the User's language back to English (only to make reading the page easier)
14. Form the NPSP Settings Page, execute the Donor Rollup Batch job again
15. Return to the contact selected in step 7 and refresh the page. Note that the Soft Credit values are now gone.

In the above, the objective is to translate the OCR Role Label (which is the recommended approach for localization) and then for an Admin using a non-English Language to update the NPSP Settings Page (which is also likely). The result of that update is that the translated label value is saved as the selected Contact Roles in the NPSP Custom Settings objects rather than the Role's API Name. When the rollup jobs use that information to select OCR records for Soft Credits, it's trying to filer for Role="La Soft Credit" whereas the database record itself actually contains "Soft Credit" (the api name for that picklist value) in the Role field.

Note that this likely applies to other picklist values that are stored in Custom Settings such as Opportunity.Type. 

https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/STG_Panel.cls#L256
https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/RLLP_OppRollup_UTIL.cls#L844

andrewyu-salesforce:
As a non-English Admin User, I want NPSP to operate consistently for all Users in my Organization regardless of which language or locale any given user has configured for themselves.

# Critical Changes

# Changes

# Issues Closed

Fixes #3021

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/STG_Panel.cls
+++ b/src/classes/STG_Panel.cls
@@ -334,7 +334,7 @@ public with sharing virtual class STG_Panel {
                 Schema.DescribeFieldResult F = Schema.sObjectType.npe4__Relationship__c.fields.npe4__Type__c;
                 List<Schema.PicklistEntry> P = F.getPicklistValues();
                 for (Schema.PicklistEntry pe : P) {
-                    listSORelTypes.add(new SelectOption(pe.getLabel(), pe.getLabel()));
+                    listSORelTypes.add(new SelectOption(pe.getValue(), pe.getLabel()));
                 }
             }
             return listSORelTypes;
@@ -353,7 +353,7 @@ public with sharing virtual class STG_Panel {
                 Schema.DescribeFieldResult F = Schema.sObjectType.Campaign.fields.Type;
                 List<Schema.PicklistEntry> P = F.getPicklistValues();
                 for (Schema.PicklistEntry pe : P) {
-                    listSOCampaignTypes.add(new SelectOption(pe.getLabel(), pe.getLabel()));
+                    listSOCampaignTypes.add(new SelectOption(pe.getValue(), pe.getLabel()));
                 }
             }
             return listSOCampaignTypes;
--- a/src/classes/STG_Panel.cls
+++ b/src/classes/STG_Panel.cls
@@ -179,25 +179,6 @@ public with sharing virtual class STG_Panel {
     }
 
     /*******************************************************************************************************
-    * @description list of SelectOptions for Opportunity Contact Roles
-    */
-    static public list<SelectOption> listSOContactRoles {
-        get {
-            if (listSOContactRoles == null) {
-                listSOContactRoles = new list<SelectOption>();
-                listSOContactRoles.add(new SelectOption('', Label.stgLabelNone));
-                Schema.DescribeFieldResult F = Schema.sObjectType.OpportunityContactRole.fields.Role;
-                List<Schema.PicklistEntry> P = F.getPicklistValues();
-                for (Schema.PicklistEntry pe : P) {
-                    listSOContactRoles.add(new SelectOption(pe.getLabel(), pe.getLabel()));
-                }
-            }
-            return listSOContactRoles;
-        }
-        private set;
-    }
-
-    /*******************************************************************************************************
     * @description list of SelectOptions for Household Object Creation Rules
     */
     static public list<SelectOption> listSOHHRules {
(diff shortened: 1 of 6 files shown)


andrewyu-salesforce:
**lurch: attache W-039023

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HrLmQAK" target="_blank">W-039023</a>

andrewyu-salesforce:
@RadSFDeveloper The code is ready for review

src/classes/STG_PanelContactRoles_CTRL.cls
```diff
@@ -194,4 +194,94 @@ public with sharing class STG_PanelContactRoles_CTRL extends STG_Panel {
         }
         return results;
     }
+
+    /**
+    * @description The current setting being stored on the Opportunity Contact Role Default Role
+    */
+    public String individualOppsContactRole {
```
RadSFDeveloper, line 201: Perhaps var names can be changed a bit so it is clear if it is a field name, field label, field picklist value or the field picklist label. Since the `individualOppsContactRole` and other similar vars below are value *labels* of the custom settings field representing OpportunityContactRole Role picklist field value(s), can we somehow rename them? For example:
- `individualOppsContactRole` -> `selectedOptionDefaultContactRole`
- `organizationOppsContactRole` -> `selectedOptionOrgDefaultContactRole`
- `matchedDonorRole` -> `selectedOptionMatchedDonorRole` 
- `hhMemberContactRole` -> `selectedOptionHHMemberContactRole`
- `honoreeOppContactRole` -> `selectedOptionHonoreeContactRole`
- `recipientOppContactRole` -> `selectedOptionNotficationRecipientRole`?

src/classes/STG_Panel.cls
```diff
@@ -467,4 +448,29 @@ public with sharing virtual class STG_Panel {
...
     }
+
+    /*******************************************************************************************************
+    * @description Get the label of the picklist value
+    * @param objectName The Sobject name of the picklist field
+    * @param fieldName The field name of the picklist field
+    * @param apiValue The value of the picklist field
+    * @return String The label value of the picklist field
+    */
+    public static String getPicklistLabelFromValue(String objectName, String fieldName, String apiValue) {
```
RadSFDeveloper, line 459: There is logic in the NPSP Settings > Donations > Contact Roles that verifies assignment. For example, a HH Member Contact Role cannot be the same as the Matched Donor Role or Honoree Role or Notification Recipient Role. These rules would need to be retested for both translated org user and the English lang user.
@ravensisco , @andrewyu-salesforce

sam-knox:
@RadSFDeveloper - Is there a GitHub Issue this can be associated with?

force2b:
@sam-knox I had though there were no GH issues for this, however it looks like I created #3021 around 2 years ago. If you agree I think this WI fixes that issue

sam-knox:
@force2b - Yes, I'll add the Fixes statement so that this Issue can be closed. Thanks!

PR #5302, opened 2020-04-02
RD2: Schedule Service Handles DayOfMonth 29 and 30 - part 2

npsp-reedestockton:
We've added explicit support for monthly recurring donation days 29 and 30 (via Day of Month field). Previously, these days were all considered the last day of the month. Now we will respect the explicit selection in all months where that is possible and treat the selection as the last day of the month in all other cases. Last Day of Month is still supported as an option.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_ScheduleService.cls
+++ b/src/classes/RD2_ScheduleService.cls
@@ -619,26 +619,7 @@ public without sharing class RD2_ScheduleService {
             isCurrent =             rdSchedule.StartDate__c <= RD2_ScheduleService.currentDate ? true : false;
             campaignId =            rdSchedule.Campaign__c;
             campaignName =          rdSchedule.Campaign__r.Name;
-            if (rdSchedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY) {
-                if (rdSchedule.DayOfMonth__c == RD2_Constants.DAY_OF_MONTH_LAST_DAY) {
-
-                    Schema.DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe(
-                        UTIL_Namespace.getNamespace() + 'RecurringDonationSchedule__c',
-                        UTIL_Namespace.getNamespace() + 'DayOfMonth__c'
-                    );
-                    List<Schema.PicklistEntry> picklistEntries = dfr.getPicklistValues();
-
-                    for (Schema.PicklistEntry ple : picklistEntries) {
-                        if (ple.getValue() == RD2_Constants.DAY_OF_MONTH_LAST_DAY) {
-                            dayOfMonth = ple.getLabel();
-                            break;
-                        }
-                    }
-                }
-                else {
-                    dayOfMonth = rdSchedule.DayOfMonth__c;
-                }
-            }
+            dayOfMonth =            rdSchedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY ? rdSchedule.DayOfMonth__c : null;
             endDate =               rdSchedule.EndDate__c;
             installmentAmount =     rdSchedule.InstallmentAmount__c;
             installmentFrequency =  rdSchedule.InstallmentFrequency__c != null ? Integer.valueOf(rdSchedule.InstallmentFrequency__c) : null;
@@ -648,7 +629,6 @@ public without sharing class RD2_ScheduleService {
             scheduleId =            rdSchedule.Id;
             startDate =             rdSchedule.StartDate__c;
         }
-    }
 
     /***
     * @description Performs a full analysis and rebuild of the Schedule records for a Recurring Donation,
--- a/src/classes/RD2_ScheduleService.cls
+++ b/src/classes/RD2_ScheduleService.cls
@@ -602,33 +602,34 @@ public without sharing class RD2_ScheduleService {
     */
     public class ActiveSchedule {
 
-        public Boolean  isCurrent {get;set;}
-        public Id       campaignId {get; set;}
-        public String   campaignName {get; set;}
-        public String   dayOfMonth {get; set;}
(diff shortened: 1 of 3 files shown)


npsp-reedestockton:
**lurch: attach W-040427

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hu4BQAS" target="_blank">W-040427</a>

npsp-reedestockton:
Please see https://github.com/SalesforceFoundation/NPSP/pull/5291 for first part of this work item.

npsp-reedestockton:
@force2b @RadSFDeveloper Could you please take a look at this part 2 PR?  Thanks!

src/classes/RD2_ScheduleService.cls
```diff
@@ -619,7 +619,26 @@ public without sharing class RD2_ScheduleService {
             isCurrent =             rdSchedule.StartDate__c <= RD2_ScheduleService.currentDate ? true : false;
             campaignId =            rdSchedule.Campaign__c;
             campaignName =          rdSchedule.Campaign__r.Name;
-            dayOfMonth =            rdSchedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY ? rdSchedule.DayOfMonth__c : null;
+            if (rdSchedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY) {
```
RadSFDeveloper, line 622: I think Day of Month display conversion belongs in the visualizer and we can leave the schedule service as is.
Also, we can extract code into a method that will determine the Day of Month display value. So, if the method is added into the visualizer `Schedule`, it would be as below. Also, we need a test in the visualizer test class where Day of Month = Last_Day since current tests pass w/o this change.
```
public Schedule(Integer num, RD2_ScheduleService.ActiveSchedule schedule) {
            this.scheduleNumber = num;
            this.isCurrent = schedule.isCurrent;
            this.amount = schedule.installmentAmount;
            this.paymentMethod = schedule.paymentMethod;
            this.campaign = schedule.campaignName;
            this.startDate = schedule.startDate;
            this.endDate = schedule.endDate;
            this.period = schedule.installmentPeriod;
            this.frequency = schedule.installmentFrequency;
            this.dayOfMonth = determineDayOfMonth(schedule);
        }

        /**
         * @description Displays Day of Month picklist values 
         * @param schedule RD Schedule Service active schedule
         * @return String
         */
        private String determineDayOfMonth(RD2_ScheduleService.ActiveSchedule schedule) {
            if (schedule.installmentPeriod != RD2_Constants.INSTALLMENT_PERIOD_MONTHLY 
                || schedule.dayOfMonth == null
            ) {
                return null;
            }
            
            if (schedule.dayOfMonth == RD2_Constants.DAY_OF_MONTH_LAST_DAY) {                
                for (Schema.PicklistEntry picklistValue : RecurringDonationSchedule__c.DayOfMonth__c.getDescribe().getPicklistValues()) {
                    if (picklistValue.getValue() == RD2_Constants.DAY_OF_MONTH_LAST_DAY) {
                        return picklistValue.getLabel();
                    }
                }
            }

            return schedule.dayOfMonth;
        }

```

npsp-reedestockton: Not sure I agree with this.  I'm open to hearing your thoughts about the logic, but I don't understand why the schedule service should continue filling the data structure with Last_Day.  

That is a value that is never usable by a consumer of this data, so it seems incorrect to me to allow it to be populated with an unusable value.

RadSFDeveloper: The `RD2_ScheduleService.ActiveSchedule` is currently used by the visualizer, so there is no impact. If there was another class consuming the service method, then using the label instead of the API value would produce incorrect results. The reasoning was to keep the schedule service use the API value while the visualizer responsible for data display would use the label.

npsp-reedestockton: So the assumption is that the visualizer will remain the only consumer.

RadSFDeveloper: The assumption is the visualizer will not be the only consumer. Visualizer needs to output the label but not the schedule service.

force2b: Trying to catch up on this. If this method is what is used to generate the Schedule data to render in the LWC UI, then we **would** want DayOfMonth to use the Label vs. the ApiName. If there's concern that the consumer of this can be expanded in the future, then we can rename this property to `dayOfMonthLabel` so it's clear what the usage is. I suspect this is going to come out of [W-040523](https://foundation--agf.na70.visual.force.com/a2x1E000001HuEkQAK) anyway. My question is why is this changed here as part of this WI?

npsp-reedestockton: Okay, good news/bad news.  @force2b and I just discussed and we will defer the change to the Last Day data to be addressed in [W-040523](https://foundation--agf.na70.visual.force.com/a2x1E000001HuEkQAK).  That's the good news.  <joke>The bad new is I still think this belongs in the schedule service.</joke>

rponti-sforg:
QE testing is complete for this PR and associated WI.  Based on the WI Description, ACs, PR comments / feedback, and Test Charter feedback, tests have passed.

For further testing information, see:

https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001Hu4BQAS/view
https://salesforce.quip.com/1uzTAzszSkcF

After looking at the Feature Test - Robot failed test (72 out 73 passed), it appears that this was an 'environmental' / 'delay' issue (i.e. could not find tab).  Upstream Robot tests have passed.  Moving forward with merging PR.

PR #5319, opened 2020-04-08
Feature/robot  tempbldr

skristem:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/robot/Cumulus/resources/GiftEntryPageObject.py
+++ b/robot/Cumulus/resources/GiftEntryPageObject.py
@@ -53,10 +53,28 @@ class GiftEntryPage(BaseNPSPPage, BasePage):
         element=self.selenium.get_webelement(locator)
         status=element.get_attribute("aria-expanded")
         if status=="false":
-            time.sleep(2)    
+            self.selenium.wait_until_page_contains("Clone")    
         self.selenium.click_link(action)
-        if action=="Edit" or "Clone":
+        if action=="Edit" or action=="Clone":
             self.selenium.wait_until_page_contains("Gift Entry Template Information")
+
+    def select_object_group_field(self,object_group,field):
+        """Select the specified field under specified object group 
+           to add the field to gift entry form and verify field is added"""
+        locator=npsp_lex_locators["gift_entry"]["form_object_dropdown"].format(object_group)
+        self.selenium.scroll_element_into_view(locator)
+        self.selenium.click_element(locator)
+        element=self.selenium.get_webelement(locator)
+        status=element.get_attribute("aria-expanded")
+        if status=="false":
+            time.sleep(2)       
+        field_checkbox=npsp_lex_locators["gift_entry"]["object_field_checkbox"].format(field)  
+        self.selenium.scroll_element_into_view(field_checkbox)   
+        self.selenium.click_element(field_checkbox)
+        field_label=object_group+': '+field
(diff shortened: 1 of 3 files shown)


robot/Cumulus/resources/locators_48.py
```diff
@@ -250,6 +250,14 @@
     'modal-form':{
         'label':'//div[./*/*[text()="{}"]]',
         },
+    'gift_entry':{
+        'button':'//*[@data-qa-locator="button {}"]/button',
+        'field_input':'//*[contains(@data-qa-locator,"{}")]//child::input',
+        'textarea':'//*[contains(@data-qa-locator,"{}")]//child::textarea',
```
sgovindankutty, line 256: Can we pass the tags directly in the function/keyword where it getting called, since the only difference between the locator is the tag name?

robot/Cumulus/resources/GiftEntryPageObject.py
```diff
@@ -1,21 +1,119 @@
...
+
+    def select_object_group_field(self,object_group,field):
+        """Select the specified field under specified object group 
+           to add the field to gift entry form and verify field is added"""
+        locator=npsp_lex_locators["gift_entry"]["form_object_dropdown"].format(object_group)
+        self.selenium.scroll_element_into_view(locator)
+        self.selenium.click_element(locator)
+        element=self.selenium.get_webelement(locator)
+        status=element.get_attribute("aria-expanded")
+        if status=="false":
```
rjanjanam, line 73: Can you please explain this condition and sleep here

robot/Cumulus/tests/browser/gift_entry/template_builder.robot
```diff
@@ -0,0 +1,48 @@
...
+    #Create Template                             
+    Go To Page                                Custom                        GE_Gift_Entry
+    Click Gift Entry Button                   Create Template
+    Wait Until Page Contains                  Gift Entry Template Information
+    Enter Value In Field
+    ...                                       Template Name=${template}
+    ...                                       Description=This is created by automation script  
+    Click Gift Entry Button                   Next: Form Fields
+    Click Gift Entry Button                   Next: Batch Header
+    Click Gift Entry Button                   Save & Close
```
rjanjanam, line 29: Can these three combined into one keyword

skristem: No because in further test cases there will be multiple actions performed between these steps and this is the only test case where we are hitting Next, Next to reach to last page.

robot/Cumulus/resources/GiftEntryPageObject.py
skristem: when we click on the dropdown for object group, aria-expanded attribute will be set to true if dropdown expands, otherwise it will be false. so after clicking on dropdown, I am checking if its true or false. If true it will go ahead and perform next operation, incase if browser is slow in opening, it will be false so it will wait for 2 seconds for it to open and then try to perform next actions

robot/Cumulus/resources/locators_48.py
skristem: didn't think about that, thanks :)

skristem: made change

PR #5326, opened 2020-04-09
Prevent Open URL Redirects in BDI_DataImport_CRTL and BDI_DataImportDeleteBTN_CTRL

dfuller114:
We have a few BDI apex classes which have URL open redirect vulnerabilities that needed to be resolved. Instead of openly using the Salesforce `retURL` parameter for redirection, we are now using an explicit URL reference to where we want to redirect the user.

There are two places in the UI where this change was made.

1. When clicking the `Close` button in the BDI VF page for processing a batch. The user should be redirected back to the batch record (classic and Lightning).

2. Closing out of the mass delete data import records page and redirecting the user back to the data import records list view (classic and Lightning).

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -143,10 +152,20 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
      * @return PageReference 
      */
     public PageReference close() {
-        String retUrl = ApexPages.currentPage().getParameters().get('retURL');
+        String uiTheme = UserInfo.getUiThemeDisplayed();
```
ptewson-sfdo, line 155: If I read this right, you are basically ignoring the retURL parameter and instead assembling the return contextually? I wonder if there's a way to still use the return URL but swap out the base URL/domain to avoid redirects outside of the Salesforce instance. 

But - are we worried about hijacking within the instance? Technically I suppose someone could host a hostile VF page and redirect there. I'm just concerned that this might have gaps, especially if (god forbid) there's a new design system anytime soon.

src/classes/BDI_DataImportDeleteBTN_TEST.cls
```diff
@@ -41,6 +41,15 @@ class BDI_DataImportDeleteBTN_TEST {
     /** @description Error message returned when user does not have Delete permission on the SObject */
     static String INSUFFICIENT_ACCESS_ERROR_MSG = 'INSUFFICIENT_ACCESS_OR_READONLY';
 
+    /** @description Identifier for when Salesforce classic is displayed for the user */
```
ptewson-sfdo, line 44: Can you reference these defined constants from the original controller (even if you have to make those test visible?)

src/classes/BDI_DataImportDeleteBTN_TEST.cls
```diff
@@ -213,16 +222,13 @@ class BDI_DataImportDeleteBTN_TEST {
...
-     *    retURL page is returned
+     *    The correct URL is used for the Salesforce classic context
      */
     @isTest
     static void TestCloseWhenRetUrlIsProvided() {
-        String retURL = '/testpage.jsp';
-        ApexPages.currentPage().getParameters().put('retURL', retURL);
-
-        assertPage(retURL, new BDI_DataImportDeleteBTN_CTRL().close());
+        assertPage(UI_CLASSIC_THEME, new BDI_DataImportDeleteBTN_CTRL().close());
```
ptewson-sfdo, line 231: For the sake of completeness, might be nice to update the name of this test to specify "Classic" and provide a second test that tests lightning version (if we stick with this theme-specific implementation)

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -143,10 +152,20 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
...
+        String url;
+        Schema.DescribeSObjectResult dataImportDescribe = UTIL_Describe.getObjectDescribe(
+                UTIL_Namespace.StrTokenNSPrefix(DATA_IMPORT_API_NAME));
```
kenlewis-sfdo, line 158: Instead of passing the object name as a string to the `StrTokenNSPrefix` method, let's use the version of `getObjectDescribe` that takes the SObjectType as its argument be used here.  Then the `DATA_IMPORT_API_NAME` final string could be removed above.

`Schema.DescribeSObjectResult dataImportDescribe = UTIL_Describe.getObjectDescribe(
                DataImport__c.SObjectType);`

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -42,6 +42,15 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     /** @description Action type specified to delete all Data Import records */
     public static final String ACTION_DELETE_ALL = 'deleteAll';
 
+    /** @description Data Import SObject API name */
+    static final String DATA_IMPORT_API_NAME = 'DataImport__c';
```
kenlewis-sfdo, line 46: This can be removed by using the object token to get the object describe result.

src/classes/BDI_DataImportDeleteBTN_TEST.cls
```diff
@@ -41,6 +41,15 @@ class BDI_DataImportDeleteBTN_TEST {
...
     static String INSUFFICIENT_ACCESS_ERROR_MSG = 'INSUFFICIENT_ACCESS_OR_READONLY';
 
+    /** @description Identifier for when Salesforce classic is displayed for the user */
+    static final String UI_CLASSIC_THEME = 'Theme3';
+
+    /** @description Identifier for when Lightning Experience is displayed for the user */
+    static final String UI_LIGHTNING_THEME = 'Theme4d';
+
+    /** @description Data Import SObject API name */
+    static final String DATA_IMPORT_API_NAME = 'DataImport__c';
```
kenlewis-sfdo, line 51: Let's use the token instead of introducing this string.  See other comment above and let me know if you have any questions!

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
dfuller114: That's correct. One of the recommended ways of avoiding an open redirect like what the security team found is to explicitly specify where the redirect will take the user. So instead of using a more open `retURL`, not knowing what value that URL could contain, we would explicitly define it. This close function is only used in one place on the page and only ever redirects back to one page. I think what we're mainly concerned with regarding open redirect is sending the user outside of the internal instance to another URL entirely. 

Here is a [trailhead](https://trailhead.salesforce.com/en/content/learn/modules/secdev_application_logic_vulnerabilities/secdev_app_logic_preventing_open_redirect) that was linked on the WI that refers to the `hard-coded redirect` approach. I don't think doing the extra sanitation on `retURL` is necessary unless we were going to be dynamically redirecting users to different pages based on different actions.

src/classes/BDI_DataImportDeleteBTN_TEST.cls
dfuller114: fixed

src/classes/BDI_DataImportDeleteBTN_TEST.cls
dfuller114: added

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
dfuller114: ah good call. fixed that

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
dfuller114: fixed!

src/classes/BDI_DataImportDeleteBTN_TEST.cls
dfuller114: fixed!

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
ptewson-sfdo: This makes sense - thanks for looping me in to that ongoing thread!

dfuller114:
@Lucia-Juarez i made changes regarding the error that you found to make the redirect more generic in how it handles various origins.

src/classes/UTIL_String.cls
```diff
@@ -79,5 +79,29 @@ public class UTIL_String {
...
+
+            if (url.startsWith('/')) {
+                internalUrl = '/' + url.replaceFirst('/+', '');
+            } else {
+                internalUrl = new Url(url).getPath();
+            }
+
+            return internalUrl;
+
+        } catch(Exception pokemon) {
```
kenlewis-sfdo, line 102: I think we generally assign caught exceptions to a variable named "e" or "ex" - should this Exception be named similarly for consistency?

ptewson-sfdo: As much as it pains me, we should probably change this to something more generic :)

ptewson-sfdo: (But at the same time, yes - when it comes to exceptions, you DO have to have catch them all)

dfuller114: fixed! Ironically did that out of habit!

dfuller114:
**lurch attach W-039640

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hs2hQAC" target="_blank">W-039310</a>

dfuller114:
**lurch: attach W-039640

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hs2hQAC" target="_blank">W-039310</a>

dfuller114:
**lurch: detach W-039310

LurchTheButler:
Detached <a href="https://foundation.my.salesforce.com/a2x1E000001Hs2hQAC" target="_blank">W-039310</a>

dfuller114:
**lurch: attach W-039640

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsZzQAK" target="_blank">W-039640</a>

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -42,6 +42,18 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     /** @description Action type specified to delete all Data Import records */
     public static final String ACTION_DELETE_ALL = 'deleteAll';
 
+    @TestVisible
```
ptewson-sfdo, line 45: One last comment - it looks like with the shift in approach we no longer need these vars to exist, am I reading that right?

dfuller114: fixed!

PR #5327, opened 2020-04-09
Bug Bounty: BDI Permission Checks 

bdvorachek:
# Critical Changes

# Changes
- Improved permission checking around deleting Data Import rows
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

**lurch: attach W-039641

### DIFF
--- a/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
+++ b/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
@@ -78,7 +78,7 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     }
 
     public Boolean getCanDelete() {
-        return UTIL_Permissions.getInstance().canDeleteSObject(UTIL_Namespace.StrTokenNSPrefix('DataImport__c'));
+        return UTIL_Permissions.getInstance().canDelete(DataImport__c.SObjectType);
     }
 
     private void displayDeleteError() {
--- a/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
+++ b/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
@@ -78,7 +78,7 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     }
 
     public Boolean getCanDelete() {
-        return UTIL_Permissions.getInstance().canDelete(DataImport__c.SObjectType);
+        return UTIL_Permissions.getInstance().canDeleteSObject(UTIL_Namespace.StrTokenNSPrefix('DataImport__c'));
     }
 
     private void displayDeleteError() {
--- a/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
+++ b/src/classes/BDI_DataImportDeleteBTN_CTRL.cls
@@ -50,10 +50,9 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     @TestVisible
     public Boolean canDelete {
         get {
-            if (canDelete == null) {
-                canDelete = SObjectType.DataImport__c.isDeletable();
+            if(canDelete == null) {
+                canDelete = checkDelete();
             }
-
             return canDelete;
         }
         private set;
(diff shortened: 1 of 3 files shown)


LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hsa4QAC" target="_blank">W-039641</a>

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -95,6 +89,18 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
         return null;
     }
 
+    private Boolean checkDelete() {
+        return UTIL_Permissions.getInstance().canDelete(DataImport__c.SObjectType);
```
kenlewis-sfdo, line 93: Can this call to the `UTIL_Permissions` `canDelete` method be housed within the getter for the local canDelete property above (line 54)?  Then line 139 below could still be `if(canDelete)` which to me seems clearer than having a call to `checkDelete` inside the conditional there.

src/classes/UTIL_Permissions.cls
```diff
@@ -167,6 +167,17 @@ public with sharing class UTIL_Permissions {
...
 
+    /**
+    * @description Instance method which can be mocked
+    * Determines whether the running user has delete permissions on the given object
+    *
+    * @param sObjectType SObjectType to check
+    * @return returns boolean to indicate whether the user has permission
+    */
+    public Boolean canDelete(Schema.SObjectType sObjectType) {
+        return canDelete(sObjectType.getDescribe().getName(), false);
```
kenlewis-sfdo, line 178: It looks like the other methods in this class call `UTIL_Describe.getObjectDescribe` when describing objects (which caches describe results).  Not totally necessary here IMO but would make this new method more consistent with the others in the same class if you want to align them.

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -95,6 +77,18 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
         return null;
     }
 
+    public Boolean getCanDelete() {
```
kenlewis-sfdo, line 80: I liked the property name `canDelete` rather than `getCanDelete` which makes its usages more readable, but that is probably me just being finicky!

bdvorachek: We've got this getter bound in visualforce, so we need the `get` prefix here unless we go to the member-variable style getter/setters.

kenlewis-sfdo: 👍🏽

src/classes/BDI_DataImportDeleteBTN_CTRL.cls
```diff
@@ -78,7 +78,7 @@ public with sharing class BDI_DataImportDeleteBTN_CTRL {
     }
 
     public Boolean getCanDelete() {
-        return UTIL_Permissions.getInstance().canDelete(DataImport__c.SObjectType);
+        return UTIL_Permissions.getInstance().canDeleteSObject(UTIL_Namespace.StrTokenNSPrefix('DataImport__c'));
```
kenlewis-sfdo, line 81: Would be great to keep the param as the token instead of the string reference here if possible.

src/classes/UTIL_Permissions.cls
```diff
@@ -168,14 +168,14 @@ public with sharing class UTIL_Permissions {
...
+    * @description Determines whether the running user has delete permissions on the given object
+    * This is an instance method which can be mocked.
     *
-    * @param sObjectType SObjectType to check
+    * @param sObjectName Name of the sObject to check
     * @return returns boolean to indicate whether the user has permission
     */
-    public Boolean canDelete(Schema.SObjectType sObjectType) {
-        return canDelete(sObjectType.getDescribe().getName(), false);
+    public Boolean canDeleteSObject(String sObjectName) {
```
kenlewis-sfdo, line 177: Can the new instance method take the object token as arg instead of a string?  Sorry if I missed this in our conversation earlier.  I don't think we should introduce any new string representations of object/field representations unless we really have to.

PR #5333, opened 2020-04-13
RD2: Find matching Opps by CommitmentId

force2b:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

Custom Field:
- Opportunity.CommitmentId__c

Apex Class:
- RD2_ElevateIntegrationService
- RD2_ElevateIntegrationService_TEST

# Deleted Metadata


force2b:
**lurch: attach W-040591

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuLbQAK" target="_blank">W-040591</a>

src/classes/RD2_QueueableService.cls
```diff
@@ -74,4 +74,41 @@ public class RD2_QueueableService {
         }
     }
 
+    /**
```
RadSFDeveloper, line 77: Based on the previous comment to rename the class and methods, the inner class would be as specified below. The `RD2_ElevateIntegrationService` methods should not depend on each other, thus the `rdIdByComittmentId` is passed to matcher method.
```
    /**
     * @description Queueable service to process inserted Recurring Donations that have a value
     * in the Elevate CommitmentId__c field.
     */
    public class ElevateOpportunityMatcher implements Queueable {
        Map<String, Id> rdIdByComittmentId;

        /***
         * @description Constructor
         * @param rdIdByComittmentId - Recurring Donation Id mapped by the Commitment Id
         */
        public ElevateOpportunityMatcher(Map<String, Id> rdIdByComittmentId) {
            this.rdIdByComittmentId = rdIdByComittmentId;
        }

        /**
         * @description Execute method called when the Queueable runs. 
         * @param context QueuableContext
         */
        public void execute(QueueableContext context) {
            match();
        }

        /**
         * @description Delegate processing of the Recurring Donations to the Evaluation Service
         */
        public void match() {
            new RD2_ElevateIntegrationService()
                .matchRecurringDonationsToOpps(rdIdByComittmentId);
        }
    }
```

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -0,0 +1,117 @@
...
+    public RD2_ElevateIntegrationService withCommitmentIdMap(Map<String, Id> commitmentIdToRecordId) {
+        this.commitmentIdToRecordId = commitmentIdToRecordId;
+        return this;
+    }
+
+    /**
+    * @description For each CommitmentId, find any existing Opportunities with the matching Commitmentid
+    * and update the RecuringDonation lookup field (only if null).
+    */
+    public void matchRecurringDonationsToOpps() {
```
RadSFDeveloper, line 57: `public void matchRecurringDonationsToOpps(Map<String, Id> rdIdByComittmentId) {`

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -0,0 +1,117 @@
...
+        ]) {
+            if (opp.CommitmentId__c == null || opp.npe03__Recurring_Donation__c != null) {
+                continue;
+            }
+            opp.npe03__Recurring_Donation__c = commitmentIdToRecordId.get(opp.CommitmentId__c);
+            oppsToUpdate.add(opp);
+        }
+
+        if (!oppsToUpdate.isEmpty()) {
+            UTIL_DMLService.updateRecordBatchAndLogErrors(oppsToUpdate, 'RD_Elevate');
```
RadSFDeveloper, line 80: Should we add the context into `ERR_Handler_API.Context`? Is this value used in other teams' code?

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    /**
+    * @description For each CommitmentId, find the matching RecurringDonation with the same CommitmentId
+    * and update the RecurringDonation lookup field on the Opp.
+    */
+    public void matchOppsToRecurringDonations() {
```
RadSFDeveloper, line 88: This method is not used anywhere. Should it be removed?

src/classes/RD2_ElevateIntegrationService_TEST.cls
```diff
@@ -0,0 +1,141 @@
...
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        insert contact;
+    }
+
+
+    /***
+    * @description Verifies that any existing Opps are linked to a new RD if the CommitmentId matches
+    */
+    @isTest
+    private static void shouldLinkOppsToRecurringDonation() {
```
RadSFDeveloper, line 57: This test should be split into various tests where each test name specifies rule when Opp is matched with a Recurring Donation or when it is omitted from matching (one test per concept).

src/classes/RD2_ElevateIntegrationService.cls
force2b: This is the method I created for the other work item. Andrew is planning on using this as his starting point.

src/classes/RD2_ElevateIntegrationService.cls
force2b: I started that way, but it felt cleaner to use the `withCommitmentIdMap` method. I will change it though if you feel it would be better with the parameter on this method.

src/classes/RD2_ElevateIntegrationService.cls
force2b: I was uncertain on this one. The Error Context field/values are not well documented. If anything, they exist for categorization of errors to help in reporting (which assumes anyone actually reports on these). I believe there was some intent to configure error notifications by Context, though that was never fully implemented based on what I can see in the code.

For this I could have gone with the existing `RD` context, which isn't inaccurate.

Looking at the list of context enums, it really feels like the original intent was to tie these to the namespace prefix of the originating class (ADDR, BDI, CONV, HH, RD, etc.). So maybe the right answer here is to create a new enum for RD2 and just use that. I don't necessarily think that adding Elevate to the context will be significantly helpful here. Unless, we decide to create an enum for `Elevate` that truly could be used by other teams. What do you think?

src/classes/RD2_ElevateIntegrationService_TEST.cls
force2b: This actually is a single concept - verifying that the right Opp is updated _while_ the other Opps are not updated. I could split these out, but then it's not testing that only the one Opp is updated which is the goal here. I could also create another test method that re-tests that a single Opp is updated, but that seems redundant since this method is testing that concept.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -201,6 +204,42 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+    */
+    private void findMatchingElevateOpps(List<npe03__Recurring_Donation__c> rds, TDTM_Runnable.Action triggerAction) {
+
+        if (triggerAction != Action.AfterInsert) {
+            return;
+        }
+
+        Map<String, Id> commitmentIdToRecordId = new Map<String, Id>();
+        for (npe03__Recurring_Donation__c rd : rds) {
+            if (rd.CommitmentId__c == null || String.isBlank(rd.CommitmentId__c)) {
```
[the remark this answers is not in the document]
force2b: The RD.CommitmentId field is unique/case-sensitive so no more than a single RD can ever have the same value.

src/classes/RD2_QueueableService.cls
force2b: I like your class/method names.

src/classes/RD2_ElevateIntegrationService_TEST.cls
RadSFDeveloper: ok, I see, so this should be `shouldLinkOppsToRDMatchingCommitmentId() `?
Can we also verify AC3 that if the Commitment ID is matching existing Opp but the Opp is already assigned to an RD, the Opp will not be matched with the just created RD? For example: `shouldNotLinkAlreadyAssignedOppsToRDMatchingCommitmentId() `?

src/classes/RD2_ElevateIntegrationService.cls
RadSFDeveloper: Sounds good @force2b , let's use the new enum for `Elevate`.

force2b:
Thanks @sam-knox. I too would have preferred an api name of `ElevateRecurringId`, but it's too late to change what's on the RD and I think this new field should match.

sam-knox:
@force2b - Ok, thanks. CommitmentId it is then!

force2b:
Thanks @RadSFDeveloper. I always appreciate your detailed reviews and code suggestions, even when I sometimes push back on them. #CleanCodeAlways

rponti-sforg:
QE testing is now complete for this PR and corresponding WI.  Tests have passed.

https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001HuLbQAK/view
https://salesforce.quip.com/uCcMAYSud0Wj

Moving forward with merging PR.  Will monitor post merge builds and will set corresponding WI status to 'Closed' once builds have completed and are posted as 'Success'

PR #5343, opened 2020-04-15
[Robot]Enhancement to Regression tests to work with CRLP setting on

rjanjanam:
Currently, NPSP regression tests are failing when customizable rollups setting is on. This change deals with validating the batch job processes with and without the setting.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/NPSPSettingsPageObject.py
```diff
@@ -146,5 +146,17 @@ def verify_gift_entry_is_not_enabled(self):
                 self.click_toggle_button("Gift Entry")
                 self.wait_for_message("Gift Entry Disabled")
             else:
-                self.builtin.log("As expected Gift Entry is Disabled")    
-                                          
\ No newline at end of file
+                self.builtin.log("As expected Gift Entry is Disabled")
+
+    def Check_Crlp_Not_Enabled_By_Default(self):
+        """Verifies that advanced mapping is not enabled by default
```
skristem, line 152: you forgot to change this to customizable rollups

robot/Cumulus/resources/NPSP.robot
```diff
@@ -183,7 +183,28 @@ Run Donations Batch Process
...
+    Wait For Batch To Process    CRLP_RD_BATCH                            Completed
+    Wait For Batch To Process    CRLP_Account_AccSoftCredit_BATCH         Completed
+    Wait For Batch To Process    CRLP_Contact_SoftCredit_BATCH            Completed
+    Wait For Batch To Process    CRLP_Account_BATCH                       Completed
+    Wait For Batch To Process    CRLP_Contact_BATCH                       Completed
+
+Run Donations Batch Process
+    Go To Page                   Custom                         NPSP_Settings
+    Open Main Menu               Donations
+    Click Link With Text         Customizable Rollups
```
skristem, line 201: Just curious, did open NPSP settings keyword didn't work here

robot/Cumulus/resources/NPSP.robot
```diff
@@ -183,7 +183,28 @@ Run Donations Batch Process
...
+    ${crlp_enabled} =            Check Crlp Not Enabled By Default
+
+    #Open NPSP Settings and run Rollups Donations Batch job Validate the batch jobs completeness based accordingly
+    Run Keyword if      ${crlp_enabled} != True
+        ...             Validate Batch Process When CRLP Unchecked
+        ...     ELSE    Validate Batch Process When CRLP Checked
```
sgovindankutty, line 207: I think this be wrapped into a keyword and pass the boolean value. I see scope for reusability of this keyword to enable settings of similar kind.

robot/Cumulus/resources/NPSPSettingsPageObject.py
rjanjanam: Thanks. Updated

robot/Cumulus/resources/NPSP.robot
rjanjanam: Understand. But I feel that in this situation it would be overcomplicating it. Run Keyword IF itself explains the logic and is hidden in the NPSPY.Robot file. The actual test just calls this keyword. My concern is if I have to design a generic keyword
I will need a method to take in the parameters (condition), as conditions can vary may not be boolean all the time and the keywords to execute both if the condition satisfies and if the condition does not satisfy.

robot/Cumulus/resources/NPSP.robot
rjanjanam: I will try it out and update

PR #5345, opened 2020-04-15
Feature/single gift payments phase1  new int1 creds

ptewson-sfdo:
# Critical Changes

# Changes

Update the pre-packaged credentials we use to connect to elevate to point to new INT1 environment

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/GE_PaymentServices.cls
+++ b/src/classes/GE_PaymentServices.cls
@@ -142,7 +142,7 @@ public with sharing class GE_PaymentServices {
             generator.writeStringField('sub', 'trickyp'); // Subject
             generator.writeStringField('sfdo.id', 'bees'); // Org ID
             generator.writeStringField('sfdo.product', 'CRM'); // Product
-            generator.writeStringField('sfdo.roles', 'Payments.Purchase,Payments.Refund,Payments.Token'); // Call Role
+            generator.writeStringField('sfdo.roles', 'Payments.Purchase'); // Call Role
 
             return generator.getAsString();
         }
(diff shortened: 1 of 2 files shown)


src/classes/GE_PaymentServices.cls
```diff
@@ -132,11 +132,9 @@ public with sharing class GE_PaymentServices {
 
             generator.writeStartObject();
             generator.writeStringField('iss', 'ApiKeyManager'); // Issuer
-            generator.writeNumberField('iat', getEpochTimestamp(Datetime.now())); // Issued at
             generator.writeNumberField('exp', getExpirationDate(1)); // Expires at
-            generator.writeStringField('aud', 'www.example.com'); // Audience
-            generator.writeStringField('sub', 'cbaxter'); // Subject
-            generator.writeStringField('sfdo.id', '00D6g0000000000EAA'); // Org ID
+            generator.writeStringField('sub', UserInfo.getUserName()); // Subject
```
voduyemi, line 136: @ptewson-sfdo I think we can call the `getUsername()` method that we created instead of calling the `getUserName` method on the UserInfo class directly.

ptewson-sfdo: Hmm. We are in the inner class here, so I don't think it's super straightforward to call the method on the parent class?

psadie-sfdo: Yeah, unfortunately I think so as well. Might be cleaner to just call `UserInfo.getUsername()` in the two lines directly and clear out the outer `getUserName()` method.

voduyemi: Yeah I think you both are right, as long as we take out the outer one, if we wanted to call the outer one, we would need to instantiate the parent class.

ptewson-sfdo: Updated!

PR #5362, opened 2020-04-21
Gift Entry Payment Services: Read configuration from application storage

kenlewis-sfdo:
The global PS_IntegrationService.setConfiguration() method stores configuration sets used for callouts to Elevate Payment Services.  This set of changes reads those configuration values from application storage when the Elevate SDK is loaded and when the callouts are made.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/GE_PaymentServices.cls
+++ b/src/classes/GE_PaymentServices.cls
@@ -49,7 +49,7 @@ public with sharing class GE_PaymentServices {
     @TestVisible protected final String SFDO_ROLES = 'sfdo.roles';
     @TestVisible static final String SFDO_USERNAME = 'sfdo.username';
     @TestVisible static final String ELEVATE_SDK = 'elevateSDK';
-    @TestVisible static final String PRODUCTID = 'productId';
+    @TestVisible protected final String PRODUCTID = 'productId';
 
     // Names of Payment Service configuration sets
     @TestVisible static final String PAYMENTS_SERVICE_NAME = 'payments';
@@ -133,6 +133,10 @@ public with sharing class GE_PaymentServices {
         return config.get(ELEVATE_SDK);
     }
 
+    public String getProductId() {
+        return config.get(PRODUCTID);
+    }
+
     /*******************************************************************************************************
     * @description Inner class to assist in constructing the Payload portion of a JSON Web Token.
     */
@@ -159,7 +163,8 @@ public with sharing class GE_PaymentServices {
             generator.writeStringField('iss', 'ApiKeyManager'); // Issuer
             generator.writeNumberField('exp', getExpirationDate(1)); // Expires at
             generator.writeStringField('sub', UserInfo.getUserName()); // Subject
-            generator.writeStringField('sfdo.id', 'bees');
+            generator.writeStringField('sfdo.id',
+                    paymentServicesInstance.config.get(paymentServicesInstance.PRODUCTID));
             generator.writeStringField('sfdo.product', 'CRM'); // Product
             if (paymentServicesInstance.config.get(
                     paymentServicesInstance.SFDO_ROLES) != null) {
--- a/src/classes/GE_PaymentServices.cls
+++ b/src/classes/GE_PaymentServices.cls
@@ -59,9 +59,6 @@ public with sharing class GE_PaymentServices {
 
     public GE_PaymentServices() {
         config = new Configuration().keyValueMap;
-        if (config.isEmpty()) {
-            throw new PaymentServicesException(System.Label.gePaymentServicesConfigError);
-        }
     }
 
     public void setPurchaseCallBody(String jsonString) {
--- a/src/classes/GE_PaymentServices.cls
+++ b/src/classes/GE_PaymentServices.cls
@@ -46,7 +46,6 @@ public with sharing class GE_PaymentServices {
     @TestVisible static final String JWT_TOKEN = 'jwttoken';
     @TestVisible static final String SFDO_MERCHANTIDS = 'sfdo.merchantids';
     @TestVisible static final String SFDO_GATEWAYIDS = 'sfdo.gatewayids';
-    @TestVisible protected final String SFDO_ROLES = 'sfdo.roles';
     @TestVisible static final String SFDO_USERNAME = 'sfdo.username';
     @TestVisible static final String ELEVATE_SDK = 'elevateSDK';
     @TestVisible protected final String PRODUCTID = 'productId';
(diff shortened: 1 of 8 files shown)


kenlewis-sfdo:
**lurch: attach W-040742

src/labels/CustomLabels.labels
```diff
@@ -6237,6 +6237,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Failure message description for when the tokenization request times out</shortDescription>
         <value>The request timed out. Please try again or contact your Salesforce administrator.</value>
     </labels>
+    <labels>
+        <fullName>gePaymentServicesConfigError</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Payment Services configuration error</shortDescription>
+        <value>Elevate customer configuration not found. Please complete Elevate onboarding.</value>
```
kenlewis-sfdo, line 6246: @melissabarber This is the one new label introduced in this PR.

src/classes/GE_PaymentServices.cls
```diff
@@ -130,14 +159,21 @@ public with sharing class GE_PaymentServices {
             generator.writeStringField('iss', 'ApiKeyManager'); // Issuer
             generator.writeNumberField('exp', getExpirationDate(1)); // Expires at
             generator.writeStringField('sub', UserInfo.getUserName()); // Subject
-            generator.writeStringField('sfdo.id', 'bees'); // Org ID
+            generator.writeStringField('sfdo.id', 'bees');
```
ptewson-sfdo, line 162: This should be the elevate onboarding "productId" parameter I believe - we are currently receiving this from int1

src/classes/GE_PaymentServices.cls
```diff
@@ -130,14 +159,21 @@ public with sharing class GE_PaymentServices {
...
             generator.writeStringField('sfdo.product', 'CRM'); // Product
-            generator.writeStringField('sfdo.roles', 'Payments.Purchase'); // Call Role
```
ptewson-sfdo, line 135: We may want to continue wiring this to "Payments.Purchase" - or something to that effect. The roles we get passed to us from elevate are technically the full set of allowed operations we can make, but the only role we need to specify here in order to make the purchase call is "Payments.Purchase".

Technically we should probably block access to the credit card widget if your configured roles *don't* include "Payments.Purchase" ... not sure how we best bridge in the interim. Technically passing in *every* role we have available is more than we need however.

src/labels/CustomLabels.labels
ptewson-sfdo: I believe also this message is a bit of a fallback - we will soon deliver the story that hides the credit card field bundle when your config isn't available. Fairly sure.

src/pages/GE_TokenizeCard.page
```diff
@@ -10,9 +11,7 @@
         </form>
     </article>
 
-    <!-- TODO: Update src with production url and/or upload as static resource -->
-    <script src="https://dmjphp3j3je7i.cloudfront.net/elevate.min.js"></script>
-
+    <script src="{!elevateSDKURL}"></script>
     <script>
         const hostname = window.location.hostname;
         const arr = hostname.split(".");
```
ptewson-sfdo, line 17: We should also replace the "clientId: 'bees' down on line 27 with a dynamic reference to the "productId" we get passed through elevate config

src/labels/CustomLabels.labels
kenlewis-sfdo: Yeah that story is here in the backlog: https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001HtRpQAK/view
Do we want to not include this custom label and just not throw an error then for this WI?

src/classes/GE_PaymentServices.cls
kenlewis-sfdo: Aren't we supposed to be sending whatever values we get from Payment Services/Transport layer?  Do we know if the purchase call works if we include extra roles besides Payments.Purchase?  Maybe on their end, since they are sending multiple roles at times, they are just validating that the correct role is included in the passed in string...

src/classes/GE_PaymentServices.cls
kenlewis-sfdo: Ok, I'll update the code and the script to insert this value into test config.

src/pages/GE_TokenizeCard.page
kenlewis-sfdo: 👍🏽

src/labels/CustomLabels.labels
melissabarber: Sorry for chiming in late on this one. @kenlewis-sfdo if you were to throw an error that's not a custom label, what would that error look like? This is the copy I worked on for this error message, "You must have {0} enabled to use these fields." where {0} is the value for commonPaymentServices.

src/classes/GE_PaymentServices.cls
ptewson-sfdo: This might be worth a lengthier convo with Cody about how exactly this is intended to be used - one thing I'm beginning to understand is that the authorizer and the actual purchase endpoint are pretty strongly de-coupled, so when you present your role here, I *think* it's in a sense saying "let me through, please, I need to perform and action that requires the Payments.Purchase role" and then authorizer checks all your security stuff to see if you are in fact allowed through this particular gate. I think? We should probably ask for a more authoritative answer, but - I wouldn't hold this up based on this given the priority to have this functionality available!

src/labels/CustomLabels.labels
ptewson-sfdo: I honestly like having this in here, even if it means a touch of cruft in the form of that label. Especially while we wait for that story to get in ahead of this, I like getting the feedback

src/classes/GE_PaymentServices.cls
kenlewis-sfdo: Ok, thanks for the thoughtful response!  I just tried sending multiple role values as a string, like we set up in the npsp.PS_IntegrationService_TEST.buildConfigurationParams test method.  I got an error back from the service stating "Cannot read property 'map' of undefined" - so multiple values don't seem to work anyway.  I pushed a change that hard codes this callout back to using "Payments.Purchase" for the sfdo.roles value - and updated the script in the test charter.

src/classes/GE_PaymentServices.cls
kenlewis-sfdo: fixed!

src/pages/GE_TokenizeCard.page
kenlewis-sfdo: fixed!

src/labels/CustomLabels.labels
kenlewis-sfdo: I removed it.  I agree that once that WI is done, this label won't ever be encountered.  It is a nice fallback, but maybe not necessary.  Now, with this exception gone, the pmt widget just spins forever when it is on a form and there is no config (the sdk isn't loaded).  I think this is acceptable for the interim until we can just block it from being added to a template when not yet configured.  But let me know if you feel strongly and I'll put it back!

kenlewis-sfdo: @melissabarber no worries at all about the timing!  Thank you for your collaboration and attention on this.  I appreciate it and I appreciate you!

PR #5365, opened 2020-04-21
Gift Entry: Batches Tab "View More" link, Sorting

### LINKED ISSUE #5292: Gift Entry "Batches" Tab only shows 10 records, based on last added "Field to Display"
In the Gift Entry Object, when a user clicks on the Batches tab to see previous batches, they are shown 10 records only, and cannot see batches in that tab beyond that. When user chooses "Select Fields to Display" and saves, the List View is updated to show the last 10 records based on that field just added through the "Select Fields to Display" option, making an inconsistent and challenging user experience.

**Steps to Reproduce**
1. In Gift Entry, create more than 10 batches with at least 1 entry to allow them to save
2. Click the "Batches" tab and see 10 of the batches that you have created
3. Click the gear icon in the Batches window and select "Select Fields to Display"
4. Add a new field and save
5. See that the List View has been refreshed, and depending on the field, some batches may no longer be visible through this view

**Workaround**
To view all batches, go to NPSP Data Import Batches Object to see all batches and enter them.

bdvorachek:
# Critical Changes

# Changes
- We added a View More link to Gift Entry’s Batches tab. The link appears when Gift Entry has more than 10 batches. When you click View More, Gift Entry loads 10 more rows, and you can continue clicking the link until no further batches are available or the maximum of 2000 rows is reached.

# Issues Closed
- Closes #5292 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geListView/geListView.js
+++ b/src/lwc/geListView/geListView.js
@@ -289,7 +289,6 @@ export default class geListView extends LightningElement {
                 if (columnEntry) {
                     if (checkNestedProperty(this.columnEntriesByName[this.sortedBy],
                         'typeAttributes', 'label', 'fieldName')) {
-                        debugger;
                         orderedByFieldApiName = columnEntry.typeAttributes.label.fieldName;
                     } else {
                         orderedByFieldApiName = columnEntry.fieldApiName;
--- a/src/lwc/geListView/geListView.js
+++ b/src/lwc/geListView/geListView.js
@@ -2,7 +2,7 @@ import { LightningElement, api, track, wire } from 'lwc';
 import { getObjectInfo } from 'lightning/uiObjectInfoApi';
 import { getRecord } from 'lightning/uiRecordApi';
 import { dispatch, handleError, generateId, showToast } from 'c/utilTemplateBuilder';
-import { format, sort, deepClone, checkNestedProperty } from 'c/utilCommon';
+import { format, deepClone, checkNestedProperty } from 'c/utilCommon';
 import LibsMoment from 'c/libsMoment';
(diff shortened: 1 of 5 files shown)


src/lwc/geListView/geListView.js
```diff
@@ -149,15 +151,17 @@ export default class geListView extends LightningElement {
     }
 
     get isShowStandardFooter() {
-        const showFooter =
-            this.showStandardFooter === 'true' ||
+        return this.showStandardFooter === 'true' ||
             this.showStandardFooter === true ||
```
vmartinov-sforg, line 155: Where does this.showStandardFooter get set? Also, why do we need to check against 3 values?

bdvorachek: I think we're compensating for the fact that @api values come across as Strings if they're set directly in the HTML for the containing component, but could be Booleans if they're bound to a variable.

On line 59 and 93 of geTemplates.html, I set the `show-standard-footer` attribute but don't give it a value, which [ends up defaulting it to Boolean true](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/js_props_boolean). As long as we're consistent about doing Boolean properties like this, I agree with dropping the extra comparisons here.

vmartinov-sforg: Does this mean that the value will always be true?

bdvorachek: Yes. If we need to set it to `false`, we can omit the attribute entirely. And if we need to dynamically toggle it, we bind it to a variable and make sure that var is always a Boolean.

PR #5369, opened 2020-04-22
Store Elevate information on DI records

dfuller114:
# Critical Changes

# Changes
When a user charges a credit card, we will store relevant response information on the corresponding data import record so that admins can reference it later for troubleshooting and auditing purposes.

# Issues Closed
AA - W-040580

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_GiftEntryController.cls
```diff
@@ -121,7 +121,7 @@ public with sharing class GE_GiftEntryController {
     /*******************************************************************************************************
     * @description Saves the data import record, dry runs the record, and returns updated
     * totals with the new row.
-    * @param dataImport: the NPSP_Data_Import__c record to update
```
bdvorachek, line 124: Since these functions are @AuraEnabled we need to know the names of the parameters in order to call them from LWC. Can we standardize on either `diRecord` or `dataImport` as the parameter name we use when passing in a data import record from LWC? I lean towards dataImport.

dfuller114: @bdvorachek good point. My vote would be `dataImport` since that's more clear.

ptewson-sfdo: I like "dataImport"!

src/classes/GE_GiftEntryController.cls
```diff
@@ -121,7 +121,7 @@ public with sharing class GE_GiftEntryController {
...
+    * @param diRecord: the NPSP_Data_Import__c record to update
```
ptewson-sfdo, line 124: Can we update this comment to match "dataImport" in the method call?

src/classes/GE_GiftEntryController.cls
```diff
@@ -178,16 +175,14 @@ public with sharing class GE_GiftEntryController {
...
     * @description Run the DataImport process on a single gift
-    * @param @param diRecord DataImport record to be processed
+    * @param diRecord DataImport record to be processed
     * @param isDryRun Boolean to set if the process should only return import results and not process
     * @param updateGift Boolean to determine our matching criteria.
     *                   true = "single match or create" and means we are updating
     *                   false = "do not match"
-    * @return void
     */
     private static void processDataImport(DataImport__c diRecord, Boolean isDryRun, Boolean updateGift) {
```
ptewson-sfdo, line 184: @bdvorachek - would it be worth updating all these calls to match "dataImport" or should we only worry about the @AuraEnabled calls? Is that over-reach?

dfuller114: fixed

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -354,6 +354,7 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     @api
     handleCatchOnSave( error ) {
         // var inits
+        const errorMessage = error.body.message;
```
psadie-sfdo, line 357: Just checking to see if this errorMessage constant is used

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -190,34 +209,74 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+
                     const form = this.template.querySelector('c-ge-form-renderer');
                     form.showSpinner = false;
                     fireEvent(null, 'paymentError', {
                         error: {
                             message: splitErrorResponse,
                             isObject: true
                         }
-                     });
+                    });
```
psadie-sfdo, line 232: What do you think about pulling lines 219 - 232 into a separate method called `dispatchPaymentErrorEvent` or something along those lines?

src/lwc/geFormRenderer/geFormRenderer.js
ptewson-sfdo, line 357:
```suggestion
```

dfuller114:
**lurch attach: W-040580

PR #5375, opened 2020-04-23
Add PMM to NPSP Installers and Trial Metadata

davidmreed:
This PR adds PMM to NPSP's two primary installers (`install` and `full-install` slugs, the former being install.salesforce.org and the latter being the Trailhead installer).

It also adds PMM to the `config_trial` flow, along with minimal metadata support: a new Profile for PMM plus small changes to existing page layouts.

For draft installer links on Staging, please contact David Reed.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


unpackaged/config/trial_pmm/layouts/Contact-___NAMESPACE___Contact Layout.layout
```diff
@@ -0,0 +1,733 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b, line 1: I'm guessing this layout is intended to replace the layout that currently exists in the `unpackaged/config/trial/layouts` folder. We'll need to make sure to keep these in sync if we decide to make changes to the normal trial layout.

davidmreed: That's accurate. I chose to keep them separate, at least for now, on the thesis that we may be asked for the ability to deploy a PMM-less trial configuration. I'd like to reconverge them if that proves not to be the case.

PR #5395, opened 2020-04-29
Update BDI processing logic to exclude DI records with a pending or unsuccessful Payment Status

dfuller114:
# Critical Changes

# Changes
BDI will not process any data import records that have a failed payment status on the data import record.  

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -621,8 +621,42 @@ global with sharing class BDI_DataImportService {
     }
 
     /**
+     * @description Before processing data import records, this method filters out any data imports that do not meet
+     * custom criteria that applies to all data imports before allowing them to be processed.
+     * @param dataImports {List<DataImport__c>} List of data import records to be filtered
+     *
+     * @return List<DataImport__c>
+    */
+    private List<DataImport__c> shouldProcess(List<DataImport__c> dataImports) {
+        List<DataImport__c> dataImportsToProcess = new List<DataImport__c>();
+
+        for (DataImport__c dataImport : dataImports) {
+            /* Use a single boolean to determine if each data import should be processed. If new criteria is added
+            * it can be added to isSuccessfulPayment using conditional operators. */
+            Boolean shouldProcess = isSuccessfulPayment(dataImport);
+
+            if (shouldProcess) {
+                dataImportsToProcess.add(dataImport);
+            }
+        }
+
+        return dataImportsToProcess;
+    }
+
+    private Boolean isSuccessfulPayment(DataImport__c dataImport) {
+        String paymentStatus = dataimport.Payment_Status__c;
+
+        return String.isBlank(paymentStatus) ||
+            dataImport.Payment_Status__c.equalsIgnoreCase(
+                GE_PaymentServices.ElevateTransactionStatus.CAPTURED.name()
+            );
+    }
+
+    /**
      * @description Provides the access levels that should be validated depending on whether
      * dry run is enabled.
+     *
+     * @return Set<BDI_DataImportFLSService.AccessLevel>
      */
     private Set<BDI_DataImportFLSService.AccessLevel> getFieldAccessLevelsToValidate() {
         Set<BDI_DataImportFLSService.AccessLevel> fieldAccessLevelsToValidate =
@@ -639,28 +673,6 @@ global with sharing class BDI_DataImportService {
         return fieldAccessLevelsToValidate;
     }
 
-    /**
-     * @description Before processing data import records, this method filters out any data imports that do not meet
-     * custom criteria that applies to all data imports before allowing them to be processed.
-     * @param dataImports {List<DataImport__c>} List of data import records to be filtered
-     *
-     * @return List<DataImport__c>
-    */
-    private List<DataImport__c> shouldProcess(List<DataImport__c> dataImports) {
-        List<DataImport__c> dataImportsToProcess = new List<DataImport__c>();
-
-        for (DataImport__c dataImport : dataImports) {
-            String paymentStatus = dataImport.Payment_Status__c;
-
-            if(String.isBlank(paymentStatus) ||
-                dataImport.Payment_Status__c.toLowerCase() == PAYMENT_SERVICES_SUCCESSFUL_CHARGE) {
-
-                dataImportsToProcess.add(dataImport);
-            }
-        }
-
-        return dataImportsToProcess;
-    }
 
     /**
      * @description Stamps Data Import records with their FLS exception details.
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -41,9 +41,6 @@ global with sharing class BDI_DataImportService {
     public static final String FM_HELP_TEXT = 'Help Text';
     public static final String FM_DATA_IMPORT_FIELD_MAPPING = 'Data Import Field Mapping';
 
-    @TestVisible
-    private static final String PAYMENT_SERVICES_SUCCESSFUL_CHARGE = 'captured';
-
     public BDI_MappingService mappingService;
 
     /*******************************************************************************************************
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -40,6 +40,8 @@
 global with sharing class BDI_DataImportService {
     public static final String FM_HELP_TEXT = 'Help Text';
     public static final String FM_DATA_IMPORT_FIELD_MAPPING = 'Data Import Field Mapping';
+
+    @TestVisible
     private static final String PAYMENT_SERVICES_SUCCESSFUL_CHARGE = 'captured';
 
     public BDI_MappingService mappingService;
@@ -499,7 +501,12 @@ global with sharing class BDI_DataImportService {
     * @param dataImports a list of Data Import records to import
     */
     public void process(Id apexJobId, Data_Import_Settings__c dataImportSettings, List<DataImport__c> dataImports) {
-        List<DataImport__c> filteredDataImports = new List<DataImport__c>();
+        // All data imports not meeting custom criteria will be filtered out before processing.
+        List<DataImport__c> filteredDataImports = shouldProcess(dataImports);
+
+        if (filteredDataImports.isEmpty()) {
+            return;
+        }
 
         // initialize our performance logger
         perfLogger = new BDI_PerfLogger(filteredDataImports[0].NPSP_Data_Import_Batch__c, filteredDataImports.size());
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -643,8 +643,16 @@ global with sharing class BDI_DataImportService {
         return dataImportsToProcess;
     }
 
+    /**
+     * @description Checks a collection of data imports and returns a filtered list of only those data import records
+     * that have been successfully charged by payment services or those with a blank payment status which were not
+     * sent to payment services because a credit card charge was not required.
+     *
+     * @param dataImport {DataImport__c} List of data import records to be filtered
+     * @return Boolean
+     */
     private Boolean isSuccessfulPayment(DataImport__c dataImport) {
-        String paymentStatus = dataimport.Payment_Status__c;
+        String paymentStatus = dataImport.Payment_Status__c;
 
         return String.isBlank(paymentStatus) ||
             dataImport.Payment_Status__c.equalsIgnoreCase(
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -652,8 +652,8 @@ global with sharing class BDI_DataImportService {
         for (DataImport__c dataImport : dataImports) {
             String paymentStatus = dataImport.Payment_Status__c;
 
-            if(!String.isNotBlank(paymentStatus) &&
-                dataImport.Payment_Status__c.toLowerCase() != PAYMENT_SERVICES_SUCCESSFUL_CHARGE) {
+            if(String.isNotBlank(paymentStatus) &&
+                dataImport.Payment_Status__c.toLowerCase() == PAYMENT_SERVICES_SUCCESSFUL_CHARGE) {
 
                 dataImportsToProcess.add(dataImport);
             }
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -652,7 +652,7 @@ global with sharing class BDI_DataImportService {
         for (DataImport__c dataImport : dataImports) {
             String paymentStatus = dataImport.Payment_Status__c;
 
-            if(String.isNotBlank(paymentStatus) &&
+            if(String.isBlank(paymentStatus) ||
                 dataImport.Payment_Status__c.toLowerCase() == PAYMENT_SERVICES_SUCCESSFUL_CHARGE) {
 
                 dataImportsToProcess.add(dataImport);
--- a/src/classes/BDI_DataImportService.cls
+++ b/src/classes/BDI_DataImportService.cls
@@ -40,6 +40,7 @@
 global with sharing class BDI_DataImportService {
     public static final String FM_HELP_TEXT = 'Help Text';
     public static final String FM_DATA_IMPORT_FIELD_MAPPING = 'Data Import Field Mapping';
+    private static final String PAYMENT_SERVICES_SUCCESSFUL_CHARGE = 'captured';
 
     public BDI_MappingService mappingService;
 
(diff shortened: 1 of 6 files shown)


dfuller114:
tagged @force2b on this review also since it involves changes to BDI

dfuller114:
**lurch attach: W-040867

src/classes/BDI_DataImportService.cls
```diff
@@ -631,6 +639,29 @@ global with sharing class BDI_DataImportService {
...
+     * @return List<DataImport__c>
+    */
+    private List<DataImport__c> shouldProcess(List<DataImport__c> dataImports) {
+        List<DataImport__c> dataImportsToProcess = new List<DataImport__c>();
+
+        for (DataImport__c dataImport : dataImports) {
+            String paymentStatus = dataImport.Payment_Status__c;
+
+            if(!String.isNotBlank(paymentStatus) &&
+                dataImport.Payment_Status__c.toLowerCase() != PAYMENT_SERVICES_SUCCESSFUL_CHARGE) {
```
voduyemi, line 656: @dfuller114 Will every DI record now be required to have a payment status of 'captured' to be processed ?

dfuller114: good call. I'll update the if check there. forgot to remove the `!`. Should be checking for not blank and status 'CAPTURED` for the payment status. Meaning if the payment status is blank it will still process.

ptewson-sfdo: Good call, Victor - I think this should be "if(String.isBlank(paymentStatus) || dataImport.Payment_Status__c.equalsIgnoreCase(PAYMENT_SERVICES_SUCCESSFUL_CHARGE)) {"

dfuller114: @voduyemi fixed!

dfuller114: @ptewson-sfdo i just updated it to `String.isNotBlank(paymentStatus) && dataImport.Payment_Status__c.toLowerCase() == PAYMENT_SERVICES_SUCCESSFUL_CHARGE`.

voduyemi: @dfuller114 Haha, your brain's working :) but I don't think both statements are not the same, we expect that DI records with blank payment statuses to get processed, the statement you wrote will filter out those records.

force2b:
Thanks @dfuller114. I don't know enough about BDI to know if this is accurate or not. I defer to @ptewson-sfdo's knowledge on that functionality.

src/classes/BDI_DataImportService.cls
ptewson-sfdo: I agree that the "&&" really needs to be an "||" here. All records with a blank payment status field need to be processed. Unless I'm missing something, your logic as its written will only process them if the value in the payment status field is "captured".

I also prefer the use of "equalsIgnoreCase()" here over the "toLowerCase() ==" as it's more readable, and handles case incongruity in both arguments (I'm likely to also suggest updating the casing on the defined constant)

src/classes/BDI_DataImportService.cls
```diff
@@ -41,6 +41,9 @@ global with sharing class BDI_DataImportService {
     public static final String FM_HELP_TEXT = 'Help Text';
     public static final String FM_DATA_IMPORT_FIELD_MAPPING = 'Data Import Field Mapping';
 
+    @TestVisible
+    private static final String PAYMENT_SERVICES_SUCCESSFUL_CHARGE = 'captured';
```
ptewson-sfdo, line 45: I think we should probably define the entire enumerated vocab for the payment status somewhere - and I think that probably belongs somewhere else (maybe GE_PaymentServices?). I think something more similar to this: 
`    /***
    * @description AsyncApexJob possible status values
    */
    public enum ApexJobStatus {
        HOLDING, QUEUED, PREPARING, PROCESSING, COMPLETED, FAILED, ABORTED
    }

    /***
    * @description AsyncApexJob possible status values indicating the batch job is in progress
    */
    public static final Set<String> IN_PROGRESS_STATUSES = new Set<String>{
        ApexJobStatus.HOLDING.name(),
        ApexJobStatus.QUEUED.name(),
        ApexJobStatus.PREPARING.name(),
        ApexJobStatus.PROCESSING.name()
    };`

ptewson-sfdo: (I don't know if we need to build that set of strings, but might have to use the .name() to compare the string value of the field with the enumeration we define)

src/classes/BDI_DataImportService.cls
```diff
@@ -631,6 +639,29 @@ global with sharing class BDI_DataImportService {
...
     }
 
+    /**
+     * @description Before processing data import records, this method filters out any data imports that do not meet
+     * custom criteria that applies to all data imports before allowing them to be processed.
+     * @param dataImports {List<DataImport__c>} List of data import records to be filtered
+     *
+     * @return List<DataImport__c>
+    */
+    private List<DataImport__c> shouldProcess(List<DataImport__c> dataImports) {
```
ptewson-sfdo, line 649: This approach - filtering the list - definitely works, but I'm wondering if we should have this method check one data import at a time, and then return a boolean based on whether it should be processed or not. I know I suggested this name, but we should either update this to include more of criteria about when we should process a record, or name it more specific to payment status. Then this method could be called at the very beginning of the processing loop - line #917 in my file, but right where the comment "// skip di's that already have an error" is currently. The main advantage to handling it that way is that it avoids additional iterations over the list of data imports, and we are very focused on performance at scale here. 

IF we end up keeping this implementation - and I think I'm leaning more towards the one-record-check version - we should rename this to something like "filterPendingPayments" or something more descriptive

but I'm wondering if we should either rename this method to "

src/classes/BDI_DataImportService.cls
ptewson-sfdo: Think of it this way - these values need to be processed:
 Payment Status: "" 
 Payment Status: "CAPTURED"

And this record should NOT be processed:
 Payment Status: "PENDING" 

For the blank status, String.isNotBlank(paymentStatus) returns "false" and `dataImport.Payment_Status__c.toLowerCase() == PAYMENT_SERVICES_SUCCESSFUL_CHARGE` also returns "false". False && False returns false and so the record won't be added to the list that's returned from the method, and won't be processed as a result.

src/classes/BDI_DataImportService.cls
vmartinov-sforg: We have the enum in javascript in geGiftEntryFormApp.js as 
const PAYMENT_TRANSACTION_STATUS_ENUM = Object.freeze({
    PENDING: 'PENDING',
    AUTHORIZED: 'AUTHORIZED',
    CANCELED: 'CANCELED',
    CAPTURED: 'CAPTURED',
    DECLINED: 'DECLINED',
    NONRETRYABLEERROR: 'NONRETRYABLEERROR',
    RETRYABLEERROR: 'RETRYABLEERROR',
    REFUNDISSUED: 'REFUNDISSUED'
});
It would be great to store them in only one place (apex) and import them in js.

src/classes/BDI_DataImportService.cls
dfuller114: @ptewson-sfdo agreed. think i confused myself because originally I was going to not add to the list if the record shouldn't be processed so I changed that part without changing the if logic.

dfuller114: just updated!

src/classes/BDI_DataImportService.cls
dfuller114: @vmartinov-sforg @ptewson-sfdo yeah we can do that. I also think `GE_PaymentServices` would be a good candidate for that since we have defined that class as containing all PS related methods and functionality.

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -113,6 +113,62 @@ private with sharing class BDI_DataImportService_TEST {
...
+    /*******************************************************************************************************************
+    * @description If any custom criteria is set that all data imports must satisfy, if that criteria is not met then
+     * BDI should not process those data import records.
+    */
+    @isTest
+    private static void shouldBypassDataImportsWithFailedCriteria() {
+        Data_Import_Settings__c diSettings = getDataImportSettings();
+
+        BDI_DataImportService service = new BDI_DataImportService(false,
+                BDI_MappingServiceHelpText.getInstance());
```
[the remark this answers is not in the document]
dfuller114: yeah good call. updated!

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -113,6 +113,62 @@ private with sharing class BDI_DataImportService_TEST {
...
+
+    /*******************************************************************************************************************
+    * @description Tests that BDI correctly processes all data import records that meet custom criteria.
+    */
+    @isTest
+    private static void shouldProcessDataImportsWithAllCriteriaMet() {
+        Data_Import_Settings__c diSettings = getDataImportSettings();
+
+        BDI_DataImportService service = new BDI_DataImportService(false,
+                BDI_MappingServiceHelpText.getInstance());
```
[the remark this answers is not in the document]
dfuller114: yeah good call. updated!

src/classes/BDI_DataImportService.cls
ptewson-sfdo: UPDATE: After further conversation, I'm back on board with the iteration approach - iterating through and filtering the DI list based on payment status.

src/classes/BDI_DataImportService.cls
```diff
@@ -612,9 +620,43 @@ global with sharing class BDI_DataImportService {
...
+
+            if (shouldProcess) {
+                dataImportsToProcess.add(dataImport);
+            }
+        }
+
+        return dataImportsToProcess;
+    }
+
+    private Boolean isSuccessfulPayment(DataImport__c dataImport) {
```
ptewson-sfdo, line 646: Even though this method is private, I think a comment header to explain the behavior like we discussed would be nice

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -113,6 +113,62 @@ private with sharing class BDI_DataImportService_TEST {
...
+
+        BDI_DataImportService service = new BDI_DataImportService(false,
+                BDI_MappingServiceAdvanced.getInstance());
+        service.injectDataImportSettings(diSettings);
+
+        List<DataImport__c> testDataImports = UTIL_UnitTestData_TEST.createDIRecordsInANewGEBatch(50);
+        insert testDataImports;
+
+        for (DataImport__c dataImport : testDataImports) {
+            dataImport.Payment_Status__c = BDI_DataImportService.PAYMENT_SERVICES_SUCCESSFUL_CHARGE;
```
ptewson-sfdo, line 159: Update this to reference the enumeration value from GE_PaymentServices

src/classes/BDI_DataImportService_TEST.cls
```diff
@@ -113,6 +113,62 @@ private with sharing class BDI_DataImportService_TEST {
...
+        }
+
+        Test.startTest();
+            service.process(null, diSettings, testDataImports);
+        Test.stopTest();
+
+        for (DataImport__c dataImport : [SELECT Status__c FROM DataImport__c]) {
+            System.assert(String.isNotBlank(dataImport.Status__c), 'The data import should have been' +
+                'processed by BDI.');
+            System.assertEquals('Imported', dataImport.Status__c, 'The data import was not successfully processed.');
```
ptewson-sfdo, line 169: Instead of 'Imported', can you use BDI_DataImport_API.bdiImported?

src/classes/BDI_DataImportService.cls
dfuller114: updated!

src/classes/BDI_DataImportService_TEST.cls
dfuller114: updated!

src/classes/BDI_DataImportService_TEST.cls
dfuller114: updated!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -82,6 +74,13 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
             this.sObjectName === DATA_IMPORT_BATCH_OBJECT.objectApiName;
     }
 
+    connectedCallback() {
+        getPaymentTransactionStatus()
```
ptewson-sfdo, line 78: Can we rename this to "getPaymentTransactionStatusEnum" or "getPaymentTransactionStatusValues"? This makes it seem like you are getting the value of of the transaction status, which is misleading

src/classes/GE_PaymentServices.cls
```diff
@@ -83,6 +94,16 @@ public with sharing class GE_PaymentServices {
         config = new Configuration().keyValueMap;
     }
 
+    @AuraEnabled(cacheable=true)
+    public static String getPaymentTransactionStatus() {
```
ptewson-sfdo, line 98: See other comment about rename to "getPaymentTransactionStatusEnum" or "getPaymentTransactionStatusValues"

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -51,16 +52,7 @@ const PAYMENT_AUTHORIZED_AT = DI_PAYMENT_AUTHORIZED_AT.fieldApiName;
...
-    PENDING: 'PENDING',
-    AUTHORIZED: 'AUTHORIZED',
-    CANCELED: 'CANCELED',
-    CAPTURED: 'CAPTURED',
-    DECLINED: 'DECLINED',
-    NONRETRYABLEERROR: 'NONRETRYABLEERROR',
-    RETRYABLEERROR: 'RETRYABLEERROR',
-    REFUNDISSUED: 'REFUNDISSUED'
-});
+const PAYMENT_TRANSACTION_STATUS_ENUM = {};
```
ptewson-sfdo, line 55: I'm not actually seeing an error in my console log, but shouldn't it be a TypeError to declare this as "const" and then attempt to assign it on line 80?

src/classes/GE_PaymentServices.cls
```diff
@@ -83,6 +94,16 @@ public with sharing class GE_PaymentServices {
...
+        Map<String, String> paymentStatusNameByEnum = new Map<String, String>();
+
+        for (ElevateTransactionStatus status : ElevateTransactionStatus.values()) {
+            paymentStatusNameByEnum.put(status.name(), status.name());
+        }
+        return JSON.serialize(paymentStatusNameByEnum);
+    }
```
kenlewis-sfdo, line 105: Is there a use-case for building and returning a key-value pair map here when the keys and values are always the same value?  Could this method just return the list of values?

`return ElevateTransactionStatus.values();`

dfuller114: @kenlewis-sfdo This is mainly to preserve the object structure that an `enum` would have if created in Javascript so that we can reference a property as a key and get the string value corresponding to that property.

dfuller114:
**lurch: attach W-040867

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuxCQAS" target="_blank">W-040867</a>

src/classes/GE_PaymentServices.cls
kenlewis-sfdo: Ok.  Does the code need that structure on the LWC side?  Or does it just validate that a status received from the service is one of the expected valid values, in which case this list could just be a set instead of a map.  

Also, I don't see this new `PAYMENT_TRANSACTION_STATUS_ENUM` being used anywhere currently - is this a placeholder?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -338,16 +337,17 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
         switch (paymentStatus) {
-            case PAYMENT_TRANSACTION_STATUS_ENUM.PENDING: return false;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.AUTHORIZED: return false;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.CANCELED: return false;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.CAPTURED: return false;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.DECLINED: return true;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.NONRETRYABLEERROR: return false;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.RETRYABLEERROR: return true;
-            case PAYMENT_TRANSACTION_STATUS_ENUM.REFUNDISSUED: return false;
+            case this.PAYMENT_TRANSACTION_STATUS_ENUM.PENDING: return false;
```
bdvorachek, line 343: I think `PENDING` should be `true` here. The next step happening here is attempting the purchase call, rather than doing BDI processing. If the purchase response is successful, then the status gets updated to be `CAPTURED` or a different status if it fails.

src/classes/GE_PaymentServices.cls
dfuller114: @kenlewis-sfdo yeah I removed that unused constant. We actually need to explicitly reference enum values in the object instead of performing a `contains` operation so that's why we need to return an object vs an array.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
dfuller114: @bdvorachek agreed. I'll make that change.

PR #5397, opened 2020-04-29
GE PMT: Clear out Payment Status and Token in 'Do Not Charge Card' Mode

voduyemi:
# Critical Changes

# Changes
- The payment status and payment authorization token will now be cleared out from Data Import records if the 'Do Not Charge Card' button is pressed while a single gift is entered.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
+++ b/src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
@@ -49,12 +49,9 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
     handleUserDisabledWidget() {
         this.toggleWidget(true);
         this.hasUserDisabledWidget = true;
-        fireEvent(
-            null,
-            'doNotChargeState',
-            {
-                isWidgetDisabled : this.hasUserDisabledWidget
-            });
+        this.dispatchApplicationEvent('doNotChargeState', {
+            isWidgetDisabled : this.hasUserDisabledWidget
+        });
     }
 
     /*******************************************************************************
@@ -64,12 +61,9 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
         this.isLoading = true;
         this.toggleWidget(false);
         this.hasUserDisabledWidget = false;
-        fireEvent(
-            null,
-            'doNotChargeState',
-            {
-                isWidgetDisabled : this.hasUserDisabledWidget
-            });
+        this.dispatchApplicationEvent('doNotChargeState', {
+            isWidgetDisabled : this.hasUserDisabledWidget
+        });
     }
 
     /*******************************************************************************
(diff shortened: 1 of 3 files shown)


voduyemi:
**lurch attach: W-040868

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuxHQAS" target="_blank">W-040868</a>

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -193,10 +194,12 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
     prepareInMemoryDataImportForUpdate(inMemoryDataImport) {
         inMemoryDataImport.Id = this.dataImportRecord.Id;
         inMemoryDataImport[PAYMENT_METHOD__C] = this.dataImportRecord[PAYMENT_METHOD__C];
-        inMemoryDataImport[PAYMENT_STATUS__C] = this.dataImportRecord[PAYMENT_STATUS__C];
+        inMemoryDataImport[PAYMENT_STATUS__C] =
+            this._isCreditCardWidgetInDoNotChargeState ? '' : this.dataImportRecord[PAYMENT_STATUS__C];
         inMemoryDataImport[PAYMENT_DECLINED_REASON__C] =
```
vmartinov-sforg, line 199: We need to clear this field as well ( I know the WI does not mention it). I forced a payment to fail by entering a negative amount, then I clicked 'Do not Charge card' and then I saved. The Declined reason ["amount: Must be greater than 0"] stayed on the DI after processing.

voduyemi: I thought of the same thing too, I believe it was brought up and we were fine with leaving the decline reason. But I'll take it out.

voduyemi: @vmartinov-sforg I just updated it.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -49,7 +49,12 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
     handleUserDisabledWidget() {
         this.toggleWidget(true);
         this.hasUserDisabledWidget = true;
-        // TODO: dispatch an event to form renderer to clear the token from the data import record
+        fireEvent(
```
psadie-sfdo, line 52: Would it be worth pulling this into its own function and call that dispatcher function here and in line 67 below?

voduyemi: Updated! @psadie-sfdo

PR #5411, opened 2020-05-04
RD2: Status to state mapping page and config

RadSFDeveloper:
# Critical Changes
- The RD2_StatusMappingSettings_CTRL Apex class is required to be enabled for user profiles that manage NPSP Settings.

# Changes
- Enable admins to configure enhanced Recurring Donation Status value mapping to 'Active', 'Lapsed' or 'Closed' State so Recurring Donation records can be processed accordingly.

# Issues Closed

# Community Ideas Delivered

# New Metadata
Apex Class:
- RD2_StatusMappingSettings_CTRL
- RD2_StatusMappingSettings_TEST

Apex Page:
- STG_PanelRD2StatusMapping

Custom Labels:
- RD2_StatusMappingColumnState
- RD2_StatusMappingColumnStatusAPIName
- RD2_StatusMappingDefinitions
- RD2_StatusMappingDeploymentAlreadyInProgress
- RD2_StatusMappingInProgressMessage
- RD2_StatusMappingIntro
- RD2_StatusMappingStateUnmapped
- RD2_StatusMappingSuccessMessage
- stgNavRD2StatusMapping

Custom Object:
- RecurringDonationStatusMapping__mdt

Custom Field:
- npe03__Recurring_Donations_Settings__c.StatusMappingDeploymentId__c
- RecurringDonationStatusMapping__mdt.State__c
- RecurringDonationStatusMapping__mdt.Status__c

Page Layout:
- RecurringDonationStatusMapping__mdt-Recurring Donation Status Mapping Layout

Lightning Web Component:
- customDatatable
- customPicklistType
- rd2StatusMappingSettings

New StaticResourceSources file:
- customDatatable/style.css

# Deleted Metadata


RadSFDeveloper:
**lurch: attach W-033837

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087vj9QAA" target="_blank">W-033837</a>

src/classes/RD2_StatusMapper_TEST.cls
```diff
@@ -30,36 +31,290 @@
...
+    }
+
+    /**
+     * @description Builds custom metadatype record
+     * @param status Status value
+     * @param state State mapping value
+     * @return RecurringDonationStatusMapping__mdt
+     */
+    private static RecurringDonationStatusMapping__mdt buildMockCustomMetadataRecord(String status, String state) {
+        return new RecurringDonationStatusMapping__mdt(
```
force2b, line 302: It's so nice that these can be mocked with a simple object instance. I wish the same were true with CRLP CMDT's.

src/classes/RD2_StatusMappingSettings_CTRL.cls
```diff
@@ -0,0 +1,429 @@
...
+        String getDeployErrorMessage();
+        void save(List<Metadata.CustomMetadata> records);
+        void saveDeploymentId();
+    }
+
+    /**
+     * @description Contains state value to label mapping
+     */
+    private static final Map<String, String> stateLabelByValue = new Map<String, String>{
+        RD2_Constants.STATUS_ACTIVE => System.Label.RD2_StatusMappingStateActive,
```
force2b, line 59: I would prefer that the code uses the actual translated label of the picklist value (these are already packaged). Adding a custom label means that there are two different translations for the same thing, used in two different places. See the `UTIL_Describe.getTranslatedPicklistLabel()` method.

src/labels/CustomLabels.labels
```diff
@@ -2103,6 +2103,105 @@
...
+        <shortDescription>Mapping datatable Status field API name</shortDescription>
+        <value>API Name</value>
+    </labels>
+    <labels>
+        <fullName>RD2_StatusMappingColumnStatusLabel</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Mapping datatable Status field label</shortDescription>
+        <value>Status</value>
```
force2b, line 2128: Rather than use a new custom label for this, do you think it would be better to use the label of the `npe03__Recurring_Donation__c.Status__c` field?

src/lwc/rd2StatusMappingSettings/rd2StatusMappingSettings.html
```diff
@@ -0,0 +1,95 @@
+<template>
+
+    <template if:true={hasMessage}>
+        <div class="slds-p-left_large slds-p-right_large" tabindex="-1">
+            <div class="desktop forcePageError" aria-live="assertive">
```
force2b, line 5: What does the `desktop` and `forcepageError` classes do here?

src/objects/RecurringDonationStatusMapping__mdt.object
```diff
@@ -0,0 +1,50 @@
...
+                    <label>Closed</label>
+                </value>
+            </valueSetDefinition>
+        </valueSet>
+    </fields>
+    <fields>
+        <fullName>Status__c</fullName>
+        <description>API name of the Recurring Donation Status that will be mapped to the state.</description>
+        <externalId>false</externalId>
+        <fieldManageability>SubscriberControlled</fieldManageability>
```
force2b, line 39: I'm not so sure about this. I believe that this should be "DeveloperControlled" to match all other CMDT fields in our package. Even as DeveloperControlled field, Users are still able to modify the values in the field and create records. If I'm not mistaken, making this SubscriberControlled means we can never modify the attributes of the field.

src/classes/RD2_StatusMappingSettings_CTRL.cls
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /**
+     * @description Builds custom metadata type record for the status to state mapping
+     * @return Metadata.CustomMetadata
+     */
+    private static Metadata.CustomMetadata buildCustomMetadata(Mapping mapping) {
+        String sObjectName = RecurringDonationStatusMapping__mdt.getSObjectType().getDescribe().getName();
+
+        Metadata.CustomMetadata cmt = new Metadata.CustomMetadata();
```
force2b, line 181: I'm not sure you need to apply the time to the label, plus it looks a bit odd if you're viewing this in Salesforce Setup. The Label doesn't need to be unique anyway. I recommend the following:
- Label: PicklistValueApiName + " -> " + State (ex: `No_Response -> Closed`)

The `Name` field is unique and I like that it has the picklist value api name.

image

src/lwc/rd2StatusMappingSettings/rd2StatusMappingSettings.js
```diff
@@ -0,0 +1,382 @@
...
+        this.handleDeploymentProgress(deploymentId);
+    }
+
+    /***
+    * @description Starts polling for the deployment job progress details until the deployment completes
+    */
+    handleDeploymentProgress(deploymentId) {
+        var self = this;
+
+        this.deploymentTimer = setTimeout(function () {
```
force2b, line 234: @RadSFDeveloper Have you looked at the `DeploymentEvent__e` usage for this rather than just plain polling? The BdiAdvancedMapping component uses this for their CMDT deployment. I think it might be more efficient than polling.

RadSFDeveloper: Platform events do not work with Lightning Out. I did implement usage of `DeploymentEvent__e` but when the LWC was tested within VF context, it did not work. Good side of time spent is that I learned how to implement platform events. Unfortunately, it just does not work with Lightning Out.

src/lwc/rd2StatusMappingSettings/rd2StatusMappingSettings.html
RadSFDeveloper: The setting was copied from the [`utilPageLevelMessage`](https://github.com/SalesforceFoundation/NPSP/blob/master/src/lwc/utilPageLevelMessage/utilPageLevelMessage.html).I cannot reused the component since we need messages to behave like a toast message not a page level message. `ShowToastEvent` cannot be used in Lightning Out.

src/objects/RecurringDonationStatusMapping__mdt.object
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <description>Recurring Donation Status to State Mapping</description>
+    <fields>
+        <fullName>State__c</fullName>
+        <description>The state that the Recurring Donation Status will be mapped to.</description>
```
RadSFDeveloper, line 6: @sam-knox , can you please review new Recurring Donation Status Mapping custom metadata SObject field description and help text?

src/labels/CustomLabels.labels
```diff
@@ -2103,6 +2103,73 @@
...
+        <shortDescription>Mapping datatable Status field containing the API Name</shortDescription>
+        <value>API Name</value>
+    </labels>
+    <labels>
+        <fullName>RD2_StatusMappingDefinitions</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Status to state mapping definitions</shortDescription>
+        <value>&lt;b&gt;Status Definitions&lt;/b&gt;&lt;br/&gt;&lt;ul&gt;&lt;li&gt;Active: The donor is actively donating. NPSP creates installment Opportunities as defined by the Recurring Donation schedule.&lt;/li&gt;
```
RadSFDeveloper, line 2128: @sam-knox , can you please review new custom labels?

RadSFDeveloper:
@sam-knox , can you please review and update the PR Description above? The new class `RD2_StatusMappingSettings_CTRL` is using `@AuraEnabled` properties so admins need to grant the access and enable it for profiles that manage NPSP Settings.

FYI @jesslopez-sf , @coriobriensfdo , @mbeaty-sf

StaticResourceSources/customDatatable/style.css
```diff
@@ -0,0 +1,32 @@
+/*
```
force2b, line 1: I like this is a shared CSS library, I just wish we didn't have to bury it in a static resource. Unfortunately, it looks like it's the only way to do it right now. https://salesforce.stackexchange.com/questions/291394/importing-a-common-css-in-lwc

src/objects/RecurringDonationStatusMapping__mdt.object
sam-knox: @RadSFDeveloper - Sure thing - I have a couple of questions here. I'm less familiar with custom metadata types. Will there be one record of this object for each mapping a user creates? For State__c, would it be more accurate to say that the State is mapped to a Status versus will be mapped to a Status? In other words, don't these records only exist once a mapping has been saved?

RadSFDeveloper: If there are no new Status values, then no RD Status Mapping custom metadata type (CMT) record will exist, the RD2 will work fine since Active/Lapsed/Closed Status are mapped by default. When a new Status value exists and an admin saves mapping, records will be saved. Here is an example:

Screen Shot 2020-05-14 at 12 00 24 PM

RadSFDeveloper:
@sam-knox , @force2b : Status to State mapping page is ready for review.

sam-knox:
@RadSFDeveloper - I can't seem to get this working in a scratch org - would you mind posting or sending me a screenshot of the page?

RadSFDeveloper:
> @RadSFDeveloper - I can't seem to get this working in a scratch org - would you mind posting or sending me a screenshot of the page?

No prob, here is the screenshot. Is RD2 enabled on the org? The page is displayed only then. Let me know if there is any issue with viewing the page or the content so I will check into it.

Screen Shot 2020-05-15 at 12 19 38 PM

RadSFDeveloper:
@sam-knox , here are additional screenshots when deployment is in progress and then when completes successfully.

Screen Shot 2020-05-15 at 12 25 45 PM

Screen Shot 2020-05-15 at 12 27 34 PM

force2b:
@RadSFDeveloper I got an error attempting to save my mappings.
image

Plus, there's a build error, which I think is a result of the merge into `feature/rd2`. https://sfdo-metaci.herokuapp.com/builds/238203

RadSFDeveloper:
Good catch @force2b ! The Status API name had space and the custom metadata record Name field does not allow spaces. It has been fixed as well as the `feature/RD2` merge issues for which builds failed.
I also made changes to *not* refresh the datatable records on error to keep mapping as user specified it on the save. When user clicks [Cancel], data will be as it was before the save attempt. The records are loaded when the Status to State Mapping page is viewed on a new org or when deployment completes. They are refreshed when the save succeeds.

Screen Shot 2020-05-18 at 11 42 09 AM

src/classes/RD2_StatusMappingSettings_CTRL.cls
```diff
@@ -0,0 +1,534 @@
...
+     */
+    @TestVisible
+    private static List<Metadata.CustomMetadata> buildCustomMetadataRecords(String jsonMapping) {
+        List<Mapping> mappings = (List<Mapping>) JSON.deserialize(
+            jsonMapping, List<Mapping>.class
+        );
+
+        List<Metadata.CustomMetadata> records = new List<Metadata.CustomMetadata>();
+        for (Mapping mapping : mappings) {
+            records.add(buildCustomMetadata(mapping));
```
force2b, line 239: Fourth attempt at this.

This all works perfectly as is, but I have a question.

Right now, this creates CMDT records for the 3 hard coded status values of Active, Lapsed and Closed. It's possible for a custom to manually edit those records in Salesforce setup. And, when they do those, changes show up in the new UI.
image

Luckily though, the validation and other services ignore the mappings for Active, Lapsed and Closed and use only the hard-coded internal mappings for those. My question is whether it's even worth creating CMDT records for these 3 hard-coded values at all. We definitely don't want customers to change what "Active" means this way. If they want to use a different label, they can easily override the label for "Active" without modifying the api name or meaning of it. Effectively, I think it might be better to just store the custom mappings without the 3 hard-coded ones.

RadSFDeveloper: The change has been done. The Save action will store the Active, Lapsed and Closed Status mapping records.

RadSFDeveloper:
@sungysang ,if you have already created a scratch org, can you re-deploy code to the scratch org again to include the latest commit for the loading spinner alt text to use the label instead of hard-coded "Loading" text (text is not displayed)?

PR #5424, opened 2020-05-06
 W-7312506 Post Install Script for Custom Metadata Type Records

mldyang:
Update "without sharing" to "inherited sharing" BDI_MigrationMappingUtility due to Insufficient Privilege error on package upgrade.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_MigrationMappingUtility.cls
+++ b/src/classes/BDI_MigrationMappingUtility.cls
@@ -34,7 +34,7 @@
 * @group-content ../../ApexDocContent/BatchDataImport.htm
 * @description Methods to handle converting data import help text mappings to custom metadata records
 */
-public class BDI_MigrationMappingUtility {
+public inherited sharing class BDI_MigrationMappingUtility {
     /*******************************************************************************************************
     * @description Metadata queued up for deployment
     */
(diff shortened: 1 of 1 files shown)


mldyang:
@deepmalamehta Hi just wanted to give you a heads up on behalf of @Julian88Tex about this PR that he is about to merge. We tested Advanced Mapping can still be enabled, and after merging Julian will verify it can be still enabled in an org with the beta package installed. We do not believe the changes would impact other elements of BDI. The changes have been vetted by @callen-sfdo @psadie-sfdo @ptewson-sfdo . Thanks!

smarovich:
@mldyang @Julian88Tex I'm trying to decide if this needs to go in the regular release notes or just the Future Features section. Will this change impact Data Import in general, or just our package install?

Julian88Tex:
@smarovich So I think the general gist of the change is that (once this works) customers will not need to perform any action to map fields that are new to elevate. In my mind, it wouldn't make sense to talk about this in current release but would be something for a future release once Elevate launches.

Julian88Tex:
@smarovich Something else to note though is that we just rolled back these changes due to issues so it definitely won't be in the next NPSP release.

PR #5425, opened 2020-05-06
Fixed length donations during RD migration dry run

### LINKED ISSUE #5409: Enhanced Recurring Donations Migration Validation Picklist Error
Existing customers who run the **Validate Migration** process on the "Upgrade to Enhanced Recurring Donation" page may receive an error regarding restricted picklist values on the **Recurring Type field**:
`Recurring Type: bad value for restricted picklist field: Fixed`

Screen Shot 2020-05-01 at 5 17 19 PM

### Workaround

- Go to Setup --> Object Manager --> Recurring Donation
- Click on Fields & Relationships
- Click on **Recurring Type**
- Scroll down to Values and click the New button. Enter: Fixed
- Save.

### Full Error
Screen Shot 2020-05-01 at 5 18 13 PM

npsp-reedestockton:
We fixed an issue with the handling of fixed length donations during Enhanced Recurring Donations migration dry run.

# Critical Changes

# Changes

# Issues Closed
- Fixes #5409 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_DataMigrationMapper.cls
+++ b/src/classes/RD2_DataMigrationMapper.cls
@@ -321,11 +321,19 @@ public class RD2_DataMigrationMapper {
         }
     }
 
+    /**
+     * @description Sets RecurringType__c to 'Open' if 'Fixed' picklist option not available during dry run.
+     * @param rd Recurring Donation in the legacy format
+     * @return void
+     */
     private void setFixedToOpenDuringDryRun(npe03__Recurring_Donation__c rd) {
         if (!new RD2_RecurringDonation(rd).isFixedLength()) {
             return;
         }
-        rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
+
+        if (!RD2_EnablementService.fixedOptionAvailable) {
+            rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
+        }
     }
 
     /**
--- a/src/classes/RD2_DataMigrationMapper.cls
+++ b/src/classes/RD2_DataMigrationMapper.cls
@@ -110,6 +110,10 @@ public class RD2_DataMigrationMapper {
         setStatus(rd);
         setOtherFields(rd);
 
+        if (isDryRunMode) {
+            setFixedToOpenDuringDryRun(rd);
+        }
+
         return rd;
     }
 
@@ -315,6 +319,13 @@ public class RD2_DataMigrationMapper {
         if (rd.RecurringType__c == RD2_Constants.RECURRING_TYPE_OPEN) {
             rd.npe03__Installments__c = null;
         }
+    }
+
+    private void setFixedToOpenDuringDryRun(npe03__Recurring_Donation__c rd) {
+        if (!new RD2_RecurringDonation(rd).isFixedLength()) {
+            return;
+        }
+        rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
     }
 
     /**
--- a/src/classes/RD2_EnablementService.cls
+++ b/src/classes/RD2_EnablementService.cls
@@ -123,6 +123,13 @@ public with sharing class RD2_EnablementService {
         private set;
     }
 
+    /**
+     * @description Evaluates whether an open Recurring Donation is fixed length regardless of context.
+     * This method must be able to respond appropriately whether or not RecurringType__c picklist option 'Fixed'
+     * is available - it will not be in orgs which were created prior to Enhanced Recurring Donations.
+     * @param rd Recurring Donation
+     * @return Boolean
+     */
     public static Boolean isFixedLength(npe03__Recurring_Donation__c rd) {
         if (!isRecurringDonations2EnabledOverride) {
             return rd.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED;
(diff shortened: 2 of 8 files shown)


npsp-reedestockton:
**lurch:attach W-041168

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hZnOQAU" target="_blank">W-041168</a>

src/classes/RD2_RecurringDonation.cls
```diff
@@ -245,7 +245,25 @@ public inherited sharing class RD2_RecurringDonation {
     * @return Boolean
     */
     public Boolean isFixedLength() {
-        return rd.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED;
+        if (!RD2_EnablementService.isRecurringDonations2EnabledOverride) {
+            return rd.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED;
+        }
+
+        if (UTIL_Describe.isValidPicklistOption(String.valueOf(
```
force2b, line 252: To reduce performance implications, I recommend doing the check for the picklist value only once rather than for each individual RD record.

npsp-reedestockton: @RadSFDeveloper @force2b 

I'd like to discuss this a bit with you and Radmila.  Since RD2_RecurringDonation.cls is a wrapper for the record, we cannot use this logic without moving it out of this class or converting it to a static method within this class.  I debated with myself back and forth on how to handle this and I am frankly not sure of the right choice. 

My initial take was the same as yours, Mike, that we should move it somewhere appropriate for a static method that is cached for the entire context.  But, I think that implies changing RD2_RecurringDonation.isFixedLength() to a static method as well, or at least a method that calls the static context.  Either that or we do not try to re-use the logic.

What do the two of you think about this?

npsp-reedestockton: @force2b @RadSFDeveloper Also, we discussed the proxy idea in our team dev meeting yesterday before you arrived.  I had initially used npe03__Open_Ended_Status__c, but there were concerns that a 3rd party validation might rely on that.

RadSFDeveloper: With regards to proxy field: That is right, changing an RD1 field might trigger client's validations or triggers that will prevent validation to succeed. Thus, the idea is to use the Open Ended Status as-is in addition to other checks to determine if the RD is a fixed-length. So, if it is a dry run and Recurring Type is Open but the Open Ended Status is None/`null`, then this is a simulated fixed-length RD. In other conditions, we will check for the Recurring Type Fixed as usual to determine if the RD is fixed-length.

With regards to performance: I agree we should be using a cache (static var/method) to check if Recurring Type "Fixed" value is present. I thought we will use a new static method in `RD2_EnablementService` for it (I have not reviewed the PR yet so perhaps it is already done). We can discuss details offline.

npsp-reedestockton:
@force2b @RadSFDeveloper Ready for your review.
@sam-knox Please review PR description.
Thanks all!

src/classes/RD2_EnablementService.cls
```diff
@@ -104,4 +104,40 @@ public with sharing class RD2_EnablementService {
         return 
             UTIL_CustomSettingsFacade.getRecurringDonationsSettings().isRecurringDonations2Enabled__c == true;
     }
+
+    /**
+     * @description Returns true if RecurringType__c 'Fixed' picklist option available
+     */
+    @TestVisible
```
force2b, line 111: `@TestVisible` isn't needed if this is public.

src/classes/RD2_EnablementService.cls
```diff
@@ -104,4 +104,40 @@ public with sharing class RD2_EnablementService {
...
+                        SObjectType.npe03__Recurring_Donation__c.fields.RecurringType__c.Name,
+                        RD2_Constants.RECURRING_TYPE_FIXED
+                );
+            }
+            return fixedOptionAvailable;
+        }
+        private set;
+    }
+
+    public static Boolean isFixedLength(npe03__Recurring_Donation__c rd) {
```
force2b, line 126: ApexDoc comment block needed here. Would probably be helpful if the description explained the logic and why it's there.

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -317,6 +321,13 @@ public class RD2_DataMigrationMapper {
         }
     }
 
+    private void setFixedToOpenDuringDryRun(npe03__Recurring_Donation__c rd) {
```
force2b, line 324: ApexDoc comment block needed here.

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -317,6 +321,13 @@ public class RD2_DataMigrationMapper {
...
+        if (!new RD2_RecurringDonation(rd).isFixedLength()) {
+            return;
+        }
+        rd.RecurringType__c = RD2_Constants.RECURRING_TYPE_OPEN;
```
force2b, line 328: If I'm not mistaken, even in DryRunMode, if the "Fixed" picklist option exists this should set RecurringType to "Fixed". We're only using "Open" in the event that the picklist value does not exist or is inactive.

src/classes/RD2_EnablementService.cls
npsp-reedestockton: That's needed in order to access the private setter.

force2b: Good point.

src/classes/RD2_DataMigrationMapper.cls
npsp-reedestockton: Yes.  Thanks for catching that!

PR #5443, opened 2020-05-11
Gift Entry: Adv Mapping - new group bug fix

### LINKED ISSUE #5416: Gift Entry tab throws null object error when standard data mapping group is empty
I've enabled the new Gift Entry feature in a sandbox.  I'm getting an error "Attempt to de-reference a null object" error when I click on the tab.  This is because I've deleted all field mappings for the standard Payment Group because we want to use the Linvio Payment object instead of the NPSP Payment.

If I put one field mapping back into the standard Payment Group, then the error goes away, but we don't want this.

vmartinov-sforg:
- Creating a new Advanced Mapping Object Group without Field Mappings causes errors in Gift Entry
# Critical Changes

# Changes

# Issues Closed
Fixes #5416 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


vmartinov-sforg:
**lurch: attach W-039432

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsGBQA0" target="_blank">W-039432</a>

melissabarber:
Thank you @vmartinov-sforg! I updated the description of the PR to remove the information from the Changes section because I think the information about the Issue covers the work done in the PR. I also removed the link markdown for the Issue because using the # will connect this to the PR (At one point the markdown for Issue linking broke the release note aggregation, but it doesn't anymore.)

PR #5445, opened 2020-05-12
Gift Entry: Reduce Initial Load Time

kenlewis-sfdo:
This set of changes aims to reduce the initial load time of the Gift Entry tab by:

1. Deferring retrieval of field and object mappings until the template builder is initialized (reducing the amount of resources loaded from the server when a User lands on the main Gift Entry tab), and
2. Changing the mechanism for determining the current namespace.  Previously the app was using BDI_ManageAdvancedMappingCtrl.getNamespaceWrapper which took a lot of server time instantiating the BDI_ManageAdvancedMappingCtrl class.  The changed components now infer the namespace by inspecting an object or field reference instead of going to Apex (reducing the overall amount of server time used on initial load). 

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


kenlewis-sfdo:
Replaced by: https://github.com/SalesforceFoundation/NPSP/pull/5472

PR #5449, opened 2020-05-12
Fix migration utility legacy object name check

### LINKED ISSUE #5529: Payment Authorization Token Error Surfaced During Advanced Mapping Enablement
When enabling Advanced Mapping via NPSP Settings, customers may see a warning about an invalid Help Text Mapping for the **Payment Authorization Token** field. This can be ignored for now and should not affect current functionality.

**Steps to Repeat**

- Go to NPSP Settings --> System Tools --> Advanced Mapping for Data Import & Gift Entry
- Click the toggle to enable Advanced Mapping

Screen Shot 2020-06-01 at 9 33 29 AM
Screen Shot 2020-06-01 at 9 33 35 AM

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed
Fixes #5529 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata


psadie-sfdo:
[Tracking W-7782995](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EIXmIAO/view)

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -913,15 +903,21 @@ public without sharing class BDI_MigrationMappingUtility {
                 }
             }
 
-            if (hasDataImportNames == true) {
-                Integer countWhitespaces = helpText.countMatches(' ');
```
dfuller114, line 917: I see you removed countMatches() from here but it's still being used on line 896 and 898. Do we still need it there when parsing out the data import object names from the special characters?

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -892,8 +883,7 @@ public without sharing class BDI_MigrationMappingUtility {
         private List<String> splitHelpText(String helpText) {
             String symbol;
             Boolean hasSpecialChar = helpText.containsAny('$&+,:;=?@#|\'<>.-^*()%!_');
```
dfuller114, line 885: Could we define the `specialCharacters` list at the top of the method and then perform a `String.join()` when defining the `hasSpecialChar` boolean so that we don't have to hard code the special characters in full twice?

dfuller114:
@psadie-sfdo just added a couple of comments

src/classes/BDI_MigrationMappingUtility.cls
vmartinov-sforg: This is pre-existing code, but I don't quite follow the reason for this line `if (parts.size() >= 2 && countWhitespaces <= 4)`. [The `split` line above set a limit to 2, and I don't get the reason for the countWhitespaces]

ptewson-sfdo: I'll let @psadie-sfdo respond definitively, but I think the criteria here were part of a *somewhat* arbitrary set of metrics hashed out through an extensive set of conversations, where the end goal is to attempt to identify the distinction between "help text that was intended to be a field mapping" and "help text that was standard help text not intended to be a field mapping". 

The first part: If parts.size() == 2, then we know the help text contained a special character that MIGHT have been intended to split the object name and the field name, which is supposed to be '.' but we gave a lot of latitude for other possible delimiters. 

The second part: I think "4" is very arbitrary, but the idea is that more than a certain threshold of spaces probably indicated an intent to set human-readable help text in the form of a sentence, but fewer than that threshold (along with a delimiting character) might indicate a misguided use of a field label like "Opportunity Close Date" instead of the API name, which of course might have no spaces.  

Honestly, at the time, and to this day still, I think we went too far in the direction of charitably attempting to view something as an attempt at help text mapping, although I suppose something like "Opportunity$Opportunity Close Date" is as likely as not to be an attempted mapping? I wouldn't mind tightening up this criteria a bit, just to avoid inadvertent interpretations of "normal" help text as an attempted mapping gone awry, which I think is basically what introduced the bug we are fixing here.

ptewson-sfdo: @psadie-sfdo - This code might benefit from some comments explaining some of the thinking in more detail, regardless

src/classes/BDI_MigrationMappingUtility.cls
ptewson-sfdo: @dfuller114 @psadie-sfdo - I like the idea of defining this list once like Daniel suggested, but I'm seeing the characters being escaped differently between the two ("." vs "\\." for instance) - not sure how that would work exactly but only like the change if we are positive it doesn't cause weird bugs.

src/classes/BDI_MigrationMappingUtility.cls
psadie-sfdo: Patrick's spot on. I'm taking another pass at this. Add comments and _possibly_ tighten up the criteria if it feels safe. Completely agree that this was trying to be overly helpful. I'll ping for a re-review when that's ready. Thanks folks!

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -874,53 +868,43 @@ public without sharing class BDI_MigrationMappingUtility {
...
+        private List<String> retrieveObjectAndFieldApiNames(String helpText) {
+            if (helpText.contains('.')) {
+                List<String> parts = helpText.split('\\.', 2);
+
+                if (parts.size() == 2) {
+                    String possibleObjectApiName = parts[0];
+                    String possibleFieldApiName = parts[1];
+                    Boolean isLegacyObjectApiName =
+                        LEGACY_DATA_IMPORT_OBJECT_NAMES.contains(possibleObjectApiName.toLowerCase());
+                    Boolean isPossibleFieldApiName = !possibleFieldApiName.contains(' ');
```
ptewson-sfdo, line 900: Can help text have leading or trailing whitespace? I wonder if we should toss a ".trim()" in here just to avoid nixing a possible mapping because it has a space at the end of it

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -874,53 +868,43 @@ public without sharing class BDI_MigrationMappingUtility {
...
+        /*******************************************************************************************************
+        * @description Retrieves an object and field api name from a DataImport__c field help text mapping.
+        *
+        * @param helpText Help text value of a DataImport__c field
+        * @return A list of strings containing an object api name and a field api name OR null if help text
+        * value doesn't contain a valid help text mapping.
+        */
+        private List<String> retrieveObjectAndFieldApiNames(String helpText) {
+            if (helpText.contains('.')) {
+                List<String> parts = helpText.split('\\.', 2);
```
psadie-sfdo, line 893:
```suggestion
                List<String> parts = helpText.trim().split('\\.', 2);
```

@ptewson-sfdo does this seem sufficient to you with regards to clearing our preceding and/or trailing spaces?

ptewson-sfdo: @psadie-sfdo - yep that looks great!

PR #5455, opened 2020-05-13
 RD2: Disable Rollups during Data Validation and Data Migration

meena-sfdc:
# Critical Changes

# Changes

## Enhanced Recurring Donations 
- To improve performance, you can disable Customizable Rollups during the Data Validation and Data Migration steps of Enhanced Recurring Donations upgrade by selecting Disable CRLP When Creating Installments in custom settings. For more information see https://powerofus.force.com/s/article/NPSP-Enhanced-RD-Upgrade-Guide

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


meena-sfdc:
**lurch: attach W-040486

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuB2QAK" target="_blank">W-040486</a>

meena-sfdc:
@RadSFDeveloper , This PR is ready for review!

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -2370,4 +2446,26 @@ private class RD2_DataMigration_TEST {
...
     }
 
+    /***
+    * @description Update two opportunities in the recurring donation to closed won for calculating rollups
+    */
+    private static void setFewOppsInRdToClosedWonForRollupTests(npe03__Recurring_Donation__c rd) {
+        List<Opportunity> oppToUpdate = new List<Opportunity>();
+        List<Opportunity> oppList =  oppGateway.getRecords(rd);
+
+        oppList.sort();
```
force2b, line 2456: What is this sorting by?

meena-sfdc: It is sorting by dates, so that I update the first two opps to closed won.

force2b: Is that defined somewhere as a custom sort interface? The default sorting of SObjects in a List collection is by the value in the "Name" field- https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_list_sorting_sobject.htm

meena-sfdc: Its not custom sort here, when I debugged, it was sorting right, that might be because the opp name has the close date in it. 
If that might not be accurate, I can add a separate method in oppGateway to sort by close dates in this case.

meena-sfdc:
@force2b, I have made the changes and ready for review!

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -2370,4 +2446,25 @@ private class RD2_DataMigration_TEST {
         return Database.insert(new Opportunity(), false);
     }
 
+    /***
+    * @description Updates first two opportunities in the recurring donation to closed won for calculating rollups
+    */
+    private static void setFewOppsInRdToClosedWonForRollupTests(npe03__Recurring_Donation__c rd) {
```
force2b, line 2452: What you think about making this more specific by changing the method signature?
```suggestion
    private static void setRdOppsToClosedWonForRollupTests(npe03__Recurring_Donation__c rd, Integer numOpps) {
```

meena-sfdc:
@force2b and @RadSFDeveloper , The changes are done and ready for review!

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -2370,4 +2446,25 @@ private class RD2_DataMigration_TEST {
         return Database.insert(new Opportunity(), false);
     }
 
+    /***
+    * @description Updates first two opportunities in the recurring donation to closed won for calculating rollups
+    */
+    private static void setRdOppsToClosedWonForRollupTests(npe03__Recurring_Donation__c rd, Integer numOpps) {
+        List<Opportunity> oppToUpdate = new List<Opportunity>();
+        List<Opportunity> oppList =  oppGateway.getRecordsByCloseDate(rd);
```
RadSFDeveloper, line 2454: Data type such as List, Map, Set are not required in the names since methods are short and we can see what each var data type is and how it can be used, so we can just call this var `opps` or `rdOpps` etc.

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -2370,4 +2446,25 @@ private class RD2_DataMigration_TEST {
...
+
+        Integer count = 0;
+
+        for (Opportunity opp: oppList) {
+            if (count < numOpps) {
+                opp.StageName = UTIL_UnitTestData_TEST.getClosedWonStage();
+                opp.CloseDate = DATE_ESTABLISHED.addMonths(1);
```
RadSFDeveloper, line 2461: Why is the `CloseDate` reset for two oldest Opps? Date Established is 3 months ago from `today`, so there should not be any need to reset the `CloseDate`.

meena-sfdc: I thought when you close an opportunity, it sets the close date as today, that's why changed to a past date.

RadSFDeveloper: This happens only for future Opps not the past ones.

meena-sfdc: When I checked debug logs, it is not resetting the closedates to today, so I will take it out then, good point,thanks!

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -2370,4 +2446,24 @@ private class RD2_DataMigration_TEST {
...
+    /***
+    * @description Updates first two opportunities in the recurring donation to closed won for calculating rollups
+    */
+    private static void setRdOppsToClosedWonForRollupTests(npe03__Recurring_Donation__c rd, Integer numOpps) {
+        List<Opportunity> oppToUpdate = new List<Opportunity>();
+        List<Opportunity> rdOpps =  oppGateway.getRecordsByCloseDate(rd);
+
+        Integer count = 0;
+
+        for (Opportunity opp: rdOpps) {
```
RadSFDeveloper, line 2458: Sorry I just realized the loop can be re-written. Will the following work:
```
        for (Integer i = 0; i < numOpps; i++) {
            rdOpps[i].StageName = UTIL_UnitTestData_TEST.getClosedWonStage();
            oppToUpdate.add(rdOpps[i]);
        }
```

meena-sfdc:
@RadSFDeveloper and @force2b , The changes are done and ready for review!

PR #5458, opened 2020-05-13
W7312506 - Post Install Script to create Migrated Custom Field Mapping records

mldyang:
1. Created new methods in BDI_MigrationMappingHelper and updated BDI_MigrationMappingUtility to deploy the correct Migrated Custom Field Mapping records
2. Updated BDI_MigrationMappingUtility to "without sharing" due to "Insufficient Privileges" error when deploying custom metadata type records via post install script. 

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


mldyang:
Hi @deepmalamehta - wanted to give you a heads up on behalf of @Julian88Tex about this PR that he is about to merge. We tested Advanced Mapping can still be enabled, and after merging Julian will verify it can still be enabled in an org with the beta package installed. We do not believe the changes would impact other elements of BDI. Thanks!

Julian88Tex:
I also checked references to `BDI_MigrationMappingUtility` since we switched it from `inherited` to `without` sharing which could pose some security risks but didn't see any place that this class was being called that could cause it to be maliciously used.

PR #5463, opened 2020-05-14
Gift Entry: make  batch tab default on GE landing page

vmartinov-sforg:
# Critical Changes

# Changes
* HGE Set Batches tab to default (not Templates tab) on GE landing page
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


vmartinov-sforg:
**lurch: attach W-041134

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hZh6QAE" target="_blank">W-041134</a>

ptewson-sfdo:
@skristem this changes the user experience (the Gift Entry tab defaults the landing to batches instead of templates) and as a result I think some robot tests are failing. Any thoughts on how to best proceed?

skristem:
@ptewson-sfdo thanks for updating me. This means that the tests need to updated with new expectation so they do not fail. I see 2 options and I will let u suggest which works best. 
1. If you have at least 1.5 to 2 days to wait to merge this PR, then I can create a child branch of this and add code to fix tests(code and get code review and merge once approved)
2. If you don't have time to wait, I can mark these tests as unstable so that these tests will not run against all builds and make changes later and move back to regular robot flow.

ptewson-sfdo:
@skristem thanks for putting those options together - I think we can afford to wait a couple days, so if you can put the updated tests together in that time frame that would be great. Thank you so much!

PR #5475, opened 2020-05-15
Eliminate extra Account trigger AfterUpdate execution

### LINKED ISSUE #4614: Multi-Line Addresses on Contact Insertion Result in Unnecessary Account Update
In the NPSP 3.157 release, we optimized DML operations around Contact with Address insertions. However, Contacts with multi-line Street Addresses are still causing an unnecessary Account update. As a result, the Order of Execution for Contacts with Single-Line Addresses vs. Multi-Line Addresses is different and may cause unexpected results.

When we do a comparison in [ADDR_Addresses_TDTM](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/ADDR_Addresses_TDTM.cls) to determine if we need to refresh the Household Account's Address, we check to see if the Billing Street value is equal to the Address record's concatenated Mailing Street 1 and 2 fields. Even though they theoretically match, Apex is not seeing it that way, so we unnecessarily do another update to the Account... 

**_My full analysis of the order of execution and potential fix will be linked on our internal Work Item._**

Single-Line:
Screen Shot 2019-09-17 at 6 55 36 PM

Multi-Line:
Screen Shot 2019-09-17 at 6 52 27 PM

### Relevant Code 
[shouldUpdateAddress() method:](https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/ADDR_Addresses_TDTM.cls#L560-L581
)
Specifically, Line 568, it's not seeing the line break as the same:
`            !equalsCaseSensitive((String) sObj.get(prefix + 'Street'), getMultilineStreet(addr)) ||
`

And as a result, we add that Account to the listAccUpdate variable on Line 337.  We then add that Account to the dmlWrapper on Lines 373-375.
 ```
if (listAccUpdate.size() > 0) {
            dmlWrapper.objectsToUpdate.addAll((List<SObject>)listAccUpdate);
}
```
This then causes all of our Account Trigger Handlers to run again on Account Update.

### PR DESCRIPTION
**Pull request by npsp-reedestockton**
We normalized line breaks in street address lines during BeforeInsert and BeforeUpdate triggers for both Contact and Account to eliminate an unnecessary AfterUpdate execution of Account trigger.

# Critical Changes

# Changes

# Issues Closed
- Fixes #4614 

# Community Ideas Delivered

# New Metadata
Apex Classes:
- UTIL_Address
- UTIL_Address_TEST

# Deleted Metadata

### CODE REVIEW

**Comment by npsp-reedestockton** [comment]
**lurch: attach W-036153

**Comment by npsp-reedestockton** [comment]
@force2b @RadSFDeveloper Ready for review
@sam-knox Please review PR description

Thanks all!

**Comment by LurchTheButler** [comment]
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001CAxLQAW" target="_blank">W-036153</a>

File: src/classes/ADDR_Account_TDTM.cls [hunk 1]
```diff
@@ -47,6 +47,7 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
     public override DmlWrapper run(List<SObject> listNew, List<SObject> listOld, 
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
 
+        boolean isBefore = (triggerAction == TDTM_Runnable.Action.BeforeUpdate || triggerAction == TDTM_Runnable.Action.BeforeInsert);
```
RadSFDeveloper, line 50: - Suggestion: `boolean isBefore` -> `Boolean isBeforeContext`
- Do you mind deleting lines 64-70 since they are not used at all?

src/classes/UTIL_Address_TEST.cls
```diff
@@ -0,0 +1,93 @@
...
+
+@isTest
+public with sharing class UTIL_Address_TEST {
+
+    /***
+    * @description Verifies that addresses which should not be altered are not.
+    */
+    @isTest
+    private static void shouldEqualInputAddress() {
+        List<Account> accountList = new List<Account>();
```
RadSFDeveloper, line 45: - `shouldEqualInputAddress()` -> `shouldNotReplaceLineBreakWithCRLF()`
- Let's name vars without data type in the name, so `accountList` -> `accounts` and `contactList` -> `contacts`.
- Similarly `accountListNew` -> `clonedAccounts` and `contactListNew` -> `clonedContacts` 
- Can the address contain a mix of the `\n` and `\r\n`?

src/classes/UTIL_Address_TEST.cls
```diff
@@ -0,0 +1,93 @@
...
+            System.assertEquals(accountList[i].BillingStreet, accountListNew[i].BillingStreet, 'Normalized street address should be equal to input');
+            System.assertEquals(contactList[i].MailingStreet, contactListNew[i].MailingStreet, 'Normalized street address should be equal to input');
+        }
+    }
+
+    /***
+    * @description Verifies that addresses which should be altered are.
+    */
+    @isTest
+    private static void shouldNotEqualInputAddress() {
```
RadSFDeveloper, line 78: - `shouldNotEqualInputAddress()` -> `shouldReplaceLineBreakWithCRLF()`
- Can we also test case when the `MailingStreet2__c` is populated?

src/classes/UTIL_Address_TEST.cls
npsp-reedestockton: If an address contains a mix of \n and \r\n we intentionally leave it unchanged.  I've added test records to verify that behavior.

src/classes/UTIL_Address_TEST.cls
npsp-reedestockton: UTIL_Address only deals with contacts and accounts.  Are you suggesting a MailingStreet2__c test in ADDR_Addreses_TEST or ADDR_Addreses_TEST2?

RadSFDeveloper: I see your point. Is there a test verifying the `ADDR_Addresses_TDTM. getMultilineStreet()` conditional on `MailingStreet2__c` since that method has been modified? If not, can tests in the an `ADDR_` test class or `UTIL_Address_TEST` verify it is as expected?

src/classes/UTIL_Address_TEST.cls
```diff
@@ -0,0 +1,95 @@
+/*
+    Copyright (c) 2014, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2020, Salesforce.org
```

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -55,14 +56,11 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
             Account acc = (Account)so;
             i++;
 
-            // BEFORE INSERT
-            if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
+            // BEFORE INSERT & BEFORE UPDATE
+            if (isBeforeContext) {
+                UTIL_Address.normalizeLineBreak(acc);
```
force2b, line 61: I'm concerned that this logic (and in ADDR_Contact_TDTM) will run on every single update of an existing Account or Contact, even if the Mailing/Billing Street was not actually modified, causing modifications to existing addresses which could cause other triggers to execute because they detect that the address was modified. recommend splitting this logic up so that the new _normalize_ logic only runs if the record is new or the Street address was actually modified:
```suggestion
            if (TDTM_Runnable.Action.BeforeInsert) {
                UTIL_Address.normalizeLineBreak(acc);
            }
            if (TDTM_Runnable.Action.BeforeUpdate) {
                if (UTIL_Address.isAddressFieldModified('Street', acc, trigger.old[i]) {
                    UTIL_Address.normalizeLineBreak(acc);
                }
            }


// ...... in UTIL_Address
public static isAddressFieldModified(String addrFieldType, SObject newRecord, SObject oldRecord) {
    Boolean isContact = (sObj.getSObjectType() == Contact.SObjectType);
    String prefix = (isContact ? 'Mailing' : 'Billing');
    String fieldName = (String) sObj.get(prefix + addrFieldType);
    
    String newValue = (newRecord.get(fieldName) != null ? (String)newRecord.get(fieldName) : null);
    String oldValue = (oldRecord.get(fieldName) != null ? (String)oldRecord.get(fieldName) : null);
    return newValue != oldValue; 
}
```

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -55,14 +56,11 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
             Account acc = (Account)so;
             i++;
 
-            // BEFORE INSERT
-            if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
+            // BEFORE INSERT & BEFORE UPDATE
```
force2b, line 59: What you you think of refactoring this for loop to be a bit more logical?
```
for (Integer i = 0; i < listNew.size(); i++) {
    Account newRecord = listNew[i];
    Account oldRecord = (listOld != null ? listOld[i] : null);
    ....
}
```

src/classes/UTIL_Address_TEST.cls
```diff
@@ -0,0 +1,95 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Utilities
+* @description Tests specific to Address utility methods.
+*/
+
+@isTest
```
force2b, line 37: Since there is no dml in this, I recommend
```suggestion
@isTest(IsParallel=true)
```

src/classes/ADDR_Account_TDTM.cls
npsp-reedestockton: There is an existing method in ADDR_Account_TDTM that checks for an address change.  Is it worth checking for a change as granular as street if we've already determined there has been an address change?  I'm wondering if your concern is adequately addressed by relying on the existing method to flag a changed address?

src/classes/UTIL_Address_TEST.cls
npsp-reedestockton: ADDR_Addresses_TEST2 tests null MailingStreet2__c null behavior in multiple unit tests.

npsp-reedestockton:
@force2b If you are okay with using existing check for address changed, this is ready for you to review again.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -60,8 +60,27 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
         Map<Id,Account> mapAccountIdAccount = null;
         boolean acctHasAddress = false;
         boolean acctChanged = false;
+        boolean isBefore = (triggerAction == TDTM_Runnable.Action.BeforeUpdate || triggerAction == TDTM_Runnable.Action.BeforeInsert);
```
npsp-reedestockton, line 63: @force2b Added logic similar to ADDR_Account_TDTM, checking for changed address in BeforeUpdate.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -60,8 +60,27 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
+
+        if (isBefore) {
```
force2b, line 65: Thanks. There's an `if` online 84 that bail's out of address management is disabled, plus there's another `if (isBefore)` block starting at line 83. Wondering if this block needs to be merged into that block; and both are after the bail-out check.

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -71,17 +72,28 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
-                Contact con = (Contact)so;
-                if (con.AccountId != null)
-                    setAccountId.add(con.AccountId);
+            for (Integer i = 0; i < listNew.size(); i++) {
+                Contact newRecord = (Contact) listNew[i];
+                Contact oldRecord = (listOld != null ? (Contact) listOld[i] : null);
+
+                // need to query account fields we need to look at
+                if (newRecord.AccountId != null)
+                    setAccountId.add(newRecord.AccountId);
```
force2b, line 83: Curly brackets needed
```suggestion
                if (newRecord.AccountId != null) {
                    setAccountId.add(newRecord.AccountId);
                }
```

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -71,17 +72,28 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
+
+                // BEFORE UPDATE
+                if (triggerAction == TDTM_Runnable.Action.BeforeUpdate && isContactAddressChanged(newRecord, oldRecord)) {
+                    UTIL_Address.normalizeLineBreak(newRecord);
+                }
             }
-            mapAccountIdAccount = new Map<Id,Account>([select Id, npe01__SYSTEM_AccountType__c,
-                (select Id from Addresses__r) from Account where Id IN :setAccountId]);
+            mapAccountIdAccount = new Map<Id,Account>([SELECT Id, npe01__SYSTEM_AccountType__c,
+                (SELECT Id FROM Addresses__r) FROM Account WHERE Id IN :setAccountId]);
```
force2b, line 96:
```suggestion
    mapAccountIdAccount = new Map<Id,Account>([
        SELECT Id, npe01__SYSTEM_AccountType__c, (SELECT Id FROM Addresses__r) 
        FROM Account
        WHERE Id IN :setAccountId
    ]);
```

src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -71,17 +72,32 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
+                }
+
+                // BEFORE UPDATE
+                if (triggerAction == TDTM_Runnable.Action.BeforeUpdate && isContactAddressChanged(newRecord, oldRecord)) {
+                    UTIL_Address.normalizeLineBreak(newRecord);
+                }
             }
-            mapAccountIdAccount = new Map<Id,Account>([select Id, npe01__SYSTEM_AccountType__c,
-                (select Id from Addresses__r) from Account where Id IN :setAccountId]);
+    mapAccountIdAccount = new Map<Id,Account>([
```
force2b, line 96: @npsp-reedestockton  Sorry, indentation is off here.

npsp-reedestockton:
@force2b @RadSFDeveloper Added missing check for address management enabled in ADDR_Account_TDTM, and refactored run method.  Please have a look.  Thanks!

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -103,7 +98,24 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
                     
         return dmlWrapper;    
     }
-    
+
+    /*******************************************************************************************************
+    * @description utility to see if an address management is enabled
+    * @param newRecord an Account
+    * @return Boolean.  true if an address is specified in any of the Account Billing fields
```
force2b, line 105: The return doesn't really match what the code does.

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -48,53 +48,48 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
...
+
+                // BEFORE UPDATE
+                if (triggerAction == TDTM_Runnable.Action.BeforeUpdate
+                        && isAccountAddressChanged(newRecord, oldRecord)) {
+                    UTIL_Address.normalizeLineBreak(newRecord);
+                }
+
+                // AFTER INSERT
+                if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
+                    if (isAccountAddressSpecified(newRecord))
```
force2b, line 81: Can you add in the missing curly braces here and on line 96?

PR #5479, opened 2020-05-18
Callable API: Execute RD2 data migration

RadSFDeveloper:
When QA automation begins I want the scratch org with a Beta managed package to have RD2 Recurring Donations data ready so that automated tests can be initiated against the org.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Label:
- RD2_EnablementPermissionRequired

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -1907,6 +1907,14 @@
         <shortDescription>Migration section warning</shortDescription>
         <value>&lt;b&gt;WARNING: Run Migration modifies existing Recurring Donation data to match the Enhanced Recurring Donations model. Backup your database if you haven't already done so.&lt;/b&gt;</value>
     </labels>
+    <labels>
+        <fullName>RD2_EnablementPermissionRequired</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error displayed when user does not have System Admin permissions</shortDescription>
```
RadSFDeveloper, line 1915: @sam-knox , can you please review the new custom label?

RadSFDeveloper:
@force2b , the branch is named as `feature/api__rad-rd2-data-migration` based on the CRLP API branch, however the `feature/api` does not exist and both PRs are set to be merged into the master. Is this expected?

RadSFDeveloper:
**lurch: attach W-040677

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuVRQA0" target="_blank">W-040677</a>

force2b:
> @force2b , the branch is named as `feature/api__rad-rd2-data-migration` based on the CRLP API branch, however the `feature/api` does not exist and both PRs are set to be merged into the master. Is this expected?

Yeah. At first I had thought we might merge them all into a parent `feature/api` branch and then into master, but in retrospect these are all independent and can just merge directly into master.

src/labels/CustomLabels.labels
sam-knox: @RadSFDeveloper - Sure thing. I assume we're checking for Customize Application, Author Apex, and Modify All Data. Anything else?

RadSFDeveloper: Only those permissions, the same three that they are checked for the NPSP Settings page access.

cumulusci.yml
```diff
@@ -388,6 +388,16 @@ tasks:
                 %%%NAMESPACE%%%Error_Settings__c errSettings = %%%NAMESPACE%%%Error_Settings__c.getOrgDefaults();
                 errSettings.%%%NAMESPACE%%%OverrideFeature_PilotEnabled__c = true;
                 upsert errSettings;
+    
+    execute_rd2_data_migration:
```
RadSFDeveloper, line 392: @codexmarcel , @force2b : can you please review changes?

src/labels/CustomLabels.labels
RadSFDeveloper: @sam-knox , the custom label text has been updated. Thanks for the feedback.

cumulusci.yml
force2b: Using a task for this is optional. It's helpful when there may be a need to run a data migration completely independent of the enablement process, but I don't necessarily think that would ever happen. I think it might be better to just have this as part of the flow without the independent task. 
@codexmarcel Are you ok with that?

RadSFDeveloper: The `execute_rd2_data_migration` has been removed and replaced with `execute_anon` to run the data migration batch.

cumulusci.yml
```diff
@@ -863,6 +864,14 @@ flows:
                     apex: >
                         %%%NAMESPACE%%%RD2_OpportunityEvaluation_BATCH rdBatchJob = new %%%NAMESPACE%%%RD2_OpportunityEvaluation_BATCH();
                         Database.executeBatch(rdBatchJob, rdBatchJob.batchSize);
+            8:
+                task: execute_anon
+                # Update enhanced Recurring Donation enablement page state
+                options:
+                    apex: >                        
+                        npe03__Recurring_Donations_Settings__c rdSettings = npe03__Recurring_Donations_Settings__c.getOrgDefaults();
```
force2b, line 872: I had originally thought we'd do this in apex as part of the API, but I like this approach. The only caveat is that any future changes to the JSON in apex will need to be replicated here.

src/labels/CustomLabels.labels
RadSFDeveloper: @ravensisco , this is the text displayed when a user does not have permission to run RD2 data migration (AC5 scenario).

RadSFDeveloper:
Hi @codexmarcel , when you have time, can you please review changes to the `yml` file?

PR #5490, opened 2020-05-19
Users without any permission can access "Advanced Mapping for Data Import & Gift Entry" in NPSP Settings

dfuller114:
A user who does not have system admin privileges should not be able to view the Advanced Mapping for Data Import & Gift Entry lightning component in NPSP Settings.
- Custom Labels
   - commonAdminPermissionErrorMessage
   - commonAdminPermissionErrorTitle
# Critical Changes

# Changes
- We improved security around NPSP Settings so users must have system admin permissions to configure Advanced Mapping.
# Issues Closed

# Community Ideas Delivered

# New Metadata

- Apex Class
    - BDI_ManageAdvancedMappingController.cls
- Lightning
   - BD_ManageAdvancedMappingHelper.js

# Deleted Metadata


dfuller114:
fyi. Working on the test class for the new apex controller and will be pushing that up soon.

dfuller114:
**lurch attach: W-040912

dfuller114:
**lurch: attach W-040912

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hv8IQAS" target="_blank">W-040912</a>

dfuller114:
@melissabarber adding you as reviewer because the error message we display will need to be a custom label.

melissabarber:
@dfuller114 how will this error be displayed to the user? Will it be an error at the top of the page and have the enable toggle disabled or will it be a full-page error when a user tries to go to the Advanced Mapping for Data Import & Gift Entry page in NPSP Settings?

dfuller114:
> @dfuller114 how will this error be displayed to the user? Will it be an error at the top of the page and have the enable toggle disabled or will it be a full-page error when a user tries to go to the Advanced Mapping for Data Import & Gift Entry page in NPSP Settings?

@melissabarber Still working with Grace on that.

src/aura/BDI_ManageAdvancedMapping/BDI_ManageAdvancedMappingHelper.js
```diff
@@ -0,0 +1,46 @@
...
+        action.setParams({
+            'userId' : userId
+        });
+
+        action.setCallback(this, function(response) {
+            let state = response.getState();
+            const SUCCESS = 'SUCCESS';
+            const ERROR = 'ERROR';
+
+            if(state === SUCCESS) {
```
psadie-sfdo, line 15:
```suggestion
            if (state === SUCCESS) {
```

Could we add a single space between the if and parenthesis? There are a few of these.

src/classes/BDI_ManageAdvancedMappingController.cls
```diff
@@ -0,0 +1,51 @@
...
+*/
+public with sharing class BDI_ManageAdvancedMappingController {
+
+    /**
+     * @description Checks if the running user has system admin permissions
+     * @return Boolean
+    */
+    @AuraEnabled
+    public static Boolean isAdminUser(Id userId) {
+        List<User> runningUser = [SELECT ProfileId FROM User WHERE Id = :userId];
```
psadie-sfdo, line 44: Should we start adding perm checks for methods that run a query? Something like `UTIL_Permissions.getInstance().canRead(User.SObjectType)` perhaps

ptewson-sfdo: I think in general, yes we should include WITH SECURITY_ENFORCED - however this is an interesting one since we are actually checking permissions with this query so throwing a security error here seems almost redundant

ptewson-sfdo: We could also refactor this with UserInfo.getProfileId() to avoid doing a SOQL query at all but that might be overkill

src/classes/BDI_ManageAdvancedMappingController.cls
```diff
@@ -0,0 +1,51 @@
+/*
+    Copyright (c) 2017, Salesforce.org
```
ptewson-sfdo, line 2: Copyright (c) 2020

src/classes/BDI_AdvancedMappingControllerTest.cls
```diff
@@ -0,0 +1,54 @@
...
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+@isTest
+private class BDI_AdvancedMappingControllerTest {
```
ptewson-sfdo, line 31: This should be named BDI_AdvancedMappingController_TEST - and also needs to be included in "package.xml"!

src/aura/BDI_ManageAdvancedMapping/BDI_ManageAdvancedMappingHelper.js
```diff
@@ -0,0 +1,46 @@
+({
+    isSysAdmin : function(component) {
+        let userId = 'asdfsdf';
```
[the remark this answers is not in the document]
dfuller114: @psadie-sfdo I'm going to remove that. That was just to force a server side error to see if the toast was working.

src/aura/BDI_ManageAdvancedMapping/BDI_ManageAdvancedMappingHelper.js
dfuller114: @psadie-sfdo Will do.

src/classes/BDI_ManageAdvancedMappingController.cls
dfuller114: @ptewson-sfdo The reason for the query in this case is to get the profileId but also to get the `User` SObject record since that's what has to be passed as a parameter to the `isSysAdmin` method in the user service.

src/aura/BDI_ManageAdvancedMapping/BDI_ManageAdvancedMappingHelper.js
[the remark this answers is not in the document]
dfuller114: @psadie-sfdo fixed!

src/classes/BDI_ManageAdvancedMappingController.cls
dfuller114: @ptewson-sfdo I've added the security check to this query. I figure the user should always have access to their own user record, but this should at least prevent the query from being flagged in a Checkmarx, PMD, or other security scan.

src/classes/BDI_ManageAdvancedMappingController.cls
dfuller114: fixed!

src/classes/BDI_AdvancedMappingControllerTest.cls
dfuller114: update the existing test class that was previously created instead of creating a new one

src/aura/BDI_ManageAdvancedMapping/BDI_ManageAdvancedMappingHelper.js
dfuller114: fixed!

PR #5517, opened 2020-05-26
resolve error on dry run in batch gift entry

ptewson-sfdo:
Update the query on data import used to submit single row dry run to include necessary fields

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


ptewson-sfdo:
**lurch: attach W-041240

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hZtHQAU" target="_blank">W-041240</a>

src/classes/BDI_DataImportService.cls
```diff
@@ -72,6 +72,27 @@ global with sharing class BDI_DataImportService {
...
+    * @description constructs the soql String to run a Data Import process for list of DataImport ids.
+    * @param dataImportIds list of data import ids to include in query
+    * @return String the soql String
+    */
+    public static String strSoqlForDataImportProcess(List<Id> dataImportIds) {
+        String dataImportQuery;
+
+        // construct or SOQL for the batch
+        dataImportQuery = 'SELECT ' + String.join(listStrDataImportFields, ',');
+        dataImportQuery += ' FROM ' + UTIL_Namespace.StrTokenNSPrefix('DataImport__c');
```
kenlewis-sfdo, line 85: Can this line use the object token instead of introducing a string reference? 
`String.valueOf(DataImport__c)`

src/classes/BDI_DataImportService.cls
```diff
@@ -72,6 +72,27 @@ global with sharing class BDI_DataImportService {
...
+    /*******************************************************************************************************
+    * @description constructs the soql String to run a Data Import process for list of DataImport ids.
+    * @param dataImportIds list of data import ids to include in query
+    * @return String the soql String
+    */
+    public static String strSoqlForDataImportProcess(List<Id> dataImportIds) {
+        String dataImportQuery;
+
+        // construct or SOQL for the batch
+        dataImportQuery = 'SELECT ' + String.join(listStrDataImportFields, ',');
```
kenlewis-sfdo, line 84: Is `listStrDataImportFields` always a full list of all of the fields on DI?  If so does this change mean that all fields will always be queried in both dry run and regular processing?

src/classes/BDI_DataImportService.cls
voduyemi: @kenlewis-sfdo I think if we want to avoid the direct string reference, we might need to use the describe info of `DataImport__c`: `DataImport__c.SObjectType.getDescribe().localName`. The `valueOf` method accepts just the `Object` parameter

ptewson-sfdo: I wasn't sure how much to keep this in parallel with the method directly below it, "strSoqlForBatchProcess()" - ideally we would refactor both to use the describe call with the localName or something, but that method is very foundational to BDI and gets called in so many contexts, I just want to be sure not to introduce a namespace bug somehow (and also want to keep these two methods in sync if possible, since they are similar in their function)

src/classes/BDI_DataImportService.cls
ptewson-sfdo: So one thing I learned during this is that we are already querying all fields in regular processing - that's actually what the strSoqlForBatchProcess(Id batchId) method is handling. When I started building this I was trying to gather all the necessary fields for dry run more carefully, but what I learned is that for at least some objects (Address for example), dry run references every mapped field.

src/classes/BDI_DataImportService.cls
vmartinov-sforg: I agree, I would go with the low-risk option given the widespread use of the method.

kenlewis-sfdo: In that case I would vote to refactor the other method.  IMO we should avoid introducing string references to known objects/fields and should be cleaning those up in old code where encountered.  Using tokens preserves referential integrity and bypasses the need to do any manual namespace checking.

ptewson-sfdo: I will make that update - I certainly agree in principal that we should avoid string references; my only reservation is breaking the seal on code that's been running in production successfully for so long! But good point - let's get this up to date.

src/classes/BDI_DataImportService.cls
```diff
@@ -72,6 +72,27 @@ global with sharing class BDI_DataImportService {
...
+        dataImportQuery += ' FROM ' + DataImport__c.SObjectType.getDescribe().localName;
```
kenlewis-sfdo, line 85: I would suggest one of two strategies here: we can either use the describe result to get the name like this, but maybe go through our `UTIL_Describe` class to get the benefit of caching, or `String.valueOf(DataImport__c.SObjectType)` to avoid the call to get the describe.  
Either way, I think we specifically want to avoid `localName` as it does NOT include the namespace even when in a namespaced context:
https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_methods_system_sobject_describe.htm

voduyemi: Thanks for calling that out @kenlewis-sfdo

PR #5519, opened 2020-05-27
Gift Entry: send gateway id parameter to Tokenize card method

vmartinov-sforg:
# Critical Changes

# Changes
* add the gateway id as a parameter to the call to Tokenize card
* display the partial gateway id in the cc widget as it's being sent to elevate

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


vmartinov-sforg:
**lurch: attach W-041520

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007haT7QAI" target="_blank">W-041520</a>

src/pages/GE_TokenizeCard.page
```diff
@@ -63,6 +79,9 @@
                 // Reject any messages from an unexpected origin
                 return;
             } else if (event.data && event.data.action === 'createToken') {
+                // TODO - remove once elevate can process the gateway id
```
ptewson-sfdo, line 82: @vmartinov-sforg I'm curious about the timing of this feature and how/when we remove it. I think part of the desire to have Elevate available in the package is to enable our broader team to demo to customers - so even though this isn't available to our broader customers we may want to be careful about this kind of visible artifact in support of testing. Not sure what the actual level of concern here is, will check with product

vmartinov-sforg: @ptewson-sfdo Checking with product sounds good. We could remove all the visible items as soon as QE is completed.

src/pages/GE_TokenizeCard.page
```diff
@@ -10,6 +10,12 @@
             <div id="payments"></div>
         </form>
     </article>
+    
+    <!--TODO - remove once elevate can process the gateway id-->
+    <!-- for QE: to display visually the gateway id when it's passed to tokenize method -->
+    <article class='slds-scope'>
+        <div><small><span id="gatewayInfo"></span></small></div>
```
ptewson-sfdo, line 17: @vmartinov-sforg Was there supposed to be content within this tag? I couldn't find it initially and I realized it's because there's nothing actually rendered within the span or div tag!

vmartinov-sforg: The text of the span is set at run-time to display the gateway is when the button is clicked. @ptewson-sfdo

PR #5547, opened 2020-06-08
Telemetry: RD2 Enablement State

andrewyu-salesforce:
As a Product Manager responsible for RD2, I need to know if customers have skipped steps during the RD2 enablement process so that I can prioritize enhancements to ensure customers are alerted to these skipped steps.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_OrgTelemetry_SVC.cls
+++ b/src/classes/UTIL_OrgTelemetry_SVC.cls
@@ -274,7 +274,7 @@ public without sharing class UTIL_OrgTelemetry_SVC {
         } else if (isMetaDeployNotCompleted()) {
             rd2EnablementStateIndicator = 1;
 
-        } else if (isNextPaymentDateRollupEnabled()) {
+        } else if (!isNextPaymentDateRollupDisabled()) {
             rd2EnablementStateIndicator = 2;
 
         } else {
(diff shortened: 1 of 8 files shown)


andrewyu-salesforce:
**lurch: attach W-041805

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hayyQAA" target="_blank">W-041805</a>

andrewyu-salesforce:
@force2b The PR is ready for review.

src/classes/UTIL_OrgTelemetry_SVC.cls
```diff
@@ -262,6 +258,86 @@ public without sharing class UTIL_OrgTelemetry_SVC {
...
+            rd2EnablementStateIndicator = 0;
+
+        } else if (isMetaDeployNotCompleted()) {
+            rd2EnablementStateIndicator = 1;
+
+        } else if (!isNextPaymentDateRollupDisabled()) {
+            rd2EnablementStateIndicator = 2;
+
+        } else {
+            rd2EnablementStateIndicator = 3;
```
force2b, line 281: @andrewyu-salesforce I was not really thinking ahead when I came up with this number scheme. For example, 3.179 will include a new custom UI with New/Edit overrides in the unpackaged metadata (and likely other small changes). It would be nice if we could eventually detect that the new UI isn't being used and include that in our results. 

Anyway, what you think of changing "3" to "10" here and the unit test to represent that RD2 is fully enabled? That gives us a little more wiggle room to add in additional metrics over time, _if needed_.

andrewyu-salesforce: Changing to 10 make sense if we are going to track more step in the future. I will do the update.

force2b: Thank you. I updated the WI description to match.

PR #5548, opened 2020-06-08
Gift Entry:  unallocated bug fix

### LINKED ISSUE #5570: HGE: Un allocated sum is not displayed correctly when batch row is open from table
Description: Un allocated sum is not displayed correctly when batch row is open from table and GAU allocations are using percents. It is working fine with amount.
Steps:
1. Create a batch gift and 2 allocations (One with Percent and one with amount).
2. Save the gift.
3. Open the gift from batch row.
4. Amount below is not accurate ( this should show sum of un allocated amount)
See screenshot.
rtaImage

vmartinov-sforg:
* Fix bug: Un allocated sum is not displayed correctly when batch row is open from table and GAU allocations are using percent.
# Critical Changes

# Changes

# Issues Closed
Fixes #5570

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


vmartinov-sforg:
**lurch: attach W-041640

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hajKQAQ" target="_blank">W-041640</a>

bdvorachek:
I found one UX bug I think we should address with this component if possible as a part of this work item. When a user loads a donation that has a GAU allocation which has a percent amount assigned to it, the Amount field should be disabled for that GAU allocation row. This is currently how it works when you key in the value in the percent field manually, but not when loading an existing donation.

The rest of this is WI is working as I expect it to 👍

src/lwc/geFormWidgetAllocation/geFormWidgetAllocation.js
```diff
@@ -349,7 +349,17 @@ export default class GeFormWidgetAllocation extends LightningElement {
                 let localFieldName = AMOUNT_FIELD.fieldApiName;
                 let currentKey = current.record.hasOwnProperty(fullFieldName) ? fullFieldName : localFieldName;
                 
-                const currentAmount = current.record[currentKey];
+                let currentAmount = current.record[currentKey];
+                if (isEmpty(currentAmount)) {
+                    // amount is empty, use the percent field
+                    let fullFieldNamePercent = `${ALLOCATION_OBJECT.objectApiName}.${PERCENT_FIELD.fieldApiName}`;
```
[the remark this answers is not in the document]
vmartinov-sforg: Done.

vmartinov-sforg:
@bdvorachek Can you please ping me when you get a chance, I have some questions re: ux bug above.

psadie-sfdo:
Functionality looks good to me and I think we could wrap this in terms of completing the WI requirements.

I ran into non-blocking bug regarding the warning label when we under/over-allocate. The widget is unable to find the label for the donation amount field and is showing up as 'undefined'. Seems to be happening because we never call `GeFormService.getFormTemplate()` which sets the `donationFieldTemplateLabel` property when we're in batch mode.

A quick fix would be to simply call `GeFormService.getFormTemplate()` for batch mode as well, around lines 184 - 189 in `geFormRenderer` but I haven't looked to see what adverse effects that may have if any.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -184,6 +184,10 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
             // When the form is being used for Batch Gift Entry, the Form Template JSON
             // uses the @wire service below to retrieve the Template using the Template Id
             // stored on the Batch.
+
+            // get the form template if in batch mode so the 
+            // donationFieldTemplateLabel gets set
+            GeFormService.getFormTemplate();
```
ptewson-sfdo, line 190: @psadie-sfdo @kenlewis-sfdo @bdvorachek @dfuller114 @voduyemi - I'm reviewing Vlad's work here and just curious about this call to GeFormService.getFormTemplate() - when there's a second call to that (with an added big ole' promise resolution thrown in) about four lines below this. Would it be possible to unify these two calls and instead dynamically assign the function that's called in the ".then()" call? And if so ... should we? 

I realize that the real purpose of geFormRenderer has drifted somewhat, but it seems like you would need to access the form template to render the form no matter what the context, and if there's different actions to take AFTER you've loaded the template it would be cleaner to have one call with different behaviors in the "then()" block depending on batch / not batch. But team consensus DEFINITELY outweighs my opinion here!

vmartinov-sforg: @ptewson-sfdo I refactored so now there's only one call to GeFormService.getFormTemplate().

PR #5555, opened 2020-06-10
Show record names instead of IDs for lookup fields in the batch table

dfuller114:
Any name or reference fields contained in the tables under the "Templates" or "Batches" tabs of the gift entry landing page now display as hyperlinks that the user can click to navigate to the relevant record detail page.

# Critical Changes

# Changes
## Gift Entry Updates
- When users review records in a table, all lookup fields will be the linked name of the record instead of the Salesforce ID.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


dfuller114:
**lurch: attach W-039798

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HsrSQAS" target="_blank">W-039798</a>

src/lwc/geListView/geListView.js
```diff
@@ -374,64 +383,96 @@ export default class geListView extends LightningElement {
...
+
     /*******************************************************************************
     * @description Method checks to see if the provided field is a reference and
     * adjusts the relevant column entry properties as needed.
     *
     * @param {object} fieldDescribe: Field describe from the schema.
     * @param {object} columnEntry: A column header entry for lightning-datatable.
     */
-    handleReferenceTypeFields(fieldDescribe, columnEntry) {
+    handleReferenceTypeFields(fieldDescribe) {
```
psadie-sfdo, line 470: I would probably rename this at this point. We're shifting towards using the 'handleX' prefix for methods that receive some kind of an event. Maybe something like `getComputedReferenceFieldTypeApiName`.

src/lwc/geListView/geListView.js
```diff
@@ -374,64 +383,96 @@ export default class geListView extends LightningElement {
...
     }
 
+
+
     /*******************************************************************************
     * @description Method checks to see if the provided field is a reference and
     * adjusts the relevant column entry properties as needed.
     *
     * @param {object} fieldDescribe: Field describe from the schema.
     * @param {object} columnEntry: A column header entry for lightning-datatable.
```
psadie-sfdo, line 468:
```suggestion
```

src/lwc/geListView/geListView.js
```diff
@@ -483,16 +524,16 @@ export default class geListView extends LightningElement {
...
-        if (fieldApiName === NAME) {
-            columnEntry.type = URL;
-            columnEntry.fieldName = URL;
+        if (columnEntry.type === URL) {
+            columnEntry.fieldName = fieldApiName + '_' + URL;
             columnEntry.typeAttributes = {
                 label: {
                     fieldName: fieldApiName
                 },
                 target: _SELF
```
psadie-sfdo, line 533: What do you think about adding a public property 'target' and applying it here if provided otherwise fallback to `_SELF`?

src/lwc/geListView/geListView.js
```diff
@@ -559,39 +600,54 @@ export default class geListView extends LightningElement {
...
-        let recordUrl = this.getRecordUrl();
 
         let records = deepClone(dataRecords);
         records.forEach(record => {
             Object.keys(record).forEach(key => {
-                if (record[key].Name) {
+                let fieldDescribe = this.objectInfo.fields[key];
+
+                if (isNotEmpty(fieldDescribe)
+                    && (fieldDescribe.nameField ||
```
psadie-sfdo, line 614: What do you think about turning the second condition here into a constant and referencing that? Maybe something like `const isReferenceField = fieldDescribe.nameField || fieldDescribe.reference;`

src/lwc/geListView/geListView.js
```diff
@@ -559,39 +600,54 @@ export default class geListView extends LightningElement {
     * 'LastModifiedDate' need to get parsed/reassigned so we can display more user
     * friendly values i.e. Name rather than RecordId or a formatted date.
     *
-    * @param {list} dataRecords: List of sObject records
+    * @param {array} dataRecords: List of sObject records
     */
-    setDatatableRecordsForImperativeCall(dataRecords) {
+     setDatatableRecordsForImperativeCall(dataRecords) {
```
psadie-sfdo, line 605:
```suggestion
    setDatatableRecordsForImperativeCall(dataRecords) {
```

src/lwc/geListView/geListView.js
dfuller114: done!

src/lwc/geListView/geListView.js
dfuller114: done!

src/lwc/geListView/geListView.js
dfuller114: yeah that sounds good to me. done!

src/lwc/geListView/geListView.js
dfuller114: yeah sounds good. might help make it a little more clear what we're checking for there. done!

src/lwc/geListView/geListView.js
dfuller114: done!

src/lwc/geListView/geListView.js
```diff
@@ -48,6 +48,12 @@ const EVENT_TOGGLE_MODAL = 'togglemodal';
 
 export default class geListView extends LightningElement {
 
+    constructor(objectApiName, userId) {
```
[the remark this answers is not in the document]
dfuller114: Yeah it can. Done!

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -1,25 +1,29 @@
 import {LightningElement, api, track} from 'lwc';
```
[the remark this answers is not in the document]
dfuller114: Done!

src/lwc/geListView/geListView.js
```diff
@@ -374,64 +383,96 @@ export default class geListView extends LightningElement {
...
+     *
+     * @param {array} fieldApiNames: List of field api names to check if reference conversation is needed
+     * @return array
+     */
+    buildNameFieldColumns(fieldApiNames) {
         this.columnEntriesByName = {};
-        this.columns = [];
+        let _columns = [];
+
+        for (let i = 0; i < fieldApiNames.length; i++) {
```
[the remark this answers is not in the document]
dfuller114: @vmartinov-sforg We wouldn't be able to use `const` in this case since we're setting the `fieldApiName` variable further down on line 419, but I can use the name syntax you recommended and just use the `let` assignment instead of `const`. Just pushed that up.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -60,10 +64,22 @@ export default class GeBatchGiftEntryTable extends LightningElement {
...
+    }
+
     connectedCallback() {
-        if (this.batchId) {
-            this.loadBatch();
-        }
+        GeFormService.getFormTemplate()
+            .then( response => {
+                    this.fieldMappings = response.fieldMappingSetWrapper.fieldMappingByDevName;
+
```
kenlewis-sfdo, line 77: Is this new `fieldMappings` field being used in the component?  I might have missed it.  I've been trying to avoid retrieving the field mappings in the table component since they are already retrieved by the renderer (which is a sibling to this component in batch mode) and the retrieval is kind of resource expensive.  If we need the field mappings in this component do you think we should import them at the top-level and pass the data structure to child components?

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -150,12 +158,14 @@ export default class GeBatchGiftEntryTable extends LightningElement {
...
+        let rows = this.setDatatableRecordsForImperativeCall([dataRow]);
+
         if (existingRowIndex !== -1) {
             this.data.splice(existingRowIndex, 1, dataRow);
             this.data = [...this.data];
         } else {
-            this.data = [dataRow, ...this.data];
-            if (this.hasData == false) {
+            this.data = [...rows, ...this.data];
+            if (this.hasData === false) {
```
kenlewis-sfdo, line 168: `hasData` is undefined when the component is initialized.  With this change, should the field be updated to have a default value of `false`?

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -150,12 +158,14 @@ export default class GeBatchGiftEntryTable extends LightningElement {
...
 
+        let rows = this.setDatatableRecordsForImperativeCall([dataRow]);
+
         if (existingRowIndex !== -1) {
             this.data.splice(existingRowIndex, 1, dataRow);
             this.data = [...this.data];
         } else {
-            this.data = [dataRow, ...this.data];
-            if (this.hasData == false) {
+            this.data = [...rows, ...this.data];
```
kenlewis-sfdo, line 167: If we're going to change this, can we use `Array.concat()` on the two arrays instead of the spread operator? I initially did this with the spread operator but concat is more performant when you don't actually need the spread, which I don't think is needed here.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
dfuller114: @kenlewis-sfdo good catch! We actually don't need to do this anymore. I changed the approach and forgot to remove it from the connectedCallback. Thanks!

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
dfuller114: done!

kenlewis-sfdo: 👍🏽

src/lwc/geListView/geListView.js
dfuller114: @psadie-sfdo while I was testing through this I found that I think we'll need to keep the for loop syntax the way it was since `fieldDescribe` can potentially be undefined if the field isn't part of the `objectInfo` property, and we have several of those extra fields which are in the object keys.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
dfuller114: @kenlewis-sfdo if we use 1concat1 on the rows for the table it'll add each new item at the end of the array and not the beginning, which would put the new items at the bottom of the table. Not sure if we want new rows to continue displaying at the top of the table, but I think changing to `concat` would have that side effect of adding at the end.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -101,7 +105,8 @@ export default class GeBatchGiftEntryTable extends LightningElement {
 
     @api
     handleSectionsRetrieved(sections) {
-        this.initColumns(this.buildColumns(sections));
+        let columns = this.buildColumns((sections));
```
vmartinov-sforg, line 108: Really minor - 
```suggestion
        const columns = this.buildColumns((sections));
```

dfuller114: @vmartinov-sforg with the resolution of the larger merge conflict, this should be resolved

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -1,84 +1,94 @@
...
-    @api
-    handleSectionsRetrieved(sections) {
-        if (!this._batchLoaded) {
-            this.loadBatch(sections);
-        }
+    constructor() {
+        super(DATAIMPORT_INFO.objectApiName);
+        /* Add the loadBatch function as a callback for the parent component to execute once it executes the
+        objectInfo wire service */
+        this.callbacks.push(this.loadBatch.bind(this));
```
bdvorachek, line 71:
```suggestion
        this.callbacks.push(this.loadBatch);
```

I don't believe `.bind` is necessary here since we're working with an arrow function, the component still works without it.

dfuller114: good call. done!

kenlewis-sfdo: It looks like this update changes the context of the call to the loadBatch method. Since it is now passed to GeListView which calls it asynchronously, it now runs anonymously - so the `this` context does need to be set explicitly to work with these changes.  

It doesn't look like we need a copy of the method, so I would try to use `.call()` rather than `.bind()`.  Even with that change, it looks like there might be some further changes needed to have this work for batch mode. For instance, `loadBatch` passes the `batchId` to the `getDataImportModel` method, but `this.batchId` won't be initialized in the constructor, so this push of the `loadBatch` method into the callbacks array in the GeListView method might need to be moved to the connectedCallback in this component.

dfuller114: @kenlewis-sfdo thanks for mentioning this. i'm pretty sure we should be okay since `loadBatch` is using an arrow expression for the function and arrow expressions use the this value from the enclosing scope, which in this case is the BGE table component class. Attached a screenshot from the page load that shows the value of `this` as the BGE table and we're able to get the batchId from that even though the function executes anonymously. 

Screen Shot 2020-06-25 at 10 48 22 AM

PR #5569, opened 2020-06-16
Provide an accessible label for form fields

npsp-reedestockton:
# Critical Changes

# Changes
 - We corrected accessibility issues for several fields on the Contact Merge, Household Management, and Levels pages.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
@CrystalPreston-Watson I would really appreciate your comments on these changes.
@force2b @andrewyu-salesforce Ready for preliminary review, but significant changes may be suggested by @CrystalPreston-Watson.
@sam-knox Would you please review the Changes section of the PR?  No label changes.

npsp-reedestockton:
**lurch: attach W-041648

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hajyQAA" target="_blank">W-041648</a>

src/pages/CON_ContactMerge.page
```diff
@@ -363,7 +364,7 @@
                                             </th>
                                             <apex:repeat var="col" value="{!row.values}">
                                                 <th class="slds-form-element__control slds-size_1-of-3 slds-cell-wrap">
-                                                    <label class="slds-radio">
+                                                    <label class="slds-radio" aria-label="{!ariaNameMap[col.objId]} {!row.fieldLabel}">
```
force2b, line 367: @CrystalPreston-Watson and @npsp-reedestockton  I'm curious how this actually reads out in VoiceOver. The generated aria-label for these is something like "Michael Smith email", assuming that one of the Contacts being merged is named "Michael Smith" and the field row is for "email". There is also a rendered value in the UI that would likely read out the email address if there is one. What I'm curious about is how this reads out if merging two Contacts with the exact same name. In the UI, I can see column 1 and column 2 and know these are two contacts with the same name being merged. If the reader says the same value for the radio button in both columns is that helpful? It might be. Knowing how to understand the page while listening to the reader is a skill in itself. Maybe it's fine. Maybe it should be something like "Contact 1 Email" and "Contact 2 Email" so that it's clear which column is being read. I don't have enough background on using a screen reader to know the right answer here.

npsp-reedestockton: When a sighted user looks at this page, the top heading for each row is the contact name.  If the contact name is the same for all contacts, the heading is the same for all contacts.  That's what I used to model the aria label.  

I don't like the idea of losing the name, but if we wanted to add info that would be fine.  Maybe something like Contact 1, then name, then field name?  I'm pretty solidly against losing the context of the user name since that would definitely mean less context.

npsp-reedestockton: @force2b I'm rethinking the Contact 1 suggestion.  Anyone using JAWS or VoiceOver to navigate this page will encounter the same thing that someone using visual navigation does - that the contacts being merged have the same name.  In order to get to this page, the user first has to go through the search/selection process, and they will have full knowledge by this point that the contacts have the same name.  We need to make sure that screen reader users know what the data represents, and that is done by providing the info from the column header and row label.

PR #5577, opened 2020-06-17
Hack to delete feature params for deploying to persistent orgs.

prescod:
This new flow makes it easy to delete feature parameters from a local repo so that you do not hit the 25 feature param limit in (some?) persistent orgs.


cumulusci.yml
```diff
@@ -714,6 +714,22 @@ tasks:
...
 
+    remove_feature_parameters_from_XML:
+        class_path: cumulusci.tasks.metadata.modify.RemoveElementsXPath
+        options:
+            output_style: salesforce
+            elements:
+                - path: src/package.xml
+                  xpath: /ns:Package/ns:types/ns:members
+                            [../ns:name='FeatureParameterBoolean' or ../ns:name='FeatureParameterInteger']
+                            [.='PilotEnabled' or
```
force2b, line 725: I'm pretty sure these two won't help with the LDV-XL org. The best one to delete is `Data_CountGERowsLast30DaysBatch`. That leaves exactly 25 feature parameters. Any org that already has those 25 (such as LDV-XL) will require that the most recent ones (`Data_CountGERowsLast30DaysBatch`) be deleted. Once it exists in the, removing from the metadata does not remove it from the org during deployment. When new FP's are added, then we can add those to this task.

cumulusci.yml
```diff
@@ -714,6 +714,22 @@ tasks:
...
+                            .='Org_LongestRunningBatchLast7Days']
+
+    delete_feature_parameter_files:
+        class_path: cumulusci.tasks.util.Delete
+        options:
+            path: ["*/*/Org_LongestRunningBatchLast7Days.featureParameterInteger",
```
force2b, line 731: Same change needed here.

cumulusci.yml
```diff
@@ -1161,6 +1175,14 @@ flows:
                     path: scripts/New_Testing_User.cls
                     apex: createTestingUser();
 
+    hack_to_delete_feature_parameters:
```
force2b, line 1178: I'm curious, what are the previous 4 steps for this? Wouldn't there just be the two new tasks?

prescod: @force2b : I'm not sure if you remember but in the early days of BASIC, it used line numbering. And the way that you made it easy to insert new lines later was to leave gaps between lines. That's what I was thinking. Perhaps it is more confusing than helpful in the context of CumulusCI.

prescod: I'm heading off to PTO so you can either leave them, change them to 1, 2 or wait for me to get back in a week.

force2b: Thanks for clarifying.

force2b:
@prescod What do you think we should do with this PR?

prescod:
I really created it to make things a bit easier for the NPSP team (including broader team like CCE etc.) We could clean it up and merge it or turn it into a Quip doc. You can consider it a POC and make the decision about whether to operationalize it or abandon it...

force2b:
@prescod If you're ok with my updated flow, let's merge this into master so it's available on any branch.

PR #5598, opened 2020-06-20
Use PUSH security for regression

davidmreed:
This PR applies PUSH security to the install of the managed beta in `regression_org`, meaning that new fields do _not_ have FLS granted to all users.

*Note* that this does not completely solve the FLS management issue, as `update_admin_profile` will still run later and grant FLS to fields on selected standard objects. This is a first step.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


davidmreed:
I'm withdrawing this PR in favor of a more sophisticated reorganization of `regression_org` to suit this need.

PR #5599, opened 2020-06-22
RD2 UI: Advanced Scheduling

force2b:
# Critical Changes

# Changes

- In the new Recurring Donations custom UI, only display the "Number of Planned Installments" field for entry when the Recurring Type is "Fixed"
- In the new Recurring Donations custom UI, add support for "Advanced" schedule entry to allow for input of the Installment Frequency ("Every") and non-Monthly type schedules.

# Issues Closed

# Community Ideas Delivered

# New Metadata

Custom Labels:
- RD2_EntryFormPeriodAdvanced
- RD2_EntryFormPeriodHelpText
- RD2_EntryFormPeriodLabel
- RD2_EntryFormPeriodPluralDaily
- RD2_EntryFormPeriodPluralMonthly
- RD2_EntryFormPeriodPluralWeekly
- RD2_EntryFormPeriodPluralYearly
- RD2_EntryFormScheduleEveryLabel
- RD2_EntryFormMissingPermissions

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_EntryFormController.cls
+++ b/src/classes/RD2_EntryFormController.cls
@@ -60,7 +60,7 @@ public with sharing class RD2_EntryFormController {
     * @param recordId RecurringDonation.RecordId
     * @return Map<String, Object> of specific type of RD data
     */
-    @AuraEnabled(Cacheable=true)
+    @AuraEnabled
     public static Map<String, Object> getRecurringData(String recordId) {
         if (recordId == null) {
             return null;
--- a/src/classes/RD2_EntryFormController.cls
+++ b/src/classes/RD2_EntryFormController.cls
@@ -37,8 +37,8 @@
 public with sharing class RD2_EntryFormController {
 
     /**
-    * @description  Gather the org setting for Recurring Donation Entry Form
-    * @param parentId The parentId to use when creating the new RD record (if New from Acount or Contact)
+    * @description  Gather the org, object and environment setting for Recurring Donation Entry/Edit Form
+    * @param parentId The parentId to use when creating the new RD record (if New from Account or Contact)
     * @return Map<String, Object>
     */
     @AuraEnabled(Cacheable=true)
@@ -47,27 +47,12 @@ public with sharing class RD2_EntryFormController {
             ? null
             : parentId.getSobjectType().getDescribe().getName();
 
-        DescribeFieldResult dfrPeriod = UTIL_Describe.getFieldDescribe(
-            'npe03__Recurring_Donation__c', 'npe03__Installment_Period__c'
-        );
-        DescribeFieldResult dfrFrequency = UTIL_Describe.getFieldDescribe(
-            'npe03__Recurring_Donation__c', UTIL_Namespace.StrTokenNSPrefix('InstallmentFrequency__c')
-        );
-
         return new Map<String, Object> {
             'isAutoNamingEnabled' => RD2_NamingService.isAutomaticNamingEnabled,
             'isMultiCurrencyEnabled' => UserInfo.isMultiCurrencyOrganization(),
             'parentSObjectType' => parentSObjectType,
-            'InstallmentPeriodPermissions' => new Map<String, Boolean> {
-                    'Createable' => dfrPeriod.isCreateable(),
-                    'Updateable' => dfrPeriod.isUpdateable(),
-                    'Visible' => dfrPeriod.isAccessible()
-                },
-            'InstallmentFrequencyPermissions' => new Map<String, Boolean> {
-                    'Createable' => dfrFrequency.isCreateable(),
-                    'Updateable' => dfrFrequency.isUpdateable(),
-                    'Visible' => dfrFrequency.isAccessible()
-                }
+            'InstallmentPeriodPermissions' => getFieldPermissionsMap('npe03__Installment_Period__c'),
+            'InstallmentFrequencyPermissions' => getFieldPermissionsMap(UTIL_Namespace.StrTokenNSPrefix('InstallmentFrequency__c'))
         };
     }
 
@@ -170,5 +155,19 @@ public with sharing class RD2_EntryFormController {
             return hhRecordTypeId;
         }
         private set;
+    }
+
+    /**
+     * @description Builld a map of field permissions to return the the LWC page to assign with rendering custom picklist fields
+     * @param fieldApiName Field ApiName
+     * @return Map<String, Object> of field permissions
+     */
+    private static Map<String, Boolean> getFieldPermissionsMap(String fieldApiName) {
+        DescribeFieldResult dfr = UTIL_Describe.getFieldDescribe('npe03__Recurring_Donation__c', fieldApiName);
+        return new Map<String, Boolean>{
+            'Createable' => dfr.isCreateable(),
+            'Updateable' => dfr.isUpdateable(),
+            'Visible' => dfr.isAccessible()
+        };
     }
 }
--- a/src/classes/RD2_EntryFormController.cls
+++ b/src/classes/RD2_EntryFormController.cls
@@ -158,7 +158,9 @@ public with sharing class RD2_EntryFormController {
     }
 
     /**
-     * @description Builld a map of field permissions to return the the LWC page to assign with rendering custom picklist fields
+     * @description Build a map of field permissions for use by LWC so it can validate access permission
+     * for custom picklist fields without a return trip to server. Fields using lightning-input-field do not
+     * require this additional check.
      * @param fieldApiName Field ApiName
      * @return Map<String, Object> of field permissions
      */
--- a/src/classes/RD2_EntryFormController.cls
+++ b/src/classes/RD2_EntryFormController.cls
@@ -38,11 +38,11 @@ public with sharing class RD2_EntryFormController {
 
     /**
     * @description  Gather the org setting for Recurring Donation Entry Form
-    * @param parentId The parentId of the new RD
+    * @param parentId The parentId to use when creating the new RD record (if New from Acount or Contact)
     * @return Map<String, Object>
     */
     @AuraEnabled(Cacheable=true)
-    public static Map<String, Object> getSetting(Id parentId) {
+    public static Map<String, Object> getRecurringSettings(Id parentId) {
         String parentSObjectType = (parentId == null)
             ? null
             : parentId.getSobjectType().getDescribe().getName();
@@ -58,25 +58,34 @@ public with sharing class RD2_EntryFormController {
     /**
     * @description Retrieve the donor type for an existing RD record
     * @param recordId RecurringDonation.RecordId
-    * @return Account (if Org is not a Household) or Contact
+    * @return Map<String, Object> of specific type of RD data
     */
     @AuraEnabled(Cacheable=true)
-    public static String getDonorType(String recordId) {
+    public static Map<String, Object> getRecurringData(String recordId) {
         if (recordId == null) {
             return null;
         }
 
+        npe03__Recurring_Donation__c rd;
+        Map<String, Object> response = new Map<String, Object>();
         try {
-            npe03__Recurring_Donation__c rd = getRecurringDonation(recordId);
-            if (rd.npe03__Organization__r.RecordTypeId == hhRecordTypeId
-                || (rd.npe03__Organization__c == null && rd.npe03__Contact__r.Account.RecordTypeId == hhRecordTypeId)
-            ) {
-                return 'Contact';
-            }
-            return 'Account';
+            rd = getRecurringDonation(recordId);
         } catch (Exception ex) {
             throw new AuraHandledException(ex.getMessage());
         }
+
+        if (rd.npe03__Organization__r.RecordTypeId == hhRecordTypeId
+            || (rd.npe03__Organization__c == null && rd.npe03__Contact__r.Account.RecordTypeId == hhRecordTypeId)
+        ) {
+            response.put('DonorType', 'Contact');
+        } else {
+            response.put('DonorType', 'Account');
+        }
+
+        response.put('Period', rd.npe03__Installment_Period__c);
+        response.put('Frequency', rd.InstallmentFrequency__c);
+
+        return response;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


force2b:
**lurch: attach W-035044
**lurch: attach W-039993

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C8w4QAC" target="_blank">W-035044</a>

src/labels/CustomLabels.labels
```diff
@@ -2015,6 +2015,70 @@
...
+        <shortDescription>Picklist value for the RD2 Entry Form custom period picklist</shortDescription>
+        <value>Advanced</value>
+    </labels>
+    <labels>
+        <fullName>RD2_EntryFormPeriodHelpText</fullName>
+        <categories>Recurring-Donations, Entry-Form</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Inline Help Text for the Monthly and Advanced Picklist</shortDescription>
+        <value>Select &quot;Advanced&quot; to enter other schedule types such as weekly, yearly, or &quot;every 3 months&quot;.</value>
```
force2b, line 2032: @sam-knox Would you mind reviewing the custom help text for this picklist? This is the help text for the field with a label of "Recurring Period" (the label below this).

sam-knox: @force2b - Sure thing. Use this for the help text:

"Select Advanced to enter other schedule types such as weekly, yearly, every 6 months, etc."

force2b: Thanks @sam-knox. Change has been made.

src/labels/CustomLabels.labels
```diff
@@ -2015,6 +2015,86 @@
         <shortDescription>Recurring Donation Entry Form message when record is inserted</shortDescription>
         <value>Recurring Donation &quot;{0}&quot; was created.</value>
     </labels>
+    <labels>
+        <fullName>RD2_EntryFormMissingPermissions</fullName>
+        <categories>Recurring-Donations, Entry-Form</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Recurring Donation Entry Form Field Permissions Error</shortDescription>
+        <value>You are missing read and/or edit access to one or more required Recurring Donation fields. Please check with your system administrator.</value>
```
force2b, line 2024: @sam-knox Sorry. One more new label to review. This message is displayed if the User is missing Read permissions to any of the required field in the custom UI - Account, Contact, Amount, Period, Frequency, etc. There are far too many to figure out which one(s) are missing and dynamically include in the error, so I kept it generic.

sam-knox: @force2b - That's ok, I don't think we need to list all the fields in the error message. Here's a slightly revised version:

"You don&apos;t have the required Recurring Donations field permissions. Contact your system administrator for assistance."

We may need to document what the required fields are though. And the minimum is Read permission - so if a user only has that permission, I assume they would be able to load the custom UI but not make any edits?

force2b: @sam-knox Thanks. Label change made. That error is only displayed when the User has no permissions to fields that are built into the UI (see list below). There is no error if they have read permissions to the field, but not edit. In that case, the field simply appears in the UI as not editable.

Fields requiring Read access:
- Account
- Contact
- Recurring Type
- Amount
- Installment Period
- Installment Frequency
- Day of Month
- Effective Date
- Payment Method
- Campaign

force2b: Typing up that list made me think about these fields more. It's impossible that some customers may want to exclude Campaign or Payment Method as field by removing access to them to end users. So, maybe there's alternative route to handle permissions. Rather than throw an exception/error if any of these fields is not accessible, I could have the page simply ignore the error and not render the fields that are not accessible. Of course, a user without read access to many of these fields will be unable to create a new RD and possibly Edit an RD, but then they'll just get a validation error for a missing field and those are already pretty well defined. For example, if they do not have Read access to DayOfMonth, they'd get an error trying to create an RD with an Installment Period of Monthly stating that a DayOfMonth value is required for a Monthly RD.

What do you think?

@npsp-reedestockton @andrewyu-salesforce

andrewyu-salesforce: I think it makes sense to not show the field if the user does not have Edit access to the field. If the user does not have the proper permission to create an RD, then it's up to the Admin to give them the right access. With this if the field is pre-populate and the user wants to update the record, the entry form will still work.

src/classes/RD2_EntryFormController.cls
```diff
@@ -38,45 +38,100 @@ public with sharing class RD2_EntryFormController {
...
+            'InstallmentPeriodPermissions' => new Map<String, Boolean> {
+                    'Createable' => dfrPeriod.isCreateable(),
+                    'Updateable' => dfrPeriod.isUpdateable(),
+                    'Visible' => dfrPeriod.isAccessible()
+                },
+            'InstallmentFrequencyPermissions' => new Map<String, Boolean> {
+                    'Createable' => dfrFrequency.isCreateable(),
+                    'Updateable' => dfrFrequency.isUpdateable(),
+                    'Visible' => dfrFrequency.isAccessible()
+                }
```
andrewyu-salesforce, line 70: Extract the Createable, Updateable, and Visible into a helper function that returns Map<String, Boolean>()
public static Map<String, Boolean> getFieldSecurity(String ApiName)

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -212,16 +234,36 @@ export default class rd2EntryForm extends LightningElement {
 
     /***
     * @description Handle component display when an error occurs
+    * @param error: Error Event
+    * @param disableSaveButton: true to disable the Save button on the page; defaults to false
     */
-    handleError(error) {
-        this.errorMessage = constructErrorMessage(error);
+    handleError(error, disableSaveButton) {
+        this.errorMessage = (!error || (!error.detail && !error.header)) ? constructErrorMessage(error) : error;
```
andrewyu-salesforce, line 241: constructErrorMessage(error) returns an Object with Error and Detail instead of the error string. It needs to go one extra step to get the detail from the return object. 
Screen Shot 2020-06-26 at 2 36 56 PM

src/lwc/rd2EntryForm/rd2EntryForm.html
```diff
@@ -18,106 +18,119 @@
...
+                            <lightning-layout horizontal-align="spread" multiple-rows="true" if:true={isFormReady}>
+                                <lightning-layout-item padding="horizontal-small" size="6" if:true={fields.status}>
+                                    <lightning-input-field field-name={fields.status.apiName} required>
                                     </lightning-input-field>
                                 </lightning-layout-item>
 
-                                <lightning-layout-item padding="horizontal-small" size="12" small-device-size="6">
-                                    <lightning-input-field field-name={fields.paymentMethod.apiName}>
+                                <lightning-layout-item padding="horizontal-small" size="6" if:true={fields.statusreason}>
+                                    <lightning-input-field field-name={fields.statusreason.apiName}>
```
andrewyu-salesforce, line 61:
```suggestion
                                    <lightning-input-field field-name={fields.statusReason.apiName}>
```

src/lwc/rd2EntryForm/rd2EntryForm.html
```diff
@@ -18,106 +18,119 @@
...
+                    <h3 class="slds-section__title slds-theme_shade">
+                        <span class="slds-truncate slds-p-horizontal_small">{customLabels.scheduleSectionHeader}</span>
+                    </h3>
+                    <div aria-hidden="false" class="slds-section__content">
+
+                        <lightning-layout horizontal-align="spread" multiple-rows="true" if:true={isFormReady}>
+                            <lightning-layout-item padding="horizontal-small" size="6" small-device-size="6" if:true={fields.amount}>
+                                <lightning-input-field field-name={fields.amount.apiName} required>
+                                </lightning-input-field>
+                            </lightning-layout-item>
```
andrewyu-salesforce, line 80: The amount input field will be null out when there is a validation error on saving the record.

src/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
```diff
@@ -144,7 +287,133 @@ export default class rd2EntryFormScheduleSection extends LightningElement {
...
+                            {label: pl.label, value: pl.value}
+                        );
+                        break;
+                }
+
+            });
+        return advancedPeriodPicklistValues;
     }
 
     /**
```
andrewyu-salesforce, line 419: isValid() should also report the validity on lightning-combobox to make sure Installment Period picklist is valid on save.
`this.template.querySelectorAll('lightning-combobox').forEach(.....)`

src/classes/RD2_EntryFormController.cls
force2b: Great suggestion

src/lwc/rd2EntryForm/rd2EntryForm.html
force2b: Good catch. Not quite sure how to handle this one though.

force2b:
Thanks for the code review @andrewyu-salesforce. All changes, with the exception of the Amount field issue, have been made. I'm not quite sure about that one. I can probably retrieve and save the value as a @tracked var and use that on the lightning-input-field, but it seems like that should really be necessary. Out of curiosity, what were you using to force a save validation error? My tests used a negative amount value, so I was wondering if that was why the value was being cleared out.

src/lwc/rd2EntryForm/rd2EntryForm.html
```diff
@@ -18,106 +18,119 @@
...
                         <h3 class="slds-section__title slds-theme_shade">
-                            <span class="slds-truncate slds-p-horizontal_small">{customLabels.scheduleSectionHeader}</span>
+                            <span class="slds-truncate slds-p-horizontal_small">{customLabels.statusSectionHeader}</span>
                         </h3>
                         <div aria-hidden="false" class="slds-section__content">
 
-                            <lightning-layout horizontal-align="spread" multiple-rows="true">
-                                <lightning-layout-item padding="horizontal-small" size="12" small-device-size="6">
-                                    <lightning-input-field field-name={fields.amount.apiName} required>
+                            <lightning-layout horizontal-align="spread" multiple-rows="true" if:true={isFormReady}>
```
andrewyu-salesforce, line 54: @force2b I forced the error on save by using a negative value in the 'Every' field. I notice the status and status reason is also clear out It seem like the error is caused by lightning-input-field being re-rendered on save. If you remove the if:true={isFormReady} on here and line 76, the field will not be re-rendered

force2b: Thanks and great suggestion. Worked perfectly.

src/classes/RD2_EntryFormController.cls
```diff
@@ -114,4 +156,18 @@ public with sharing class RD2_EntryFormController {
         }
         private set;
     }
+
+    /**
+     * @description Builld a map of field permissions to return the the LWC page to assign with rendering custom picklist fields
```
npsp-reedestockton, line 161: This is confusing to me.  Would it be correct to reword as "Builld a map of field permissions for use by LWC so it can validate access permission without a return trip to server."

src/lwc/rd2EntryFormDonorSection/rd2EntryFormDonorSection.js
```diff
@@ -44,34 +45,31 @@ export default class rd2EntryFormDonorSection extends LightningElement {
     contactLabel;
 
     /**
-     * @description Get settings required to enable or disable fields and populate their values
+     * @description If editing an existing record retrieve the Donor Type from the record so it can defaul the custom
```
npsp-reedestockton, line 48: default

src/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
```diff
@@ -78,56 +170,107 @@ export default class rd2EntryFormScheduleSection extends LightningElement {
...
+        if (!this.isNew || !this.defaultInstallmentPeriodValue) {
+            return;
+        }
+        if (this.defaultInstallmentPeriodValue !== PERIOD_MONTHLY) {
+            this.updateScheduleFieldVisibility(RECURRING_PERIOD_ADVANCED, this.defaultInstallmentPeriodValue);
+        }
+    }
+
+    /**
+     * @description Returns a boolean to disable entry a field based on the FLS
```
npsp-reedestockton, line 226: entry into a field

src/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
```diff
@@ -78,56 +170,107 @@ export default class rd2EntryFormScheduleSection extends LightningElement {
...
+     * @returns True to disable entry into the field
+     */
+    shouldDisableField(fieldPerms) {
+        return (this.isNew ? !fieldPerms.Createable : !fieldPerms.Updateable);
+    }
+
+    /**
+     * @description Returns a boolean to hide entry a field based on the FLS
```
npsp-reedestockton, line 234: hide a field

force2b:
@npsp-reedestockton Thanks for the review. All changes made

src/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.html
```diff
@@ -7,38 +7,86 @@
...
-                        </lightning-input-field>
+                    <lightning-layout-item padding="horizontal-small" size="12" large-device-size={scheduleRowColumnSize}
+                                           if:false={hidePeriodPicklistField}>
+                        <template if:true={isAdvancedMode}>
+                            <lightning-combobox
+                                name="advancedPeriodSelect"
+                                class="advanced-mode-fields rd2-entry-custom-field-padding"
+                                label={fields.period.label}
+                                value={customPeriodAdvancedMode}
+                                placeholder=""
```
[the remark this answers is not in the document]
force2b: Looks better without the placeholder. Thanks.

PR #5608, opened 2020-06-24
W-041630 [NPSP] PYALLOC - Modify Allocations Trigger for Payment Amou…

jbrandolinisf:
…nt $0 Behavior
- When Payment Allocations are enabled:
- New Payments with Amount $0 will have a $0 Default Allocation
- Changing from a non zero to a zero $ amount will resize percent allocations and block if non default fixed amount allocations are present
- Changing from a non zero to a zero $ amount will remove the default allocation
- Updated API version of AllocationsUtil, AllocationsUtil_TEST class
- Tests added to cover revised Payment Allocation $0 Amount scenario changes:
      - changePaymentAmountFromNonZeroToZeroshouldRemoveDefaultAllocation
      - generateErrorForNonZeroFixedAmountAllocationIfPaymentAmountIsZero
      - changePaymentAmountFromNonZeroToZeroshouldResizePercentAllocations
      - defaultAllocationsForZeroAmountPaymentShouldRemainOnCreate

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


jbrandolinisf:
**lurch: attach W-041630

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007haiWQAQ" target="_blank">W-041630</a>

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -226,18 +225,18 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
...
+                if ((pmt.npe01__Payment_Amount__c != null)) {
                     pmtsNeedingAllocations.add(pmt);
                 }
             }
 
             //UPDATE case: if the payment amount changed, add to the list for processing
             if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-                Decimal oldAmount = valueOrDefaultIfNull((Decimal) oldlist[i].get('npe01__Payment_Amount__c'), 0);
-                Decimal newAmount = valueOrDefaultIfNull(pmt.npe01__Payment_Amount__c, 0);
+                Decimal oldAmount = (Decimal) oldlist[i].get('npe01__Payment_Amount__c');
```
force2b, line 236: I think this will throw an NPE if `npe01__Payment_Amount__c` is null. Why not just use the following this way it doesn't crash if the value is null
`Decimal oldAmount = oldlist[i].npe01__Payment_Amount__c;`

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -1088,6 +1088,228 @@ private with sharing class ALLO_Allocations_TEST {
...
 
+    /**
+    * @description Tests that Allocations are updated and preserved when changing a Payment from a
+    * non zero to a zero amount, and that default allocation is removed
+    */
+    @isTest
+    private static void changePaymentAmountFromNonZeroToZeroshouldResizePercentAllocations() {
+        List<General_Accounting_Unit__c> gaus = new List<General_Accounting_Unit__c>();
+
+        General_Accounting_Unit__c defaultGau = new General_Accounting_Unit__c(Name='General');
```
force2b, line 1099: There is a `UTIL_UnitTestData_TEST.createGAUs()` method that you can use for these. Plus, all four of these new test methods have extremely similar data setup. Might be worth centralizing this is a helper method.

src/classes/ALLO_Allocations_TDTM.cls
jbrandolinisf: Thanks for the feedback.  I am pretty sure the get will just return null if there is no value present in the field map.  I can change to your suggested format since it is easier to read, but I believe I just mirrored what was already done in the Opportunity part of the triggers (I will double check though).

src/classes/ALLO_Allocations_TEST.cls
jbrandolinisf: I will see what I can do with this today/tomorrow (lot of meetings today).

src/classes/ALLO_Allocations_TDTM.cls
force2b: Thanks. It's true that `oldlist[i].npe01__Payment_Amount__c` will return null, however casting the null as a Decimal (`(Decimal)null`) is what will throw the exception. Since the field always exists on the record, it's more efficient to just retrieve it directly and avoid the casting.

jbrandolinisf: I am happy to make the change you are asking me to make, but as a point of order what you are stating is incorrect.  You absolutely CAN cast null to a Decimal type, or virtually any other.  It will not throw an exception.  The value will simply be null. 

Whether you should or not is a different question...

force2b: I stand very corrected. I could have sworn that this used to throw an exception - specifically when an object type is cast to something concrete. It definitely works fine now, so no need to change anything here.

jbrandolinisf: Hehe.  You are awesome!!!  No worries on my end.  I so value your feedback, and really appreciate the chance to talk back and forth when there is something like this!!!!

jbrandolinisf:
@force2b - This is ready for re-review when you have time

src/classes/ALLO_Allocations_TEST.cls
```diff
@@ -37,6 +37,53 @@
 @isTest
 private with sharing class ALLO_Allocations_TEST {
 
+    // These variables are used in a "Setup Results" object map to find different
+    // test data that was created
+
+    /** @description Key to find GAUs with */
+    private static final String GAUS_KEY {
```
force2b, line 44: In a unit test context, is this any different than the following?
`private static final String GAUS_KEY = 'GAUS'`?

src/classes/ALLO_Allocations_TDTM.cls
```diff
@@ -791,13 +795,14 @@ public class ALLO_Allocations_TDTM extends TDTM_Runnable {
 
             if (mapWrapper.containsKey(pmt.Id)) {
                 alloWrapper pmtWrap = mapWrapper.get(pmt.Id);
-                Decimal amt = valueOrDefaultIfNull(pmt.npe01__Payment_Amount__c, 0);
-                //null or zero amount: delete all associated allocations
-                if (amt == 0) {
+                Decimal amt = pmt.npe01__Payment_Amount__c;
+                // null delete all associated allocations
```
force2b, line 799: Seems like a good opportunity to abstract out some behavior. The method itself is a huge and could use cleanup, but for now I think it's safer to refactor what's being modified.
```
            Decimal amt = pmt.npe01__Payment_Amount__c;
            if (amt == null) {
                handleNullAmount(pmtWrap);
                continue;
            } else { 
```
```
/**
* @description If amount is null then delete all associated allocations
*/
private void handleNullAmount(alloWrapper pmtWrap) {
    dmlWrapper.objectsToDelete.addAll((List<SObject>)pmtWrap.listAllo);
    if (settings.Default_Allocations_Enabled__c && pmtWrap.defaultAllo != null) {
        dmlWrapper.objectsToDelete.add(pmtWrap.defaultAllo);
        dmlWrapper.objectsToDelete.addAll((List<SObject>)pmtWrap.defaultDupesById.values());
    }
}

jbrandolinisf: Yeah, I was hoping to eventually replace some of this with the service layer we are using via callable, but I can do that.

src/classes/ALLO_Allocations_TEST.cls
jbrandolinisf: No, not really.  I was being a little over obsessive about lazy initialization for class variables to keep heap allocation low...but it really isn't necessary for these, so I will change them.

src/classes/ALLO_Allocations_TDTM.cls
jbrandolinisf: I don't want it to do the continue statement, but will otherwise do as you suggest.  This is because I still want it to go through the default allocations block.  I agree this whole thing could be simplified, but I would rather refactor parts to share the service logic long term.

jbrandolinisf:
@force2b  - Ok, I incorporated your latest feedback.

PR #5610, opened 2020-06-25
Feature/rd2 elevate integration

vmartinov-sforg:
# Critical Changes

# Changes
Implement two methods in PS_IntegrationServiceConfig.cls:
* isIntegrationEnabled()
* hasIntegrationPermissions()

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


vmartinov-sforg:
**lurch: attach W-042119

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hbcdQAA" target="_blank">W-042119</a>

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -66,12 +92,81 @@ public with sharing class PS_IntegrationServiceConfig {
...
+
+
+
+    /**
+    * @description Inner class responsible for retrieving Payment Services configuration from
+    *              application storage and building and exposing a set of key-value
+    *              configuration pairs.
+    */
+    @TestVisible
+    private class Configuration {
```
force2b, line 109: @vmartinov-sforg Since this class is potentially called by standard end users (via other logic), and the outer class uses `with sharing`, it might be helpful to declare this inner class as `without sharing` to ensure that the queries against the config objects always work.

vmartinov-sforg: @force2b What would be the purpose to call this class for any other purpose? I'm reluctant to use `without sharing` and raise security concerns.

force2b: My concern was that the class gets called from the trigger which is executed when a standard User makes a changes to the RD. If that standard user has no permissions to the `Payment_Services_Configuration__c` object, it is possible the query against that object could fail. Given the usage though, the intent is to determine if the current user is the integration, regardless of whether the current user actually has permissions to the `Payment_Services_Configuration__c` object. My proposal is to use `without sharing` only on the inner class to ensure that the query against `Payment_Services_Configuration__c` always works regardless of what permissions the current user has. It's also feasible to use a separate inner class that has `without sharing` just for the query and have that return the result back to this inner class. That might make it clearer for security review.

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -66,12 +90,86 @@ public with sharing class PS_IntegrationServiceConfig {
     * @return Boolean
     */
     public Boolean hasIntegrationPermissions() {
```
pedrodalcol, line 92: Should this method be called isIntegrationUser? There is only one integration user (the user that last registered Elevate). However, there could be multiple users allowed to integrate with elevate (any user that has the correct perm set/s assigned to it). This method will be used to answer the former not the latter.

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -44,9 +70,7 @@ public with sharing class PS_IntegrationServiceConfig {
     private Boolean isEnabled {
         get {
             if (isEnabled == null) {
-                //Integer configCount = Database.countQuery('SELECT Count() FROM Payment_Services_Configuration__c');
-                Integer configCount = 0;//TODO: hard-coded, disable the integration service until full implementation
-                isEnabled = configCount > 0;
+                return new Configuration().keyValueMap.keySet().containsAll(REQUIRED_CONFIG_KEYS);
```
force2b, line 73: I believe that the Configuration wrapper class is being instantiated twice by the two methods (isEnabled and hasIntegrationPermissions). Each new  configuration instance will query the config object, resulting in an extra query. What do you think of creating a method in the outer class such as the following that both methods can use to retrieve the configuration without incurring the additional query.
```
private static Configuration psConfig {
    get {
        if (psConfig == null) {
            psConfig = new Configuration();
        }
        return psConfig;
    } set;
}
```
}

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -66,12 +90,86 @@ public with sharing class PS_IntegrationServiceConfig {
...
+                }
+                return keyValueMap;
+            }
+            private set;
+        }
+
+        public Payment_Services_Configuration__c lastModifiedRecord {
+            get {
+                if (lastModifiedRecord == null) {
+                    List<Payment_Services_Configuration__c> lastModifiedRecords = getLastModifiedConfigRecord();
```
force2b, line 146: For this method and `getConfigRecordsByName`, Is the intent that every time this method is called it re-queries the object, even in the same execution context? I'm wondering if it makes sense to store the result of the query into a class var and allow that results to be re-used on all subsequent calls into this method/class - the goal being to reduce the number of queries. I can see where that might not be advantageous if the data could change mid-context, but if I understand the usage correctly I think that is not likely.

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -66,12 +90,86 @@ public with sharing class PS_IntegrationServiceConfig {
...
+                    if (makanaKey != null && keyValueMap.get(API_KEY) == null) {
+                        keyValueMap.put(API_KEY, makanaKey);
+                    }
+                }
+                return keyValueMap;
+            }
+            private set;
+        }
+
+        public Payment_Services_Configuration__c lastModifiedRecord {
```
force2b, line 143: Would you mind adding apex doc comment blocks on these inner class methods?

src/classes/PS_IntegrationServiceConfig.cls
vmartinov-sforg: We don't query the config object when we instantiate the Configuration object. The constructor is empty. @force2b

force2b: True. However each individual instance of Configuration has its own versions of the instance methods. So, the following two calls will create two instances of configuration and each will query the `Payment_Services_Configuration__c` object. 
- `new Configuration().keyValueMap.keySet().containsAll(REQUIRED_CONFIG_KEYS);`
- `new Configuration().lastModifiedRecord`
My main concern is the addition of two soql queries per RD record when both methods are called from triggers and batch jobs. My general proposal is to move to a static instance of the query so that there is never more than 1 query against the `Payment_Services_Configuration__c` object even if the isEnabled and hasIntegrationPermission permissions are called multiple times in the same batch operation.

vmartinov-sforg: I understand, I can make changes to achieve these goals, but please keep in mind however that if the configuration records get updated in 'real-time' these methods might not return the correct values, as the results will be 'cached'.

force2b: Only if the records were updated in the same execution context, which probably can't happen. For example, if a batch job is running and processing 2000 Recurring Donation records with a batch size of 200, the result of the query against `Payment_Services_Configuration__c` is cached for the life of that batch child iteration - 200 records only. Any static vars are reset at the start of each individual child iteration (basically at the start of the execute() method). The same would be true for Trigger executions. Effectively, static var life is very short - microseconds to a possible 2 minute max cpu time for a batch iteration.

src/classes/PS_IntegrationServiceConfig.cls
vmartinov-sforg: We are checking `lastModifiedRecord == null`. Wouldn't that prevent the query execution for the second time?

force2b: Excellent point. The getter will only execute the first time.

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -34,6 +34,32 @@
 * @description Determines the Elevate Payment Integration Services status and permissions
 */
 public with sharing class PS_IntegrationServiceConfig {
+    
+    // Names of Payment Service configuration sets
+    @TestVisible static final String PAYMENTS_SERVICE_NAME = 'payments';
+    @TestVisible static final String MAKANA_SERVICE_NAME = 'makana';
+
+     // Properties used to configure headers in
```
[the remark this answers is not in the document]
vmartinov-sforg: Done.

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -61,17 +85,124 @@ public with sharing class PS_IntegrationServiceConfig {
...
+            private set;
+        }
+
+        /* @description Returns the last modified configuration record
+        */
+        public Payment_Services_Configuration__c lastModifiedRecord {
+            get {
+                if (lastModifiedRecords == null) {
+                    lastModifiedRecords = getLastModifiedConfigRecord();
+                    lastModifiedRecord = lastModifiedRecords.isEmpty() ? null : lastModifiedRecords[0];
```
force2b, line 180: @vmartinov-sforg  I think this particular line is the issue from before. If `lastModifiedRecords` is an empty collection, the class var is reset to null which means the getter will requery each time.

vmartinov-sforg: The issue was that I was checking `lastModifiedRecord ` not `lastModifiedRecords`. `lastModifiedRecords` is a class variable, so we are OK - the query gets executed only once whether we have records or not.

src/classes/GE_PaymentServices_TEST.cls
```diff
@@ -52,19 +52,19 @@ public class GE_PaymentServices_TEST {
 
     // Key-Values by service name
     private static final Map<String, List<Map<String, String>>> testConfig =
```
ptewson-sfdo, line 54: I'm wondering if this test config would be better being defined in the PS_IntegrationServiceConfig test class? And maybe some of these test methods (shouldUsePaymentServicesConfig) as well?

vmartinov-sforg: Done.

PR #5611, opened 2020-06-25
W-041881 - [NPSP] Add View with additional fields to Custom Metadata Opportunity Stage to State

jbrandolinisf:
# Critical Changes

# Changes
- Adds "All" Records Listview to Custom Metadata Type Opportunity Stage to State

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/package.xml
+++ b/src/package.xml
@@ -3738,6 +3738,7 @@
         <members>Level__c.All</members>
         <members>Opportunity.To_Be_Acknowledged</members>
         <members>Trigger_Handler__c.All</members>
+        <members>Opportunity_Stage_To_State_Mapping__mdt.All</members>
         <name>ListView</name>
     </types>
     <types>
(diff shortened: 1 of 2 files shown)


jbrandolinisf:
@lparrott - Let me know if you think we should go with something other than All for the list view name.

src/package.xml
```diff
@@ -3738,6 +3738,7 @@
         <members>Level__c.All</members>
         <members>Opportunity.To_Be_Acknowledged</members>
         <members>Trigger_Handler__c.All</members>
+        <members>Opportunity_Stage_To_State_Mapping__mdt.All</members>
```
lparrott, line 3741: It's too bad this doesn't overwrite the default "All" list, but I can't think of a better name 🤷‍♂️  "Stages and States"?

jbrandolinisf: I considered All Stage to State Mappings.  If we like that better, so it does not duplicate it, I can change it to that.

jbrandolinisf: I do think it alpha sorts them though, and goes with the first one if none are selected, so changing it could cause it to not be selected by default.  Right now, the custom All view is selected.

lparrott: Good point, let's just leave it like this then.

PR #5629, opened 2020-06-28
Fix table header and row issues on Contact Merge

npsp-reedestockton:
- We corrected several accessibility issues on the Contact Merge page: improper association of table headers and rows, a missing heading, and an active link improperly located in a table header row.

# Critical Changes

# Changes
- We corrected several accessibility issues on the Contact Merge page.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/pages/CON_ContactMerge.page
+++ b/src/pages/CON_ContactMerge.page
@@ -278,7 +278,9 @@
                                 <table class="slds-table slds-table_bordered slds-max-medium-table_stacked-horizontal">
                                     <thead>
                                         <tr>
-                                            <th></th>
+                                            <th class="slds-text-heading_label" scope="col">
+                                                <apex:outputText value="{!$Label.mtchSelectCbxTitle}"/>
+                                            </th>
                                             <th class="slds-text-heading_label" scope="col">Name</th>
                                             <apex:repeat value="{!$ObjectType.Contact.FieldSets.ContactMergeFoundFS}" var="field">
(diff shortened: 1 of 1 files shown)


npsp-reedestockton:
**lurch:attach W-042097

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hbarQAA" target="_blank">W-042097</a>

npsp-reedestockton:
@force2b @andrewyu-salesforce This is ready for review. Thanks!

@sam-knox "Select" label heading has been added to the search results table on Contact Merge and "Select All" row heading has been added to the Select Contacts comparison table on Contact Merge.  Both are existing labels, but this usage is new.

force2b:
@npsp-reedestockton Overall the functionality works, however the one issue I see is that the two new header rows use a larger font size. As a result they look a bit odd to me.
image

npsp-reedestockton:
Thanks for pointing that out, Mike.  I've been looking at this page for so long it didn't even register that the font size was wrong.

force2b:
Thanks @npsp-reedestockton. The "Select All" line is better, but the Contact name row is still large relative to the other text on the page.
image

npsp-reedestockton:
@force2b Take a quick peek at the production version.  I think you'll find that the contact name line is the same size.  If you still want it to change, okay, but that's not a result of the current code modifications.

PR #5631, opened 2020-06-29
RD: Auto Naming Setting

andrewyu-salesforce:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_Constants.cls
+++ b/src/classes/RD2_Constants.cls
@@ -115,9 +115,4 @@ public inherited sharing class RD2_Constants {
     * @description Contains default value for the Next Donation Date Match Range Days field
     */
     public static final Integer DEFAULT_NEXT_DONATION_DATE_MATCH_RANGE_DAYS = 3;
-
-    public static final String RD_NAMING_DISABLE = 'Disable Auto Naming';
-    public static final String RD_NAMING_OPTION_1 = '{!npe03__Donor_Name__c} {!npe03__Amount__c} - \"Recurring\"';
-    public static final String RD_NAMING_OPTION_2 = '{!npe03__Donor_Name__c} {!npe03__Installment_Period__c} Recurring Donation - {!npe03__Amount__c}';
-    public static final String RD_NAMING_OPTION_3 = '{!npe03__Donor_Name__c} {!npe03__Amount__c} {!npe03__Installment_Period__c}';
 }
--- a/src/classes/RD2_Constants.cls
+++ b/src/classes/RD2_Constants.cls
@@ -116,22 +116,8 @@ public inherited sharing class RD2_Constants {
     */
     public static final Integer DEFAULT_NEXT_DONATION_DATE_MATCH_RANGE_DAYS = 3;
 
-    public static List<String> RDNamingOptions {
-        get {
-            if (RDNamingOptions == null) {
-                RDNamingOptions = new List<String> {
-                    'Disable Auto Naming',
-                    '{!npe03__Donor_Name__c} {!npe03__Amount__c} - "Recurring"',
-                    '{!npe03__Donor_Name__c} {!npe03__Installment_Period__c} Recurring Donation - {!npe03__Amount__c}',
-                    '{!npe03__Donor_Name__c} {!npe03__Amount__c} {!npe03__Installment_Period__c}'
-                };
-            }
-        }
-        set;
-    }
-
-    public static final Stirng RD_NAMING_DISABLE = 'Disable Auto Naming';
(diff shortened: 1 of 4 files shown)


andrewyu-salesforce:
**lurch: attach W-035238

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001C9J1QAK" target="_blank">W-035238</a>

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -120,6 +120,18 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>RecurringDonationNaming__c</fullName>
+        <description>The format to use for Recurring Donation naming.</description>
```
andrewyu-salesforce, line 125: @sam-knox Can you review the description and inline help text for this RD Setting custom field. It is used to store the RD auto naming format.

andrewyu-salesforce:
**lurch: attach W-042396

andrewyu-salesforce:
**lurch: detach W-035238

LurchTheButler:
Detached <a href="https://foundation.my.salesforce.com/a2x1E000001C9J1QAK" target="_blank">W-035238</a>

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000007hc64QAA" target="_blank">W-042396</a>

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -120,6 +120,18 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>RecurringDonationNaming__c</fullName>
```
force2b, line 124: @andrewyu-salesforce Since there is already a checkbox that indicates the "state" of RD naming, I think this field should be something like `RecurringDonationNamingFormat__c`; possibly with a label such as "Recurring Donation Name Format".
cc @sam-knox

src/classes/RD2_Constants.cls
```diff
@@ -116,4 +116,8 @@ public inherited sharing class RD2_Constants {
     */
     public static final Integer DEFAULT_NEXT_DONATION_DATE_MATCH_RANGE_DAYS = 3;
 
+    public static final String RD_NAMING_DISABLE = 'Disable Auto Naming';
```
force2b, line 119: If the default value of this field is null, is that sufficient to indicate that naming is disabled. The actual label in the picklist can come from `stgRDNamingDisable`, but the value to store would be null. If true, then you probably don't need a constant for this.

Also, do you think these constants really belong in the RD2_NamingService class?

andrewyu-salesforce: It makes sense to put the constant into RD2_NamingService. For the disable String, I was debating between empty string or auto populate the field with Disable the first time user save the RD setting in NPSP page if auto naming is off. I think empty strings makes more sense since it will be less conditions to check.

src/objects/npe03__Recurring_Donations_Settings__c.object
andrewyu-salesforce: I think both work. I include 2 screen shots on what it will look at in the UI. @sam-knox 
Screen Shot 2020-06-30 at 12 53 03 PM
Screen Shot 2020-06-30 at 12 52 52 PM

sam-knox: We're just talking about the label and help text for the custom setting here right? For the field label in NPSP Settings, I do like "Recurring Donation Name Format" better. The label used for Opportunity names is "Opportunity Name Format".

andrewyu-salesforce:
@force2b @sam-knox The label is updated

rjanjanam:
There is one Robot test failure and is due to flakiness of the environment and related to the code change. 
Validated the change
Looks good

sam-knox:
@andrewyu-salesforce - I just tried this in a scratch org and Automatic Naming is turned on, but in NPSP Settings, the default Name Format is "Disable Auto Naming". Shouldn't one of the actual formats be selected by default here?

andrewyu-salesforce:
@sam-knox I am still in development of the actual logic under the same branch. This PR is to get the Custom Setting field into master before the freeze. The actual logic is developed in this #5636

sam-knox:
@andrewyu-salesforce - Understood. Thanks for the clarification!

PR #5639, opened 2020-07-02
Gift Entry: Select columns for display in Batch Gifts table

### LINKED ISSUE #5647: Gift Entry: Objects Without Name Fields Cause "No such column 'name'" Error
When Lookup fields whose related object do not have a "Name" field are included in a Template, the Gift Entry screen shows a "No such column 'name' on entity <ObjectName>" error:

**Steps to reproduce:**
1. Create a Lookup field on Data Import that is related to an object without a "Name" field (like Case)
2. Create a Lookup field (to Case) on Opportunity
3. Create a Field Mapping for these two fields
4. Create a Template that includes the new Case Lookup field
5. Create a Batch from that Template
6. Open the new Batch and observe the "No such column" error

**Expected behavior:**
Lookups that reference objects without Name fields should not throw an error and should instead use whatever field is used in place of "Name" on these objects (for example, CaseNumber on Case)

kenlewis-sfdo:
This set of changes adds a menu to the Batch Gifts table with
an action item that can be used to customize the columns that
are displayed in the table and makes lookup fields display as
links in the table.

[W-7780501](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHvYIAW/view)

# Critical Changes

# Changes

# Issues Closed
Fixes #5647

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -140,19 +141,7 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
     batchGiftSubmit(event) {
         const table = this.template.querySelector('c-ge-batch-gift-entry-table');
         this.dataImportRecord = event.detail.dataImportRecord;
-        saveAndDryRunDataImport({batchId: this.recordId, dataImport: this.dataImportRecord})
```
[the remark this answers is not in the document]
kenlewis-sfdo: This move is because after the dataImportModel comes back from the server it needs to be appended with the additional properties used by columns to display links.  That requires the Data Import object info, which is already in the Table but does not exist in the App.  So either the App needed to get access to the DI object info (by adding another getObjectInfo) or the function needed to move.  I think the latter is better for the structure of our App.  

My opinion on functionality being housed at the App level is evolving ever so slightly. :) Shared state should be managed at the App level, but functionality/logic that is only used by one child component should just be housed in that child component to keep them as loosely coupled as possible and to separate out concerns. This change is a further move toward those two goals.  

The App is still doing an update after loading a row into the form, which could/should also be moved down here to the table, but other than that almost everything concerning displaying/managing the entered gifts (DI rows) is done by the Table component.  None of that logic is shared and there is very little shared state (just the amount and count totals which the Table communicates up to the App via events).  That way the App is working more as an orchestrator - so when a new gift is saved in Batch mode, the App passes the DI to the table which saves the record and then manipulates the response directly in order to display the rows correctly.  Happy to discuss this further or try to add detail to this thought process maybe in our refactoring doc...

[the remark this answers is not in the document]
ptewson-sfdo: Thanks - I think that explanation was incredibly clear and catches me up to the latest-and-greatest thinking! Thank you!

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -116,34 +119,47 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
+                                fieldMapping.Source_Field_API_Name :
+                                isObjectMappedLookup ?
+                                        fieldMappingName : null;
                         // Add the source fields used in the Template
-                        if (!fieldApiNames.contains(
-                                fieldMapping.Source_Field_API_Name.toLowerCase())) {
-                            fieldApiNames.add(
-                                    fieldMapping.Source_Field_API_Name.toLowerCase());
+                        if (!fieldApiNames.contains(sourceFieldName.toLowerCase())) {
+                            fieldApiNames.add(sourceFieldName.toLowerCase());
```
ptewson-sfdo, line 136: I don't think we should make this change RIGHT now, but for the future - a combination of sourceFieldName + targetFieldName might be a better key here, to account for possible cases where the same DI field is mapped to multiple locations (as is frequently the case with the Record Imported fields)

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -460,6 +493,8 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
     */
     private static List<DataImport__c> getDataImportRecordsByBatchId(Id batchId, Integer offset) {
         String query = getDataImportQuery(batchId, null, loadSize, offset);
+        System.debug(LoggingLevel.ERROR, '*** variable value: ' + query);
+        System.debug(LoggingLevel.ERROR, '^^^^^^ query ^^^^^^');
```
ptewson-sfdo: Need to remove these debugs!

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -100,11 +100,44 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
+                    WITH SECURITY_ENFORCED
+            ];
+        } catch (QueryException qe) {
+            // If User doesn't have access to Version field, treat as 1.0.
+            isQueryException = true;
+        }
+
+        if (isQueryException || batch.Batch_Gift_Entry_Version__c < 2) {
+            BGE_BatchGiftEntry_UTIL.checkFieldPermissions(
+                    UTIL_Namespace.StrTokenNSPrefix('DataImport__c'),
```
ptewson-sfdo, line 119: Schema.SObjectType.DataImport__c.getName()?

ptewson-sfdo: Or ... String.valueOf(DataImport__c.SObjectType)? Does that work too?

PR #5645, opened 2020-07-06
Feature/vmartinov/batch status

vmartinov-sforg:
* Add a new Batch Status Field to the Batch Data Import object and set the values automatically based on the bdi record statuses in the batch.
# Critical Changes

# Changes
- We added a new Batch Status field that can be used in Gift Entry and other features that use NPSP Data Import Batches. The Batch Status is based on the Import Status of the records within the batch. If one record fails to be processed, the status for the batch is "Failed - Needs Review." In Gift Entry, you may want to add this new field to the listview for Batches for quick reference. Learn more about updating these columns in Customize Columns on Batches & Templates: https://powerofus.force.com/s/article/NPSP-Configure-Gift-Entry#ariaid-title7
# Issues Closed

# Community Ideas Delivered

# New Metadata
- DataImportBatch__c.Status__
- geWarningGauAllocFieldUpdatesDisabled label

# Deleted Metadata


vmartinov-sforg:
GUS WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHvdIAG/view

src/objects/DataImportBatch__c.object
```diff
@@ -430,6 +430,36 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>Status__c</fullName>
+        <description>The status of the batch, for example, Open or Completed.</description>
```
melissabarber, line 435:
```suggestion
        <description>The Batch Status is based on the Import Status of the NPSP Data Import records within the batch. If one record fails to import, the status for the batch is "Failed - Needs Review."</description>
```

src/objects/DataImportBatch__c.object
```diff
@@ -430,6 +430,36 @@
...
+        <externalId>false</externalId>
+        <inlineHelpText>The status of the batch, for example, Open or Completed.</inlineHelpText>
```
melissabarber, line 437:
```suggestion
        <inlineHelpText>The Batch Status is based on the Import Status of the NPSP Data Import records within the batch. If one record fails to import, the status for the batch is "Failed - Needs Review."</inlineHelpText>
```

src/objects/DataImportBatch__c.object
```diff
@@ -430,6 +430,36 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>Status__c</fullName>
```
melissabarber, line 434:
```suggestion
        <fullName>Batch_Status__c</fullName>
```

src/objects/DataImportBatch__c.object
```diff
@@ -430,6 +430,36 @@
...
+        <description>The status of the batch, for example, Open or Completed.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The status of the batch, for example, Open or Completed.</inlineHelpText>
+        <label>Status</label>
```
melissabarber, line 438:
```suggestion
        <label>Batch Status</label>
```

src/objects/DataImportBatch__c.object
melissabarber: @vmartinov-sforg Would having this field name more clearly call out that it's the batch's status make reporting easier for customers?

src/objects/DataImportBatch__c.object
vmartinov-sforg: Done

src/objects/DataImportBatch__c.object
vmartinov-sforg: Done.

src/objects/DataImportBatch__c.object
vmartinov-sforg: Done.

src/objects/DataImportBatch__c.object
vmartinov-sforg: Done.

melissabarber:
@vmartinov-sforg I'm all good on this PR. Thanks!

src/classes/BDI_DataImportBatch_SEL.cls
```diff
@@ -60,7 +60,8 @@ public with sharing class BDI_DataImportBatch_SEL {
                     Records_Failed__c,
                     Records_Successfully_Processed__c,
                     Post_Process_Implementing_Class__c,
-                    Run_Opportunity_Rollups_while_Processing__c
+                    Run_Opportunity_Rollups_while_Processing__c,
```
ptewson-sfdo, line 63: @vmartinov-sforg just curious, does this resolve the issue we've been seeing on Batch clone and Batch edit when it's not through the Gift Entry interface?

vmartinov-sforg: @ptewson-sfdo Can you remind me what issue you're referring to?

src/classes/BDI_DataImport_TDTM.cls
```diff
@@ -0,0 +1,208 @@
...
+      // get all the batch records
+      List<DataImportBatch__c> batches = [SELECT Id, Batch_Status__c
+                                          FROM DataImportBatch__c
+                                          WHERE Id IN :batchIdSet];
+
+      // get all the di records in all the batches
+      List<DataImport__c> diRecords = [SELECT Id, NPSP_Data_Import_Batch__c, Status__c 
+                                          FROM DataImport__c
+                                          WHERE NPSP_Data_Import_Batch__c IN :batchIdSet];
+
```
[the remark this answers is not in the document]
ptewson-sfdo: (although maybe you still have to do the following iteration to get the aggregate figures, so it might not make that much of a difference)

src/labels/CustomLabels.labels
```diff
@@ -7100,6 +7100,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Warning message for deleting a section if it contains required fields</shortDescription>
         <value>Warning: By deleting this section, you are also removing required fields from the template. These fields need to be added into another section before you can save the template.</value>
     </labels>
+    <labels>
+        <fullName>geWarningGauAllocFieldUpdatesDisabled</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Warning for preventing update to some fields when GAUs exist on Donation</shortDescription>
```
melissabarber, line 7108:
```suggestion
        <shortDescription>Warning about some uneditable fields when donation selected already has a GAU Allocation assigned</shortDescription>
```

melissabarber:
@vmartinov-sforg & @ptewson-sfdo I added a suggestion for the shortDescription of the custom label. Thanks!

ptewson-sfdo:
@vmartinov-sforg just FYI I added a custom label to this PR that, while technically unrelated, gets all our forward-looking translation collateral in one branch for the localization team. Happy to fill you in on details when you get back!

src/labels/CustomLabels.labels
vmartinov-sforg: done.

src/classes/BDI_DataImport_TDTM.cls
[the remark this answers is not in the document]
vmartinov-sforg: Done.

src/classes/BDI_DataImport_TDTM.cls
```diff
@@ -0,0 +1,208 @@
...
+    private void setBatchStatus(List<DataImport__c> diOldList, 
+                                List<DataImport__c> diNewList) {
+
+      List<DataImport__c> diList = new List<DataImport__c>();
+      Map<Id, DataImport__c> diNewMap = new Map<Id, DataImport__c>(diNewList);
+
+      for(DataImport__c diOld : diOldList) {
+        DataImport__c diNew = diNewMap.get(diOld.Id);
+        if (diOld.Status__c != diNew.Status__c) {
+          diList.add(diOld);
```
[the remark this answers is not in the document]
vmartinov-sforg: Done.

src/classes/BDI_DataImport_TDTM.cls
```diff
@@ -0,0 +1,208 @@
...
+     */
+    private void setBatchStatus(List<DataImport__c> diOldList, 
+                                List<DataImport__c> diNewList) {
+
+      List<DataImport__c> diList = new List<DataImport__c>();
+      Map<Id, DataImport__c> diNewMap = new Map<Id, DataImport__c>(diNewList);
+
+      for(DataImport__c diOld : diOldList) {
+        DataImport__c diNew = diNewMap.get(diOld.Id);
+        if (diOld.Status__c != diNew.Status__c) {
```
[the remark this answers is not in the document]
vmartinov-sforg: Yes, good catch. Done.

melissabarber:
@vmartinov-sforg and @ptewson-sfdo I'm good to go on this PR. Thanks for adding me as a reviewer!

src/classes/BDI_DataImportBatch_TDTM.cls
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * @description In the before trigger context, sets the status for the batch
+     * @param List<DataImportBatch__c> 
+     */
+    private void handleBefore(List<DataImportBatch__c> batches) {
+        for (DataImportBatch__c b : batches) {
+            b.Batch_Status__c = BATCH_STATUS_OPEN;
+        }
```
kenlewis-sfdo, line 68: Just curious here - could this field be given a default value instead of using a trigger to always set new records to a single value?

src/classes/BDI_DataImportBatch_TDTM.cls
```diff
@@ -0,0 +1,70 @@
...
+* @date 2020
+* @group Batch Data Import
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Trigger Handler for the DataImportBatch SObject
+*/
+public class BDI_DataImportBatch_TDTM extends TDTM_Runnable {
+    
+    public static final String BATCH_STATUS_OPEN = 'Open';
+    public static final String BATCH_STATUS_COMPLETED = 'Completed';
+    public static final String BATCH_STATUS_FAILED = 'Failed - Needs Review';
```
kenlewis-sfdo, line 41: Could/should these statuses be exposed as custom labels so they can be translated?

src/classes/BDI_DataImport_TDTM.cls
```diff
@@ -0,0 +1,225 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Batch Data Import
+* @group-content ../../ApexDocContent/BatchDataImport.htm
+* @description Trigger Handler for the DataImport SObject
+*/
+public class BDI_DataImport_TDTM extends TDTM_Runnable {
```
kenlewis-sfdo, line 37: Should this class name be more specific to the functionality, like BDI_DataImportBatchStatus_TDTM?  I know we have a few TDTM classes that are named generally, but I think having TDTM classes that are named and implemented as specific and siloed helps leverage the adjustability of TDTM (re-ordering logic using the Trigger Handler entries).

src/classes/BDI_DataImportBatch_TDTM.cls
vmartinov-sforg: We could make these strings to be labels, however the di statuses were already stored as strings as well, see BDI_DataImport_API.cls (global static final string bdiImported = 'Imported'; etc). We should be consistent on both fronts I think. I am open to making all the statuses labels, but that would be a bigger refactoring as they are referenced elsewhere in the bdi code. @kenlewis-sfdo 
cc: @ptewson-sfdo.

src/classes/BDI_DataImport_TDTM.cls
vmartinov-sforg: I agree, I'll rename the class. Good point.

vmartinov-sforg: done.

src/classes/BDI_DataImportBatch_TDTM.cls
vmartinov-sforg: Good point, and I did consider it. I looked at what we do for the di status picklist, and we do not set a default value in that case, that's why I went with the trigger for the batch. I am open to go with the default value though, but can we quickly chat tomorrow? Let's also talk about the labels.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -0,0 +1,224 @@
...
+
+        return null;
+    }
+
+    /**
+     * @description updates the batch status on di records insert, delete or undelete
+     * @param diList the old or new list
+     */
+    private void setBatchStatus(List<DataImport__c> diList) {
+      Set<Id> batchIdSet = new Set<Id>();
```
ptewson-sfdo, line 70: This comment applies to the total work here I think - the tab settings look like they are a 2 spaces and I believe our standard is to use 4 spaces?

src/classes/BDI_DataImportBatch_TDTM.cls
ptewson-sfdo: This is interesting - I think that DI statuses are strings and not labels, at least in part, because the values are heavily used - even primarily used - as flags in the code for conditional logic, etc. As a result translating them causes all sorts of problems - even if we consistently tested against the label value, different users can be in different languages. I definitely would be cautious about making that change for the DI labels without carefully taking into account the potential impact of translations and other label overrides. 

This seems like a similar case here, and so I might be on a starting position of preferring them declared as strings here (or perhaps enum?). But I imagine we do want translated values to appear to user - so we may need to cook up a plan to make that work.

ptewson-sfdo: Also - given that the eventual destination of these values is in a picklist, could we just grab the strings here from the picklist describe information?

src/classes/BDI_DataImportBatchStatus_TDTM.cls
vmartinov-sforg: Right, Not sure how I ended up with 2 spaces. I will update all the files.

src/classes/BDI_DataImportBatch_TDTM.cls
vmartinov-sforg: We can get the list from the describe, however how do I identify the 'Open' picklist value for example without a comparison with a string 'Open'?

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -0,0 +1,224 @@
+/*
+    Copyright (c) 20120 Salesforce.org
```
kenlewis-sfdo, line 2:
```suggestion
    Copyright (c) 2020 Salesforce.org
```

src/classes/BDI_DataImportBatch_TDTM.cls
ptewson-sfdo: I'm going to mark this "resolved" based on our conversation today - although there's a broader discussion to be had I believe we agreed not to tie it to this delivery

Issue #15, opened 2013-09-26
As a sysadmin I want to choose how to be notified when an error occurs in Cumulus so I can decide my own workflow


mrbelvedere:
Sysadmin should have the option to select multiple ways of being notified when an error occurs, these ways could include:
- Email
- Log-Only
- Chatter
  (Much of this can be pulled from Recurring Donation)
  ---------------
  Initial page design implemented in Sprint 8.

PR #5656, opened 2020-07-07
Custom error messages for Validation Rules not displaying on Gift Entry

### LINKED ISSUE #5605: Custom error messages for validation rules not displaying on Gift Entry
Validation Rules on NPSP object are appearing at the right time on Gift Entry (when the Condition is met), but the error message chosen doesn't appear.

NOTE: We should test this for both page-level validation messages and field-level validation messages; also need to do some regression on other error message display (with/without CC widget, allocations, errors on DI save and errors on processing, etc). 

Pre-Conditions:
AM & GE should be enabled

Steps to reproduce:
1- Create a Validation rule on the NPSP Data Import object (for example: "npsp__Donation_Amount__c > 10")
2- Enter a custom error message for the Validation Rule
3- Attempt to enter a gift through Batch or Single gift that complies with the Condition (for example: amount 30)

Expected Result: Gift should not be entered and custom error message should be displayed at the top of the page
Actual result: Gift is not entered but error message at the top is blank

dfuller114:
If a validation rules causes a DML exception when submitting a gift, the page-level error message will no longer be blank.

# Critical Changes

# Changes

# Issues Closed
Fixes #5605 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/utilCustomErrors/exceptionDataError.js
```diff
@@ -9,21 +10,14 @@ class ExceptionDataError extends Error {
...
-            const apexExceptionWrapper = JSON.parse(apexException.body.message);
+        if (isNotEmpty(apexException.body)) {
+            const apexExceptionWrapper = validateJSONString(apexException.body.message);
+            this.stackTrace = validateJSONString(apexException.body.stackTrace);
             this.exceptionType = apexExceptionWrapper.exceptionType;
-            this.errorMessage = apexExceptionWrapper.errorMessage;
-            this.DMLErrorMessageMapping = apexExceptionWrapper.DMLErrorMessageMapping;
-            this.DMLErrorFieldNameMapping = apexExceptionWrapper.DMLErrorFieldNameMapping;
+            this.isUserDefinedException = apexException.body.isUserDefinedException;
+            this.DMLErrorMessageMapping = apexExceptionWrapper.DMLErrorMessageMapping;;
```
psadie-sfdo, line 19:
```suggestion
            this.DMLErrorMessageMapping = apexExceptionWrapper.DMLErrorMessageMapping;
```

src/lwc/utilCustomErrors/exceptionDataError.js
```diff
@@ -9,21 +10,14 @@ class ExceptionDataError extends Error {
...
+            this.DMLErrorFieldNameMapping = apexExceptionWrapper.DMLErrorFieldNameMapping;;
```
psadie-sfdo, line 20:
```suggestion
            this.DMLErrorFieldNameMapping = apexExceptionWrapper.DMLErrorFieldNameMapping;
```

src/classes/GE_GiftEntryController.cls
```diff
@@ -65,16 +65,17 @@ public with sharing class GE_GiftEntryController {
...
+            Boolean hasTimedout = ex.getMessage().contains(TIMED_OUT);
 
             if (hasTimedout) {
                 return getPurchaseCallTimeoutResponse(ex, dataImportRecordId);
             } else {
-                String JSONExceptionData = ERR_ExceptionData.createExceptionWrapperJSONString(ex);
-                throw returnDmlException(JSONExceptionData);
+                return null;
             }
+        } catch(DmlException ex) {
```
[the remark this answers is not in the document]
dfuller114: @voduyemi the issue there that I saw is that we are throwing an exception returned from `returnDmlException()` which only returns a `DmlException`. If we don't have any specific way we would handle other exceptions like a `JSONException`, for example, that exception would still throw on line 77 and display unhandled to the UI. It's only the specific exceptions where we've written methods to handle those types that I've added the more specific type for the catch parameter.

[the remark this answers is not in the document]
voduyemi: I believe `returnDmlException` should not be used in this instance since the exception caught should be based on the action/call performed. This might not need happen in this WI to minimize the impact of your change since we specifically handle purchase call errors in a certain way to control what happens after a failed purchase attempt. The exception on line 77 will also not get thrown if a DMLException is not caught on line 76. If we still want to use `returnDmlException`, we can use the parent `Exception` class and to act like a 'catch-all' if other errors such as 
 the `JSONException` happens.

src/classes/GE_GiftEntryController.cls
```diff
@@ -65,16 +65,17 @@ public with sharing class GE_GiftEntryController {
             GE_PaymentServices.ResponseWrapper response = paymentServices.createTransaction();
 
             return JSON.serialize(response);
-        } catch (Exception ex) {
-            Boolean isCalloutException = ex instanceof System.CalloutException;
-            Boolean hasTimedout = isCalloutException && ex.getMessage().contains(TIMED_OUT);
+        } catch (CalloutException ex) {
+            Boolean hasTimedout = ex.getMessage().contains(TIMED_OUT);
```
bdvorachek, line 69:
```suggestion
            Boolean hasTimedOut = ex.getMessage().contains(TIMED_OUT);
```

Camelcase on word borders here.

src/classes/GE_GiftEntryController.cls
```diff
@@ -65,16 +65,17 @@ public with sharing class GE_GiftEntryController {
...
 
             if (hasTimedout) {
```
bdvorachek, line 71:
```suggestion
            if (hasTimedOut) {
```

src/classes/GE_GiftEntryController.cls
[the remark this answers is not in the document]
dfuller114: Good point. I think the `AuraHandledException` is better to use as a catch-all so I could just replace `returnDmlException()` with `returnAuraException()` and use the `Exception` generic type. Main goal is to try and think about how to be more discrete about what errors we expect to occur and how we want to handle different exception types. Something I've been wanting to get better about. How does that sound?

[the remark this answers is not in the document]
voduyemi: @dfuller114 That works, as long as we make sure that we are still able to handle the error thrown with `returnAuraException` gracefully in the front-end.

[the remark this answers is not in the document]
dfuller114: @psadie-sfdo do you see any issue with changing this to an `AuraHandledException` instead of `DmlException`?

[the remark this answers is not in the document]
psadie-sfdo: Nope, I think that's definitely more appropriate than the `DmlException`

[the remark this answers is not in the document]
kenlewis-sfdo: I would just add to this thread that I would avoid using `GE_GiftEntryController.returnAuraException` or `GE_GiftEntryController.returnDmlException`.  I added a line (#15) to our refactor doc.  I think we should avoid using these and look to remove them.

src/classes/GE_GiftEntryController.cls
```diff
@@ -147,11 +148,11 @@ public with sharing class GE_GiftEntryController {
             upsertDataImport(dataImport);
             return BGE_DataImportBatchEntry_CTRL.runSingleDryRun(dataImport.Id, batchId);
 
-        } catch (Exception ex) {
+        } catch (DmlException ex) {
             // create custom lwc exception and throw it
             String JSONExceptionData = ERR_ExceptionData.createExceptionWrapperJSONString(ex);
 
-            throw returnDmlException(JSONExceptionData);
+//            throw returnDmlException(JSONExceptionData);
```
ptewson-sfdo, line 155: @dfuller114 This is causing a deploy error since the method demands a string type! Do we have a solid understanding of what we want our error handling behavior to be in the case where we land in this catch?

dfuller114: @ptewson-sfdo i think that failure might have been due to a leftover artifact that I just removed and pushed up. The deploy should be successful now. Here, we're expecting a DML exception so when it goes into the catch we'll return a `DMLException` data type

src/lwc/utilCustomErrors/exceptionDataError.js
```diff
@@ -9,22 +10,16 @@ class ExceptionDataError extends Error {
     */
     constructor(apexException) {
         super();
-        if (apexException.body && apexException.body.exceptionType) {
```
ptewson-sfdo, line 12: @dfuller114 I'm worried about removing this block - before your updates here, this method looks like it was built to handle exceptions that were provisioned with serialized ERR_ExceptionData (which you've preserved), and also handle exceptions in their raw format, which was handled in this block that's being removed. How sure are we that we don't have to handle basic, "untreated" exceptions?

src/classes/GE_GiftEntryController.cls
```diff
@@ -65,16 +65,17 @@ public with sharing class GE_GiftEntryController {
...
-            Boolean hasTimedout = isCalloutException && ex.getMessage().contains(TIMED_OUT);
+        } catch (CalloutException ex) {
+            Boolean hasTimedout = ex.getMessage().contains(TIMED_OUT);
 
             if (hasTimedout) {
                 return getPurchaseCallTimeoutResponse(ex, dataImportRecordId);
             } else {
-                String JSONExceptionData = ERR_ExceptionData.createExceptionWrapperJSONString(ex);
-                throw returnDmlException(JSONExceptionData);
+                return null;
```
ptewson-sfdo, line 74: @dfuller114 What's the logic behind returning "null" here? What happens from the user perspective if we hit this line? This seems like any CalloutException that's not a timeout gets caught and dead-ends unhelpfully here, rather than returning to the user in a way that provides a helpful message (that said, I didn't manage to actually hit any weird behavior here)

src/lwc/utilCustomErrors/exceptionDataError.js
dfuller114: @ptewson-sfdo the idea was to simplify the logic here and remove duplicate logic since it doesn't seem like these two blocks are necessary. We are populating the majority of all the properties of the `Error` parent in both blocks, which will then be handled in whatever LWC component consumes it. In the second block, the `message` property is parsed which is what the `validateJSONString` function does if the string is valid JSON. So it seemed cleaner to have the first block populate all the properties if we have a `body` and `exceptionType` (whether there are values there or not), then have the second block directly set the `apexException`  to the `message` property if that scenario applies which is a different scenario. The `if, if/else` blocks don't seem different enough to warrant an entirely separate block. @psadie-sfdo you might can confirm. think we discussed this some when I made the change.

src/classes/GE_GiftEntryController.cls
dfuller114: @ptewson-sfdo looking back at this, I don't like returning `null` since it's an anti-pattern. I just pushed up a change for this where we just conditionally set a `timeoutResponse` string variable if that `TIMED_OUT` boolean is true; otherwise, the string is just an empty string that gets returned.

src/lwc/utilCustomErrors/exceptionDataError.js
ptewson-sfdo: I think that as this is written, it will choke on exceptions that haven't been provisioned with the serialized content in the message (validateJSONString() will return `false` and then the attempts to reference properties of apexExceptionWrapper will fail. I think we an protect against that, however, by re-writing this block (below the call to super()) with something like: 
```

       const apexExceptionWrapper = validateJSONString(apexException.body.message);
        if (apexExceptionWrapper === false) {
            this.exceptionType = apexException.body.exceptionType;
            this.errorMessage = apexException.body.message;
            this.stackTrace = apexException.body.stackTrace;
            this.isUserDefinedException = apexException.body.isUserDefinedException;
            this.DMLErrorMessageMapping = {};
            this.DMLErrorFieldNameMapping = {};

        } else if (apexException.body && apexException.body.message) {
            // This looks like an instance of the apex wrapper class ERR_ExceptionData
            this.exceptionType = apexExceptionWrapper.exceptionType;
            this.errorMessage = apexExceptionWrapper.errorMessage;
            this.DMLErrorMessageMapping = apexExceptionWrapper.DMLErrorMessageMapping;
            this.DMLErrorFieldNameMapping = apexExceptionWrapper.DMLErrorFieldNameMapping;

        } else {
            this.message = apexException;
        }
```

Unfortunately, I think based on the conversation with the team we should roll back the rest of the changes here for the time being. I think we've collectively identified a need for refactor with error handling (and de-prioritized the field-level messaging, which I think introduced a lot of the complexity in the first place), but we need to approach that with caution and enough time to deliver that change thoughtfully. For this Work Item we probably need to focus on the updates that are constrained to closing the bug that originated this work - which I believe the suggested change here does.

src/classes/GE_GiftEntryController.cls
```diff
@@ -147,7 +149,7 @@ public with sharing class GE_GiftEntryController {
             upsertDataImport(dataImport);
             return BGE_DataImportBatchEntry_CTRL.runSingleDryRun(dataImport.Id, batchId);
 
-        } catch (Exception ex) {
+        } catch (DmlException ex) {
```
ptewson-sfdo, line 152: From our conversation with the team, I think we need to revert these changes and keep this as generic "Exception" for the time being

src/classes/GE_GiftEntryController.cls
```diff
@@ -199,11 +201,10 @@ public with sharing class GE_GiftEntryController {
     @AuraEnabled
     public static DataImport__c upsertDataImport(DataImport__c dataImport) {
         try {
-
             upsert dataImport Id;
 
             return dataImport;
-        } catch (Exception e) {
+        } catch (DmlException e) {
```
ptewson-sfdo, line 207: Revert this change and keep this as generic "Exception" for the time being

src/classes/GE_GiftEntryController.cls
```diff
@@ -226,7 +227,7 @@ public with sharing class GE_GiftEntryController {
             List<DataImport__c> listDI = new List<DataImport__c>{dataImport};
 
             BDI_DataImport_API.processDataImportRecords(diSettings, listDI, isDryRun);
-        } catch (Exception e) {
+        } catch (DmlException e) {
```
ptewson-sfdo, line 230: Revert this change and keep this as generic "Exception" for the time being

src/lwc/utilCustomErrors/exceptionDataError.js
```diff
@@ -9,22 +10,22 @@ class ExceptionDataError extends Error {
...
-            this.exceptionType = apexException.body.exceptionType;
-            this.errorMessage = apexException.body.message;
+
+
```
ptewson-sfdo, line 14: Remove extra blank line

src/classes/GE_GiftEntryController.cls
```diff
@@ -65,16 +65,18 @@ public with sharing class GE_GiftEntryController {
...
+            String timeoutResponse = '';
+
+            if (hasTimedOut) {
+                timeoutResponse = getPurchaseCallTimeoutResponse(ex, dataImportRecordId);
             }
+
+            return timeoutResponse;
+        } catch(AuraHandledException ex) {
+            String JSONExceptionData = ERR_ExceptionData.createExceptionWrapperJSONString(ex);
+            throw returnAuraException(JSONExceptionData);
```
ptewson-sfdo, line 79: I'm no fan of the original block, but to avoid the risk of causing a bug we need to revert to the original code here.

src/lwc/utilCustomErrors/exceptionDataError.js
dfuller114: done

src/classes/GE_GiftEntryController.cls
dfuller114: done

src/classes/GE_GiftEntryController.cls
dfuller114: done

src/classes/GE_GiftEntryController.cls
dfuller114: done

ptewson-sfdo:
> Looks good to me! Also looking forward to the deeper refactor, thank you for all the thought and work thus far!
> 
> _One thing that I noticed which may not be an issue, but wanted to pose to ya'll. For the remaining scenarios where we're still explicitly passing through and throwing our own `returnXException` and the exception type that gets thrown doesn't match with X, the custom message I passed doesn't come through. Instead I receive the generic `Script-exception thrown`. I believe this is working as expected, but wasn't 100% certain._
> 
> _Quick example to see this in action is to throw an AuraHandledException error inside the `upsertDataImport` method of `GE_GiftEntryController.cls`_

@psadie-sfdo - I don't think that is desired, honestly - I think we would prefer to include the specified message in all cases, and I don't believe the "Script-exception thrown" is specified anywhere in our product definition. There might be some cases as a fallback behavior in areas where we don't have a more specific error defined, but honestly even then I personally would prefer to present the message to users rather than something completely generic. 

Personally I would make that a general goal for our refactor - display specific message where the label is defined, display the message from the exception if it's a case we haven't specifically defined. Happy to have more conversation about that however!

PR #5657, opened 2020-07-10
Fix Enhanced Recurring Donations Creates 2 Opportunities When New

### LINKED ISSUE #5597: Enhanced Recurring Donation Generates 2 Installment Opportunities on Create
When creating a Recurring Donation, if there is an update made to either a field that the Enhanced Recurring Donation considers a key field, or a field that is mapped in the Field Mapping section of NPSP Settings for Recurring Donations.

**Steps to Recreate**
1. Follow the Setup steps for enabling Enhanced Recurring Donations (https://powerofus.force.com/s/article/NPSP-Enhanced-RD-Upgrade-Guide)
2. After upgrading, go to NPSP Settings > Recurring Donations > Recurring Donation Custom Field Mappings and Map **Recurring Donation Name** to Opportunity **Name**
3. Ensure that the Recurring Donation setting "Install Opportunity Auto-Creation = Always Create Next Installment
4. In Setup > Custom Settings > Recurring Donation Settings > Manage the settings and disable "Automatic Recurring Donation Naming"

5. Create a Workflow Rule 
a. **Rule Name** = RD Type Open
b. **Active** = TRUE
c. **Rule Criteria** = Recurring Donation: Recurring Type _equals_ Open
d. **Object** = Recurring Donation
e. **Evaluation Criteria** = Evaluate the rule when a record is created, and any time it's edited to subsequently meet criteria
6. Create a Field Update linked to the Workflow from previous step
a. **Name** = RD Name - Open
b. **Object** = Recurring Donation
c. **Field to Update** = Recurring Donation: Recurring Donation Name
d. **Field Data Type** = Text
e. **Re-evaluate Work Rules after Field Change** = FALSE
f. Formula Value = `npe03__Organization__r.Name &" OPEN RecDon for " &" $"& TEXT(npe03__Amount__c) &" "& TEXT(npe03__Installment_Period__c)`
7. Create a Second Workflow Rule
a. **Rule Name** = RD-Open Exists on Opportunity
b. **Active** = TRUE
c. **Rule Criteria** = (Opportunity: Recurring Donation Installement Name NOT EQUAL to `NULL`) AND (Opportunity: Recurring Donation Installment Name CONTAINS `of 0)`)
d. **Object** = Opportunity
e. **Evaluation Criteria** = Evaluate the rule when a record is created
8. Create a Field Update linked to the Workflow from previous step
a. **Name** = Opportunity Type is New Funding
b. **Object** = Opportunity
c. **Field to Update** = Opportunity: Type
d. **Field Data Type** = Picklist
e. **Re-evaluate Work Rules after Field Change** = FALSE
f. **Formula Value** = `New Funding`
9. Create a Recurring Donation:
a. **Date Established** = <Today's Date>
b. **Recurring Type** = Open
c. **Number of Planned Installments** = 1 (default, do not change)
d. **Amount** = $100
e. **Installment Period** = Monthly
f. **Day of Month** = <1 day after today's date>
g. **Effective Date** = <same date as Date Established>
h. **Recurring Donation Name** = <any value>
10. Save the Recurring Donation
11. Wait a few seconds (in some cases longer depending on if there are other queued processes waiting

**Cause**
Because the Recurring Donation Field Mapping has the Recurring Donation Name mapping to the Opportunity Name, the [RD2_OpportunityEvaluationService.cls](https://github.com/SalesforceFoundation/NPSP/blob/3172d24e313a05faca3b47c2b0b328e64289a61a/src/classes/RD2_OpportunityEvaluationService.cls) reassesses the creation of the Installment Opportunities of the Recurring Donation. Two queueable handlers are created to create Opportunities because there is both a Workflow setting to update the Recurring Donation Name and Opportunity Name, at the same time the Recurring Donation Field Mapping is updating the Opportunity Name. Running asynchronously before the record is actually inserted is what is causing the duplication of actions to generate the Opportunities. This line in the class shows where the Field Mappings are being re-evaluated in the same section, following a list of key fields that automatically trigger a reevaluation - https://github.com/SalesforceFoundation/NPSP/blob/3172d24e313a05faca3b47c2b0b328e64289a61a/src/classes/RD2_OpportunityEvaluationService.cls#L321

**Workaround**
Removing multiple updates made to the same key fields or mapped fields via Recurring Donation settings, via workflow, process builder, code, or Recurring Donation Field Mapping, as listed in the code here:
https://github.com/SalesforceFoundation/NPSP/blob/3172d24e313a05faca3b47c2b0b328e64289a61a/src/classes/RD2_OpportunityEvaluationService.cls#L301

npsp-reedestockton:
# Critical Changes

# Changes
- We fixed an issue with Enhanced Recurring Donations that would sometimes result in the creation of duplicate Opportunities when creating a new Recurring Donation.

# Issues Closed
Fixes #5597

# Community Ideas Delivered

# New Metadata
Custom Field:
- npe03__Recurring_Donations_Settings__c.InstallmentOppFirstCreateMode__c

# Deleted Metadata


npsp-reedestockton:
@force2b @RadSFDeveloper Ready for review.  Thank you!!

npsp-reedestockton:
@smarovich Sam Knox suggested we reach out to you for Doc needs in his absence.  Would you please review the PR description here or suggest another member of the team to review?  Thank you!!!

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -455,6 +471,9 @@ public inherited sharing class RD2_OpportunityEvaluationService {
         if (isBatchContext) {
             return !rdSettings.isDisableAllInstallments;
 
+        } else if (rd.isNew() && triggerAction == TDTM_Runnable.Action.AfterUpdate) {
```
force2b, line 474: I'm just wondering if it's better to control the flow of calling the service to the [RD2_RecurringDonations_TDTM](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/RD2_RecurringDonations_TDTM.cls#L380) class rather than here. Assuming it's possible to validate the same `isNew` condition in that class, we could just not call the Queueable from the trigger on an AfterUpdate context. Evaluating `isNew` may not be feasible in the TDTM class too, but I wanted to throw it out there. If I follow the logic correctly, `isNew` is determined by there being no related Opps. If true, that likely can't be determined in the TDTM class anyway.

npsp-reedestockton: RD2_RecurringDonations_TDTM needs to call RD2_OpportunityEvaluationService from both AfterInsert and AfterUpdate.  In the case of a potentially changed schedule, it needs to take a look at the existing Opportunities and take appropriate action.

You're right, isNew() is based on whether there are related opportunities.

My conclusion was that the approach least likely to have unintended side effects was to isolate the narrow set of circumstances that produced the bug, and that is the combination of a Recurring Donation that RD2_RecurringDonation determines is new based upon no related Opportunities and a call from RD2_RecurringDonations_TDTM with an AfterUpdate action type.

RadSFDeveloper:
[W-7782150](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EIHmIAO/view)

src/classes/RD2_OpportunityEvaluationService.cls
force2b: Makes sense. This also confirms that my initial guess on a cause was way off. I had originally thought that it was calling the queueable only when it thought a new Opp was needed, but that was wrong. The queueable itself makes the proper determination rather than the trigger. As you've identified though, it's getting hit from both an after insert and an after update.

What would happen in the following scenario (not that I think this is likely)?
- A new RD is created and a new Opp is created with it
- The open Opp is deleted (manually or otherwise)

Would a new opp be created by the trigger, or would that happen when the nightly job runs?

RadSFDeveloper: Related to the initial comment and the reply above, if the Opp eval service is called after the update or new RDs in the case Schedule changes, should we actually skip the Opp eval service call after the insert and call it only once within the transaction, ie `afterUpdate`? If so, then we would keep the trigger actions within the trigger scope and the Opp eval queueable and the service would not need to know of the trigger context. Thoughts?

npsp-reedestockton: I'd like us to meet to discuss this offline.  I'm concerned that we are about to initiate a major change to fix an isolated bug.

force2b:
@npsp-reedestockton FYI: Build failures. https://sfdo-metaci.herokuapp.com/builds/278804/tests

npsp-reedestockton:
@force2b @RadSFDeveloper This PR is ready for review.  Thanks!

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -34,6 +34,18 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>InstallmentOppFirstCreateAsync__c</fullName>
```
RadSFDeveloper, line 38: `InstallmentOppCreateFirstOption__c` since async is just one of the options?

src/classes/RD2_Constants.cls
```diff
@@ -53,6 +53,15 @@ public inherited sharing class RD2_Constants {
         Disable_All_Installments
     }
 
+    /**
+    * @description Enum to specify asynchronous behavior for the creation of the first installment opportunity
+    */
+    public enum FirstInstallmentAsyncOptions {
```
RadSFDeveloper, line 59: `FirstInstallmentOppCreateOptions` (or `InstallmentOppCreateFirstOptions` ) since async is just one of the options.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -382,14 +382,42 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+        }
+
+        if (settings.firstInstallmentAsync == RD2_Constants.FirstInstallmentAsyncOptions.Asynchronous_Always.name()
+            && (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs())
+        ) {
+            return false;
+        }
+
+        if (settings.firstInstallmentAsync == RD2_Constants.FirstInstallmentAsyncOptions.Asynchronous_When_Bulk.name()
+            && System.isBatch()
```
force2b, line 412: Rather than just `isBatch` I was thinking when `trigger.new.size()` > 1. This way it catches the scenario where an insert of multiple RD's happens even from outside of a batch job.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -382,14 +382,42 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
         RD2_QueueableService.EvaluateInstallmentOpportunities service =
             new RD2_QueueableService.EvaluateInstallmentOpportunities(rdIds, isScheduleChangedById);
 
-        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs()) {
-            System.enqueueJob(service);
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert && evaluateOppsSynchronously()) {
```
RadSFDeveloper, line 385: Previously the queueable limit was being checked on both insert and update, but currently it is being checked for the insert only.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -382,14 +382,42 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+            System.enqueueJob(service);
         }
     }
 
+    /***
+    * @description Determines whether Opportunities should be evaluated synchronously or asynchronously
+    * @return Boolean
+    */
+    @TestVisible
+    private Boolean evaluateOppsSynchronously() {
```
force2b, line 398: Maybe `shouldEvaluateOppsSynchronouslyOnInsert()`?

src/classes/RD2_Settings.cls
```diff
@@ -121,6 +121,24 @@ public with sharing class RD2_Settings {
         private set;
     }
 
+    /***
+    * @description Installment Opp Auto Creation Option,
```
RadSFDeveloper, line 125: The `@description` should be updated for this method.

src/classes/RD2_Constants.cls
```diff
@@ -53,6 +53,15 @@ public inherited sharing class RD2_Constants {
...
+        Synchronous_Always,
```
force2b, line 60: To match the help text on the field, this should be:

- Synchronous
- Asynchronous
- Asynchronous_When_Bulk

src/classes/RD2_OpportunityEvaluationService_TEST.cls
```diff
@@ -289,9 +289,9 @@ private class RD2_OpportunityEvaluationService_TEST {
      */
     private static npe03__Recurring_Donation__c setupDataAndCloseRD(String closeAction) {
         RD2_Settings_TEST.setUpConfiguration(new Map<String, Object> {
-            'npe03__Open_Opportunity_Behavior__c' => closeAction
+            'npe03__Open_Opportunity_Behavior__c' => closeAction,
+            'InstallmentOppFirstCreateAsync__c' => RD2_Constants.FirstInstallmentAsyncOptions.Asynchronous_Always.name()
```
force2b, line 293:
```suggestion
            'InstallmentOppFirstCreateMode__c' => RD2_Constants.FirstInstallmentAsyncOptions.Asynchronous.name()
```

src/classes/RD2_Settings.cls
```diff
@@ -121,6 +121,24 @@ public with sharing class RD2_Settings {
...
+    * If random string is input, the setting will be default to Always Create Next installment
+    */
+    @TestVisible
+    public String firstInstallmentAsync {
```
RadSFDeveloper, line 129: `firstInstallmentOppCreateOption` or `installmentOppCreateFirstOption` since async is one of the modes.

force2b: What do you think of `firstInstallmentCreateMode`?

src/classes/RD2_Settings_TEST.cls
```diff
@@ -354,4 +354,10 @@ public with sharing class RD2_Settings_TEST {
         UTIL_CustomSettingsFacade.reloadCustomSettingsOnRecordTypeCleanup();
         UTIL_CustomSettingsFacade.getRecurringDonationsSettingsForTest(rdSetting);
     }
+
+    public static void insertFirstRDAsynchronously() {
```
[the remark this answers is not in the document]
force2b: What about `configureForASyncFirstOpportunityCreate()`?

src/objects/npe03__Recurring_Donations_Settings__c.object
force2b: Or `InstallmentOppFirstCreateMode__c`. Either way

npsp-reedestockton:
@force2b @RadSFDeveloper Code review feedback addressed. Could you review again please? Thanks!

Julian88Tex:
@npsp-reedestockton @RadSFDeveloper @force2b remember me!?

I'm down to QE this as soon as it's merged :-)

rponti-sforg:
@codexmarcel Thank you Marcel.  Have a great weekend!

PR #5668, opened 2020-07-15
Gift Entry: List View Sort Bug Fix

voduyemi:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geListView/geListView.js
+++ b/src/lwc/geListView/geListView.js
@@ -10,7 +10,7 @@ import {
     format,
     deepClone,
     isNotEmpty,
-    showToast,
+    showToast
 } from 'c/utilCommon'
 import { fireEvent } from 'c/pubsubNoPageRef'
 import LibsMoment from 'c/libsMoment';
@@ -625,9 +625,7 @@ export default class geListView extends LightningElement {
                 }
             });
         });
-
-        this.records = records;
-
+        this.records = this.sortRows(records);
         return records;
     }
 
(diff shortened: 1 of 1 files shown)


src/lwc/geListView/geListView.js
```diff
@@ -315,15 +316,17 @@ export default class geListView extends LightningElement {
...
+                if (columnEntry && !this._orderBy) {
                     if (hasNestedProperty(this.columnEntriesByName[this.sortedBy],
                         'typeAttributes', 'label', 'fieldName')) {
                         orderedByFieldApiName = columnEntry.typeAttributes.label.fieldName;
                     } else {
                         orderedByFieldApiName = columnEntry.fieldApiName;
                     }
-                    orderBy = `${orderedByFieldApiName} ${this.sortedDirection}`;
+                } else {
+                    orderedByFieldApiName = this._orderBy;
```
kenlewis-sfdo, line 327: Should this be an inner if block?  Like
if (columnEntry) {
  if (!this._orderBy) {
  } else {
  }
}
Would columnEntry ever be false?

voduyemi: It can be in an inner block, `columnEntry` will never be false, but `_orderBy` can be `undefined`. I thought it would be cleaner if we combined both if statements. If it affects readability we can change it.

voduyemi:
@kenlewis-sfdo Thanks for catching that, we probably want to preserve the sort order even if we retrieve more records. I just pushed up that update. That way it's up to the user to reset the sort order if they want to.

PR #5669, opened 2020-07-15
Prevent user from creating or editing an Enhanced Recurring Donation if metadata has not been deployed

npsp-reedestockton:
# Critical Changes

# Changes
- When Enhanced Recurring Donations is enabled, a validation has been added to ensure that the Unpackaged Metadata required for Enhanced Recurring Donations has been deployed, such as picklist value, label, helptext and layout changes. This addresses an edge case where some customers have only partially completed the Enhanced Recurring Donations enablement process, leaving their org in a state where data may be partially invalid.

# Issues Closed

# Community Ideas Delivered

# New Metadata

Custom Labels: 
- RD2_EnablementIncomplete

# Deleted Metadata


npsp-reedestockton:
#W-7807220

npsp-reedestockton:
@force2b @RadSFDeveloper This is ready for your review.

@smarovich Would you please review the PR change description and the label?

Thank you all!!

src/labels/CustomLabels.labels
```diff
@@ -1753,6 +1753,14 @@
         <shortDescription>Enablement section warning</shortDescription>
         <value>&lt;b&gt;WARNING: NPSP doesn't provide a way to revert back to legacy Recurring Donations.&lt;/b&gt;</value>
     </labels>
+    <labels>
+        <fullName>RD2_EnablementIncomplete</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Enhanced Recurring Donations enablement incomplete</shortDescription>
+        <value>You cannot create or update Recurring Donations because you haven't finished enabling Enhanced Recurring Donations. Go to NPSP Settings | Recurring Donations | Upgrade to Enhanced Recurring Donations and complete all enablement steps.</value>
```
force2b, line 1762:
```suggestion
        <value>You cannot create or update Recurring Donations because you haven&apos;t finished enabling Enhanced Recurring Donations. Go to NPSP Settings | Recurring Donations | Upgrade to Enhanced Recurring Donations and complete all enablement steps.</value>
```

src/classes/RD2_EnablementService.cls
```diff
@@ -147,4 +147,28 @@ public with sharing class RD2_EnablementService {
...
+
+            if (String.isBlank(jsonState)) {
+                state.isMetaConfirmed = false;
+            }
+        } catch (Exception e) {
+            state.isMetaConfirmed = false;
+        }
+
+        return state.isMetaConfirmed == true;
+    }
```
force2b, line 173: Slight revision to cache the state in a static var so that each subsequent call to the method does not need to retrieve and de-serialize the result.
```suggestion
    private static RD2_EnablementDelegate_CTRL.EnablementState enablementState;
    public static Boolean isMetaDataDeployed() {
        
        if (enablementState != null) {
            return enablementState.isMetaConfirmed == true;
        }

        enablementState = new RD2_EnablementDelegate_CTRL.EnablementState();
        enablementState.isMetaConfirmed = false;

        try {
            String jsonState = UTIL_CustomSettingsFacade.getRecurringDonationsSettings().RecurringDonations2EnablementState__c;

            if (String.isNotBlank(jsonState)) {
                enablementState = (RD2_EnablementDelegate_CTRL.EnablementState) JSON.deserialize(jsonState, RD2_EnablementDelegate_CTRL.EnablementState.class);
            }
        } catch (Exception e) { }

        return enablementState.isMetaConfirmed == true;
    }
```

src/classes/RD2_EnablementService_TEST.cls
```diff
@@ -367,6 +367,7 @@ public with sharing class RD2_EnablementService_TEST {
         RD2_EnablementService.enablementService = (RD2_EnablementService) Test.createStub(
             RD2_EnablementService.class, new Stub().withRecurringDonations2Enabled()
         );
+        RD2_ValidationService.isRD2MetadataDeployed = true;
```
force2b, line 370: Nice!

src/classes/RD2_ValidationService.cls
```diff
@@ -196,6 +212,7 @@ public with sharing class RD2_ValidationService {
                 continue;
             }
 
+            validateRD2MetadataDeployed(rd);
```
force2b, line 215: Do you think it's necessary to validate on a change too? I can see where this might catch RD's inserted between the time the org first enabled RD2 and when this change hits customer orgs, but (at least given the data we have now), it doesn't appears that like these customers have a lot of data if any. Mainly I'm thinking about it from a performance perspective.

src/classes/RD2_EnablementService.cls
npsp-reedestockton: @force2b The only place this is called is RD2_ValidationService, and the below is how that is implemented.

```
    public static Boolean isRD2MetadataDeployed {
        get {
            if (isRD2MetadataDeployed == null) {
                isRD2MetadataDeployed = RD2_EnablementService.isMetaDataDeployed();
            }
            return isRD2MetadataDeployed;
        }
        set;
    }

So aren't we already caching?
```

src/classes/RD2_ValidationService.cls
npsp-reedestockton: I'm agnostic about validating on change.  I suppose someone could get into a state where we needed to validate on update by doing a bulk load of inserts with triggers off, then re-enabling triggers for interactive access.

src/classes/RD2_EnablementService.cls
```diff
@@ -147,4 +147,27 @@ public with sharing class RD2_EnablementService {
 
         return true;
     }
+
+    /**
+     * @description Evaluates whether Enhanced Recurring Donation metadata has been deployed
+     * @return Boolean
+     */
+    private static RD2_EnablementDelegate_CTRL.EnablementState enablementState;
+    public static Boolean isMetaDataDeployed() {
```
RadSFDeveloper, line 156: This method does more than it should, ie it retrieves the enablement state and checks it for the metadata deployment. It can be split into the following so it is clear when the enablement state is retrieved and what the method should be doing. Btw, the method is replaced with the property so it can be set for the validation.

```
    /**
     * @description Evaluates whether Enhanced Recurring Donation metadata has been deployed
     * @return Boolean
     */
    public static Boolean isMetaDataDeployed {
        get {
            if (isMetaDataDeployed == null) {
                isMetaDataDeployed = (getEnablementState().isMetaConfirmed == true);
            }
            return isMetaDataDeployed;
        }
        set;
    }

    /**
     * @description Retrieves RD2 enablement state from custom settings
     * @return RD2_EnablementDelegate_CTRL.EnablementState
     */
    private static RD2_EnablementDelegate_CTRL.EnablementState getEnablementState() {
        RD2_EnablementDelegate_CTRL.EnablementState enablementState = new RD2_EnablementDelegate_CTRL.EnablementState();
        enablementState.isMetaConfirmed = false;

        try {
            String jsonState = UTIL_CustomSettingsFacade.getRecurringDonationsSettings().RecurringDonations2EnablementState__c;
            if (String.isNotBlank(jsonState)) {
                enablementState = (RD2_EnablementDelegate_CTRL.EnablementState) JSON.deserialize(jsonState, RD2_EnablementDelegate_CTRL.EnablementState.class);
            }
        } catch (Exception e) { }

        return enablementState;
    }
```

src/classes/RD2_DataMigrationBase_BATCH.cls
```diff
@@ -152,6 +152,7 @@ public abstract class RD2_DataMigrationBase_BATCH implements Database.Batchable<
         //enable enhanced Recurring Donations during the batch transaction
         if (!RD2_EnablementService.isRecurringDonations2Enabled && isDryRunMode) {
             RD2_EnablementService.isRecurringDonations2EnabledOverride = true;
+            RD2_ValidationService.isRD2MetadataDeployed = true;
```
RadSFDeveloper, line 155: The validation service domain is not the enablement, thus, based on the recommended changes in the enablement service below, this can be replaced with `RD2_EnablementService.isMetaDataDeployed = true;`.
Btw, nice catch that validation needs to succeed before actual enablement occurs.

src/classes/RD2_EnablementService_TEST.cls
RadSFDeveloper: `RD2_EnablementService.isMetaDataDeployed = true;`

src/classes/RD2_ValidationService.cls
```diff
@@ -36,6 +36,19 @@
 */
 public with sharing class RD2_ValidationService {
 
+    /***
+    * @description Status of Enhanced Recurring Donations metadata deployment
+    */
+    public static Boolean isRD2MetadataDeployed {
```
RadSFDeveloper, line 42: This property can be removed.

src/classes/RD2_ValidationService.cls
```diff
@@ -209,6 +225,17 @@ public with sharing class RD2_ValidationService {
         }
     }
 
+    /***
+    * @description Validates Enhanced Recurring Donations metadata has been deployed
+    * @param rd The Recurring Donation to validate
+    * @return void
+    */
+    private void validateRD2MetadataDeployed(npe03__Recurring_Donation__c rd) {
+        if (!isRD2MetadataDeployed) {
```
RadSFDeveloper, line 234: `if (!isRD2MetadataDeployed) {` -> `if (!RD2_EnablementService.isMetaDataDeployed) {`

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -113,6 +113,57 @@ private with sharing class RD2_ValidationService_TEST {
...
+    private static void shouldFailWhenMetadataNotDeployedOnInsert() {
+        String errMessage = '';
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+
+        npe03__Recurring_Donation__c rd = getRecurringDonationBuilder()
+            .withContact(getContact().Id)
+            .withInstallmentPeriodYearly()
+            .build();
+
+        RD2_ValidationService.isRD2MetadataDeployed = false;
```
RadSFDeveloper, line 129: `RD2_EnablementService.isMetaDataDeployed = false;`

src/classes/RD2_ValidationService_TEST.cls
```diff
@@ -113,6 +113,57 @@ private with sharing class RD2_ValidationService_TEST {
...
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+
+        npe03__Recurring_Donation__c rd = getRecurringDonationBuilder()
+            .withContact(getContact().Id)
+            .withInstallmentPeriodYearly()
+            .build();
+
+        insert rd;
+
+        RD2_ValidationService.isRD2MetadataDeployed = false;
```
RadSFDeveloper, line 155: `RD2_EnablementService.isMetaDataDeployed = false;`

src/classes/RD2_EnablementService.cls
force2b: Nice abstraction

src/classes/RD2_ValidationService.cls
force2b: @RadSFDeveloper What do you think?

npsp-reedestockton: I'm less agnostic about this now.  If it is possible, given the current enablement flow, to get into a state where RDs can hit an update, we need to do this at least until the enablement flow is altered.  And we already know there have been records created, so we already know some records meet this test.  We need to validate on update.

npsp-reedestockton:
@force2b @RadSFDeveloper Code review feedback incorporated.  Please review.  Thanks!

src/classes/RD2_ValidationService.cls
force2b: We can leave this for now, but maybe this gets added to the "refactor ideas" doc as a reminder that it should be pulled out at some point in the future once we get the enablement process under control.

PR #5681, opened 2020-07-20
Updated NPSP Translations

dmkud:
# Critical Changes

# Changes

- Updated Translations for all supported languages:
  - Field Labels
  - Field Help Text
  - Object Labels
  - Picklist Labels
  - Custom Buttons and Links
  - Custom Labels
  - Tab Labels
  - Packaged Page Layout Section Titles

# Issues Closed

# New Metadata

# Deleted Metadata

# Unaggregated Pull Requests
* LQA fixes for summer20 drop1 [[PR5707](https://github.com/SalesforceFoundation/NPSP/pull/5707)]


src/objectTranslations/Address__c-nl_NL.objectTranslation.orig
```diff
@@ -0,0 +1,1096 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b, line 1: @dmkud I can strip these out of the PR, but I thought you should know that they were included in the last commit.

dmkud: Noted.

unpackaged/config/trial/standardValueSetTranslations/ContactRole-nl_NL.standardValueSetTranslation
```diff
@@ -0,0 +1,43 @@
...
+        <masterLabel>Other</masterLabel>
+        <translation>Overige</translation>
+    </valueTranslation>
+    <valueTranslation>
+        <masterLabel>Matched Donor</masterLabel>
+        <translation>Gekoppelde donor</translation>
+    </valueTranslation>
+    <valueTranslation>
+        <masterLabel>Soft Credit</masterLabel>
+        <translation>"Soft credit"</translation>
```
force2b, line 29: @dmkud Can you confirm this is correct? I see something similar (without the quotes) on the German translation. The other languages all have translations for this value.

localization/translation-cleanup.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
force2b, line 1: @codexmarcel Are you OK with me adding a new folder to the root of the repository to contain files unique to the localization process? I'll add a readme.md file too.

force2b: @codexmarcel To take this whole script a bit further, if we can add a new scratch org definition to the repo that has sampleData disabled, then I can drop the first couple of lines in this bash file that require JQ.

src/objectTranslations/Payment_Services_Configuration__c-es.objectTranslation
```diff
@@ -2,39 +2,39 @@
...
     <layouts>
         <layout>Payment Configuration Layout</layout>
         <sections>
-            <label><!-- Custom Links --></label>
+            <label>Vínculos personalizados</label>
             <section>Custom Links</section>
         </sections>
     </layouts>
-    <nameFieldLabel><!-- Payment Services Configuration Name --></nameFieldLabel>
+    <nameFieldLabel>Nombre de configuración de Servicios de pago</nameFieldLabel>
```
force2b, line 39: @dmkud This is being truncated down to 40 characters - "Nombre de configuración de Servicios de ". It might be necessary to change the translation to be a max of 40.

src/objectTranslations/Payment_Services_Configuration__c-fr.objectTranslation
```diff
@@ -2,40 +2,40 @@
...
         <layout>Payment Configuration Layout</layout>
         <sections>
-            <label><!-- Custom Links --></label>
+            <label>Liens personnalisés</label>
             <section>Custom Links</section>
         </sections>
     </layouts>
-    <nameFieldLabel><!-- Payment Services Configuration Name --></nameFieldLabel>
-    <startsWith><!-- Consonant --></startsWith>
+    <nameFieldLabel>Nom de configuration des services de paiement</nameFieldLabel>
```
force2b, line 39: @dmkud This is being truncated down to 40 characters - "Nom de configuration des services de pai". It might be necessary to change the translation to be a max of 40.

unpackaged/config/trial/standardValueSetTranslations/ContactRole-nl_NL.standardValueSetTranslation
dmkud: This seems to be expected, because there are quotes in other Dutch labels mentioning Soft Credit.

localization/translation-cleanup.sh
codexmarcel: @force2b new folder in root with a readme is just fine, that's tidy. The less dependencies the better, so let's drop those lines and we'll get `hasSampleData` flipped _false_. Standby...

force2b: Great!. I seem to remember that @davidmreed had some concerns with dropping hasSampleData in the org definitions, but a new scratch org type just for localization would work if we can't modify the existing one.

Also, this pattern (the scratch org def and this script file) will be repeated in the five underlying repositories.

src/objectTranslations/Payment_Services_Configuration__c-es.objectTranslation
dmkud: I committed the changed labels. Is it possible to add 40 character limit attribute to those labels?
Also, that's exactly what could have been caught using pseudo localization.

src/objectTranslations/Payment_Services_Configuration__c-fr.objectTranslation
dmkud: Fixed

force2b:
Linked To:
* [W-7852547](https://gus.lightning.force.com/a07B0000008HdJvIAK)
* [W-7850752](https://gus.lightning.force.com/a07B0000008TBwfIAG)

unpackaged/post/first/translations/de.translation
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b, line 1: @codexmarcel Because the new round of translations include translations in the `unpackaged/config/post` folder, this may create problems if this is deployed into an org without translation workbench enabled. For example, the "beta", "trial", "release" and many other scratch org definitions do not have translation workbench enabled. As a result, running `deploy_post` task fails against these orgs.

I'm a bit uncertain on where else this task may be used. If it's pretty limited to just our own configuration we may be able to modify the scratch org definitions. However, if this is used in many other areas, it may be necessary to remove these translations from the `post/first` folder. I suppose they could be included in the main trial org configuration since the trial config is always deployed after `post/first`.
cc @davidmreed

force2b: Since the pre and post folders both had translations in master, I'm re-running the install_beta task one more time from this branch to make sure that it still fails.

davidmreed: The biggest issue I see here is that `unpackaged/post` is also deployed in _both_ NPSP installers - the primary customer-facing one and the one for Trailhead Playgrounds. I don't think it is safe to make the assumption that those orgs have Translation Workbench turned on.

We don't currently have a preflight check for Translation Workbench. If we did (I don't know at present how to detect it!) we could break out these translations into a separate task and conditionally deploy them if and only if Translation Workbench were turned on.

What's the objective for these translations look like? Are they primarily supporting the TSO, or were we aiming to deliver them via MetaDeploy as well?

force2b: Good questions. Pre and Post were included in the translations because they both have unpackaged metadata. In particular, the `pre/account_record_types/` folder has the Household Account and Organization record types. It is odd that the pre folders already had the translation files. The only change in that folder structure is that the customObjectTranslation block was added to the package.xml.

I just confirmed that `install_beta` fails on the  `unpackaged/post/first` deployment step due to Translation Workbench not being installed. I suspect that the metadata api allows the deployment of the object translations in the `unpackaged/pre` folder, but is failing on the global (quickAction) translations in the post folder. That seems to be the only difference. I think the best solution - at least for now - is to just remove these two translations from the post config folder. They're very minor, and we've never had them in orgs.

force2b:
Work Items: 
- [W-7852547](https://gus.lightning.force.com/a07B0000008HdJvIAK)
- [W-7850752](https://gus.lightning.force.com/a07B0000008TBwfIAG)

PR #5682, opened 2020-07-20
RD2: Create and edit Pause on Recurring Donation

RadSFDeveloper:
# Critical Changes

# Changes
- In Enhanced Recurring Donations, you can now pause a Recurring Donation. Pausing a Recurring Donation means that you can honor a donor's request to temporarily suspend their donations for a specified period of time. To enable pause functionality, you must assign the `npsp.RD2_PauseForm_CTRL` to user profiles that manage Recurring Donations. Follow these instructions:
  - Go to Setup | Profiles | select the profile | Apex Class Access | Edit | select `npsp.RD2_PauseForm_CTRL` and move into the Enabled Apex Classes | Save.

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Apex class:
	- RD2_PauseForm_CTRL

- Aura Component:
	- RD2_PauseForm

- Lightning Web Component:
	- rd2PauseForm

- Custom Field:
	- RecurringDonationSchedule__c.IsPause__c
	- RecurringDonationSchedule__c.StatusReason__c

- Custom Labels:
	- commonInsufficientPermissions
	- commonSelectOptionPlaceholder
	- RD2_InstallmentStatusSkipped
	- RD2_PauseDeactivationSuccessMessage
	- RD2_PausePermissionRequired
	- RD2_PauseSaveSuccessMessage
	- RD2_PauseSelectedInstallmentTextPlural
	- RD2_PauseSelectedInstallmentTextSingular
	- RD2_PausedReasonLabel

- Quick Action:
	- npe03__Recurring_Donation__c.Pause

# Deleted Metadata


RadSFDeveloper:
Tracking:
- [W-7780227](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHqyIAG/view)
- [W-7780259](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHrUIAW/view)
- [W-7780257](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHrSIAW/view)

RadSFDeveloper:
@mpusto , @clairerowlett : Can you please verify the Pause page behaviour and let me know if there is anything that needs to be changed? Please note there is outstanding pending work that will be worked on today, however, I hope it is not relevant for the early work feedback.
Pending work:
- Highlight selected installments in blue on Pause dialog 
- Highlight select installments in gray and set font to italic on Upcoming Installments component
- Refresh the Upcoming Installments when a Pause is saved/deactivated

@force2b , @npsp-reedestockton : the following is pending for the back-end:
- Test class for `RD2_PauseForm_CTRL`
- Method and params description

src/labels/CustomLabels.labels
```diff
@@ -2158,6 +2158,14 @@
         <protected>true</protected>
         <shortDescription>Validate Recurring Donations data if it is valid install period</shortDescription>
         <value>{0} is not a supported value for Installment Period. Select Daily, Weekly, 1st and 15th, Monthly, or Yearly.</value>
+    </labels>
+     <labels>
+        <fullName>RD2_InstallmentStatusSkipped</fullName>
```
RadSFDeveloper, line 2163: @sam-knox , can you please verify content for seven new custom labels?

src/quickActions/npe03__Recurring_Donation__c.Pause.quickAction
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<QuickAction xmlns="http://soap.sforce.com/2006/04/metadata">
+    <description>Enables pausing of an enhanced Recurring Donation record</description>
```
RadSFDeveloper, line 3: @sam-knox , can you please check the description of the [Pause] button?

src/objects/RecurringDonationSchedule__c.object
```diff
@@ -441,6 +441,29 @@
             </valueSetDefinition>
         </valueSet>
     </fields>
+    <fields>
+        <fullName>IsPause__c</fullName>
```
RadSFDeveloper, line 445: @sam-knox , can you please verify the description and help text for two new fields on RD Schedule?

src/labels/CustomLabels.labels
RadSFDeveloper: @sam-knox , here are screenshots containing labels:
Screen Shot 2020-07-20 at 5 58 37 PM

Screen Shot 2020-07-20 at 5 56 10 PM

Upcoming Installments component displaying skipped Installments:
Screen Shot 2020-07-20 at 5 56 40 PM

When user selects an installment, specifies Paused Reason and clicks [Save]:
Screen Shot 2020-07-20 at 5 57 13 PM

When user deselects all skipped installments and saves, the pause will be deactivated:
Screen Shot 2020-07-20 at 6 15 33 PM

Page displayed when the RD is closed preventing pause setup:
Screen Shot 2020-07-20 at 6 19 26 PM

Here is the screenshot displaying the error message when a user does not have permission on the Apex class. Note this is not a label but wanted to let you know how the page will look like.
Screen Shot 2020-07-20 at 5 52 37 PM

src/lwc/rd2PauseForm/rd2PauseForm.html
```diff
@@ -0,0 +1,105 @@
+<template>
```
RadSFDeveloper, line 1: @rjanjanam , can you please verify elements on the page can be used for automation tests? Thanks.

RadSFDeveloper:
@sam-knox , labels and descriptions have been updated per your feedback. Can you please review again?

src/lwc/rd2PauseForm/rd2PauseForm.html
rjanjanam: Can see the data-qa-locators and access them. Looks good. Thanks @RadSFDeveloper

src/classes/RD2_PauseForm_CTRL.cls
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        return jsonData;
+    }
+
+    /**
+     * @description 
+     * @return 
+     */
+    public static Boolean isClosed(Id rdId) {
```
force2b, line 80: What is this method used for?

src/classes/RD2_PauseForm_CTRL.cls
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * @description Returns the Paused Reason value and label select options
+     * @return List<Option>
+     */
+    private static List<Option> getPausedReasonOptions() {
+        List<Option> options = new List<Option>();
+        List<Schema.PicklistEntry> entries = npe03__Recurring_Donation__c.ClosedReason__c.getDescribe().getPicklistValues();
+
+        for (Schema.PicklistEntry entry : entries) {
```
force2b, line 135: I think this can be replaced by a call to `UTIL_Describe.getSelectOptions()`, or could this just be handled in JavaScript?

src/classes/RD2_PauseForm_CTRL.cls
```diff
@@ -0,0 +1,270 @@
...
+    public static void savePause(String jsonPauseData) {
+        try {
+            PauseData pause = (PauseData) JSON.deserialize(jsonPauseData, PauseData.class);
+
+            pauseHandler.cancelPauseSchedule(pause.rdId);
+
+            //create new pause schedule
+            RecurringDonationSchedule__c schedule = buildPauseSchedule(pause);
+            if (schedule.StartDate__c != null && schedule.EndDate__c != null) {
+                insert schedule;
```
force2b, line 183: The part I'm a little unsure is whether a standard User has insert permissions into a protected object

unpackaged/config/rd2_post_config/profiles/Admin.profile
```diff
@@ -108,4 +112,14 @@
         <field>npe03__Recurring_Donation__c.npe03__Total__c</field>
         <readable>true</readable>
     </fieldPermissions>
+    <fieldPermissions>
```
force2b, line 115: Since the `RecurringDonationSchedule__c` is protected, it's not possible (or needed) to grant permissions to any fields on this object in unpackaged metadata. These 2 (and the 2 on the Standard profile) will need to be removed.

src/classes/RD2_PauseForm_CTRL.cls
```diff
@@ -0,0 +1,270 @@
...
+        return dataByKey;
+    }
+
+    /**
+     * @description 
+     * @param 
+     * @return void
+     */
+    @AuraEnabled
+    public static void savePause(String jsonPauseData) {
```
npsp-reedestockton, line 174: Why isn't the logic for schedule handling delegated to RD2_ScheduleService?

src/classes/RD2_PauseForm_CTRL.cls
RadSFDeveloper: The `UTIL_Describe.getSelectOptions()` contains a loop and since the `SelectOption` cannot be returned to the LWC (I am getting `Apex Type unsupported in JSON: System.SelectOption` error), so there should be another loop to convert options from the `SelectOption` to the (value, label) object list. The `getPicklistValues` can be used from the LWC, but then the page needs to check for one more Boolean if everything is loaded in addition to checking if has access, is closed, and is loading. The current implementation returns all Paused Reason data at once.
I am suggesting keeping the implementation as is unless there is a cleaner way to coordinate load and display of the Paused Reason.

src/classes/RD2_PauseForm_CTRL.cls
RadSFDeveloper: The `isClosed()` checks if the Recurring Donation is closed.

src/classes/RD2_PauseForm_CTRL.cls
RadSFDeveloper: Thank you! The existing pause deactivation (update) and new pause insert have been moved into the schedule service.

src/classes/RD2_PauseForm_CTRL.cls
RadSFDeveloper: The pause schedule update/insert has been moved to the schedule service after pause tests have been added.

RadSFDeveloper:
@force2b , @rponti-sforg 
> What happens if a User without Edit permissions to the RD object clicks the Pause button?
An Insufficient Permission error will be displayed and no RD data will be displayed.  

> What happens if a User with no Read permissions on the RD.Status, RD.Amount or some schedule related RD fields?
> What happens if a User without Edit permissions to the RD.Status field attempts to skip one or more installments? Are there other fields that might be considered required on the RD record?

When the user has create and edit access on the RD SObject,  the same error as on Upcoming Installments will be displayed when user does not have read access on RD Amount, Day of Month and Payment Method fields. No RD data will be displayed. 

When the above fields are readable but the RD "Status Reason" (`ClosedReason__c`) is not readable, then the Paused Reason picklist values are still displayed and I think they should. Let me know if this is not correct.

When a user does not have the Read permission, they will not be able to view the RD detail page, so the [Pause] button will not be accessible. However, the change has been made in the controller to check for Read, Create and Edit permissions on the RD object when determining the access.

src/labels/CustomLabels.labels
```diff
@@ -2216,12 +2232,20 @@
         <value>Select installments from the list below. You can pause for up to 12 consecutive installments.</value>
     </labels>
     <labels>
-        <fullName>RD2_PauseHeader</fullName>
+        <fullName>RD2_PausePermissionRequired</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>User does not have a permission to edit or create Recurring Donations</shortDescription>
+        <value>You don&apos;t have permissions to Pause Recurring Donations. Please contact your system administrator for create and edit access on Recurring Donation.</value>
```
RadSFDeveloper, line 2240: @sam-knox , can you please verify the new custom label text displayed when user does not have read, edit and create permissions on the Recurring Donation SObject? Also, there is another new `lblInsufficientPermissions` and I added a "."  for the `RD2_PauseSaveSuccessMessage` and `RD2_PauseDeactivationSuccessMessage`.

Screen Shot 2020-07-23 at 12 50 19 PM

force2b:
@RadSFDeveloper FYI - build error
`MetadataParseError: Could not process MDAPI response: Update of ApexClass RD2_PauseForm_CTRL: Error on line 128, col 9: Illegal assignment from List<System.SelectOption> to List<RD2_PauseForm_CTRL.Option>`

src/classes/RD2_PauseForm_CTRL.cls
force2b: Understood. Makes me wonder if we eventually need a utility method for getPicklistValuesForLwc(object, field) that returns the wrapper class. Though even in that model there is no support for the _default_ value. That's probably not an issue here, but it's something to consider if we eventually want to move to a utility method.

RadSFDeveloper:
> @RadSFDeveloper FYI - build error
> `MetadataParseError: Could not process MDAPI response: Update of ApexClass RD2_PauseForm_CTRL: Error on line 128, col 9: Illegal assignment from List<System.SelectOption> to List<RD2_PauseForm_CTRL.Option>`

@force2b , @rponti-sforg , sorry about that. I thought all SelectOption changes have been reverted. The latest commit should fix it.

src/classes/RD2_ScheduleService.cls
```diff
@@ -909,9 +865,242 @@ public without sharing class RD2_ScheduleService {
...
+        * @description Deactivates an active Pause Schedule (if any) and 
+        * saves the specified Pause Schedule.
+        * @param rdId Recurring Donation Id
+        * @param pauseSchedule The Pause Schedule on the Recurring Donation
+        * @return void
+        */
+        public void savePauseSchedule(Id rdId, RecurringDonationSchedule__c pauseSchedule) {
+            Savepoint sp = Database.setSavepoint();
+            try {
+                cancelPauseSchedule(rdId);
```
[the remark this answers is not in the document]
RadSFDeveloper: The Enhanced RD page layout can be changed to add Recurring Donation Schedule related list (add fields) on an unmanaged scratch org and all schedules can be verified.  @force2b , @rponti-sforg , @npsp-reedestockton 

- The `getPauseSchedule()` returns current/future pause schedule. It is assumed at most one active pause schedule can be present. If today is 7/28, the latest pause should be returned.
- When a new pause is saved, a current/future pause (if any) is deactivated, meaning its schedule End Date is set to yesterday. The existing pause schedule is not deleted.

Deactivation and saving a new pause schedule does work when I test in UI. Is there a malfunction in the above behaviour?

src/classes/RD2_ScheduleService.cls
```diff
@@ -909,9 +865,242 @@ public without sharing class RD2_ScheduleService {
...
+
+        /***
+        * @description Indicates if the schedule is active starting from the reference date
+        * @param schedule Recurring Donation schedule
+        * @param referenceDate Reference Date
+        * @return Boolean
+        */
+        private Boolean isWithinDateRange(RecurringDonationSchedule__c schedule, Date referenceDate) {
+            return schedule.EndDate__c == null
+                || (referenceDate <= schedule.EndDate__c && schedule.StartDate__c <= schedule.EndDate__c);
```
force2b, line 972: @RadSFDeveloper The `if` condition here has the following condition, which I would assume is true 100% of the time.
`schedule.StartDate__c <= schedule.EndDate__c`

The way I read this method is:
- If EndDate is null, return true
- If EndDate is not null and the referenceDate is before or equal to the EndDate, return true (because StartDate must always be before or equal to EndDate).
- If EndDate is not null and the referenceDate is after the EndDate, return false.

Could this impact how the method is used?

RadSFDeveloper: If the End Date is in past, the schedule is not returned regardless of the Start Date.
If the End Date is today or in future *and* the Start Date <= End Date, the Pause Schedule is a valid schedule and it should be returned/handled. So the above conclusion is correct, let me know if it does not work as expected.

When testing and code reviewing, change Enhanced RD page layout to add RD Schedule related list and verify schedules upon creation, and deactivation. Also, you can change the Start Date or even End Date manually to simulate current or past valid schedules. Past valid  and deactivated schedules should not be used in any schedule and installment processing. Current and future valid schedules should be handled. Current and future invalid (inactive) schedules should not be handled. Also, there is an assumption that at most one active Pause Schedule can exist.

FYI @rponti-sforg

RadSFDeveloper: Also, when the RD is closed, any active schedule including the Pause Schedule is deactivated.
FYI @force2b , @npsp-reedestockton , @rponti-sforg

src/classes/RD2_PauseForm_CTRL.cls
```diff
@@ -0,0 +1,287 @@
...
+                SELECT Status__c
+                FROM npe03__Recurring_Donation__c
+                WHERE Id = :rdId
+                LIMIT 1
+            ];
+
+            isClosed = (new RD2_RecurringDonation(rd)).isClosed();
+
+        } catch (Exception e) {
+            //the record has been deleted;
```
force2b, line 115: Though unlikely, is it worth throwing an AuraHandledException here, even if it just returns `e.getMessages()[0]`?

src/classes/RD2_PauseForm_TEST.cls
```diff
@@ -0,0 +1,504 @@
...
+			RD2_PauseForm_CTRL.PauseData pause = getPauseData(rd.Id);
+			System.assertEquals(false, pause.hasAccess, 'The user should not have access: ' + pause);
+			System.assertEquals(null, pause.isRDClosed, 'The Recurring Donation closed status should not be specified');
+			System.assertEquals(null, pause.startDate, 'The Pause Start Date should not be set');
+			System.assertEquals(null, pause.resumeAfterDate, 'The Pause Resume After Date should not be set');
+			System.assertEquals(null, pause.pausedReason, 'The Paused Reason should not be initialized');
+		}
+	}
+
+	/****
```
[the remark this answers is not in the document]
force2b: @RadSFDeveloper Most of this file is using tabs for indentation instead of spaces.

src/classes/RD2_PauseForm_TEST.cls
```diff
@@ -0,0 +1,504 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Recurring Donations
+* @description Unit tests for the Pause Recurring Donation feature
+*
+*/
+@IsTest(IsParallel=false)
```
force2b, line 37: Did you find a case where this was not working with parallel set to true?

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -1714,6 +1701,385 @@ private with sharing class RD2_ScheduleService_TEST {
...
+        schedules.add(pause);
+
+        RD2_ScheduleService.currentDate = pause.StartDate__c.addDays(-1);
+
+        List<RD2_ScheduleService.Installment> installments = service.getVisualizedInstallments(
+            RD2_ScheduleService.currentDate, NUM_INSTALLMENTS, schedules
+        );
+        System.assertEquals(NUM_INSTALLMENTS, installments.size(), 'Number of installments should match: ' + installments);
+
+		for (Integer i = 0, pauseCounter = 4; i < pauseCounter; i++) {
```
force2b, line 1859: There are sporadic uses of tabs for indentations in this file as well

src/lwc/rd2PauseForm/rd2PauseForm.js
```diff
@@ -0,0 +1,421 @@
...
+    maxRowSelection = 12;
+    selectedIds = [];
+    @track selectedRowsSummary = null;
+    @track columns = [];
+    @track installments;
+
+    @track error = {};
+
+    /***
+    * @description
```
force2b, line 65: Most of the functions in this file do not have a description. Are those coming later?

src/lwc/rd2PauseForm/rd2PauseForm.js
```diff
@@ -0,0 +1,421 @@
...
+    * @description 
+    */
+    connectedCallback() {
+        this.init();
+    }
+
+    /***
+    * @description Group various calls to Apex
+    */
+    init = async () => {
```
force2b, line 74: I'm intrigued by the use of `async()` in some of the functionality here. How is this used to improve page functionality and/or is it worth writing up a document on when/why we should use `async` in LWC controllers?

src/lwc/rd2PauseForm/rd2PauseForm.html
force2b: I do find the use of _both_ a `data-id` and `data-qa-locator` that are so very similar, yet slightly different, to be confusing. Is there value in having two different `data` attributes used for similar purposes, but in different areas? For example, the JS controller can just as easily query for elements by `data-qa-locator` as it can by `data-id`. If there is a desire to keep these as two completely different `data` attributes, it's probably worth spelling this out in a shared document somewhere to define how each are should be used by developers.

src/lwc/rd2PauseForm/rd2PauseForm.js
RadSFDeveloper: I did not measure response time, but when not using `async()`, the form took a bit longer to load. Also, waiting for the PauseData ensures the page rendering completes and is useable (otherwise we need to have flags for the load complete (form ready)). The answer to this question and `data-id` vs `data-qa-locator` might require team discussion. I can start discussion offline. In meantime, the form works in various scenarios and the xpath can be determined by the automation, and if there is any change (which I doubt) based on the discussion, we can change it before merging complete Pause functionality into the `master`.

force2b: Thanks for clarifying. I'm effectively wondering when we should consider using `async` and when we shouldn't.

src/classes/RD2_PauseForm_TEST.cls
force2b: Thanks. I was wondering if we should consider moving tests that create/modify Users into a separate test class (ex: `RD2_PauseForm_TEST2` that does not run parallel and then leave the main class as parallel=true. I'm not really sure it would help with build performance, but it's a thought.

src/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -551,16 +551,9 @@ public inherited sharing class RD2_OpportunityEvaluationService {
...
-            'InstallmentFrequency__c, ' +
-            'InstallmentPeriod__c, ' +
-            'DayOfMonth__c, ' +
-            'PaymentMethod__c, ' +
-            'StartDate__c, ' +
-            'EndDate__c ' +
-            'FROM RecurringDonationSchedules__r)';
+        Set<String> fields = new RD2_ScheduleService.ScheduleQueryHandler().getFields();
+
+        return '(SELECT ' + String.join(new List<String>(fields), ', ') + ' FROM RecurringDonationSchedules__r)';
```
npsp-reedestockton, line 556: great!!

src/classes/RD2_ScheduleService.cls
```diff
@@ -909,9 +865,242 @@ public without sharing class RD2_ScheduleService {
...
+            recurringDonationId = rdSchedule.RecurringDonation__c;
+            scheduleId = rdSchedule.Id;
+            startDate = rdSchedule.StartDate__c;
+        }
+    }
+
+    /***
+    * @description Handles Pause Recurring Donation functionality
+    */
+    public class PauseScheduleHandler {
```
npsp-reedestockton, line 939: I love the way you have encapsulated this functionality!!!

src/labels/CustomLabels.labels
sam-knox: Are there other places where we're checking for permissions that the `lblInsufficientPermissions` header could be used? Seems like this is a generic label that could be applied across the product wherever we are doing permission checks.

RadSFDeveloper: The search in codebase does not return any part of code where "Insufficient Permissions" or `Insufficient*` is used so no other code is changed. However, it is a generic label and can be used in future WIs.

sam-knox: @melissabarber ^^

sam-knox: @RadSFDeveloper Gift Entry uses a similar label, `commonAdminPermissionErrorTitle`, "Check Your Permissions".

For this one, maybe we could change the label name to `commonInsuffientPermissions`

The Gift Entry team has started to use the "common" prefix for anything that could be used generically. There's several other labels like that in customLabels.

@force2b

melissabarber: We have commonAdminPermissionErrorTitle "Check Your Permissions" but as Sam pointed out in our side conversations, users can't really check their permissions unless they're admins and "Insufficient Permissions" makes more sense. @ptewson-sfdo we might want to make this part of HoneyGems' cleanup work when our two teams nail down more consistent permission errors messaging.

RadSFDeveloper: @sam-knox , @melissabarber : the label is renamed to `commonInsufficientPermissions`. Thank you for your feedback.

RadSFDeveloper: Similarly, the new `lblSelectOptionPlaceholder` containing "Select Option" value is renamed to `commonSelectOptionPlaceholder`.

sam-knox: Thanks great, thanks @RadSFDeveloper!

rponti-sforg:
@RadSFDeveloper @force2b It looks like the Robot test completed successfully by looking at this: https://sfdo-metaci.herokuapp.com/builds/289576/flows and that 73 of 73 tests have passed.  On the PR page, it appears to be still running ... What do you think?  Move forward with merge?

force2b:
@rponti-sforg This is merging into a parent branch vs. master, plus looking at the history, the last 4 robot builds in this branch all passed.. I recommend merging.

PR #5686, opened 2020-07-21
GE: Fix for Batch Processing bug

voduyemi:
Fixed issue with namespaced donation matching rule defaults in non-namespaced orgs

Not needed in release notes. We don’t encourage folks to use our code non-packaged.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geBatchWizard/geBatchWizard.js
+++ b/src/lwc/geBatchWizard/geBatchWizard.js
@@ -37,6 +37,7 @@ const ID = 'id';
 const MAX_STEPS = 2;
 const CANCEL = 'cancel';
 const COMMA_SEPARATOR = ';';
+const NAMESPACE_PREFIX = 'npsp__';
 
 export default class geBatchWizard extends NavigationMixin(LightningElement) {
 
@@ -380,7 +381,7 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
             const matchingRuleFields = donationMatchingRule.split(COMMA_SEPARATOR);
             matchingRuleFields.forEach(field => {
                 strippedFields += stripNamespace(field,
-                  namespace + '__') + COMMA_SEPARATOR;
+                  NAMESPACE_PREFIX) + COMMA_SEPARATOR;
             });
             dataImportBatch.fields[
               DATA_IMPORT_BATCH_DONATION_MATCHING_RULE.fieldApiName] =
--- a/src/lwc/geBatchWizard/geBatchWizard.js
+++ b/src/lwc/geBatchWizard/geBatchWizard.js
@@ -15,7 +15,7 @@ import {
     getNamespace,
     getNestedProperty,
     isNull,
-    stripNameSpace
+    stripNamespace
 } from 'c/utilCommon'
 import GeLabelService from 'c/geLabelService';
 
@@ -379,7 +379,8 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
             let strippedFields = '';
             const matchingRuleFields = donationMatchingRule.split(COMMA_SEPARATOR);
             matchingRuleFields.forEach(field => {
-                strippedFields += stripNameSpace(field) + COMMA_SEPARATOR;
+                strippedFields += stripNamespace(field,
+                  namespace + '__') + COMMA_SEPARATOR;
             });
             dataImportBatch.fields[
               DATA_IMPORT_BATCH_DONATION_MATCHING_RULE.fieldApiName] =
--- a/src/lwc/geBatchWizard/geBatchWizard.js
+++ b/src/lwc/geBatchWizard/geBatchWizard.js
@@ -37,7 +37,6 @@ const ID = 'id';
 const MAX_STEPS = 2;
 const CANCEL = 'cancel';
 const COMMA_SEPARATOR = ';';
-const NAMESPACE_PREFIX = 'npsp__';
 
 export default class geBatchWizard extends NavigationMixin(LightningElement) {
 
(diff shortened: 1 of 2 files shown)


voduyemi:
Link to the WI : https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EIXlIAO/view

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -352,6 +359,33 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
+        return dataImportBatch;
+    }
+
+    /***************************************************************************
+     * @description Strips namespace prefix from the Data Import Batch donation matching
+     * rule fields if in a namespaced context
+     * @param dataImportBatch
+     * @returns {object}
+     */
+    stripNameSpacedDonationMatchingRuleFields (dataImportBatch) {
```
psadie-sfdo, line 373: One final question. What do you think about renaming this to something like below?
```suggestion
    stripNamespaceFromDonationMatchingRuleFields (dataImportBatch) {
```

ptewson-sfdo:
@voduyemi I'm just now setting up a non-namespaced org to test this out, but I'm curious if the original issue with processing batches was just through Gift Entry or if the namespace issue on the matching rule fields would apply to Batches created through old BGE or batches that weren't attached to one of our gift entry products? 

I'm just noticing that the namespace updates are all happening front-end here and want to make sure we aren't still leaving an issue with other paths for processing batches?

voduyemi:
@ptewson-sfdo I made the changes in the front-end because that was the only place I could find where we were creating a new Data Import Batch record based on the defaults, In old BGE, we actually set the donation matching rule in the VF page controller `BDI_BatchOverride_CTRL`:  ` if (diBatch.Donation_Matching_Rule__c == null) {
            diBatch.Donation_Matching_Rule__c = UTIL_Namespace.StrTokenNSPrefix('Donation_Amount__c') + ';' +
                UTIL_Namespace.StrTokenNSPrefix('Donation_Date__c');
        }`
Also before we process the batch we always check for the data import setting on the batch which in this case will be the right matching rule. I believe the only way for us to get into that weird namespace state is if we manually update the donation matching rule on the batch.

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -352,6 +359,33 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
+    stripNamespaceFromDonationMatchingRuleFields (dataImportBatch) {
+        const donationMatchingRule = dataImportBatch.fields[
+          DATA_IMPORT_BATCH_DONATION_MATCHING_RULE.fieldApiName];
+        const namespace = getNamespace(DATA_IMPORT_BATCH_INFO.objectApiName);
+        
+        if (isNull(namespace)) {
+            let strippedFields = '';
+            const matchingRuleFields = donationMatchingRule.split(COMMA_SEPARATOR);
+            matchingRuleFields.forEach(field => {
+                strippedFields += stripNamespace(field,
```
ptewson-sfdo, line 382: I'm a bit confused here - since we are inside the `if (isNull(namespace))` block here, isn't the value of `namespace` always going to be null?  And as a result, is the call to stripNamespace() going to actually do anything? 

I did some console logging in a non-namespaced scratch org and got these values: 
- Namespace: null
- Matching rule fields: npsp__Donation_Amount__c,npsp__Donation_Date__c
- Stripped Fields: npsp__Donation_Amount__c;npsp__Donation_Date__c

Running through the steps in the WI, The batch seems to process successfully, but it looks like the value for the donation matching rule isn't actually being changed at all. Am I missing something? 

@voduyemi @psadie-sfdo

voduyemi: @ptewson-sfdo You're right, I was referencing an already `null` namespace variable, I should be passing in the `npsp__` string. It's fixed now.

voduyemi: @ptewson-sfdo The batches will still get process as long as we're not trying to match donations (triggering the review donations modal). That's the reason why you were still able to process/dry run batches.

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -25,11 +30,14 @@ import DATA_IMPORT_BATCH_GIFT_INFO from '@salesforce/schema/DataImportBatch__c.G
...
 import DATA_IMPORT_MATCHING_BEHAVIOR_INFO from '@salesforce/schema/DataImportBatch__c.Donation_Matching_Behavior__c';
 import DATA_IMPORT_BATCH_TABLE_COLUMNS_FIELD from '@salesforce/schema/DataImportBatch__c.Batch_Table_Columns__c';
+import DATA_IMPORT_BATCH_DONATION_MATCHING_RULE from '@salesforce/schema/DataImportBatch__c.Donation_Matching_Rule__c';
 
 const NAME = 'name';
 const ID = 'id';
 const MAX_STEPS = 2;
 const CANCEL = 'cancel';
+const COMMA_SEPARATOR = ';';
+const NAMESPACE_PREFIX = 'npsp__';
```
ptewson-sfdo, line 40: @voduyemi If there's any way we can avoid hard-coding the namespace here that would be great ... I realize this update is following my feedback on the previous iteration of this work, but I would love see if we can avoid the hard-coded reference here. @psadie-sfdo is this a case that we've addressed in the past, any advice here?

voduyemi: @ptewson-sfdo I think the problem here is in a non-namespaced org, I'm not sure we have a way of retrieving the namespace prefix programatically and we need this namespace prefix specifically in a non-namespaced org. I'm open to suggestions on how to it can be achieved if anyone knows how.

PR #5688, opened 2020-07-23
Set hasSampleData to `false` for all scratch org definitions

codexmarcel:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


skristem:
@codexmarcel I noticed a failing Robot test on this PR and that seems to be a valid one. It looks like for Accounts object it is missing My Accounts view option.

PR #5707, opened 2020-08-03
LQA fixes for summer20 drop1

dmkud:
LQA - fixed translations for summer20 drop1


src/objectTranslations/npe03__Recurring_Donation__c-nl_NL.objectTranslation
```diff
@@ -245,15 +245,15 @@
         </picklistValues>
     </fields>
     <quickActions>
-        <label>Terugkerende donatie sluiten</label>
+        <label><!-- Close Recurring Donation --></label>
```
force2b, line 248: @dmkud Not sure why these translations were removed from dutch. They weren't removed for any of the other languages on this object.

dmkud: Looks like the translations for these labels are coming from another file, which was not included in PR. I will commit it now as well.

PR #5712, opened 2020-08-04
Update Opportunity Naming sharing from "with sharing" to "inherited sharing"

### LINKED ISSUE #5533: Opty Naming Class: "Attempt to de-reference a null object." error when Site Guest User creates an Opportunity
**Issue:** When Padlock Site Guest User security settings are enabled and Default Site Guest User record owner option is used for force.com sites, the NPSP Classes built as "with sharing" are failing. Query's run as the Site Guest User that need to locate a record owned by another User (due to default enabled) cannot find the record as Sharing Rules fire after Triggers.

**Steps to Reproduce** (in an org that has NPSP and an Opty Naming rule configured in NPSP Settings):
1. Enable Sites and create a Site URL to create a Site Guest User
2. Enable Padlock ("Secure guest user record access" on Sharing Settings page in Setup)
3. Enable "Assign new records created by Salesforce Sites guest users to a default owner in the org" setting on Sites page. Define a default on the site itself. 
4. Use the Site URL to create an Opportunity record (for example as part of an ISV partner integration), and note the error. While the error is received, the Opty record should be created as expected. The NPSP Opty naming class "Opp_OpportunityNaming" will have failed to run, preventing the name of the new Opty from being set per the naming rule.
5. Delete the Opty Naming Rule in NPSP Settings or disable the default setting and test again. Note there is no error when the Opty is created.

_For further details see internal GUS Investigation W-7544682_

**Cause:** The NPSP's Opty Naming class Opp_OpportunityNaming is setup as "With Sharing", which prevents the Site Guest User running the naming class from locating the newly created Opty during it's SOQL Query. 
**Potential Solution:** Modify the Opty Naming class Opp_OpportunityNaming to use "Inherited Sharing". Initial testing of this solution has been successful, however additional risk assessment and testing is needed.

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed
fixes #5533 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/OPP_OpportunityNaming.cls
+++ b/src/classes/OPP_OpportunityNaming.cls
@@ -35,7 +35,7 @@
 * @description Main logic class for opportunity naming
 */
 //TODO: finish implementation of interface so other implementing classes can be used. this will require making classes global and a heirarchical setting that we don't yet have. TBD if this is necessary.
-public with sharing class OPP_OpportunityNaming implements OPP_INaming {
+public inherited sharing class OPP_OpportunityNaming implements OPP_INaming {
 
     private static final SObjectType recordTypeSObjectType = UTIL_Describe.getSObjectType('RecordType');
 


psadie-sfdo:
[Tracking W-7646294](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B000000894knIAA/view)

PR #5730, opened 2020-08-11
Gift Entry: Update payment integration service upsert to insert and update

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


psadie-sfdo:
[Tracking W-7849944](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008TBE7IAO/view)

src/classes/PS_IntegrationService.cls
```diff
@@ -54,33 +54,143 @@ global class PS_IntegrationService {
...
+    * @param serviceKeys A set of Payment_Services_Configuration__c.Service_Key__c values
+    */
+    private static Payment_Services_Configuration__c[] getExistingRecords(Set<String> serviceKeys) {
+        return [
+            SELECT Id,
+                Service_Key__c,
+                Key__c,
+                Value__c
+            FROM Payment_Services_Configuration__c
+            WHERE Service_Key__c = :serviceKeys
```
vmartinov-sforg, line 131: This works, but could we use 'IN' instead of '=' ?

src/classes/PS_IntegrationService.cls
```diff
@@ -54,33 +54,143 @@ global class PS_IntegrationService {
...
+    /*******************************************************************************************************
+    * @description Applies the existing configuration record ids to the corresponding newly generated
+    * configuration in preparation for an update.
+    *
+    * @param existingConfigurations List of existing Payment_Services_Configuration__c records
+    * @param configurationsToUpsertByServiceKey Map of newly generated Payment_Services_Configuration__c records
+    * from the parameters provided by the caller of 'setConfiguration'.
+    */
+    private static void applyExistingIdsToNewConfigurations(
+        Payment_Services_Configuration__c[] existingConfigurations,
```
vmartinov-sforg, line 144: Even though Arrays and Lists are almost the same in apex, I would consider using List<> instead of [] for consistency. We are using `List<Payment_Services_Configuration__c>` in code above.
This comment applies to some of the methods below.

src/classes/PS_IntegrationService.cls
```diff
@@ -54,33 +54,143 @@ global class PS_IntegrationService {
...
+    * @description Collects a set of Service_Key__c fields from a list of Payment_Services_Configuration__c
+    * records
+    *
+    * @param existingConfigurations List of existing Payment_Services_Configuration__c records
+    */
+    private static Set<String> collectServiceKeys(Payment_Services_Configuration__c[] existingConfigurations) {
+        Set<String> existingConfigurationKeys = new Set<String> {};
+
+        for (Payment_Services_Configuration__c configuration : existingConfigurations) {
+            existingConfigurationKeys.add(configuration.Service_Key__c);
```
vmartinov-sforg, line 164: Sets allow `null` to be added, should we check for null before we add the value to the Set?

src/classes/PS_IntegrationService.cls
```diff
@@ -54,33 +54,145 @@ global class PS_IntegrationService {
...
+    }
+
+    /*******************************************************************************************************
+    * @description Queries for existing Payment_Services_Configuration__c records
+    *
+    * @param serviceKeys A set of Payment_Services_Configuration__c.Service_Key__c values
+    */
+    private static List<Payment_Services_Configuration__c> getExistingRecords(Set<String> serviceKeys) {
+        return [
+            SELECT Id,
```
kenlewis-sfdo, line 126: For future reference - Id is always returned in SOQL queries so doesn't need to be explicitly named in the field set (I forget this often :)

ptewson-sfdo: It doesn't hurt to include it though - I usually do just because ... I guess I feel anxious if it isn't in the field list even though I know it's always included

PR #5737, opened 2020-08-11
RD2: Display the first donation date after the pause ends

RadSFDeveloper:
As a user I want to see when the next donation date will occur after the Pause so that I can be sure I am setting up the right schedule.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/rd2PauseForm/rd2PauseForm.html
```diff
@@ -66,15 +66,21 @@
                     <lightning-layout multiple-rows="true">
 
                         <lightning-layout-item padding="horizontal-large" class="slds-p-top_large" size="12">
-                            <template if:true={selectedRowsSummary}>
+                            <div if:true={selectedRowsSummary} style="font-style: italic;">
```
force2b, line 69: I've read that for jest tests to work properly, `if:true` and `if:false` should only be used with the `<template>` tag. So, here it might be better to leave the template as the outer wrapper and add an inner wrapper of div or span with the style tag.

src/lwc/rd2PauseForm/rd2PauseForm.js
```diff
@@ -276,6 +300,33 @@ export default class Rd2PauseForm extends LightningElement {
...
+
+            if (lastSelectedId < this.installments.length) {
+                firstDateAfterPause = this.installments[lastSelectedId].donationDate;
+
+            } else if (lastSelectedId == this.installments.length) {
+                firstDateAfterPause = this.firstDonationDateBoundary;
+            }
+
+            if (firstDateAfterPause !== null) {
+                this.firstDonationDateMessage = this.labels.firstDonationDateMessage.replace('{0}', firstDateAfterPause);
```
force2b, line 325: How does this handle date formatting in other locales (such as yyyy-mm-dd, etc.)?

RadSFDeveloper:
Tracking: [W-7780258](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EHrTIAW/view)

src/lwc/rd2PauseForm/rd2PauseForm.html
RadSFDeveloper: Did not know that, thank you, the change has been done.

src/lwc/rd2PauseForm/rd2PauseForm.js
RadSFDeveloper: Excellent point, the fix has been applied.

force2b:
@RadSFDeveloper Thanks for making the changes. I'm not if the date format change is working though. In this org, I changed the users Locale to "English (United Kingdom)", but the component still shows a date in the wrong format.
image

RadSFDeveloper:
The date formatting has been fixed. Should have used locale instead of the lang. Can you please verify the changes again @force2b ?

src/lwc/rd2PauseForm/rd2PauseForm.html
```diff
@@ -67,14 +68,25 @@
 
                         <lightning-layout-item padding="horizontal-large" class="slds-p-top_large" size="12">
                             <template if:true={selectedRowsSummary}>
-                                <lightning-formatted-rich-text value={selectedRowsSummary} data-id="rowSummary"
-                                    data-qa-locator="richtext Row Summary">
-                                </lightning-formatted-rich-text>
+                                <div style="font-style: italic;">
+                                    <lightning-formatted-rich-text value={selectedRowsSummary} data-id="rowSummary"
```
andrewyu-salesforce, line 72: Just a suggestion. Since `<lightning-formatted-rich-text>` is used, instead of having a `<div>` with in line style attribute. Maybe you can append `<em></em> `tag to `{selectedRowsSummary}`. Same for line 81 to 86. What do you think?

PR #5748, opened 2020-08-14
Make regression_org install beta without FLS changes

davidmreed:
This PR revises the regression_org flow to install the managed beta last, with PUSH install semantics - meaning no FLS changes upon install. 

NOTE: this will significantly change the behavior of regression orgs. You will not have FLS to any new fields or sObjects introduced since the previous production release. While this makes it possible to catch specific classes of user-facing FLS issues, it also means that your manual and Robot regressions may fail until FLS is granted.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


davidmreed:
@rponti-sforg It can be tested in this branch.

PR #5752, opened 2020-08-16
RD2: Update open Opp to the Next Donation Date on schedule change

### LINKED ISSUE #5750: Enhanced Recurring Donations: RD schedule update incorrectly updates open Opp Close Date
### Summary
When "Day of Month", "Installment Frequency" and/or "Installment Period" is updated on a Recurring Donation having a *closed* current/future Opportunity and an *open* Opportunity, then the open Opportunity Close Date is updated to the same date as the closed Opportunity. The schedule update on the RD in this conditions will cause duplicate Opportunities for the same Close Date.

### Steps to Repeat
**Scenario 1: RD record without an active Pause**
1. Create a Recurring Donation so that Day of Month is within 3 days from today. For example Day of Month = 15 given today is 8/14.
2. Close the open Opportunity with the Close Date 8/15/2020.
3. The Opportunity Close Date is automatically set to today and a new open Opportunity for 9/15/2020 is automatically created.
4. Click [Edit] on the Recurring Donation page 
5. Change "Day of Month" to 3 and change "Effective Date" to 9/3/2020
6. Click [Save]
7. Verify the Next Donation Date is 9/3/2020 after the Recurring Donation page is manually refreshed
8. Verify the open Opportunity Close Date:

_Expected Result:_
- The open Opportunity Close Date is updated to 9/3/2020

_Actual Result:_
- The open Opportunity Close Date is updated to 8/15/2020

**Scenario 2: RD active with an active Pause**

1. Create a Recurring Donation so that Day of Month is within 3 days from today. For example Day of Month = 15 given today is 8/14.
2. Click [Pause] and select installments: 9/15 and 10/15
3. Click [Save]
4. Close the open Opportunity with the Close Date 8/15/2020.
5. The Opportunity Close Date is automatically set to today and a new open Opportunity for 11/15/2020 is automatically created.
6. Click [Pause] and *deselect* installment: 10/15
7. Click [Save]
8. Verify the Next Donation Date is 10/15/2020 after the Recurring Donation page is manually refreshed
9. Verify the open Opportunity Close Date:

_Expected Result:_
- The open Opportunity Close Date is updated to 10/15/2020

_Actual Result:_
- The open Opportunity Close Date is updated to 8/15/2020

### Workaround:
Manually update the Close Date on the open Opportunity to the expected Close Date.

RadSFDeveloper:
When "Day of Month", "Installment Frequency" and/or "Installment Period" is updated on a Recurring Donation having a closed current/future Opportunity and an open Opportunity, then the open Opportunity Close Date should be updated to the Next Donation Date value.

# Critical Changes

# Changes

# Issues Closed
- Fixes #5750 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_OpportunityService.cls
+++ b/src/classes/RD2_OpportunityService.cls
@@ -143,7 +143,9 @@ public with sharing class RD2_OpportunityService {
 
         Integer numberOfInstallments = 1;
         for (npe03__Recurring_Donation__c rd : rds) {
-            RD2_OpportunityMatcher.Record nextDonationDateRecord = scheduleService.getNextDonationDateRecord(rd, rd.RecurringDonationSchedules__r);
+            RD2_OpportunityMatcher.Record nextDonationDateRecord = scheduleService.getNextDonationDateRecord(
+                new RD2_RecurringDonation(rd), rd.RecurringDonationSchedules__r
+            );
 
             if (nextDonationDateRecord == null
                 || !nextDonationDateRecord.isNew()
@@ -256,11 +258,8 @@ public with sharing class RD2_OpportunityService {
             RD2_ScheduleService.Installment earliestUnmatchedInstallment;
 
             for (RD2_OpportunityMatcher.Record record : matcher.getRecords()) {
-                isOpenOppMatched = false;
 
-                if (record.hasInstallment() 
-                    && (record.isNew() || !record.getInstallment().isSkipped)
-                ) {
+                if (record.hasInstallment() && record.isNew()) {
                     earliestUnmatchedInstallment = getEarliestInstallment(record.getInstallment(), earliestUnmatchedInstallment);
                 }
 
@@ -269,7 +268,7 @@ public with sharing class RD2_OpportunityService {
                 }
 
                 Opportunity opp = record.getOpportunity();
-                if (record.hasInstallment() && !record.getInstallment().isSkipped) {
+                if (record.hasInstallment()) {
                     installmentByOppId.put(opp.Id, record.getInstallment());
                     isOpenOppMatched = true;
 
@@ -311,7 +310,7 @@ public with sharing class RD2_OpportunityService {
         Boolean isEarlier = earliestInstallment == null
             || currentInstallment.nextDonationDate < earliestInstallment.nextDonationDate;
 
-        if (isEarlier && !currentInstallment.isSkipped) {
+        if (isEarlier) {
             earliestInstallment = currentInstallment;
         }
 
--- a/src/classes/RD2_RecurringDonation.cls
+++ b/src/classes/RD2_RecurringDonation.cls
@@ -57,6 +57,11 @@ public inherited sharing class RD2_RecurringDonation {
     }
 
     /**
+    * @description Recurring Donation record current and future Opportunities
+    */
+    List<Opportunity> opportunities = new List<Opportunity>();
+
+    /**
     * @description A list of the fields that have been changed by one of the revise methods.
     */
     private Set<Schema.SObjectField> changedFields = new Set<Schema.SObjectField>();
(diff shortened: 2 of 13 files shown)


RadSFDeveloper:
Attached to: [W-7963522](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UDPOIA4/view)

src/classes/RD2_OpportunityService.cls
```diff
@@ -256,11 +258,8 @@ public with sharing class RD2_OpportunityService {
             RD2_ScheduleService.Installment earliestUnmatchedInstallment;
 
             for (RD2_OpportunityMatcher.Record record : matcher.getRecords()) {
-                isOpenOppMatched = false;
 
-                if (record.hasInstallment() 
-                    && (record.isNew() || !record.getInstallment().isSkipped)
-                ) {
+                if (record.hasInstallment() && record.isNew()) {
```
force2b, line 262: I noticed that in this method, the previous logic to check for a skipped installment has been removed. I can't find where that logic still executes elsewhere though. Is the premise here that you no longer check for a paused installment in the opp service - which doesn't seem right here.

RadSFDeveloper: That's a valid point and I planned to indicate more information about code changes, did not have time.
The `isSkipped` is not required to be checked since we get active installments only from `getVisualizedInstallments()`. The `isSkipped` installment checks bring complexity in the conditional statements and it is applicable only if we get all installment instead ie `getAllVisualizedInstallments()`. So instead of checking which conditional statement is causing the error, it was cleaner to remove all `isSkipped` checks in the method. Let me know if you suggest different approach.

src/classes/RD2_RecurringDonation.cls
```diff
@@ -66,7 +71,28 @@ public inherited sharing class RD2_RecurringDonation {
...
+
+    /**
+    * @description Constructor
+    * @param rd Recurring Donation record
+    * @param opportunities Recurring Donation current and future Opportunities
+    */
+    public RD2_RecurringDonation(npe03__Recurring_Donation__c rd, List<Opportunity> opportunities) {
         this.rd = rd;
+
+        this.opportunities = opportunities == null
```
RadSFDeveloper, line 85: When the Next Donation Date is determined in the RD2 trigger on update, then `npe03__Donations__r` list is not set. Thus, it is better to store Opportunities in a variable without requiring `npe03__Recurring_Donation__c`  to be rebuilt with `npe03__Donations__r`  populated.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -163,23 +163,25 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
+                    .revisePlannedInstallments()
+                    .reviseNextDonationDateBeforeInsert(scheduleService);
 
-            } else if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
-                rdRecord
+            } else if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {                
+                RD2_RecurringDonation rdRecord = new RD2_RecurringDonation(rd, oppsByRDId.get(rd.Id))
+                    .revisePlannedInstallments()
                     .reviseStatus(currentDate)
                     .reviseNextDonationDate(scheduleService, schedulesByRdId.get(rd.Id));
```
RadSFDeveloper, line 186: Even when an RD has a closed Opportunity for the first future installment, the Next Donation Date would be set to the first installment date in the RD2 trigger on update, since Opportunities were not retrieved for the NDD revision so the first installment was considered as open. However, the Opp eval service would overwrite the value relatively soon since the service is queued from the RD2 trigger. So now the Opportunities are fetched from `RD2_OpportunityEvaluationService().getOpportunitiesByRDId()`.

The validation service also fetches Opportunities, so we added one more SOQL in the transaction. However, I think it is better to keep responsibilities of fetching Opps separately.

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -418,10 +428,15 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
             for (npe03__Recurring_Donation__c rd : rds) {
-                if (evalService.hasKeyFieldChanged(rd, oldRdsById.get(rd.Id))) {
+                npe03__Recurring_Donation__c oldRd = oldRdsById.get(rd.Id);
+
+                if (evalService.hasKeyFieldChanged(rd, oldRd)) {
                     rdIds.add(rd.Id);
                 }
-                if (rd.npe03__Next_Payment_Date__c != oldRdsById.get(rd.Id).npe03__Next_Payment_Date__c) {
+
+                if (rd.npe03__Next_Payment_Date__c != oldRd.npe03__Next_Payment_Date__c
```
RadSFDeveloper, line 437: Adding the check for fields changing the schedule fixed the issue tested in [this test](https://github.com/SalesforceFoundation/NPSP/blob/5a7e6004051e595592f39c6a3d3d3fe542445535/src/classes/RD2_OpportunityService_TEST.cls#L619). If the Effective Date (`StartDate__c`) is not set, then test pass w/o the fix, however setting the Effective Date uncovers an error  that is fixed by retrieving Opportunities to calculate the NDD and then as a fail safe, check fields changing the schedule to see if Opportunities need to be evaluated.

src/classes/RD2_RecurringDonation.cls
npsp-reedestockton: I like this change a lot.  Makes the implementation much more general I think.

PR #5754, opened 2020-08-17
Accessibility: Update Visualforce pages to use slds and Lightning stylesheets

npsp-reedestockton:
# Critical Changes

# Changes
 - We switched to Lightning Stylesheets on our Visualforce pages in order to take advantage of accessibility improvements.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
Tracking [W-7780322](https://gus.lightning.force.com/a07B0000008EHsgIAG)

npsp-reedestockton:
@force2b @RadSFDeveloper This is ready for your review.  Thanks!

src/pages/LD_LeadConvertOverride.page
```diff
@@ -49,7 +30,7 @@
                 <div class="slds-p-horizontal_large">
 
                     <div class="slds-form-element slds-is-required">
-                        <c:UTIL_FormField sObjType="Contact" sObj="{!dummyCon}" field="OwnerId" labelOverride="{!$Label.leadConvertOwner}" />
+                        <c:UTIL_FormField sObjType="Contact" sObj="{!dummyCon}" field="OwnerId" labelOverride="{!$Label.leadConvertOwner}" overrideRequired="true"/>
```
RadSFDeveloper, line 33: Is `UTIL_FormField` component setting fields as required by default when using new `npsp-common.css` and lightning stylesheet?

src/pages/MTCH_FindGifts.page
```diff
@@ -22,17 +26,21 @@
             background-color: inherit;
             z-index: 100;
         }
+
+        .margin-left-zero {
```
RadSFDeveloper, line 30: Let's add the comment to clarify why this was added.

src/components/BDI_SettingsUI.component
```diff
@@ -254,4 +254,11 @@
         }, false);
     </script>
 
+    <script>
```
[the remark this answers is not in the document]
force2b: Definitely. It's effectively a hack to work-around an issue in the dueling stylesheets. It it worth moving this into a JS file in the static resource and then just loading it on the pages as
`<apex:includeScript value="{!URLFOR($Resource.CumulusStaticResources, 'RemoveDuplicateRequiredMark.js')}"/>`

src/pages/PMT_PaymentWizard.page
```diff
@@ -1,5 +1,9 @@
-<apex:page controller="PMT_PaymentWizard_CTRL" title="{!IF(isWriteoffWizard,$Label.pmtWizardSectionWriteoff,$Label.pmtWizardSubtitle)}"
-           tabStyle="Opportunity" standardStylesheets="true" docType="html-5.0">
+<apex:page controller="PMT_PaymentWizard_CTRL"
+    title="{!IF(isWriteoffWizard,$Label.pmtWizardSectionWriteoff,$Label.pmtWizardSubtitle)}"
+    tabStyle="Opportunity"
+    standardStylesheets="true"
```
RadSFDeveloper, line 4: @npsp-reedestockton , is there a reason to keep `standardStylesheets` as true?

src/pages/PSC_ManageSoftCredits.page
```diff
@@ -1,8 +1,13 @@
-<apex:page standardController="Opportunity" extensions="PSC_ManageSoftCredits_CTRL" showHeader="true" standardStylesheets="true" docType="html-5.0">
+<apex:page standardController="Opportunity"
+    extensions="PSC_ManageSoftCredits_CTRL"
```
RadSFDeveloper, line 2: The manage Soft Credits has Contact marked as required, which I think is correct since it cannot be empty. The field border is red on error while it is not on production. @mpusto  any objections?

Branch org:
Screen Shot 2020-08-18 at 4 56 21 PM


Main (production) org:
Screen Shot 2020-08-18 at 4 56 30 PM

force2b:
@npsp-reedestockton Looks like there is a build failure due to a markup problem on the BDI_BatchOverride page
https://sfdo-metaci.herokuapp.com/builds/298664

force2b:
@RadSFDeveloper On that last screen capture with the date picker differences, the two fields in the two screen caps are different. I would be surprised if that makes a difference, but can you confirm that the pickers are still different if you use the "Date of First Payment" field in both branches?

RadSFDeveloper:
> @RadSFDeveloper On that last screen capture with the date picker differences, the two fields in the two screen caps are different. I would be surprised if that makes a difference, but can you confirm that the pickers are still different if you use the "Date of First Payment" field in both branches?

This is the branch org date picker on the same field and it still differs, which is fine but wanted to double check. 
Screen Shot 2020-08-19 at 9 03 45 AM

src/pages/PSC_ManageSoftCredits.page
RadSFDeveloper: @rponti-sforg , the above will be different when compared to an org built of the `master`.

npsp-reedestockton:
@sam-knox Would you please review the Changes section of the PR?

npsp-reedestockton:
@rjanjanam There is one failing robot test here, but it appears to be BGE related rather than something about the work on this PR.

rjanjanam:
> @rjanjanam There is one failing robot test here, but it appears to be BGE related rather than something about the work on this PR.

@npsp-reedestockton I checked the failure. It is failing for some other reason for the batch process taking too long. Robot tests look good .

PR #5772, opened 2020-08-20
Gift Entry Accessibility: Ensure Tabs have a proper id/name 

bdvorachek:
GUS WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008TVgdIAG/view
- Fixed an issue where tabs in the template builder did not have a valid id attribute, which could cause issues with Accessibility.
# Critical Changes

# Changes
- We improved Gift Entry Template Builder so screen readers tell users the tab or step they're on.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -528,17 +528,13 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
...
         switch (tabValue) {
-            case this.TabEnums.INFO_TAB:
-                this.activeTab = this.TabEnums.INFO_TAB;
-                break;
-            case this.TabEnums.FORM_FIELDS_TAB:
-                this.activeTab = this.TabEnums.FORM_FIELDS_TAB;
-                break;
-            case this.TabEnums.BATCH_SETTINGS_TAB:
-                this.activeTab = this.TabEnums.BATCH_SETTINGS_TAB;
+            case this.tabs.INFO.id:
```
ptewson-sfdo, line 531: Does javascript have the switch behavior that every case falls through to the next one until you call break? So all three of these first three cases execute the "this.activeTab = tabValue;" line? 

I *might* prefer including the "this.activeTab = tabValue; break;" in every case just for clarify if that's so, but it's certainly not a big detail.

bdvorachek: Yes, it does do switch fall-through until a break is encountered. I'm leaning towards a comment to indicate fall-through is happening rather than writing out `this.activeTab = tabValue; break;` for each case. Would that work?

PR #5775, opened 2020-08-21
Calculate code coverage

davidmreed:
This PR sets a minimum NPSP code coverage level of 86% application-wide. Current coverage is 87%; this allows for any imprecision while ensuring we do not meaningfully drop code coverage.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


cumulusci.yml
```diff
@@ -547,6 +547,7 @@ tasks:
                 - "UNABLE_TO_LOCK_ROW"
                 - "connection was cancelled here"
             retry_always: True
+            required_org_code_coverage_percent: 75
```
davidmreed, line 550:
```suggestion
            required_org_code_coverage_percent: 86
```

force2b (approved):
Very nice. The only thing that would make it better is if the code coverage can be rendered in the PR UI somewhere. I realize that may require a whole other set of changes though.

cumulusci.yml
```diff
@@ -550,7 +550,9 @@ tasks:
                 - "unable to obtain exclusive access to this record"
                 - "UNABLE_TO_LOCK_ROW"
                 - "connection was cancelled here"
+                - "None"
```
codexmarcel, line 553: Ooof.

codexmarcel (approved):
`required_org_code_coverage_percent` is so satisfying.

PR #5777, opened 2020-08-22
W-7849631: Process RDs in BDI (3a of 5) - Update install script

grivero:
[W-7849631](https://gus.my.salesforce.com/a07B0000008WOQaIAO): Process RDs in BDI (3a of 5) - Update NPSP install script to push new object and field mappings

The code (aka installation script), creates migrated recurring_donation object and/or migrated fields dynamically only when they don't exist AND when all the original mapping references exist.  

Notes: https://salesforce.quip.com/wjGWAPKE6f0d

Only 4 classes should appear in Files changed tab:
- BDI_MigrationMappingUtility.cls
- BDI_MigrationMappingUtility_TEST.cls
- STG_InstallScript.cls
- STG_InstallScript_TEST.cls

How to Test this using Dev Console:
--
- Execute the following script in order to upsert migrated Recurring Donation object and migrated Recurring Donation fields.
  - use dev and dev_namespaced orgs 
- You can test deleting different fields mappings and executing the script, should create only the missing mappings.
- If all the migrated objects and fields are present, when running this script it should not do nothing.
- You can also test deleting all migrated field mappings for the migrated recurring-donation object so then you can delete the migrated object mapping and recreate the entire migrated-structure using the script:
```
List<String> newDefaultMappingStrings = new List<String>{          
    'Recurring_Donation_Amount',
    'Recurring_Donation_Date_Established',
    'Recurring_Donation_Day_of_Month',
    'Recurring_Donation_Effective_Date',
    'Recurring_Donation_Elevate_Recurring_Id',
    'Recurring_Donation_End_Date',
    'Recurring_Donation_Installment_Frequency',
    'Recurring_Donation_Installment_Period',
    'Recurring_Donation_Name',
    'Recurring_Donation_Payment_Method',
    'Recurring_Donation_Planned_Installments',
    'Recurring_Donation_Recurring_Type',
    'Recurring_Donation_Status',
    'Recurring_Donation_Status_Reason',
    'Recurring_Donation_Campaign'
};

BDI_MigrationMappingHelper helper = new BDI_MigrationMappingHelper();
BDI_MigrationMappingUtility util = new BDI_MigrationMappingUtility(helper);
util.migrateNewDefaultToCustomMetadata(newDefaultMappingStrings);
```

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


grivero:
@mldyang  -- Do you remember why we didn't include Payment_Check_Reference_Number and Payment_Method to the field mappings? maybe they already existed?  

I was doing some tests and I realized that we don't have those field names in the set of fields to migrate.

mldyang:
> @mldyang -- Do you remember why we didn't include Payment_Check_Reference_Number and Payment_Method to the field mappings? maybe they already existed?
> 
> I was doing some tests and I realized that we don't have those field names in the set of fields to migrate.

@grivero yes! Those are fields that already exist in NPSP, ie - we did not add them as part of our Payments work. Payment is already an object and those two are the fields that get mapped. Because of this, Advanced Mapping would map these fields for end-users so no need for our Post Install script to do it.

grivero:
Looks like I need to take another look into some unit tests. I'll do that asap

grivero:
@mldyang - I just sync this PR so you can see only my changes. 
I had to fix unit tests for STG_InstallScript_TEST, given that we can't hardcode the same set of fields that the stubbed class returns (Account_1_City__c, Account_1_Country__c).

I continue doing some manual testings on namespaced orgs and worked fine too 👍

src/classes/BDI_MigrationMappingUtility.cls
```diff
@@ -428,6 +465,44 @@ public without sharing class BDI_MigrationMappingUtility {
...
+            // create instance of object mapping
+            BDI_MigrationMappingUtility.DataImportObjectMapping diom = 
+                new BDI_MigrationMappingUtility.DataImportObjectMapping(rdObjectMapping, true);                
+
+            // get metadata object to be enqueued
+            Metadata.CustomMetadata objectMetadata = diom.getMetadataRecord();
+            // get unique migrated name using full name
+            String uniqueName = objectMetadata.fullName;
+            // split by '.' and get the last array node
+            uniqueName =  uniqueName.split('\\.')[uniqueName.split('\\.').size()-1] ;
```
mldyang, line 495: Just want to double check... we can't utilize the existing data import object mapping wrapper and therefore generateUniqueRecordName(), right?

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -504,9 +504,10 @@ public with sharing class STG_InstallScript_TEST {
                         new BDI_MigrationMappingHelperMock());
         BDI_MigrationMappingUtility util= new BDI_MigrationMappingUtility(mock_BDI_MigrationMappingHelper);
 
-        Test.testInstall(new STG_InstallScript(util), new Version(1,0), true);
-        System.assertEquals(2, util.queuedMetadataTypesForDeploy.size(),
-                'We have 2 metadata records queued for deployment.');
+        Test.testInstall(new STG_InstallScript(util), new Version(1,0), true);     
+        // we should have 2 field mappings queued for deployment   
+        System.assertEquals(2, util.queuedMetadataTypesForDeploy.size(), 
+            'We should have 2 mapping metadata records queued for deployment.');
```
mldyang, line 510: Is this the spot you're referring to "fix unit test in STG_InstallScript_TEST?"

Is it feasible to modify the stub class to include a new default object + new default field so we can see if both the object and the field are migrated? There are a few different stubs in BDI_MigrationMappingUtility_TEST that we might be able to reference.

src/classes/BDI_MigrationMappingUtility.cls
grivero: We can't 👎  The inner class attributes are defined without public attributes and without getter-setter so is not reachable. 
I didn't want to modify it given that is being used in several places.

src/classes/STG_InstallScript_TEST.cls
grivero: I tried that last-night, looks like a really big effort and I couldn't make it work. It's not just mocking some fields/object, we need to mock an entire BDI_MigrationMappingHelper class (including the methods). 
I also tried modifying the existent ones but it breaks a lot of unit tests so I end it up doing the same as you did previously, just re-used what is there.

PR #5808, opened 2020-09-04
Add Save & Activate Button to Batch Number Format Configuration Page

kenlewis-sfdo:
[W-7954555](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008U82fIAC/view)

# Critical Changes

# Changes
- Add Save & Activate button.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/labels/CustomLabels.labels
+++ b/src/labels/CustomLabels.labels
@@ -4985,12 +4985,12 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>Batch Number Formats</value>
     </labels>
     <labels>
-        <fullName>batchNumberSettingsSave</fullName>
+        <fullName>batchNumberSettingsActivate</fullName>
         <categories>Batch Number Settings</categories>
         <language>en_US</language>
         <protected>true</protected>
-        <shortDescription>Link to save a batch number format</shortDescription>
-        <value>Save Batch Number Format</value>
+        <shortDescription>Link to save and activate the new batch number format</shortDescription>
+        <value>Save &amp; Activate</value>
     </labels>
     <labels>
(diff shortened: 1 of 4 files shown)


melissabarber:
Label looks great to me, @kenlewis-sfdo! Though, I think we'll actually end up creating the label in a separate PR that @ptewson-sfdo is working on for label freeze.

psadie-sfdo (commented):
Looks good to me! My only question is around the automation locators. I'm thinking they might be too generic for the NPSP Settings page. Since everything is loaded upfront we could potentially have a scenario where multiple save buttons etc end up having the same locators in the future. _until we can refactor NPSP Settings_

src/labels/CustomLabels.labels
```diff
@@ -4985,12 +4985,12 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <value>Batch Number Formats</value>
     </labels>
     <labels>
-        <fullName>batchNumberSettingsSave</fullName>
```
ptewson-sfdo, line 4988: Are you intending to remove the "batchNumberSettingsSave" label and replace it with the "batchNumberSettingsActivate" label?

ptewson-sfdo (approved):
This looks good to me! Just going to verify that the branch deploys!

src/labels/CustomLabels.labels
kenlewis-sfdo: Yep, that was the intent - I just renamed it when I made this change end of last sprint.  Let me know if this will be ok or if you think it will cause any issues - I think this is the same way you have/had it in the other labels pr.

PR #5813, opened 2020-09-08
RD2: Display Elevate credit card widget

RadSFDeveloper:
Display Elevate credit card widget when a user creates a new Recurring Donation and select "Credit Card" as the payment method on an org connected to the Elevate payment services.

# Critical Changes

# Changes
The new `PS_ElevateService` and `UTIL_AuraEnabledCommon` Apex class should be added to required profiles' enabled classes.

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Apex Class:
    - PS_ElevateService
    - UTIL_AuraEnabledCommon
<br/>

- Lightning Web Component:
    - psElevateTokenHandler
    - rd2ElevateCreditCardForm
<br/>

- Custom Label:
    - RD2_ElevateDisableButtonLabel
    - RD2_ElevateDisabledMessage
    - RD2_ElevateEnableButtonLabel

# Deleted Metadata


RadSFDeveloper:
Tracking [W-8010251](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UhPQIA0/view)

src/classes/PS_ElevateService.cls
```diff
@@ -0,0 +1,94 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Elevate Payment Services
+* @description Contains Elevate payment services methods 
+*/
+public with sharing class PS_ElevateService {
```
RadSFDeveloper, line 36: There are common Payment Services methods that will be used by RD2 and probably other features, but currently they are contained in `GE` classes. I *copied* the common methods into a new Apex class. I have not extracted them from the `GE` classes, but I would suggest they are updated to use the common code in another WI. 

This new controller should be added to the required profiles' enabled Apex classes.

src/classes/PS_ElevateService.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
RadSFDeveloper, line 3: @force2b , should we be using API version 49 for new classes/LWCs?

src/classes/RD2_EntryFormController.cls
```diff
@@ -56,24 +56,32 @@ public with sharing class RD2_EntryFormController {
...
+    @AuraEnabled(cacheable=true)
     public static Map<String, Object> getRecurringSettings(Id parentId) {
         String parentSObjectType = (parentId == null)
             ? null
             : parentId.getSobjectType().getDescribe().getName();
 
+        // Checks whether the Elevate credit card can be displayed
+        // on the form to enter a new Recurring Donation in NPSP and the related commitment in Elevate.
+        // The RD2 controller should be using isIntegrationEnabled() only when GA.
+        Boolean isElevateCustomer = RD2_ElevateIntegrationService.isIntegrationEnabled()
```
RadSFDeveloper, line 74: The `PilotEnabled` feature parameter and its override on a sandbox will be used to enable Elevate credit card widget *in addition* to having the org connected to Elevate. When functionality becomes GA, the `PilotEnabled` check can be removed.

src/labels/CustomLabels.labels
```diff
@@ -1617,6 +1617,30 @@
         <shortDescription>Error when the functionality is not supported for Elevate Recurring Donation.</shortDescription>
         <value>This functionality is currently not supported for Recurring Donations generated from Elevate Payment Services.</value>
     </labels>
+    <labels>
+        <fullName>RD2_ElevateDisableButtonLabel</fullName>
```
RadSFDeveloper, line 1621: @sam-knox , can you please verify three new labels applicable to RD2 entry form explicitly?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1015,12 +1016,21 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
                     index: index
                 });
             });
+
+        } else if (errorResponse) {
```
RadSFDeveloper, line 1020: @ptewson-sfdo , the GE form error handling was changed to account to an error response not containing message and index, so that one error returned from the iframe error response can be displayed.

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -0,0 +1,168 @@
+
+import { fireEvent } from 'c/pubsubNoPageRef';
+import { getNamespace, isFunction, isNull } from 'c/utilCommon';
+
+import PAYMENT_AUTHORIZATION_TOKEN_FIELD from '@salesforce/schema/DataImport__c.Payment_Authorization_Token__c';
```
RadSFDeveloper, line 5: The namespace is verified by checking `DataImport__c.Payment_Authorization_Token__c`. This is valid even when the user does not have read access to the NPSP Data Import SObject or read access to the field. @rponti-sforg , please verify  when testing the feature for a user that is not an admin.

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -0,0 +1,168 @@
...
+/***
+* @description Max number of ms to wait for the response containing a token or an error
+*/
+const TOKENIZE_TIMEOUT_MS = 10000;
+
+
+/***
+* @description Payment services Elevate credit card tokenization service
+*/
+class psElevateTokenHandler {
```
RadSFDeveloper, line 28: @ptewson-sfdo , a new Elevate token handler was created for the common methods. I have not changed the GE form, but in the case the Visualforce origin handling is changed and this PR is not merged into the `master` yet, feel free to copy the class and use/update the common methods.

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -0,0 +1,168 @@
...
+            );
+
+            this.tokenCallback = message => {
+                clearTimeout(timer);
+
+                if (message.error) {
+                    reject(handleError(message));
+
+                } else if (message.token) {
+                    resolve(message.token);
```
RadSFDeveloper, line 137: @ptewson-sfdo , the resolve contains the token, no field setting in the new common token handler class.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
```diff
@@ -0,0 +1,67 @@
...
+                                    <h1 class='slds-text-body_regular' data-qa-locator='heading'>{labels.elevateWidgetLabel}</h1>
+                                </lightning-layout-item>
+                                <lightning-layout-item size='6'>
+                                    <template if:false={isDisabled}>
+                                        <lightning-button variant="base" label={labels.elevateDisableButtonLabel}
+                                            alternative-text={labels.elevateDisableButtonLabel}
+                                            icon-name='utility:close' icon-position="right"
+                                            onclick={handleUserDisabledWidget}
+                                            class='slds-button slds-button_base slds-float_right hide-card-button'
+                                            data-qa-locator='button Do Not Use Elevate'>
```
RadSFDeveloper, line 21: @rjanjanam , can you please verify the `data-qa-locators` make sense?

RadSFDeveloper (commented):
@ptewson-sfdo , @force2b , @rponti-sforg , @andrewyu-salesforce 

Adding card holder input fields (as placeholders, they are not used by changes in this WI) are left, but please can you start the review/testing in meantime and I will make additional changes (if any) based on the code review. Please note this PR will be merged into `feature/RD2` and not into `master` this sprint. I also assume the `feature/RD2` will not be merged into `master` in the next sprint, but this will be confirmed during the next sprint planning. The feature is protected from being displayed to users even when the org is the Elevate customer by utilizing the `PilotEnabled` feature parameter.

sam-knox (approved):

sam-knox:
@RadSFDeveloper - Per the notes in Changes, do all NPSP customers need to assign `PS_ElevateService` to profiles, or only customers who want to use Elevate?

RadSFDeveloper:
> @RadSFDeveloper - Per the notes in Changes, do all NPSP customers need to assign `PS_ElevateService` to profiles, or only customers who want to use Elevate?

Great question @sam-knox , the service call will be made only when the org is connected to Elevate.

RadSFDeveloper:
@rjanjanam , can you please verify [robot test failures](https://sfdo-metaci.herokuapp.com/builds/312242/tests) are not caused by this PR changes?  Thank you!

src/classes/PS_ElevateService.cls
force2b: @RadSFDeveloper Because this class queries the `Organization` object, which is a System object, I think the class should be defined as either `without` or `inherited sharing` to avoid the same issue we ran into with the ApexAsyncJob system object. I will note that other classes that query this object are mostly implied inherited. There is one class that uses `with sharing`, so I suspect it would work even if left as is.

src/classes/PS_ElevateService.cls-meta.xml
force2b: Yeah. I think we should

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
+    <isExposed>false</isExposed>
```
force2b, line 4: A description should be included on any new LWC's.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.css
```diff
@@ -0,0 +1,41 @@
...
+
+.credit-card-container {
+    border: 1px solid #E2E1E0;
+    border-radius: 0.25rem;
+    min-height: 5rem;
+    box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.1);
+}
+
+.credit-card-header {
+    background-color: #E2E1E0;
```
force2b, line 19: What if the User changes the theme in their org? Normally that would cause other colors to shift throughout the app, but here the background color is this section would remain the same even in an alternate theme. Is there a SLDS style that can be used to define this or a similar color so it's not hardcoded/fixed?

force2b (commented):
Looking very nice. I need to do a second review of this and test in an org, but wanted to get you my initial comments asap.

src/classes/PS_ElevateService.cls
ptewson-sfdo: Depending on timing, I wonder if we (Honeygems) might want to put in a PR as a child of this work to pivot to the new PS-prefixed variants. At the time we initially built this there wasn't enough intel to know if we were really going to share concepts like "isElevateCustomer" outside of Gift Entry, but since it is now clear that we will and are sharing these functions, I agree that this switch makes sense. But if we gather all that work into this PR we can solve for it now, rather than waiting until this work merges when we will inevitably be focused on other things.

ptewson-sfdo (approved):
This looks good from my side! Will want to coordinate on this side of things but verified that the Gift Entry form continues to work and that the error handling appears to be restored. Thanks!

src/classes/PS_ElevateService.cls
RadSFDeveloper: @patrick-yan-sf , feel free to create branches off the `feature/RD2` to incorporate work by Honeygems related to Elevate Payment Services based off the common classes. The `feature/RD2` will be main branch for Elevate commitment creation work.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.css
RadSFDeveloper: Excellent point! The change has been made to use [design tokens](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/tokens_standard_force_base.htm) and the fallback colors (in the case token rendering does not work).

RadSFDeveloper:
@force2b , changes have been applied. The `PS_ElevateService` has been changed to use `without sharing`.

src/classes/PS_ElevateService.cls
```diff
@@ -0,0 +1,94 @@
...
+* @description Contains Elevate payment services methods 
+*/
+public without sharing class PS_ElevateService {
+
+    /***
+    * @description Checks whether an org is a customer of Elevate Payment Services
+    * @return Boolean
+    */
+    @AuraEnabled
+    public static Boolean isElevateCustomer() {
```
force2b, line 43: Is this method used anywhere?

andrewyu-salesforce (approved):
@RadSFDeveloper The PR looks great to me. I like how you already have the implementation layout for the future WIs such as sending token and error handling.

src/classes/PS_ElevateService.cls
RadSFDeveloper: It is not used currently (it did at some part of the RD2 credit widget but decided to use `RD2_ElevateIntegrationService.isIntegrationEnabled()`), but most importantly it was created so Gift Entry code referencing common Elevate methods can start using the `PS_ElevateService`. Should I remove it for now, and leave this to Honeygems to add when they refactor the code to use the common methods?
CC @ptewson-sfdo

force2b: Thanks for the clarification. There are other parts of the code that return back the elevate status as part of wrapper class structure. Wondering if that's the pattern to stick with even for future code.

I suppose, along the same lines, the `getOrgDomainInfo()` method is more of a utility method than something  specific to PaymentServices, though it is being used specifically by PS right now. Rather than include the AuraEnabled method in this class, which would then require permissions granted, could there be a `UTIL_LWC` (or `LWC_Utilities') class that has exposed utility methods, even if those methods delegate to other service classes? That could also help with permissions going forward.

RadSFDeveloper: I understand for `getOrgDomainInfo()` that is a generic even though it is not used outside of the Elevate wrapper currently, but I assume we will have many Elevate related AuraEnabled methods so we do need a class specific to Elevate. I would avoid mixing Elevate AuraEnabled and generic methods in one class.

force2b: That's exactly my concern here. `getOrgDomainInfo` is generic and is in this class, while this class is meant to be specific to PaymentServices related methods.

force2b: That noted, if this class is going to expose AuraEnabled methods we can leave `getOrgDomainInfo` in here for now and eventually move out. The benefit of moving now though is that we can add the new utility class permission to our unpackaged metadata config for new RD2 users.

RadSFDeveloper: The new `UTIL_AuraEnabledCommon` has been created a generic utility class for common methods regardless of the LC/LWC domain. Its name and/or `PS_ElevateService` can be changed before merging into `master`, and we will discuss this with Honeygems next week at the dev meeting. CC @ptewson-sfdo

force2b (approved):
Thanks for making those changes @RadSFDeveloper. I'm approving with one remaining question - is a new unit test class needed for the `UTIL_AuraEnabledCommon` class?

RadSFDeveloper:
> Thanks for making those changes @RadSFDeveloper. I'm approving with one remaining question - is a new unit test class needed for the `UTIL_AuraEnabledCommon` class?

Valid question. Tests do not exist for `getOrgDomainInfo` currently. It is a matter of executing the method and verifying the values are not null. I will create the test class, but if we change the name of the common util class, the test class name needs to change as well.

RadSFDeveloper:
@rponti-sforg , Elevate credit card widget is ready for testing.

PR #5814, opened 2020-09-08
W-8017430 - P1 Bug - npsp__PaymentImportStatus set to Created even when errors

grivero:
[W-8017430](https://gus.my.salesforce.com/a07B0000008UmGKIA0) - Bug - P1 - BDI npsp__PaymentImportStatus__c is set to Created even when Payment record cannot be modified

- @callen-sfdo, @mldyang:
  - couldn't replicate error on unit test, errors at DI level are being caught before given that all the DI and Payment custom fields has same length and properties.
  -  20 Unit Tests failing when using Status = Update, won't update Status for now as @callen-sfdo suggested:
> Class: BDI_Donations_TEST
Exists logics with Status == "Matched", "Best Match Used", "User-Selected Match", "Created"
- ways to test this:
  - Validation Rule: 🟢  works, (use formula suggested on WI)
  - Removing Permissions on Sys Admin profile for some Payment Object Field `(for example: Payment Method)`: 🔴  did not work. Fails when running BDI but it doesn't process BDI. No modification to DI records, only updates Failure reason (for exampe: `You do not have permissions to modify npe01__OppPayment__c.npe01__Payment_Method__c.`
  - Trying to break string lengths or picklist values with apex or UI: 🔴  did not work, fails at Data Import insert level. We can'r run this type of DI records on BDI.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


callen-sfdo (approved):
Looks good.

Julian88Tex (approved):
- Verified Payment status error displays when validation rule blocks it's creation in both DI status, payment status, and respective error message fields.
- Verified Payments can still be created successfully when no validation rule to block exists.
- Verified matching to Opportunities/Payments and setting them to Closed Won/Paid respectively works.
- Worked with @deepmalamehta the BDI expert (thanks!) to validate above testing strategy.
- Validated unstable robot BDI tests pass or are low risk (custom object).

PR #5816, opened 2020-09-08
Support Address Platform Encryption

andrewyu-salesforce:
As an admin I want to use platform encryption on Address fields to make my system more secure.

Updated custom formula fields to use `IsBlank` instead of `Len` on Address Mailing address fields:
- Formula_MailingAddress__c
- Formula_MailingStreetAddress__c

# Critical Changes

# Changes
 - We updated the Address Mailing field on the Address object to use `IsBlank` instead of `Len` to support Address fields for Shield Platform Encryption.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


andrewyu-salesforce:
@RadSFDeveloper The PR is ready for review.

src/classes/ACCT_AccountMerge_TEST.cls
```diff
@@ -183,35 +183,46 @@ private with sharing class ACCT_AccountMerge_TEST {
...
-        Address__c addr = [select Id, Default_Address__c, MailingStreet__c from Address__c where  MailingStreet__c = '123 abc'];
-        system.assertEquals(false, addr.Default_Address__c);
+        Address__c targetAddr = null;
+        for (Address__c addr : [SELECT Id, MailingStreet__c FROM Address__c WHERE Default_Address__c = false]) {
+            if (addr.MailingStreet__c == '123 abc') {
+                targetAddr = addr;
+                break;
+            }
+        }
+        System.assertNotEquals(null, targetAddr, 'There should a non default address returned');
```
RadSFDeveloper, line 212: Let's update the assert message to `A non-default address should be returned.`

src/classes/CONV_Account_Conversion_BATCH_TEST.cls
```diff
@@ -55,17 +55,17 @@ public with sharing class CONV_Account_Conversion_BATCH_TEST {
         CONV_Account_Conversion_BATCH_TEST.setPreHHAccountModel(true);
 
         // Insert 150 test Contacts with a different owner than the Current User
-        List<Contact> cons = UTIL_UnitTestData_TEST.CreateMultipleTestContacts(150);
+        List<Contact> cons = UTIL_UnitTestData_TEST.CreateMultipleTestContacts(50);
```
RadSFDeveloper, line 58: @andrewyu-salesforce , let's use a var to store the number of contacts to create and then check in the following assert statements. The comment on line 57 still contains the "150" reference. Is there a reason to create 50 instead of 150 as before?

@force2b , is there a reason for creating 150 Contacts in the original test?

src/classes/CONV_Account_Conversion_BATCH_TEST.cls
```diff
@@ -242,17 +246,31 @@ public with sharing class CONV_Account_Conversion_BATCH_TEST {
...
+
+        address__c targetAddress;
+        for (Address__c address : [SELECT Id, MailingStreet__c, Default_Address__c FROM Address__c]) {
+            if (address.MailingStreet__c == '123 Main St.') {
+                targetAddress = address;
+                break;
+            }
+        }
+    
+        System.assertNotEquals(null, targetAddress, 'A address with matching mailing adress should be created');
```
RadSFDeveloper, line 262: `An address with the matching mailing street should be created.`

src/classes/CONV_Account_Conversion_BATCH_TEST.cls
```diff
@@ -242,17 +246,31 @@ public with sharing class CONV_Account_Conversion_BATCH_TEST {
...
+        System.assertEquals(true, targetAddress.Default_Address__c, 'The address should be the default address');
+        System.assertEquals(targetAddress.Id, [SELECT Current_Address__c FROM Contact WHERE Id = :c.Id].Current_Address__c,
+            'The default address should be the current address on contact');
+
+        System.assert(CAO_Constants.isHHAccountModel(),
+            'The Org Setting should be set to Household Account Model');
+        System.assertEquals(HH_Households.NO_HOUSEHOLDS_PROCESSOR, UTIL_CustomSettingsFacade.getHouseholdsSettings().npo02__Household_Rules__c,
+            'The Household Rules should be set correctaly');
```
RadSFDeveloper, line 270: Typo

RadSFDeveloper (commented):
Nice work @andrewyu-salesforce , below are minor comments related to test. Does the address verification service work when the address is encrypted?

src/classes/CONV_Account_Conversion_BATCH_TEST.cls
force2b: According to GitBlame, it was set to 150 around [7 years ago](https://github.com/SalesforceFoundation/NPSP/commit/cd692c4dc290a9ae81698c0f12ecf5a9295313ee) for bulk testing, though I'm not really sure how 150 is truly bulk as compared to 50. I think of bulk testing more in the > 200 to allow for proper trigger recursion tests. That's tested by the TDTM_DMLgt200_TEST class though. So, for here I think 50 is as fine as 150, but I am curious why the change was made.

andrewyu-salesforce: It was updated to 50 because this tests consistently getting CPU time out during the development. I also think there is no difference between 150 and 50 because we already have the bulk test in place.

andrewyu-salesforce:
@RadSFDeveloper @force2b The PR is updated and is ready for review.

src/classes/ADDR_Validator_Batch.cls
```diff
@@ -33,59 +33,90 @@
...
+    /**
+    * @description Chunk size to use when running in LDV Mode
+    */
+    private static final Integer CHUNK_SIZE = 10000000;
 
-    private String query;
+    /**
+    * @description Control whether or not to skip the already verified address 
+    */
+    private static Boolean skipPreviouslyVerified = false;
```
force2b, line 50: This should be an instance var vs. a static var so that its value is retained as _this_ is used to chain the next batch.

src/classes/ADDR_Validator_Batch.cls
```diff
@@ -33,59 +33,90 @@
...
-    }
-
-    /*******************************************************************************************************
-    * @description Collects the records to be passed to the execute method in batches.
-    * @param bc The context for the batch. Provided by the platform.
-    * @return Database.QueryLocator The record set processed in chunks for each iteration of the batch.
-    */
-    public Database.QueryLocator start(Database.BatchableContext bc) {
-        return Database.getQueryLocator(query);
+        skipPreviouslyVerified = skipPreviouslyVerified;
```
force2b, line 57:
```suggestion
        this.skipPreviouslyVerified = skipPreviouslyVerified;
```

force2b (changes requested):
Looks good overall. Just recommending a change to a var declaration.

src/classes/CONV_Account_Conversion_BATCH_TEST.cls
RadSFDeveloper: @andrewyu-salesforce , changes with regards to "let's use a var to store the number of contacts to create and then check in the following assert statements. The comment on line 57 still contains the "150" reference" are still outstanding.

src/classes/GE_LookupController_TEST.cls
```diff
@@ -92,21 +92,26 @@ private class GE_LookupController_TEST {
...
+        
+        insert opps;
 
         // doSearch utilizes SOSL, which means results must be hardcoded in tests
-        Test.setFixedSearchResults(new List<Id>{accounts[0].Id});
+        Test.setFixedSearchResults(new List<Id>{opps[0].Id});
 
         Test.startTest();
-        List<GE_LookupController.LookupResult> results = GE_LookupController.doSearchSOQL('test', 'Account');
+        List<GE_LookupController.LookupResult> results = GE_LookupController.doSearchSOQL('test', 'Opportunity');
```
RadSFDeveloper, line 110: @patrick-yan-sf , by this change the the GE controller test class will pass on an encrypted org. Not sure if you have a related WI and if more changes are pending.

@andrewyu-salesforce , thanks for making changes.

andrewyu-salesforce:
@force2b The code is updated and is ready for another review

RadSFDeveloper (approved):
Looks good, thanks @andrewyu-salesforce !

force2b (approved):
The code changes all look good. I'm approving for that part, but I do want to confirm that the address validation api has been tested given the changes to the formula fields.

andrewyu-salesforce:
@rponti-sforg This WI is ready for testing

andrewyu-salesforce:
Attach To [W-7969604](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UHQvIAO/view)

andrewyu-salesforce:
@force2b Both Google API and Smart Street API works

force2b:
Thanks for confirming @andrewyu-salesforce!

PR #5853, opened 2020-09-17
RD2: Support Cardholder Name in the Credit Card widget

force2b:
- Add support for capturing a single field for "Cardholder Name" in the RD credit card widget and passing that value to the Tokenization process

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

Custom Label:
- commonCardholderName

# Deleted Metadata


force2b:
Work Item: [W-7863547](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008TJGCIA4/view)

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -1,5 +1,6 @@
-import { api, track, LightningElement } from 'lwc';
+import {api, track, LightningElement, wire} from 'lwc';
```
andrewyu-salesforce, line 1: `wire` and `getObjectInfo` is not used anymore in the file.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
```diff
@@ -26,11 +27,21 @@ <h1 class='slds-text-body_regular' data-qa-locator='heading'>{labels.elevateWidg
                         </div>
                     </lightning-layout-item>
 
+                    <template if:false={isDisabled}>
+                        <lightning-layout-item padding="horizontal-small" size="6">
+                            <lightning-input name="cardholderName"
+                                             label={labels.cardholderNameLabel}
+                                             data-qa-locator="Elevate Cardholder Name"
```
andrewyu-salesforce, line 34: data-qa-locator should start with the field type. In this case it will be 'text Elevate Cardholder Name'

andrewyu-salesforce (commented):
@force2b Nice work. Just have some small feedback on the data-qa-locator and unused adapter in LWC.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
force2b: I can change that, though honestly I'm not sure why it uses the data type. It would make sense if there were more than one field in the UI for the same "label", but that's not generally the case.  I think I questioned that in the Quip doc as well.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
force2b: Good catch.

force2b:
Thanks for the review @andrewyu-salesforce. Changes made.

andrewyu-salesforce (approved):
Approve

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -71,6 +72,8 @@ export default class rd2EntryForm extends LightningElement {
 
     @track isElevateWidgetDisplayed = false;
     isElevateCustomer = false;
+    elevateAuthToken;
```
RadSFDeveloper, line 75: Let's rename `elevateAuthToken` to `paymentMethodToken` as used by the Elevate payments services API.

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -218,6 +221,21 @@ export default class rd2EntryForm extends LightningElement {
         this.isElevateWidgetDisplayed = this.isElevateCustomer === true
             && !this.isEdit // The Elevate widget is applicable to new RDs only
             && paymentMethod === 'Credit Card';// Verify the payment method value
+
+        if (this.isElevateWidgetDisplayed === true) {
+            // TO-DO: Prepopulate the Cardholder Name with the currently selected Contact or Organization Name
```
RadSFDeveloper, line 226: I am debating if we should keep this commented out code, but I guess it is fine since the next WI for the cardholder name should be worked on relatively soon.

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -71,6 +72,8 @@ export default class rd2EntryForm extends LightningElement {
 
     @track isElevateWidgetDisplayed = false;
     isElevateCustomer = false;
+    elevateAuthToken;
+    elevateCardholderName;
```
RadSFDeveloper, line 76: Can we rename `elevateCardholderName` to just  a `cardholderName`?

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -387,6 +405,7 @@ export default class rd2EntryForm extends LightningElement {
             : insertSuccessMessage.replace("{0}", recordName);
 
         showToast(message, '', 'success', []);
+        console.log('>> Elevate Token: ' + this.elevateAuthToken);
```
RadSFDeveloper, line 408: The log for the token is not required.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
```diff
@@ -26,11 +27,21 @@ <h1 class='slds-text-body_regular' data-qa-locator='heading'>{labels.elevateWidg
                         </div>
                     </lightning-layout-item>
 
+                    <template if:false={isDisabled}>
+                        <lightning-layout-item padding="horizontal-small" size="6">
+                            <lightning-input name="cardholderName"
+                                             label={labels.cardholderNameLabel}
+                                             data-qa-locator="text Elevate Cardholder Name"
```
RadSFDeveloper, line 34: I think this should be ` data-qa-locator="input Elevate Cardholder Name"`.

src/labels/CustomLabels.labels
```diff
@@ -7028,6 +7028,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Label for the Template Name field</shortDescription>
         <value>Template Name</value>
     </labels>
+    <labels>
+        <fullName>gePaymentRequestCardholderName</fullName>
```
RadSFDeveloper, line 7032: I guess it is fine for the label to belong to GE labels (not `common`) even though it will be used on both forms.
@sam-knox , @melissabarber : can you please review the new label?

RadSFDeveloper (commented):
Nice work @force2b ! Below are minor comments for the var naming mostly.

src/labels/CustomLabels.labels
force2b: I thought about moving to RD2, but it seems so specific to "gifts" in general (vs. Common for things that are very generic) so I ended up leaving it as a "ge" label.

src/lwc/rd2EntryForm/rd2EntryForm.js
force2b: I left it in only because I had started working on it while it was still part of this WI. Now that it's moved into the other WI, I figured it might still be useful here.

force2b:
Thanks for the code review @RadSFDeveloper. All changes made

RadSFDeveloper (approved):
Looks good, thanks @force2b !

src/labels/CustomLabels.labels
melissabarber: Thanks for the heads up about this. Thanks for leaving it as ge, but definitely something to keep in mind for components like this that will be shared across GE and RD entry screens. FYI, @ptewson-sfdo @psadie-sfdo @vmartinov-sforg @bdvorachek @dfuller114 @kenlewis-sfdo @voduyemi

melissabarber: K, let's try that again. My comment didn't make a whole lot of sense. I was operating under the assumption that this was a label we created for GE when we were setting up Elevate functionality in GE. Since it's new and it could be used by GE or RD entry screens, should this have the "common" prefix instead of GE? @psadie-sfdo do you recall why we *don't* have a label like this in GE?

psadie-sfdo: @melissabarber - we don't have a label for this because our credit card widget doesn't have a cardholder name input field. We construct the cardholder name on the fly and send it along with our elevate requests in the format of "{contact first name} {contact last name}" or "{organization account name} {organization account name}".

If I remember correctly, the sentiment behind that decision was that we wanted to eliminate the step for a gift entry user since we had all of that information already. I think we may need to re-visit that decision soon.

For the shared RD/GE label, could we consider prefixing with `npsp`? Specifically for labels like this which doesn't really fit `common`.

src/pages/GE_TokenizeCard.page
```diff
@@ -62,12 +63,16 @@
 
         /*******************************************************************************
         * @description Method listens for messages from a credit card widget LWC
+        * Event.data contains at least an "action" and optinally a "nameOnCard" value.
```
kenlewis-sfdo, line 66:
```suggestion
        * Event.data contains at least an "action" and optionally a "nameOnCard" value.
```

src/labels/CustomLabels.labels
force2b: There are no current labels that start with `npsp`, but interesting there is a very similar one that starts with common: `commonCreditNumber`. So it probably does make sense to just call this one `commonCardholderName`.

psadie-sfdo (approved):
Approve!

PR #5862, opened 2020-09-19
RD2: Prevent Elevate credit card widget display

npsp-reedestockton:
# Critical Changes

# Changes
We limited display of the Elevate credit card component to USD currency transactions, Contacts with US addresses, and 'Open' Recurring Donation types.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


RadSFDeveloper:
Attached to [W-8021184](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UoVZIA0/view)

src/classes/RD2_EntryFormController.cls
```diff
@@ -150,6 +150,24 @@ public with sharing class RD2_EntryFormController {
...
 
+    /**
+     * @description Validate that the current contact has a US address
+     * @return True if Contact address is a US address
+     */
+    @AuraEnabled
+    public static Boolean contactHasUSAddress(Id recordId) {
+        Contact rdContact = [SELECT MailingState, MailingCountry FROM Contact WHERE Id = :recordId];
+        if (String.isNotBlank(rdContact.MailingState)
+            && (rdContact.MailingCountry == 'US'
```
force2b, line 161: To be on the safe side, I think this should also check for:
- U.S.
- U.S.A.
- United States of America

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -211,13 +217,66 @@ export default class rd2EntryForm extends LightningElement {
...
+        let currencyCode;
+
+        if (this.isMultiCurrencyEnabled) {
+            currencyCode = this.template.querySelector('lightning-input-field[data-id="currencyField"]').value;
+
+        } else {
+            currencyCode = CURRENCY;
+        }
+
+        return currencyCode === CURRENCY_CODE_USD;
```
force2b, line 279: I realize that it's not needed today, but do you think this should be more of a contains in list? Something like:
```
const ELEVATE_SUPPORTED_CURRENCIES = [ 'USD' ];
...
...
...
return ELEVATE_SUPPORTED_CURRENCIES.includes(currencyCode);
```

force2b (commented):
Just a couple of early comments.

src/classes/RD2_EntryFormController.cls
RadSFDeveloper: That is correct, it is re-written to use `ADDR_Validator_UTIL.isUSCountry(country);`. However, I am checking the code now if we should validate the country in the JS instead of Apex...

src/lwc/rd2EntryForm/rd2EntryForm.js
RadSFDeveloper: Valid point, it has its benefits with regards to extending the functionality down the road. I need to rename all methods evaluating the currency then. But also, should we write the address validation code to be generic even though requirements are to allow US address and USD currency only?

src/classes/RD2_EntryFormController.cls
force2b: Just for performance reasons, I think it should stay in JS. Your point (below) about supporting other countries down the road is a good one. Maybe this could use a constant such `ELEVATE_SUPPORTED_COUNTRIES` that is an array of just US countries for now. 

I see an opportunity for shared content between this widget and GE if we use a utility component, but at the same time I'm not sure how GE is handling address/currency verification to display the CC widget.

RadSFDeveloper: @ptewson-sfdo , @psadie-sfdo : do you validate the currency and the Contact address for a GE purchase?

psadie-sfdo: I don't believe we do. As of right now we set the currency code for every purchase transaction using `UserInfo.getDefaultCurrency()`.

RadSFDeveloper: @force2b , the change has been made to validate the Contact address. We check Mailing Country only.
The Contact is validated regardless if the "Donor Type" is Account or Contact. The AC5 was not specific about the "Donor Type" so the implementation is not specific either.

CC @mpusto , @rponti-sforg

src/lwc/rd2EntryFormDonorSection/rd2EntryFormDonorSection.js
```diff
@@ -86,6 +86,7 @@ export default class rd2EntryFormDonorSection extends LightningElement {
         } else if (parentSObjType === CONTACT_OBJECT.objectApiName) {
             this.contactId = this.parentId;
             this.donorType = 'Contact';
+            this.dispatchContactChangeEvent(this.contactId);
```
RadSFDeveloper, line 89: @rponti-sforg , the new Recurring Donation test cases should cover creating it from a:
- Contact detail page where MailingCountry = US
- Contact detail page where MailingCountry = CA
- Contact detail page where MailingCountry is empty
- Account detail page
- Recurring Donations tab

RadSFDeveloper: Moreover, it should be verified that Recurring Donations on an org not connected to Elevate continue to be working as before, ie the functionality is not broken.

src/lwc/rd2EntryFormDonorSection/rd2EntryFormDonorSection.html
```diff
@@ -25,30 +25,33 @@
 
                     <template if:true={isContactDonor}>
                         <lightning-layout-item padding="horizontal-small" size="12" small-device-size="6" if:true={fields.contact} >
-                            <lightning-input-field field-name={fields.contact.apiName} value={contactId} required tabindex="0">
+                            <lightning-input-field 
+                                field-name={fields.contact.apiName} 
+                                value={contactId} 
+                                required tabindex="0" 
+                                onchange={handleContactChange}>
```
force2b, line 32: Very minor, but how about `handleDonorChange` since the same event is used for both the Account and Contact?

force2b (commented):
Very nice. One small comment.

src/lwc/rd2EntryFormDonorSection/rd2EntryFormDonorSection.html
RadSFDeveloper: It is only used for the Contact lookup field value change, not Account lookup field value change. If there is confusion when reading the code, let me know where and I will fix it.

force2b: Oh, I saw the line below that said "isAccountDonor" and thought that the input was for the Account, but it's still for the Contact, so you can ignore this.

force2b (approved):

src/lwc/rd2EntryFormDonorSection/rd2EntryFormDonorSection.html
RadSFDeveloper: I changed code a bit so Established Date field is constructed only twice. It is still difficult to see there are two Contact fields, could not make any change to make it as one.

RadSFDeveloper:
@rponti-sforg , the "Prevent credit card widget display" PR is ready for testing.

PR #5897, opened 2020-10-01
RD2: Set Elevate Recurring Id on the new Recurring Donation

RadSFDeveloper:
# Critical Changes

# Changes
 - We added these fields to the Recurring Donation object, and Read access to the Standard and System Administrator profiles:
   - Card Expiration Month
   - Card Expiration Year
   - Card Last 4

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom Field:
    - npe03__Recurring_Donation__c.CardExpirationMonth__c
    - npe03__Recurring_Donation__c.CardExpirationYear__c
    - npe03__Recurring_Donation__c.CardLast4__c
<br/>

- Custom Label:
    - RD2_ErrorRDIdIsMissingForCommitment

# Deleted Metadata

### DIFF
--- a/src/classes/PS_CommitmentRequest.cls
+++ b/src/classes/PS_CommitmentRequest.cls
@@ -489,19 +489,27 @@ public inherited sharing class PS_CommitmentRequest {
         }
 
         /** 
-         * @description Returns credit card expiration date
-         * @return String
-         */
-        public String getCardExpirationDate() {
-            return hasCardData() ? cardData.getExpirationDate() : null;
-        }
-
-        /** 
          * @description Returns credit card last 4 digits
          * @return String
          */
         public String getCardLast4() {
-            return hasCardData() ? cardData.getLast4() : null;
+            return hasCardData() ? cardData.last4 : null;
+        }
+
+        /** 
+         * @description Returns credit card expiration month
+         * @return String
+         */
+        public String getCardExpirationMonth() {
+            return hasCardData() ? cardData.expirationMonth : null;
+        }
+
+        /** 
+         * @description Returns credit card expiration year
+         * @return String
+         */
+        public String getCardExpirationYear() {
+            return hasCardData() ? cardData.expirationYear : null;
         }
     }
 
@@ -512,21 +520,5 @@ public inherited sharing class PS_CommitmentRequest {
         public String last4;
         public String expirationMonth;
         public String expirationYear;
-
-        /** 
-         * @description Returns credit card expiration date
-         * @return String
-         */
-        public String getExpirationDate() {
-            return expirationMonth + '/' + expirationYear;
-        }
-
-        /** 
-         * @description Returns credit card last 4 digits
-         * @return String
-         */
-        public String getLast4() {
-            return last4;
-        }
     }
 }
--- a/src/classes/PS_CommitmentRequest.cls
+++ b/src/classes/PS_CommitmentRequest.cls
@@ -104,6 +104,11 @@ public inherited sharing class PS_CommitmentRequest {
     private static Integer DEFAULT_CURRENCY_MULTIPLIER = 100;
 
     /***
+     * @description Contains max length for a first/last name passed to the Elevate commitment request
+     */
+    private static Integer MAX_NAME_LENGTH = 60;
+
+    /***
      * @description Payment Services configuration
      */
     @TestVisible
@@ -351,8 +356,8 @@ public inherited sharing class PS_CommitmentRequest {
                 return this;
             }
 
-            this.firstName = getMaxLenName(contact.FirstName);
-            this.lastName = getMaxLenName(contact.LastName);
+            this.firstName = trimNameField(contact.FirstName);
+            this.lastName = trimNameField(contact.LastName);
 
             return this;
         }
@@ -367,8 +372,8 @@ public inherited sharing class PS_CommitmentRequest {
                 return this;
             }
 
-            this.firstName = getMaxLenName(organization.Name);
-            this.lastName = getMaxLenName(organization.Name);
+            this.firstName = trimNameField(organization.Name);
+            this.lastName = trimNameField(organization.Name);
 
             return this;
         }
@@ -378,8 +383,8 @@ public inherited sharing class PS_CommitmentRequest {
          * @param name First or Last Name
          * @return String
          */
-        private String getMaxLenName(String name) {
-            return String.isBlank(name) ? name : name.left(60);
+        private String trimNameField(String name) {
+            return String.isBlank(name) ? name : name.abbreviate(MAX_NAME_LENGTH);
         }
 
         /**
--- a/src/classes/PS_CommitmentRequest.cls
+++ b/src/classes/PS_CommitmentRequest.cls
@@ -35,13 +35,61 @@
 */
 public inherited sharing class PS_CommitmentRequest {
 
+    /***
+     * @description Commitment type
+     */
     public enum CommitmentType {
         SUSTAINER
     }
 
+    /***
+     * @description Commitment payment method. Currently only credit card is supported.
+     */
     public enum PaymentMethod {
         CARD
     }
+
+    /***
+     * @description Commitment Status
+     * Commitments start in the NEW state, then transition to ACTIVE after the first payment.
+     * The Status cannot be an enum since NEW is a reserved keyword.
+     */
+    public static String STATUS_NEW = 'NEW';
+    public static String STATUS_ACTIVE = 'ACTIVE';
+    public static String STATUS_INACTIVE = 'INACTIVE';
+    public static String STATUS_FAILING = 'FAILING';
+
+    /***
+     * @description Commitment Status Reason
+     */
+    public enum CommitmentStatusReason {
+        COMMITMENT_CREATED,
+        PAID_AS_AGREED,
+        PARTIALLY_PAID,
+        PAYMENT_ERROR,
+        PAYMENT_DECLINE,
+        DONOR_CANCELED,
+        MULTIPLE_FAILURES,
+        NEVER_PAID
+    }
+
+    /***
+     * @description Commitment schedule frequency
+     */
+    public enum Frequency {
+        DAY, WEEK, MONTH, YEAR
+    }
+
+    /***
+     * @description Translates the Installment Period into the Elevate API frequency value
+     */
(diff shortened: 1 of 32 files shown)


RadSFDeveloper:
Tracking [W-8035154](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UwdTIAS/view)

src/classes/PS_CommitmentRequest.cls
```diff
@@ -311,12 +367,21 @@ public inherited sharing class PS_CommitmentRequest {
...
             return this;
         }
 
+        /**
+         * @description Return max number of character allowed for the first/last name in Elevate
+         * @param name First or Last Name
+         * @return String
+         */
+        private String getMaxLenName(String name) {
+            return String.isBlank(name) ? name : name.left(60);
```
force2b, line 382: Probably best to put `60` into a constant at the top of the class just in case it changes in Elevate in the future.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -311,12 +367,21 @@ public inherited sharing class PS_CommitmentRequest {
...
 
             return this;
         }
 
+        /**
+         * @description Return max number of character allowed for the first/last name in Elevate
+         * @param name First or Last Name
+         * @return String
+         */
+        private String getMaxLenName(String name) {
```
force2b, line 381:
```suggestion
        private String trimNameField(String name) {
```

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -22,6 +22,30 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>CardExpirationDate__c</fullName>
+        <description>The expiration month and year of the card used for this Recurring Donation.</description>
```
force2b, line 27: Should the Help for these two fields use "credit card" instead of just "card"?
@sam-knox ?

@RadSFDeveloper If the field only has MM/YY or MM/YYYY should the field length be either 5 or 7?

src/classes/RD2_EntryFormController.cls
```diff
@@ -281,47 +290,108 @@ public with sharing class RD2_EntryFormController {
...
+            return;
+        }
+
+        try {
+            PS_CommitmentRequest.ResponseBody commitment = new PS_CommitmentRequest().getCommitment(response);
+
+            if (commitment.hasId()) {
+                npe03__Recurring_Donation__c rd = new npe03__Recurring_Donation__c(
+                    Id = recordId,
+                    CommitmentId__c = commitment.id,
```
force2b, line 358: Looking at the unpackaged config, these two new fields are defined as Readonly in the Admin/Standard profile permissions - which is probably fine. However, this method that updates these fields is in a class defined as "With Sharing". I think these need to be updated in an inner class (somewhere) defined as "Without Sharing" (rather than grant edit access to all users).

unpackaged/config/rd2_post_config/objects/npe03__Recurring_Donation__c.object
```diff
@@ -58,6 +58,30 @@
         <trackTrending>false</trackTrending>
         <type>Date</type>
     </fields>
+    <fields>
```
force2b, line 61: I'm pretty sure that you do not need to include these fields in the object unpackaged metadata. It's worth testing, but there are other fields granted permission in the profiles (EndDate, CurrentYearValue, NextYearValue, etc.) that are not included in the object file.

force2b (commented):
Very nice. Just some comments on naming, unpackaged, and help text.

src/classes/PS_CommitmentRequest.cls
RadSFDeveloper: Done, I was going bak and forth with naming this one. `trim` keyword was a part of many variations :)

src/objects/npe03__Recurring_Donation__c.object
RadSFDeveloper: It is `MM/YYYY `. I can change it to 7 to be exact.

force2b: Out of curiosity, do you think we should parse this into two fields:
cardExpirationMonth__c (2) and cardExpirationYear__c (4)

sam-knox: @force2b - Let me ask if there's a current convention. I was going to say use "credit card" but some people use debit cards.

sam-knox: Yeah, let's go with just "card" @force2b

npsp-reedestockton (approved):
Great work!

src/labels/CustomLabels.labels
```diff
@@ -2207,6 +2207,14 @@
         <shortDescription>The RD2 Batch job can only be executed when Enhanced RD is enabled</shortDescription>
         <value>You must enable Enhanced Recurring Donations before you can run the Enhanced Recurring Donations batch job.</value>
     </labels>
+    <labels>
+        <fullName>RD2_ErrorRDIdIsMissingForCommitment</fullName>
```
RadSFDeveloper, line 2211: @sam-knox , this new custom label is used to verify that RD2 entry form does create and passes a Recurring Donation Id before a commitment request can be sent to Elevate API.

RadSFDeveloper (commented):
@sam-knox , @force2b : the code has been updated based on the feedback. Can you please review again?

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -34,6 +34,15 @@
...
 */
 public inherited sharing class RD2_ElevateIntegrationService {
+
+    /***
+     * @description Contains a prefix appended to a GUID to indicate the RD is in draft
+     * and has a temp commitment Id so future installment Opportunities are not
+     * created for this Recurring Donation and any user can change it
+     * from a temp value to the actual value.
+     */
+    public static final String TEMP_COMMITMENT_ID_PREFIX = '_TEMP_';
```
force2b, line 44: What would you think of `_PENDING_` for this instead of `_TEMP`?

force2b (commented):
Really nice. One last comment.

src/classes/RD2_ElevateIntegrationService.cls
RadSFDeveloper: Perfect, will make the change now.

force2b (approved):
Great work on this!!

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -22,6 +22,42 @@
...
+    <fields>
+        <fullName>CardExpirationMonth__c</fullName>
+        <description>The expiration month of the card used for this Recurring Donation.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The expiration month of the card used for this Recurring Donation.</inlineHelpText>
+        <label>Card Expiration Month</label>
+        <length>2</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
```
sam-knox, line 34: Minor thing, but if we make these fields a number type, they can be sorted easier in reports.

@RadSFDeveloper @force2b

src/labels/CustomLabels.labels
sam-knox: @RadSFDeveloper - What circumstances would produce this error? Would it mean that the integration is broken?

sam-knox: Or perhaps it would mean that the RD record itself couldn't be created (due to validation error or somthing)?

src/objects/npe03__Recurring_Donation__c.object
RadSFDeveloper: I followed the field definition as it is on the Payment SObject , for example [`Card_Expiration_Month__c`](https://github.com/SalesforceFoundation/NPSP/blob/master/src/objects/npe01__OppPayment__c.object#L104). To be consistent, let's follow what has already been done.

src/labels/CustomLabels.labels
RadSFDeveloper: If the RD is not created or there is an error either in the LWC or when passing the RD Id, the commitment creation must be prevented in such case. And even though this is a programatic error, this is another check to ensure data integrity and correct process.

RadSFDeveloper: @sam-knox , that ^ is correct. Either validation or something happened that prevented RD creation or it was created by LWC is not passing it to the commitment creation. So if this happens (it should not) , the commitment request should not be sent to Elevate.

src/objects/npe03__Recurring_Donation__c.object
sam-knox: Got it - sounds good.

sam-knox (changes requested):
The new fields all look good. One requested change for the label, line 2216:

"There was an error sending the Recurring Donation ID to Elevate. Contact your system administrator for assistance."

RadSFDeveloper:
> The new fields all look good. One requested change for the label, line 2216:
> 
> "There was an error sending the Recurring Donation ID to Elevate. Contact your system administrator for assistance."

@sam-knox , we are not sending the RD Id to Elevate but we do need to ensure it exists before the commitment request is sent. Should the custom label change to indicate the RD Id is not passed due to an internal error during RD creation or something like that?

sam-knox:
@RadSFDeveloper - Good point, we're not actually sending the ID itself. Which of these two do you think captures the error the best?

1. There was an error sending the Recurring payment request to Elevate. Contact your system administrator for assistance.

2. There was an error creating the Recurring Donation record. No Recurring payment request was sent to Elevate. Contact your system administrator for assistance.


Even though we call them "commitments", that term doesn't appear anywhere in the Elevate UI. It's represented as either "Recurring" or "Recurring payment".

src/labels/CustomLabels.labels
RadSFDeveloper: @sam-knox , both suggestions are valid, chose the second one since it has more info about the cause and impact.

sam-knox:
Sounds good to me @RadSFDeveloper!

sam-knox (approved):

PR #5898, opened 2020-10-01
[DO NOT MERGE] Gift Entry: convert record type id from lookups to picklists in the form

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geFormFieldPicklist/geFormFieldPicklist.js
```diff
@@ -90,25 +125,60 @@ export default class GeFormFieldPicklist extends LightningElement {
...
     }
 
-    set recordTypeId(id) {
-        this._recordTypeId = id || this.defaultRecordTypeId;
+    // ================================================================================
+    // PRIVATE METHODS
+    // ================================================================================
+    handleValueChange(event) {
+        event.stopPropagation();
+        this.value = event.detail.value;
```
kenlewis-sfdo, line 149: What do you think about adding a `displayValue` prop here to the detail object and including the RecordTypeName? 
(I think the built-in change event only sends the value, right?) Then the new `updateFormState` method in the renderer could just append to state and it would include both the RT val and displayVal without needed to look up the RTName in the renderer?

psadie-sfdo: As discussed, we'll hold off on making additional changes to the existing picklist flow.

psadie-sfdo:
[Tracking W-7963596](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UDVkIAO/view)

kenlewis-sfdo (commented):
This looks really good to me!!  I really like the nice clean new methods, the headers in the picklists file and of course the jest tests!  Nice work!   I know you've tested our work together already, but I'm also deploying them together to an org to independently verify.  Will take one more quick pass in the a.m. tomorrow, but this looks great.

psadie-sfdo:
> This looks really good to me!! I really like the nice clean new methods, the headers in the picklists file and of course the jest tests! Nice work! I know you've tested our work together already, but I'm also deploying them together to an org to independently verify. Will take one more quick pass in the a.m. tomorrow, but this looks great.

Sounds good! Please do! I'm doing another run through this morning as well. Note that the two things I had to do consistently while testing both were to resolve the small conflict in `geFormFieldPicklist` around the public methods header and also to update the property that was being passed into the `geFormFieldPicklist`components in the `geFormField` template from `value` to `valueFromFormState`.

PR #5900, opened 2020-10-01
Gift Entry: Centralize State Management draft

kenlewis-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormFieldPicklist/geFormFieldPicklist.js
+++ b/src/lwc/geFormFieldPicklist/geFormFieldPicklist.js
@@ -57,11 +57,6 @@ export default class GeFormFieldPicklist extends LightningElement {
         return `${this.objectName}.${this.fieldName}`;
     }
 
-    handleValueChange(event) {
-        this.value = event.detail.value;
-        this.dispatchEvent(new CustomEvent('onchange', event)); // bubble up to ge-form-field
-    }
-
     @api
     reportValidity(){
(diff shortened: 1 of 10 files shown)


src/lwc/geFormFieldPicklist/geFormFieldPicklist.js
```diff
@@ -57,11 +57,6 @@ export default class GeFormFieldPicklist extends LightningElement {
         return `${this.objectName}.${this.fieldName}`;
     }
 
-    handleValueChange(event) {
```
psadie-sfdo, line 60: @kenlewis-sfdo - I think we should keep this, stop the event from propagating, and just update the custom event name accordingly.

As it is, the [event](https://git.soma.salesforce.com/aura/lightning-global/blob/6410b243701b052ab8dd9915eb5fec92ba8f2319/ui-lightning-components/src/main/modules/lightning/combobox/combobox.js#L402) (link courtesy of Ben) from the `lightning-combobox` is set to `bubbles: true` and `composed: true` by default. There's a small chance that it'll get unintentionally handled somewhere higher in our hierarchy.

kenlewis-sfdo:
Superseded by [5908](https://github.com/SalesforceFoundation/NPSP/pull/5908)

PR #5901, opened 2020-10-01
RD2: Prevent a Canceled Elevate RD from being re-activated

npsp-reedestockton:
We prohibited re-activating a Recurring Donation linked to Elevate.

# Critical Changes

# Changes
 - To preserve data integrity with Elevate, you can't edit the Status for an Elevate connected Recurring Donation when Status is Closed.

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom Label:
    - RD2_ElevateRDClosedStatusCannotBeChanged

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_ElevateIntegrationService.cls
+++ b/src/classes/RD2_ElevateIntegrationService.cls
@@ -120,6 +120,22 @@ public inherited sharing class RD2_ElevateIntegrationService {
         return !isChanged;
     }
 
+    /***
+     * @description Determines whether an Elevate Recurring Donation is changing from Closed to some other state
+     * @param rd Changed Recurring Donation
+     * @param oldRd Old Recurring Donation
+     * @return Boolean true if RD Status is changing from Closed to another state on an Elevate RD
+     */
+    public Boolean isElevateStatusChangeFromClosed(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
+        if (String.isNotBlank(oldRd.CommitmentId__c)
+            && new RD2_RecurringDonation(oldRd).isClosed()
+            && !new RD2_RecurringDonation(rd).isClosed()
+        ) {
+            return true;
+        }
+        return false;
+    }
+
     /**
     * @description For each CommitmentId, find any existing Opportunities with the matching Commitmentid
     * and update the RecuringDonation lookup field (only if null).
--- a/src/classes/RD2_ElevateIntegrationService.cls
+++ b/src/classes/RD2_ElevateIntegrationService.cls
@@ -141,22 +141,6 @@ public inherited sharing class RD2_ElevateIntegrationService {
         return isReactivated;
     }
 
-    /***
-     * @description Determines whether an Elevate Recurring Donation is changing from Closed to some other state
-     * @param rd Changed Recurring Donation
-     * @param oldRd Old Recurring Donation
-     * @return Boolean true if RD Status is changing from Closed to another state on an Elevate RD
-     */
-    public Boolean isElevateStatusChangeFromClosed(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
-        if (String.isNotBlank(oldRd.CommitmentId__c)
-            && new RD2_RecurringDonation(oldRd).isClosed()
-            && !new RD2_RecurringDonation(rd).isClosed()
-        ) {
-            return true;
-        }
-        return false;
-    }
-
     /**
     * @description For each CommitmentId, find any existing Opportunities with the matching Commitmentid
     * and update the RecuringDonation lookup field (only if null).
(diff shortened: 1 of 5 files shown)


npsp-reedestockton:
[W-8035271](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UwhVIAS/view)

npsp-reedestockton:
@RadSFDeveloper @force2b Ready for review.  Thanks!

@sam-knox Would you please review the label for this PR?  Thanks!

src/classes/RD2_ValidationService.cls
```diff
@@ -193,6 +193,11 @@ public with sharing class RD2_ValidationService {
             npe03__Recurring_Donation__c rd = rds[i];
             npe03__Recurring_Donation__c oldRd = oldRds[i];
 
+            if (elevateService.isElevateStatusChangeFromClosed(rd, oldRd)) {
```
force2b, line 196: Is there a check anywhere to confirm that Elevate Integration is enabled? My concern is this validation will block customers that may choose to put a value in the CommitmentId even if they're not using Elevate Payments. I believe that in other places we do check for ElevatePayments enabled before enforcing validations. Though if that's not correct - and we ignore the elevate enabled status for other validations - then it can be ignored here. Basically just looking for consistency.

src/labels/CustomLabels.labels
```diff
@@ -1601,6 +1601,14 @@
         <shortDescription>Warning displayed on Customizable Rollups page when enhanced RDs are enabled</shortDescription>
         <value>&lt;b&gt;WARNING: You have Enhanced Recurring Donations enabled. If you disable Customizable Rollups, rollups to Recurring Donations will no longer function.&lt;/b&gt;</value>
     </labels>
+    <labels>
+        <fullName>RD2_ElevateClosedStatusChangeError</fullName>
```
RadSFDeveloper, line 1605: Let's rename the custom label to match the requirement (expected behaviour). For example: `RD2_ElevateRDClosedStatusCannotBeChanged` or `RD2_ElevateRDClosedStatusIsLocked`  or `RD2_ElevateRDStatusIsLockedWhenClosed`.

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -120,6 +120,22 @@ public inherited sharing class RD2_ElevateIntegrationService {
         return !isChanged;
     }
 
+    /***
+     * @description Determines whether an Elevate Recurring Donation is changing from Closed to some other state
+     * @param rd Changed Recurring Donation
+     * @param oldRd Old Recurring Donation
+     * @return Boolean true if RD Status is changing from Closed to another state on an Elevate RD
+     */
+    public Boolean isElevateStatusChangeFromClosed(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
```
RadSFDeveloper, line 129: This renamed method that is using `RD2_RecurringDonation` to check if it is Elevate record. It also checks if the Elevate is enabled as suggested by @force2b . The result is contained in a Boolean var that reminds what is being checked. Also, the ApexDoc is updated.

```

    /***
     * @description Determines whether an Elevate Recurring Donation is changing from closed state to another state
     * @param rd Changed Recurring Donation
     * @param oldRd Old Recurring Donation
     * @return Boolean
     */
    public Boolean isElevateRecordReactivated(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
        if (!isIntegrationEnabled()) {
            return false;
        }

        RD2_RecurringDonation rdRecord = new RD2_RecurringDonation(rd);
        RD2_RecurringDonation oldRDRecord = new RD2_RecurringDonation(oldRd);

        Boolean isReactivated = oldRDRecord.isElevateRecord()
            && oldRDRecord.isClosed()
            && !rdRecord.isClosed();
        
        return isReactivated;
    }

```

sam-knox (changes requested):
Label looks good, but please add a period to the end of the sentence. Line 1610.

npsp-reedestockton:
@RadSFDeveloper @force2b Thank you for the review suggestions.  All have been implemented.  Please take another look when you have a moment.

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -120,6 +120,43 @@ public inherited sharing class RD2_ElevateIntegrationService {
...
+        return isReactivated;
+    }
+
+    /***
+     * @description Determines whether an Elevate Recurring Donation is changing from Closed to some other state
+     * @param rd Changed Recurring Donation
+     * @param oldRd Old Recurring Donation
+     * @return Boolean true if RD Status is changing from Closed to another state on an Elevate RD
+     */
+    public Boolean isElevateStatusChangeFromClosed(npe03__Recurring_Donation__c rd, npe03__Recurring_Donation__c oldRd) {
```
RadSFDeveloper, line 150: Let's remove this method and replace its call in the tests with `isElevateRecordReactivated()`.

npsp-reedestockton: Thank you!!

RadSFDeveloper (approved):
Looks good, thanks @npsp-reedestockton !

force2b (approved):
Very nice

PR #5905, opened 2020-10-02
[RD2] CallableApi action to Retrieve Schedule Records

force2b:
# Critical Changes

# Changes

* Implement a new Callable Api service to return Enhanced Recurring Donation schedule record details for a specified set of Recurring Donation Id's. 

# Issues Closed

# Community Ideas Delivered

# New Metadata

Apex Classes:
- RD2_ApiService
- RD2_ApiService_TEST

# Deleted Metadata


force2b:
[W-8100851](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008dNDtIAM/view)

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,140 @@
...
+        public Decimal installmentAmount { get; set; }
+        public Integer installmentFrequency { get; set; }
+        public String installmentPeriod { get; set; }
+        public String paymentMethod { get; set; }
+        public Date startDate { get; set; }
+        public String statusReason { get; set; }
+        public String currencyCode { get; set; }
+
+        public Schedule(RecurringDonationSchedule__c rdSchedule, npe03__Recurring_Donation__c rd) {
+            isActive = rdSchedule.Active__c;
```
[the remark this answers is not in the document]
force2b: Really good point. I would like the schedule data to be accurate - reflecting what is truly active at that point. Need to look at the schedule service a little closer.

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,140 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group Recurring Donations
+* @description All services specific to Enhanced Recurring Donation Api's
+*
+*/
+public inherited sharing class RD2_ApiService {
```
npsp-reedestockton, line 37: I have concerns about the approach taken to this API.  I'm not sure it best serves the stated purpose to assist in forecasting calculations to just provide access to the schedule data as is.  There is complex logic that is deeply embedded in our handling of schedules and I'm not sure providing the raw data "as is" provides the best API.  

Would it perhaps be useful to consider a variation of getVisualizedInstallments that does a forecast based upon input parameters???

force2b:
@RadSFDeveloper and @npsp-reedestockton All changes have been made. This is now ready for a second review.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -0,0 +1,291 @@
...
+        for (Schedule s : schedulesByRd.get(rd.Id)) {
+            if (s.isActive) {
+                activeCount++;
+            }
+            if (s.isCurrent) {
+                currentCount++;
+            }
+        }
+        System.assertEquals(1, currentCount, 'There should be one current schedules returned');
+        System.assertEquals(3, activeCount, 'There should be two active schedules returned');
```
npsp-reedestockton, line 87: Text does not match assert.  Should be "three active schedules" I think.

npsp-reedestockton (changes requested):

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @description Returns query on Recurring Donation Schedule
+     * Return all schedules for the RD Schedule visualization.
+     * The Schedule Service will filter them out if not applicable within the time period.
+     * @return String RD Schedule subquery used in the Recurring Donation SOQL
+     */
+    public String getScheduleSubQuery() {
+        Set<String> fields = new RD2_ScheduleService.ScheduleQueryHandler().getFields();
+        fields.add('CreatedDate');
```
npsp-reedestockton, line 110: I apologize for not thinking of this earlier, but would it make some sense to add ModifiedDate as well?  

I was originally thinking we did not need it because we don't modify schedules, but we do.  When we deactivate a schedule, we do so by changing the EndDate__c to today()-1, so ModifiedDate would show us when that happened.  

We can usually infer that from StartDate__c of the new schedule, so it isn't critical that we have this, but it would be a bit more transparent I think.  And we might encounter some scenarios with re-activated RDs where it would be useful to have.

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,178 @@
...
+        if (!RD2_EnablementService.isRecurringDonations2Enabled) {
+            return JSON.serialize(schedulesByRdId);
+        }
+
+        RD2_ScheduleService.PauseScheduleHandler pauseHandlerService = new RD2_ScheduleService.PauseScheduleHandler();
+
+        String soql = getRecurringDonationQuery();
+        for (npe03__Recurring_Donation__c rd : Database.query(soql)) {
+
+            Boolean isRdCurrentlyPaused = pauseHandlerService.hasActivePause(
```
RadSFDeveloper, line 58: When a pause is valid (Start Date <= End Date) and in future (referenceDate <= End Date), then the pause schedule is considered as an active pause. However, it does not necessarily mean it is the current pause. Similarly is true for a regular active schedule: we can have up to two active future schedules but only one is current.  In summary, neither `isActive` nor `isActivePause`  conveys if the schedule is current, ie neither method checks if "Start Date <= referenceDate <= EndDate". 

So we need another method,  or order of commands to find the active pause schedule, and if one exists, check if the current schedule is paused. Otherwise, if no pause exists, then no additional check is to be done.

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,178 @@
...
+        public String paymentMethod { get; set; }
+        public Date startDate { get; set; }
+        public String statusReason { get; set; }
+        public String currencyCode { get; set; }
+        public Datetime createdDate { get ;set; }
+
+        public Schedule(RecurringDonationSchedule__c rdSchedule, npe03__Recurring_Donation__c rd) {
+            isActive = rdSchedule.Active__c;
+            isCurrent = rdSchedule.StartDate__c <= RD2_ScheduleService.currentDate
+                && (rdSchedule.EndDate__c == null || rdSchedule.EndDate__c >= RD2_ScheduleService.currentDate);
```
RadSFDeveloper, line 143: Do you please mind using "<=" instead, ie `|| RD2_ScheduleService.currentDate <= rdSchedule.EndDate__c`?

src/classes/RD2_ApiService.cls
force2b: Good catch. I misread the meaning and logic in that method. The method description is `Indicates if the Recurring Donation is an active pause schedule` which I read has the RD is currently paused. I'll update the description to clarify.

I do agree that another method is needed to see if the pause is currently active.

force2b: I'll add that the `isWithinDateRange` method name is very misleading. If I provide a date to that method I would expect it to tell me if that date is within the range of the start and end dates on the record. That is definitely not what it does though. It actually returns true if the record is either current or in the future based on the provided date. 

What would you think of me renaming the method to `isCurrentOrFuture()`? Note that this method is duplicated in another place in the class so I would rename both.
@npsp-reedestockton

RadSFDeveloper: I have not checked the code for existing method, wondering if we can leave it as is, or it is required to be renamed. With regards to renaming `isWithingDateRange()`, it is perfectly fine and valid. How about we use `isScheduled()` instead ("scheduled" is one of the synonyms of "future" :)))

force2b: It's not *required* to rename, but I definitely feel that the method name does not reflect the behavior. I like `isScheduled` as a replacement for `isWithinDateRange`.

src/classes/RD2_ApiService.cls
```diff
@@ -0,0 +1,181 @@
...
+        for (npe03__Recurring_Donation__c rd : Database.query(soql)) {
+
+            Boolean isRdCurrentlyPaused = pauseHandlerService.isCurrentlyPaused(
+                rd.RecurringDonationSchedules__r, RD2_ScheduleService.currentDate
+            );
+
+            List<RD2_ApiService.Schedule> schedules = new List<RD2_ApiService.Schedule>();
+            for (RecurringDonationSchedule__c s : rd.RecurringDonationSchedules__r) {
+                RD2_ApiService.Schedule schedule = new RD2_ApiService.Schedule(s, rd);
+                if (isRdCurrentlyPaused && schedule.isActiveAndNotPaused()) {
```
RadSFDeveloper, line 65: I think we just need to check `if (isRdCurrentlyPaused  && schedule.isCurrent && !schedule.isPause) {` in order to set the `isCurrent` to false. We might not need a new method for this one line.

src/classes/RD2_ScheduleService_TEST.cls
```diff
@@ -2014,6 +2014,45 @@ public with sharing class RD2_ScheduleService_TEST {
...
+    */
+    @isTest
+    private static void shouldIndicateIfCurrentlyPaused() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+        RD2_ScheduleService.currentDate = DATE_ESTABLISHED;
+
+        npe03__Recurring_Donation__c rd = getRecurringDonationMonthlyBuilder().build();
+
+        List<RecurringDonationSchedule__c> schedules = new RD2_ScheduleService().buildNewSchedules(rd);
+        schedules.add(buildPauseSchedule(rd.Id, Date.today(), Date.today().addMonths(2)));
```
RadSFDeveloper, line 2028: When a pause for the next installment is added, it has the next installment Close Date. For example, if a pause for a monthly RD with Day of Month = 1 is paused in Nov, the pause schedules will have Start Date = End Date = Nov 1. So, if the pause is built for the next installment Opp instead using `Date.today()`, I believe the test would fail.

src/classes/RD2_ScheduleService.cls
```diff
@@ -990,7 +990,8 @@ public without sharing class RD2_ScheduleService {
     public class PauseScheduleHandler {
 
         /***
-        * @description Indicates if the Recurring Donation is an active pause schedule
+        * @description Indicates if the Recurring Donation has an active pause schedule
+        * at some point in the future (not necessarily in a current pause state).
```
npsp-reedestockton, line 994: Very clear.  Thanks.

src/classes/RD2_ScheduleService_TEST.cls
npsp-reedestockton: Wouldn't buildPauseSchedule() as used in this test always result in a currently active pause regardless of the cadence of upcoming installments?  I would think that the pause would always begin today and extend for two months, covering whatever installments would otherwise occur during that span.

RadSFDeveloper: The pause schedules starts on the first paused installment close date and ends on the last paused installment date.

RadSFDeveloper (approved):
Great work, thank you @force2b !

force2b:
@npsp-reedestockton This is ready for re-review

npsp-reedestockton (approved):
Thanks for working through this so carefully.  Nice work!

PR #5906, opened 2020-10-02
RD2: Handle cancel commitment errors

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


RadSFDeveloper:
Tracking [W-7993904](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UVdQIAW/view)

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -183,39 +186,226 @@ public inherited sharing class RD2_ElevateIntegrationService {
...
     /**
-    * @description Sends HttpRequest to Elevate to cancel Commitments for close Recurring Donations
+    * @description Cancels commitments in Elevate connected to the closed Recurring Donations
     * @param closedRDs Closed RDs with the user specified closed reason
     * @return void
     */
     public void cancelCommitments(List<npe03__Recurring_Donation__c> closedRDs) {
-        UTIL_Http.RequestService requestService = new UTIL_Http.RequestService();
-        List<UTIL_Http.Response> responses = new List<UTIL_Http.Response>();
+        new CancelCommitmentService().cancelCommitments(closedRDs);
```
force2b, line 203: @RadSFDeveloper How does the UI handle errors? The method is synchronous, but it only writes a message to the log. Is the intent that the page doesn't display a toast or other "error" style message, but instead just reverts the RDStatus to "Active" and updates the StatusReason field with the returned error?

RadSFDeveloper: This `cancelCommitments` is called from the async `RD2_QueueableService.CancelCommitmentService`. The "Close Recurring Donation" QuickAction is prevented only if there are issues with closing the RD. The "Status Reason" is set to "Pending Elevate Process". When the commitment cancel succeeds, then it is set to the user entered value at the "Close Recurring Donation". On an error, the "Status Reason" is set to the error, but the RD state stays closed.

force2b: Thanks for the explanation @RadSFDeveloper

force2b (approved):

andrewyu-salesforce (changes requested):
@RadSFDeveloper The error handling looks great. Just a small typo in the description.

andrewyu-salesforce (approved):
@RadSFDeveloper The change looks great.

RadSFDeveloper:
@rponti-sforg , the PR is ready for testing.

PR #5927, opened 2020-10-05
Access to Payment field on Allocation object required to process a GE batch

dfuller114:
# Critical Changes

# Changes
A user will no longer receive an FLS error when processing a batch if the user does not have access to the allocation payment field and the field is not on the gift entry form template.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


kenlewis-sfdo (approved):
These changes look great to me! Thanks for creating that thorough design document and for your analysis and consideration of options!

ptewson-sfdo (approved):
Looks good!

PR #5929, opened 2020-10-06
Adoption Application Status Component

dcaballeroSf:
# Critical Changes

# Changes

- Creating a Lightning Component for displaying the Application Status to P10.

# Issues Closed

# Community Ideas Delivered

# New Metadata
Apex Classes:

- GS_ApplicationStatusController
- GS_ApplicationStatusControllerTest

LWC

- gsApplicationStatus
- gsModal

Custom Labels:

- gsApplicationStatus
- gsApplicationStatusModalHeader
- gsApplyForFreeLicenses
- gsCheckStatus
- gsClose
- gsDaysAdded
- gsDaysRemainingInFreeTrial
- gsFollowUpApplicationStatus
- gsLearnMore
- gsNoApplicationSubmitted
- gsSubmitted

Custom Object :
- Testing_Configuration__c

Feature Parameter:

- P10_Application_Date

# Deleted Metadata


dcaballeroSf:
[W-8039006](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008IUiwIAG/view)
[W-7977130](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008IErjIAG/view)
[W-7977149](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008IEsDIAW/view)

src/featureParameters/P10_Application_Date.featureParameterDate
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FeatureParameterDate xmlns="http://soap.sforce.com/2006/04/metadata">
+    <dataflowDirection>LmoToSubscriber</dataflowDirection>
+    <masterLabel>Application to P10 program date retrieved from BizTech</masterLabel>
+    <value>2020-10-23</value>
```
force2b, line 5: Why is this using a relatively arbitrary default value?

src/classes/GS_ApplicationStatusController.cls
```diff
@@ -0,0 +1,176 @@
...
+
+    /** 
+     * @description Queries the Organization object to get the TrialExpirationDate and the IsSandbox field
+     * @returns Organization with TrialExpirationDate and IsSandbox queried.
+     */
+    private Organization getOrgInformation() {
+        return [SELECT TrialExpirationDate,
+                       IsSandbox
+                FROM   Organization
+                WITH_SECURITY_ENFORCED];
```
force2b, line 150: I'm not so sure about using `WITH_SECURITY_ENFORCED` here. This is a system object which technically nobody probably access to. We recently had an issue where standard users were getting exceptions thrown because code running as them was querying `AsyncApexJob`, which I consider at the same system level as this object. Have you verified that a standard user (without any system admin level permissions) can still use this page without error?

src/classes/GS_ApplicationStatusController.cls
```diff
@@ -0,0 +1,176 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Returns if it's running on a Scratch org.
+     * @return true if code runs on a Scratch org, false otherwise
+     */
+    private Boolean isScratchOrg() {
+        Organization org = getOrgInformation();
```
force2b, line 138: There is probably a better implementation of this method in `UTIL_Namespace.isScratchOrg()`. We worked with core to verify an acceptable way to validate this.

src/objects/Testing_Configuration__c.object
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
force2b, line 1: I'm still trying to get through this PR, but why would there be an object called "Testing Configuration"? Is "Testing" what this is actually used for?
Also, a description is required for this object even if it is protected. I suppose that would help with my previous question too.

src/classes/GS_ApplicationStatusController.cls
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * @description Returns the parsed Testing_Configuration.
+     * @return   Map<String, Object> Parsed Testing_Configuration Custom_Setting. it is a json format.
+     * Examples: 
+     * {"Sandbox":false} simulates an Active org.
+     * {"Sandbox":false,"TrialExpiration":"10/15/2020, 00:01 am”} simulates a trial org which is not applied for licenses.
+     * {"Sandbox":false,"TrialExpiration":"10/15/2020","Application":"10/02/2020"} Simulates a trial org which has applied for licenses.
+     */
+    private Map<String, Object> getTestingConfiguration() {
```
force2b, line 123: Is this object strictly used to simulate testing? If so, I'm not sure I'm comfortable with the addition of a custom object for that. Let's talk about other ways to accomplish this without building in a new (never-removable) custom object into the package.

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,25 @@
+@IsTest
```
force2b, line 1: Class and method comment blocks are required

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,25 @@
...
+        insert tc; 
+
+        GS_ApplicationStatusController.ApplicationStatus status = GS_ApplicationStatusController.getApplicationStatus();
+        System.assert(status != null, 'Status should not be null');
+        System.assert(!status.isSandbox, 'Status should not be sandbox');
+        System.assertEquals(Date.parse('10/15/2020'), date.newInstance(status.trialExpirationDate.year(), status.trialExpirationDate.month(), status.trialExpirationDate.day()), 'Trial expiration date not equal');
+        System.assert(status.applicationDate == null, 'application date should be null');
+    }
+
+        @IsTest public static void testGetApplicationStatusTrialOrgApplicated() {
```
force2b, line 15: Method indication is off, plus @IsTest should be on a separate line

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,25 @@
+@IsTest
+public with sharing class GS_ApplicationStatusControllerTest {
+
+    @IsTest public static void testGetApplicationStatusTrialOrgNonApplicated() {
```
force2b, line 4: @IsTest should be on a separate line

src/labels/CustomLabels.labels
```diff
@@ -12731,4 +12731,107 @@ Note: The NPSP Settings tab is visible in the Nonprofit Success Pack application
         <shortDescription>stgValidationHHAccountHHRules</shortDescription>
         <value>The Account Model is set to &apos;Household Account&apos;, which requires Household Rules to be set to &apos;No Contacts&apos;.   When using Household Accounts, there is no need to have an additional Household Object.</value>
     </labels>
+
```
force2b, line 12734: There should be no blank lines between labels, plus these new labels must be placed within the file in the right spot alphabetically.

src/labels/CustomLabels.labels
```diff
@@ -12731,4 +12731,107 @@ Note: The NPSP Settings tab is visible in the Nonprofit Success Pack application
...
+    <labels>
+        <fullName>gsNoApplicationSubmitted</fullName>
+        <categories>gettingStarted</categories>
+        <language>en_US</language>
+        <protected>false</protected>
```
force2b, line 12739: All new custom labels must be protected

src/labels/CustomLabels.labels
```diff
@@ -12731,4 +12731,107 @@ Note: The NPSP Settings tab is visible in the Nonprofit Success Pack application
...
+        <shortDescription>gsNoApplicationSubmitted</shortDescription>
```
force2b, line 12740: All custom labels require a valid description to explain the usage.

src/lwc/gsApplicationStatus/__test__/gsApplicationStatus.test.js
```diff
@@ -0,0 +1,68 @@
+import { createElement} from 'lwc';
```
force2b, line 1: It's great to see Jest tests here!

force2b: Though, comment blocks are needed for the file and methods

src/lwc/gsApplicationStatus/gsApplicationStatus.css
```diff
@@ -0,0 +1,71 @@
...
+.info .slds-button {
+   
+    width: 100%;
+    font-size: 15px;
+    margin-bottom: 24px;
+    padding: 10px;
+}
+
+.slds-button_outline-brand {
+    border-color: purple;
```
force2b, line 38: Why override the button colors? Doesn't this violate SLDS design patterns, plus potentially create accessibility issues with contrast between the background color and text color? There is a reason that we try to stick with standard SLDS styles 99.9% of the time.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,30 @@
...
+    <template if:false={isActiveInstance}>
+        <div class="container slds-is-relative">
+            <div class="spinner" if:true={isLoading}>
+                <lightning-spinner alternative-text="Loading"></lightning-spinner>
+            </div>
+            <div class="image-container"><img class="image" src={img}/></div>
+            <div if:false={isApplicationSubmitted} class="info pending">
+                <p>{labels.gsNoApplicationSubmitted} | <a>{labels.gsLearnMore}</a></p>
+                <p class="daysLeft">{diffInDays} {labels.gsDaysRemainingInFreeTrial}</p>
+                <a class="slds-button slds-button_brand" href="https://powerofus.force.com/PowerOfUsApply" target="_blank">
```
force2b, line 11: Why not just using <lightning-button> as well as all standard SLDS base components in all of this?

force2b (changes requested):
I have a ton of questions and potential issues on this PR. Besides the new custom object, which I'm not really sure should be there, I have to admit I'm getting less and less comfortable with the choice to not use LWC base components in the LWC mock-up. Why was this chosen? We have a policy of always sticking with base components for LWC. Not using them opens up to huge tech debt issues. I feel pretty strongly that this needs to be reconsidered.

src/classes/GS_ApplicationStatusController.cls
dcaballeroSf: This is a Custom Setting. We add this after discussing with Marcel and David Reed. The component has different behavior according if the ORG is a Trial/Active and if the Application for the P10 licenses is available(this is coming from a new feature parameter). The idea is to give the QA a way to injects the values they need to test the paths we have.

force2b: Thanks. I'll check in with David and Marcel on this. Let's hold the PR in the mean time.

src/classes/GS_ApplicationStatusController.cls
dcaballeroSf: This component is going to be seen only by System Administrators, so the WITH_SECURITY_ENFORCED should not be an Issue. I also wrote a test with Run As standard user and the fields were visible. I'm pretty sure that WITH_SECURITY_ENFORCED would cause no harm on this one and I avoid a PMD warning, but it can be removed if required.

src/labels/CustomLabels.labels
```diff
@@ -12731,4 +12731,107 @@ Note: The NPSP Settings tab is visible in the Nonprofit Success Pack application
...
+        <value>Days Remaining in Free Trial</value>
+    </labels>
+
+    <labels>
+        <fullName>gsApplyForFreeLicenses</fullName>
+        <categories>gettingStarted</categories>
+        <language>en_US</language>
+        <protected>false</protected>
+        <shortDescription>gsApplyForFreeLicenses</shortDescription>
+        <value>Apply for Your Salesforce Licenses</value>
```
smarovich: I would make this: Apply for Salesforce Licenses

src/labels/CustomLabels.labels
```diff
@@ -12731,4 +12731,107 @@ Note: The NPSP Settings tab is visible in the Nonprofit Success Pack application
...
+        <value>Close</value>
+    </labels>
+
+    <labels>
+        <fullName>gsFollowUpApplicationStatus</fullName>
+        <categories>gettingStarted</categories>
+        <language>en_US</language>
+        <protected>false</protected>
+        <shortDescription>gsFollowUpApplicationStatus</shortDescription>
+        <value>To follow up on your application status, please email us at: something@salesforce.com</value>
```
smarovich: We need a real email address here. According to the website, it should be powerofus@salesforce.com.

smarovich (changes requested):

src/classes/GS_ApplicationStatusController.cls
dcaballeroSf: I'm thinking now, What do you think if instead of using the CustomSetting, we get the overriding data from a static resource which is not in the package? It will provide testeability now without putting in the package any new object that we wont be able to delete it in the future?

src/lwc/gsApplicationStatus/gsApplicationStatus.css
dcaballeroSf: Regarding this comment, We asked Mary Heid about this and she answered: 

> There are two questions at play here, I believe.
> #1 - does the color meet guidelines? Yes.
> #2 - SLDS design patterns - <I’m not an SLDS developer, but> if we use the SLDS design tokens, when those update, we get those updates for ‘free.’ For example, I believe the the SLDS team is working to update tokens to meet the newer WCAG 2.1 guidelines. While that’s not an issue in this particular use case, it speaks to the point and value of using those tokens. Additionally, I believe the tokens inherently support accessibility features such as inverted colors when a user invokes high contrast mode in their browser/os..

The color does not violates any color guidelines but still has the cons of not having an specific design-token for it, the design was previously approved  (you can talk to Stephanie Zeitz about it), so I would stick to the purple

dcaballeroSf:
I went thru most of the pr comments. @force2b still waiting for confirmation about the Testing_Configuration and my suggestion for finishing this pr

src/featureParameters/P10_Application_Date.featureParameterDate
force2b: @dcaballeroSf Was there a reason this is being populated with a default value? I would expect it to default to an empty/null value and then be populated through the LMO.

src/classes/GS_ApplicationStatusController.cls
force2b: Since this is a system object, where it's not even possible to set security on this object, I recommend removing the `WITH SECURITY ENFORCED` clause in the query.

src/classes/GS_ApplicationStatusController.cls
```diff
@@ -0,0 +1,215 @@
...
+                return testingConfiguration;
+            }
+        }
+        set {
+            testingConfiguration = value;
+        }
+    }
+
+    /**
+     * @description Read-only property with the Org queried with the fields: IsSandbox and TrialExpirationDate.
```
force2b, line 64: Separation of Concerns.
Overall, I think all of the organization state specific methods (isActiveOrg, isSandbox, getExpirationDate, getOrgInformation, etc.) belong in a different class, such as `UTIL_OrgState`. Unit tests would be in a related `UTIL_OrgState_TEST` class. This class would then contain only `@auraEnabled` methods that delegate to the `UTIL_OrgState` class, plus other private static methods that are very specific to the controller operation.

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
+/*
```
force2b, line 1: The format we follow in NPSP for test classes is to use `_TEST` at the end: `GS_ApplicationStatusController_TEST`.

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+ * @author Salesforce.org
+ * @description TestClass for GS_ApplicationStatusController.
+ */
+@IsTest
+public with sharing class GS_ApplicationStatusControllerTest {
+    /**
```
force2b, line 36: blank line needed here

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+        Map<String, Object> testingConfiguration = new Map<String, Object> { 'Sandbox' => false, 
+                                                                             'TrialExpiration' => '10/15/2020, 00:01 AM'};
+
+        GS_ApplicationStatusController.testingConfiguration = testingConfiguration;
+        
+        GS_ApplicationStatusController.ApplicationStatus status = GS_ApplicationStatusController.getApplicationStatus();
+        System.assert(status != null, 'Status should not be null');
+        System.assert(!status.isSandbox, 'Status should not be sandbox');
+        System.assertEquals(Date.parse('10/15/2020'), date.newInstance(status.trialExpirationDate.year(), status.trialExpirationDate.month(), status.trialExpirationDate.day()), 'Trial expiration date not equal');
+        System.assert(status.applicationDate == null, 'application date should be null');
```
force2b, line 50:
```suggestion
        System.assertEquals(null, status.applicationDate, 'application date should be null');
```

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+     */
+    @IsTest 
+    public static void testGetApplicationStatusTrialOrgNonApplicated() {
+        Map<String, Object> testingConfiguration = new Map<String, Object> { 'Sandbox' => false, 
+                                                                             'TrialExpiration' => '10/15/2020, 00:01 AM'};
+
+        GS_ApplicationStatusController.testingConfiguration = testingConfiguration;
+        
+        GS_ApplicationStatusController.ApplicationStatus status = GS_ApplicationStatusController.getApplicationStatus();
+        System.assert(status != null, 'Status should not be null');
```
force2b, line 47:
```suggestion
        System.assertNotEquals(null, status, 'Status should not be null');
```

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+        System.assert(!status.isSandbox, 'Status should not be sandbox');
```
force2b, line 48:
```suggestion
        System.assertEquals(false, status.isSandbox, 'Status should not be sandbox');
```

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+ * @author Salesforce.org
+ * @description TestClass for GS_ApplicationStatusController.
+ */
+@IsTest
```
force2b, line 34: Since this test class issues no DML, let's mark as `IsParallel=true`

src/classes/GS_ApplicationStatusControllerTest.cls
```diff
@@ -0,0 +1,71 @@
...
+       
+
+        Map<String, Object> testingConfiguration = new Map<String, Object> { 'Sandbox' => false, 
+                                                                             'TrialExpiration' => '10/15/2020, 00:01 AM',
+                                                                             'Application' => '10/02/2020'};
+        
+        GS_ApplicationStatusController.testingConfiguration = testingConfiguration;
+        
+        GS_ApplicationStatusController.ApplicationStatus status = GS_ApplicationStatusController.getApplicationStatus();
+        System.assert(status != null, 'Status should not be null');
```
force2b, line 66: Same assert comments from above

src/featureParameters/P10_Application_Date.featureParameterDate
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FeatureParameterDate xmlns="http://soap.sforce.com/2006/04/metadata">
+    <dataflowDirection>LmoToSubscriber</dataflowDirection>
+    <masterLabel>Application to P10 program date retrieved from BizTech</masterLabel>
```
force2b, line 4:
```suggestion
    <masterLabel>P10 Application date</masterLabel>
```

src/labels/CustomLabels.labels
```diff
@@ -7487,6 +7487,94 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Gift Processing configuration error</shortDescription>
         <value>Your org is missing required configuration for Gift Processing. Contact your admin for help.</value>
     </labels>
+    <labels>
+        <fullName>gsApplicationStatus</fullName>
+        <categories>gettingStarted</categories>
```
force2b, line 7492: The category in the other PR is different than this. I believe that one has "Getting Started Page". Please be sure that the category used for all custom labels across all PR's matches.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,37 @@
+<template >
+    <template if:false={isActiveInstance}>
+        <lightning-card>
+            <div class="spinner" if:true={isLoading}>
```
force2b, line 4: Better to use a `<template>` tag to wrap this in for visibility than to use `if:...` in the div tag - throughout this and all component html

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,37 @@
...
+                <lightning-spinner alternative-text="Loading"></lightning-spinner>
+            </div>
+            <div class="slds-grid slds-grid_align-center slds-grid_vertical">
+                <div class="slds-col slds-align_absolute-center">
+                    <img class="image" src={img}/>
```
force2b, line 9: alt text or other accessibility attribute is required here

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,37 @@
...
+                        <div><p class="slds-text-heading_small slds-m-bottom_medium daysLeft">{diffInDays} {labels.gsDaysRemainingInFreeTrial}</p></div>
+                        <div><a tabindex="0" class="slds-button slds-button_brand slds-size_full slds-text-heading_small slds-p-around_medium" href="https://powerofus.force.com/PowerOfUsApply" target="_blank">{labels.gsApplyForFreeLicenses}</a></div>
+                    </div>
+                </div>
+                <div if:true={isApplicationSubmitted} class="slds-col slds-align_absolute-center slds-size_full">
+                    <div class="slds-text-align_center slds-size_full slds-p-around_small">
+                        <div><p class="slds-text-heading_small slds-m-bottom_medium">{labels.gsApplicationStatus}<span class="slds-text-color_success">{labels.gsSubmitted}</span></p></div>
+                        <div><p class="slds-text-heading_small slds-m-bottom_medium daysLeft">{diffInDays} {labels.gsDaysRemainingInFreeTrial}</p></div>
+                        <div><button onclick={onCheckStatusClick}
```
force2b, line 22: lightning-button?

src/lwc/gsApplicationStatus/gsApplicationStatus.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * 
+     * @param {Object} result - Retrieved result object from Salesforce, in order to this to work it has to have a 'trialExpirationDate' field.
+     * @returns remaining days between today and expiration day or -1.
+     */
+    calculateTrialRemainingDays(result) {
+        if (result.trialExpirationDate) {
+            const date = new Date(result.trialExpirationDate)
```
force2b, line 80: semi-colons needed at the end of these lines

src/lwc/gsApplicationStatus/gsApplicationStatus.js
```diff
@@ -0,0 +1,104 @@
...
+        return -1;
+    }
+
+    /**
+     * Check if application has been submitted, using data from Salesforce org.
+     * @param {Object} result the data retrieved from salesforce
+     * @returns True if data has an applicationDate field, false otherwise.
+     */
+    checkApplicationSubmitted(result) {
+        return result.applicationDate != null;
```
force2b, line 94:
```suggestion
        return result.applicationDate !== null;
```

src/lwc/gsModal/gsModal.html
```diff
@@ -0,0 +1,31 @@
...
+                </button>
+                <h2 id="modal-heading-01" class="slds-text-heading_medium slds-hyphenate">{header}</h2>
+             </header>
+             <!-- modal body start -->
+             <div class="slds-modal__content slds-p-around_medium slds-text-align_center slds-p-around_xx-large" id="modal-content-id-1">
+                <slot></slot>
+            </div>
+             <!-- modal footer start-->
+             <footer class="slds-modal__footer">
+                <button class="slds-button slds-button_brand" onclick={closeModal}>{buttonLabel}</button>
```
force2b, line 23: lightning-button?

src/lwc/gsModal/gsModal.html
```diff
@@ -0,0 +1,31 @@
+<template>
+  <!-- modal start -->        
+    <template if:true={bShowModal}>
+       <section role="dialog" tabindex="-1" aria-labelledby="modal-heading-01" aria-modal="true" aria-describedby="modal-content-id-1" class="slds-modal slds-fade-in-open">
+          <div class="slds-modal__container">
+             <!-- modal header start -->
+             <header class="slds-modal__header">
+                <button class="slds-button slds-button_icon slds-modal__close slds-button_icon-inverse" title={butonLabel} onclick={closeModal}>
+                   <lightning-icon icon-name="utility:close"
+                      alternative-text="close"
```
force2b, line 10: There should be an existing custom label you can use for this rather than to hardcode the alternate text. When a non-english user is using a screen reader, we want the page to read the button alternative text in right language.

src/package.xml
```diff
@@ -1346,6 +1348,7 @@
         <members>Schedulable__c.Class_Name__c</members>
         <members>Schedulable__c.Frequency__c</members>
         <members>Schedulable__c.Last_Time_Run__c</members>
+        <members>Testing_Configuration__c.Configuration__c</members>
```
force2b, line 1351: I think we eliminated this object and field from the PR already, correct?

src/package.xml
```diff
@@ -3936,6 +3957,7 @@
     <types>
         <members>CumulusStaticResources</members>
         <members>DefaultGiftEntryTemplate</members>
+        <members>gsResources</members>
```
force2b, line 3960: I recommend spelling out the name of the static resource. I believe this will require renaming the file and folder under StaticResourceSources too.
```suggestion
        <members>GettingStartedResources</members>
```

force2b (changes requested):
This is definitely taking shape, but I did have a lot of comments and requested changes.

Also, I see that the PR has a changed version of the `src/staticresources/CumulusStaticResources.resource` zip file, however there are no changes in the `StaticResourceSources` folder that would explain what changes were made to that zip file. is this a legacy changed file that needs to be removed or were there changes made to that zip file that are missing from the PR?

force2b:
@dcaballeroSf and @fparodi-salesforce There are conflicts on this PR that need to be resolved.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
bethbrains: If the image is decorative only and has no critical meaning, it can have an empty alt tag (`alt=""`) to tell the screen reader to skip it. Otherwise the screen reader will try to read it and it will sound like "Image <filename>" or something like that.

If the image *does* have meaning to the user, the alt text needs to describe it.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
bethbrains: I think you should be able to put the `if:true` directly into the `lightning-spinner` tag.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,37 @@
...
+            </div>
+            <div class="slds-grid slds-grid_align-center slds-grid_vertical">
+                <div class="slds-col slds-align_absolute-center">
+                    <img class="image" src={img}/>
+                </div>
+                <div if:false={isApplicationSubmitted} class="slds-col slds-align_absolute-center slds-size_full">
+                    <div class="slds-text-align_center slds-size_full slds-p-around_small">
+                        <div><p class="slds-text-heading_small slds-m-bottom_medium">{labels.gsNoApplicationSubmitted} | <a tabindex="0">{labels.gsLearnMore}</a></p></div>
+                        <div><p class="slds-text-heading_small slds-m-bottom_medium daysLeft">{diffInDays} {labels.gsDaysRemainingInFreeTrial}</p></div>
+                        <div><a tabindex="0" class="slds-button slds-button_brand slds-size_full slds-text-heading_small slds-p-around_medium" href="https://powerofus.force.com/PowerOfUsApply" target="_blank">{labels.gsApplyForFreeLicenses}</a></div>
```
bethbrains, line 15: You shouldn't need `tabindex` here or up on line 13. What behavior are you seeing that requires special handling?

src/lwc/gsApplicationStatus/gsApplicationStatus.html
bethbrains: Please use `lightning-button` and just put `{labels.gsCheckStatus}` into the `name` attribute.

src/lwc/gsModal/gsModal.html
bethbrains: @force2b what do you think about borrowing PMM's reusable modal component? https://github.com/SalesforceFoundation/PMM/blob/master/force-app/main/default/lwc/modal/modal.html It's quite similar, and has a few extra features that might be useful to NPSP in the future.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,37 @@
+<template >
+    <template if:false={isActiveInstance}>
+        <lightning-card>
+            <div class="spinner" if:true={isLoading}>
+                <lightning-spinner alternative-text="Loading"></lightning-spinner>
+            </div>
+            <div class="slds-grid slds-grid_align-center slds-grid_vertical">
```
bethbrains, line 7: Please use `lightning-layout` and `lightning-layout-item` for layouts, not directly `slds-grid`/`slds-col`. Much easier to read and maintain. https://developer.salesforce.com/docs/component-library/bundle/lightning-layout/example

src/lwc/gsApplicationStatus/__tests__/gsApplicationStatus.test.js
```diff
@@ -0,0 +1,94 @@
...
+            const button = element.shadowRoot.querySelector('.slds-button');
+            expect(button.innerHTML).toBe('c.gsCheckStatus');
+            const daysLeft = element.shadowRoot.querySelector(".daysLeft");
+            expect(daysLeft.innerHTML).toContain('15');
+            const img = element.shadowRoot.querySelector('img').getAttribute('src');
+            expect(img).toBe('CumulusStaticResources/gsResources/gift_illustration_2.svg');
+        })
+        
+    });
+});
```
bethbrains, line 94: Could you import and run `sa11y`? https://developer.salesforce.com/blogs/2020/10/automated-accessibility-testing-with-sa11y.html

bethbrains (changes requested):

src/lwc/gsApplicationStatus/gsApplicationStatus.html
dcaballeroSf: I can't on this one. The button design is not a standard salesforce lightning-button like. All the variants make the button look different than the design

src/lwc/gsApplicationStatus/gsApplicationStatus.html
dcaballeroSf: On Chrome without the tabindex I was not reaching the elements using tab key.

bethbrains:
Is there a spot this is rendering in the UI yet so I can take a look?
Thanks!

src/lwc/gsApplicationStatus/gsApplicationStatus.html
bethbrains: Is there a spot in the UI where this component is rendering yet, so I can take a look? Thanks!

force2b:
@dcaballeroSf There are 140 changed files in this PR, as well as some merge conflicts. Looks like the branch is out of sync which is causing the mass number of changed files.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
dcaballeroSf: It is a component for a lightning page, there is no page yet though

src/lwc/gsApplicationStatus/__tests__/gsApplicationStatus.test.js
dcaballeroSf: I added sa11y and locally it works. But i broke the build on github.

src/classes/GS_NonprofitTrialOrgService.cls
```diff
@@ -0,0 +1,173 @@
...
+*/
+/**
+ * @date 2020
+ * @group Getting Started
+ * @author Salesforce.org
+ * @description Service class for getting the Licensing Status of the org
+ */
+public with sharing class GS_NonprofitTrialOrgService {
+
+    private static final String STATIC_RESOURCE_NAME = 'gsApplicationConfiguration';
```
force2b, line 38:
```suggestion
    private static final String STATIC_RESOURCE_NAME = 'GettingStartedTestingConfigurationJSON';
```

src/classes/GS_NonprofitTrialOrgService.cls
```diff
@@ -0,0 +1,173 @@
...
+    private static final String APPLICATION_DATE = 'P10_Application_Date';
+    private static final String SANDBOX_PARAMETER = 'Sandbox';
+    private static final String APPLICATION_PARAMETER = 'Application';
+    private static final String TRIAL_EXPIRATION_PARAMETER = 'TrialExpiration';
```
force2b, line 42: Rather than deserialize this structure into a map, I recommend using a simple private inner class:
```
private class TestingConfig {
    public Date applicationDate;
    public Boolean isSandbox;
    public String application;
    public Date trialExpiration;
}
```
Then just deserialize:
```
TestingConfig testConfig = (TestingConfig)JSON.deserialize(staticResources[0].Body.toString(), TestingConfig.class);
```

And you can reference properties without type casting:
`Date applicationDate = testConfig.applicationDate;`

This would be significantly easier to follow in the code as well as debug and test.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -0,0 +1,46 @@
+<template >
+    <template if:false={isActiveInstance}>
+        <lightning-card>
+            <div class="spinner" >
```
force2b, line 4: Why wrap the spinner in a div with a class of spinner? I don't see a custom CSS style for spinner? Does this do something else?

force2b (changes requested):
Taking shape, but there are some thing that need to be adjusted:
- Avoid date.parse()
- Use a private class to deserialize the testing config json
- Some LWC questions
- This PR still contains a changed file for `src/staticresources/CumulusStaticResources.resource`, yet there are no changed files in the `StaticResourceSources` folder to match the change to that resource. Either something is missing from this PR or that static resource should not be here. That must be resolved before this PR can be merged.
- I believe I suggested renaming the static resource in another PR (probably to `GettingStartedResources`). There is no value in using `gs` versus `GettingStarted` when it comes to the static resource. If anything using `gs` makes it harder to know what the files/resource is for.

dcaballeroSf:
@force2b I see 4 new files in StaticResourcesSources which are the causing of the update in the static resource

force2b:
@dcaballeroSf Sorry. I forgot that the `StaticResourcesSources` older is just one big folder with everything in the `CumulusStaticResources` zip file. So, in this case, your 4 files in the `StaticResourceSources/gsResources/` folder are all in the main `CumulusStaticResources` zip file. That's fine. No need to move those or rename anything.

src/classes/GS_NonprofitTrialOrgService.cls
```diff
@@ -0,0 +1,182 @@
...
+        } else {
+            return org.TrialExpirationDate;
+        }
+    }
+
+    /**
+     * @description Returns a Map<String,Object> retrieved from a Static Resource called 'GettingStartedTestingConfigurationJSON'
+     * @return   TestingConfig testing configuration after parsing the Static Resource Json
+     * Examples: 
+     * {"Sandbox":false} simulates an Active org.
```
force2b, line 136: These comments need updating to match the format of the `TestingConfiguration` inner class properties.

force2b (changes requested):

src/lwc/modal/__tests__/modal.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+        // Return a promise to wait for any asynchronous DOM updates. Jest
+        // will automatically wait for the Promise chain to complete before
+        // ending the test and fail the test if the promise rejects.
+        return global.flushPromises().then(async () => {
+            // Query header slot element
+            const headerSlotEl = element.shadowRoot.querySelector('slot[name="header"]');
+            expect(headerSlotEl).not.toBeNull();
+
+            // TODO: Determine if we can require either the slot or header be populated?
```
force2b, line 48: Are these todo's still open or are these intended to be resolved in another PR?

src/lwc/modal/modal.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata" fqn="modal">
+    <apiVersion>46.0</apiVersion>
```
force2b, line 3: This should be api 49. (Or 50 if you're setting all class/components in the PR to 50). They all have to match at either 49 or 50.

force2b (changes requested):

src/lwc/modal/__tests__/modal.test.js
dcaballeroSf: No idea, I'm copying the modal from CaseMan (https://github.com/SalesforceFoundation/CaseMan/tree/master/force-app/main/default/lwc/modal) as Beth recommendation

force2b: Fair enough. @bethbrains Any concerns if we delete the todo from this component or is it worth keeping as a reminder that something eventually needs to be done?
Also @dcaballeroSf Can you add a comment block to the top of the JS file for the parent modal component to indicate that it was copied from CaseMan just so we remember in the future?

src/lwc/modal/modal.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata" fqn="modal">
+    <apiVersion>49.0</apiVersion>
```
force2b, line 3: @bethbrains What would be a good label/description to put into this component that isn't specific to the adoption pages?

src/lwc/modal/__tests__/modal.test.js
bethbrains: yeah, it should be fine to remove the TODO. @jjbennett likely wrote this test.

src/lwc/modal/modal.js-meta.xml
bethbrains: @force2b it's not exposed so it doesn't need a label/descrip. Is it NPSP's practice to include them regardless? It would serve as dev docs only.

bethbrains (changes requested):

src/lwc/modal/modal.js-meta.xml
force2b: We do try to include them, even as dev docs. In this case, other teams wouldn't necessarily know that this exists as a shared component that can be used outside of the getting started page context.

bethbrains: Got it. Then I'd say label is `Generic Modal` and description could be `A generic, reusable modal component that follows SLDS since there is no lightning base component for this yet.` Does that sound ok to you?

force2b: Thank you @bethbrains 
FYI @dcaballeroSf

force2b (approved):
Very nice. Thank you for working through all the changes with us @dcaballeroSf

src/labels/CustomLabels.labels
```diff
@@ -7583,6 +7591,94 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <protected>true</protected>
+        <shortDescription>Header of the modal label for the application status component</shortDescription>
+        <value>Check Application Status</value>
+    </labels>
+    <labels>
+        <fullName>gsApplyForFreeLicenses</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Apply for licencses button label for the application status component</shortDescription>
```
smarovich, line 7615: Typo in this short description: misspelled "licenses"

src/labels/CustomLabels.labels
```diff
@@ -7583,6 +7591,94 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <protected>true</protected>
+        <shortDescription>Days added label for the application status component</shortDescription>
+        <value>30 Days Added</value>
+    </labels>
+    <labels>
+        <fullName>gsDaysRemainingInFreeTrial</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>gsDaysRemainingInFreeTrial</shortDescription>
```
smarovich, line 7647: I'd recommend using a short desc that is similar to the others: Days remaining in the free trial label for the application status component

bethbrains (approved):
i left a question in quip for @force2b re: yarn.lock, but otherwise this is looking good to me.

force2b:
Reading up on what `yarn.lock` is used for, I think changes to it should be committed to the repo.

davidmreed:
> Reading up on what `yarn.lock` is used for, I think changes to it should be committed to the repo.

`yarn.lock` should be committed, since there is a change to `package.json`. `package-lock.json` should _not_ be committed, because we use `yarn` instead of `npm`.

dcaballeroSf:
I removed the package-lock.json, are we good to merge it?

src/labels/CustomLabels.labels
```diff
@@ -7716,16 +7716,24 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <shortDescription> Days remaining in the free trial label for the application status component</shortDescription>
         <value>Days Remaining in Free Trial</value>
     </labels>
+    <labels>
+        <fullName>gsEmailAddress</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Power of us email address</shortDescription>
+        <value>powerofus@salesforce.com</value>
```
force2b, line 7728: I'm curious why some URL's are hardcoded into the html file and other URL's (or emails) are handled as a custom label? There should be consistency throughout all of the Getting Started pages on how these are handled. Changes to the email address in the future can be made just as easily in either area.

src/lwc/gsApplicationStatus/gsApplicationStatus.html
```diff
@@ -1,51 +1,44 @@
-<template >
+<template>
     <template if:false={isActiveInstance}>
-        <lightning-card>
+        <div class="container">
```
force2b, line 3: Why go back to the non base component structure?

src/lwc/gsApplicationStatus/gsApplicationStatus.js-meta.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
-    <masterLabel>Component that displays the Trial Org and Application Status</masterLabel>
+    <masterLabel>NPSP Application status</masterLabel>
+    <description>Component that displays the Trial Org and Application Status</description>
     <apiVersion>49.0</apiVersion>
```
force2b, line 5: Noting that there is a mix of API versions throughout the Getting Started pages. I'd prefer to see everything at the same api level. 50 is fine at this point.

src/labels/CustomLabels.labels
```diff
@@ -816,6 +816,14 @@
         <shortDescription>CampaignMemberStatusOmit</shortDescription>
         <value>omit</value>
     </labels>
+    <labels>
+        <fullName>Close</fullName>
```
force2b, line 820:
```suggestion
        <fullName>btnClose</fullName>
```

src/labels/CustomLabels.labels
```diff
@@ -816,6 +816,14 @@
...
+        <categories>Error</categories>
```
force2b, line 821:
```suggestion
        <categories>Button</categories>
```

src/labels/CustomLabels.labels
```diff
@@ -816,6 +816,14 @@
...
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Close button description</shortDescription>
```
force2b, line 824:
```suggestion
        <shortDescription>Close button label</shortDescription>
```

src/labels/CustomLabels.labels
```diff
@@ -7655,6 +7663,102 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+    <labels>
+        <fullName>gsCheckStatus</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Check Status label for the application status component</shortDescription>
+        <value>Check Status</value>
+    </labels>
+    <labels>
+        <fullName>gsClose</fullName>
```
force2b, line 7699: Why have two labels for a button label for close? There was one above with an api name of "Close" that I requested to be changed to "btnClose". Why not just have one label for "btnClose" that is used **everywhere** for a close button?

force2b (changes requested):

src/lwc/gsApplicationStatus/gsApplicationStatus.html
dcaballeroSf: Design needed a shadow which was not able to make it using a lightning-card

src/labels/CustomLabels.labels
force2b: @dcaballeroSf What is your thought on this?

dcaballeroSf: @force2b done

src/labels/CustomLabels.labels
```diff
@@ -7655,6 +7663,110 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <shortDescription>Header of the modal label for the application status component</shortDescription>
+        <value>Check Application Status</value>
+    </labels>
+    <labels>
+        <fullName>gsApplicationUrl</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Url of power of us</shortDescription>
+        <value>https://www.salesforce.org/power-of-us/</value>
```
force2b, line 7688: I was thinking you'd move the email to the html. I recommend not putting URL's and email addresses in custom labels. This has been called out as security issue because a malicious user  could modify a URL (by using the Overrride option in the org) so that any other users that click that link redirect them to a website that does something the user did not intend it to do. In other areas where we've hardcoded URL's (primarily in older visualforce pages), we actually push those through an allowlist component that strips out unsupported URL's. For an LWC though, we should just put the URL into the HTML. It's easily updated in either location.

force2b (changes requested):

src/labels/CustomLabels.labels
davidmreed: Agreed, we really should not use Custom Labels for anything other than a translatable string.

src/labels/CustomLabels.labels
```diff
@@ -7655,6 +7663,102 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <shortDescription>No application submitted label for the application status component</shortDescription>
+        <value>No Application Submitted</value>
+    </labels>
+    <labels>
+        <fullName>gsPowerOfUsUrl</fullName>
+        <categories>Getting Started Page</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Url of power of us</shortDescription>
+        <value>https://www.salesforce.org/power-of-us/</value>
```
force2b, line 7744: I thought you were moving both the email and urls into the html?

force2b (changes requested):

force2b (approved):

PR #5936, opened 2020-10-09
GE, RD2: Consolidate common payments services code

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_GiftEntryController.cls
```diff
@@ -351,7 +346,7 @@ public with sharing class GE_GiftEntryController {
 
             return JSON.serialize(opportunityWrappers);
         } catch (Exception ex){
-            returnAuraException(ex.getMessage());
+            UTIL_AuraEnabledCommon.buildAuraException(ex.getMessage());
```
RadSFDeveloper, line 349: Should this statement include the `throw` exception instead of just getting/building it?

src/classes/GE_GiftEntryController.cls
```diff
@@ -964,7 +959,7 @@ public with sharing class GE_GiftEntryController {
             Custom_Column_Header__c columnHeader = new Custom_Column_Header__c();
             columnHeader.Field_Api_Name__c = field;
             columnHeader.Index__c = index;
-            columnHeader.Name = generateId();
+            columnHeader.Name = UTIL_String.getGUID();
```
RadSFDeveloper, line 962: Using the new `UTIL_String.getGUID()` and removed the same code implemented in `generateId()`.

src/classes/PS_IntegrationServiceConfig_TEST.cls
```diff
@@ -161,17 +103,17 @@ public with sharing class PS_IntegrationServiceConfig_TEST {
         PS_IntegrationService.setConfiguration(testConfig);
 
         Test.startTest();
-        GE_PaymentServices service = new GE_PaymentServices();
+        PS_IntegrationServiceConfig.Service service = new PS_IntegrationServiceConfig.Service();
```
RadSFDeveloper, line 106: The `PS_IntegrationServiceConfig.Service` replaced the configuration getter methods as `GE_PaymentServices` returned it before.

src/classes/UTIL_Http.cls
```diff
@@ -54,21 +63,23 @@ public without sharing class UTIL_Http {
...
         * @param request HttpRequest
         * @return Response HttpResponse wrapper
         */
         public Response sendRequest(HttpRequest request) {
-            UTIL_Http.Response response = new UTIL_Http.Response();
+            Response response;
 
             try {
                 HttpResponse requestResponse = new Http().send(request);
+
```
RadSFDeveloper, line 75: In order to test the Payments Services request timed out response, I could not test it without adding the timeout value in ms before the request is sent, ie before line 74 as `request.setTimeout(1000);`. 

The Gift Entry error handling is as expected:
GEPurchaseCalloutTimedOut

The RD entry form error handling is as expected:
RDCommitmentCalloutTimedOut

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -254,12 +236,35 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
...
     }
 
-    dispatchApplicationEvent (eventName, payload) {
+    /**
+     * @description Returns the cardholder name.
+     * If the Donor Type = Account, the first and last name are empty.
+     * If the Donor Type = Contact, the Contact first name (if specified) 
+     * and the last name are provided to the Payments Services card tokenization iframe.
+     */
+    getCardholderName() {
```
RadSFDeveloper, line 253: The cardholder name is passed to the card tokenization process. It is constructed with the contact's first name (if the contact has it) and the last name if the Donor Type = Contact. If the Donor Type = Account, an empty cardholder name is passed to the tokenization process with used "Not provided" as default for the card tokenization.

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -118,19 +118,23 @@ class psElevateTokenHandler {
...
 
         const tokenPromise = new Promise((resolve, reject) => {
 
             const timer = setTimeout(() =>
-                reject(this.handleTimeout(handleError)),
+                reject(handleError({
+                    error: this.labels.tokenRequestTimedOut,
+                    isObject: false
+                })),
                 TOKENIZE_TIMEOUT_MS
```
RadSFDeveloper, line 138: When testing the tokenization timeout, the `TOKENIZE_TIMEOUT_MS` value can be set from 10000 to 1000. 

The Gift Entry displays the tokenization timeout error:
GETokenTimedOut


The RD form displays the tokenization error:
RDTokenTimedOut

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -82,7 +82,10 @@ export default class rd2ElevateCreditCardForm extends LightningElement {
     * @description Method sends a message to the visualforce page iframe requesting a token.
     */
     requestToken() {
+        this.clearError();
```
RadSFDeveloper, line 85: When an invalid card is entered,  an error is displayed on the RD form as expected. If the Amount is changed to a negative number but the credit card information is fixed, the error on the Elevate credit card widget should not be displayed.
@rponti-sforg

src/pages/GE_TokenizeCard.page
```diff
@@ -18,7 +18,7 @@
         let domain = arr[0].replace('--npsp', '');
         domain = domain.replace('--c', '');
         const lightningOrigin = `https://${domain}.lightning.force.com`;
-        let nameOnCard = 'test name';
+        let nameOnCard = 'Not provided';
```
RadSFDeveloper, line 21: Since the VF page is changed in this PR, used the opportunity to change the default value from `test name` to `Not provided`. I can revert back the constant if that is required.

src/pages/GE_TokenizeCard.page
```diff
@@ -69,10 +69,14 @@
             if (event && event.origin !== lightningOrigin) {
                 // Reject any messages from an unexpected origin
                 return;
+
             } else if (event.data && event.data.action === 'createToken') {
                 if (event.data.nameOnCard) {
                     nameOnCard = event.data.nameOnCard;
+                } else {
+                    nameOnCard = 'Not provided';//default
```
RadSFDeveloper, line 77: The name needs to be reset every time the card tokenization is required. See AC3 in the WI for the scenario.

unpackaged/config/ps_testing/classes/PS_TestPaymentService.cls
```diff
@@ -36,13 +36,28 @@
 */
 public with sharing class PS_TestPaymentService {
 
-    private static final String BASE_PURCHASE_CALL_BODY = '{"amount":25,"email":"test@test.test","firstName":"John","lastName":"Doe","metadata":{},';
-    private static final String TOKEN_FIELD_NAME = 'token';
-    private static final String STATUS_FIELD_NAME = 'status';
+    private static final String BASE_PURCHASE_CALL_BODY = '{"amount":2500,"email":"test@test.test","firstName":"John","lastName":"Doe","metadata":{},';
```
RadSFDeveloper, line 39: The amount must be multiplied with the currency denominator in order for Elevate to process a credit card charge (the $.25 cannot be processed).

RadSFDeveloper (commented):
These comments give feedback about changes done and testing scenarios. There is one question though about the exception.

RadSFDeveloper:
Tracking [W-8120905](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008fsf8IAA/view)

RadSFDeveloper:
> Hi @RadSFDeveloper - I am looking at this but have a couple broader questions just for context. There's more updated lines to some of the Gift Entry pieces than I was expecting - GE_GiftEntryController.cls and geFormWidgetTokenizeCard as examples. Do you have a sense for whether these changes are mostly style updates, renaming and moving methods around ... or are they more structural?
> 
> Also, does the RD2 form use the geFormWidgetTokenize card LWC?

No prob, let me clarify change. Let me know if there are questions. 
- This PR is deleting methods and updating `geFormWidgetTokenize` to use common `psElevateTokenHandler.js` to get the VF domain, URLs, tokenize card, request token. 
- The RD2 form is using a separate credit card widget LWC `rd2ElevateCreditCardForm` which is calling common `psElevateTokenHandler.js` to get the VF domain, URLs, tokenize card, request token.  
- `GE_PaymentsServices` is updated to use common `PS_Request` and `UTIL_Http` to build a purchase request, send it and return the response. 
- In addition to consolidating code, there is change in behaviour for the GE form: The credit card holder name is passed to the credit card tokenization when Donor Type = Contact. I can revert it back to pass nothing to the tokenization, and then "Not provided" will be always used for the GE.
- Both GE and RD2 are passing "Not provided" to the tokenization instead of "test name" when the Donor Type = Account.

What needs to be retested is:
- GE form continues to work as before with regards to form display and error handling on an org connected to Elevate and on org not connected to Elevate. (full regression testing)
- GE form passes the contact name to the tokenization process when Donor Type = Contact
- GE and RD2 forms pass "Not provided" to the tokenization process when Donor Type = Account
- RD2 form continues to work as before on an org connected to Elevate and on org not connected to Elevate.

Also, if there is a better way to rewrite the common front-end code for the tokenization and common back-end code to handle purchase, and create/edit/cancel commitment requests and response, please let me know.

src/classes/GE_GiftEntryController.cls
```diff
@@ -60,63 +64,74 @@ public with sharing class GE_GiftEntryController {
...
+            throw buildDmlException(JSONExceptionData);
 
-            if (hasTimedOut) {
-                return getPurchaseCallTimeoutResponse(ex, dataImportRecordId);
-            } else {
-                String JSONExceptionData = ERR_ExceptionData.createExceptionWrapperJSONString(ex);
-                throw returnDmlException(JSONExceptionData);
-            }
+        } finally {
+            processPurchaseTimeoutResponse(response, dataImportRecordId);
```
ptewson-sfdo, line 93: Just to make sure I'm tracking this, these changes update the criteria we are using to identify a timed out response -  formerly the purchase was considered to have timed out when the CalloutException was thrown and the "timed out" text appeared in the message; after these changes it's looking at the response code of the response. I just want to make sure this shift is both intentional, and still in line with how we intended to process timeouts (CC @psadie-sfdo)

src/classes/GE_GiftEntryController.cls
```diff
@@ -60,63 +64,74 @@ public with sharing class GE_GiftEntryController {
...
-    *
-    * @return String: http response wrapper
+    * @return void
     */
-    private static String getPurchaseCallTimeoutResponse(Exception ex, String dataImportRecordId) {
-        GE_PaymentServices.ResponseWrapper response = new GE_PaymentServices.ResponseWrapper();
-        response.statusCode = TIMEOUT_STATUS_CODE;
-        response.status = ex.getMessage();
+    private static void processPurchaseTimeoutResponse(UTIL_Http.Response response, String dataImportRecordId) {
+        if (response.statusCode != UTIL_Http.STATUS_CODE_REQUEST_TIMEOUT) {
```
ptewson-sfdo, line 106: I think the description block of the header for this method should be updated to include this behavior, that the method returns without further action when the response status code is not STATUS_CODE_REQUEST_TIMEOUT

src/classes/GE_GiftEntryController.cls
ptewson-sfdo: Probably yes ... I don't think this catch block has much utility without throwing that exception (CC @kenlewis-sfdo )

src/classes/GE_GiftEntryController.cls
ptewson-sfdo: I think this changer makes sense? (CC @psadie-sfdo )

src/classes/GE_GiftEntryController_TEST.cls
```diff
@@ -511,6 +567,48 @@ private with sharing class GE_GiftEntryController_TEST {
...
+        return (Id)dataImportBatch.Id;
+    }
+
+    /***
+    * @description creates a new di record for tests
+    * @param firstname the firstname to specify for Contact1
+    * @param lastname the lastname to specify for Contact1
+    * @return DataImport__c the new Data Import record
+    */
+    public static DataImport__c newDI(String firstname, String lastname, Id batchId) {
```
ptewson-sfdo, line 603:
```suggestion
    public static DataImport__c createDataImport(String firstname, String lastname, Id batchId) {
```

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1004,10 +1004,19 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
      * @param event The paymentError event object
      */
     handleAsyncWidgetError(event) {
-        let errorResponse = isNotEmpty(event.error.message[1]) ? event.error.message[1] : null;
+        let errorMessage = this.CUSTOM_LABELS.commonUnknownError;
```
ptewson-sfdo, line 1007: I thiiiiiiink this makes sense but just wanted to tag @psadie-sfdo and @kenlewis-sfdo to make sure!

ptewson-sfdo (commented):
I left a handful of comments and couple questions, but overall this looks like a nice shift to shared infrastructure. I tagged a couple of the honeygems devs on a few of the items that I wanted their expertise on, but overall not seeing anything big to address here!

src/pages/GE_TokenizeCard.page
ptewson-sfdo: I like making the switch here, but I think the value needs to match what Elevate is expecting, which I believe is "[Not Provided]". I'm not actually clear on if the string we use needs to match exactly some expected value, but to be safe can we use that string (with the square brackets) until we hear otherwise?

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -20,75 +24,95 @@ import {
...
 
-const TOKENIZE_TIMEOUT = 10000; // 10 seconds
-const TOKENIZE_CARD_PAGE_NAME = 'GE_TokenizeCard';
-
 export default class geFormWidgetTokenizeCard extends LightningElement {
     @api cardHolderName;
 
     @track domain;
     @track visualforceOrigin;
+    visualforceOriginUrls;
```
ptewson-sfdo, line 32:
```suggestion
    _visualforceOriginUrls;
```

ptewson-sfdo: Our team prefers to flag internal properties of LWC components with the "_" prefix - this is in line with the style that 
 the LWC team at Salesforce uses when developing their components.

src/classes/GE_GiftEntryController.cls
psadie-sfdo: This looks good to me!

src/classes/GE_GiftEntryController.cls
RadSFDeveloper: The new [UTIL_Http.sendRequest()](https://github.com/SalesforceFoundation/NPSP/blob/70a0f7503bcc97841671f10542d4838c7e0e55da/src/classes/UTIL_Http.cls#L91) and the [timeout exception](https://github.com/SalesforceFoundation/NPSP/blob/70a0f7503bcc97841671f10542d4838c7e0e55da/src/classes/UTIL_Http.cls#L92) are created based on the original GE controller code. Difference is in the response `status` and `body` construction. The `sendRequest()` timeout error response `body` contains the exception error message, however, the GE controller builds the timeout response so that status contains the error message and the response `body` contains a generic timed out error. Thus, additional processing is done in the `processPurchaseTimeoutResponse()`.

So, `sendMessage()` sends the request and :
1.  if the Elevate returns a success response, the method just returns the response
2.  if the Elevate returns an error response, the method just returns the response
3.  if the Elevate returns a timed out response, the method just returns the response
4.  if the Callout exception is generated and the error message contains "timed out", a timed out response with status code 408, status "Request Timeout" and body the exception error message is returned.
5.  if any unexpected exception is generated and it is not a timed out, then a  bad request response is returned. I think this is different: a DML exception is returned currently in `master` on an unexpected send request exception.

GE `processPurchaseRequest()`:
- continues to build purchase request body and sends the purchase request.
- continues to throw DML exception on an unexpected error except in the above case (5)
- the format of the timed out error response and its conditions should not be changed.

Based on your comment, I thought of changing the code to not use `UTIL_Http.sendRequest()` but to use `HttpResponse requestResponse = new Http().send(request);` as before and built timed out response. But I think the only issue is the case (5), bad request response over DML exception when sending the request. Let me know what you think, I can update code not to use common `UTIL_Http.sendRequest()`. @psadie-sfdo , @patrick-yan-sf

src/lwc/geFormRenderer/geFormRenderer.js
RadSFDeveloper: thanks, it would be great if this method is double or triple checked that it is coded as expected.
The method would fail (in console, the Elevate card widget would just display the spinner) if the error message is not an array. I set tokenization timeout to be 1000. I did reproduce the same issue on `feature/RD2`, but have not tested if the error is reported when timeout is 1000 on `master`.

GETokenTimeoutErrorNotDisplayedOnValidCard_Main

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
RadSFDeveloper: The var is used now in the [psElevateTokenhandler](https://github.com/SalesforceFoundation/NPSP/blob/70a0f7503bcc97841671f10542d4838c7e0e55da/src/lwc/psElevateTokenHandler/psElevateTokenHandler.js#L92). Should I stil update the var to `_visualforceOriginUrls`?

src/lwc/geFormRenderer/geFormRenderer.js
RadSFDeveloper: The same issue is reproducible on an org built off the `master`.
Screen Shot 2020-10-13 at 6 31 08 PM

src/classes/GE_GiftEntryController.cls
psadie-sfdo: This looks good to me! I like UTIL_Http and would prefer that we keep using it moving forward. From what I'm seeing we're still logging the unexpected exceptions/errors in the Data Import record and also outputting an error to the UI.

The only thing I'd like to comment on here is with regards to `processPurchaseTimeoutResponse`. I'm not sure about always be making this call regardless of the status of the response. I would perhaps consider pulling the check on the response status code out and perform that check before calling `processPurchaseTimeoutResponse`.

`
if (response.statusCode == UTIL_Http.STATUS_CODE_REQUEST_TIMEOUT) {
                processPurchaseTimeoutResponse(response, dataImportRecordId);
            }
`

RadSFDeveloper: Thanks for the feedback @psadie-sfdo ! I understand your point, previously the timeout error was processed on the callout exception but now it is every time on the purchase response validation. 

In the previous code version, the timeout as the result of the callout exception with the timed out message and the timeout returned from Elevate API differed, but now every timeout response (status code 408) will have status containing the error message and the response body containing the generic timed out error returned to the LWC. Does that present any issue?

src/classes/GE_GiftEntryController.cls
```diff
@@ -37,17 +37,21 @@
...
-        String url = System.Url.getOrgDomainUrl().getHost();
-        String[] urlParts = url.split('\\.');
-        String podName = [SELECT InstanceName FROM Organization].InstanceName.toLowerCase();
-        return new OrgDomainInfo (urlParts[0], podName);
+
+    /*******************************************************************************************************
+    * @description Indicates if the org is connected to Elevate
+    * @return Boolean
+    */
+    @AuraEnabled
```
force2b, line 48: Could this be cachable?

RadSFDeveloper:
@ptewson-sfdo , @psadie-sfdo , @force2b : the code is ready for review again. Thanks.

src/classes/GE_PaymentServices.cls
```diff
@@ -54,142 +48,142 @@ public with sharing class GE_PaymentServices {
...
+    /**
+     * @description Checks whether an org is a customer of Elevate Payment Services.
+     *              After onboarding within Elevate, these configuration key-value
+     *              pairs should exist in the org.  If this required set of keys is not
+     *              in the org, NPSP cannot load the Elevate SDK or successfully make
+     *              callouts to Elevate services.
+     */
+    public static Boolean isElevateCustomer {
+        get {
+            PS_IntegrationServiceConfig ps = new PS_IntegrationServiceConfig();
```
force2b, line 74: I recommend storing a static instance of this getter to avoid re-instantiating PS_IntegrationServiceConfig() every time. It does look like the PS_IntegrationServiceConfig.config class is static, so I'm pretty sure this won't cause additional queries, but I do think that avoiding a new instantiation on multiple calls to this method will improve performance, even if just slightly.

src/classes/PS_Request.cls
```diff
@@ -198,10 +215,10 @@ public inherited sharing class PS_Request {
         * @param commitmentId Recurring Donation Commitment Id (required for cancel commitment endpoint only)
         * @return String Full endpoint
         */
-        private String getEndpoint(ElevateEndpoint endpoint, String commitmentId) {
+        public String getEndpoint(ElevateEndpoint endpoint, String commitmentId) {
             String value;
             if (endpoint == ElevateEndpoint.TOKENIZE) {
-                value = '/v1/payments/public/tokens';
+                value = '/v1/payments/verified/tokens';
```
force2b, line 221: Does it make sense to move all endpoint URL's to constants at the top of the class so that they're easier to find/review/change in the future?

src/classes/GE_GiftEntryController.cls
RadSFDeveloper: @ptewson-sfdo ^

src/classes/GE_PaymentServices.cls
RadSFDeveloper: @ptewson-sfdo , @force2b : The `isElevateCustomer` is unchanged. I would keep it as is. Let me know.

src/classes/GE_GiftEntryController.cls
ptewson-sfdo: I believe so! The impact might be more on the testing side, where an update to the elevate parameters in the back-end might not immediately make the UI update until that cache is updated. I wouldn't mind making this function cacheable however.

src/classes/PS_Request.cls
RadSFDeveloper: It has been done.
@rponti-sforg , Cancel Commitment needs to be retested (happy path).

src/classes/GE_PaymentServices.cls
ptewson-sfdo: I could go either way (or we could also store a static instance of the PS_IntegrationServiceConfig class).... Mike's suggestion is a good one, but I also think the impact is probably minimal. Given that there's a complex QA process for this work and a tightening window, I'm maybe more on Radmila's side to minimize changes where the impact isn't thought to be high.

src/classes/GE_GiftEntryController.cls
RadSFDeveloper: The method was set as `(cacheable=true)`.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
ptewson-sfdo: Interesting ... should we just move this property to psElevateTokenHandler as well? I think the only use for it was in evaluated the origin of the post message, which is now being done in psElevateTokenHandler. It seems like it should be a 'private' property of that component?

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -108,119 +132,81 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
...
-                            [DATA_IMPORT_PAYMENT_STATUS_FIELD.fieldApiName]: this.PAYMENT_TRANSACTION_STATUS_ENUM.PENDING
-                        });
-                    }
-                };
+        const iframe = this.template.querySelector(`[data-id='${this.CUSTOM_LABELS.commonPaymentServices}']`);
 
-            });
+        //The cardholder name is always empty for the purchase Payments Services card tokenization iframe
+        //even though when it is accessible by the Gift Entyr form for Donor Type = Contact.
+        const nameOnCard = null;
```
RadSFDeveloper, line 160: @ptewson-sfdo , @psadie-sfdo , @kenlewis-sfdo , @deepmalamehta :
The GE cardholder name will always be a default. The change in behaviour is to use different constant in the `GE_TokenizeCard` VF page, ie  replace `test name` with `[Not Provided]`. Please let me know if there are any questions or additional change should be done.

src/classes/GE_PaymentServices.cls
RadSFDeveloper: @kenlewis-sfdo  has very valuable comments how to improve the class and initialize class variables. It is most beneficial to apply the change in a new tech debt/refactoring WI.

src/classes/PS_Request.cls
```diff
@@ -64,9 +64,31 @@ public inherited sharing class PS_Request {
...
+    public static String ElevateEndpointCommitmentCancel = '/v1/payments/verified/commitments/{0}/cancel';
 
     private static Integer RECOMMENDED_TIMEOUT_MS = 20000;
 
+    /***
+    * @description Returns payments service endpoint URL
+    * @param endpoint Elevate Endpoint enum 
+    * @return String
+    */
+    public String getEndpoint(ElevateEndpoint endpoint) {
```
RadSFDeveloper, line 80: @kenlewis-sfdo , feel free to create a new tech debt/refactoring WI for the common class renaming, for example `PS_Request` -> `PS_PaymentRequest` . Renaming can be done on any purchase (`GE_PaymentsServices`) or commitment request  (`PS_CommitmentRequest`) related work. Or just feel free to rename/refactor classes (if at all) when working on a PR that changes `GE_PaymentsServices` class. CC @psadie-sfdo , @ptewson-sfdo

src/classes/GE_GiftEntryController.cls
```diff
@@ -60,73 +64,84 @@ public with sharing class GE_GiftEntryController {
...
+    * @description Sends the purchase request to Elevate and logs an error on a timeout error.
+    * This method is used in tests instead of the sendPurchaseRequest() in order to skip JSON.deserialize(response)
+    * due to it erroring out because the response header contains the Object data type.
+    * @param requestBodyParameters: JSON containing parameters for the purchase call request body
+    * @param dataImportRecordId: Record ID of the related Data Import
+    * @return UTIL_Http.Response
+    */
+    @TestVisible
+    private static UTIL_Http.Response processPurchaseRequest(String requestBodyParameters, String dataImportRecordId) {
+        UTIL_Http.Response response = new UTIL_Http.Response();
```
RadSFDeveloper, line 82: @kenlewis-sfdo , Great points on if and how the response should be built (response wrapper vs HttpResponse) and where it should be checked (Apex class vs JS). Feel free to change the response creation/checks/return for the purchase request, however note that create and cancel commitment functionality is using common `PS_Request` and `UTIL_Http` classes.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
RadSFDeveloper: I was attempting to do that before during coding of the psElevateTokenHandler but it did not work. Let me try again. Might need to consult you or @psadie-sfdo  if I cannot make it work.

RadSFDeveloper: The VF origin and origin URL vars have been moved into `psElevateTokenHandler`. Can you please review changes again?

src/classes/GE_GiftEntryController.cls
kenlewis-sfdo: The new line that throws looks good to me!

src/pages/GE_TokenizeCard.page
kenlewis-sfdo: The new change with the brackets included looks good!  Thanks Radmila!

RadSFDeveloper:
Tracking [W-7974134](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008IE0LIAW/view)

RadSFDeveloper:
Tracking [W-8134765](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000089leQIAQ/view)

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
ptewson-sfdo: That's accurate as far as I know - we don't plan on supplying a value there until the gateway metadata endpoint is available to get more accurate information about what additional fields are expected by the gateway for verification, including cardholder name.

src/classes/GE_GiftEntryController.cls
kenlewis-sfdo: Sounds good!  I think this is fine to leave as is for now and I will add my suggestions to our Gift Entry refactor list in Quip.

kenlewis-sfdo (commented):
Overall I think this PR looks great!  Thanks for all of this work, Radmila!  I'll add all of the suggestions I brought up in our discussion into our Gift Entry refactor list doc and create a few refactor WIs for future grooming and discussion.  I don't have any requested changes that would block this PR from moving forward as is.

ptewson-sfdo (approved):
Looks good! Nice work to bring everything into sync!

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: This change looks good to me!

psadie-sfdo (approved):
The latest changes look good! Thank you for all the effort Radmila!

RadSFDeveloper:
@deepmalamehta , the GE form is ready for testing. The W-7974134 and W-8134765 are attached to this PR. Feel free to set them to Fixed or QA in Progress state.

@rponti-sforg , RD2 (create/cancel commitment, error handling) on the org connected to Elevate and RD2 on an org not connected to Elevate is ready for testing.

PR #5939, opened 2020-10-12
Add new unpackaged metadata per QE request

codexmarcel:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


unpackaged/config/qa/objects/Lead.object
```diff
@@ -0,0 +1,1020 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davidmreed, line 1: It looks like we are only intending to deploy a single field on Lead. Can we excise the remainder of the metadata other than the handful of required attributes? I see some Data.com field references here that will be a problem when that feature ultimately goes away.

codexmarcel: The benefits of code review strike again!

skristem (approved):
tested with qa_org and regression_org flows

PR #5950, opened 2020-10-14
Gift Entry: Render Record Types as Picklists and Centralize State Management

psadie-sfdo:
[W-7963596](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UDVkIAO/view)
[W-8017324](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008UmDaIAK/view)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

# Unaggregated Pull Requests
* Gift Entry: add jest tests for picklist type form fields [[PR5962](https://github.com/SalesforceFoundation/NPSP/pull/5962)]
* source of truth for selectedDonation is now formRenderer and is passe… [[PR5974](https://github.com/SalesforceFoundation/NPSP/pull/5974)]


src/lwc/geFormField/__tests__/geFormField.test.js
```diff
@@ -0,0 +1,69 @@
...
+        return Promise.resolve().then(() => {
+            pl.dispatchEvent(new CustomEvent('change', {value: newPicklistValue}));
+        });
+
+        expect(picklistChangeHandler).toBeCalled();
+
+        const formFieldChangeDetailObject = {
+            [fieldMapping.Source_Field_API_Name]: {value: newPicklistValue}
+        }
+        expect(picklistChangeHandler).toBeCalledWith(formFieldChangeDetailObject);
```
psadie-sfdo, line 66: @kenlewis-sfdo @bdvorachek - It looks to me like this block of code isn't reachable with the `return` on line 57. Could ya'll take another look? Running into a bit of a wall trying to test that the correct events are firing and being handled when a picklist option is selected myself. If ya'll have some time to sync up tomorrow, I'd appreciate it!

src/lwc/geFormField/geFormField.js
```diff
@@ -40,43 +41,46 @@ const DATETIME = 'datetime-local';
...
-    @track picklistValues = [];
+    @track _value;
     @track objectDescribeInfo;
     @track richTextValid = true;
     @track _disabled = false;
     @api element;
     @api targetFieldName;
     _defaultValue = null;
     _recordTypeId;
+    _picklistValues
```
voduyemi, line 52:
```suggestion
    _picklistValues;
```

src/lwc/geFormField/geFormField.js
```diff
@@ -536,29 +519,15 @@ export default class GeFormField extends LightningElement {
...
-    setRecordTypeIdOnChildComponents() {
-        if (this.isPicklist) {
-            this.template.querySelector('c-ge-form-field-picklist')
-                .recordTypeId = this.recordTypeId;
-        }
-    }
-
     fireLookupRecordSelectEvent() {
+        if (this.targetFieldApiName === 'RecordTypeId') {
+            console.log('*** ' + 'not firing lookup rcrd select for target rtid' + ' ***');
```
voduyemi, line 528: Can we remove this `console.log`?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -929,28 +916,46 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+            this.setFormStateToInitialFieldValuesForObjMapping(objectMappingDeveloperName);
         }
-
-        sectionsList.forEach(section => {
-            section.reset(fieldMappingDevNames, setDefaults);
-        });
         this.widgetData = {};
     }
 
+    resetLocalState() {
```
voduyemi, line 930: What do you think about calling this function `resetFormState` ?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1766,4 +1749,247 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
     get qaLocatorSaveButton() {
         return `button ${this.saveActionLabel}`;
     }
+
+    /*******************************************************************************
+     * @description Object used to hold current values for all fields on the form.
+     */
+    @track
+    _formState = {}
```
voduyemi, line 1757: Can we move the `_formState` variable to the top of the class since its a top-level variable?

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1766,4 +1749,247 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+
+    sourceFieldFor(fieldMappingDevName) {
+        return GeFormService.fieldMappings[fieldMappingDevName].Source_Field_API_Name;
+    }
+
+    /*******************************************************************************
+     * @description Analyzes the sections property to get initial values and set them
+     * in the formState property.
+     */
+    initializeState() {
```
voduyemi, line 1871: Can we call this function `initializeFormState` so we're inline with `updateFormState` and `resetFormState`?

voduyemi (changes requested):
Thanks a lot @kenlewis-sfdo and @psadie-sfdo! This is a lot of work and I'm pumped about how it's shaping out, I left some comments about naming, styling and structure. I was also thinking of the size of `geFormField` and `geFormRenderer` js files, do we think at some point we could off-load some of their responsibilities into smaller helper modules that we could import ?

src/lwc/geFormField/__tests__/geFormField.test.js
kenlewis-sfdo: Yeah, good catch @psadie-sfdo I'll take a look at this test and see if I can re-work it to make it into something meaningful.  Thanks!

src/lwc/geFormField/geFormField.js
kenlewis-sfdo: Yes, I'm actually going to remove this if block since Record Types are no longer lookups.

src/lwc/geFormRenderer/geFormRenderer.js
kenlewis-sfdo: Yep I'll move it up!

src/lwc/geFormRenderer/geFormRenderer.js
kenlewis-sfdo: Yep good call!

src/lwc/geFormRenderer/geFormRenderer.js
kenlewis-sfdo: Yeah, I like this suggestion also.  I initially intentionally did not call it that because it is (currently) not just resetting/initializing the new formState object, it is also updating state that is currently stored in these other local variables (and I just extracted most of this logic from its previous location to reduce the size of the function.  But, as we continue with this work, the calls to these methods on 931, 932, and 933 should go away since "reset" will only need to call this.initializeState once all state is stored in the formState object.  So, I'll update it, thanks!

src/lwc/geFormField/geFormField.js
```diff
@@ -523,41 +506,16 @@ export default class GeFormField extends LightningElement {
         }
 
         // reset lookups and recordtype fields
-        if (this.isLookupRecordType || this.isLookup) {
+        if (this.isLookup) {
```
kenlewis-sfdo, line 509: @psadie-sfdo is this function needed at all anymore?

src/lwc/geFormField/__tests__/geFormField.test.js
bdvorachek: So I did a bit of research on this and discovered that we need to fast-forward time whenever we're testing this component due to the debounce-ing we're doing in the change handler. We'd put `jest.useFakeTimers();` in the header of this file, and then `jest.runOnlyPendingTimers();` after dispatching the change event.

bdvorachek (changes requested):

src/lwc/geFormField/geFormField.js
kenlewis-sfdo: We should check in early next week, but my thoughts:

* the first block `if (setDefaults) {` should be updated to `if (!this.isPicklist && setDefaults) {` since picklists value will now be reset in the renderer.
* the second block `if (this.isLookup) {` should be removed because I don't think that `querySelector('[data-id="inputComponent"]')` is needed anymore 
* the third block `if (this.isPicklist) {` can also be removed since picklists get their reset value from formState now

src/lwc/geFormField/__tests__/geFormField.test.js
psadie-sfdo: Thanks for looking into this Ben! Gonna give that a shot today!

bdvorachek (approved):

voduyemi (approved):

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1785,4 +1791,324 @@ export default class GeFormRenderer extends NavigationMixin(LightningElement) {
...
+        let flatObj = {};
+        for (const [key, value] of Object.entries(obj)) {
+            if (value !== null && value !== undefined && value.hasOwnProperty('value')) {
+                flatObj[key] = value.value;
+            } else {
+                flatObj[key] = value;
+            }
+        }
+        return flatObj;
+    }
```
ptewson-sfdo, line 2071: Would this be something to possibly move to utilCommon or another shared location?

ptewson-sfdo (approved):
This looks great! Thanks team!

PR #5953, opened 2020-10-15
PMT: CAPTCHA for iframe

dfuller114:
# Critical Changes

# Changes
We are now passing a JWT in the Elevate tokenize call in order to bypass CAPTCHA verification in the Elevate iFrame in Gift Entry.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_PaymentServices.cls
```diff
@@ -122,13 +122,23 @@ public with sharing class GE_PaymentServices {
...
+        JWTPayload jwtPayload = new JWTPayload(config.get(PS_IntegrationServiceConfig.PRODUCTID),
+                PS_IntegrationServiceConfig.PURCHASE_ROLE);
         String jwt = UTIL_Jwt.getSignedJWTFromString(
                 jwtPayload.getAsString(),
                 config.get(PS_IntegrationServiceConfig.API_KEY)
         );
+        return jwt;
+    }
 
+    public String getJwtForToken() {
```
dfuller114, line 135: quick explanation of why `getJwtForToken()` and `getJwtForPurchase()` are being used with some duplicate code. We have to pass an explicit role to get a JWT now since we are generating jwts for more than one role. Since we're using the SDK to generate the JWT for tokenize we can't pass the role as a parameter since that would require Visualforce remoting, which would require the method to be `static` which it isn't and we are using a constructor Builder pattern moving forward for the PS_Request class anyway. That's going to make tokenizing from the SDK a special case. We likely can make the regular `getJwt()` function more generic once we merge with @RadSFDeveloper's code but I have created it this way for now for simplicity sake until we merge with her code.

src/classes/GE_PaymentServices.cls
```diff
@@ -122,13 +122,23 @@ public with sharing class GE_PaymentServices {
         return config.get(PS_IntegrationServiceConfig.BASE_URL);
     }
 
-    public String getJwt() {
-        JWTPayload jwtPayload = new JWTPayload(config.get(PS_IntegrationServiceConfig.PRODUCTID));
+    public String getJwtForPurchase() {
+        JWTPayload jwtPayload = new JWTPayload(config.get(PS_IntegrationServiceConfig.PRODUCTID),
+                PS_IntegrationServiceConfig.PURCHASE_ROLE);
```
psadie-sfdo, line 127:
```suggestion
            PS_IntegrationServiceConfig.PURCHASE_ROLE);
```

src/classes/GE_PaymentServices.cls
```diff
@@ -122,13 +122,23 @@ public with sharing class GE_PaymentServices {
...
         String jwt = UTIL_Jwt.getSignedJWTFromString(
                 jwtPayload.getAsString(),
                 config.get(PS_IntegrationServiceConfig.API_KEY)
```
psadie-sfdo, line 130:
```suggestion
            jwtPayload.getAsString(),
            config.get(PS_IntegrationServiceConfig.API_KEY)
```

src/classes/GE_PaymentServices.cls
```diff
@@ -122,13 +122,23 @@ public with sharing class GE_PaymentServices {
...
         );
+        return jwt;
+    }
 
+    public String getJwtForToken() {
+        JWTPayload jwtPayload = new JWTPayload(config.get(PS_IntegrationServiceConfig.PRODUCTID),
+                PS_IntegrationServiceConfig.TOKEN_ROLE);
```
psadie-sfdo, line 137:
```suggestion
            PS_IntegrationServiceConfig.TOKEN_ROLE);
```

src/classes/GE_PaymentServices.cls
```diff
@@ -122,13 +122,23 @@ public with sharing class GE_PaymentServices {
...
+        String jwt = UTIL_Jwt.getSignedJWTFromString(
+                jwtPayload.getAsString(),
+                config.get(PS_IntegrationServiceConfig.API_KEY)
```
psadie-sfdo, line 140:
```suggestion
            jwtPayload.getAsString(),
            config.get(PS_IntegrationServiceConfig.API_KEY)
```

src/classes/GE_PaymentServices.cls
```diff
@@ -144,14 +154,20 @@ public with sharing class GE_PaymentServices {
...
+
     /*******************************************************************************************************
     * @description Inner class to assist in constructing the Payload portion of a JSON Web Token.
     */
     public with sharing class JWTPayload {
         String sfdoId;
+        String role;
 
-        public JWTPayload(String sfdoId) {
+        public JWTPayload(String sfdoId, String role) {
```
psadie-sfdo, line 168: Should we pluralize `role`? I know it's technically a single string, but multiple roles comma-separated can be contained within this string right?

src/pages/GE_TokenizeCard.page
```diff
@@ -76,9 +77,14 @@
         * @description Creates a token and posts the response to geFormWidgetTokenizeCard
         */
         function handleTokenCreation() {
+            let auth = {
```
psadie-sfdo, line 80:
```suggestion
            const auth = {
```

psadie-sfdo (changes requested):
Just a few minor comments, for now, pending Radmila's PR merging. Will take another look at that point. Thanks!

src/classes/GE_PaymentServices.cls
```diff
@@ -100,6 +100,10 @@ public with sharing class GE_PaymentServices {
         return configService.getGatewayIds();
     }
 
+    public String getMerchantIds() {
+        return configService.getMerchantIds();
+    }
+
```
RadSFDeveloper, line 106: Missing another method:
```
/**
     * @description Returns JSON Web Token to payments services tokenize card iframe
     * @return String
     */
    public String getJwtForToken() {
        return new PS_Request.Builder().getJWT(PS_Request.ElevateEndpoint.TOKENIZE);  
    }
    
```

src/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -51,6 +51,8 @@ public with sharing class PS_IntegrationServiceConfig {
     public static final String SFDO_USERNAME = 'sfdo.username';
     public static final String ELEVATE_SDK = 'elevateSDK';
     public static final String PRODUCTID = 'productId';
+    public static final String PURCHASE_ROLE = 'Payments.Purchase';
```
RadSFDeveloper, line 54: These constants are not required after the latest changes.

src/classes/GE_PaymentServices.cls
dfuller114: @RadSFDeveloper yeah thanks. Was holding off until I made `getJWT()` public in `PS_Request`. Just updated!

src/classes/PS_IntegrationServiceConfig.cls
dfuller114: @RadSFDeveloper updated!

src/pages/GE_TokenizeCard.page
dfuller114: @psadie-sfdo updated!

src/classes/GE_PaymentServices.cls
dfuller114: @psadie-sfdo `JWTPayload` isn't being used anymore so we don't have to worry about this one.

src/classes/GE_PaymentServices.cls
dfuller114: @psadie-sfdo this code has been removed after merging with @RadSFDeveloper's PR.

src/classes/GE_PaymentServices.cls
dfuller114: this code has been removed after merging with @RadSFDeveloper's PR.

src/classes/GE_PaymentServices.cls
dfuller114: this code has been removed after merging with @RadSFDeveloper's PR.

src/classes/GE_PaymentServices.cls
dfuller114: this code has been removed after merging with @RadSFDeveloper's PR.

src/classes/GE_PaymentServices.cls
RadSFDeveloper: Everything looks good @dfuller114 . Wondering if additional change to [PS_TestPaymentServices](https://github.com/SalesforceFoundation/NPSP/blob/0bd334447467506e0944c62efe5e98c1e9355b04/unpackaged/config/ps_testing/classes/PS_TestPaymentService.cls#L66) is required to handle the CAPTCHA requirements?

src/pages/GE_TokenizeCard.page
```diff
@@ -87,10 +88,15 @@
         * @description Creates a token and posts the response to a credit card widget LWC
         */
         function handleTokenCreation() {
-// TODO Remove this line in W-8021204
-console.log('>>> Tokenize with Name On Card: "' + nameOnCard + '"');
+            // TODO Remove this line in W-8021204
+            console.log('>>> Tokenize with Name On Card: "' + nameOnCard + '"');
```
ptewson-sfdo, line 92: @RadSFDeveloper @dfuller114 - do we want this console log to stay in here? Do we have a plan to keep it from hitting production?

RadSFDeveloper: Is there any issue in keeping the debug log as is? There is a future work on the RD entry form to provide card holder name to the tokenization, so it is beneficial to keep it for that purpose. However, if there is any issue with the debug statement being included in the release, feel free to remove it in this PR.
CC @force2b

ptewson-sfdo (approved):
Looks good! Want to remove the console log if possible, but if we make that update, there's no reason to hold up QA!

src/pages/GE_TokenizeCard.page
ptewson-sfdo: We generally don't ship code with log statements, but if it's serving a purpose I'm ok with it remaining on a temporary basis!

RadSFDeveloper: Since it is a temp debug and it can be added when the future work starts, we can add the console output just for that WI, so @dfuller114 , do you mind removing the debug statement if it is not too late?

force2b: Agreed. We can just add it back when needed for QA.

psadie-sfdo:
Looks good to me as well! Would also like removal of the console log as Patrick mentioned, but I wouldn't consider it a blocker.

psadie-sfdo (approved):
Looks good to me as well! Would also like removal of the console log as Patrick mentioned, but I wouldn't consider it a blocker.

src/pages/GE_TokenizeCard.page
dfuller114: just pushed an update to remove the console log

RadSFDeveloper (approved):
Looks good, thanks Daniel!

PR #5962, opened 2020-10-19
Gift Entry: add jest tests for picklist type form fields

psadie-sfdo:
# Critical Changes

# Changes
- Adds jest tests for picklist type form fields
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


kenlewis-sfdo (approved):
These look great to me!!

src/lwc/geFormField/__tests__/geFormField.test.js
```diff
@@ -54,16 +89,181 @@ describe('c-ge-form-field', () => {
         const pl = element.shadowRoot.querySelector('lightning-combobox');
 
         const newPicklistValue = 'newPicklistValue';
+        const customEventDetail = { detail: { value: newPicklistValue } };
+
+        return Promise.resolve().then(() => {
+            pl.dispatchEvent(new CustomEvent('change', customEventDetail));
```
bdvorachek, line 95: use `dispatchChangeEvent` for this line, also, I think it is ok to move this line upwards so it is outside of the `Promise.resolve.then(() => ` block.

src/lwc/geFormField/__tests__/geFormField.test.js
```diff
@@ -54,16 +89,181 @@ describe('c-ge-form-field', () => {
...
+                mockFormState(element, 'Picklist_Option_2');
+            })
+            .then(() => {
+                expect(picklist.value).toBe('Picklist_Option_2');
+            });
+    });
+
+});
+
+expect.extend({
```
bdvorachek, line 215: :+1: cool use of expect.extend.

src/lwc/geFormField/__tests__/geFormField.test.js
```diff
@@ -54,16 +89,181 @@ describe('c-ge-form-field', () => {
...
+        return {
+            message: () =>
+                `expected the picklist to contain options: ${JSON.stringify(expected)}
+                actual options: ${JSON.stringify(actual)}`,
+            pass,
+        };
+    }
 });
+
+const getShadowRoot = (element) => {
```
bdvorachek, line 237: I don't have a good spot in mind for the more generic functions here at the moment, but I'd like to see them migrated out somewhere. Putting them as a global inside of `jest.setup.js` like we did with `clearDOM` doesn't feel right, but I'm open to being wrong on this. It'd be nice if we could `import { getShadowRoot, shadowQuerySelector } from [somewhere] since I think we'll be re-using these.

bdvorachek (changes requested):

src/lwc/geFormField/__tests__/geFormField.test.js
psadie-sfdo: The lightning-global repo has a bunch of these extensions and other useful utils. Only pulled the ones that seemed immediately useful for these tests 🙏

bdvorachek (approved):

PR #5969, opened 2020-10-20
[RD - Tech Debt] Refactor RD2 tests so they are not re-run in builds

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/PS_CommitmentRequest_TEST.cls
+++ b/src/classes/PS_CommitmentRequest_TEST.cls
@@ -33,7 +33,7 @@
 * @group Elevate Payment Services
 * @description Tests verifying Elevate payment services commitment request
 */
-@IsTest(IsParallel=true)
+@IsTest(IsParallel=false)
 private with sharing class PS_CommitmentRequest_TEST {
 
     private static final String PAYMENT_TOKEN = 'token-abcd-efgh-ijkl-mnop-qrst';
--- a/src/classes/RD2_ApiService_TEST.cls
+++ b/src/classes/RD2_ApiService_TEST.cls
@@ -39,14 +39,22 @@ private class RD2_ApiService_TEST {
 
     private static final String API_QUERY_SCHEDULES = 'rd2.queryschedules';
     private static final TEST_SObjectGateway.RecurringDonationScheduleGateway scheduleGateway = new TEST_SObjectGateway.RecurringDonationScheduleGateway();
+    private static final TEST_SObjectGateway.RecurringDonationGateway rdGateway = new TEST_SObjectGateway.RecurringDonationGateway();
 
     /****
     * @description Creates data required for unit tests
     */
     @TestSetup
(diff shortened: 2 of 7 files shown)


RadSFDeveloper:
Attached to [W-8071518](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008d2D8IAI/view)

src/classes/PS_CommitmentRequest_TEST.cls
```diff
@@ -33,7 +33,7 @@
 * @group Elevate Payment Services
 * @description Tests verifying Elevate payment services commitment request
 */
-@IsTest(IsParallel=true)
```
RadSFDeveloper, line 36: The `PS_CommitmentRequest_TEST` was recently created but it was re-run in most builds. This is not expected since the test class is relatively small (three tests) and it is using `testSetup()` method to create required data for all tests. So, this was a lucky break to realize that not necessarily we need to change RD2 test classes to avoid DMLs, but also chance parallel testing attribute to `false`. After changing the parallel attribute, none of the changed classes is re-run, therefore, the total build time is decreased.

So based on one [build](https://sfdo-metaci.herokuapp.com/builds/342561/flows) in another PR, the re-run was applied to "Retrying 263 failed methods from 19 test classes" and took about 21 min.
Another random chosen [build](https://sfdo-metaci.herokuapp.com/builds/338372/flows) had "Retrying 342 failed methods from 19 test classes" and took 59 min.
Another PR random [build](https://sfdo-metaci.herokuapp.com/builds/342549/flows) had "Retrying 305 failed methods from 19 test classes" and took 30 min.

A [build](https://sfdo-metaci.herokuapp.com/builds/342016/flows) in *this PR* has "Retrying 116 failed methods from 17 test classes" and it took 40 min.
Another [build](https://sfdo-metaci.herokuapp.com/builds/342461/flows) had "Retrying 79 failed methods from 13 test classes which took 6min.

So even though we cannot influence how long other tests re-runs will take, we did decrease number of re-run tests and increased the probability of builds running faster. Based on builds in this PR, time to run tests (excluding installed package deployments and post deploy cleanup) takes from 1h 10min to 1h 20min.

CC @force2b , @andrewyu-salesforce , @npsp-reedestockton , @ptewson-sfdo

andrewyu-salesforce (approved):
@RadSFDeveloper Interesting observation of the isParallel tag. All tests pass and no RD2 tests were rebuild during the run.

src/classes/PS_CommitmentRequest_TEST.cls
force2b: Great catch @RadSFDeveloper

force2b (approved):
Excellent work on this!

PR #5972, opened 2020-10-21
Add Last Elevate Event Played field

npsp-reedestockton:
# Critical Changes

# Changes
 - We added a new field, Last Elevate Event Played, to the Recurring Donation object. This field supports Elevate integration with NPSP. It's a system field so you don't need to add to your page layouts. We included Read access on the Standard and System Administrator profiles.

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom field:
    - npe03__Recurring_Donation__c.LastElevateEventPlayed__c

# Deleted Metadata

### DIFF
--- a/src/objects/npe03__Recurring_Donation__c.object
+++ b/src/objects/npe03__Recurring_Donation__c.object
@@ -337,6 +337,18 @@
         <unique>false</unique>
     </fields>
     <fields>
+        <fullName>LastElevateEventPlayed__c</fullName>
+        <description>The last Elevate event played resulting in an update to this Recurring Donation.</description>
+        <externalId>true</externalId>
+        <inlineHelpText>The last Elevate event played resulting in an update to this Recurring Donation.</inlineHelpText>
+        <label>Last Elevate Event Played</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
         <fullName>NextYearValue__c</fullName>
         <description>This Recurring Donation&apos;s total expected amount for the next calendar or fiscal year (read only).</description>
         <externalId>false</externalId>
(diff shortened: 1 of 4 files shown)


npsp-reedestockton:
@RadSFDeveloper @andrewyu-salesforce @sam-knox This is ready for review.

npsp-reedestockton:
Tracking [W-8246650](https://gus.lightning.force.com/lightning/_classic/%2Fa07B0000008h6ESIAY)

andrewyu-salesforce (approved):
looks great

RadSFDeveloper (approved):
Awesome work, thank you @npsp-reedestockton !

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -336,6 +336,18 @@
         <type>Number</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>LastElevateEventPlayed__c</fullName>
```
RadSFDeveloper, line 340: @callen-sfdo , do you please mind double checking the new field on RD that will contain an Elevate Event Id? Thanks.

sam-knox: @RadSFDeveloper - I assume we don't need to tell admins to put this field on their page layouts right?

sam-knox (approved):

src/objects/npe03__Recurring_Donation__c.object
RadSFDeveloper: That is correct @sam-knox , the field is needed for the RD integration framework for an org connected to Elevate. It should not be added to any page layout.

sam-knox: @RadSFDeveloper - As part of documentation, do we need to tell admins to add Read and Edit access to this field for their Elevate integration user, or is this going to be added to the Elevate permission sets?

sam-knox (approved):
Updated the Changes section.

src/objects/npe03__Recurring_Donation__c.object
RadSFDeveloper: The field is relevant for the integration user, so for the regular user, the check will be in trigger that will not check field permission, so there is no need for adding permissions. When the RD MetaDeploy is run, the field will be added to the system admin and standard user profile. So, for existing Elevate customers, we can specify that they add read/edit field permission to the field.

PR #5985, opened 2020-10-26
Resolve GEM unit test failures

vmartinov-sforg:
# Critical Changes

# Changes
Added unit tests in NPSP that mirror the GEM failing tests, to be able to correct the issue.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/UTIL_Describe.cls
+++ b/src/classes/UTIL_Describe.cls
@@ -286,14 +286,14 @@ public class UTIL_Describe {
             } else {
                 accessibleFields.add(field.getName());
             }
-    	}
-  
+        }
+        
     	String allFields = '';
     	for (String fieldName : accessibleFields) {
             allFields += fieldName + ','; 
         }
         allFields = allFields.substring(0, allFields.length() - 1);
-    	return allFields;
+        return allFields;
   	}
 
     /*******************************************************************************************************
--- a/src/classes/UTIL_Describe.cls
+++ b/src/classes/UTIL_Describe.cls
@@ -272,7 +272,6 @@ public class UTIL_Describe {
     /*******************************************************************************************************
      * @description Returns a comma delimited string with all accessible fields of an object, used for queries.
      * @param sobjectname Name of the object to retrieve field list for.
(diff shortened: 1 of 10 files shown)


vmartinov-sforg:
Work Item: [W-8248351](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008h7bIIAQ/view)

src/classes/UTIL_Describe.cls
```diff
@@ -269,6 +269,33 @@ public class UTIL_Describe {
         return fieldMap;
     }
 
+    /*******************************************************************************************************
+     * @description Returns a comma delimited string with all accessible fields of an object, used for queries.
+     * @param sobjectname Name of the object to retrieve field list for.
+     * @return String A comma delimited String with all fields of an object.
+     */
+    public static String getAllFieldsQueryStringFromObject(String sObjectName) {
```
ptewson-sfdo, line 277: I have some potential reservations about adding this method to the Describe library. I realize it's here because there's a similar method in GEM and we want to utilize similar setup for the new unit test to verify that we aren't failing in the equivalent test that was failing in GEM. However, it appears that its purpose is to support a "SELECT  *" type functionality, and that's not considered a great practice in Salesforce development.  

I realize that we are doing this essentially already both in GEM and in parts of BDI, but I might prefer to avoid adding a method that's so clearly designed to support that practice -  one option is that we could simply bring this code inline within the unit test, although we also might want to consider grabbing the globally available  "listStrDataImportFields" property of the BDI_DataImportService class and doing a String.join() of those results as well?

ptewson-sfdo (commented):
One suggested change, other than that looks great!

src/classes/UTIL_Describe.cls
vmartinov-sforg: Didn't know listStrDataImportFields already existed...

ptewson-sfdo (approved):
Looks good!

bdvorachek (approved):

PR #5988, opened 2020-10-27
Gift Entry: Batch Status Query Returns Non-Selective Query

### LINKED ISSUE #5811: Batch Status Query Returns Non-Selective Query or 50,000 Row Limit Error in Customer Orgs with Thousands of Data Import Records
The query throwing the error is in [BDI_DataImportBatchStatus_TDTM](https://github.com/SalesforceFoundation/NPSP/blob/61d3c4fd51ce07ab5f6a611f355ee6dbb5fb4cba/src/classes/BDI_DataImportBatchStatus_TDTM.cls), Lines 123-127.
```
SELECT Id, Batch_Status__c, 
	(SELECT Id, NPSP_Data_Import_Batch__c, Status__c FROM NPSP_Data_Imports__r) 
FROM DataImportBatch__c 
WHERE Id = :tmpVar1
```

**Issue 1:**
If you have more than 200,000 NPSP Data Import records in an org and try to Process a Batch:

```
npsp.TDTM_DataImport: execution of AfterInsert caused by: System.QueryException: Non-selective query against large object type (more than 200000 rows). Consider an indexed filter or contact salesforce.com about custom indexing. Even if a field is indexed a filter might still not be selective when: 1. The filter value includes null (for instance binding with a list that contains null) 2. Data skew exists whereby the number of matching rows is very large (for instance, filtering for a particular foreign key value that occurs many times) (npsp)
```
Screen Shot 2020-09-04 at 1 31 12 PM

Other scenarios where I replicated:
* If you have more than 200,000 records associated with a batch where the Batch Status = ‘Completed’ and you attempt to add a new Data Import record. 
* If you have more than 200,000 Data Import records in general. And you have a Batch with Status of “Open” and > 50,000 Data Import records associated with it and you associate a new Data Import record with the batch  (Sometimes with fewer records as well…)
* I also replicated while Processing a Batch:
Screen Shot 2020-09-04 at 1 37 30 PM

**Issue 2:** 
If you have more than 50,000 NPSP Data Import records associated with one NPSP Data Import Batch (but less than 200,000 Data Import records as a whole) and you try to insert another NPSP Data Import record tied to that batch, you hit:
`npsp.TDTM_DataImport: System.LimitException: npsp:Too many query rows: 50001`

----
Neither of these issues surfaced before adding the above query that we use to set the Batch Status field on the NPSP Data Import object. While it would be a best practice for customers to limit the total number of NPSP Data Import records associated to an NPSP Data Import Batch to 10,000 (per Salesforce documentation), customers may hit this now during their initial import (if data is tied to a batch) OR if they haven’t deleted Data Import records in the past.

### Workarounds:
1. Limit the number of NPSP Data Import records associated with an NPSP Data Import Batch to 10,000 or less.
2. Delete NPSP Data Import records that are no longer needed.

### Steps to Repeat:

**50,000 Row Error**
1. Insert 1 NPSP Data Import Batch record.
2. Insert 50,000 NPSP Data Import records associated with that Batch record. (You may need to disable the Trigger Handlers on the NPSP Data Import and NPSP Data Import Batch objects to avoid errors). Re-enable them when the 50,000 records are inserted.
3. Insert 1 more Data Import record associated with the NPSP Data Import Batch
4. Confirm the 50,000 row limit. 

**Non-selective query error:**
1. Disable the Trigger Handlers on the NPSP Data Import and NPSP Data Import Batch objects.
 2. Create a new NPSP Data Import Batch record.
3. Insert 200,000 NPSP Data Import records  and associate with that Batch record.
4. Re-enable the Trigger Handlers you disabled.
5. Go back to your original NPSP Data Import batch record (that has 50K records) and attempt to insert an NPSP Data Import record. You should receive the non-selective query error.

bdvorachek:
GUS WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008gwsZIAQ/view

# Critical Changes

# Changes

# Issues Closed
Closes #5811 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_DataImportBatchStatus_TDTM.cls
+++ b/src/classes/BDI_DataImportBatchStatus_TDTM.cls
@@ -46,18 +46,25 @@ public class BDI_DataImportBatchStatus_TDTM extends TDTM_Runnable {
      ********************************************************************************************************************/
     public override TDTM_Runnable.DmlWrapper run(List<SObject> newSObjects, List<SObject> oldSObjects,
             TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
-        if (triggerAction == TDTM_Runnable.Action.AfterInsert
-                || triggerAction == TDTM_Runnable.Action.AfterUndelete) {
-            setBatchStatus(newSObjects);
+        
+        switch on triggerAction {
+            when AfterInsert {
+                setBatchStatus(newSObjects);
+            }
+
+            when AfterUpdate {
+                setBatchStatus(oldSObjects, newSObjects);
+            }
+
+            when AfterUndelete {
+                setBatchStatus(newSObjects);
+            }
+
+            when AfterDelete {
+                setBatchStatus(oldSObjects);
+            }
         }
 
-        if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-            setBatchStatus(oldSObjects, newSObjects);
-        }
-
-        if (triggerAction == TDTM_Runnable.Action.AfterDelete) {
-            setBatchStatus(oldSObjects);
-        }
         return null;
     }
 
--- a/src/classes/BDI_DataImportBatchStatus_TDTM.cls
+++ b/src/classes/BDI_DataImportBatchStatus_TDTM.cls
@@ -64,11 +64,6 @@ public class BDI_DataImportBatchStatus_TDTM extends TDTM_Runnable {
         return null;
     }
 
-    /**
-     * @description Update Batch Status if a relevant field on the Data Import has been updated.
-     * @param diOldList Trigger.old
-     * @param diNewList Trigger.new
-     */
     private void setBatchStatusOnUpdate(List<DataImport__c> diOldList,
             List<DataImport__c> diNewList) {
 
@@ -76,35 +71,11 @@ public class BDI_DataImportBatchStatus_TDTM extends TDTM_Runnable {
         updateBatchStatus(batchIdSet);
     }
 
-    private Boolean needsStatusUpdate(DataImport__c diOld, DataImport__c diNew) {
-        if (diOld.NPSP_Data_Import_Batch__c != diNew.NPSP_Data_Import_Batch__c) {
-            // di records changed parents
-            if (diOld.NPSP_Data_Import_Batch__c != null) {
-                return true;
-            }
-            if (diNew.NPSP_Data_Import_Batch__c != null) {
-                return true;
-            }
-        } else if (diOld.Status__c != diNew.Status__c) {
-            // status was changed; same parent
-            if (diNew.NPSP_Data_Import_Batch__c != null) {
-                return true;
-            }
-        }
-
-        return false;
-    }
-
     private void setBatchStatus(List<DataImport__c> diList) {
         Set<Id> batchIdSet = getBatchIdsToUpdate(diList);
         updateBatchStatus(batchIdSet);
     }
 
-    /**
-     * @description updates the batch status on di records insert, delete or undelete
-     * @param diList the old or new list
-     * @return {Set<Id>} A set of Batch Ids that need to have their status updated.
-     */
     private Set<Id> getBatchIdsToUpdate(List<DataImport__c> diList) {
         Set<Id> batchIdSet = new Set<Id>();
 
@@ -131,12 +102,31 @@ public class BDI_DataImportBatchStatus_TDTM extends TDTM_Runnable {
         return batchIdSet;
     }
 
+    private Boolean needsStatusUpdate(DataImport__c diOld, DataImport__c diNew) {
+        if (diOld.NPSP_Data_Import_Batch__c != diNew.NPSP_Data_Import_Batch__c) {
+            // di records changed parents
+            if (diOld.NPSP_Data_Import_Batch__c != null) {
+                return true;
+            }
+            if (diNew.NPSP_Data_Import_Batch__c != null) {
+                return true;
+            }
+        } else if (diOld.Status__c != diNew.Status__c) {
+            // status was changed; same parent
+            if (diNew.NPSP_Data_Import_Batch__c != null) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     /**
      * @description Update the batch status based on statuses of the data import records associated with that batch
      * @param batchIdSet
-     * when one DI is imported, and none are failed or open -> then the Batch Status is Completed
-     * when one DI status is blank or Dry Run Validated or Dry Run error, and none are failed -> then the Batch Status is Open
-     * when any DI is Failed -> then the Batch Status is Failed - needs review
+     * When one DI is imported, and none are failed or open -> then the Batch Status is Completed
+     * When one DI status is blank or Dry Run Validated or Dry Run error, and none are failed -> then the Batch Status is Open
+     * When any DI is Failed -> then the Batch Status is Failed - Needs Review
      *
      */
     private void updateBatchStatus(Set<Id> batchIdSet) {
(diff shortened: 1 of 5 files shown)


bdvorachek:
> one with an empty status

Updated the query that scans for "Open" rows to also include `NULL`. I don't think comparing with `NULL` should have a big negative impact on selectivity since we're still using `LIMIT 1`, but will run through the LDV org to see if this has any effect.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,273 @@
...
+         * Read the status of a single data import row from the inner query on the batch,
+         * and set one of the batch status flags to indicate that the specified type of row was found.
+         * @param dataImport The data import row to get the status of.
+         */
+        void readRowStatus(DataImport__c dataImport) {
+            String status = dataImport.Status__c;
+
+            if (isOpen(status)) {
+                this.hasOpenRow = true;
+            }
```
vmartinov-sforg, line 263: We could add a return statement here to avoid checking the status again below.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,273 @@
...
+
+            if (isImported(status)) {
+                this.hasImportedRow = true;
+            }
```
vmartinov-sforg, line 267: Same here.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,273 @@
...
+         * @return {String} One of 'Failed - Needs Review', 'Open', or 'Completed'
+         */
+        String getBatchStatus() {
+            // if the batch has one failed row, the batch is failed regardless of the state of the other rows
+            if (this.hasFailedRow) {
+                return BDI_DataImport_API.BATCH_STATUS_FAILED;
             }
+
+            // if all rows are imported, and there are no failed or open rows, the batch is completed
+            if (this.hasImportedRow && !this.hasFailedRow && !this.hasOpenRow) {
```
vmartinov-sforg, line 285: Would it be simpler to check only for hasOpenRow here, since we checked the Failed above, and return BDI_DataImport_API.BATCH_STATUS_OPEN if true?

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,273 @@
...
+                return BDI_DataImport_API.BATCH_STATUS_COMPLETED;
+            }
+
+            // if none of the above apply, the batch is considered open
+            return BDI_DataImport_API.BATCH_STATUS_OPEN;
```
vmartinov-sforg, line 290: If we make the change above, here we could just return BDI_DataImport_API.BATCH_STATUS_COMPLETED.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,273 @@
...
+        String getBatchStatus(Id batchId) {
+            return statusesById.get(batchId).getBatchStatus();
         }
 
-        return BDI_DataImportBatch_TDTM.BATCH_STATUS_OPEN;
+        /**
+         * Reads in a list of Data Import batches, and creates or updates a BatchDataImportStatus instance for each batch
+         * @param batches A list of Data Import batches to determine the status of.
+         */
+        void ingest(List<DataImportBatch__c> batches) {
```
vmartinov-sforg, line 221: Could we use a more descriptive name for this method, maybe addBatches()  ?

vmartinov-sforg (changes requested):
This looks good and it works correctly based on my tests. A few minor suggestions.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
bdvorachek: @vmartinov-sforg An empty batch is technically an open batch. I don't think we can change the default case to be that the batch is completed because empty batches would get marked as completed.  That said, I do have some simplifications here that should make this read better.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,276 @@
...
+            // The order of these conditionals is IMPORTANT. Failed rows have precedence over Open rows when
+            // determining batch status, and Open rows have precedence over Imported rows. This means all rows
+            // must be marked as Imported for a batch to be marked as Completed, no Failed or Open rows allowed.
+            if (this.hasFailedRow) {
+                return BDI_DataImport_API.BATCH_STATUS_FAILED;
             }
+
+            if (!this.hasOpenRow) {
+                return BDI_DataImport_API.BATCH_STATUS_COMPLETED;
+            }
```
kenlewis-sfdo, line 291: How will this hold up if more Data Import Statuses are added in the future?  Or if custom DI Statuses are used in a customer org?  

It seems that this block is returning Failed if the batch has >= 1 failed row, Open if the batch has >= 1 open row, and "Completed" in all other cases.  I know we've been analyzing this scenario and we've probably already answered this, but want to make sure that A) that understanding is correct, and B) that is our expected behavior.  Since, if a Customer decides to overwrite the DI Status of a failed DI row with some custom value (but the row is still a failed row), that Batch's Batch Status might be updated by this code to be "Completed".

kenlewis-sfdo (commented):
This looks good, Ben! Thanks for this work!  I left a few comments mostly as questions and to verify that I understand the intended functionality.

vmartinov-sforg (approved):

ptewson-sfdo (approved):
Approved from my end, though worth looking at current comments!

src/classes/BDI_DataImportBatchStatus_TDTM.cls
bdvorachek: New statuses would need a priority in the list of statuses. Failed Data Import Statues are most important,  Open Statuses next Imported Statues are the least important. If we had an Intermediate DI/Batch status prioritized between Failed and Open, we'd check for the presence of Intermediate status Data Import rows before checking for Open rows, and return `BDI_DataImport_API.BATCH_STATUS_INTERMEDIATE;` here if one was found. As long as any further Batch Statuses can be prioritized in the list of statuses and the rules around Batch Statuses don't get more complicated, I think this way works.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
```diff
@@ -35,193 +35,276 @@
...
+            if (diOld.NPSP_Data_Import_Batch__c != null) {
+                return true;
+            }
+            if (diNew.NPSP_Data_Import_Batch__c != null) {
+                return true;
+            }
+        } else if (diOld.Status__c != diNew.Status__c) {
+            // status was changed; same parent
+            if (diNew.NPSP_Data_Import_Batch__c != null) {
+                return true;
```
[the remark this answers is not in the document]
bdvorachek: I thought about this a bit further. We can skip checking further statuses only if all batches in context have a failed row. 

https://github.com/SalesforceFoundation/NPSP/compare/feature/ben/giftEntryLDV...feature/ben/giftEntryLDV__shortCircuit here's a diff of what I think this change looks like.

src/classes/BDI_DataImportBatchStatus_TDTM.cls
kenlewis-sfdo: 👍🏽

src/classes/BDI_DataImportBatchStatus_TDTM.cls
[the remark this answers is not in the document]
kenlewis-sfdo: Yeah, I like these additions!  But all good - maybe you can hold onto these for a future enhancement.  Thanks for taking a look and considering!

PR #5998, opened 2020-11-02
[RD] Provide API access to RD Installments

npsp-reedestockton:
We provided API access to projected RD Installments in order to facilitate complex revenue forecasting requirements.
# Critical Changes

# Changes
 - We provided API access to projected Recurring Donation installments in order to facilitate complex revenue forecasting requirements.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
Tracking [W-8167087](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008gMESIA2/view)

npsp-reedestockton:
@RadSFDeveloper @andrewyu-salesforce Ready for review.
@force2b Not expecting your review, but just in case you are interested...

src/classes/RD2_ApiService.cls
```diff
@@ -145,18 +145,79 @@ public inherited sharing class RD2_ApiService {
...
-                if (UserInfo.isMultiCurrencyOrganization()) {
-                    currencyCode = (String) rd.get(UTIL_Currency.CURRENCY_ISO_CODE_FIELD);
-                } else {
-                    currencyCode = Userinfo.getDefaultCurrency();
-                }
-            } catch (Exception ex) {
-                // ignore any errors if for some reason we can't retrieve the currency code
-                // because usage in the ActiveSchedule inner class is only for api use
+            List<RD2_ApiService.Installment> installments = new List<RD2_ApiService.Installment>();
+            for (RD2_ScheduleService.Installment installment : service.getVisualizedInstallments(
```
force2b, line 186: I suspect that the logic in the `getVisualizedInstallments()` will default the startDate and also exclude paused installments, but it might be worth adding something here so it's clear that is the behavior. Could be just something to set the startDate to the current date if null and whatever is needed to excluded paused installments.

npsp-reedestockton: That behavior is a part of getVisualInstallments().  This is the version used calculate yearly totals.  Start Date default to currentDate is a behavior of RD2_ScheduleService.

Are you suggesting I document that or that I don't rely on it and provide my own implementation?

force2b: No. However, reading this new method in the API I'm unclear on the start date and whether skipped installments are included or not.

sam-knox (approved):

RadSFDeveloper:
@npsp-reedestockton ,  [a test](https://sfdo-metaci.herokuapp.com/builds/347640/tests) has failed. It seems related to this PR changes.

src/classes/CallableApiParameters.cls
```diff
@@ -148,10 +153,28 @@ public class CallableApiParameters {
...
 
+    /**
+     * @description Retrieve the specified Parameter as a Date value
+     * @param paramName Map key for the params collection
+     * @return Date
+     */
+    public Date getDate(String paramName) {
+        validateRequiredParameter('Date', paramName);
+        try {
+            if (params.get(paramName) != null) {
```
RadSFDeveloper, line 164: No need to check the param has the value since it is already checked in the `validateRequiredParameter`.

src/classes/Callable_API.cls
```diff
@@ -105,6 +105,19 @@ global with sharing class Callable_API implements System.Callable {
                 RD2_ApiService rd2ApiSvc = new RD2_ApiService();
                 return rd2ApiSvc.querySchedules(rdIds);
 
+            } when 'rd2.queryinstallments' {
+                Set<Id> rdIds = paramService
+                    .withRequiredParamCheck()
+                    .getSetId(CallableApiParameters.PARAM_RECORD_IDS);
+                Date startDate = paramService
+                    .withoutRequiredParamCheck()
```
RadSFDeveloper, line 113: Suggestion: since required attribute is `false` by default, not need to explicitly check `withoutRequiredParamCheck()`.

src/classes/RD2_ApiService.cls
```diff
@@ -55,7 +55,7 @@ public inherited sharing class RD2_ApiService {
 
             List<RD2_ApiService.Schedule> schedules = new List<RD2_ApiService.Schedule>();
             for (RecurringDonationSchedule__c s : rd.RecurringDonationSchedules__r) {
-                schedules.add(new RD2_ApiService.Schedule(s, rd));
+                schedules.add(new RD2_ApiService.Schedule(s, rd, getCurrencyCode(rd)));
```
RadSFDeveloper, line 58: Since the RD is passed into the `RD2_ApiService.Schedule`, then `getCurrencyCode(rd)` should be the schedule responsibility, ie part of that class. Or make the method static so it can be accessed from `Schedule` and `Installment` inner classes.
Suggestion: the `RD2_ApiService.Schedule` inner class can be called `Schedule` unless there are compile issues.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,116 @@ private class RD2_ApiService_TEST {
...
+
+        System.assertNotEquals(0, installmentsByRd.size(),
+            'Recurring donation installments should be returned when Start Date is missing');
+    }
+
+    /**
+     * @description Validate an empty installment response when End Date is missing
+     */
+    @IsTest
+    private static void shouldReturnNoInstallmentsWithMissingEndDate() {
```
RadSFDeveloper, line 319: Why not make both start and end date as optional, and if the end date is not specified, then return default 12 installments?
However, the End Date is required based on the required param check in the `Callable_API`, so I am wondering why no exception is thrown when the end date is not provided?

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,116 @@ private class RD2_ApiService_TEST {
...
+        npe03__Recurring_Donation__c rd = rdGateway.getRecords()[0];
+
+        Map<Id, List<Installment>> installmentsByRd = retrieveInstallmentsUsingApi(
+            date.today(),
+            date.today().addYears(2),
+            rd.Id);
+        System.assertEquals(25, installmentsByRd.get(rd.Id).size(),
+            '25 Recurring donation installments should be returned when there is no pause schedule');
+
+        RD2_ScheduleService_TEST.createPauseSchedule(rd.Id, Date.today().addMonths(1), Date.today().addMonths(2));
```
RadSFDeveloper, line 347: Let's create two tests out of this one since there are two concepts tested: installments when params are valid and installments when the RD is paused for two installments.
Let's store 25 and 23 as vars so it is easier to know what these numbers represent.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -268,6 +379,30 @@ private class RD2_ApiService_TEST {
...
     }
 
+    /**
+     * @description Utility method to retrieve installments via Callable_API
+     * @param startDate date
+     * @param endDate date
+     * @param rdId Recurring Donation Id
+     * @return Deserialized api response
+     */
+    private static Map<Id, List<Installment>> retrieveInstallmentsUsingApi(
```
RadSFDeveloper, line 389: Let's name this helper method as what the `Callable_API` name is: `queryInstallments()`. The arguments can be specified on the same line then.

RadSFDeveloper (commented):
Nice work on the retrieving the installments. Here are code review comments, and one request to make end date optional. @mpusto , @force2b , @andrewyu-salesforce

src/classes/RD2_ApiService_TEST.cls
npsp-reedestockton: I don't believe a default end date makes sense.  We could be dealing with any period - weekly, monthly, yearly - so 12 installments could represent a wide range of dates.  Since the purpose of this API is to assist in custom revenue forecasting scenarios, a specific end date makes sense to me.

npsp-reedestockton: Pretty sure that the reason no exception is thrown in this test currently is that an end date parameter is being provided.  It is null, but it exists.  Exception would be thrown when a required parameter is not provided at all.

npsp-reedestockton:
@RadSFDeveloper Code review suggestions have been addressed.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,143 @@ private class RD2_ApiService_TEST {
...
+    private static void shouldReturnNoInstallmentsWithMissingEndDate() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+        npe03__Recurring_Donation__c rd = rdGateway.getRecords()[0];
+
+        Map<Id, List<Installment>> installmentsByRd = queryInstallments(
+            date.today(),
+            null,
+            rd.Id);
+
+        System.assertNotEquals(0, installmentsByRd.size(),
```
RadSFDeveloper, line 328: Based on latest changes, when End Date is not provided, should default date be used, ie date that is start date *next* year, so 12 installments should be returned?

src/classes/RD2_ApiService.cls
```diff
@@ -145,18 +145,108 @@ public inherited sharing class RD2_ApiService {
...
+    /**
+     * @description Calculates endDate to produce 1 year of installment data (startDate + 1 year - 1 day).
+     * One day is subtracted to ensure that one year of installments is included. Without the adjustment,
+     * an extra installment is reported in most scenarios.
+     * @param startDate Date
+     * @return endDate Date
+     */
+    private Date getDefaultEndDate(Date startDate) {
+        Date endDate = startDate.addYears(1);
+        if (startDate.day() == endDate.day()) {
```
RadSFDeveloper, line 223: The method description and this statement not align. It seems based on the description, that one day is always subtracted.

RadSFDeveloper (commented):
Awesome work @npsp-reedestockton ! Two minor comments have been logged.

npsp-reedestockton:
@RadSFDeveloper I've addressed the code review suggestions.  Thanks for the good catches!

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,144 @@ private class RD2_ApiService_TEST {
...
+    private static void shouldReturnValidInstallmentsWithMissingStartDate() {
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+        npe03__Recurring_Donation__c rd = rdGateway.getRecords()[0];
+
+        Map<Id, List<Installment>> installmentsByRd = queryInstallments(
+            null,
+            date.today().addYears(2),
+            rd.Id);
+
+        System.assertNotEquals(0, installmentsByRd.size(),
```
RadSFDeveloper, line 311: I would expect 24 installments since start date should default to today?

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,144 @@ private class RD2_ApiService_TEST {
...
+            'Recurring donation installments should be returned when Start Date is missing');
+    }
+
+    /**
+     * @description Validate a valid installment response when End Date is missing
+     */
+    @IsTest
+    private static void shouldReturnNoInstallmentsWithMissingEndDate() {
```
RadSFDeveloper, line 319: Let's update the test name to indicate that one year worth of installments will be returned when end date is not specified.

RadSFDeveloper (commented):
Thank you @npsp-reedestockton  for making updates. Two comments related to tests are logged, can you please take a look?

npsp-reedestockton:
@RadSFDeveloper Next set of changes ready for review.  These were really good catches.  One of the tests would have failed on the first of every month.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,149 @@ private class RD2_ApiService_TEST {
...
+
+        System.assertEquals(0, installmentsByRd.size(),
+            'There should be no recurring donation installments returned when RD Id Set is empty');
+    }
+
+    /**
+     * @description Validate a valid installment response when start date is not passed in
+     */
+    @IsTest
+    private static void shouldReturn25ValidInstallmentsWithDefaultStartDate() {
```
RadSFDeveloper, line 302: Let's rename the test to be more generic, such as `shouldReturnValidInstallmentsWithDefaultStartDate()`.

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,149 @@ private class RD2_ApiService_TEST {
...
+        final Integer installmentCount = 24;
+        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
+        npe03__Recurring_Donation__c rd = rdGateway.getRecords()[0];
+
+        Map<Id, List<Installment>> installmentsByRd = queryInstallments(
+            date.today(),
```
RadSFDeveloper, line 308: Should start date be null?

src/classes/RD2_ApiService_TEST.cls
```diff
@@ -246,6 +247,149 @@ private class RD2_ApiService_TEST {
...
+            date.today().addYears(2).addDays(-1),
+            rd.Id);
+
+        for (Installment ins: installmentsByRd.get(rd.Id)) {
+            System.debug(ins);
```
RadSFDeveloper, line 313: Let's remove debug code.

RadSFDeveloper (approved):
Nice work @npsp-reedestockton ! Approving so testing can start, however there are minor comments for a test method.

src/classes/RD2_ApiService_TEST.cls
npsp-reedestockton: Fixed.  Thanks!

PR #6001, opened 2020-11-03
RD2: Display entry form modal

### LINKED ISSUE #5968: Enhanced Recurring Donations - Unable to edit a Recurring Donation if the RD2_EntryForm has already been loaded in an Edit context in the same browser session
### Summary of the Issue

If a user loads the RD2_EntryForm in a record Edit context and then tries to load it again, by editing the same or another recurring donation, in the same session they will not be able to do so. The RD2_EntryForm model will briefly display and then disappear so the user cannot edit the record. Trying to click Edit again on any Recurring Donation will result in the same behaviour, so the user is prevented from editing more than one recurring donation in the same browser session.

Note: The user does not need to actually made an edit to the record when loading the RD2_EntryForm for the 1st time, simply loading the form and then cancelling the form will also result in them not being able to load the form again when editing any record.

### Steps to Replicate

- Ensure Enhanced Recurring Donations is enabled and the RD2_EntryForm is in place as a Lightning Experience Override for the Edit action/button
- Create a new Recurring Donation
- Click Edit on the Recurring Donation, and then Save or Cancel
- Click Edit on the same Recurring Donation, or any other Recurring Donation
- Observe the RD2_EntryForm modal displays briefly but then disappears
- Clicking on Edit again results in the same behaviour

**Potential Workarounds until the issue is addressed**

After editing a Recurring Donation, do one of the following, and then repeat this each time you need to edit more than one Recurring Donation in the same session:

- Force refresh the page (Windows: ctrl + F5; Mac/Apple: Apple + R or command + R; Linux: F5)
- Clear the browser cache
- Logout of Salesforce and then back in

andrewyu-salesforce:
As a user I want to edit more than one Recurring Donation during a browser session to accomplish my daily tasks.

# Critical Changes
 - We applied changes for the Recurring Donation modal displayed when user clicks New or Edit buttons, so it does not automatically close when the page is loaded in the same lightning session.

# Changes

# Issues Closed
- Fixes #5968 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


andrewyu-salesforce:
@npsp-reedestockton @RadSFDeveloper The PR is ready for review.

src/lwc/rd2EntryForm/rd2EntryForm.html
```diff
@@ -1,176 +1,197 @@
...
+        aria-labelledby="modal-heading-01"
+        aria-modal="true"
+        aria-describedby="modal-content-id-1"
+        class="slds-modal slds-fade-in-open"
+        onkeyup={handleKeyUp}
+    >
+        <div class="slds-modal__container">
+            <header class="slds-modal__header slds-clearfix">
+                <lightning-button-icon
+                                class="slds-modal__close"
```
RadSFDeveloper, line 14: Let's move indentation to be as for other elements ( one tab instead of three in this case).

src/lwc/rd2EntryForm/rd2EntryForm.css
```diff
@@ -11,4 +11,13 @@
     width: 100%;
     text-align: center;
     top: calc(50% + 3em);
+}
+
+.slds-modal__content {
```
RadSFDeveloper, line 16: It seems the width of the modal is smaller than previously, this when the Recurring Period is Advanced, the Installment Period is not aligned with other fields.
@ravensisco , please add into the test charter a test for a user with non-English locale and see how fields are rendered. Also, please ensure the form is rendered as expected in different browsers.

Screen Shot 2020-11-04 at 10 15 17 AM

src/aura/RD2_EntryForm/RD2_EntryForm.cmp
```diff
@@ -14,14 +14,12 @@
...
-
-    <lightning:overlayLibrary aura:id="overlayLib" />
+    <aura:attribute name="parentId" type="Id" description="Reference to the parent Id if opened from related list" />
+    <aura:attribute name="renderModal" type="Boolean" default="false" description="Boolean to render Modal" />
 
     <aura:handler name="init" value="{!this}" action="{!c.doInit}" />
-
-    <c:utilDedicatedListener eventName="rd2EntryFormEvent" onreceiveevent="{!c.handleModalEvent}" />
+ 
+    <aura:if isTrue="{!v.renderModal}">
```
RadSFDeveloper, line 22: When a New form is open and data populated, then click Cancel, Save or close the modal on the upper right corner, click New again, the form is pre-populated with data entered in the previous data form.

Moreover, when I saved *new* RD with Elevate credit card data populated, then viewed the RD tab again and clicked New the following screen was displayed:
Screen Shot 2020-11-04 at 10 30 43 AM

RadSFDeveloper (commented):
Hi @andrewyu-salesforce  -- it seems the form state persists between entries. Can you please check into comments below?

andrewyu-salesforce:
@RadSFDeveloper @npsp-reedestockton The PR is updated for review again. The problem with the modal in weird state is that Salesforce does not refresh the component when clicking the 'New' button on RD list view. Instead, the component is loaded from cache and no `init/connectedcallback` function is fired. The fix I put on is forcing the page to reset when the modal is closed. This way the when the modal is loaded again from Cache, the form will be set to the initial state.

src/lwc/rd2EntryForm/rd2EntryForm.css
andrewyu-salesforce: add modal-small to increase the width

RadSFDeveloper (commented):
All previously tested scenarios pass: the modal state does not persist when a user clicks New or Edit. There is one minor issue with Campaign field: the field flickers when lookup fields are displayed and the Elevate widget is displayed (regardless if the user clicks "Do not use Elevate"). The issue is not present when Custom Fields section has fields. However, I have only two campaigns, so not sure how the Campaign field would display if more campaigns are displayed. Can you please take a look? Otherwise, everything works great! Thank you!

andrewyu-salesforce:
@RadSFDeveloper Thanks for the feedback on the flickering. The issue is fixed by removing !important in the css file.

andrewyu-salesforce:
@RadSFDeveloper @npsp-reedestockton The code is ready for review again.

RadSFDeveloper (approved):
Looks awesome, thank you so much @andrewyu-salesforce !

npsp-reedestockton (approved):
Excellent work, Andrew!!

PR #6015, opened 2020-11-11
Add preflight checks

davidmreed:
This PR is dependent on the merge and release of [this CumulusCI PR](https://github.com/SFDO-Tooling/CumulusCI/pull/2162).

We now preflight-check before installing NPSP:

- Chatter must be turned on for all install plans
- Advanced Currency Management must be turned off for all plans
- NPSP must be installed to run the Reports & Dashboards plan
- NPSP must _not_ be installed to run the Trailhead installer.

Installer for testing (*VPN required*): (to be republished)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


cumulusci.yml
```diff
@@ -1209,6 +1215,15 @@ plans:
             - when: "'.my.' not in org_config.instance_url"
               action: error
               message: "Please enable My Domain in your org prior to installing."
+            - when: "not tasks.check_chatter_enabled()"
+              action: error
+              message: "Nonprofit Success Pack requires Chatter. Please enable Chatter in your org and try again."
+            - when: "not 'npsp' in org_config.installed_packages"
+              action: error
+              message: "Nonprofit Success Pack is already installed in this org. This installer cannot be used with orgs that already have NPSP."
+            - when: "'CurrencyType' in tasks.check_available_sobjects()"
```
codexmarcel, line 1224: Cool!

codexmarcel (approved):

cumulusci.yml
force2b: @davidmreed  Is this the correct object for ACM? I thought it was `DatedConversionRate`? It's also important to consider that once ACM is enabled in an org this object remains visible even if the admin disables ACM. As a result, with this check, it will never be possible to install NPSP in an org that ever had ACM enabed.

davidmreed: Ah, I didn't know that, @force2b. Thank you for catching this. `CurrencyType` does successfully discriminate between ACM-on and ACM-off, but I didn't realize it stayed visible with ACM disabled. We'll have to try a different strategy here, possibly a Tooling API query against Settings.

force2b: https://trailblazer.salesforce.com/issues_view?id=a1p3A0000017yARQAY

force2b: I'm fairly sure that `CurrencyType` is always visible in an org multi-currency enabled, even if not ACM.

davidmreed: I did some testing and found a better strategy.

`CurrencyType` and `DatedConversionRate` are _both_ visible in multicurrency orgs, even with ACM off. However, whether or not `DatedConversionRate` is `createable` is a reliable discriminator. It is createable only with ACM on, even if ACM was previously on in the org and has been turned off.

I'll build a new preflight.

davidmreed:
Closing for now - requires new CumulusCI release.

PR #6018, opened 2020-11-11
RD2: Add validation migration check for inconsistent RD data

### LINKED ISSUE #5947: Enhanced Recurring Donations: Amount field value set incorrectly when Closed RD includes any value other than 1 in "Number of Planned Installments" field before migration
[W-8215323](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008VB5VIAW/view)
[W-8222341](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EyOeIAK/view)

**Issue:**

After Enhanced RD Migration, Closed Recurring Donations had the following unexpected changes made: 
- Amount field displays the value from “Installment Amount” field (unexpected and incorrect)
- Installment Amount updated to 0 (unexpected but correct)
- Number of Planned Installments updated to Null (unexpected but correct)

When the Closed status RD had any value in the "Number of Planned Installments" other than 1, it caused the "Installment Amount" field value to be set to a value other than null, and other than the "Amount" field value on the RD. This in itself isn't an issue, but becomes an issue during the RD1 - RD2 migration process.

**Cause:**

Amount field value should be maintained during migration for all Closed RD's that are migrated when they don't meet the criteria to use Installment Amount value as defined in the guide:
- Open Ended Status = None
- Schedule Type = Divide By or is null

Instead, it's using the "Installment Amount" field incorrectly for all Closed RD's when the "Number of Planned Installments" is a value of anything other than 1.
Screen Shot 2020-10-14 at 9 01 51 AM

**Steps to Repro:**

1. In an org with Legacy RD's enabled, create a RD for $1000.
- Set the Open Ended Status to Open
- Installment Period to Monthly (this happens for Annual and Quarterly as well in testing)
- Leave the Schedule Type as None (which will be Null)
- Leave the default for the Number of Planned Installments field, which should be 1
2. Save and note that the Installment Amount field should be the same as the Amount field, $1000 in this example. 
3. Edit the RD and set the Open Ended Status to Closed. Note the Installment Amount is left as $1000.
4. Edit the RD and change the Number of Planned Installments value from 1 to any other number, say 10. Note that the Installment Amount is changed to $100.
5. Perform migration to Enhanced RD, and review example Opty.
- Expected Result: RD is migrated to RD2 with Amount left as $1000.
- Actual Result: RD is migrated to RD2 with Amount set to $10.

Customer specific details available on WI if needed.

npsp-reedestockton:
# Critical Changes

# Changes
 - We updated the Enhanced Recurring Donation upgrade (step 1, Validation) to identify inconsistent data on Closed Recurring Donations. IMPORTANT: If you are in the process of updating to Enhanced Recurring Donations and have already completed step 1 (Validation), run the validation again after this code update is pushed to your org (DATE). Otherwise, the validation might have missed identifying the inconsistent data.

# Issues Closed
- Fixes #5947

# Community Ideas Delivered

# New Metadata
- Custom Labels:
    - RD2_DataMigrationInvalidPlannedInstallments

# Deleted Metadata


npsp-reedestockton:
Tracking [W-8222341](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EyOeIAK/view)

npsp-reedestockton:
@andrewyu-salesforce @RadSFDeveloper Ready for your review.  Thanks!
@sam-knox This PR has a label for you to review.  Thanks!

RadSFDeveloper:
Hi @npsp-reedestockton  -- one [test](https://sfdo-metaci.herokuapp.com/builds/353204/tests) is failing for the PR. Can you please check it out?

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -131,6 +131,8 @@ public class RD2_DataMigrationMapper {
      */
     private void setInstallmentFields(npe03__Recurring_Donation__c rd) {
 
+        validateAmountCanBeCalculated(rd);
```
RadSFDeveloper, line 134: Is there a reason to validate the amount at this line? How about checking the value before current line 156, ie before setting RD amount?

src/classes/RD2_DataMigrationMapper.cls
```diff
@@ -159,6 +161,20 @@ public class RD2_DataMigrationMapper {
         }
     }
 
+    /**
+     * @description Determines whether Amount can be calculated.  Throws an exception if not.
+     * @param rd Recurring Donation
+     */
+    private void validateAmountCanBeCalculated(npe03__Recurring_Donation__c rd) {
+
+        if (rd.npe03__Open_Ended_Status__c == RD_Constants.OPEN_ENDED_STATUS_CLOSED
```
RadSFDeveloper, line 170: Based on the `setOtherFields()` method, a fixed RD is the one with `rd.npe03__Open_Ended_Status__c == RD_Constants.OPEN_ENDED_STATUS_NONE || rd.npe03__Open_Ended_Status__c == null` so when an RD is closed, we do not know if it was a fixed or an open RD. Based on the `setOtherFields()`, the "Number of Planned Installments"(`npe03__Installments__c`) can be populated or null for an open RD. Thus, I think we also need to check that `rd.npe03__Installments__c != null`. The lack of that check is the reason why the existing test is failing.

src/classes/RD2_DataMigration_TEST.cls
```diff
@@ -1932,6 +1932,55 @@ private class RD2_DataMigration_TEST {
...
+            .withDateEstablished(DATE_ESTABLISHED)
+            .build();
+
+        npe03__Recurring_Donation__c convertedRD;
+        try {
+            convertedRD = new RD2_DataMigrationMapper(rd)
+                .convertToEnhancedRD();
+        }
+        catch (RD2_DataMigrationMapper.MigrationException e) {
+            System.assertEquals(true,
```
RadSFDeveloper, line 1976: Let's check exceptions like we usually do with checking an `actualException` after the `try/catch` block.

RadSFDeveloper (commented):
Hi @npsp-reedestockton  -- below are comments related to changes of the current code. I have questions about requirements, and wondering if a check needs to be done for a fixed active RD not just a closed RD. Will clarify them in the WI.

src/classes/RD2_DataMigrationMapper.cls
npsp-reedestockton: I'm fine to change this if it helps with readability, but it does mean that we will sometimes do unnecessary processing.

RadSFDeveloper: Let's keep related code together in a cluster ;)

src/classes/RD2_DataMigrationMapper.cls
npsp-reedestockton: I'm not sure I agree with the conclusion here.  I think the failing test no longer represents a valid outcome.  Let's discuss offline during dev office hours.

RadSFDeveloper: A closed RD can be a fixed or open RD before being closed. If an open legacy RD is closed and "Number of Planned Installments" is not provided (null), then this new validation check will fail validation/migration process but it should not be.

npsp-reedestockton: My reading of the failing test is that it precisely creates the conditions for AC1 of the current work item.  And that AC specifies that the migration should fail.

src/labels/CustomLabels.labels
```diff
@@ -1577,6 +1577,14 @@
         <shortDescription>Open Ended Status cannot be converted to enhanced Recurring Donation format</shortDescription>
         <value>Recurring Donation Open Ended Status &quot;{0}&quot; cannot be converted to the Enhanced Recurring Donations format. You must use one of these values: Open, Closed, or None.</value>
     </labels>
+    <labels>
+        <fullName>RD2_DataMigrationInvalidPlannedInstallments</fullName>
+        <categories>Recurring-Donations, Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Planned Installments cannot be migrated to enhanced Recurring Donation format</shortDescription>
+        <value>Number of Planned Installments must be 1 when Open Ended Status is Closed and Divide By/Multiply By is not populated.</value>
```
sam-knox, line 1586: @npsp-reedestockton - Open Ended Status = Closed just means closed. We're looking for fixed-lengths RDs here? In that case we should be looking for Open Ended Status = None.

npsp-reedestockton: @sam-knox My understanding is that the issue we are trying to fix is a result of the fact that Open Ended Status = Closed RDs may have been either open or fixed.  We can't calculate amount unless Divide By/Multiply By and Planned installments gives us a way to do that.  @RadSFDeveloper has substantive questions about this PR as well, so I think we need to take the discussion into the work item and get straight on the ACs.

src/classes/RD2_DataMigrationMapper.cls
npsp-reedestockton: I've concluded that my understanding of the requirements was incorrect and your interpretation is correct.  I've updated the code to reflect your original suggestion.

src/labels/CustomLabels.labels
npsp-reedestockton: @sam-knox  I've updated the error label.

sam-knox (changes requested):
Here's my suggestion for this label:

"When Open Ended Status is Closed and Schedule Type is not populated, Planned Installments must be 1."

The field is Schedule Type where Divide By/Multiply By are picklist values.

npsp-reedestockton:
@RadSFDeveloper @sam-knox Requested changes have been implemented.

RadSFDeveloper (approved):
Looks good, thank you @npsp-reedestockton !

sam-knox (changes requested):
Sorry Reede, my fault here. I got the field label wrong for Planned Installments. It should be Number of Planned Installments.

"When Open Ended Status is Closed and Schedule Type is not populated, Number of Planned Installments must be 1."

npsp-reedestockton:
Thanks, @sam-knox.  "Number of Planned Installments" change made to label.

andrewyu-salesforce (approved):
@npsp-reedestockton Nice work

npsp-reedestockton:
@sam-knox The change you requested on this PR has been made.  Could you please approve when you get a moment?

sam-knox (approved):

PR #6041, opened 2020-11-16
RD2: Set origin type to CRM on Elevate recurring commitment

RadSFDeveloper:
We are setting Elevate origin type to "CRM Connect" when a recurring commitment is created from NPSP.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


RadSFDeveloper:
Tracking [W-8235298](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008F7isIAC/view)

src/classes/PS_CommitmentRequest.cls
```diff
@@ -155,6 +155,7 @@ public inherited sharing class PS_CommitmentRequest {
             : UserInfo.getDefaultCurrency();
 
         reqBody.withCurrency(currencyCode);
+        reqBody.withMetadata(new Metadata());
```
npsp-reedestockton, line 158: Would it be better for the metadata to be in the buildCreateRequest method?  It seems more suited to the sort of data being built there.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -521,4 +537,21 @@ public inherited sharing class PS_CommitmentRequest {
...
         public String expirationYear;
     }
+
+    /***	
+     * @description Metadata property on the commitment request	
+     */	
+    public class Metadata {	
+        public String originType;	
+        public String originDisplayName;	
+        public String campaignCode;
```
npsp-reedestockton, line 547: are originDisplayName and campaignCode used anywhere at this point?

src/classes/PS_CommitmentRequest.cls
```diff
@@ -296,6 +297,11 @@ public inherited sharing class PS_CommitmentRequest {
         public String currencyCode;
         public List<Schedule> schedules;
 
+        /***	
+         * @description Optional Metadata containing origin type and other properties
```
npsp-reedestockton, line 301: I think it would be better to either enumerate all properties or just refer to Metadata properties.

npsp-reedestockton (changes requested):

src/classes/PS_CommitmentRequest.cls
RadSFDeveloper: Not as of this time, but there is a pending work for the Campaign code on create (W-8334258). I can remove attributes, but campaign code is a placeholder for the future work.

src/classes/PS_CommitmentRequest.cls
RadSFDeveloper: That is a good suggestion, I thought of changing this so it is set as a default, but it will not work when campaign code is to be passed. Since fields such as gateway, donor info, currency, etc are set in the `getRequestBody()`, so `Metadata`  should be set here as well and minimize changes when future WI is worked on. So, even though we can set `Metadata` in various methods, it is better to leave it where it will be needed for the campaign code. Let me know if I overlooked something.

src/classes/PS_CommitmentRequest.cls
RadSFDeveloper: The unused Metadata properties have been removed, can you please review the PR again? Thanks.

src/classes/PS_CommitmentRequest.cls
npsp-reedestockton: Yes, that makes sense.

npsp-reedestockton (approved):

andrewyu-salesforce (approved):
@RadSFDeveloper Great work. Everything looks good to me

PR #6068, opened 2020-11-20
CCI: Flows to run the full CRLP Validation script without manual intervention.

force2b:
- Customizable Rollup Validation Script:
    - New Command: `cci run flow test_customizable_rollups --org {orgalias}`. Requires a scratch org with unmanaged NPSP deployed.
    - Executes a full _regression_ against Customizable Rollups by creating a set of data and executing both Legacy and Customizable Rollups against the data, verifying that the rollup fields on target Account, Contact, RD and GAU records are accurate.
    - Alternate Flow: `test_customizable_rollups_multicurrency`: Support Multi-Currency Orgs (automatically creates a CAD currency with the expected exchange rate)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


davidmreed:
Do we have a defined way to execute these tests and surface failures automatically? I worry about test plans that don't have automation attached to them.

force2b:
@davidmreed These two flows haven't been attached to anything that runs automatically as a _build_ during a commit. For the moment, they automate something that has been a tedious manual process. That said, I wouldn't mind having these run automatically on merges into master as an extra validation step. Not counting the org build time, the flow itself completes in around 5-10 minutes. Is there a way to connect these to the build process? And, if so is it something I can do?

Also, if it's possible, there is value in running these on commits into a branch that have any files changed matching a pattern of `CRLP_*.cls`. Not a requirement, but would be useful when there are changes to CRLP related classes.

force2b:
@davidmreed and @prescod This is another PR I'd like to get merged, and subsequently automated. Theoretically we can automate without merging, but I've found it helpful to be able to execute the new `test_customizable_rollups` flow from any branch. I've just had to do it the hard way by building my scratch org off of one branch, changing to this branch, running the flow, and then changing back. The new flow doesn't introduce any changes to everyday functionality. 

The value in automating is that it validates calculations on rollups. It's useful to make sure we haven't broken rollups, but at the same time that is not something we've broken in the past. Sill, being able to run this maybe on merges into master would be useful - though it needs to run against unmanaged code. Automation can come later - and maybe there's a more formal way to request that now if I'm reading the new confluence document correctly.

prescod (approved):

davidmreed:
I'm going to go ahead and approve this and I'd like to revisit getting it to run in automation and appropriately capture the results of the testing.

davidmreed (approved):

PR #6080, opened 2020-11-30
[GS][W-8515435] Add 'WITH SECURITY_ENFORCED' to SOQL

fparodi-salesforce:
# Critical Changes

# Changes
 Add 'WITH SECURITY_ENFORCED' to SOQL
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/GS_NonprofitTrialOrgService.cls
+++ b/src/classes/GS_NonprofitTrialOrgService.cls
@@ -138,11 +138,10 @@ public with sharing class GS_NonprofitTrialOrgService {
      * {"Sandbox":false,"TrialExpiration":"2020-11-29T00:00:00.000Z", "applicationDate":"2020-10-20"} Simulates a trial org which has applied for licenses.
      */
     private static TestingConfig getTestingConfiguration() {
-        List<StaticResource> staticResources = [
-            SELECT Body FROM StaticResource
-            WHERE Name = :STATIC_RESOURCE_NAME
-            WITH SECURITY_ENFORCED
-        ];
+        List<StaticResource> staticResources = [SELECT Body
+                                                FROM StaticResource
+                                                WHERE Name = :STATIC_RESOURCE_NAME
+                                               ];
         if (!staticResources.isEmpty()) {
             return (TestingConfig) JSON.deserialize(staticResources[0].Body.toString(), TestingConfig.class);
         }
@@ -162,11 +161,9 @@ public with sharing class GS_NonprofitTrialOrgService {
      * @return Organization with TrialExpirationDate and IsSandbox queried.
      */
     private static Organization getOrgInformation() {
-        return [
-            SELECT TrialExpirationDate, IsSandbox
-            FROM   Organization
-            WITH SECURITY_ENFORCED
-        ];
+        return [SELECT TrialExpirationDate,
+                       IsSandbox
+                FROM   Organization];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


fparodi-salesforce:
[W-8515435](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008kQ1vIAE/view)

src/classes/GS_NonprofitTrialOrgService.cls
```diff
@@ -161,9 +162,11 @@ public with sharing class GS_NonprofitTrialOrgService {
      * @return Organization with TrialExpirationDate and IsSandbox queried.
      */
     private static Organization getOrgInformation() {
-        return [SELECT TrialExpirationDate,
-                       IsSandbox
-                FROM   Organization];
+        return [
+            SELECT TrialExpirationDate, IsSandbox
+            FROM   Organization
+            WITH SECURITY_ENFORCED
```
force2b, line 168: This clause should only be used on standard and custom objects, not system objects.

src/classes/GS_NonprofitTrialOrgService.cls
```diff
@@ -138,10 +138,11 @@ public with sharing class GS_NonprofitTrialOrgService {
      * {"Sandbox":false,"TrialExpiration":"2020-11-29T00:00:00.000Z", "applicationDate":"2020-10-20"} Simulates a trial org which has applied for licenses.
      */
     private static TestingConfig getTestingConfiguration() {
-        List<StaticResource> staticResources = [SELECT Body
-                                                FROM StaticResource
-                                                WHERE Name = :STATIC_RESOURCE_NAME
-                                               ];
+        List<StaticResource> staticResources = [
+            SELECT Body FROM StaticResource
```
force2b, line 142: This clause should only be used on standard and custom objects, not system objects.

force2b (changes requested):

force2b (changes requested):
There's another query on line 69 that should be updated.

force2b (approved):
This looks good. 

I want to make an important note now that I'm aware that there will be a similar page for End Users. The concept with the Admin pages was that only a "true" admin would ever use these pages so the additional query/dml security makes sense. However, when it comes to the End User page, I think that we may want to reconsider the use of `WITH SECURITY_ENFORCED` for that pages controller. The premise is that we'd want _any_ end user to be able to access the new page even if an Admin hasn't explicitly granted the User crud and fls to the CMDT and `GetStartedCompletionChecklistState__c` objects - which will be especially true in smaller P10 orgs where there might not even be admin present. My preference would be that for _that_ specific EndUser page the controller uses `Without Sharing` on the class and that queries/dml are done without reference to `With Security_Enforced`' - if everyone agrees.
cc @dcaballeroSf @smarovich @ameyersalesforce

ameyersalesforce:
> This looks good.
> 
> I want to make an important note now that I'm aware that there will be a similar page for End Users. The concept with the Admin pages was that only a "true" admin would ever use these pages so the additional query/dml security makes sense. However, when it comes to the End User page, I think that we may want to reconsider the use of `WITH SECURITY_ENFORCED` for that pages controller. The premise is that we'd want _any_ end user to be able to access the new page even if an Admin hasn't explicitly granted the User crud and fls to the CMDT and `GetStartedCompletionChecklistState__c` objects - which will be especially true in smaller P10 orgs where there might not even be admin present. My preference would be that for _that_ specific EndUser page the controller uses `Without Sharing` on the class and that queries/dml are done without reference to `With Security_Enforced`' - if everyone agrees.
> cc @dcaballeroSf @smarovich @ameyersalesforce

I mentioned this in Slack too, but I think we should use Without Sharing for both the admin and end user page (or whatever ensures that anyone can access both pages). For the admin page, our target audience are admins who are JUST starting - they have no idea about permissions and we don't want to present additional hurdles to them - we want to make it as easy as possible. Plus, there isn't any content on the page that is protected or high risk or anything like that. 

For end users, we want end users to be able to access it, so - once again, I think we should NOT enforce permissions on these pages.

force2b:
Thanks @ameyersalesforce. @dcaballeroSf To make these pages truly visible to anyone, we'll also need to change the page controllers to use `without sharing` (alternatively move queries into an inner class that is defined as `without sharing`).

PR #6082, opened 2020-12-01
[Robot] New status mapping test case with custom picklist status values

rjanjanam:
This test case is to ensure that the status mappings for the exiting values
 remain intact . The newly added status mapping values appear in the status
 mappings section and remain unmapped.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


rhakeshj (approved):

robot/Cumulus/resources/ObjectMangerPageObject.py
```diff
@@ -149,4 +149,44 @@ def create_custom_field(self, **kwargs):
 			elif type.lower() == 'formula':
 				self.create_formula_field(kwargs['Field_Name'],kwargs['Formula'])
 			elif type.lower() == 'text':
-				self.create_text_field(kwargs['Field_Name'])
\ No newline at end of file
+				self.create_text_field(kwargs['Field_Name'])
+		
+	
+	@capture_screenshot_on_error
+	def add_picklist_values(self, **kwargs):
```
skristem, line 156: what would happen if the same keyword is executed again and again in the same org?

skristem, line 156: Also, can these picklist values be added as part of org setup?

robot/Cumulus/resources/locators_50.py
```diff
@@ -259,6 +259,9 @@
         'search_result': '//tbody/tr/td/a/span[contains(text(),"{}")]',
         'formula_txtarea': '//textarea[@id = "{}"]',
         'object_result': '//th/a[text()="{}"]',
+		'field_result': '//td/a/span[text()="{}"]',
```
skristem, line 262: please adjust formatting

robot/Cumulus/resources/ObjectMangerPageObject.py
rjanjanam: It will fail may be because there would be a duplicate constraint. But on metaCI the org is in a clean state so there won't be any problem. The testcase specially wants these values and that is part of the acceptance criteria.

rjanjanam: I was not sure if there is a way to delete these picklist values from UI

rjanjanam: Modified it to have unique picklist values so that rerunning test works always

rjanjanam: modified it to be part of setup

robot/Cumulus/resources/locators_50.py
rjanjanam: This looks fine on the editor. Not sure why it is showing it that way here

skristem (approved):
looks good

PR #6086, opened 2020-12-02
[Robot] GE batch numbers test case

skristem:
This test case adds two batch number formats and activates one. Verifies new batch created after the format is activated carry the batch number in the specified format and starting number. Deactivate the active number format, Activate the other format and verify the new batch carries the correct format. Deactivate the number format, Verify new batch does not have a batch number assigned.
W-8246942
To execute test run command :
cci task run robot -o suites robot/Cumulus/tests/browser/gift_entry/batch_numbers.robot --org <org name>

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/NPSPSettingsPageObject.py
```diff
@@ -247,4 +247,25 @@ def verify_status_to_state_mappings(self, **kwargs):
                 print(f"actual mapping found for {key} is {actual_value}")
                 assert value == actual_value, "Expected {} value to be {} but found {}".format(key,value, actual_value)
             else:
-                print("Right keys under status mapping fields table not found")    
\ No newline at end of file
+                print("Right keys under status mapping fields table not found")
+
+    @capture_screenshot_on_error
+    def set_batch_number_format_status(self,number_format,status):
+        """Deactivates the given batch number format by clicking on dropdown and selecting option
```
rjanjanam, line 254: This method deactivates and activates both depending on the status . The description talks about only deactivation. Please update to reflect both.

robot/Cumulus/resources/NPSP.py
```diff
@@ -1669,4 +1669,22 @@ def query_and_store_records_to_delete(self,object_name,**kwargs):
             for i in records:
                 self.salesforce.store_session_record(object_name,i)
 
+    @capture_screenshot_on_error
+    def verify_table_contains_row(self,table_name,record,**kwargs):
```
rjanjanam, line 1673: Is this keyword generic enough for any usecase where we wanted to validate table contains row ?

robot/Cumulus/resources/NPSP.py
```diff
@@ -1669,4 +1669,22 @@ def query_and_store_records_to_delete(self,object_name,**kwargs):
...
+        and record field contains specified value. Example usage:
+        Verify Table Contains Row | Batches | MyBatch | Batch Number=MyBatch-01
+        """
+        for key,value in kwargs.items():
+            locator=npsp_lex_locators['datatable'].format(table_name,record,key)
+            actual=self.selenium.get_text(locator)
+            print(f'actual value is {actual}')
+            if actual==value:
+                print(f'Table contains {record} with expected {key}={value}')
+            elif value=='None' and actual=='':
```
rjanjanam, line 1684: Curious what would be the scenario where value='None' and actual=' '  ?
for empty string validation you can use if not <my_string>

deepmalamehta (approved):
Looks good !!

robot/Cumulus/resources/NPSPSettingsPageObject.py
```diff
@@ -247,4 +247,25 @@ def verify_status_to_state_mappings(self, **kwargs):
...
+        And verifies that the format has been deactivated"""
+        locator=npsp_lex_locators['npsp_settings']['table_dropdown'].format(number_format)
+        self.selenium.get_webelement(locator).click()
+        if status.lower()=='deactivate':
+            self.selenium.wait_until_page_contains("Deactivate")
+            self.selenium.click_link("Deactivate")
+            self.builtin.sleep(1,"To allow time for format to be deactivated")
+            self.npsp.verify_table_contains_row("datatable",number_format,Active="False")
+        elif status.lower()=='activate':
```
vmartinov-sforg, line 263: Is the syntax 'elif', or should it be 'else if' ?

kenlewis-sfdo (approved):
From my perspective, this looks great.  Thanks, Sravani!  I was able to run the test and see it complete successfully, and the scenario under test looks good to me.  I see that Ratna has a few comments but I don't have any changes to request.  Thanks again!

robot/Cumulus/resources/NPSP.py
vmartinov-sforg, line 1684: Is the syntax 'elif', or should it be 'else if' ?

rhakeshj (approved):

vmartinov-sforg (commented):
Looks good, I have a couple of syntax related questions.

robot/Cumulus/resources/NPSP.py
skristem: In the test case when I am passing key value pairs I have to pass like Batch Number=abc1 but if I want to validate that field doesn't contain a value I am using Batch Number=None then when I see the expected value as None the expectation is to see the actual value as empty string. I should see empty string only when I pass None as value

robot/Cumulus/resources/NPSPSettingsPageObject.py
skristem: in python we use 'elif' for 'else if'

robot/Cumulus/resources/NPSP.py
skristem: it works as long as it is a lightning table and has qalocator. I am using this in the Gift Entry and NPSP settings page. Since we are moving towards adding identifiers, I have added this here

rjanjanam (approved):

robot/Cumulus/resources/NPSPSettingsPageObject.py
skristem: thanks for catching, I modified after creating it and forgot to update the documentation

robot/Cumulus/resources/NPSP.py
skristem: python uses elif for else if

vmartinov-sforg (approved):

schlossmj (approved):

PR #6088, opened 2020-12-03
[GETTING STARTED] Fixing failing test

dcaballeroSf:
# Critical Changes

# Changes
Fixing failing testcase in some types of orgs
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/GS_ApplicationStatusController_TEST.cls
+++ b/src/classes/GS_ApplicationStatusController_TEST.cls
@@ -44,7 +44,7 @@ public with sharing class GS_ApplicationStatusController_TEST {
         // This test relies on mocked data which is only being mocked in Scratchs org.
         // As these real data depends on the org which the test is running, the test is fragile in any other org type
         // avoiding the test.
-        if (!UTIL_Namespace.isScratchOrg() && String.isBlank(UTIL_Namespace.getNamespace())) {
+        if (!UTIL_Namespace.isScratchOrg() || String.isBlank(UTIL_Namespace.getNamespace())) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/classes/GS_ApplicationStatusController_TEST.cls
```diff
@@ -41,6 +41,13 @@ public with sharing class GS_ApplicationStatusController_TEST {
     @IsTest 
     private static void testGetApplicationStatusTrialOrgNonApplicated() {
         
+        // This test relies on mocked data which is only being mocked in Scratchs org.
```
force2b, line 44: Is this comment fully accurate? Mocked data would never be dependent on an org because it's mocked (i.e., completely fake). That can't be the issue here though. What is the root cause that makes this test fail in the packaging org and not in a scratch org?

force2b (changes requested):

src/classes/GS_ApplicationStatusController_TEST.cls
dcaballeroSf: In the controller code, I am only allowing to inject the mocked data if the org is of type: "Scratch", I dont want final customers messing with a testing solution we found to give testeability to a component which depends on many external conditions. 

The each method is assuming that some status of the org(p10 applied, expiration date, etc) to check if the returning values from the controller is correct. As these conditions probably arent met in running org, the test fails.

force2b: Thanks for clarifying. How about changing this comment to:
```
// This test relies on mocked org information data in the Controller class. That class will only apply mocked data
// when the current org is a Sandbox org (by checking the Organization object) and there is no package 
// namespace. As a result, this test has to be skipped for other types of orgs to avoid an error.
```

force2b (changes requested):
Thanks for clarifying. I wrote up a new comment block to use.

force2b (approved):

force2b (approved):

PR #6089, opened 2020-12-03
W-8530908 - Mapping new fields to Payment for ACH and other tweaks

callen-sfdo:
- Created three fields on the Payment Object to receive status, declined reason, and ACH code.
- Created three new field mappings to populated those fields from the DataImport object
- Modified two field labels on the DataImport object to better reflect their content.
- Cleaned up the order of fields and some white space in the Data import object file to match the SFDC version.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/objects/npe01__OppPayment__c.object
```diff
@@ -196,6 +210,30 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Elevate_Payment_API_Declined_Reason__c</fullName>
+        <description>Provides detail about why a transaction was declined.</description>
```
Julian88Tex, line 215: @callen-sfdo Is this a decline reason for the API or for the individual transaction? If so wondering why it's called "API Declined Reason".

src/customMetadata/Data_Import_Field_Mapping.Payment_Elevate_API_Declined_Reason.md
```diff
@@ -0,0 +1,29 @@
...
+        <field>Required__c</field>
+        <value xsi:type="xsd:string">No</value>
+    </values>
+    <values>
+        <field>Source_Field_API_Name__c</field>
+        <value xsi:type="xsd:string">npsp__Payment_Declined_Reason__c</value>
+    </values>
+    <values>
+        <field>Target_Field_API_Name__c</field>
+        <value xsi:type="xsd:string">npsp__ Elevate_Payment_API_Declined_Reason__c</value>
```
Julian88Tex, line 23:
```suggestion
        <value xsi:type="xsd:string">npsp__Elevate_Payment_API_Declined_Reason__c</value>
```

Julian88Tex: @callen-sfdo looks like there's an extra space here

src/objects/DataImport__c.object
```diff
@@ -1441,7 +1441,7 @@
         <fullName>Payment_Declined_Reason__c</fullName>
```
sam-knox, line 1441: @callen-sfdo - Shouldn't the API name get updated the match the label too?

src/objects/npe01__OppPayment__c.object
sam-knox: @Julian88Tex - It's for the individual transaction. We're using "API" in the label to make it clear that it's the value directly from the API (the Elevate backend) not what users see in the front end. At some point later, we may have another field for the declined reason as it's rendered in the UI. 

These API fields are meant to be used for troubleshooting because they're the only way to reconcile the state of records between the two systems in an unambiguous way.

sam-knox (approved):
Labels look fine, but it would be great if we can update:

Payment_Declined_Reason__c
to
Payment_Elevate_API_Declined_Reason__c

force2b:
@sam-knox If that field has already been included in release, the API cannot be changed.

grivero (approved):
Looks good

src/objects/DataImport__c.object
callen-sfdo: @sam-knox Yeah as mike mentioned, we generally do not change API names unless its something that has not been released yet.  Basically changing the API name is the same as creating a new field and deprecating an old one, and any referencing code would need to be moved over.  It wouldn't necessarily be a big deal with just our code, but clients may be relying on it at this point.

src/customMetadata/Data_Import_Field_Mapping.Payment_Elevate_API_Declined_Reason.md
callen-sfdo: @Julian88Tex yep, thats a typo, thanks for the catch

Julian88Tex (approved):
Verified on my end.

PR #6108, opened 2020-12-09
W-8342739 - Elevate ACH (NAACP MVP) - BDI changes to handle payment edits for ACH status changes

callen-sfdo:
Code work to support Elevate ACH payment processing in BDI.  These changes include:
- If Payment Elevate ID is detected, normal Opportunity/Payment matching is skipped entirely and it will be assumed that any Id supplied in the DonationImported__c or PaymentImported__c are the Ids to update.
- If Payment Elevate Id is detected, then auto-payment creation by triggers on the opportunity will be supressed and Payments will instead be created by the BDI code (using an existing method in PMT_PaymentCreator)
-  If the Payment type is ACH then it will bypass the Payment status filter that currently only allows blank or 'CAPTURED' payment status 
-  The DonationImportStatus__c will now be set to "Updated" if the opportunity was successfully updated during BDI Processing.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_PaymentServices.cls
```diff
@@ -45,7 +45,8 @@ public with sharing class GE_PaymentServices {
         DECLINED,
         NONRETRYABLEERROR,
         RETRYABLEERROR,
-        REFUNDISSUED
+        REFUNDISSUED,
+        SUBMITTED
```
KietTruongSFDO, line 49: There is also a SETTLED status isn't there?

src/classes/BDI_Donations.cls
```diff
@@ -711,5 +773,136 @@ public class BDI_Donations {
...
+        }
+        return nonElevateDIs;
+    }
+
+    /*******************************************************************************************************
+    * @description Populates new and updates existing payments as needed for elevate related transactions.
+    * @param dataImportsById map of Data import records by their ids. 
+    */
+    private void processPaymentsForElevateDIs (DataImport__c[] dataImports) {
+        Map<Id, npe01__OppPayment__c> paymentsForUpsertByDiId = new Map<Id, npe01__OppPayment__c>();
```
KietTruongSFDO, line 808: This `paymentsForUpsertByDiID` is not being used  right?

src/classes/BDI_Donations.cls
```diff
@@ -143,13 +143,16 @@ public class BDI_Donations {
...
         List<DataImport__c> dataImportDonations = validateDonationsToProcess(matchBehavior);
 
+        List<DataImport__c> dataImportForMatching = filterOutElevatePaymentDIs(dataImportDonations);
+
         // call the extension to match against existing donations
-        if (matchBehavior != BDI_DataImport_API.DoNotMatch) {
+        if (matchBehavior != BDI_DataImport_API.DoNotMatch && dataImportForMatching.size() > 0) {
             BDI_PerfLogger.PerfLog logger = BDI_DataImportService.perfLogger.newPerfLog('iMatchDonations.match');
+            System.debug('Calling Matching logic for Donations');
             iMatchDonations.match(bdi, dataImportDonations, opportunitiesByDataImportId, paymentsByDataImportId);
```
KietTruongSFDO, line 156: iMatchDonations.match is taking in `dataImportDonations`, is it supposed to take in `dataImportForMatching` instead?

src/classes/BDI_Donations.cls
```diff
@@ -173,7 +176,8 @@ public class BDI_Donations {
                     dataImport.PaymentImportStatus__c = Label.bdiCreated;
                 } else {
                     // validate against our desired matching behavior (exact match, best match, new, etc.)
-                    if (!isValidMatchBehavior(dataImport, matchBehavior)) {
+                    // Elevate payments are specifically excluded
```
KietTruongSFDO, line 179: Is the comment correct or is the logic incorrect? it looks like all non-elevate payments are excluded, not the other way as the comment suggests

src/classes/BDI_Donations.cls
```diff
@@ -727,8 +719,7 @@ public class BDI_Donations {
             }
             if (dataImport.PaymentImported__c != null
                     || dataImport.DonationImportStatus__c == Label.bdiMatchedApplyNewPayment
-                    || (isElevatePaymentDI(dataImport) && 
-                        dataImport.PaymentImportStatus__c == null)) {
+                    || isElevatePaymentDI(dataImport)) {
```
KietTruongSFDO, line 722: isElevatePaymentDI doesn't need to be at the beginning of the if-statement?

callen-sfdo: @KietTruongSFDO Just simplified this since all Elevate Payments should be fine entering.

src/classes/BDI_Donations.cls
callen-sfdo: @KietTruongSFDO fixed this.

src/classes/BDI_Donations.cls
callen-sfdo: @KietTruongSFDO Yeah it was a leftover, removed it.

src/classes/GE_PaymentServices.cls
callen-sfdo: @KietTruongSFDO I'm actually not going to add anything for these now since I don't need it for the logic.

src/classes/BDI_Donations.cls
callen-sfdo: @KietTruongSFDO I revised the wording to be more clear.

src/classes/BDI_Donations.cls
```diff
@@ -711,5 +767,108 @@ public class BDI_Donations {
...
+        for (DataImport__c di : disWithOpptForQuery) {
+            opptIdsForQuery.add(di.DonationImported__c);
+        }
+
+        Set<String> fieldNamesForQuery = new Set<String>{'Id','StageName','CloseDate','Amount','IsClosed'};
+
+        // Add all the payment mappings to the list of fields to be queried since they will
+        // need to be copied to the payment.
+        for (npe01__Payment_Field_Mapping_Settings__c pfms : PMT_PaymentCreator.paymentMappings.values()) {
+            fieldNamesForQuery.add(pfms.npe01__Opportunity_Field__c);
```
KietTruongSFDO, line 828: we should probably .toLowerCase() these fields and the initial set of fields in "fieldNamesForQuery" to make it case insensitive when adding into the set.

KietTruongSFDO (approved):
One minor change, but approved.

src/classes/BDI_Donations.cls
callen-sfdo: Yeah, probably a good call, made the change.

PR #6122, opened 2020-12-15
Build UI to Select Gateway

dfuller114:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
```diff
@@ -0,0 +1,82 @@
...
+            </lightning-layout-item>
+        </template>
+        <lightning-layout-item size="4">
+            <template if:true={isReadState}>
+                <lightning-input variant="label-inline" read-only="true" label="Gateway ID" value={gatewayId}
+                                 type="text"></lightning-input>
+            </template>
+            <template if:true={isEditState}>
+                <lightning-input variant="label-inline" label="Gateway ID" value={gatewayId}
+                                 type="text" data-id="gatewayIdEditField"></lightning-input>
```
kenlewis-sfdo, line 74: Is it possible to use only one `lightning-input` in the template here and control the read-only property using the `isReadState` field? (and maybe rename the field to "isReadOnly")

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
```diff
@@ -0,0 +1,82 @@
...
+                recently. To override this, click Edit and enter the gateway ID for the one you'd like to use instead.
+                Find the gateway ID in the Elevate Admin Console | Gateway Management.</p>
+        </lightning-layout-item>
+        <template if:true={isReadState}>
+            <lightning-layout-item size="12" class="slds-text-align_center">
+                <lightning-button label="Edit" data-id="editGatewayButton" onclick={handleEdit}></lightning-button>
+            </lightning-layout-item>
+        </template>
+        <lightning-layout-item size="4">
+            <template if:true={isReadState}>
```
kenlewis-sfdo, line 68: Can these two templates that use the same if conditional (`template if:true={isReadState}`) be combined into a single template?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    _isEditState;
+    get isEditState() {
+        return this._isEditState
+    }
+    set isEditState(value) {
+        this._isEditState = value;
+        this._isReadState = false;
+    }
```
kenlewis-sfdo, line 63: Is it necessary to have both of these properties (isReadState, isEditState)?  Would it be possible to merge these into a single "isReadOnly" property?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,120 @@
...
+
+    async handleSave(event) {
+        try {
+            let gatewayId = this.template.querySelector("[data-id='gatewayIdEditField']").value;
+            await setGatewayId({ gatewayId: gatewayId});
+
+            this.isSuccess = true;
+        } catch(ex) {
+            console.log(ex);
```
kenlewis-sfdo, line 108: Should this console.log be removed?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,120 @@
...
+            this.isError = true;
+        }
+    }
+
+    async getGatewayId() {
+        try {
+            this.gatewayId = await getGatewayIdFromConfig();
+        } catch(ex) {
+            // handleError(ex);
```
kenlewis-sfdo, line 117: Is there an expected exception to be caught here?  If not, this try/catch could probably be removed.

kenlewis-sfdo (commented):
Looking good, Daniel!  I left a few initial comments - let me know if you have any questions!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,120 @@
...
+    get isError() {
+        return this._isError;
+    }
+    set isError(value) {
+        this._isError = value;
+
+        if (value) { this.isSuccess = false; }
+    }
+
+    _errorMessage;
```
bdvorachek, line 85: Some of these simpler private variables may not need getters/setters.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,120 @@
...
+import { LightningElement } from 'lwc';
+import { handleError } from 'c/utilTemplateBuilder';
+import messageLoading from '@salesforce/label/c.labelMessageLoading';
+import setGatewayId from '@salesforce/apex/PS_GatewayManagement.setGatewayId';
+import getGatewayIdFromConfig from '@salesforce/apex/PS_GatewayManagement.getGatewayIdFromConfig';
+
+export default class GePaymentGatewayManagement extends LightningElement {
+
+    showSpinner = false;
+    gatewayId;
```
psadie-sfdo, line 39: Should we prefix these with an underscore?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js-meta.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
psadie-sfdo, line 3: Bump to the latest apiVersion 50

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
```diff
@@ -300,6 +300,12 @@ const dispatch = (context, name, detail, bubbles = false, composed = false) => {
 * @param {object} error: Event holding error details
 */
 const handleError = (error) => {
```
psadie-sfdo, line 302: More of a note for the future, but I'm not a huge fan of this particular `handleError` being imported outside of template builder use-cases. Ideally, we'd either move `handleError` out of `utilTemplateBuilder` and into `utilCommon` or start a new method entirely for general use.

src/pages/STG_PanelGatewayManagement.page-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexPage xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
psadie-sfdo, line 3: Bump to the latest apiVersion 50

src/pages/STG_SettingsManager.page
```diff
@@ -267,6 +267,16 @@
...
+                                <li class="slds-tree__item"
+                                    role="treeitem"
+                                    aria-level="2"
+                                    id="idPanelGatewayManagementNav">
+
+                                    <a href="#"
+                                       onclick="ShowPanel('idPanelGatewayManagement');return false;"
+                                       class="slds-grow slds-m-left_large"> Elevate Gateway Management
+                                    </a>
+                                </li>
```
psadie-sfdo, line 279: Personally not a huge fan of the formatting on this page, but what do you think about adhering to the established one-line?

psadie-sfdo (changes requested):
Just a few things!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
dfuller114: @kenlewis-sfdo good idea. Just pushed an update for this!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
dfuller114: @kenlewis-sfdo Pushed an update to use a single isReadOnly getter so now we're using if:true to render the edit button and if:false to render the Save/Cancel button group.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @kenlewis-sfdo Just pushed an update that removed the isEditState() getter all together

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @kenlewis-sfdo Haven't completed the error handling and validations yet so this will be filled in once that's done

src/classes/PS_GatewayManagement.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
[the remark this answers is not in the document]
dfuller114: @bdvorachek thanks! updated!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js-meta.xml
dfuller114: done!

src/pages/STG_PanelGatewayManagement.page-meta.xml
dfuller114: done!

src/pages/STG_SettingsManager.page
dfuller114: @psadie-sfdo yeah I danced around what to do there. Didn't want to change the formatting of the rest of the page. I'll fix that to keep it consistent with the rest of the page. Just updated.

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
dfuller114: @psadie-sfdo I added a new method called `buildErrorMessage` into that component where we can get back the error message string without having to throw a toast but without disrupting all the other references to `handleError`, but I do agree the more we can pull utilities out of the `utilTemplateBuilder` the better but that would be a pretty high LOE at this point since we're importing this in a bunch of places.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @psadie-sfdo these variables are used in the UI template so they aren't private to the javascript so I think in those cases we wouldn't use the underscore

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @bdvorachek Agreed. Just pushed an update for that

src/lwc/utilTemplateBuilder/utilTemplateBuilder.js
psadie-sfdo: +++ on that update and completely agree that we shouldn't tackle pulling `handleError` out in this work.

Thanks!!

dfuller114:
At this point, the code work should be complete except for apex unit test coverage.

vmartinov-sforg (changes requested):
The code looks good - just A couple of minor comments. The Gateway Management Page does not load for me, I just get a blank page. I see an error in the console when the npsp settings load. 

 `Uncaught Error: $Lightning.use() already invoked with application: npsp:RD2_EnablementApp
    at Object.use (computing-fun-4943-dev-ed--npsp.visualforce.com/lightning/lightning.out.delegate.js?v=dDIdorNC3N22LalQ5i3slQ:133)
    at Object.use (lightning.out.js:59)
    at STG_SettingsManager:6316`

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
```diff
@@ -0,0 +1,89 @@
...
+            </lightning-spinner>
+            <p class='loading-text slds-text-heading_medium'>{loadingText}</p>
+        </div>
+        <lightning-layout-item size="12">
+            <h1 class="slds-text-heading_medium slds-p-top-medium">Elevate Gateway Management</h1>
+        </lightning-layout-item>
+        <lightning-layout-item size="12">
+            <p class="slds-text-body_regular slds-p-vertical_medium">NPSP uses the gateway that was added to Elevate most
+                recently. To override this, click Edit and enter the gateway ID for the one you'd like to use instead.
+                Find the gateway ID in the Elevate Admin Console | Gateway Management.</p>
```
psadie-sfdo, line 66: Reminder to leave inline comments about the hardcoded UI text.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
```diff
@@ -0,0 +1,89 @@
...
+            <lightning-input variant="label-inline"
+                             read-only={isReadOnly}
+                             label="Gateway ID"
+                             value={gatewayId}
+                             type="text"
+                             data-id="gatewayIdField">
+            </lightning-input>
+            <template if:false={isReadOnly}>
+                <lightning-button-group>
+                    <lightning-button label="Save" variant="brand" onclick={handleSave}></lightning-button>
```
voduyemi, line 83: Can we use custom labels for these button labels?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,130 @@
...
+
+    _isError;
+    get isError() {
+        return this._isError;
+    }
+    set isError(value) {
+        this._isError = value;
+
+        if (value) { this.isSuccess = false; }
+    }
```
kenlewis-sfdo, line 82: Could isError be a simple getter that analyzes the errorMessage property instead of storing both?  

```javascript
get isError() {
        return !!this.errorMessage;
    }
```

kenlewis-sfdo (commented):
This is looking really good, Daniel.  I left one comment re: whether the _isError prop is needed, and I had one question re: the second A/C on the WI.  Should customers be prevented from saving a new Gateway Id when they do not already have one configured (i.e., they are not yet an Elevate customer)?

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    async getGatewayId() {
+        try {
+            this.gatewayId = await getGatewayIdFromConfig();
+        } catch(ex) {
+            this.errorMessage = buildErrorMessage(ex);
+            this.isError = true;
+        }
```
kenlewis-sfdo, line 123: If there is not a specific exception here that is expected to occur then I think this try catch block should be removed.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
dfuller114: I didn't see any custom labels we could use that had common enough names and since we aren't using any new labels for the most part I just hard coded

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
dfuller114: done!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: Good call. done!

src/classes/STG_PanelGatewayManagement_CTRL_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
psadie-sfdo, line 3:
```suggestion
    <apiVersion>50.0</apiVersion>
```

vmartinov-sforg, line 3:
```suggestion
    <apiVersion>50.0</apiVersion>
```

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @kenlewis-sfdo I thought of going that route but it would've required a little more robust solution both in this component and the alert banner component that I felt wasn't necessary with this being a simpler temporary feature in general that will only ever have these two alert banner states.

dfuller114: > This is looking really good, Daniel. I left one comment re: whether the _isError prop is needed, and I had one question re: the second A/C on the WI. Should customers be prevented from saving a new Gateway Id when they do not already have one configured (i.e., they are not yet an Elevate customer)?

Since we are keeping this feature simple overall we are just letting the admin override the current gateway id with whatever they enter into the field. We aren't doing any additional checks or validations other than not allowing them to save a blank value.

src/classes/STG_PanelGatewayManagement_CTRL_TEST.cls-meta.xml
dfuller114: done!

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,177 @@
...
+            this.isElevateCustomer = await checkForElevateCustomer();
+
+            this.hasAccess = !!(this.isElevateCustomer && this.isSystemAdmin);
+        } catch(ex) {
+            this.errorMessage = buildErrorMessage(ex);
+            this.isError = true;
+        }
+
+        if (this.hasAccess) {
+            this.getGatewayId();
```
voduyemi, line 69: I think we're missing the `async` keyword here and also not handling any error scenarios with this call. Do you also think we could call `await getGatewayIdFromConfig()` directly here instead since we're not using `getGatewayIdFromConfig` anywhere else. Also I would love to see us combine all three server calls in the `connectedCallback` into one instead of making three server trips. What are your thoughts?

src/lwc/utilAlertBanner/utilAlertBanner.html
```diff
@@ -1,13 +1,31 @@
...
-            <div class={alertTheme} role="alert">
-                <span class="slds-assistive-text">{alertAssistText}</span>
-                <span class="slds-icon_container slds-icon-utility-error slds-m-right_x-small">
-                    <lightning-icon icon-name={alertIcon} variant={alertVariant} size={alertSize}></lightning-icon>
-                </span>
-                <h2>{alertMessage}</h2>
-            </div>
-        </lightning-layout-item>
-    </lightning-layout>
+    <template if:true={isVisible}>
```
voduyemi, line 2: This is just a UI discrepancy I noticed, in other sections of NPSP Settings, we display UI messages using this component: 
image

We seem to be using a different component for this WI: 
image

Maybe we should align with the existing UI ? What are your thoughts?

voduyemi (changes requested):
Thanks Daniel! I've got a few comments around the server trips in the gateway management LWC and how we are displaying toasts.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
dfuller114: @voduyemi Regarding the error scenarios, since there aren't any specific error scenarios we would anticipate here, @kenlewis-sfdo mentioned we should probably just remove the try/catch which I agree with since we want to avoid any more generic error handling. There's really no way we could get a failure here just retrieving the gatewayId. Also, since the `async` keyword is on the connected callback function, this function isn't one we have to set an `await` on since there's no other logic after it. It will just populate the field on the component load when the call finishes

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.html
```diff
@@ -0,0 +1,119 @@
...
+                        <lightning-button
+                                label="Save"
+                                variant="brand"
+                                onclick={handleSave}
+                                data-qa-locator="button gateway management save">
+                        </lightning-button>
+                        <!-- Temporary Hardcoded Text until this functionality is moved out of NPSP entirely -->
+                        <lightning-button label="Cancel"
+                                          onclick={handleCancel}
+                                          data-qa-locator="button gateway management cancel">
```
bdvorachek, line 112: :+1: I think this is ok for QA locators that don't have any dynamic properties going forward.

src/classes/PS_GatewayManagement_TEST.cls
```diff
@@ -0,0 +1,129 @@
...
+    static void shouldRetrieveCorrectGatewayId() {
+        Test.startTest();
+
+        PS_GatewayManagement.setGatewayId(TEST_GATEWAY_ID);
+
+        Test.stopTest();
+
+        System.assertEquals(TEST_GATEWAY_ID, PS_GatewayManagement.getGatewayIdFromConfig(), 'When this method is ' +
+                'invoked, the gateway Id stored in CRM should be the gateway returned.');
+    }
```
kenlewis-sfdo, line 90: It looks like `PS_GatewayManagement.getGatewayIdFromConfig()` calls `PS_IntegrationServiceConfig.Service().getGatewayIds()` within its method body.  If that is the case, could one of these tests be removed?

dfuller114: @kenlewis-sfdo I think at this point we'll avoid any further code changes unless we feel they are necessary.

src/lwc/utilAlertBanner/utilAlertBanner.html
dfuller114: Discussed in our call today. We are going to leave the style difference as is since we are using LWC here and the top alert is using a Visualforce component.

bdvorachek (approved):

vmartinov-sforg (approved):
Looks Good - thank you Daniel, and everyone else who reviewed / tested this item.

src/lwc/utilAlertBanner/utilAlertBanner.js
```diff
@@ -15,6 +15,8 @@ export default class AlertBanner extends LightningElement {
     @api icon;
     @api size;
     @api message;
+    @api isDismissable = false;
+    @api isVisible = false;
```
RadSFDeveloper, line 19: @dfuller114 , when there is an error for the credit card info on the RD entry form, the error on the alert banner is not displayed in the Elevate widget (above the card holder name field). It is due to `isVisible` being false by default. Can you please check into when you have time? 
Screen Shot 2020-12-17 at 11 07 15 AM

src/pages/STG_SettingsManager.page
```diff
@@ -267,6 +267,7 @@
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelErrorNotifNav"><a href="#" onclick="ShowPanel('idPanelErrorNotif');return false;" class="slds-grow slds-m-left_large">{!$Label.stgNavErrorNotify}</a></li>
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelTDTMNav"><a href="#" onclick="ShowPanel('idPanelTDTM');return false;" class="slds-grow slds-m-left_large">{!$Label.stgNavTriggerConfig}</a></li>
                                 <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelDataImportAdvancedMappingNav"><a href="#" onclick="ShowPanel('idPanelDataImportAdvancedMapping');return false;" class="slds-grow slds-m-left_large">{!$Label.bdiAdvancedMapping}</a></li>
+                                <li class="slds-tree__item" role="treeitem" aria-level="2" id="idPanelGatewayManagementNav" data-qa-locator="elevate gateway management setting"><a href="#" onclick="ShowPanel('idPanelGatewayManagement');return false;" class="slds-grow slds-m-left_large"> Elevate Gateway Management</a></li>
```
RadSFDeveloper, line 270: The subpage name should be a custom label.

src/lwc/gePaymentGatewayManagement/gePaymentGatewayManagement.js
```diff
@@ -0,0 +1,172 @@
...
+
+        let gatewayIdField = this.template.querySelector("[data-id='gatewayIdField']");
+
+        if (isEmpty(gatewayIdField.value)) {
+
+            gatewayIdField.setCustomValidity('Invalid Gateway ID.');
+            gatewayIdField.reportValidity();
+
+
+            this.errorMessage = 'Enter a valid gateway ID.'
```
RadSFDeveloper, line 151: Should the error message be a custom label?

vmartinov-sforg: @RadSFDeveloper We decided to not create custom labels as this work is temporary and will be replaced by Gateway Routing Rule sets.

src/lwc/utilAlertBanner/utilAlertBanner.js
RadSFDeveloper: Thanks for the change @dfuller114 , the alert banner is displayed on the RD entry form.
Screen Shot 2020-12-17 at 11 27 59 AM

psadie-sfdo (approved):
This is looking good to me! Just one piece of non-blocking feedback below.

voduyemi (approved):
Looks good!

PR #6158, opened 2021-01-07
Added missing credit card constant

voduyemi:
WI: [W-8615890](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008mAAYIA2/view)

Added a new constant to identify credit card transaction responses
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormRenderer/geFormRenderer.js
+++ b/src/lwc/geFormRenderer/geFormRenderer.js
@@ -122,6 +122,7 @@ const DONATION_DONOR_TYPE_ENUM = Object.freeze({
 });
 const CREDIT_CARD_WIDGET_NAME = 'geFormWidgetTokenizeCard';
 const ACH_CONSENT_MESSAGE = 'true';
+const PAYMENT_METHOD_CREDIT_CARD = 'Credit Card';
 
 export default class GeFormRenderer extends LightningElement{
     // these three fields are used to query the donor record
@@ -2268,7 +2269,7 @@ export default class GeFormRenderer extends LightningElement{
     }
 
     isCreditCardTransaction() {
-        return this.selectedPaymentMethod() === PAYMENT_METHODS.CREDIT_CARD;
+        return this.selectedPaymentMethod() === PAYMENT_METHOD_CREDIT_CARD;
     }
 
     processDataImport = async () => {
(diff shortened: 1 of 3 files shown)


src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -122,6 +122,7 @@ const DONATION_DONOR_TYPE_ENUM = Object.freeze({
 });
 const CREDIT_CARD_WIDGET_NAME = 'geFormWidgetTokenizeCard';
 const ACH_CONSENT_MESSAGE = 'true';
+const PAYMENT_METHOD_CREDIT_CARD = 'Credit Card';
```
ptewson-sfdo, line 125: This constant is also defined in src/lwc/rd2EntryForm/rd2EntryForm.js and src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js (and also within the script in GE_TokenizeCard.page)... assuming these are equivalent defined values (corresponding to the picklist value in the Payment Method picklist, right?) is there some way to reduce the number of duplicated constants? 

Also assuming this is tied to the picklist value, is there a risk of that value getting translated?

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -2,7 +2,7 @@ import { LightningElement, api, track, wire } from 'lwc';
 import CURRENCY from '@salesforce/i18n/currency';
 import { registerListener } from 'c/pubsubNoPageRef';
 import { isNull, showToast, constructErrorMessage, format } from 'c/utilCommon';
-import { HTTP_CODES } from 'c/geConstants';
```
voduyemi, line 5: @npsp-reedestockton I made this change to move this common constant `PAYMENT_METHOD_CREDIT_CARD` to `geConstants` since both teams use it, what do you think?

ptewson-sfdo: For my 2 cents, I think this makes sense as long as they are both referencing the same value - which for Gift Entry, is the value of the picklist option on the Payment npe01__Payment_Method__c field (I think). However, if they are pointing to different fields it might be nice to use different constants just in case the values  drift in the future.

psadie-sfdo (commented):
Changes look good to me including the changes to rd2EntryForm.js, but withholding approval until we get feedback from one of NGOhana folks.

ptewson-sfdo (approved):
Looks good, pending any feedback on those RD2 changes!

ptewson-sfdo (approved):
Looks good!

PR #6186, opened 2021-01-16
PoC for using Allocations service for GAU Allocations for RD in BDI

callen-sfdo:
A proof of concept for implementing GAU Allocations for RD In BDI by disabling all allocation triggers before BDI Processing, then using the existing additional objects code for inserting GAU Allocations, then running ALLO_AllocationsService manually and re-enabling allocations triggers.

If an error is encountered during allocations processing then the allocations are manually deleted and the error is set in one of the import status fields.  The downside of this approach is that any client triggers that rely on allocations would presumably fire even on the insert of an invalid allocation (such as percent > 100%) since it wouldn't be caught before insert.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


coriobriensfdo:
> Thanks for the contribution! It looks like @e-ferrando @fparodi-salesforce is an internal user so signing the CLA is not required. However, [we need to confirm this](https://docs.google.com/forms/d/e/1FAIpQLSfB_vIrwFLweXIGt7ID98JhISCU2zm2ErOugAnGAcCvAzr7NQ/viewform).

Ignore. The user is no longer working with Salesforce and therefore no longer has perms to this org. Manually removed the cla:missing label as this is not considered an external contribution.

PR #6142, opened 2020-12-23
POC: Edit Elevate Commitment

RadSFDeveloper:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

PR #6193, opened 2021-01-20
[RD] Initiate Elevate commitment before Recurring Donation creation 

RadSFDeveloper:
The change in order record creation order has been applied to ensure commitment is created in Elevate before it is created in NPSP so Elevate is the record source of truth.

# Critical Changes

 - For Elevate customers: enable at least read permissions for these Recurring Donation fields on your user profiles:
   - Elevate Recurring Id
   - Card Last 4
   - Card Expiration Month
   - Card Expiration Year   

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Apex class:
- RD2_DataRegulationService

# Deleted Metadata


RadSFDeveloper:
Tracking [W-8554656](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008l8I5IAI/view)

src/classes/RD2_EntryFormController_TEST.cls
```diff
@@ -375,28 +266,19 @@ private with sharing class RD2_EntryFormController_TEST {
...
-        String jsonResponse = RD2_EntryFormController.createCommitment(rd.Id, PAYMENT_METHOD_TOKEN);
+        String jsonResponse = RD2_EntryFormController.handleCommitment(JSON.serialize(rd), PAYMENT_METHOD_TOKEN);
         Test.stopTest();
 
         List<Error__c> errors = errorGateway.getRecords();
         System.assertEquals(1, errors.size(),
             'An error record should be created for an error response.');
-        System.assertEquals(rd.Id, errors[0].Related_Record_Id__c,
-            'The error record should contain RD record Id.');
+        System.assertEquals(rd.npe03__Contact__c, errors[0].Related_Record_Id__c,
```
RadSFDeveloper, line 280: The change has been made that a donor Id is set as the record Id in the error log when commitment is created but Recurring Donation creation failed.

src/labels/CustomLabels.labels
```diff
@@ -2221,15 +2221,20 @@
...
+        <value>Saving Elevate recurring payment...This may take a few minutes.</value>
     </labels>
     <labels>
         <fullName>RD2_EntryFormSaveCommitmentFailedMessage</fullName>
         <categories>Recurring-Donations, Entry-Form</categories>
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Message displayed as prefix to errors returned from Elevate API</shortDescription>
-        <value>However, an Elevate recurring payment could not be created due to the following errors:</value>
+        <value>A recurring payment has been created in Elevate. Elevate Recurring Id: {0}.
```
RadSFDeveloper, line 2232: A new line is added in the custom label and `lightning-formatted-text` used to display new line and bulletin point. The `<br/>` and other html tags were not used, but could be with  `lightning-formatted-rich-text` since the error log in the email and in NPSP Settings > Error Log would contain tags and the message would not be user friendly.

src/classes/RD2_ElevateIntegrationService.cls
```diff
@@ -145,11 +137,9 @@ public inherited sharing class RD2_ElevateIntegrationService {
 
         if (!isChanged) {
             String oldCommitmentId = oldRd.CommitmentId__c;
-            Boolean isTempIdOverwrite = String.isNotBlank(oldCommitmentId)
-                && oldCommitmentId.startsWith(TEMP_COMMITMENT_ID_PREFIX);
 
-            isChanged = rd.CommitmentId__c != oldCommitmentId
-                && !isTempIdOverwrite;
+            isChanged = String.isNotBlank(oldCommitmentId)
```
RadSFDeveloper, line 141: @rponti-sforg , please ensure non-integration user can create an Elevate commitment.

src/classes/RD2_EntryFormController.cls
```diff
@@ -354,24 +304,83 @@ public with sharing class RD2_EntryFormController {
...
+        /**
+        * @description Handles validation and Elevate recurring commitment creation
+        * @param jsonRecord Recurring Donation fields as entered by the user before the record is saved
         * @param paymentMethodToken Payment Method Token
-        * @return PS_CommitmentRequest.RequestBody
+        * @return UTIL_Http.Response Payments API response
         */
-        public PS_CommitmentRequest.RequestBody getRequestBody(Id recordId, String paymentMethodToken) {
-            PS_CommitmentRequest request = new PS_CommitmentRequest();
+        public UTIL_Http.Response handleCommitment(String jsonRecord, String paymentMethodToken) {
```
RadSFDeveloper, line 313: The code added in this WI is very similar to Edit Commitment POC #6142, but it did diverge a bit. Thus, when edit commitment feature is being implemented, please do not copy code over `rd2EntryForm`, and `RD2_EntryFormController` but feel free to add pieces here and there understanding the copied and changed code.

RadSFDeveloper (commented):
Code is ready for review.

src/labels/CustomLabels.labels
RadSFDeveloper, line 2232: @sam-knox , can you please review the label as well as PR description? All user profiles regardless if they use Elevate need to have read-only access on specified fields (this is due to using the fields in the form). Please let me know if there are questions.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -178,16 +190,22 @@ public inherited sharing class PS_CommitmentRequest {
     */
     @TestVisible
     private void setDonorInfo(RequestBody reqBody, npe03__Recurring_Donation__c rd) {
-        Boolean isContactDonor = rd.npe03__Contact__c != null
+        Contact contactDonor = rd.npe03__Contact__c != null
+            ? getContact(rd.npe03__Contact__c)
+            : null;
+
+        Boolean isContactDonor = contactDonor != null
```
andrewyu-salesforce, line 197: I find the `isContactDonor` and the `contactDonor` variables confusing. I think changing `isContactDonor` to `isHouseholdDonor` will be clearer.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -286,6 +260,45 @@ public inherited sharing class PS_CommitmentRequest {
...
     }
 
+    /***
+    * @description Retrieves Contact data specified as the donor on the Recurring Donation record
+    * that is not yet saved into database
+    * @param contactId Contact Id
+    * @return Contact
+    */
+    private Contact getContact(Id contactId) {
+        List<Contact> contacts = new List<Contact>([
```
andrewyu-salesforce, line 270: `new List<Contact>` part is unnecessary because Apex automatically cast the query result into SObject List even if no record is found from the query.
Same with `getAccount()` method below

src/classes/PS_CommitmentRequest.cls
```diff
@@ -286,6 +260,45 @@ public inherited sharing class PS_CommitmentRequest {
...
+            SELECT
+                AccountId, FirstName, LastName
+            FROM Contact
+            WHERE Id = :contactId
+            LIMIT 1
+        ]);
+        
+        return contacts == null || contacts.isEmpty()
```
andrewyu-salesforce, line 278: Does not need the null check because the query will always return a List even if no record is found.
Same with `getAccount()` method below

src/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -260,92 +260,136 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
-            //map the Day of Month value of "31" to the Last_Day value
-            if (rd.Day_of_Month__c == '31') {
-                rd.Day_of_Month__c = RD2_Constants.DAY_OF_MONTH_LAST_DAY;
-            }
 
-            //set the default installment frequency when value is not provided on the RD insert
-            if (rd.InstallmentFrequency__c == null && triggerAction == TDTM_Runnable.Action.BeforeInsert) {
-                rd.InstallmentFrequency__c = RD2_Constants.DEFAULT_INSTALLMENT_FREQUENCY;
-            }
+    /***
```
force2b, line 268: Would would you think of moving this to a separate class (`RD2_DataRegulatorService`?)? It feels like this behavior isn't necessarily part of TDTM specifically, but instead a service used by TDTM and others.

src/labels/CustomLabels.labels
sam-knox: @RadSFDeveloper - Ok, will do. Are there plans to include these new fields in the unmanaged metadata so that at least the Sys Admin profile will get the correct FLS when upgrading to RD2?

Also, are these fields part of the Enhanced Recurring Donation page layout? Would a non-Elevate customer need to remove them?

andrewyu-salesforce (commented):
@RadSFDeveloper Thanks for the implementation. A small feedback on the `setDonorInfo()` method.

src/labels/CustomLabels.labels
```diff
@@ -2221,15 +2221,20 @@
...
+        
+        However, a matching Recurring Donation wasn&apos;t created in NPSP due to the following error:
+        ● {1}
+        
+        You can keep this form open, fix the error, and save the Recurring donation again. Alternatively, cancel and contact your System Administrator for help.</value>
```
sam-knox, line 2237:
```suggestion
        You can keep this form open, fix the error, and save the Recurring Donation again. Alternatively, cancel and contact your System Administrator for help.</value>
```

sam-knox (changes requested):
Just one minor change on line 2237. Capitalize the word "Donation" in Recurring Donation.

src/labels/CustomLabels.labels
RadSFDeveloper: I have updated Admin and Standard Profiles deployed in RD2 meta deploy to include the "Elevate Recurring Id" (`CommitmentId__c`). Other fields were already there.

The Elevate fields are not added to the RD page layout. The admin for the org connected to Elevate needs to add them manually (if required).

RadSFDeveloper: @sam-knox , to clarify, the specified fields visibility is required only for RD2 customers.

npsp-reedestockton (approved):
Great work @RadSFDeveloper

andrewyu-salesforce (approved):
@RadSFDeveloper Nice work!!

sam-knox (approved):

src/classes/RD2_DataRegulationService.cls
```diff
@@ -0,0 +1,164 @@
+/*
+    Copyright (c) 2020 Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2021 Salesforce.org
```

src/classes/RD2_DataRegulationService.cls
```diff
@@ -0,0 +1,164 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Recurring Donations
+* @description Adjusts Enhanced Recurring Donations records so they can be processed
+* in triggers, or data adjusted and validated outside of the trigger context.
+*
+*/
+public with sharing class RD2_DataRegulationService {
```
force2b, line 38: As a service class, I recommend using `inherited sharing` so that this use the perms of the calling class rather than enforcing it's own unique requirements on user perms.

force2b (changes requested):
Looking good. Just one minor change needed and suggestion for another.

force2b (approved):

src/labels/CustomLabels.labels
RadSFDeveloper: @sam-knox , the latest testing on a scratch org and sandbox indicate the fields need to be visible only for RD2 customers using Elevate. Can you please update the PR description? We will also check into sandboxes after the release to confirm.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -122,28 +122,61 @@ public inherited sharing class PS_CommitmentRequest {
...
     /***
-    * @description Creates an HttpRequest for a new commitment creation
+    * @description Creates an HttpRequest for a new commitment creation or
+    * existing commitment update depending on the prodived commitment Id
+    * @param commitmentId Elevate recurring commitment Id
     * @param jsonRequestBody JSON containing parameters for the purchase call request body
     * @return HttpRequest
     */
-    public HttpRequest buildCreateRequest(String jsonRequestBody) {
+    public HttpRequest buildRequest(String commitmentId, String jsonRequestBody) {
```
RadSFDeveloper, line 145: The latest change, primarily in the [commit](https://github.com/SalesforceFoundation/NPSP/pull/6193/commits/a3620f1cb257fe732f1813f6b0e108e76ab3e4cf), has been done to issue an edit request instead of bypassing the commitment API request when commitment is created, and the RD was changed  after erroring out first (see AC7).
This way we are building a path for the edit that will work with create instead of patching the code to work for AC7 but later on changing it when edit changes are introduced or when user changes the Amount or schedule after fixing an error for RD insert failure for which commitment has been created. Please let me know if there is any issue with this approach.
@rponti-sforg , the PR can be retested (AC7 and happy path ).

rponti-sforg: @RadSFDeveloper Re-testing.  Will update you with any questions or concerns.

PR #6196, opened 2021-01-21
Gift Entry: Add RecordType field, mappings to QA/Regression flows

bdvorachek:
Testing Notes: 
* Run `cci flow run qa_org` then `cci task run test_data_qa_org` against your target org.
* Enable Advanced Mapping & Gift Entry
* Add new fields to your desired gift entry template 
  * CustomObject1__c.C1_rt_picklist__c is a picklist which is controlled by the current Record Type
  * CustomObject1__c.RecordTypeId
  * And CustomObject1Imported so you can select one of the existing CustomObject1 records created by `test_data_qa_org`

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


datasets/qa_org/test_data.sql
```diff
@@ -0,0 +1,2204 @@
...
+CREATE TABLE "CustomObject1__c_rt_mapping" (
+    record_type_id VARCHAR(255) NOT NULL,
+    developer_name VARCHAR(255),
+    PRIMARY KEY (record_type_id)
+);
+INSERT INTO "CustomObject1__c_rt_mapping" VALUES('Flower', 'Flower');
+INSERT INTO "CustomObject1__c_rt_mapping" VALUES('Vegetable', 'Vegetable');
+INSERT INTO "CustomObject1__c_rt_mapping" VALUES('Fruit', 'Fruit');
+
+-- fix for allowing namespaced scratch orgs to deploy these record types
```
bdvorachek, line 2195: @psadie-sfdo Here's one possible solution for the namespaced scratch org problem with record types.

psadie-sfdo (approved):
Looks good to me!

datasets/qa_org/test_data.sql
psadie-sfdo: I'd appreciate feedback from someone in RE on this one

davidmreed: Could you share some additional context on the problem?

I would strongly encourage all teams to stop using namespaced scratch orgs. 2GP orgs are far superior for their intended purpose. There are some wrinkles for NPSP as 2GP orgs use regression configuration rather than QA, but that seems like a fruitful opportunity to unify those configurations.

vmartinov-sforg (approved):
Looks Good!

datasets/qa_org/test_data.sql
bdvorachek: @davidmreed - The name of the record type table is the problem. In namespaced scratch orgs, the name of the CustomObject1__c object when you `Describe` it is `npsp__CustomObject1__c, and that value is what gets used to find the record type table - `npsp__CustomObject1__c_rt_mapping` which doesn't exist.

davidmreed: That is a bug in CumulusCI. I will create a WI to fix it properly.

dfuller114 (approved):
looks good from mystery perspective

PR #6209, opened 2021-02-02
NPSP Localization - Spring '21

dmkud:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/globalValueSetTranslations/Payment_ACH_Code-ja.globalValueSetTranslation
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<GlobalValueSetTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
+    <valueTranslation>
+        <masterLabel>Prearranged Payment and Deposit Entry (PPD)</masterLabel>
+        <translation>Prearranged Payment and Deposit Entry (PPD)</translation>
```
force2b, line 5: @dmkud This appears to be missing the japanese translations

src/globalValueSetTranslations/Payment_ACH_Code-nl_NL.globalValueSetTranslation
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<GlobalValueSetTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
+    <valueTranslation>
+        <masterLabel>Prearranged Payment and Deposit Entry (PPD)</masterLabel>
```
force2b, line 4: Missing Dutch Translations

force2b (commented):
First glance and noticed what might be a couple of issues. Will do a deeper dive once we run the metadata cleanup scripts

dmkud (commented):
Left in English by the linguist intentionally.

force2b:
Thank you for the updates on these @dmkud.

dmkud:
@force2b 
I pushed the Adoption drop 2 files from the closed PR (https://github.com/SalesforceFoundation/NPSP/pull/6218) here.
cc: @bdvorachek

metadeploy/labels_de.json
```diff
@@ -117,6 +165,34 @@
...
+        "Nonprofit Success Pack is required to install NPSP Reports & Dashboards. Please install NPSP and try again.": {
+            "message": "Nonprofit Success Pack ist erforderlich, um NPSP-Berichte und -Dashboards zu installieren. Installieren Sie NPSP, und versuchen Sie es noch einmal.",
+            "description": "shown if validation fails"
+        },
+        "Advanced Currency Management must be disabled to install Nonprofit Success Pack, but can be enabled after installation.": {
+            "message": "Für die Installation von Nonprofit Success Pack muss das erweiterte Währungsmanagement deaktiviert sein, Sie können es aber nach der Installation aktivieren.",
+            "description": "shown if validation fails"
+        },
+        "This installer is intended only for Trailhead Playground orgs. Your org doesn't appear to be a Trailhead Playground. Continuing with the installation can damage your org.": {
+            "message": "",
```
bdvorachek, line 194: This message doesn't have a translation for de as well as es/fr/ja/nl.

src/objectTranslations/Addr_Verification_Settings__c-de.objectTranslation
```diff
@@ -3,42 +3,42 @@
...
         <caseType>Nominative</caseType>
         <plural>false</plural>
-        <value><!-- Address Verification Settings --></value>
+        <value>Adressüberprüfungseinstellung</value>
     </caseValues>
     <caseValues>
         <caseType>Nominative</caseType>
         <plural>true</plural>
-        <value><!-- Address Verification Settings --></value>
+        <value><!-- Adressüberprüfungseinstellung --></value>
```
bdvorachek, line 11: should these still be commented out?

bdvorachek, line 11: Should these be commented out still?

src/objectTranslations/GetStartedChecklistItem__mdt-iw.objectTranslation
```diff
@@ -3,22 +3,145 @@
...
         <article>None</article>
         <plural>false</plural>
-        <value><!-- NPSP Get Started Admin Checklist Item --></value>
+        <value><!-- NPSP Get Started Checklist Item --></value>
     </caseValues>
     <caseValues>
         <article>Definite</article>
         <plural>false</plural>
-        <value><!-- NPSP Get Started Admin Checklist Item --></value>
+        <value><!-- הNPSP Get Started Checklist Item --></value>
```
bdvorachek, line 11: This ה character turns up in a few spots that I wouldn't expect it to in the Hebrew translations alongside english text. All have been commented out so far.

force2b: @bdvorachek  I've seen that in places as well. Technically we don't support Hebrew as a supported languages, but the files are still in our repo because it was started at one point.

dmkud:
@force2b Mike, I pushed two commits with fixes for metadeploy and plural forms.
cc: @bdvorachek

src/objectTranslations/Allocation__c-fr.objectTranslation
```diff
@@ -1,73 +1,69 @@
...
 <CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
-    <caseValues>
-        <plural>false</plural>
-        <value>Allocation UCG</value>
-    </caseValues>
-    <caseValues>
-        <plural>true</plural>
-        <value>Allocations UCG</value>
-    </caseValues>
-    <fieldSets>
```
force2b: @dmkud I noticed that the field set label was not translated on this object. I also noticed that on most objects, the field set label is cryptic and probably not translatable, but that's not the case here.

dmkud: @force2b Allocations is a loan word in French, so it's correct

force2b: Thank you @dmkud

bdvorachek (commented):
Ran through the tests on this branch, got three failures. Think I have a fix for the third one w/ the System Admin profile, but still looking into the first two.

- CRLP_RollupCMT_TEST.shouldConvertPicklistApiNamesIntoLabels
  - System.AssertException: Assertion Failed: The value label should have been set from the Api Name: Expected: 1;
Último día del mes, Actual: 1;
Last Day Of Month
  - For some reason we're getting back the English label here instead of the Spanish when we run the tests in Spanish language.
- ERR_ExceptionData_TEST.testExceptionWrapperCreationForDMLException
  - System.AssertException: Assertion Failed: Expected: Required fields are missing: [Name], Actual: Faltan campos obligatorios: [Name]
  - Looks like we need to add a branch to this test that accounts for the language being set to Spanish.
- PS_GatewayManagement_TEST.shouldReturnAsAdmin
  - System.QueryException: List has no rows for assignment to SObject
2021-02-12 13:49:27: 	StackTrace: Class.UTIL_UnitTestData_TEST.buildUser: line 756, column 1
Class.UTIL_UnitTestData_TEST.buildUser: line 745, column 1
Class.UTIL_UnitTestData_TEST.createUserWithoutInsert: line 693, column 1
Class.PS_GatewayManagement_TEST.shouldReturnAsAdmin: line 72, column 1
  - This was an instance of the System Admin profile being hardcoded, updating to use UTIL_Profile's method of getting back the Sysadmin profile.

force2b:
@bdvorachek Thanks for the info on the 3 failing tests:

- CRLP_RollupCMT_TEST: I think I've seen this one before. Though, what's really odd is that I just spun up a scratch org, changed the language to Spanish and ran this test. No failure. Strange.
- ERR_ExceptionData_TEST: What is the error here? The method doesn't look like it has anything hardcoded in English other than the field Api Name, which wouldn't change anyway. Also, same result as above here. No error when I ran the test in my scratch org. 
- PS_GatewayManagement_TEST. This was I was able to reproduce.

🤷

force2b (approved):

bdvorachek (approved):

PR #6210, opened 2021-02-02
Dispatch Scheduled Jobs from Apex

npsp-reedestockton:
# Critical Changes

# Changes
Thank you to Koen Wesselman for adding the ability to dispatch scheduled jobs from Apex with this modification to the Callable_API class.

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Label
- CallableApiMethodNotImplemented

# Deleted Metadata


npsp-reedestockton:
Track [W-8712689](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008wnimIAA/view)

npsp-reedestockton:
@force2b @andrewyu-salesforce Ready for review.  Would appreciate your feedback.
@sam-knox @mattwashington There is a new label here.  Would appreciate your review.

Thanks all!!!

src/classes/CallableApiParameters.cls
```diff
@@ -49,6 +49,9 @@ public class CallableApiParameters {
     public static final String PARAM_PROCESSING_BOUNDARY = 'ProcessingBoundary';
     public static final String PARAM_DISABLE_ALLOCATION_TRIGGERS = 'DisableAllocationTriggers';
     public static final String PARAM_SCHEDULE_JOBS = 'ScheduleJobs';
+    public static final String PARAM_SCHEDULED_JOB_NAME = 'ScheduledJobName';
+    public static final String PARAM_SCHEDULED_JOB_CRON_EXPRESSION = 'ScheduledJobCronExpression';
+    public static final String PARAM_SCHEDULED_JOB_CLASS = 'ScheduledJobClass';
```
force2b, line 54: I think we can remove the "ScheduledJob" part of the param values to since the params are only used as part of the job.schedule task:
```suggestion
    public static final String PARAM_SCHEDULED_JOB_NAME = 'JobName';
    public static final String PARAM_SCHEDULED_JOB_CRON_EXPRESSION = 'CronExpression';
    public static final String PARAM_SCHEDULED_JOB_CLASS = 'ApexClass';
```

src/classes/Callable_API.cls
```diff
@@ -146,6 +146,24 @@ global with sharing class Callable_API implements System.Callable {
                 Boolean scheduleJobsWhenComplete = paramService.getBoolean(CallableApiParameters.PARAM_SCHEDULE_JOBS, true);
                 return new CRLP_EnablementService().enable(scheduleJobsWhenComplete);
 
+            } when 'job.schedule' { 
+                // This method schedules the specified Schedulable class using the specified job name & cron expression.
+                // It should only be used for Schedulable classes that are in the NPSP Managed Package.
+                // It returns the Job ID of the Scheduled Job.
+                String scheduledJobName = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_NAME);
+                String scheduledJobCronExpression = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CRON_EXPRESSION);
+                String scheduledJobClass = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CLASS);
```
force2b, line 155: Following with the above comment regarding the parameter names, I think the var names can be shorted as well.
```suggestion
                String jobName = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_NAME);
                String cronExpression = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CRON_EXPRESSION);
                String apexClass = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CLASS);
```

src/classes/Callable_API.cls
```diff
@@ -146,6 +146,24 @@ global with sharing class Callable_API implements System.Callable {
...
+                
+                Type scheduledClassType = Type.forName(scheduledJobClass);
```
force2b, line 157:
```suggestion
                Type apexClassType = Type.forName(scheduledJobClass);
```

src/classes/Callable_API_TEST.cls
```diff
@@ -34,13 +34,15 @@
...
-@IsTest(isParallel=true)
+// isParallel is FALSE as shouldScheduleJob method touches CronJobDetail which cannot be accessed when testing in parallel
+@IsTest(isParallel=false)
 private with sharing class Callable_API_TEST {
 
     private static final String TDTM_TRIGGER_ACTION = 'TDTM.DisableAllTriggers';
     private static final String CRLP_IS_ENABLED_ACTION = 'CRLP.iscrlpenabled';
     private static final String RD2_EXECUTE_MIGRATION_ACTION = 'RD2.ExecuteDataMigration';
     private static final String RD2_ENABLE_ACTION = 'Settings.EnableEnhancedRecurringDonations';
+    private static final String JOB_SCHEDULE_ACTION = 'Job.Schedule';
```
force2b, line 45:
```suggestion
    private static final String JOB_SCHEDULE_ACTION = 'Apex.ScheduleJob';
```

src/classes/Callable_API.cls
```diff
@@ -146,6 +146,24 @@ global with sharing class Callable_API implements System.Callable {
                 Boolean scheduleJobsWhenComplete = paramService.getBoolean(CallableApiParameters.PARAM_SCHEDULE_JOBS, true);
                 return new CRLP_EnablementService().enable(scheduleJobsWhenComplete);
 
+            } when 'job.schedule' {
```
force2b, line 149:
```suggestion
            } when 'apex.schedulejob' { 
```

force2b (changes requested):
Overall I think this looks good. I requested some parameter and api name changes, but the overall logic looks solid. I have two thoughts though:

1. Because the unit test for this new change forces the test class to not run with IsParallel=true, does it make sense to move that one unit test into a different test class rather than relegate the entire class to not run in parallel? 
2. With the above in mind, could we delegate the entire job logic to another class. For example, could `Callable_API` delegate to `CallableApiApexDelegationService` with a unit test class of `CallableApiApexDelegationService_TEST`. I think it's feasible that we'll eventually support a new api action of `Apex.ExecuteBatch`, which would fit nicely into that same new service class.

npsp-reedestockton:
@force2b Code review suggestions have been implemented.  

Really like refactoring into delegation service to support the possibility of extension!

I debated making CallableApiApexDelegationService global to provide the same level of visibility, but since it is never directly called outside the managed package, decided that was a poor approach, so made it public.  Happy to change if you feel differently about this.

src/classes/CallableApiApexDelegationService.cls
```diff
@@ -0,0 +1,66 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group API
+* @group-content ../../ApexDocContent/API.htm
+* @description API class that handles apex jobs
+*/
+
+public with sharing class CallableApiApexDelegationService {
```
force2b, line 38:
```suggestion
public inherited sharing class CallableApiApexDelegationService {
```

src/classes/CallableApiApexDelegationService.cls
```diff
@@ -0,0 +1,66 @@
...
+        String apexClass = paramService.getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CLASS);
+
+        Type apexClassType = Type.forName(apexClass);
+        if (apexClassType == null) {
+            throw new Callable_API.MalformedMethodInvocationException(
+                    String.format(System.Label.CallableApiJobScheduleInvalidClass, new List<String> { apexClass })
+            );
+        }
+
+        Schedulable schedulableClass = (Schedulable) apexClassType.newInstance();
```
force2b, line 63: It might be good to wrap this in a try catch block as well in case the cron expression or job name causes a failure - unless the expectation is that the pass through exception is valid - which it might be. Alternatively we could put the cron expression through a [regex](https://stackoverflow.com/questions/14203122/create-a-regular-expression-for-cron-statement/19046939) to validate it.

src/classes/CallableApiApexDelegationService.cls
```diff
@@ -0,0 +1,66 @@
...
+    * @description Handles scheduling apex jobs specified in params
+    * @param params name value map of arguments
+    * @return Object returned is String with CronTrigger.Id
+    ********************************************************************************************************/
+    public static Object handleApexScheduleJob(Map<String, Object> params) {
+        // This method schedules the specified Schedulable class using the specified job name & cron expression.
+        // It should only be used for Schedulable classes that are in the NPSP Managed Package.
+        // It returns the Job ID of the Scheduled Job.
+
+        CallableApiParameters paramService = new CallableApiParameters(params);
```
force2b, line 50: I recommend including `withRequiredParamCheck()` since all 3 parameters are required here.

src/classes/CallableApiApexDelegationService.cls
```diff
@@ -0,0 +1,66 @@
...
+
+public with sharing class CallableApiApexDelegationService {
+
+    /*******************************************************************************************************
+    * @description Handles scheduling apex jobs specified in params
+    * @param params name value map of arguments
+    * @return Object returned is String with CronTrigger.Id
+    ********************************************************************************************************/
+    public static Object handleApexScheduleJob(Map<String, Object> params) {
+        // This method schedules the specified Schedulable class using the specified job name & cron expression.
```
force2b, line 46: Are these comments needed here too?

src/labels/CustomLabels.labels
```diff
@@ -5157,6 +5157,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>campaignMemberStatusResponded</shortDescription>
         <value>campaignMemberStatusResponded</value>
     </labels>
+    <labels>
+        <fullName>CallableApiJobScheduleInvalidClass</fullName>
+        <categories>API</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>CallableApi Job.Schedule Invalid Class</shortDescription>
+        <value>{0} is not an Apex Class</value>
```
force2b, line 5166:
```suggestion
        <value>An Apex class for &quot;{0}&quot; cannot be found or does not implement the Scheduleable interface</value>
```

force2b (changes requested):
I like the new structure with the 2 new classes. A few more recommended changes though

src/classes/CallableApiApexDelegationService.cls
npsp-reedestockton: An invalid cron expression causes an apex error that is already clear than an invalid cron expression was the issue.  Do you think there is value in throwing our own error?

force2b: That's fine then. Mostly just making sure that's accounted for. We shouldn't do this now, but it would be interesting if we could support cron expressions like "IN_MINUTES:N" where it parses out the minutes value and dynamically generates a cron expression to run the job once _N_ minutes from the current time. We can wait for a customer to request that kind of functionality.

npsp-reedestockton: Yes, that would be nice.  Looking forward to extending and improving this functionality!

npsp-reedestockton:
@force2b Change requests have made.   Would appreciate another pass from you.  Thanks!

src/classes/CallableApiApexDelegationService.cls
```diff
@@ -0,0 +1,68 @@
...
+
+        String jobName = paramService
+            .withRequiredParamCheck()
+            .getString(CallableApiParameters.PARAM_SCHEDULED_JOB_NAME);
+        String cronExpression = paramService
+            .withRequiredParamCheck()
+            .getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CRON_EXPRESSION);
+        String apexClass = paramService
+            .withRequiredParamCheck()
+            .getString(CallableApiParameters.PARAM_SCHEDULED_JOB_CLASS);
```
force2b, line 56: You only have to include the call to `.withRequiredParamCheck()` one time in the instantiation on line 46.

force2b (changes requested):
Sorry, one more change needed.

src/classes/CallableApiApexDelegationService.cls
npsp-reedestockton: Thanks!   Misremembered how it works.  Appreciate the correction!

force2b (approved):

src/labels/CustomLabels.labels
```diff
@@ -5157,6 +5157,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>campaignMemberStatusResponded</shortDescription>
         <value>campaignMemberStatusResponded</value>
     </labels>
+    <labels>
+        <fullName>CallableApiJobScheduleInvalidClass</fullName>
+        <categories>API</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>CallableApi Job.Schedule Invalid Class</shortDescription>
+        <value>An Apex class for &quot;{0}&quot; cannot be found or does not implement the Scheduleable interface</value>
```
sam-knox, line 5166: @npsp-reedestockton - Just so I understand, the {0} variable here is the name of an invalid class?

andrewyu-salesforce (approved):
@npsp-reedestockton Great work

src/labels/CustomLabels.labels
npsp-reedestockton: @sam-knox That's correct.  The parameter is whatever the caller passed in as the apex class name.

npsp-reedestockton:
@force2b Added logic to throw ParameterException in the edge cases of an empty job name or cron expression.  Both of those cases return unhelpful apex errors by default.  Also added unit tests to validate behavior in those scenarios.

Would appreciate another peek at this.  Thank you!!

force2b (approved):
Great work!

sam-knox (changes requested):
Here's my recommended text for the error label:

"The Apex class {0} doesn't exist or doesn't implement the Schedulable interface."

npsp-reedestockton:
@sam-knox Altered the label per your request.
@force2b @andrewyu-salesforce The label change introduced a test failure which I have corrected.
@rponti-sforg No changes to non-test code other than the label.

src/labels/CustomLabels.labels
```diff
@@ -5163,7 +5163,7 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>CallableApi Job.Schedule Invalid Class</shortDescription>
-        <value>An Apex class for &quot;{0}&quot; cannot be found or does not implement the Scheduleable interface</value>
+        <value>The Apex class {0} doesn''t exist or doesn''t implement the Schedulable interface.</value>
```
force2b, line 5166: @npsp-reedestockton  The apostrophe's need to be escaped, though why are there two?

npsp-reedestockton: Actually, in my testing, escaping the apostrophe does not work in a context where the error is logged.  "\'" and &apos; both result in "doesnt".  

The double apostrophe is what actually makes the contraction appear correctly.

npsp-reedestockton: The perfect compromise.  Double escaped apostrophes.  ;-)

sam-knox (approved):

koenwesselman:
@npsp-reedestockton , this may not be the place to raise it - but the CallableApiApexDelegationService does not seem to be a global class, so its not currently publicly accessible in the sandbox beta release.

koenwesselman:
@npsp-reedestockton - Oops, ignore my previous comment! Its not the delegator that's global because we're still passing through the Callable_API.. My mistake!

Thanks so much for your changes and taking my minor contribution to the next level in incorporating it in NPSP! Much appreciated and will make a huge impact for our project in terms of overnight Scheduled Job performance! :-)

npsp-reedestockton:
@koenwesselman Glad to hear things are okay.

The reaction to your contribution has been tremendously enthusiastic.  Thanks again for your _major_ contribution!!

PR #6212, opened 2021-02-04
Enforce Security Permission on NPSP Settings Pages

andrewyu-salesforce:
A user without edit rights to an object are not allowed to perform operations that modify those types of records in NPSP Settings.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


andrewyu-salesforce:
Attach to [WI-8783337](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000094XDjIAM/view)

andrewyu-salesforce:
@npsp-reedestockton @force2b The PR is ready for review

src/classes/STG_PanelCreatePayments_CTRL.cls
```diff
@@ -34,26 +34,35 @@
...
+    /**
     * @description Returns the string Id of the Donor Statistics panel. 
     */
     public override string idPanel() { return 'idPanelCreatePayments'; }
     
-    /*********************************************************************************************************
+    /**
+    * @description NPSP namespace: an empty string if unmanaged, or 'npsp' if managed
+    */
+    public String getNamespace() {
```
force2b, line 45: Since all of these controller classes extend from `STG_Panel`, I recommend putting the `getNamespace()` method in there this way it's available to all the classes. I think that'll still work for the getter, but the worst case is that this method in each individual controller simply returns the call from a method in the parent `STG_Panel` class.

src/classes/STG_PanelMakeDefaultAllocations_TEST.cls
```diff
@@ -0,0 +1,70 @@
+/*
+    Copyright (c) 2016, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2021, Salesforce.org
```

src/classes/STG_PanelRDBatch_CTRL_TEST.cls
```diff
@@ -0,0 +1,70 @@
+/*
+    Copyright (c) 2016, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2021, Salesforce.org
```

src/classes/STG_PanelRenameHouseholds_CTRL_TEST.cls
```diff
@@ -0,0 +1,70 @@
+/*
+    Copyright (c) 2016, Salesforce.org
```
force2b, line 2: 2021

src/classes/STG_PanelCreatePayments_CTRL.cls
```diff
@@ -62,4 +71,18 @@ public with sharing class STG_PanelCreatePayments_CTRL extends STG_Panel {
...
         return null;
     }
+
+    /**
+    * @description Check if the running user has update access to the Opportunity sObject and Read access to the
+    * Payment object
+    * @return Boolean
+    */
+    public Boolean getHasAccess() {
+        String oppObject = Schema.SObjectType.Opportunity.getName();
```
force2b, line 81: It might also be useful to move all of the object declarations like these - for all of the classes - into the `STG_Panel` class as well - as a static var. This way, the call to `Schema.SObjectType.Opportunity.getName()` (and so on) is only made one time for all fo the controllers. That may aid in performance because NPSP Settings loads up every single page at the beginning. As a result, it's effectively making these calls multiple times for every page as it loads its controller.

src/classes/STG_PanelCreatePayments_CTRL_TEST.cls
```diff
@@ -0,0 +1,70 @@
+/*
+    Copyright (c) 2016, Salesforce.org
```
force2b, line 2: The copyright on all the new test classes should be 2021

src/pages/STG_PanelCreatePayments.page
```diff
@@ -1,21 +1,45 @@
 <apex:page controller="STG_PanelCreatePayments_CTRL" docType="html-5.0" standardStylesheets="false" >
+    <apex:includeLightning/>
+    <script>
```
force2b, line 3: I know this is a `<script>` block, but I'm wondering if we can put this little block into a VF component that all the individual pages load as
`<c:STG_PermissionsCheckAndNotify hasAccess={hasAccess} />`.

If this works, then any other STG page using the same block should be updated to use this (even if that page wasn't changed in this PR).

force2b (changes requested):
Nice work. Because there's a lot of duplication, I'm suggesting some ways to create common code for re-use across the pages.

src/classes/STG_PanelRDBatch_CTRL.cls
```diff
@@ -135,4 +137,17 @@ public with sharing class STG_PanelRDBatch_CTRL extends STG_Panel {
...
+    * @description Check if the running user has update access to the Opportunity sObject and Recurring Donation sObject
+    * Contact sObject
+    * @return Boolean
+    */
+    public Boolean getHasAccess() {
+        String oppObject = Schema.SObjectType.Opportunity.getName();
+        String rdObject = Schema.SObjectType.npe03__Recurring_Donation__c.getName();
+
+        return UTIL_Permissions.canUpdate(oppObject, false) &&
+        UTIL_Permissions.canUpdate(rdObject, false);
```
npsp-reedestockton, line 150: Nitpick - Should probably indent UTIL...

npsp-reedestockton (changes requested):
Looks good!  Just one minor formatting issue noted.

npsp-reedestockton (approved):

src/classes/STG_PanelCreatePayments_CTRL.cls
andrewyu-salesforce: all `getNamespace()` method are moved to `STG_Panel`

src/classes/STG_PanelCreatePayments_CTRL.cls
andrewyu-salesforce: I was reading the dev doc and it seem like token describe is lightweight and execute on compile time. Therefore, I do not think there will be any performance gain to create the static variables in the parent class. [documentation](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_dynamic_describe_objects_understanding.htm#:~:text=Apex%20provides%20two%20data%20structures,is%20used%20for%20token%20describes.)

src/pages/STG_PanelCreatePayments.page
andrewyu-salesforce: I am unable to create a vf component for to load the lightning out. From the error it seem like the component will be loaded on the initial with `STG_Manager` and the error saying Lightning-out is already hooked to something else will be thrown.

andrewyu-salesforce:
@npsp-reedestockton @force2b Thanks for the first review. The code is updated and ready for second review. I am able to moved the `getNamespace()` method to `STG_Panel` but have no luck with moving the lightning script to a separate component due to `lightning-use`. I also had some comment on moving the token describe to the parent class.

src/classes/STG_Panel.cls
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2014, Salesforce.org
+    Copyright (c) 2021, Salesforce.org
```
force2b, line 2: Older classes do not have to be updated to 2021 - it was only for new classes created in this PR.

src/pages/STG_PanelCreatePayments.page
force2b: Thanks for checking into it.

force2b (commented):
Looks good. Really just the one comment about reverting the copyright year on pre-existing classes.

andrewyu-salesforce:
@force2b The copyright dates are reverted.

force2b (approved):

PR #6218, opened 2021-02-09
Adoption Localization drop 2 - Spring '21

dmkud:
# Critical Changes

Translated labels from drop 2 for Adoption Spring '21

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
@dmkud Thanks, but for LQA we need to keep the two drops in a single branch/pr. I'm not sure if we can just merge this one into the first or if you have to re-do the translation into the other branch, but there won't be a way for us to build an LQA org that has translations from two different branches. Plus, it's double the work on our end to review and update two large PR's (which also impacts our timeline significantly).
CC @bdvorachek

force2b:
@dmkud Closing this PR since the translations will all go into the existing branch.

PR #6222, opened 2021-02-10
[Robot] Changing the cci command to enable_rd2 on the org the robot tests are run

rjanjanam:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


robot/Cumulus/resources/NPSP.py
```diff
@@ -1697,6 +1697,14 @@ def verify_table_contains_row(self,table_name,record,**kwargs):
...
-
-
-
+            
+    def get_org_name(self):
+        """
+        Obtains the orgname info from the cumulsci's get org info and returns the value
+        """
+        org_info = self.cumulusci.get_org_info()
+        if "sfdx_alias" in org_info.keys():
```
boakley, line 1706: The cumulusci library has an attribute named `org` which has a `name` attribute. You can simply return `self.cumulusci.org.name` instead of getting the org info and parsing the result.

robot/Cumulus/resources/RecurringDonationsPageObject.py
```diff
@@ -18,10 +18,16 @@ class RDListingPage(BaseNPSPPage, ListingPage):
     @capture_screenshot_on_error
     def wait_for_rd2_modal(self):
         """Based on the button name (Cancel)  or (Save) on the modal footer, selects and clicks on the respective button"""
-        self.builtin.sleep(2,"Wait Needed for now to wait for the new modal")
+        self.builtin.sleep(3,"Wait Needed for now to wait for the new modal")
         btnlocator = npsp_lex_locators["button-with-text"].format("Save")
         self.selenium.scroll_element_into_view(btnlocator)
         self.selenium.wait_until_element_is_visible(btnlocator,60)
+        for i in range(3):
```
boakley, line 25: Why is this necessary, since you're already waiting up to 60 seconds for the element to be visible?

boakley (changes requested):

robot/Cumulus/resources/RecurringDonationsPageObject.py
rjanjanam: This logic is added to help handle the unpredicted load times of the Recurring donations modal.

boakley: I still don't quite understand. You're already waiting for it to be visible. Are you saying that the element could be visible but that `self.npsp.check_if_element_displayed(btnlocator)` can still return `False`? What else is `check_if_element_displayed` doing besides checking that it is visible?

If this extra check really is necessary, please add a comment which explains why it is necessary. Otherwise, someone else may come by later and think it's a redundant check and remove it.

rjanjanam: Removed the redundant wait and added a comment for the logic I added in the loop

robot/Cumulus/resources/RecurringDonationsPageObject.py
```diff
@@ -18,10 +18,16 @@ class RDListingPage(BaseNPSPPage, ListingPage):
     @capture_screenshot_on_error
     def wait_for_rd2_modal(self):
         """Based on the button name (Cancel)  or (Save) on the modal footer, selects and clicks on the respective button"""
-        self.builtin.sleep(2,"Wait Needed for now to wait for the new modal")
+        self.builtin.sleep(3,"Wait Needed for now to wait for the new modal")
```
schlossmj, line 21: Any way an implicit wait could be used here? (Wait For Element to be Present, etc)

rjanjanam: Changed it with a better way of waiting

robot/Cumulus/tests/browser/zrecurring_donations/zcreate_fixed_erd_record.robot
```diff
@@ -56,7 +56,10 @@ Create Fixed Recurring Donation With Monthly Installment
     ...                                     Payment Method=Credit Card
     ...                                     Day of Month=${day_of_month}
     Click Rd2 Modal Button                  Save
+
     Wait Until Modal Is Closed
+    Capture Page Screenshot
+    Sleep                                   1
```
boakley, line 62: Did you intend to leave this sleep statement in?

boakley (approved):
Looks good to me.

robot/Cumulus/tests/browser/zrecurring_donations/zcreate_fixed_erd_record.robot
rjanjanam: yes

robot/Cumulus/resources/RecurringDonationsPageObject.py
schlossmj: 👍

schlossmj (approved):

PR #6223, opened 2021-02-10
Refactor NPSP Household Naming To Use Domain Classes

kenlewis-sfdo:
Introduces Accounts, Contacts, Households, HouseholdSettings,
HouseholdName, etc.  The addition of these new Types is intended to 
clarify the logic that lives within this feature area and begin to move 
core NPSP functionality toward a domain-driven architecture.

WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000095HOeIAM/view
Domain diagram: https://lucid.app/lucidchart/invitations/accept/0e818758-b0ef-47c6-a4b1-cd3e92011005

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata
* HH_HouseholdNaming.cls (Renamed to HouseholdNamingService.cls)
* HH_HouseholdNaming_TEST.cls (Renamed to HouseholdNamingService_TEST.cls)
* HH_NameSpec.cls (Renamed to HH_INamingImpl.cls)
* HH_NameSpec_TEST.cls (Renamed to HH_INamingImpl_TEST.cls)


src/classes/Contacts.cls
```diff
@@ -0,0 +1,748 @@
...
+    }
+
+    public void buildOldHouseholdCollections() {
+        oldHouseholds.addAll(getOldHouseholdIds());
+        buildOldHouseholdSizeMap();
+    }
+
+    private void buildOldHouseholdSizeMap() {
+        for (AggregateResult a : getMemberCountForOldHouseholds()) {
+            oldHouseholdSizeById.put((id) a.get('oldhh'), (integer) a.get('membercount'));
```
vmartinov-sforg, line 134: Let's declare constants for 'oldhh' and 'membercount'.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,748 @@
...
+            if (isInsertTrigger) {
+                updateNamesAfterHouseholdInsert(householdSaveResults);
+            }
+        }
+    }
+
+    private List<Contact> processHouseholdSaveResults(List<Contact> contacts,
+            Database.SaveResult[] householdSaveResults) {
+        List<Contact> contactsToBeUpdatedWithHouseholdId = new List<Contact>();
+        Integer i = 0;
```
vmartinov-sforg, line 299: This variable is not used and can be removed.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,748 @@
...
+                householdsNeedingNamesUpdated.add(householdIdFor(con));
+            }
+        }
+    }
+
+    private Boolean isHouseholdNamingFieldsChanged(Contact con) {
+        return isNamingExclusionsChanged(con) || isNamingFieldChanged(con);
+    }
+
+    Set<String> householdNamingFieldsInUse;
```
vmartinov-sforg, line 488: Let's move the class variable to the top of the class.

src/classes/HH_HouseholdNaming.cls
```diff
@@ -75,60 +78,72 @@ public without sharing class HH_HouseholdNaming {
...
+        hhupdatelist.addAll(getListHHObj(hhids));
+        hhupdatelist.addAll(getListHHAcct(hhids));
+        return hhupdatelist;
+    }
 
-	            if (!customnameparam.contains('Informal_Greeting__c'))
-	                h.put('npo02__Informal_Greeting__c', getInformalName(listCon));
-            }
+    private List<SObject> getListHHAcct(List<Id> hhids) {
+        List<SObject> listHHAcct = [select Id, Name, npo02__SYSTEM_CUSTOM_NAMING__c from Account where id IN :hhids];
```
vmartinov-sforg, line 119: Suggestion - format the soql per our standard.

src/classes/HH_HouseholdNaming.cls
```diff
@@ -75,60 +78,72 @@ public without sharing class HH_HouseholdNaming {
...
+        return listHHAcct;
+    }
 
-            // now we also store the number of members on the Household!
-            h.put(UTIL_Namespace.StrTokenNSPrefix('Number_of_Household_Members__c'), listCon != null ? listCon.size() : 0);
-        }
+    private List<SObject> getListHHObj(List<Id> hhids) {
+        List<SObject> listHHObj = [select Id, Name, npo02__SYSTEM_CUSTOM_NAMING__c from npo02__Household__c where id IN :hhids];
```
vmartinov-sforg, line 124: Suggestion - format the soql per our standard.

src/classes/HH_HouseholdNaming.cls
```diff
@@ -210,236 +181,149 @@ public without sharing class HH_HouseholdNaming {
...
+    public static void refreshAllHouseholdNaming(Boolean isActivation) {
         // the household batch expects a list of Contact's with just Id, LastName, HHId__c available.
-        string strSoql = 'SELECT Id, LastName, ' + UTIL_Namespace.StrTokenNSPrefix('HHId__c') + ' FROM Contact WHERE ' +
-            UTIL_Namespace.StrTokenNSPrefix('HHId__c') + ' != NULL ';
+        String strSoql = 'SELECT Id, LastName, ' + UTIL_Namespace.StrTokenNSPrefix('HHId__c') + ' FROM Contact WHERE ' +
+                UTIL_Namespace.StrTokenNSPrefix('HHId__c') + ' != NULL ';
         HH_HouseholdNaming_BATCH batch = new HH_HouseholdNaming_BATCH(strSoql, isActivation);
-        integer batchSize = 200;
-        id batchProcessId = database.executeBatch(batch, batchSize);
+        Integer batchSize = 200;
```
vmartinov-sforg, line 234: let's use a constant for 200. There may be one declared already.

vmartinov-sforg (commented):
Looks good - I only have a few comments.

src/classes/Contacts.cls
kenlewis-sfdo: I factored these down a bit for additional clarity but since the Aggregate query doesn't seem to be able to use the constant, I think it is ok to leave these as they are now.

src/classes/Contacts.cls
kenlewis-sfdo: Tried removing but looks like it IS actually used!  Ha - I didn't see that either - used to grab the correct result a few lines below this.

src/classes/Contacts.cls
kenlewis-sfdo: 👍🏽

src/classes/HH_HouseholdNaming.cls
kenlewis-sfdo: 👍🏽

src/classes/HH_HouseholdNaming.cls
kenlewis-sfdo: 👍🏽

src/classes/HH_HouseholdNaming.cls
kenlewis-sfdo: Actually, I think this should be removed since the default batch size is 200, but I don't really want to change this within this work stream.  I'm going to leave it as is for now with a goal of factoring this out later.

src/classes/HH_HouseholdNaming.cls
```diff
@@ -37,13 +37,16 @@
...
+    String nameReplacementText = System.Label.npo02.NameReplacementText;
+
     /*******************************************************************************************************
     * @description future method to update the household names for the specified households
     * @param hhids the list of household Id's (either Household__c or Account Id's)
     * @return void
     */
     @future
-    public static void FutureUpdateNames(list<id> hhids){
+    public static void FutureUpdateNames(List<Id> hhids){
```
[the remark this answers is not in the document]
kenlewis-sfdo: 👍🏽

src/classes/HH_HouseholdNaming.cls
```diff
@@ -65,7 +68,7 @@ public without sharing class HH_HouseholdNaming {
     * @param hhids the list of household Id's (either Household__c or Account Id's)
     * @return void
     */
-    public void UpdateNames(list<id> hhids){
+    public void UpdateNames(List<Id> hhids){
```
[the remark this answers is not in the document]
kenlewis-sfdo: 👍🏽

voduyemi (commented):
Thanks @kenlewis-sfdo  this looks really good! I had some more comments around variable renaming and extracting some logic into a method (Should have done this in my first pass). Something else I would love to see maybe during the next phase of this refactor would be to abstract all the queries in these classes into some sort of data access layer(selector classes) so we don't repeat the same queries across all these classes and for clarity.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,729 @@
...
+    * @description Creates a new Household Object for each contact, unless excluded by recordType
+    * @param contacts List of Contacts
+    * @param isInsertTrigger Whether called from the afterInsert trigger (vs. afterUpdate trigger) on Contacts
+    * @return void
+    ********************************************************************************************************/
+    private void insertHouseholdsForContacts(List<Contact> contacts, Boolean isInsertTrigger) {
+        List<npo02__Household__c> householdInserts = getHouseholdsForInsert(contacts);
+        if (householdInserts.size() > 0) {
+            Database.SaveResult[] householdSaveResults =
+                    UTIL_DMLService.insertRecords(householdInserts, false);
```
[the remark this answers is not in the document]
kenlewis-sfdo: Does this collection need to be upcasted back to an SObject collection before doing the DML?

kenlewis-sfdo:
Thanks for the review @voduyemi !  I made most of those changes - please take another look!  Yes, Selector classes should be factored out and will be in future WI.

src/classes/Contacts.cls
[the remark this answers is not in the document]
voduyemi: Well my thinking was we give the DML service the type of collection it expects but on a closer look, we may not need to make this change.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,732 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Domain class for Contacts in NPSP.
+*/
+public inherited sharing class Contacts extends fflib_SObjects {
```
force2b, line 35: Overall I like how the class encapsulates Contact related logic, however I have a few thoughts:

- It's a huge class and it may be worth considering abstracting common logic out into its own service classes - naming, address, etc.
- I see your point on not adding block comments, but there are a lot of methods here that could use comments. We've always settled that every method requires a block comment, and I'm not sure we should try to change that rule here. Yes, some comments on some methods may be redundant, but I think there are enough complex methods in here that commenting will be beneficial.
- There's a lot of logic that [I believe] is unique to the older account models (Individual and Bucket). What would you think of abstracting those out into another class or even an inner class - as a way of making it super clear while reading the code that these are methods in support of legacy models?

src/classes/HH_Container_LCTRL.cls
```diff
@@ -518,26 +518,16 @@ public with sharing class HH_Container_LCTRL {
     public static SObject getHHNamesGreetings(SObject hh, list<Contact> listCon) {
         try {
             // update naming exclusions field based on exclusion checkboxes
-            for (Contact con : listCon) {
-                HH_Households_TDTM.copyNamingExclusionsFromCheckboxes(con);
-            }
+            Contacts contacts = new Contacts(listCon);
```
force2b, line 521: Instantiations of this domain class as just `contacts` will be easily confused will collections of Contacts. I know you would prefer to keep the class name as is, but either way I think instances of the class have to be named differently - such a `contactsDomain` or `contactsService` or `domainContacts`, etc. Just something to differentiate from a `List<Contact>`.

src/classes/HH_Container_LCTRL.cls
```diff
@@ -518,26 +518,16 @@ public with sharing class HH_Container_LCTRL {
...
+            contacts.copyNamingExclusionsFromCheckboxes();
 
-            HH_HouseholdNaming hhNaming = new HH_HouseholdNaming();
-
-            if (!hhNaming.isAutomaticNaming()) {
+            HouseholdSettings settings = new HouseholdSettings();
```
force2b, line 524:
```suggestion
            HouseholdSettings hhSettings = new HouseholdSettings();
```

src/classes/HH_HouseholdNaming.cls
```diff
@@ -65,70 +62,87 @@ public without sharing class HH_HouseholdNaming {
...
-	            string customnameparam = '';
-	            if (h.get('npo02__SYSTEM_CUSTOM_NAMING__c') != null){
-	                customnameparam = string.valueOf(h.get('npo02__SYSTEM_CUSTOM_NAMING__c'));
-	            }
+        TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.HH, false);
+    }
 
-	            if (!customnameparam.contains('Name'))
-	                h.put('Name', getHHName(listCon));
+    private Map<Id, List<Contact>> getHhIDContactMap(List<Id> hhids) {
```
force2b, line 98: These methods need a block comment to clarify usage

src/classes/HH_HouseholdNaming.cls
force2b: It's worth defining as a constant or getter in the `HH_HouseholdNaming_BATCH` class. That'll make it easier for us to eventually implement a setting for this.

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,268 @@
+/*
```
force2b, line 1: I believe this is a service class as opposed to a domain class, so a class name such as `HouseholdNameService` or `HouseholdName_SVC` would be more appropriate.

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,268 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Household Name value object in NPSP.
+*/
+public with sharing class HouseholdName {
```
force2b, line 35: Inherited Sharing.

Overall, `with sharing` should only be used on controller classes for custom UI's. Beyond that all other classes should be `inherited sharing` unless the class is a utility or speciality class of some kind that specifically requires access to records/objects the User might not otherwise have access to.

src/classes/HouseholdSettings.cls
```diff
@@ -0,0 +1,71 @@
+public with sharing class HouseholdSettings {
```
force2b, line 1: Inherited sharing

force2b (commented):
To start, I really love the abstractions! The logic will be so much easier to read and follow when this done. Super fantastic work.

Beyond that, I had a number of renaming recommendations and some additional thoughts on further abstractions. I'm holding the review at this first level of detail, but will continue to dig in deeper on each iteration.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,732 @@
...
+            if (oldHouseholdIsEmpty(hhid))
+                oldHouseholdSizeById.put(hhid, 0);
+        }
+    }
+
+    private Boolean oldHouseholdIsEmpty(Id hhid) {
+        return !oldHouseholdSizeById.containskey(hhid);
+    }
+
+    private Integer memberCountFrom(AggregateResult a) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: Good idea!  I'll take a look.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,732 @@
...
+
+    public void onBeforeUpdate() {
+        synchronizeNamingExclusionsFieldsBeforeUpdate();
+    }
+
+    public void onAfterInsert() {
+        processHouseholdsAfterInsert();
+    }
+
+    private void processHouseholdsAfterInsert() {
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yeah this could definitely be factored down so that it is not doing multiple things.  I'll try to refactor this down a bit more.

src/classes/Accounts.cls
```diff
@@ -0,0 +1,228 @@
+/*
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yep, happy to keep the discussion going!  :) If it helps, think about the codebase a few months from now or a year from now.  Every instance where code outside of the Accounts class is manipulating collections of Account records directly will be refactored to use an instance of Accounts.  This in-between state where we have some Lists/Maps named 'accounts' and some instances of Accounts named 'accounts' is temporary and will be resolved as we continue to push deeper into domain-driven territory.  :)

src/classes/Contacts.cls
kenlewis-sfdo: * Yes, this is a huge class.  Since we are rolling this WI over to next Sprint, I will start a next round of refactor and see if there is more logic that we can abstract out to reduce its size.  Overall, this will be an ongoing process.  While we might not get all the way to our perfect refactoring point in this work, we are moving in the right direction, and we should begin to regularly update our domain model and refactor code when the two get out of line: https://lucid.app/lucidchart/d10afa5b-bbc2-443f-9817-619c148235c5/edit?page=0_0#?folder_id=home&browser=icon
* We should definitely not add block comments to every method.  NPSP needs to move away from that rule in my opinion.  Every time we think "this method does multiple things and needs a comment" that is a sign that the method needs to be refactored.  We want methods that are single chunks of code that do one thing and have expressive names. I'll do another pass and continue to try to break down the methods that do multiple things.  
* Splitting out classes by Account model type is an interesting idea.  I'll look and see if I can start that process.

src/classes/HouseholdName.cls
kenlewis-sfdo: Ok - Intellij auto generates classes as with sharing by default so I need to remember to update this regularly (or maybe look to see if I can change that setting for auto-generated classes... :)

src/classes/Contacts.cls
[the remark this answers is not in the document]
kenlewis-sfdo: Factored this method out - see if the onAfterInsert method looks better now! :)

src/classes/Contacts.cls
```diff
@@ -0,0 +1,732 @@
...
+        }
+    }
+
+    private Boolean hasExclusionCheckboxesChecked(Contact con) {
+        return isExcludedFromHouseholdName(con)
+                || isExcludedFromInformalGreeting(con)
+                || isExcludedFromFormalGreeting(con);
+    }
+
+    private static void synchronizeNamingExclusionStringToCheckboxes(
```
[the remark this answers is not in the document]
kenlewis-sfdo: good catch - updated!

src/classes/HouseholdName.cls
kenlewis-sfdo: Yeah, this might call for a renaming of HH_HouseholdNaming - I'm going to look at that class and this one and see if a few renames and moving a couple methods would have these all make a bit more sense together.

src/classes/HH_HouseholdNaming.cls
kenlewis-sfdo: Extracted it into a method!

src/classes/Contacts.cls
```diff
@@ -0,0 +1,732 @@
...
+            if (shouldCopyNamingExclusionsFromCheckboxes(con)) {
+                copyNamingExclusionsFromCheckboxes(con);
+            } else if (shouldCopyNamingExclusionsFromMultiSelectPicklist(con)) {
+                synchronizeNamingExclusionStringToCheckboxes(
+                        new HouseholdNamingExclusions(namingExclusionsFor(con)), con);
+            }
+        }
+    }
+
+    private Boolean shouldCopyNamingExclusionsFromMultiSelectPicklist(Contact con) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: I like this idea, but haven't made this change yet - I haven't been able to logically come up with a good name for this class, and the functionality is really tightly knit to Contacts.  Happy to look at this again if in your next review you still think there is a good abstraction to pursue here.

src/classes/HouseholdName.cls
kenlewis-sfdo: Renamed HH_HouseholdNaming to HouseholdNamingService, which clarifies its role, especially in relation to the HouseholdName value object.  Let me know what you think!

src/classes/HH_Container_LCTRL.cls
kenlewis-sfdo: For two main reasons I am still hesitant to do this.  
1. As mentioned in another comment, all instances of `List<Contacts>` should change over time to be instances of `Contacts` - there will also be fewer usages over time as behavior is moved to be encapsulated inside the Contacts domain class
2. This kind of naming is bakes the Type into the variable name.  Apex is strongly typed, so putting the Type in the name is a bit redundant, and, while we are seeing a lot of general var names of 'contacts' here, the local variable name can always be named to be more expressive for the variables usage, i.e., `Contacts contactsForNewsletter = new Contacts(List<Contact>)`

src/classes/HouseholdCustomNamingString.cls
```diff
@@ -0,0 +1,91 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Household Custom Naming String value object in NPSP.
+*/
+public inherited sharing class HouseholdCustomNamingString {
```
voduyemi, line 35: I think we can call this class `HouseholdCustomNamingUtil`. What do you think?

src/classes/HouseholdNamingExclusionsString.cls
```diff
@@ -0,0 +1,90 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Domain class for HouseholdNamingExclusions in NPSP.
+*/
+public inherited sharing class HouseholdNamingExclusionsString {
```
voduyemi, line 35: I think we can call this class `HouseholdNamingExclusionsUtil`. What do you think?

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+        }
+        if (!customNamingString.contains('Formal_Greeting__c')) {
+            household.put('npo02__Formal_Greeting__c', householdName.asFormalGreeting());
+        }
+        if (!customNamingString.contains('Informal_Greeting__c')) {
+            household.put('npo02__Informal_Greeting__c', householdName.asInformalGreeting());
+        }
+    }
+
+    public HouseholdCustomNamingString getCustomNamingString(SObject household, SObject oldRecord) {
```
voduyemi, line 408: Since we expect to get a custom naming string here, Do we think we can return a string here instead of the instance of the class?

src/classes/HouseholdCustomNamingString.cls
```diff
@@ -0,0 +1,91 @@
...
+    private String value;
```
force2b, line 36: What does "value" represent here? A different var name would help on this one.

src/classes/HouseholdCustomNamingString.cls
```diff
@@ -0,0 +1,91 @@
...
+
+    public HouseholdCustomNamingString(String str) {
+        this.value = str == null ? ';' : str + ';';
+    }
+
+    public String value() {
```
force2b, line 42: I believe you can combine this and the private on line 36 with a getter/setter combo:
```
public String value {
    get {
        return this.value;
    } private set;
}
```

src/classes/HouseholdCustomNamingString.cls
```diff
@@ -0,0 +1,91 @@
...
+    public Boolean contains(String val) {
+        return customNameSet().contains(val);
+    }
+
+    public void removeName() {
+        replace('Name;', ';');
+    }
+
+    public void removeFormalGreeting() {
+        replace('Formal_Greeting__c;', ';');
```
force2b, line 73: I'm not 100% sure where these are being referenced from, but is there a concern that the namespace needs to be applied to these?

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,115 @@
...
+    public HouseholdName(Contacts contacts, HH_INaming service) {
+        this.contacts = contacts;
+        this.service = service == null ? new HH_NameSpec() : service;
+    }
+
+    /*******************************************************************************************************
+    * @description returns the Household Name for the list of Contacts
+    * @return String
+    */
+    public String value() {
```
force2b, line 53: How does this method  differ from other methods, such as `asInformalGreeting()`? The name is super generic. Is this effectively `getHouseholdName()`?

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,115 @@
...
+        return service.getHouseholdName(getContactsForHouseholdNameGeneration());
+    }
+
+    private List<Contact> getContactsForHouseholdNameGeneration() {
+        List<Contact> contactsForHouseholdNameGeneration = new List<Contact>();
+        for (Contact con : (List<Contact>) contacts.getRecords()) {
+            if (contactShouldBeUsedForHouseholdNameGeneration(con))
```
force2b, line 60: curly brackets needed

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,115 @@
...
+    * @return String
+    */
+    public String asInformalGreeting() {
+        return service.getHouseholdInformalGreeting(getContactsForInformalName());
+    }
+
+    private List<Contact> getContactsForInformalName() {
+        List<Contact> listCon = new List<Contact>();
+        for (Contact con : (List<Contact>) contacts.getRecords()) {
+            if (isIncludedInInformalGreeting(con))
```
force2b, line 82: curly brackets needed

src/classes/HouseholdName.cls
```diff
@@ -0,0 +1,115 @@
...
+    * @return String
+    */
+    public String asFormalGreeting() {
+        return service.getHouseholdFormalGreeting(getContactsForFormalName());
+    }
+
+    private List<Contact> getContactsForFormalName() {
+        List<Contact> contactsForFormalName = new List<Contact>();
+        for (Contact con : (List<Contact>) contacts.getRecords()) {
+            if (isIncludedInFormalGreeting(con))
```
force2b, line 104: curly brackets needed

src/classes/HouseholdNamingExclusionsString.cls
```diff
@@ -0,0 +1,90 @@
...
+    private String value;
```
force2b, line 36: Maybe something like
```suggestion
    private String namingTemplate;
```

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
+/*
+    Copyright (c) 2011, Salesforce.org
```
force2b, line 2: 2021

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2011 (2.3)
```
force2b, line 32: Is this a new class or just renamed? I'm assuming it's renamed, but slightly new - in which case the year is 2021 and the version can be removed.

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+* @group Households
+* @description Controls generation of household names for lists of households
+* NOTE: This class handles both household naming for the Household__c object
+* as well as for HH Account objects.
+*/
+public without sharing class HouseholdNamingService {
```
force2b, line 38: At some point we should understand whether the naming classes truly require "without". I would expect them to be called by either a) a trigger; or b) a batch operation. In both cases, "inherited" sharing would be sufficient. I'm a bit nervous about changing the sharing declaration on an existing class, but as with the above comment I'm not clear on whether this is new or not.

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    private List<SObject> householdsOrAccountsFor(List<Id> householdOrAccountIds) {
+        List<SObject> hhupdatelist = new List<SObject>();
+        hhupdatelist.addAll(householdsFor(householdOrAccountIds));
+        hhupdatelist.addAll(accountsFor(householdOrAccountIds));
+        return hhupdatelist;
+    }
+
+    private List<Account> accounts;
```
force2b, line 135: Good example of where a different var name should be used to avoid confusion with the `Accounts` class.

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+    private List<SObject> accountsFor(List<Id> hhids) {
```
force2b, line 136:
```suggestion
    private List<SObject> getAccountsFor(List<Id> hhids) {
```

Though, both of these methods do not guarantee the result. I can pass one set of Id's to this method and get a response. Then pass a completely different set of Ids and it will return back the same response from the first call. Given that, what is the value of storing the result in a static var here? The calling method should be the only place responsible for caching the result.

src/classes/HouseholdNamingService.cls
```diff
@@ -0,0 +1,512 @@
...
+                    'Id', 'HHId__c', 'npo02__Naming_Exclusions__c'
+            };
+            selectFields.addAll(getHouseholdNamingContactFields());
+
+            queryForContactNamingFields = new UTIL_Query()
+                    .withSelectFields(selectFields)
+                    .withToLabelSalutation()
+                    .withFrom('Contact')
+                    .build();
+            }
```
force2b, line 278: Outdent

src/classes/HouseholdSettings.cls
```diff
@@ -0,0 +1,71 @@
+public inherited sharing class HouseholdSettings {
```
force2b, line 1: Copyright block needed

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -425,7 +425,7 @@ public with sharing class LD_LeadConvertOverride_CTRL {
 
             // fix the account that lead conversion created
             if (strConId == 'NEW_CONTACT' && isIndividualAccount) {
-                string strSoql = 'select id, firstname, lastname,';
+                string strSoql = 'select id, firstname, lastname, accountId,';
```
force2b, line 428:
```suggestion
                string strSoql = 'SELECT Id, Firstname, Lastname, AccountId,';
```

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -425,7 +425,7 @@ public with sharing class LD_LeadConvertOverride_CTRL {
...
                 if (Schema.sObjectType.Contact.fields.getMap().get('RecordTypeId') != null)
                    strSoql += ' recordtypeId,';
```
force2b, line 430:
```suggestion
                if (Schema.sObjectType.Contact.fields.getMap().get('RecordTypeId') != null) {
                   strSoql += ' RecordtypeId,';
               }
```

force2b (changes requested):
Wow. This is **really** starting to look great. I have a lot of comments and a bunch of requested changes, but there is really nothing significant in my comments. I would like to see a bit more abstraction for some inner services to avoid clouding of class purpose - especially in the domain classes. I think that will help with separation of concerns and probably for testing down the road. Great work!!

src/classes/HouseholdNamingExclusionsString.cls
```diff
@@ -0,0 +1,90 @@
...
+        value += 'Household__c.Formal_Greeting__c;';
+    }
+
+    public void excludeInformalGreeting() {
+        value += 'Household__c.Informal_Greeting__c;';
+    }
+
+    public String getDisplayValue() {
+        prepareFinalValue();
+        return value;
```
bdvorachek, line 74: At the very end of all of this, we assign this value to `npo02__Naming_Exclusions__c`. Have you thought about renaming `value` to be `exclusions`? I think that name reads more fluently in this file.

src/classes/HouseholdNamingExclusionsString.cls
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private void prepareFinalValue() {
+        if (value.endsWith(';')) {
```
bdvorachek, line 78: The String class has received a few new methods since this code was first written, and we should keep an eye out for opportunities to use them. This fn could be written as `return value.removeEnd(';');`

src/classes/Accounts.cls
```diff
@@ -0,0 +1,219 @@
...
+        this(records, oldlist, null, null);
+    }
+
+    @TestVisible
+    private Accounts(List<Account> records, List<Account> oldlist, HouseholdNamingService
+            service, HouseholdSettings settings) {
+        super(records, Account.SObjectType);
+        this.oldMap = oldlist == null ? new Map<Id, Account>() : new Map<Id, Account>
+                (oldlist);
+        this.householdNamingService = service == null ? new HouseholdNamingService() : service;
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yep, this is a good call.  Nice to avoid the unnecessary constructor if we can.  That being said, I just don't love having the null check if it can be avoided.  I just updated these tonight (before reading the comments on the PR) to have the no param constructor pass a new instance to the private constructor, thus eliminating the need to do the null check. 
 Then the test instantiation would be new `HousholdNamingService(new NamingServiceMock())`.  Let me know what you think of that and if you'd still prefer to go this route.  Eventually I'd like to use force-di to inject these collaborator class instances, which will make both of these techniques unnecessary!

src/classes/ContactsInLegacyHouseholds.cls
```diff
@@ -0,0 +1,594 @@
...
+        fireRollups();
+        updateHouseholdNames();
+        deleteOldEmptyHouseholds();
+    }
+
+    private void deleteOldEmptyHouseholds() {
+        delete getEmptyHouseholdsToDelete(contactsThatChangedHouseholds());
+    }
+
+    private static Integer timesRun = 0;
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yes, good eye - this is one thing I added during this refactor work.  This method runs recursively on merges and began hitting soql limits with the refactored work.  It is only used for this method and only to prevent rampant recursion in that merge case. I don't think its worth moving to another class, I think it fits best right here where it is the only place it is used.  I could rename it, but I think when you look at its usage in the method right below it is pretty clear.  But happy to rename if you feel strongly and have a name suggestion! :)

src/classes/HouseholdCustomNamingString.cls
kenlewis-sfdo: Renamed this today to HouseholdNamingOverrides.  Trying to find a good name for it, open to other suggestions.  In DDD value objects are immutable instances that provide a singular value.  Thats what this is!  The value is generated from the passed in objects, but once it is constructed, it is just an object representing a value - in this case it is the string that represents the delimited set of name/greeting fields that are overridden (and is then stored in the confusing system_custom_naming field).

src/classes/HouseholdCustomNamingString.cls
kenlewis-sfdo: I'm open to discussing this, but I think NPSP should lean more toward exposing everything on its classes via methods, and not have a combo of fields/methods exposed.  It keeps the apis cleaner and hides the internals better - callers don't need to know if something is a field or a calculated value, and everything being exposed as a method protects that internal implementation detail.

src/classes/HouseholdCustomNamingString.cls
kenlewis-sfdo: I don't think so...  :-\  I don't like the looks of these either, but I was really careful and didn't change anything around these (other than in one place the string was being initialized as '' and in the other as ';', which is what I settled on here.  Its been in place like this for a long time... :)

src/classes/HouseholdName.cls
kenlewis-sfdo: Yes, and could be named as such.  Open to either.  But I kinda like that this way it is:
HouseholdName().value();
HouseholdName().asInformalGreeting();
HouseholdName().asFormalGreeting();
Its a value object, so really the value() method is the main one, and the other two are "household name (value) for specific contexts" 
Open to changing this if you think it makes more sense another way

src/classes/HouseholdNamingService.cls
kenlewis-sfdo: Yep - I think I did this when trying to get to the bottom of the issue with the merge recursion that led to the introduction of that "timesRun" flag.  I'll take these vars back out.

src/classes/Accounts.cls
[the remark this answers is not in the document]
kenlewis-sfdo: You're right, its way better to remove the private constructors.  I'll make this change

src/classes/ContactsInLegacyHouseholds.cls
```diff
@@ -0,0 +1,594 @@
...
+                    memberCountFrom(aggregateResult));
+        }
+        for (Id hhid : getOldHouseholdIds()) {
+            if (!oldHouseholdSizeById.containskey(hhid))
+                oldHouseholdSizeById.put(hhid, 0);
+        }
+        return oldHouseholdSizeById;
+    }
+
+    private Integer memberCountFrom(AggregateResult a) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: I've tried a few times now to get this factored out into a class but I keep getting halfway and then just feeling funny about it.  But I hear what you are saying and will try again.  If I can't get this into a good abstraction today then it might be helpful to jump on a call and refactor it together!

src/classes/HouseholdNamingExclusionsString.cls
kenlewis-sfdo: This is a great point and a good catch!  This is intended to be a value object but the fact that the Contacts class is using it in this way and managing a lot of the behavior that the Exclusions object should be doing itself is a sign of feature envy. :) I'm going to take a crack at moving that behavior further into this class so that the oddness of its variable name and usages will be removed from the Contacts class - stay tuned!  Would love to see what you think afterwards if I can make it look good.  :)

src/classes/Contacts.cls
```diff
@@ -0,0 +1,266 @@
...
+            exclusions.excludeFormalGreeting();
+        }
+        if (isExcludedFromInformalGreeting(con)) {
+            exclusions.excludeInformalGreeting();
+        }
+        return exclusions;
+    }
+
+
+    public void onAfterUpdate() {
```
force2b, line 181: Let's keep the "on" methods all at the top of the class.

src/classes/ContactsInLegacyHouseholds.cls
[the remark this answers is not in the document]
force2b: Thanks for the explanation. For now, maybe rename to `updateHouseholdNamesCounter` and add a comment block above to explain why it's needed to prevent gov limit errors due to recursion during merges.

src/classes/ContactsInLegacyHouseholds.cls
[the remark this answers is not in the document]
force2b: Fair enough. It's easy for me to read the code in review and say refactor, but reality on the ground (or in the ide) may be a bit different. I think you're out next week, but we can talk about this more when you're back - maybe even a little screen share coding - to see if there's a way to pull it off.

src/classes/HH_INamingImpl.cls
```diff
@@ -31,9 +31,9 @@
 * @author Salesforce.org
 * @date 2014
 * @group Households
-* @description HH_NameSpec class that implements template based naming for Households.
+* @description NPSPs default implementation of the HH_INaming interface used to generate household names.
 */
-public with sharing class HH_NameSpec implements HH_INaming {
+public with sharing class HH_INamingImpl implements HH_INaming {
```
force2b, line 36: I do like the "impl" naming structure (Java-ish maybe), but it hasn't fit our general naming pattern. But, it is an implementation of an interface so it makes sense from that perspective. It's probably a good practice for us to use going forward. For example, the 3 address verification classes could also follow this pattern.

src/classes/HouseholdName.cls
force2b: I see your point, but I also think of it as there are 3 different naming formats - standard name, formal, and informal. So it makes more sense to me to have all methods following a similar naming pattern.

src/classes/HouseholdNamingOverrides.cls
```diff
@@ -0,0 +1,185 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Value object for the String used to signify which Household Name / Greeting fields are overridden.
+*/
+public inherited sharing class HouseholdNamingOverrides {
```
force2b, line 35: I'm honestly not sure what this class does. What is an override for the naming? I thought there were just the 3 different types of names and some exclusionary settings.

src/classes/HouseholdNamingService.cls
force2b: This still says 2011 while the @date below has 2021.

force2b: If this entire class is just a rename of a legacy class, then both should probably be 2011.

src/classes/Households.cls
```diff
@@ -0,0 +1,106 @@
...
+
+    private void setCustomNamingField() {
+        householdNamingService.setCustomNamingField(getRecords(), oldMap);
+    }
+
+    private Boolean isAdvancedHouseholdNamingEnabled() {
+        return settings.isAdvancedHouseholdNaming();
+    }
+
+    public void onAfterUpdate() {
```
force2b, line 84: All the "on" methods should be at the top of the class

force2b (commented):
Getting better and better every time! Great work!!

force2b:
@davidmreed and @codexmarcel I want to call out the large number of apex class renames in this PR. This will merge into 232, which will eventually merge into master. We'll definitely want to allow for additional time post merge to resolve potential issues in the packaging org in case some of the classes are not deleted/deprecated.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,216 @@
...
+
+    private Boolean orgUsesLegacyHouseholds() {
+        return settings.isAllProcessor() || settings.isAllIndividualsProcessor();
+    }
+
+    private Id householdIdFor(Contact con) {
+        return con.npo02__Household__c;
+    }
+
+    public void synchronizeNamingExclusions() {
```
force2b, line 111: This method probably needs a block comment to explain what it's doing and why. The same true for `mergeExclusions` and some other related ones.

src/classes/HouseholdNamingExclusions.cls
```diff
@@ -30,61 +30,59 @@
 /**
 * @author Salesforce.org
 * @date 2021
-* @description Domain class for HouseholdNamingExclusions in NPSP.
+* @description Represents household naming exclusions as a set of boolean values for Contacts in NPSP.
 */
-public inherited sharing class HouseholdNamingExclusions {
-    private String value;
+public with sharing class HouseholdNamingExclusions implements INamingExclusions {
```
force2b, line 35: Why "With" sharing?

src/classes/Contacts.cls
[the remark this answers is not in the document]
kenlewis-sfdo: After a lot of back and forth, I moved this into the Households domain class and then replaced the Aggregate query.  It was being performed twice and had this manual augmentation to insert 0s for empty households since they didn't come back in the agg result.  I think the new change is much cleaner, which uses a subquery when the Households domain class is instantiated.  Let me know what you think!

see `Households.getHouseholdsById`

src/classes/ContactsInLegacyHouseholds.cls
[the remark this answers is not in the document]
kenlewis-sfdo: I'm going to resolve this convo for now - please see the other comment about how I removed the Agg query altogether.

src/classes/ContactsInLegacyHouseholds.cls
```diff
@@ -0,0 +1,594 @@
...
+    private void calculateRollups(Set<Id> householdIds) {
+        if (rollupsShouldRunSynchronously(householdIds)) {
+            RLLP_OppRollup rg = new RLLP_OppRollup();
+            rg.rollupHouseholds(getHouseholdMap(householdIds));
+        } else {
+            RLLP_OppRollup.rollupHouseholdsFuture(householdIds);
+        }
+    }
+
+    private Map<Id, npo02__Household__c> getHouseholdMap(Set<Id> householdIds) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: Moved these into the Households class!

src/classes/HouseholdCustomNamingString.cls
kenlewis-sfdo: Renamed again to `UserControlledHouseholdNamingFields`. I think we are honing in on an understandable name for this class, and just trying to figure out its name has provided a log of clarity (for me anyway :) on its purpose.  

Also renamed the value var to 'asConcatenatedString'.

The usage looks quite a bit better now I think in `HouseholdNamingService.setCustomNamingStringValue`

src/classes/HouseholdNamingExclusionsString.cls
kenlewis-sfdo: See the usage in `Contacts.synchronizeNamingExclusions` now after recent changes and see if it looks ok.  Is is now `setExclusionsStringFor(con, exclusionsAsString.value());` which feels clear to me.

src/classes/LD_LeadConvertOverride_CTRL.cls
```diff
@@ -425,14 +425,14 @@ public with sharing class LD_LeadConvertOverride_CTRL {
 
             // fix the account that lead conversion created
             if (strConId == 'NEW_CONTACT' && isIndividualAccount) {
-                string strSoql = 'select id, firstname, lastname, accountId,';
+                string strSoql = 'SELECT Id, Firstname, Lastname, AccountId,';
```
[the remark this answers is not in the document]
kenlewis-sfdo: Cleaned up this whole file!

src/classes/HouseholdNamingExclusions.cls
kenlewis-sfdo: just defaulted.  updated!

davidmreed:
Can we confirm that all of the deleted Apex classes (rename is a delete followed by a deploy) have not been used as the controller in a Visualforce page, Lightning component, or Visualforce component? If that is the case, the two-step process is required.

force2b:
@davidmreed  Thank you. I can confirm that none of them would have been referenced by a VF page itself and none contained @AuraEnabled methods.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,230 @@
...
+    public void onAfterInsert() {
+        if (orgUsesLegacyHouseholds()) {
+            ContactsInLegacyHouseholds contactsInLegacyHouseholds =
+                    new ContactsInLegacyHouseholds(getRecords());
+            contactsInLegacyHouseholds.onAfterInsert();
+        }
+    }
+
+    public void onAfterUpdate() {
+        new ContactsInLegacyHouseholds(getRecords(), oldMap)
```
force2b, line 76: The `onAfterInsert` method first checks to see if `orgUsesLegacyHouseholds()` before calling the `ContactsInLegacyHouseholds` class. Why isn't that needed here?

force2b (commented):
I definitely like where all this is heading. One common thing I'm observing is that the new classes do not appear to be the same in structure. For consistency, the structure of the class should be along the lines of:

- static vars
- public instance vars
- private instance vars
- getter/setters
- constructors
- "on" methods
- public methods
- private [helper] methods.
- any static methods

Mixing the public and private is OK if they're grouped together, and ideally marked with a comment or `/* ==== */`  type block to show the logical grouping by functionality.

src/classes/Contacts.cls
```diff
@@ -0,0 +1,216 @@
...
+        this(records, null);
+    }
+
+    public Contacts(List<Contact> records, List<Contact> oldList) {
+        super(records, Contact.SObjectType);
+        this.oldMap = oldList != null ?
+                new Map<Id, Contact>(oldList) : new Map<Id, Contact>();
+    }
+
+    public void onBeforeInsert() {
```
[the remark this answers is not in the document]
kenlewis-sfdo: I think these are at the top, just below the constructors.  Is that ok or do you want to move them above the constructors?

src/classes/Contacts.cls
kenlewis-sfdo: This is leftover after first putting everything in Contacts and then extracting ContactsInLegacyHouseholds.  I removed the pass-throughs which should make it clearer.  The checks against settings are maybe extracted but not net new.  Now that we have both of these classes, there is some logic that concerns all Contacts, and some that only concerns (legacy) Contacts that use Households.

src/classes/HouseholdNamingExclusionsCheckboxes.cls
```diff
@@ -0,0 +1,88 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Represents household naming exclusions as a set of boolean values for Contacts in NPSP.
+*/
+public with sharing class HouseholdNamingExclusionsCheckboxes implements INamingExclusions {
```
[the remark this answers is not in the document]
kenlewis-sfdo: This should be fixed!

src/classes/HouseholdNamingService.cls
kenlewis-sfdo: This class is renamed from HH_HouseholdNaming - the sharing declaration is unchanged.  There is a public method in this class that executes the batch to refresh all household naming.  Whatever in here needs to truly be without should probably be factored out into its own smaller class.  But would need to do some more investigation to figure out what really needs to have without sharing.

kenlewis-sfdo: To add to this comment, this class is going to need some further refactoring for sure. It is in an okay place for now but besides the "without sharing" logic being factored out into its own class, this class queries household members and have some other stuff that should probably be pushed into the domain classes instead of here in the service class. I'm going to leave it for now in the interest of stabilizing this branch to finalize review and testing.

src/classes/Accounts.cls
```diff
@@ -79,16 +79,64 @@ public inherited sharing class Accounts extends fflib_SObjects {
...
+                resetAccountTypeAndIndividualFields(acc);
+            }
+        }
+    }
+
+    /******************************************************************************************************
+    * @description utility to return the recordtypeId for a new account, given the account model.
+    * @return Id. if none is explicitly specified, returns null.
+    */
+    public static Id rtIdForNewAccount() {
```
force2b, line 108: I might have called this out elsewhere, but I've got 2 questions on these 2 methods:

- Do they need to be public?
- Do the need to be static?

If yes to both, I would make the `rtIdForAccountModel` method into a getter that saves the data in a static var so it doesn't need to run though the describe and subsequent logic on each call to the method.

src/classes/Households.cls
```diff
@@ -0,0 +1,201 @@
...
+            HouseholdNamingService householdNamingService = new HouseholdNamingService();
+            householdNamingService.updateHouseholdNameAndMemberCount(new List<Id>(ids()));
+        }
+    }
+
+    public Set<Id> ids() {
+        return getIdFieldValues(npo02__Household__c.Id);
+    }
+
+    private static List<npo02__Household__c> getHouseholdsById(Set<Id> ids) {
```
force2b, line 145: Does this need to be static?

force2b (commented):
Really looking great. Just a couple of comments.

src/classes/Accounts.cls
kenlewis-sfdo: Yeah, this is a bit of an odd method - I moved these from `ACCT_IndividualAccounts_TDTM.cls` because it felt like they fit better here.  But I'm hesitant to change this further now - they are used in a few classes, and there is a bit of a larger refactor here that might come into play in the Address Management refactor WI coming up.  Classes are using these to set the RT on a new Account, but the Accounts class should probably have a method that that does that instead.  Let me know if you want me to dig into this further and try to start that refactor here.

src/classes/Households.cls
kenlewis-sfdo: In this configuration, yes, because it is called inside one of the constructors.  I've been hesitant to introduce Selector classes yet because I'd really like to have the full AEP library before doing that, but I introduced the very beginnings of a HouseholdSelector class here to avoid this, which should also help with future tests for this class.  Let me know what you think.

force2b: Can the constructor not call an instance method?

src/classes/HouseholdSelector.cls
```diff
@@ -0,0 +1,48 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Selector class for Households in NPSP.
+*/
+public with sharing class HouseholdSelector {
```
force2b, line 35: @kenlewis-sfdo Can you make this `inherited sharing`?

src/classes/HouseholdService.cls
```diff
@@ -0,0 +1,266 @@
...
+    private Boolean isNotExcludedByRecordType(Contact con) {
+        return !isRecordTypesOnContacts()
+                || !settings.contactRecordTypesToExclude().contains(recordTypeIdFor(con));
+    }
+
+    private Id recordTypeIdFor(Contact con) {
+        return (Id) con.get(RECORD_TYPE_ID);
+    }
+
+    private Boolean isRecordTypesOnContacts() {
```
force2b, line 262: @kenlewis-sfdo Calling getMap() more than once is going to be a cpu expensive operation. I recommend an alternative approach using a static var so that the evaluation of whether the field exists only happens once. You could also compare the cpu time of calls to getMap() vs. just wrapping a call to `get('RecordTypeId')` in a try/catch block.

src/classes/HouseholdSelector.cls
kenlewis-sfdo: Yep! will push this up shortly.

src/classes/HouseholdService.cls
```diff
@@ -259,8 +259,13 @@ public inherited sharing class HouseholdService {
         return (Id) con.get(RECORD_TYPE_ID);
     }
 
+    private Boolean isRecordTypeInUseOnContacts;
```
force2b, line 262: @kenlewis-sfdo  I recommend using a static var so it's retained across instances of the class.

PR #6241, opened 2021-02-23
W-8157839 & W-8320388 - Fix DryRun and Existing RD bugs for RDInBDI

callen-sfdo:
Fixes two issues found during testing for Recurring Donations in BDI:
W-8320388 - Fixes an issue where running BDI with the dry run = true would result in a null pointer exception.  This was due to the return for dry run mode not being in the correct place.
W-8157839 - Fixes issue where putting an existing Recurring Donation Id in the recurring donation imported field would not link the new donation to the exsiting RD.  This was due to the matching code that set the Recurring Donation Import Status to 'Matched' was incorrect and therefore the Donation code did not identify it as being valid matched parent record.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


grivero:
Changes looks good but unit tests are failing, let me know once you go over them and I can review again

grivero (approved):
Looks good

PR #6249, opened 2021-02-24
Fix Contact delete behavior in LEX when delete override is enabled

### LINKED ISSUE #6128: Deleting a Contact in Lightning Displays an Error but is Successfully Deleted
When a customer deletes a Contact record and it's not the last Contact in a Household, the Contact is successfully deleted, but an error message is displayed:

_"Looks like there's a problem.
Unfortunately, there was a problem. Please try again. If the problem continues, get in touch with your administrator with the error ID shown here and any other related details. The requested resource does not exist"_

Screen Shot 2020-12-16 at 4 42 07 PM

The URL is the deleted Contact's URL, in Lightning form:
_https://flow-momentum-43332-dev-ed.lightning.force.com/lightning/r/Contact/0031100001jy84YAAQ/view_

### Root Cause
Something seems to have changed (recently?) with the params that are captured from the Visualforce page. We include an override on the Contact Delete button that points to our [CON_DeleteOverride Visualforce page](https://github.com/SalesforceFoundation/NPSP/blob/master/src/pages/CON_DeleteContactOverride.page), which has an Apex Controller of [CON_DeleteContactOverride_CTRL](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CON_DeleteContactOverride_CTRL.cls).

On the Visualforce page, we render an outputPanel if the sforceNav variable is not null. If the sforceNav variable === 'back', we call sforce.one.back(true). Otherwise, we navigate to the Contact tab.
```
 <apex:outputPanel rendered="{!sforceNav != null}" id="navigator">
            <script type="text/javascript">
                (function() {
                    var sforceNav = '{!sforceNav}';
                    if (sforceNav === 'back') {
                        sforce.one.back(true);
                    } else {
                        sforce.one.navigateToURL('{!URLFOR($Action.Contact.Tab, $ObjectType.Contact)}', true);
                    }
                })();
            </script>
        </apex:outputPanel>
```

In this case, the code is incorrectly setting sforceNav to "back" within the controller's [deleteContact() method.](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CON_DeleteContactOverride_CTRL.cls#L214-L219). In the past, the retURL variable would have contained the Contact ID, so we would set its value to "home" instead of "back". But, further up in the code in the [setNavigationParams() method](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CON_DeleteContactOverride_CTRL.cls#L145-L156), we set retURL based on the vfRetURLInSFX parameter returned from the VF page. However, that param is not present anymore. [I'll list out the params that _are_ present in the Work Item.

```
 private void setNavigationParams(Map<String, String> params) {
        // In LEx and SF1 the returl param name differs from classic
        String param = sforceUser ? 'vfRetURLInSFX' : 'retURL';
        retURL = params.get(param);

        //if we don't have a return URL, go back to Contacts home.
        if (String.isBlank(retURL)) {
            retURL = CONTACT_HOME;
        }

        failRetUrl = params.get('failRetURL');
    }
```

**Relevant Debug Lines:**
10:28:07.0 (7263834)|VARIABLE_ASSIGNMENT|[147]|param|"vfRetURLInSFX"
10:28:07.0 (7270069)|STATEMENT_EXECUTE|[148]
10:28:07.0 (7292166)|SYSTEM_METHOD_ENTRY|[148]|Map<String,String>.get(Object)
10:28:07.0 (7319126)|SYSTEM_METHOD_EXIT|[148]|Map<String,String>.get(Object)
10:28:07.0 (7332140)|VARIABLE_ASSIGNMENT|[148]|this.retURL|null|0x4a355c58
10:28:07.0 (7351068)|SYSTEM_METHOD_ENTRY|[151]|String.isBlank(String)
10:28:07.0 (7370337)|SYSTEM_METHOD_EXIT|[151]|String.isBlank(String)
10:28:07.0 (7375695)|STATEMENT_EXECUTE|[151]
10:28:07.0 (7377708)|STATEMENT_EXECUTE|[152]
10:28:07.0 (7388935)|VARIABLE_ASSIGNMENT|[152]|this.retURL|"/003/o"|0x4a355c58

### Steps to Reproduce
1. Create a Contact (First Name = Jess; LastName = Lopez);
2. Create another Contact associated with the Lopez Household Account.
3. Delete the second Contact by using the Delete button on the Contact Detail page.
4. An error appears.

In an org with unmanaged code, I modified the deleteContact() method and set sforceNav explicitly to LEX_HOME (assuming this is what it was defaulting to prior to a recent update). The page then redirected properly to the Contacts tab and did not throw an error.
```
//sforceNav = retURL.contains(queryContact.Id) ? LEX_HOME : LEX_BACK;
sforceNav = LEX_HOME;
```

bdvorachek:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008lg4cIAA/view

# Critical Changes

# Changes

# Issues Closed
Closes #6128 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CON_DeleteContactOverride_CTRL.cls
+++ b/src/classes/CON_DeleteContactOverride_CTRL.cls
@@ -60,11 +60,11 @@ public with sharing class CON_DeleteContactOverride_CTRL {
 
     /** @description Boolean that is set to true if user theme is LEx or SF1 */
     @TestVisible 
-    private Boolean sforceUser {
+    private Boolean lexOrSF1User {
         get {
-            return sforceUser = sforceUser == null 
+            return lexOrSF1User = lexOrSF1User == null
                 ? UserInfo.getUiThemeDisplayed().contains('Theme4')
-                : sforceUser;
+                : lexOrSF1User;
         }
         set;
     }
@@ -143,9 +143,7 @@ public with sharing class CON_DeleteContactOverride_CTRL {
     */
     @TestVisible
     private void setNavigationParams(Map<String, String> params) {
-        // In LEx and SF1 the returl param name differs from classic
-        String param = sforceUser ? 'vfRetURLInSFX' : 'retURL';
-        retURL = params.get(param);
+        retURL = params.get('retURL');
 
         //if we don't have a return URL, go back to Contacts home.
         if (String.isBlank(retURL)) {
@@ -211,8 +209,9 @@ public with sharing class CON_DeleteContactOverride_CTRL {
 
             delete queryContact;
 
-            if (sforceUser) {
-                sforceNav = retURL.contains(queryContact.Id) ? LEX_HOME : LEX_BACK;
+
+            if (lexOrSF1User) {
+                sforceNav = LEX_HOME;
                 return null;
             } else {
                 return getRedirect();
@@ -251,7 +250,7 @@ public with sharing class CON_DeleteContactOverride_CTRL {
             return null;
         }
 
-        if (sforceUser) {
+        if (lexOrSF1User) {
             sforceNav = LEX_HOME;
             return null;
 
@@ -270,7 +269,7 @@ public with sharing class CON_DeleteContactOverride_CTRL {
     * @return PageReference 
     */ 
     public PageReference onCancel() {
-        if (sforceUser) {
+        if (lexOrSF1User) {
             sforceNav = LEX_BACK;
             return null;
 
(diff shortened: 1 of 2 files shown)


sokosaurus (commented):
Code looks good to me. Given that I'm new to this code base, I'm holding off on designating an actual approval to give the other reviewers an opportunity to review.

One thing - I think the work item referenced in your PR description should be [this](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008lg4cIAA/view) instead of [this](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000096OPhIAM/view).

psadie-sfdo (approved):
Looks good to me!

voduyemi (approved):
Changes look good to me! Thanks! @bdvorachek

PR #6251, opened 2021-02-24
Add error icon to batch table

dfuller114:
Error icons will now appear for each batch table row in Batch Gift Entry if there is a BDI or payment error on the data import record corresponding to the table row.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BGE_BatchGiftEntry_UTIL.cls
+++ b/src/classes/BGE_BatchGiftEntry_UTIL.cls
@@ -94,6 +94,8 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
                 fieldApiNames.add(field.toLowerCase());
             }
 
+            fieldApiNames.addAll(getDataImportElevateFields());
+
             // Must explicitly use BDI_MappingServiceAdvanced here (as opposed to
             // BDI_DataImportService.getDefaultMappingService()) because of the need
             // to reference the fieldMappingByDevName property.
@@ -161,6 +163,18 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
             }
         }
         return fieldApiNames;
+    }
+
+    private static List<String> getDataImportElevateFields() {
+        List<String> elevateFieldApiNames = new List<String>();
+        if (GE_PaymentServices.isElevateCustomer) {
+            elevateFieldApiNames = new List<String>{
+                    DataImport__c.Payment_Declined_Reason__c.getDescribe().getName(),
+
+            };
+        }
+
+        return elevateFieldApiNames;
     }
 
     /*******************************************************************************************************
--- a/src/classes/BGE_BatchGiftEntry_UTIL.cls
+++ b/src/classes/BGE_BatchGiftEntry_UTIL.cls
@@ -73,89 +73,84 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
         // Batches created using BGE_DataImportBatchEntry store fields used by its Data
         // Import records in Active_Fields__c, while Batches created in GE_GiftEntry
         // use fields included in the Form_Template__c record related to the Batch.
-        DataImportBatch__c batch = [
-                SELECT Batch_Gift_Entry_Version__c,
-                        Form_Template__r.Template_JSON__c
-                FROM DataImportBatch__c
-                WHERE Id = :batchId
-        ];
+        DataImportBatch__c batch = getDataImportBatch(batchId);
+        List<String> dataImportCoreFields = getCoreDataImportFields(includeRelationshipFields);
 
-        if (batch.Batch_Gift_Entry_Version__c == null ||
-                batch.Batch_Gift_Entry_Version__c < 2.0) {
-
-            fieldApiNames = getCoreDataImportFields(includeRelationshipFields);
+        if (!isNewBatchGiftEntryVersion(batch)) {
+            fieldApiNames.addAll(dataImportCoreFields);
             fieldApiNames.addAll(getActiveFieldNamesFromBatch(batchId));
 
-        } else if (batch.Batch_Gift_Entry_Version__c >= 2.0) {
+            return fieldApiNames;
+        }
 
-            fieldApiNames.add(DataImport__c.Additional_Object_JSON__c.getDescribe().getName());
+        fieldApiNames.addAll(getDataImportElevateFields());
+        fieldApiNames.add(getNewBatchGiftEntryVersionFields());
 
-            for (String field : getCoreDataImportFields(includeRelationshipFields)) {
-                fieldApiNames.add(field.toLowerCase());
-            }
+        for (String field : dataImportCoreFields) {
+            fieldApiNames.add(field.toLowerCase());
+        }
 
-            fieldApiNames.addAll(getDataImportElevateFields());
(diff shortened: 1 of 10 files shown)


src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -73,87 +73,84 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
+                    // Add the source fields used in the Template but only if the source field name is valid
+                    if (!fieldApiNames.contains(sourceFieldName.toLowerCase()) && field != null) {
+                        fieldApiNames.add(sourceFieldName.toLowerCase());
+                    }
 
-                        BDI_FieldMapping fieldMapping =
-                                fieldMappingByDevName.get(fieldMappingName);
+                    // Add the Name fields on the related object(s) for all Lookup fields,
+                    // used as display value(s) for Lookup fields loaded in the Form UI
+                    if ((fieldMapping != null &&
```
voduyemi, line 142: I think we can extract the lines of code that populate `relatedObjectNameField` into two methods, `isRelatedObjectNameField` which returns a `boolean` and `getRelatedObjectNameField` that returns a `String`

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -73,87 +73,84 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
-            for (GE_Template.Section formSection : template.layout.sections) {
-                for (GE_Template.Element formElement : formSection.elements) {
-                    if (formElement.dataImportFieldMappingDevNames == null) {
+                    if (fieldMapping == null && !isObjectMappedLookup) {
                         continue;
                     }
-                    // Get the field mappings, used to reference the source fields
-                    for (String fieldMappingName :
-                            formElement.dataImportFieldMappingDevNames) {
+                    String sourceFieldName = fieldMapping != null ?
```
voduyemi, line 128: This can also be extracted into its own method `getSourceFieldName`

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -8,7 +8,7 @@ import getDataImportRows from '@salesforce/apex/BGE_DataImportBatchEntry_CTRL.ge
 import saveAndDryRunDataImport from '@salesforce/apex/GE_GiftEntryController.saveAndDryRunDataImport';
 
 import { handleError } from 'c/utilTemplateBuilder';
-import { isNotEmpty, isUndefined } from 'c/utilCommon';
+import {isNull, isNotEmpty, isUndefined, isEmptyObject, apiNameFor} from 'c/utilCommon';
```
voduyemi, line 11: Can we remove `isNull` and `isEmptyObject` since they are unused?

voduyemi (changes requested):
Thanks @dfuller114 ! Looks really good, had some comments around the `getDataImportFields` refactor and some general clean-up. Let me know if would like to chat about the refactor.

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -163,10 +160,48 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
+        );
+        return template;
+    }
+
+    private static String getNewBatchGiftEntryVersionFields() {
+        return DataImport__c.Additional_Object_JSON__c.getDescribe().getName();
+    }
+
+    private static Boolean isNewBatchGiftEntryVersion(DataImportBatch__c batch) {
+        return batch.Batch_Gift_Entry_Version__c >= 2.0;
```
sokosaurus, line 176: This is something realllyyyyy minor...

Can you imagine a future scenario where would ever need to increment this number value? If so - perhaps we could consider moving it to a constant at the header of the class, or somewhere else.

sokosaurus (commented):
Thanks for letting me review this PR. Looks like you took an opportunity to do some refactoring in here, which is great. 

Being new to stubbing/mocking, it was also really insightful to see how you're able to leverage those frameworks.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -103,6 +100,61 @@ export default class GeBatchGiftEntryTable extends LightningElement {
...
+        this._data = dataImportRows;
+    }
+
+    tableRowErrors;
+    assignDataImportErrorsToTableRows(dataImportRows) {
+        let errors = {rows: {}};
+        this.getDataImportRowsWithErrors(dataImportRows).forEach(row => {
+            Object.assign(errors.rows,  {
+                [row.Id] : {
+                    title: 'Process Error',
```
voduyemi, line 122: Can we add the custom label for this hardcoded string?

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -163,10 +160,48 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
+        ];
+        return batch;
+    }
+
+    @TestVisible
+    private static List<String> getDataImportElevateFields() {
+        List<String> elevateFieldApiNames = new List<String>();
+        if (GE_PaymentServices.getInstance().isElevateCustomer()) {
+            elevateFieldApiNames = new List<String>{
+                DataImport__c.Payment_Declined_Reason__c.getDescribe().getName()
```
kenlewis-sfdo, line 194: This can be `String.valueOf(DataImport__c.Payment_Declined_Reason__c)`.  No need to get the whole describe.

src/classes/BGE_BatchGiftEntry_UTIL.cls
```diff
@@ -73,87 +73,84 @@ public with sharing class BGE_BatchGiftEntry_UTIL {
...
         // Import records in Active_Fields__c, while Batches created in GE_GiftEntry
         // use fields included in the Form_Template__c record related to the Batch.
-        DataImportBatch__c batch = [
-                SELECT Batch_Gift_Entry_Version__c,
-                        Form_Template__r.Template_JSON__c
-                FROM DataImportBatch__c
-                WHERE Id = :batchId
-        ];
+        DataImportBatch__c batch = getDataImportBatch(batchId);
+        List<String> dataImportCoreFields = getCoreDataImportFields(includeRelationshipFields);
```
kenlewis-sfdo, line 77: These extracted methods are looking nice!

src/classes/BGE_BatchGiftEntry_UTIL.cls
dfuller114: @sokosaurus This version number shouldn't ever change. Since it's only used in one place in the class we probably don't need a constant at this point I don't think.

src/classes/BGE_BatchGiftEntry_UTIL.cls
dfuller114: updated!

voduyemi (approved):
Looks good to me! Thanks @dfuller114

sokosaurus (approved):
Looks great. Approved!

PR #6268, opened 2021-03-04
W-8926650: Adding Latest Version Played field to Recurring Donation

lparrott:
Adding a 'Latest Version Played' field to Recurring Donation, which will store the Version number of the latest Event that caused an update. Will be used to prevent circular updates from Elevate/CCS.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
- Custom Field
    - npe03__Recurring_Donation__c.LatestElevateVersionPlayed__c

# Deleted Metadata


npsp-reedestockton (approved):
Looks fine to me!

npsp-reedestockton:
Adding @sam-knox and @mattwashington for label review.

sam-knox:
Yep - looks fine!

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -360,6 +360,20 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>LatestVersionPlayed__c</fullName>
+        <description>The last Elevate event version played resulting in an update to this Recurring Donation.</description>
```
force2b, line 365: I noticed that the description/helptext on the `LastElevateEventPlayed__c` field is really really close to this new field. Only the word "event" makes them different. I think it might avoid confusion (for us in the future as well as for customers) if there was a bit more in the description/helptext on these two fields to further clarify their usage and especially their differences.
CC @mattwashington

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -360,6 +360,20 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>LatestVersionPlayed__c</fullName>
```
force2b, line 364: Since the other field has an api name of `LastElevateEventPlayed__c` should this fields api name be `LastElevateVersionPlayed__c` for consistency?

force2b (changes requested):
We have some unpackaged metadata that gets deployed to customer orgs during RD2 enablement. Looking at that metadata, it's currently granting _read_ access to the `LastElevateEventPlayed__c` field. So, I think we should do the same for the `LatestVersionPlayed__c` field. The metadata is in the `unpackaged/config/rd2_post_config` folder. You should be able to follow the same pattern as the other field.

src/objects/npe03__Recurring_Donation__c.object
lparrott: Works for me!

src/objects/npe03__Recurring_Donation__c.object
```diff
@@ -360,6 +360,20 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>LastElevateVersionPlayed__c</fullName>
+        <description>An Integer that represents the last Elevate event version played resulting in an update to this Recurring Donation.</description>
```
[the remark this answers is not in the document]
sam-knox: I think we could work some of that description into the help text too. Want to take a crack at it @mattwashington?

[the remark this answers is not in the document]
lparrott: @mattwashington and I just talked through it, new commit was pushed!

force2b (approved):

PR #6271, opened 2021-03-08
Better Multi-currency support in Batch Gift Entry

### LINKED ISSUE #6132: Gift Entry - Currency Incorrectly Displaying and Converting Batch Totals
**Issue**
If an org has enabled multi-currency, the values of gifts entered, and totalled under "Total Batch Amount" does not correctly display, or convert, the amount entered in the batch.

As a result of the conversion not correctly correlating by:

1. The Currency defined via the Currency ISO Code field is not changing the Total Batch Amount's currency to match the entered gift's currency/expected batch currency
2. If the batch is set to "Require Expected Totals Match" then the incorrectly converted amount causes issues hindering the batch from being closed and processed

<br>

image

<br>
<br>

**Steps to Reproduce**

1. Enable Advanced Mapping and Gift Entry in Org
2. Go to Setup > Company Information > Edit and check Enable Multiple Currencies and Save
3. From the same page: Setup > Company Information > click "Currency Setup and add a new currency (such as GBP - British Pound) and add a conversion rate of 0.752
4. Update the currency used for running user in British Pounds on their user record
5. Goto Gift Entry
a.  Click on existing template
b.  Go to Batch Settings > Batch Header
c. Check Currency ISO Code
d.  Save
6. Create a new batch
a.  Pick the template you have updated
b.  Give it a batch name
c.  Currency ISO Code - currency you have added in currency setup
d.  Click Next(leave defaults)
e.  Save
7. Enter new gift info
8. Save and notice the "Gifts Entered for" section and compare Donation Amount and Total Batch Amount field shown in different currencies

**Workaround**
The gift values properly carry over to the Data Import Record once saved, and upon importing properly generate based on the User's currency defined on their User record; however, the Total Batch Amount value does not display properly, meaning that if a User has the "Require Expected Totals Match" option on the batch, they are not able to process it. There are two options for different scenarios:

1. Remove the "Require Expected Totals Match" from the Batch Info to allow for the proper processing of the Data Import Records
2. To properly calculate the Total Batch Amount, generate an "NPSP Data Imports with NPSP Data Import Batch" report that can be filtered by the batch and summarize the totals to verify

### LINKED ISSUE #6236: Gift Entry: Editing Batch Info Clears Batch Number field if it’s been added to a Page Layout
### Issue Summary:
If a customer has enabled the automatic Batch Number feature for Gift Entry, the value may be cleared from an NPSP Data Import Batch record when clicking the _Edit Batch Info_ button and saving. This seems to occur if the Batch Number field has been added to a Page Layout for the NPSP Data Import Batch object. 

### Root Cause
In the handleSave() method of geBatchWizard.js, we set the value of dataImportBatchCreateDefaults via the getRecordCreateDefaults LWC method. (https://developer.salesforce.com/docs/component-library/documentation/en/lwc/reference_wire_adapters_create_record_values). This returns object metadata associated with the default Page Layout. If a customer has added the Batch Number field (npsp__Batch_Number__c) to the page layout, we retrieve it and it is reset to a default value of null, despite being set to a value previously. 

### Steps to Repeat:
1. Enable Advanced Mapping and Gift Entry in NPSP Settings.
2. Save and Activate Batch Numbering (Display Format = BTC-{00000}; Starting Number = 1)
3. Go to Setup —> Object Manager —> NPSP Data Import Batch —> Page Layout —> NPSP Data Import Batch. Remove the Batch Number field if it’s currently on the Page Layout. Save.
4. Go to the Gift Entry Tab. 
5. On the Batches Tab, click the cog and Select Fields To Display. Add the Batch Number field. 
6. Click the New Batch button. 
    1. Choose the default template and click Next.
    2. Name = Test Batch 1. Click Next.
    3. Leave the defaults on the Set Default Values screen. Click Save. 
    4. Go back to the Gift Entry Tab and confirm that your batch has a Batch Number in the batches table (should be BTC-00001). 
    5. Click on the Batch again. 
    6. Click the Edit Batch Info button. Click Next and Save (no need to change anything). 
    7. Go back to the Gift Entry Tab and confirm that the Batch Number is still present and is still BTC-00001.
7. Now add the Batch Number field to the NPSP Data Import Batch Page Layout in Setup —> Object Manager —> NPSP Data Import Batch —> Page Layouts.
8. Go back to the Gift Entry Tab. 
9. Click on your Batch and click the Edit Batch Info button. Click Next and Save (no need to change anything). 
10. Go back to the Gift Entry Tab an confirm that your Batch Number has now vanished. 

### Workaround:
Remove the Batch Number field from your Page Layouts on the NPSP Data Import Batch object.

voduyemi:
WI: [W-8598455](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008lpYHIAY/view)

This PR fixes the issue with mismatched donation amount currencies in Batch Gift Entry. If a mismatch in batch currency and gift currency is detected, a warning message is displayed at the top of the form and entering a gift with a different currency than the batch is prevented
Custom Labels
- geErrorBatchGiftEntryCurrencyMismatch
- geWarningBatchGiftEntryCurrencyMismatch
# Critical Changes

# Changes

# Issues Closed
Fixes #6132 
Fixes #6236
# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -442,6 +453,12 @@ export default class GeBatchGiftEntryTable extends LightningElement {
             type: this.getColumnTypeFromFieldType(element.dataType)
         };
 
+        if (column.fieldName === DONATION_AMOUNT.fieldApiName) {
+            column.typeAttributes = {
+                currencyCode: this.batchCurrencyISOCode
+            };
+        }
+
```
dfuller114, line 461: Can we rename `fieldWrapper` used in this function as well as the buildColumnsFromSections() function to `fieldMapping` to make it more clear what is contained in the JSON object?

src/classes/GE_GiftEntryController.cls
```diff
@@ -173,6 +178,20 @@ public with sharing class GE_GiftEntryController {
...
 
+    /**
+     * @description Validates that the currency of a batch matches the currently logged in
+     * user's currency
+     * @param batchId : ID of the NPSP_Data_Import_Batch__c
+     */
+    private static void validateCurrencies (Id batchId) {
+        String batchCurrencyIsoCode = getBatchCurrencyIsoCode(batchId);
+        if (batchCurrencyIsoCode != UserInfo.getDefaultCurrency()) {
+           throw new CurrencyMismatchException(String.format(
```
dfuller114, line 189: I see that we are throwing an exception for mismtached ISO code here. Are we doing this as well as the warning message in the UI? Assumed that this exception would also override the warning message.

dfuller114 (changes requested):
just a few smaller comments. Refactor on dataImportModel is looking good!

src/classes/GE_GiftEntryController.cls
```diff
@@ -1035,5 +1054,21 @@ public with sharing class GE_GiftEntryController {
         }
     }
 
+    @AuraEnabled(Cacheable= true)
+    public static String getBatchCurrencyIsoCode (Id batchId) {
+        if (UserInfo.isMultiCurrencyOrganization()) {
+            String query = new UTIL_Query()
```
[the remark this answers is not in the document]
voduyemi: We switched to dynamic SOQL here to accommodate the quirkiness of referencing the `CurrencyIsoCode` field in single-currency orgs.

melissabarber:
@voduyemi @dfuller114 @sokosaurus the two custom labels in this PR look good to me!

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -342,20 +342,39 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
-            this.handleRecordCreate(recordObject);
+            this.handleRecordCreate(this.buildCreateRecord());
         }
     }
 
+    /**
+     * @description Builds a new data import batch record to be saved
+     * @returns {DataImportBatch record}
+     */
+    buildCreateRecord() {
```
kenlewis-sfdo, line 356: Are these changes related to the overwriting of batch defaults?  Is there a separate section in the test charter to test how these changes effect batch default values?

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -219,7 +229,7 @@ export default class GeBatchGiftEntryTable extends LightningElement {
     }
 
     deleteDIRow(rowToDelete) {
-        const isRowToDelete = row => row.Id == rowToDelete.Id;
+        const isRowToDelete = row => row.Id === rowToDelete.Id;
```
kenlewis-sfdo, line 232: Was there an issue that prompted this change?

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
```diff
@@ -125,6 +127,14 @@ export default class GeBatchGiftEntryTable extends LightningElement {
             .catch(error => handleError(error));
     }
 
+    retrieveBatchCurrencyIsoCode() {
+        retrieveBatchCurrencyIsoCode({batchId: this.batchId})
```
kenlewis-sfdo, line 131: Did you consider adding the batch ISO code to the data import model? Since the batch and its rows are being queried at initialization time already it might not be a bad idea to add this as an attribute to the model instead of a separate query.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -88,58 +88,63 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
         }
 
         DataImportModel dataImportModel = new DataImportModel();
-        Integer rowOffset = 0;
-
         dataImportModel.id = batchId;
-        dataImportModel.dataImportRows = getDataImportRows(batchId, rowOffset);
+        dataImportModel.dataImportRows = getDataImportRows(batchId, 0);
```
kenlewis-sfdo, line 92: How does this change effect infinite scrolling? Is the offset not needed anymore?

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -88,58 +88,63 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
...
         dataImportModel.setColumns();
         dataImportModel.labels = getBatchDataImportLabels();
         dataImportModel.isNamespaced = String.isNotBlank(UTIL_Namespace.getNamespace());
-        dataImportModel.totalCountOfRows = getDataImportBatchTotalCount(batchId);
+        dataImportModel = getDataImportBatchTotals(dataImportModel);
```
kenlewis-sfdo, line 96: I know we discussed maybe having an new separate method for the controller to use to get the batch total for this work item.  If we are going to go this route and modify the model that is used by current _and_ old BGE, then we'll have to make sure we test old BGE also.

kenlewis-sfdo (commented):
Looking good, Victor!  I left a few comments/questions.  Let me know if you'd like to discuss!

src/classes/BGE_DataImportBatchEntry_CTRL.cls
voduyemi: This shouldn't affect infinite scrolling since the offset was only set in a local variable `Integer offset = 0` in this method.

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
voduyemi: To be safe when I do equality comparison in javascript I use `===` to guard against any unexpected type coercion.

src/lwc/geBatchWizard/geBatchWizard.js
voduyemi: Yes, these changes are related to batch field values getting reset through the batch wizard. Yes, it is called out in the test charter.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
kenlewis-sfdo: Oh I see this one is only used for the initial call.👍🏽

src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.js
voduyemi: I decided to go the separate query route since we will be moving away from this data structure in the near future once we retire old BGE. I'm open to chatting about it though because I wrestled with the decision.

dfuller114 (approved):
Looks good to me! Great work Victor!

sokosaurus (approved):
There's a ton of really good stuff in here. Great work! Approved!

src/classes/BGE_DataImportBatchEntry_CTRL.cls
```diff
@@ -374,6 +407,7 @@ public with sharing class BGE_DataImportBatchEntry_CTRL {
         dataImportModel.totalCountOfRows = getDataImportBatchTotalCount(batchId);
         AggregateResult[] aggResult = getDataImportBatchTotalAmount(batchId);
         dataImportModel.totalRowAmount = (Decimal) aggResult[0].get('amount');
+        dataImportModel.batchTotalRowAmount = getBatchTotalRowAmount(batchId);
```
kenlewis-sfdo, line 410: Does this property need to be added onto the model here? It looks like this method might only be used by old BGE.

src/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -21,9 +21,9 @@ import {
     getNamespace,
     getNestedProperty,
     isNotEmpty,
-    isNull,
-    stripNamespace
-} from 'c/utilCommon'
+    isNull, isUndefined,
```
kenlewis-sfdo, line 24: Looks like isUndefined might be unused.

src/classes/BGE_DataImportBatchEntry_CTRL.cls
voduyemi: Yes it does, It's called  by `runSingleDryRun` which is we call every time we add a row to a batch in Gift Entry.

kenlewis-sfdo (approved):
Looks good to me Victor! Thanks for all the thoughtful work on this one. Much appreciated!

PR #6274, opened 2021-03-09
BGE Retirement Alert

sokosaurus:
Link to WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000096EDyIAM/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTabHelper.js
```diff
@@ -15,6 +15,15 @@
             }
         });
         $A.enqueueAction(action);
+
+        const modalViewedKey = 'hasViewedRetModal';
```
kenlewis-sfdo, line 19: Not a necessary change, but could consider naming this `hasViewedRetirementPrompt` just to stay consistent with the other references to the prompt

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
```diff
@@ -81,6 +82,24 @@
         </div>
     </header>
 
+    <aura:if isTrue="{!v.showRetirementPrompt}">
+        <c:utilPrompt 
+            aura:id="retirementPrompt" 
+            id="retirementPrompt"
+            headerTitle="Batch Gift Entry is retiring"
+            promptMessage="On July 18, 2021, legacy Batch Gift Entry will be retired and will no longer work in NPSP.
```
kenlewis-sfdo, line 90: Did we decide to go with a label for this to be translatable?  Or just have it be text?  I know there was a discussion on the WI but wanted to double check the decision there.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.css
```diff
@@ -1,3 +1,11 @@
 .THIS .slds-scrollable_y {
     min-height: 200px
+}
+
+#retirementPrompt.THIS {
+    position: absolute;
+    bottom: 0;
+    right: 0;
+    z-index: 1000;
+    max-width: 450px;
```
kenlewis-sfdo, line 10: I haven't deployed this and tested in the UI yet - but what is the behavior if the viewport is shrunk down significantly?  Does the prompt stay in the same place as the window size changes?

kenlewis-sfdo (commented):
Looking good @sokosaurus!  I left a few questions but nothing major.  Let me know if you'd like to discuss anything!

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
sokosaurus: I don't know if a decision was made but I think a label makes the most sense, even if the number of non-english customers is ~3. I'll get this PR updated with that.

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.css
sokosaurus: Good question. The prompt stays fixed to the bottom right as the viewport initially shrinks -- and then as the prompt nears the size of the viewport, it shifts to fill up the space.

**Full Viewport**

Screen Shot 2021-03-09 at 9 40 10 AM

**Minimized Viewport**

Screen Shot 2021-03-09 at 9 40 17 AM

Please let me know if you think any adjustments need to made here. Completely open to exploring other options if there is a better path forward.

kenlewis-sfdo: I think this looks good - thanks for verifying.  Lets double check with Lauren - maybe in our review mtg on Thursday - and have her confirm that this looks good also.  Thanks!

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
```diff
@@ -54,6 +54,7 @@
     <aura:attribute name="sortDirection" type="String" default="desc" access="private"/>
     <aura:attribute name="isLoadingDataTable" type="Boolean" default="false" access="private"/>
     <aura:attribute name="enableInfiniteLoading" type="Boolean" default="true" access="private"/>
+    <aura:attribute name="showRetirementPrompt" type="boolean" default="false" access="private"/>
```
bdvorachek, line 57:
```suggestion
    <aura:attribute name="showRetirementPrompt" type="Boolean" default="false" access="private"/>
```

Match case here.

sokosaurus:
Please see #6278

PR #6278, opened 2021-03-09
BGE Retirement Alert

sokosaurus:
Link to WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000096EDyIAM/view

# Critical Changes

# Changes
- **Note: this is about the legacy Batch Gift Entry feature and will not affect batches in Gift Entry or NPSP Data Importer (BDI).** On July 13, 2021, Batch Gift Entry will be retired. We added a warning on the Batch Gift Entry tab about the upcoming feature retirement. Learn more about the [retirement of Batch Gift Entry on the Power of Us Hub](https://powerofus.force.com/s/feed/0D51E00005acDV1SAM).

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
+++ b/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
@@ -54,7 +54,7 @@
     <aura:attribute name="sortDirection" type="String" default="desc" access="private"/>
     <aura:attribute name="isLoadingDataTable" type="Boolean" default="false" access="private"/>
     <aura:attribute name="enableInfiniteLoading" type="Boolean" default="true" access="private"/>
-    <aura:attribute name="showRetirementPrompt" type="boolean" default="false" access="private"/>
+    <aura:attribute name="showRetirementPrompt" type="Boolean" default="false" access="private"/>
 
     <!-- Start Markup -->
     <lightning:overlayLibrary aura:id="overlayLib"/>
--- a/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
+++ b/src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
@@ -54,6 +54,7 @@
     <aura:attribute name="sortDirection" type="String" default="desc" access="private"/>
     <aura:attribute name="isLoadingDataTable" type="Boolean" default="false" access="private"/>
     <aura:attribute name="enableInfiniteLoading" type="Boolean" default="true" access="private"/>
+    <aura:attribute name="showRetirementPrompt" type="boolean" default="false" access="private"/>
 
     <!-- Start Markup -->
     <lightning:overlayLibrary aura:id="overlayLib"/>
@@ -80,6 +81,19 @@
             </div>
         </div>
     </header>
+
+    <aura:if isTrue="{!v.showRetirementPrompt}">
+        <c:utilPrompt 
+            aura:id="retirementPrompt" 
+            id="retirementPrompt"
+            headerTitle="{!$Label.c.bgeRetirementWarningHeader}"
+            promptMessage="{!$Label.c.bgeRetirementWarningBody}"
+            confirmButtonTitle="Confirm" 
+            hasBackdrop="true" 
+            variant="error" 
+            onconfirm="{!c.handleConfirmClick}"
+        ></c:utilPrompt>
+    </aura:if>
 
     <!--Data Table-->
     <lightning:datatable aura:id="dataTable"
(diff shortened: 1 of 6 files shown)


sokosaurus:
Replaces #6274

src/aura/BGE_BatchGiftEntryTab/BGE_BatchGiftEntryTab.cmp
```diff
@@ -81,6 +82,19 @@
         </div>
     </header>
 
+    <aura:if isTrue="{!v.showRetirementPrompt}">
+        <c:utilPrompt 
+            aura:id="retirementPrompt" 
+            id="retirementPrompt"
+            headerTitle="{!$Label.c.bgeRetirementWarningHeader}"
+            promptMessage="{!$Label.c.bgeRetirementWarningBody}"
+            confirmButtonTitle="Confirm"
```
psadie-sfdo, line 91: Took a quick look and I'm surprised we don't have a standalone common label for 'Confirm'. I think 'Confirm' is worthy as a common label `commonConfirm`.

Custom labels we prefix with 'common' are expected to used across NPSP and are not specific to any given feature. What do you think?

@melissabarber - additional thoughts?

bdvorachek, line 91: I'm surprised we don't have a label for the word "Confirm" already. One thing to note is that once BGE is retired, we can deprecate these labels. To that end, should we label this one up too?

psadie-sfdo (commented):
One more thing actually

src/labels/CustomLabels.labels
```diff
@@ -5093,6 +5093,22 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>Button to process a batch.</shortDescription>
         <value>Process Batch</value>
     </labels>
+    <labels>
+        <fullName>bgeRetirementWarningBody</fullName>
+        <categories>Batch Data Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Explanation of Batch Gift Entry retirement &amp; recommendations</shortDescription>
+        <value>On July 13, 2021, legacy Batch Gift Entry will be retired and will no longer work in NPSP. If you have any in-progress batches at that time, you can still access those records, but you can't process them using Batch Gift Entry. To ensure a smooth transition, enable and use the newer Gift Entry in advance of July 13th. Learn more about Gift Entry in the &lt;a href=&apos;https://powerofus.force.com/NPSP_Documentation&apos; target=&apos;_blank&apos;&gt;NPSP Documentation&lt;/a&gt; on the Power of Us Hub.</value>
```
ptewson-sfdo, line 5102: Does the markup within this label cause trouble for anything that attempts to block XSS attacks here? Do we have an improved system for including markup w/in labels? (@psadie-sfdo I feel like you know about this?)

melissabarber:
@sokosaurus & @ptewson-sfdo reviewed the labels, and they're good to go (after you sort out the linking syntax).

src/labels/CustomLabels.labels
psadie-sfdo: I believe this is good! My understanding from our last check-in with the localization folks (quite some time ago) is that the danger from this approach is if we ourselves were to write JS directly in there or if were to unescape like `<apex:outputText escape="false" value="{!Label. bgeRetirementWarningBody}">` wherever this label is used.

ptewson-sfdo: Thank you! Sounds good, marking this resolved!

psadie-sfdo (approved):
Looks good! Thanks David!

src/labels/CustomLabels.labels
melissabarber, line 5102:
```suggestion
        <value>On July 13, 2021, legacy Batch Gift Entry will be retired and will no longer work in NPSP. If you have any in-progress batches at that time, you can still access those records, but you can't process them using Batch Gift Entry. To ensure a smooth transition, enable and use the newer Gift Entry in advance of July 13th. Learn more about Gift Entry in the &lt;a href=&apos;https://powerofus.force.com/NPSP_Documentation&apos; target=&apos;_blank&apos;&gt;NPSP Documentation&lt;/a&gt;.</value>
```

sokosaurus: Committed your change to this branch. Thanks for the update @melissabarber. Also, on this topic @ptewson-sfdo  - Melissa and I chatted about this yesterday and wanted to get your thoughts on whether or not you think this label might need to be localized. The non-english/international install base looks to small fyi.

melissabarber:
Hey gang! Confirmed with Ali that this PR should merge into the main branch instead of the feature branch. This message is to warn users of the upcoming retirement that'll happen with the 232 release.

CC @schlossmj

sokosaurus:
Please see #6279

PR #6292, opened 2021-03-15
CCI: Mock core instrumentation class for an org without a namespace

force2b:
- Mock (as unmanaged) the Core Splunk logging class (`SfdoLogUtils`) for instrumentation to allow for compilation.
- The actual Core class (of the same name) will only exist the context of the "npsp" namespace. To support development in an org without a namespace, the mocked class in this PR provides a simple stub interface to reflect the core class and methods. It's sole purpose is to  allow the package Apex to compile when there is no defined namespace in the org. (i.e., deploy the package code into the org).
- The `SfdoLogUtils` apex class mocked in this PR is a stub only and will not log anything.
- The `SfdoLogUtils` apex class in this PR should **never** be deployed to the packaging org.
- Work Item: [W-9776327](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000UTO0YAO/view)

# Critical Changes

# Changes

- Implement an Instrumentation Service for NPSP to support collecting aggregate data to improve the understanding of product feature usage. More information can be found at LINK_TBD

# Issues Closed

# Community Ideas Delivered

# New Metadata

Apex Classes:
- SfdoInstrumentationService
- SfdoInstrumentationService_TEST
- SfdoInstrumentationEnum

# Deleted Metadata


cumulusci.yml
```diff
@@ -825,6 +832,12 @@ flows:
                     api_names: QA,Gift_Entry
 
 
+    deploy_unmanaged:
```
force2b, line 835: @davidmreed Can you confirm that the 'deploy_unmanaged` flow will only execute when deploying unmanaged npsp into an org - one that doesn't have a namespace?

davidmreed: No. Deploy unmanaged will also run for deployments into namespaced scratch orgs.

force2b: Glad I asked! Is there a flow/task that I can hook into here that will **only** deploy into an unmanaged org (including not deploying into the packaging org)?

force2b: Though, now that I think of it, that might actually be good as is. My goal is for the class in this PR to never be in the packaging org. If it exists in a scratch org with a namespace, that probably would be ok.

davidmreed: You could include a task in `deploy_unmanaged` with a `when` clause, like `when: not org_config.namespace`. I think that'd work fine.

force2b: Good idea. Testing that now.

force2b: Worked perfectly, thank you!! 
I'm still finalizing the method signatures for the `sfdoLogUtils` class in core, but ideally this will be ready for merging in the next couple of weeks.

cumulusci.yml
```diff
@@ -587,6 +587,13 @@ tasks:
                 errSettings.%%%NAMESPACE%%%OverrideFeature_PilotEnabled__c = true;
                 upsert errSettings;
 
+    deploy_core_mock_class:
+        description: Deploy a mock apex class that will prevent compilation errors when the packaged code references file-based apex in the core stack for instrumentation purposes
+        class_path: cumulusci.tasks.salesforce.Deploy
+        group: Salesforce Metadata
+        options:
+            path: unpackaged/pre/core_apex_class_mock
```
force2b, line 595: @davidmreed Out of curiosity, am I able to update the `Codeowners` file to add an entry for `unpackaged/pre/core_apex_class_mock` or is that only for RE related folders?

davidmreed: You can do that. I'll probably have to approve the change (which I'm happy to do), but it's a perfectly legitimate use of CODEOWNERS.

PR #6294, opened 2021-03-16
Fixed template builder failures

schlossmj:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/robot/Cumulus/tests/browser/gift_entry/template_builder.robot
+++ b/robot/Cumulus/tests/browser/gift_entry/template_builder.robot
@@ -6,11 +6,15 @@ Library         cumulusci.robotframework.PageObjects
 Suite Setup     Run keywords
 ...             Open Test Browser
 ...             API Check And Enable Gift Entry
+...             Setup Test Data
 Suite Teardown  Capture Screenshot and Delete Records and Close Browser
 
-*** Variables ***
-${template}       Robot Template
-${new_template}   Robot Copy Template
+*** Keywords ***
+Setup Test Data
+    ${template} =           Generate Random String
+    Set Suite Variable      ${template}
+    ${new_template} =       Generate Random String
+    Set Suite Variable      ${new_template}
 
 *** Test Cases ***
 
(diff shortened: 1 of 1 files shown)


robot/Cumulus/tests/browser/gift_entry/template_builder.robot
```diff
@@ -40,11 +44,13 @@ Create Clone and Delete Template
     Click Gift Entry Button                   Save & Close
     Wait Until Page Contains                  This name has been used by another template. Please enter a unique name.
     Enter Value In Field                      Template Name=${new_template}
+    Click Gift Entry Button                   Next: Form Fields
+    Click Gift Entry Button                   Next: Batch Settings
```
skristem, line 48: why add these 2 steps when we are not doing anything on those tabs?

robot/Cumulus/tests/browser/gift_entry/template_builder.robot
```diff
@@ -40,11 +44,13 @@ Create Clone and Delete Template
...
     Click Gift Entry Button                   Save & Close
     Current Page Should Be                    Landing                        GE_Gift_Entry
     Click Link                                Templates
     Wait Until Page Contains                  ${new_template}
-    
+    Store Template Record Id                  ${new_template}
```
skristem, line 53: This step was not originally added as the template is being deleted in the next step but makes sense if the test fails before going to the next step then it gets deleted as part of teardown

deepmalamehta (approved):

robot/Cumulus/tests/browser/gift_entry/template_builder.robot
schlossmj: Yep, that was the intention with this. I'm seeing lots of cleanup failures when I run tests, so I'm working to ensure it gets deleted no matter what.

robot/Cumulus/tests/browser/gift_entry/template_builder.robot
schlossmj: This was the main cause of the failure - the form throws an error if you don't click to the Form Fields tab and skip right to saving. I had it mimic creating the previous template to ensure parity.

skristem (approved):

PR #6298, opened 2021-03-16
Refactor NPSP Address Management To Use Domain Classes 

kenlewis-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/BDI_DataImportService.cls
```diff
@@ -1777,21 +1777,24 @@ global with sharing class BDI_DataImportService {
                 addr.Address_Type__c = 'Home';
 
                 if (con1 != null) {
-                    ADDR_Addresses_TDTM.copyAddressAddrSObj(addr, con1, 'Mailing', 'npe01__Primary_Address_Type__c');
+                    NPSP_Address address = new NPSP_Address(addr);
+                    address.copyOntoSObject(con1, 'Mailing','npe01__Primary_Address_Type__c');
```
kenlewis-sfdo, line 1781: Eventually this should be: `npspContact.mailingAddress(address);`, instead of telling the Address to copy itself onto the contact, we will tell the contact to use the Address.

src/classes/anAddress.cls
```diff
@@ -0,0 +1,27 @@
+public interface anAddress {
```
kenlewis-sfdo, line 1: I was originally thinking we might have two implementations - one for our regular address and one for seasonal. But it seems like we might just have one concrete class for NPSP_Address.  In that case this interface can be removed, but curious to hear others thoughts.

src/classes/AddressService.cls
```diff
@@ -0,0 +1,60 @@
...
+    /*******************************************************************************************************
+    * @description Utility to copy Address fields between Contacts and Accounts (or vice-versa).
+    * Handles instances where State and Country picklists are enabled, and multiline street addresses.
+    * @param sobjSrc the source Contact or Account
+    * @param strFieldPrefixSrc the address fields to copy from, ie., Mailing, Other, Shipping, Billing
+    * @param sobjDst the destination Contact or Account
+    * @param strFieldPrefixDst the address fields to copy to, ie., Mailing, Other, Shipping, Billing
+    * @return void
+    */
+    public void copyAddressStdSObj(sobject sobjSrc, string strFieldPrefixSrc, sobject sobjDst, string strFieldPrefixDst) {
```
kenlewis-sfdo, line 46: There isn't a great place for this method right now.  As I mention in my other comment, we should probably have Entities (NPSP_Account, NPSP_Contact) that have public methods that accept an instance of NPSP_Address, and then set the value of the objects address fields internally.  

This AddressService seems like a logical place to house this method until that architecture is available.

src/classes/NPSP_Address.cls
```diff
@@ -0,0 +1,361 @@
...
+    public NPSP_Address(Address__c address) {
+        this.addr = address;
+    }
+
+    public Boolean isSeasonal() {
+        return (this.addr.Seasonal_Start_Month__c != null && this.addr.Seasonal_Start_Day__c != null &&
+                this.addr.Seasonal_End_Month__c != null && this.addr.Seasonal_End_Day__c != null);
+    }
+
+    public Address__c getRecord() {
```
kenlewis-sfdo, line 48: This method should eventually be made private, but is public for now until we have all of the public methods on this class to expose the public properties.

So in usages, instead of `getRecord().MailingStreet__c`, it will eventually be `npspAddress.mailingStreet();`

For now, getRecord() is available until this object is more fully flushed out with all of those public methods.

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1161 @@
...
+        if (errors.errorsExist) {
+            putErrorsOnAddresses(errors, addressesByHouseholdAccountId(), new Map<Id, Address__c>());
+        }
+    }
+
+    /********************************************************************************************************
+    * @description Retreive the defined default address type picklist value to use when creating a new
+    * Address record.
+    */
+    public static String defaultAddressType {
```
force2b, line 147: There are a lot of Static methods in this class. As a domain class, shouldn't these be instance methods, or if they're just simple util methods they would go into a different class? Though, maybe that's your intent with this first iteration?

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -176,14 +176,15 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
         list<Address__c> listAddr = new list<Address__c>();
         for (Account acc : listAcc) {
             Address__c addr = new Address__c();
-            addr.Household_Account__c = acc.Id;
-            addr.Default_Address__c = true;
-            addr.Latest_Start_Date__c = system.today();
-            addr.Latest_End_Date__c = null;
-            ADDR_Addresses_TDTM.copyAddressStdSObjAddr(acc, 'Billing', addr, null);
-            listAddr.add(addr);         
+            NPSP_Address npspAddress = new NPSP_Address(addr);
```
force2b, line 179: If we eventually want to move this to an ORM, the calls to `getRecord().{fieldname}` would need to be replaced with something like `.setAccount(acc.id)`, etc. In this example though, I'm not sure we're really gaining any value by instantiating `NPSP_Addresses` in this if-block context. It's still directly manipulating the field instances and in the end we just add the record to the listAddr collection and the instance of NPSP_Address goes away at the curly brace on line 185. It honestly might be more efficient to stick with the original code here, unless we're going to implement a full ORM at this point - but that would require a lot more changes across this class.

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1161 @@
...
+        return addr.MailingStreet__c + '\r\n' + addr.MailingStreet2__c;
+    }
+
+    /*******************************************************************************************************
+    * @description specifies whether state and country picklists are enabled in this org.
+    * @return true if enabled
+    */
+    public static Boolean isStateCountryPicklistsEnabled {
+        get {
+            if (isStateCountryPicklistsEnabled == null)
```
force2b, line 442: MIght was well add curly braces while we're here

src/classes/ContactRepository.cls
```diff
@@ -0,0 +1,101 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Contact repository responsible for retrieval and storing of Contacts for specified Accounts
```
force2b, line 33: Can you say more about how this is used? Is it more like a Household (Contacts within an Account)?

src/classes/NPSP_Address.cls
force2b: Ah. This is basically buried in a question I had further up (which I'll try to remember to update before committing the review).

src/classes/NPSP_Address.cls
```diff
@@ -0,0 +1,361 @@
...
+        return (cParts > 0 && cParts < 4);
+    }
+
+    /*******************************************************************************************************
+    * @description utility to create the month and day strings to a decimal we can use in comparisons.
+    * @param m the month ordinal as a string
+    * @param d the day ordinal as a string
+    * @return decimal
+    */
+    private static decimal decimalOfMonthDay(string m, string d) {
```
force2b, line 293: Should this be static and/or in this class?

src/classes/anAddress.cls
force2b: I don't think we'll need this. Seasonal vs. non-Seasonal Addresses are all still addresses. The only difference is the start/end periods, which are technically still attributes of an address even if they don't have a value.

force2b (commented):
Got through a quick review before eod.

There's a lot to wrap my head around in this one, especially since some of the domain and ORM models are not yet fully in place. A few overall notes though:
- For new classes, I recommend fixing the obvious style problems: missing curly brackets, upper case keywords in soql, title case data types, etc.
- Try to use consistent method names for retrieval, comparison, and sets. I realize that a lot of this is copy/paste so the naming inconsistencies come from the original code. If you see stuff that looks odd though, definitely adjust. That'll help with reading and self-documentation.

src/classes/ContactRepository.cls
kenlewis-sfdo: This was an inner class of `ADDR_Addresses_TDTM.cls`.  I extracted it, but it should be otherwise unchanged. Besides tests, looks like it is only used once inside `Addresses.cleanupAccountAddresses` - it could be clarified with some small refactoring - could be renamed `ContactSelector` and not be instantiated with that set of Account Ids.  I'll try that.

src/classes/ADDR_Account_TDTM.cls
kenlewis-sfdo: More logic has been moved into the NPSP_Address class now, including the rest of this block.  The getRecord() methods will need to stay for a while, until we get all appropriate properties exposed as public methods, at that point we'll be able to make getRecord() private.

src/classes/Addresses.cls
kenlewis-sfdo: Yes, most everything should be instance, I haven't changed many static vars/methods from static to instance or vice versa.  This is probably a result of moving this var. It came over from ADDR_Addresses_TDTM.defaultAddressType

kenlewis-sfdo:
See #6330 for all combined work.

PR #6301, opened 2021-03-17
Customizable Rollups: Force Certain Records to run in Skew

npsp-reedestockton:
We provided a way to force certain Account and Contact records to use skew mode for rollups.
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Fields
  - Account.CustomizableRollups_UseSkewMode__c
  - Contact.CustomizableRollups_UseSkewMode__c

# Deleted Metadata


npsp-reedestockton:
Tracking [W-7783082](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EIZsIAO/view)

src/classes/CRLP_RollupBatch_SVC.cls
```diff
@@ -130,10 +130,15 @@ public class CRLP_RollupBatch_SVC {
                 jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
                 jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
             String objName = recordId.getSobjectType().getDescribe().getName();
-            String soql = 'SELECT Id FROM ' + objName + ' WHERE Id = :recordID ';
+            String soql = 'SELECT Id FROM ' + objName + ' WHERE Id = :recordID AND ('
+                + UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c')
+                + ' = true ';
```
npsp-reedestockton, line 135:
```suggestion
                + ' = true';
```

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -499,8 +499,20 @@ public inherited sharing class CRLP_RollupProcessor {
         SObject updatedRecord = handler.getPopulatedSObject();
 
         // Determine if the updated SObject record is different than the parent.
-        // If not, set it to null
-        if (CRLP_Rollup_SVC.resultsNeedUpdate(parent, updatedRecord, rollupDefs) == false) {
+        Boolean needsUpdate = CRLP_Rollup_SVC.resultsNeedUpdate(parent, updatedRecord, rollupDefs);
+
+        // Set skew mode flag on skew mode records not currently set
+        String qualifiedSkewField = UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c');
+        if (rollupType != CRLP_RollupProcessingOptions.RollupType.GAU
```
force2b, line 506: @npsp-reedestockton  The only other thing to check for here is that the job is running in SkewMode. I think that's defined in the `mode` var.

npsp-reedestockton: Thanks!  I lost track of the fact that I wasn't in a skew mode job for a second there...

npsp-reedestockton:
@lparrott @andrewyu-salesforce @force2b Please review when you have a chance.  Thanks!

@mattwashington Please review labels and help on the two new fields.  Thanks!

lparrott (approved):

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -488,6 +488,15 @@ public inherited sharing class CRLP_RollupProcessor {
         // has changed or not.
         Id parentId = parent.Id;
 
+        Boolean hasSkewFlag = parent.getSObjectType() == Account.SObjectType || parent.getSObjectType() == Contact.SObjectType;
```
force2b, line 491: For performance reasons, some of this can be moved into a method that is set a single time. Rollups are always run for the same object in the same context. So, instead of doing this compare and the `isSet()` call 200 times in an execution context (assuming a batch size of 200 times), we can do it just once the first time the method is called and re-use the result for all subsequent calls. In the below, I figure this method can call `getSkewModeValues(parent)` at the top to handle setting the values if needed. I also renamed some vars here.

```
private final String qualifiedSkewField;
private Boolean useForceSkewModeField;
private Boolean isSetSkewModeField;

private void getSkewModeValues(SObject parent) {
    if (qualifiedSkewModeField == null) {
        this.useForceSkewModeField = (parent.getSObjectType() == Account.SObjectType || parent.getSObjectType() == Contact.SObjectType);
        this.isSetSkewModeField = parent.isSet(qualifiedSkewField);
        this.qualifiedSkewField = UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c');
    }
}
```

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -499,8 +508,20 @@ public inherited sharing class CRLP_RollupProcessor {
...
+        Boolean needsUpdate = CRLP_Rollup_SVC.resultsNeedUpdate(parent, updatedRecord, rollupDefs);
+
+        // Unconditionally set force skew flag if parent is Account/Contact and running in skew mode
+        if (mode == CRLP_RollupProcessingOptions.BatchJobMode.SkewMode && hasSkewFlag) {
+            updatedRecord.put(qualifiedSkewField, true);
+
+            // If force skew flag was not initially true, return for update unconditionally
+            if (!forceSkew) {
+                return updatedRecord;
+            }
```
force2b, line 520:
```suggestion
            needsUpdate = true;
```

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -488,6 +488,15 @@ public inherited sharing class CRLP_RollupProcessor {
         // has changed or not.
         Id parentId = parent.Id;
 
+        Boolean hasSkewFlag = parent.getSObjectType() == Account.SObjectType || parent.getSObjectType() == Contact.SObjectType;
+        Boolean forceSkew = false;
```
force2b, line 492: Mostly just renaming the var here to make it more clear what is being forced

```suggestion
        Boolean isForceSkewModeSet = false;
```

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -488,6 +488,15 @@ public inherited sharing class CRLP_RollupProcessor {
...
+        String qualifiedSkewField = UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c');
+
+        // Save the skew flag setting
+        if (hasSkewFlag && parent.isSet(qualifiedSkewField)) {
+            forceSkew = (Boolean) parent.get(qualifiedSkewField);
```
force2b, line 497:
```suggestion
            isForceSkewModeSet = (Boolean) parent.get(qualifiedSkewField);
```

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -499,8 +508,20 @@ public inherited sharing class CRLP_RollupProcessor {
         SObject updatedRecord = handler.getPopulatedSObject();
 
         // Determine if the updated SObject record is different than the parent.
-        // If not, set it to null
-        if (CRLP_Rollup_SVC.resultsNeedUpdate(parent, updatedRecord, rollupDefs) == false) {
+        Boolean needsUpdate = CRLP_Rollup_SVC.resultsNeedUpdate(parent, updatedRecord, rollupDefs);
+
+        // Unconditionally set force skew flag if parent is Account/Contact and running in skew mode
+        if (mode == CRLP_RollupProcessingOptions.BatchJobMode.SkewMode && hasSkewFlag) {
```
force2b, line 514: I think if you check for isForceSkewModeSet being false here, then the rest of the code in the if block only runs when the field needs to be set to true, which also means that an update is required.

```suggestion
        if (mode == CRLP_RollupProcessingOptions.BatchJobMode.SkewMode && hasSkewFlag && !isForceSkewModeSet) {
```

force2b (changes requested):
This looks really nice. I made some suggestions for caching vars to improve performance as well as some renaming.

I have a question though; or maybe just a new testing path:
* When we deploy this new field to customer orgs, no Users will no have access to it.
* However, when any user clicks the [Recalculate Rollups] button on the Account or Contact record, the controller will call a method that queries the record to determine if Skew mode is needed or not. That update query includes the new ForceSkewMode field.
* If the User has no access to that field, will that cause the query to fail? My gut says it should work because we're not using StripInaccessibleFields() or WITH SECURITY ENFORCED. However, it's worth double checking that the query still works when the button on the Account/Contact is used even if the User has no access to the field. Just to be on the safe side.

npsp-reedestockton:
@force2b Implemented code review suggestions, with a couple of minor modifications to fix scenarios other than Account/Contact parent record.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -499,14 +516,37 @@ public inherited sharing class CRLP_RollupProcessor {
...
+     */
+    private void getSkewModeValues(SObject parent) {
+        if (this.qualifiedSkewField == null) {
+            this.isSetSkewModeField = false;
+            this.qualifiedSkewField = UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c');
+            this.useForceSkewModeField = (parent.getSObjectType() == Account.SObjectType || parent.getSObjectType() == Contact.SObjectType);
+            if (this.useForceSkewModeField && parent.isSet(this.qualifiedSkewField))
+            {
+                this.isSetSkewModeField = (Boolean) parent.get(this.qualifiedSkewField);
+            }
```
force2b, line 546: This particular block shouldn't be in this method because it will vary for each individual parent record. It should be up in the `completeRollupForSingleSummaryRecord` method.

force2b (changes requested):

src/classes/CRLP_RollupProcessor.cls
npsp-reedestockton: Good catch.  Thanks.  That also means your earlier suggestion to update skew values once should ignore isSetSkewModeField and leave that to be handled on a record-by-record basis.

npsp-reedestockton:
@force2b I've tested the scenario where a user has no access to CustomizableRollups_UseSkewMode__c.  It still works fine.  Tested on both Account and Contact.

force2b (commented):
Thanks for testing the user access scenario. The logic looks fantastic. I made a couple of renaming suggestions only.

force2b (approved):
Great work!

rponti-sforg:
QE testing is now complete for this PR and subsequent WI.  Tests are passing based on WI ACs, additional tests cases, test charter feedback, and PR comments.

For further WI and testing details, see:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008EIZsIAO/view
https://salesforce.quip.com/UTaTA9akLOlh

PR was merged successfully.  The primary (required) post merge builds have passed.  There are a couple of **tag: uat/3.195-Beta_13** builds that were restarted.  

Setting WI status to **Closed**.

PR #6303, opened 2021-03-17
RD2: RD Updating Entry Form to allow credit card edits

lparrott:
- We updated the Enhanced Recurring Donation Edit page to allow editing Credit Card information.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
@lparrott I'm not quite sure I understand the branching on this PR. This branch is not a child of Andrew's, but I also wasn't sure if Andrews is supposed to be in 232 since it was basically recreating a bunch of [changes that I made](https://gus.lightning.force.com/a07B0000008lUKsIAM) and that are only in the `feature/232` branch. I think we need to figure out the branching and where changes exist to know where these PR's should exist. (CC @andrewyu-salesforce)

Though, if you do need to create a child of another branch, be sure to use child-branch naming to avoid an accidental overwrite from merges into master.

force2b (changes requested):
I think this PR is supposed to be a layer on top of Andrews work, but the code doesn't exactly match up. For example, the changes in the RD2_EntryFormController class in this PR do not match to the exact same changes in Andrews version of that file. I'm sure there are others. I think this is a result of not using a child branch of Andrews. 

We may have to shelve this PR until Andrews is tested and merged, or possibly vice-versa.  The other option might be to create a child branch off of Andrews and then recreate the changes in this PR off of that. I suppose you may be able to rename this branch and then re-base the PR. Either way though I'm seeing code that is both the same and different in two PR's, with one merging into the other. That will definitely create issues beyond just merge conflicts.

andrewyu-salesforce (changes requested):
@lparrott great work. Just couple feedback/suggestion on the method call and custom label used.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
```diff
@@ -13,34 +13,87 @@
...
-                                </lightning-button>
-                            </p>
+                            <template if:true={isEditMode}>
+                                <lightning-layout vertical-align='center' multiple-rows='true'>
+                                    <lightning-layout-item size="4" class=" slds-text-color_weak slds-p-bottom_small">
+                                        <label for="last4ach" id="achlast4Label"
+                                            class="slds-form-element__label slds-text-body_regular slds-text-color_default slds-p-vertical_none slds-m-bottom_x-small">
+                                            Card Last 4
+                                        </label><br/>
+                                        <lightning-formatted-text id="last4ach"
```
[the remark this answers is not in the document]
lparrott: I copied the qa-locator from `rd2ElevateInformation`, should it be different than that?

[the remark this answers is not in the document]
andrewyu-salesforce: For `rd2ElevateInformation`, ACH and Credit Card last 4 is displayed conditionally. If you look to the credit card last 4 section([here](https://github.com/SalesforceFoundation/NPSP/blob/131c922ba797f108c196bf27574df68f3db8749f/src/lwc/rd2ElevateInformation/rd2ElevateInformation.html#L125)) you will see the id reference is `last4cc`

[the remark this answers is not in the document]
lparrott: Ok, I just pushed an update for this and your other feedback, let's chat tomorrow if I'm still missing something!

andrewyu-salesforce (approved):
@lparrott everything looks good.

src/classes/RD2_EntryFormController.cls
```diff
@@ -86,7 +86,8 @@ public with sharing class RD2_EntryFormController {
             'InstallmentPeriodPermissions' => getFieldPermissionsMap('npe03__Installment_Period__c'),
             'InstallmentFrequencyPermissions' => getFieldPermissionsMap(UTIL_Namespace.StrTokenNSPrefix('InstallmentFrequency__c')),
             'customFieldSets' => getCustomFieldSectionFields(),
-            'isElevateCustomer' => RD2_ElevateIntegrationService.isIntegrationEnabled()
+            'isElevateCustomer' => RD2_ElevateIntegrationService.isIntegrationEnabled(),
+            'isPilotEnabled' => UTIL_FeatureEnablement.isPilotEnabled()
```
force2b, line 90: I recommend matching the pattern implemented by @andrewyu-salesforce here - https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/RD2_ElevateInformation_CTRL.cls#L65

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -129,6 +133,19 @@ export default class rd2EntryForm extends LightningElement {
...
     @track error = {};
 
+    @api 
+    get nextDonationDate() {
+        const localDate = new Date(this._nextDonationDate);
+        return new Date(localDate.getUTCFullYear(), localDate.getUTCMonth(), localDate.getUTCDate());
+    }
+
+    @api 
+    get cardExpDate(){
```
force2b, line 143: Does normal edit/insert still work if the User does not have edit permissions to these fields?

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -153,12 +170,17 @@ export default class rd2EntryForm extends LightningElement {
                 this.customFields = response.customFieldSets;
                 this.hasCustomFields = Object.keys(this.customFields).length !== 0;
                 this.isElevateCustomer = response.isElevateCustomer;
+                this.isPilotEnabled = response.isPilotEnabled;
```
force2b, line 173:
```suggestion
                this.isEditEnabled = response.isEditEnabled;
```

force2b (commented):
Nice. So much easier to review now.
Anyway, a few suggestions.

src/lwc/rd2EntryForm/rd2EntryForm.js
lparrott: Yes, but I think we may want to rethink how permissions are being checked in regard to this widget.

src/classes/RD2_EntryFormController.cls
lparrott: I like it! Will update this afternoon.

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -330,8 +334,9 @@ export default class rd2EntryForm extends LightningElement {
     */
     evaluateElevateEditWidget(paymentMethod) {
         let statusField = getFieldValue(this.record, FIELD_STATUS);
+        this.commitmentId = getFieldValue(this.record, FIELD_COMMITMENT_ID);
 
-        if(this.isEdit && statusField != STATUS_CLOSED){
+        if(this.isEditEnabled && this.commitmentId !== null && this.isEdit && statusField != STATUS_CLOSED){
```
force2b, line 339:
```suggestion
        if (this.isEditEnabled && this.commitmentId !== null && this.isEdit && statusField !== STATUS_CLOSED){
```

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -129,6 +134,24 @@ export default class rd2EntryForm extends LightningElement {
 
     @track error = {};
 
+    @api
```
force2b, line 137: block comments needed for the methods in this class

force2b (changes requested):
Looking very good. Just a couple of minor things

src/lwc/rd2EntryForm/rd2EntryForm.js
lparrott: I definitely need to look at the style guide again, thanks for pointing this out.

force2b (approved):
Fantastic work on this @lparrott

PR #6307, opened 2021-03-19
Require Opportunity Amount Access for Payment Wizard

voduyemi:
WI: [W-8855391](https://gus.lightning.force.com/a07B00000095uWsIAI)

# Critical Changes
Added a full-page blocker if the currently logged-in user lacks access(read and edit) to the opportunity amount field when using the payment wizard feature.

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/PMT_PaymentWizard_CTRL.cls
+++ b/src/classes/PMT_PaymentWizard_CTRL.cls
@@ -61,13 +61,13 @@ public with sharing class PMT_PaymentWizard_CTRL {
     * @description Permissions class used to determine if the running user has permissions
     */
     @TestVisible
-    private UTIL_Permissions perms {
+    private UTIL_Permissions permissions {
         get {
-            if (perms == null) {
-                perms = new UTIL_Permissions();
+            if (permissions == null) {
+                permissions = new UTIL_Permissions();
             }
 
-            return perms;
+            return permissions;
         }
         set;
     }
@@ -277,7 +277,7 @@ public with sharing class PMT_PaymentWizard_CTRL {
     @TestVisible
     private void setupPaymentWizard() {
         String id = ApexPages.currentPage().getParameters().get('id');
-        Opportunity[] o = Database.query(makeOppQuery() + ' WHERE Id = :id');
+        Opportunity[] o = Database.query(buildOpportunityQuery() + ' WHERE Id = :id');
 
         isWriteoffWizard = false;
         isScheduleWizard = true;
@@ -308,7 +308,7 @@ public with sharing class PMT_PaymentWizard_CTRL {
     @TestVisible
     private void setupWriteoffWizard() {
         String id = ApexPages.currentPage().getParameters().get('id');
-        Opportunity[] o = Database.query(makeOppQuery() + ' WHERE Id = :id');
+        Opportunity[] o = Database.query(buildOpportunityQuery() + ' WHERE Id = :id');
 
         isWriteoffWizard = true;
         isScheduleWizard = false;
@@ -338,9 +338,10 @@ public with sharing class PMT_PaymentWizard_CTRL {
     * @description Creates an opportunity query with payment mapping and currency fields, filtered on id paramter.
     * @return string An opportunity query string.
     */
-    private String makeOppQuery() {
+    private String buildOpportunityQuery() {
         // requery the Opportunity for all fields that we need.
-        String query = 'SELECT name, amount, stagename, npe01__payments_made__c, npe01__amount_outstanding__c, description, closedate, npe01__Number_of_Payments__c, isClosed, isWon';
+        String query = 'SELECT name, amount, stagename, npe01__payments_made__c, ' +
+            'npe01__amount_outstanding__c, description, closedate, npe01__Number_of_Payments__c, isClosed, isWon';
 
         // add any opportunity fields from payment field mappings
         for (String s : PMT_PaymentCreator.paymentMappings.keyset()) {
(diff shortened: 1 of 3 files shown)


src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -534,52 +529,47 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
+        if (canCheckPermissionsFor(payment)) {
+            return hasAccessTo(payment, paymentFields())
+                && hasAccessTo(opportunity, opportunityFields());
         }
+        return false;
+    }
 
-        return accessResult;
+    private Boolean canCheckPermissionsFor(SObjectType objectType) {
+        return perms.canCreate(objectType) && perms.canDelete(objectType);
```
psadie-sfdo, line 548: I would maybe spell out permissions instead of perms, but super minor and not a blocker.

psadie-sfdo (commented):
Thanks for putting this together Victor! This looks good to me based on the conversations in the work item.

src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -339,7 +338,7 @@ public with sharing class PMT_PaymentWizard_CTRL {
     * @description Creates an opportunity query with payment mapping and currency fields, filtered on id paramter.
     * @return string An opportunity query string.
     */
-    public string makeOppQuery() {
+    private String makeOppQuery() {
```
dfuller114, line 341: Could we spell out "Opportunity" here instead of "Opp"?

src/classes/PMT_PaymentWizard_CTRL.cls
dfuller114: agree

src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -534,52 +529,47 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
-                npe01__oppPayment__c.fields.npe01__written_off__c
-            };
-            if (!(perms.canRead(pmt, sObjectFields) && perms.canUpdate(pmt, sObjectFields))) {
-                return accessResult;
-            }
-            accessResult = true;
+    private Boolean hasFeatureAccess() {
+        SObjectType payment = npe01__OppPayment__c.getSObjectType();
+        SObjectType opportunity = Opportunity.getSObjectType();
+        if (canCheckPermissionsFor(payment)) {
```
dfuller114, line 540: Just reading through this, I'm not sure if this "canCheckPermissionsFor" is clear in its intent in the naming. I see what you're going for in checking for create/delete first before checking for read/update, but for the sake of readability here I'm wondering if it would be better to just have the "hasAccessTo" check without the additional "create" / "delete" check and we could roll the "create" and "delete" check into the "hasAccessTo" method which would then comprise all of the required checks.

dfuller114 (changes requested):
just a couple comments from me. Everything is looking good!

src/classes/PMT_PaymentWizard_CTRL.cls
voduyemi: I agree with your comment here, I'll move the call to `canCheckPermissionsFor` into `hasAccessTo`.

voduyemi: Updated!

src/classes/PMT_PaymentWizard_CTRL.cls
voduyemi: Updated!

src/classes/PMT_PaymentWizard_CTRL.cls
voduyemi: Updated!

src/classes/PMT_PaymentWizard_CTRL.cls
```diff
@@ -535,51 +532,47 @@ public with sharing class PMT_PaymentWizard_CTRL {
...
+    private Boolean hasAccessTo(SObjectType sObjectType, Set<SObjectField> sObjectFields) {
+        if (canCheckPermissionsFor(sObjectType)) {
+            return permissions.canRead(sObjectType, sObjectFields)
+                && permissions.canUpdate(sObjectType, sObjectFields);
         }
+        return false;
+    }
 
-        ApexPages.addMessage(new ApexPages.Message(ApexPages.Severity.ERROR, errorMsg));
+    private Boolean canCheckPermissionsFor(SObjectType objectType) {
```
dfuller114, line 554: Can we remove this method if it's no longer being used?

dfuller114 (changes requested):
one other quick change on the method that I don't think we're using anymore and then I think I'll be good to go!

src/classes/PMT_PaymentWizard_CTRL.cls
voduyemi: `canCheckPermissionsFor` is still being used in `hasAccessTo`. That check is a gated boolean, instead of cluttering the `if` statement with ` permissions.canCreate(objectType) && permissions.canDelete(objectType);`. I think it is better to have it in its own private method.

dfuller114: Got ya. Missed that and Intellij didn't find usages for some reason.

dfuller114 (approved):
looks good to me! Thanks Victor!

psadie-sfdo (approved):
Second smoke test looked good! Thanks!

ptewson-sfdo (approved):
This looks good to me!

PR #6309, opened 2021-03-22
Add detail section to Batch Gift Entry header

psadie-sfdo:
[W-8953529](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000097ob2IAA/view)
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GiftBatchSelector.cls
```diff
@@ -0,0 +1,62 @@
...
+
+    public List<DataImportBatch__c> selectGiftBatchesById(Set<Id> ids) {
+        return [
+            SELECT Id
+            FROM DataImportBatch__c
+            WHERE Id IN :ids
+        ];
+    }
+
+    public List<AggregateResultProxy> selectBatchGiftsGroupedByStatusBy(Id batchId) {
```
voduyemi, line 46: This query in this method feels a little out of place here for me. I believe this selector class is supposed to retrieve `DataImportBatch__c` records but this method retrieves aggregate results of `DataImport__c`. Don't we think this query should live in a `DataImportSelector` class?

src/classes/AggregateResultProxy.cls
```diff
@@ -0,0 +1,16 @@
+public class AggregateResultProxy {
```
voduyemi, line 1: I like your use of this proxy here. It's done this way so it supports aggregate queries of any `sObject` right ?

src/classes/GiftBatches.cls
```diff
@@ -0,0 +1,55 @@
...
+
+    public GiftBatches(List<DataImportBatch__c> records) {
+        super(records, DataImportBatch__c.SObjectType);
+    }
+
+    public Map<Id, Map<String, Integer>> getTotalsMap() {
+        Map<Id, Map<String, Integer>> totalsByBatchId = new Map<Id, Map<String, Integer>>();
+
+        for (SObject batch : getRecords()) {
+            List<AggregateResultProxy> totals = new GiftBatchSelector().selectBatchGiftsGroupedByStatusBy(batch.Id);
```
voduyemi, line 49: We might hit some governor limits here, since `new GiftBatchSelector().selectBatchGiftsGroupedByStatusBy(batch.Id);` is essentially a query. We might have to bulkify this call instead

src/classes/GE_GiftEntryController.cls
```diff
@@ -40,6 +40,10 @@ public with sharing class GE_GiftEntryController {
     public static final String PURCHASE_CALL_TIMEOUT_MESSAGE = System.Label.geErrorRequestTimedout;
     private static final String BOOLEAN_DATA_TYPE = 'BOOLEAN';
 
+    @AuraEnabled(cacheable=true)
+    public static Map<String, Integer> getGiftBatchTotalsBy(String batchId) {
+        return new GiftBatchService().getGiftBatchTotalsMapFor(batchId);
```
voduyemi, line 45: Should we passing a `String` `Id` to an an argument of type `Id`. I'm sure it works but it just feels weird to me. Maybe `Id.valueOf(batchId)` to be safe?

voduyemi (commented):
Thanks @psadie-sfdo for all your efforts. I left some comments. I'm still reviewing but I thought I put my comments out there before I start testing.

src/classes/AggregateResultProxy.cls
psadie-sfdo: Yep, exactly! I mainly wanted this proxy in order to support being able to mock methods that would normally accept an AggregateResult object or AggregateResult list as a param since we can't instantiate an AggregateResult like other sObjects.

src/classes/GiftBatches.cls
psadie-sfdo: Ah totally right! I'll make that update

src/classes/GE_GiftEntryController.cls
psadie-sfdo: What do you think of casting to `Id`? Pushed up an update.

src/classes/GiftBatches.cls
psadie-sfdo: Updated!

src/classes/GE_GiftEntryController.cls
```diff
@@ -40,6 +40,10 @@ public with sharing class GE_GiftEntryController {
     public static final String PURCHASE_CALL_TIMEOUT_MESSAGE = System.Label.geErrorRequestTimedout;
     private static final String BOOLEAN_DATA_TYPE = 'BOOLEAN';
 
+    @AuraEnabled(cacheable=true)
+    public static Map<String, Integer> getGiftBatchTotalsBy(String batchId) {
+        return new GiftBatchService().getGiftBatchTotalsMapFor((Id) batchId);
```
kenlewis-sfdo, line 45: Can Gift Batch Service be an injectable class level variable instead of newing up an instance here?

src/classes/GiftBatches.cls
```diff
@@ -0,0 +1,58 @@
...
+    public GiftBatches(Set<Id> ids, GiftBatchSelector selector) {
+        this(selector.selectGiftBatchesById(ids));
+    }
+
+    public GiftBatches(List<DataImportBatch__c> records) {
+        super(records, DataImportBatch__c.SObjectType);
+    }
+
+    public GiftBatchGroupedTotals getGroupedTotals() {
+        List<AggregateResultProxy> resultProxies = new GiftBatchSelector()
```
kenlewis-sfdo, line 46: Can the passed in selector instance be used here instead of newing?

src/classes/GiftBatches.cls
```diff
@@ -0,0 +1,58 @@
...
+            .selectBatchGiftsGroupedByStatusBy(giftBatchIds());
+        return new GiftBatchGroupedTotals(resultProxies);
+    }
+
+    private Set<Id> giftBatchIds() {
```
kenlewis-sfdo, line 51: Take a look at fflib_SObjects.getIdFieldValues :)

src/classes/GiftBatchGroupedTotals.cls
```diff
@@ -0,0 +1,48 @@
...
+            addTo(queryResultsByBatchId, queryResultProxy);
+        }
+        return queryResultsByBatchId;
+    }
+
+    private void addTo(
+        Map<Id, List<AggregateResultProxy>> queryResultsByBatchId,
+        AggregateResultProxy queryResultProxy) {
+
+        Id batchId = (Id) queryResultProxy.getValuesByAlias().get(NPSP_DATA_IMPORT_BATCH);
```
kenlewis-sfdo, line 32: `Id batchId = (Id) queryResultProxy.getValuesByAlias().get(String.valueOf
                (DataImport__c.NPSP_Data_Import_Batch__c));`

I'd skip the class variable And just use string.valueof on the field token here.

src/classes/AggregateResultProxy.cls
```diff
@@ -0,0 +1,16 @@
...
+    private Map<String,Object> valuesByAlias = new Map<String,Object> ();
+
+    public AggregateResultProxy withAggregateResult(AggregateResult groupedResult) {
+        String groupedResultJson = JSON.serialize(groupedResult);
+        this.valuesByAlias = (Map<String,Object>) JSON.deserializeUntyped(groupedResultJson);
+        return this;
+    }
+
+    public AggregateResultProxy withAliasAndValue(String alias, Object val) {
```
kenlewis-sfdo, line 10: If this is only used in tests can it be made private and test visible?

src/classes/GiftBatchStatusGroup.cls
```diff
@@ -0,0 +1,28 @@
+public inherited sharing class GiftBatchStatusGroup {
+    final private String status;
+    final private String paymentStatus;
+    private Integer recordCount;
+
+    public GiftBatchStatusGroup(AggregateResultProxy result) {
+        Map<String,Object> resultProxy = result.getValuesByAlias();
+        paymentStatus = (String) resultProxy.get(Schema.SObjectType.DataImport__c.fields.Payment_Status__c.getName());
```
kenlewis-sfdo, line 8:
```suggestion
        paymentStatus = (String) resultProxy.get(String.valueOf(DataImport__c.Payment_Status__c));
```

src/classes/GiftBatchStatusGroup.cls
```diff
@@ -0,0 +1,28 @@
...
+    public Boolean isImported() {
+        return status == 'Imported';
+    }
+
+    public Boolean isProcessingFailed() {
+        return status == 'Failed';
+    }
+
+    public Boolean isPaymentFailed() {
+        return paymentStatus == 'DECLINED';
```
kenlewis-sfdo, line 26: We sure that this is exactly what the payment status will be after a failed credit card transaction events are processed?

src/classes/GiftBatchGroupedTotals.cls
```diff
@@ -0,0 +1,48 @@
+public inherited sharing class GiftBatchGroupedTotals {
```
kenlewis-sfdo, line 1: All of the new classes need the copyright at the top and then a minimal apex dock header.

src/classes/GiftBatchService.cls
```diff
@@ -0,0 +1,42 @@
...
+* @author Salesforce.org
+* @date 2021
+* @description Service class for Gift Batches in NPSP.
+*/
+public with sharing class GiftBatchService {
+
+    public Map<String, Integer> getGiftBatchTotalsMapFor(Id batchId) {
+        GiftBatches batches = new GiftBatches(new Set<Id> {batchId}, new GiftBatchSelector());
+        GiftBatchGroupedTotals groupedTotals = batches.getGroupedTotals();
+        return groupedTotals.totalsFor(batchId);
```
kenlewis-sfdo, line 40: Can this just return the instance of GiftBatchGroupedTotals, And then the controller could return That instance to the client-side As a data structure?

psadie-sfdo: Yeah, I was hoping to do something along those lines as well but the catch there is that if we're returning an instance of something then we'd have to mark all the relevant properties as auraenabled and public.

That's kiiind of what I was trying to do with `GiftBatchTotals` but instead went with a map.

src/classes/GiftBatchStatusGroup.cls
psadie-sfdo: Yes, according to this [quip doc](https://salesforce.quip.com/771RADnOwEsG#FSHACAgtkdh).

But I do wish there was a more definitive way to get all the possible status values. Perhaps we should just establish our own internal enum at the very least that's used across all classes needing to reference payment status.

psadie-sfdo: scratch that, we do have one stashed inside `GE_PaymentServices`

melissabarber:
@psadie-sfdo looks good to go! Thanks.

src/classes/GiftBatchStatusGroup.cls
dfuller114: as a side note in the BGE CC work I've pulled the enum out to a top level instead of having it as an inner class https://github.com/SalesforceFoundation/NPSP/commit/4187846676d50d9effc76d07f893177bf443f20a

src/classes/GiftBatchAggregateGroupName.cls
```diff
@@ -0,0 +1,5 @@
+public enum GiftBatchAggregateGroupName {
```
kenlewis-sfdo, line 1: Even though this file is tiny, should probably still have the copyright at the top.

kenlewis-sfdo (approved):
Great work, Prince.  Lets keep an eye on those 2gp tests.  Assuming those get cleared, this is approved on my end!

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -0,0 +1,109 @@
...
+            await flushPromises();
+            const buttons = element.shadowRoot.querySelectorAll('lightning-button');
+            expect(buttons.length).toBe(3);
+        });
+
+        it('renders detail row', async () => {
+            getGiftBatchTotalsBy.mockResolvedValue(APEX_GIFT_BATCH_TOTALS_BY_SUCCESS);
+            const element = setup();
+
+            await flushPromises();
```
bdvorachek, line 79: Nice!

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -0,0 +1,109 @@
...
+    { virtual: true }
+);
+
+describe('c-ge-batch-gift-entry-header', () => {
+    afterEach(() => {
+        clearDOM();
+        jest.clearAllMocks();
+    });
+
+    function flushPromises() {
```
bdvorachek, line 39: This function is defined globally in jest.setup.js

bdvorachek (approved):
Nice tests and cleanup effort in the header component! I gave them a run against the version I'd like to bump up to in next sprint and they're solid green. The re-definition of flushPromises() in the test is pretty minor, and nothing I'd hold this up over.

kenlewis-sfdo (approved):
Thanks for fixing those tests!

PR #6314, opened 2021-03-23
Add Elevate information widget to flexipage

npsp-reedestockton:
This is intended to fill a gap in our ability to automate Elevate testing.  The goal is to provide a fully automated way to configure the Recurring Donations flexipage with the Elevate Information widget installed.
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/unpackaged/config/rd2_elevate_post_config/package.xml
+++ b/unpackaged/config/rd2_elevate_post_config/package.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+    <types>
+        <members>NPSP_Recurring_Donation</members>
+        <name>FlexiPage</name>
+    </types>
+    <version>48.0</version>
+</Package>
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -142,6 +142,14 @@ tasks:
             path: unpackaged/config/rd2_post_config
             namespace_inject: $project_config.project__package__namespace
 
+    deploy_rd2_elevate_config:
+        description: Deploys updates to enable Elevate for Enhanced Recurring Donations
+        class_path: cumulusci.tasks.salesforce.Deploy
+        group: NPSP
+        options:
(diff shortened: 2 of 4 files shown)


npsp-reedestockton:
@force2b There is a reason this is starting out as a draft PR.  This is not related to a work item.  The intent is to address the need for automation to have a fully configured way to test without manual intervention.  Happy to have input on a better way to handle this.

npsp-reedestockton:
@force2b Would appreciate your review of this item.  I think I've done what you had in mind, but would like to make sure.

unpackaged/config/rd2_elevate_post_config/package.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+    <types>
+        <members>NPSP_Recurring_Donation</members>
+        <name>FlexiPage</name>
+    </types>
+    <version>48.0</version>
```
force2b, line 7: I recommend using v51

force2b (commented):
I like that this can could be used later for other Elevate specific config, which makes me wonder if this folder/task is really more for "Elevate QA/Dev" and not really specific to RD2. For now, we're just adding this one new component to the record page, but maybe we want to automate other simple changes in Elevate orgs for dev and testing such as add other tabs, preconfigure some things, update perms, etc. Your new task could be that starting point.

force2b:
Also, whatever the new task ends up as, can you add it to the [automation.md](https://github.com/SalesforceFoundation/NPSP/blob/master/documentation/automation.md) file?

unpackaged/config/rd2_elevate_post_config/package.xml
npsp-reedestockton: @force2b I agree about considering wider application for this in the future.  It would be nice, for example, to take the config information we generate from the transport org, save it in .json format or some such, and use it in a flow to fully configured the connect Elevate org with remote site settings, etc.

PR #6330, opened 2021-03-26
Refactor NPSP Address Management To Use Domain Classes (ADDR_Contact_TDTM)

kenlewis-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -49,507 +49,450 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Contacts 
     * @return dmlWrapper.  
     ********************************************************************************************************/
-    public override DmlWrapper run(List<SObject> listNew, List<SObject> listOld, 
-        TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+    DmlWrapper dmlWrapper = new DmlWrapper();
```
kenlewis-sfdo, line 52: This class is still currently in progress.  Continuing to split and pull the logic out, pushing it mostly into the Contacts class.  More changes on the way.

src/classes/NPSP_Address.cls
```diff
@@ -0,0 +1,377 @@
...
+    /*******************************************************************************************************
+    * @description Utility to copy Address fields from an Address object to a Contact or Account.
+    * Handles instances where State and Country picklists are enabled, and multiline street addresses.
+    * @param addr the Address object to copy from
+    * @param sobjDst the destination Contact or Account
+    * @param strFieldPrefix the address fields to copy to, ie., Mailing, Other, Shipping, Billing
+    * @param strFieldAddrType an optional Address Type field on sobjDst to copy to
+    * @return void
+    */
+    public void copyOntoSObject(SObject sobjDst, String strFieldPrefix, String strFieldAddrType) {
```
psadie-sfdo, line 136: Would this later be replaced in the various entity classes (`NPSP_Account`, `NPSP_Contact`, etc) where we'd initiate and pass in an `NPSP_Address` to copy an address?

psadie-sfdo (commented):
Thank you for all of this work Ken! Quick pass-through and have a few minor comments below. 

Posted in the hummingbird slack channel for another question.

src/classes/NPSP_Address.cls
kenlewis-sfdo: Yep, exactly! I just pushed a few commits that get us a little closer to that.

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1196 @@
+/*
```
force2b, line 1: This class is just too large and a mixed collection of private, public, instance and static methods; some of which are quite long. I think it's going to be necessary to break this class down into something that is manageable - where each is focused on a specific behavior. My main concern with this at the moment is that it doesn't necessarily implement much from a "Domain" perspective that addresses flow and logic - especially around making it easier to make adjustments to the logic.

src/classes/ContactRepository.cls
```diff
@@ -0,0 +1,101 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Contact repository responsible for retrieval and storing of Contacts for specified Accounts
```
force2b, line 33: I'm still not necessarily comfortable with this class name. Maybe others have thoughts on another name?

src/classes/Contacts.cls
```diff
@@ -50,10 +61,133 @@ public inherited sharing class Contacts extends fflib_SObjects {
         synchronizeNamingExclusions();
     }
 
+    public void onBeforeInsertAddrConTDTM() {
```
force2b, line 64: What are the TDMT() methods here? I think these should have block comments to make usage clear since they're clearly different than the other onModeAction method.

src/classes/Contacts.cls
```diff
@@ -245,4 +379,537 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+            }
+        }
+        return new Contacts(contactsInHouseholdAccounts);
+    }
+
+    private Account accountFor(NPSP_Contact npspContact) {
+        return accountsByIds().get(npspContact.accountId());
+    }
+
+    public Contacts thatHaveAccounts() {
```
force2b, line 433:
```suggestion
    public Contacts getContactsWithAccount() {
```

src/classes/ADDR_Account_TDTM.cls
```diff
@@ -146,21 +147,22 @@ public class ADDR_Account_TDTM extends TDTM_Runnable {
...
+           !Addresses.equalsCaseSensitive(acc1.BillingCountry, acc2.BillingCountry);
+
+        if (!isChanged && Addresses.isStateCountryPicklistsEnabled) {
             isChanged = 
-	            !ADDR_Addresses_TDTM.equalsCaseSensitive(string.valueOf(acc1.get('BillingStateCode')), string.valueOf(acc2.get('BillingStateCode'))) || 
-	            !ADDR_Addresses_TDTM.equalsCaseSensitive(string.valueOf(acc1.get('BillingCountryCode')), string.valueOf(acc2.get('BillingCountryCode')));
+	            !Addresses.equalsCaseSensitive(String.valueOf(acc1.get('BillingStateCode')), String.valueOf(acc2.get('BillingStateCode'))) ||
+	            !Addresses.equalsCaseSensitive(String.valueOf(acc1.get('BillingCountryCode')), String.valueOf(acc2.get('BillingCountryCode')));
         }    
         return isChanged;
```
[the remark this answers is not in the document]
kenlewis-sfdo: Great idea! I just introduced an NPSP_Household entity, and I think would make sense to ask the Household if its Address changed. So I moved this into that entity. Take a look and let me know what you think!

src/classes/NPSP_Contact.cls
```diff
@@ -0,0 +1,117 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Contact entity class in NPSP.
+*/
+public with sharing class NPSP_Contact {
```
force2b, line 35: inherited sharing

src/classes/Contacts.cls
kenlewis-sfdo: This is a filter method that returns a subset of the underlying records.  Its usage is 
`Contacts consInAccts = contacts.thatHaveAccounts();`.  

`getContactsWithAccount()` sounds to me more like a query method that would return all of the underlying records with their parent Account record.

src/classes/ContactRepository.cls
kenlewis-sfdo: Yeah, this is really a mini HouseholdAccounts domain object.  It is instantiated with a set of AccountIds, then queries all Contacts in those Accounts, and has a method to get the Contacts for a given Account/Household.

It should really be `Accounts accts = new Accounts(acctIds);` then `accts.contactsForHouseholdOf(acctId);`

kenlewis-sfdo: Maybe the logic can be moved into the existing Accounts class.

src/classes/Addresses.cls
kenlewis-sfdo: Yep, this is definitely still a work in progress.  Migrating all of the logic out of the two ADDR_ TDTM classes AND fully refactoring every extracted/moved method proved to be too much to fit in at once.  The code is currently in a state where it has been extracted and moved, but not yet fully factored down.  As this class continues to be factored down, more logic gets pushed down into the NPSP_Address entity and more methods get extracted and de-duped in this class.  This process is underway, but not complete.   See the video I posted in the Hummingbird Slack channel and this commit that shows logic continue to get pushed down into the entity: [551677c](https://github.com/SalesforceFoundation/NPSP/pull/6330/commits/551677c6d335dced8e7974ee176986a62f693153)

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1196 @@
...
+            return addr.MailingStreet__c;
+        }
+        return addr.MailingStreet__c + '\r\n' + addr.MailingStreet2__c;
+    }
+
+    /*******************************************************************************************************
+    * @description specifies whether state and country picklists are enabled in this org.
+    * @return true if enabled
+    */
+    public static Boolean isStateCountryPicklistsEnabled {
```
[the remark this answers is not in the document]
kenlewis-sfdo: UTIL_Describe is already 700+ lines long and could probably be broken up into smaller classes.  I put this into a new class for now, OrgSettings, but open to keeping the conversation going.

src/classes/OrgSettings.cls
```diff
@@ -0,0 +1,97 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Org Settings class in NPSP.
+*/
+public with sharing class OrgSettings {
```
force2b, line 35:
```suggestion
public inherited sharing class OrgSettings {
```

force2b, line 35: Is the intent here that this will hold all general sfdc org related config methods?
Wondering if this should be an instance class, which might open up some mocking (via injection) options later?

src/classes/NPSP_Address.cls
```diff
@@ -0,0 +1,399 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Address entity class in NPSP.
+*/
+public inherited sharing class NPSP_Address {
+
+    private Address__c addr;
```
bdvorachek, line 37: Would this SObject eventually be replaced with member variables?

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1133 @@
...
+    */
+    public static Boolean shouldUpdateAddress(SObject sObj, Address__c addr) {
+        if (sObj == null || addr == null) {
+            return false;
+        }
+        Boolean isContact = (sObj.getSObjectType() == Contact.SObjectType);
+        String prefix = (isContact ? 'Mailing' : 'Billing');
+
+        Boolean isDifferent = (
+                !equalsCaseSensitive((String) sObj.get(prefix + 'Street'), getMultilineStreet(addr)) ||
```
bdvorachek, line 170: This looks like another spot that we'd want to push down into the NPSP Address Entity.

src/classes/Addresses.cls
```diff
@@ -0,0 +1,1133 @@
...
+
+        return mapAccIdHHInfo;
+    }
+
+    /*******************************************************************************************************
+    * @description helper class to keep track of the default and seasonal address for a household
+    */
+    public class HHInfo {
+        public Id accId;
+        public Address__c addrDefault;
```
bdvorachek, line 379: Would this eventually use NPSP_Address as a member instead of the SObject?

kenlewis-sfdo: Great question. This class actually looks like a data structure intended to hold a subset of the information that w should be in an NPSP_Account.  I would think you would want to access the default addresfor an account through a method like npspAccount.defaultAddress().  And I'm not exactly surhow this object is used with a single seasonal address since an account could have multiple seasonal addresses.  I don't know that I'll get into replacing this data structure in this work stream but it is a great callout and something to look at for potential future refactoring.

src/classes/OrgConfig.cls
```diff
@@ -63,6 +76,10 @@ public with sharing class OrgSettings {
         private set;
     }
 
+    public Boolean isStateCountryPicklistsEnabled() {
```
force2b, line 79: Why still static?

src/classes/TDTM_Runnable.cls
```diff
@@ -148,10 +148,18 @@ global abstract class TDTM_Runnable {
         * @return void
         */
         public void groupByType() {
-            objectsToInsert = groupSObjects(objectsToInsert);
-            objectsToUpdate = groupSObjects(objectsToUpdate);
-            objectsToDelete = groupSObjects(objectsToDelete);
-            objectsToUndelete = groupSObjects(objectsToUndelete);
+            if (!objectsToInsert.isEmpty()) {
```
force2b, line 151: I recommend using the null safe operator in these if blocks just in case
```suggestion
            if (objectsToInsert?.isEmpty() == false) {
```

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -216,6 +216,17 @@ public class TDTM_TriggerHandler {
...
 
+    private static Map<String, Type> typeMap = new Map<String, Type>();
+    private static Type getTypeForClassNamed(String str) {
+        if (typeMap.get(str) != null) {
+            return typeMap.get(str);
+        } else {
+            Type typeInst = Type.forName(str);
+            typeMap.put(str, typeInst);
+            return typeInst;
+        }
```
force2b, line 227:
```suggestion
        if (typeMap.get(str) == null) {
            Type typeInst = Type.forName(str);
            typeMap.put(str, typeInst);
        }
        return typeMap.get(str);
```

force2b (commented):
Looking good. Made a few code change recommendations. I'll need to do a full review again once you're ready for it.

src/classes/Addresses.cls
kenlewis-sfdo: Yep, great call.  I just moved things around a bit.  Check out the latest commit and let me know what you think.

src/classes/OrgConfig.cls
kenlewis-sfdo: Should be fixed now!  About to push up one more round of changes then will stop for final review!

src/classes/TDTM_Runnable.cls
kenlewis-sfdo: These variables are instantiated as empty lists.  Is the null safe operator necessary?

src/classes/OrgSettings.cls
kenlewis-sfdo: Yep! This should be finished now!

kenlewis-sfdo:
Keep the feedback coming if you have any! I will probably only make absolutely necessary changes at this point, but comment if you see anything to keep in mind for future work or have questions!

src/classes/TDTM_Runnable.cls
force2b: If you're confident the null check is never required, then no.

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -216,6 +216,15 @@ public class TDTM_TriggerHandler {
         return null;
     }
 
+    private static Map<String, Type> typeMap = new Map<String, Type>();
+    private static Type getTypeForClassNamed(String str) {
+        if (typeMap.get(str) == null) {
```
force2b, line 221: Just in case the class name that comes from the Trigger_Handler__c object is not always in the same case, maybe force it to lower case:
```suggestion
        str = str.toLowerCase();
        if (typeMap.get(str) == null) {
```

force2b (commented):
Just a couple of quick questions/comments.

bdvorachek (approved):
Thanks for everything that went into this Ken! I had one comment around an unnecessary `this.getRecord()` in NPSP_Address, but overall this checks out. Big improvements to readability and organization. The stuff that sticks out to me yet are some of the longer methods, specifically refreshCurrentHHAddress.

src/classes/TDTM_Runnable.cls
kenlewis-sfdo: Just double checked - they definitely can be set to null, and the fields are exposed globally instead of accessed using setters so the null safe operator is the way to go!  Just pushed that up in a commit.  Thank you for the catch and suggestion.

src/classes/NPSP_Address.cls
```diff
@@ -82,6 +93,78 @@ public inherited sharing class NPSP_Address {
...
+        if (addr.Address_Type__c != null
+                && strCleanup(addr.Address_Type__c) != strCleanup(otherAddress.Address_Type__c)
+                && otherAddress.Address_Type__c != null) {
+            cChange++;
+        }
+        return cChange <= 1;
+    }
+
+    public String getAddrKey() {
+        Address__c addr = this.getRecord();
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yep, good call, Ben!  Thank you again for the review, contribution and collaboration!

force2b (approved):
🎉 Awesome! 🎉

PR #6338, opened 2021-03-29
RD2: Removing Pilot feature flag check for editing Elevate Recurring Donations

lparrott:
# Critical Changes

# Changes

- We added the ability to edit the Credit Card information of Recurring Donations connected to Elevate.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/rd2ElevateInformation/rd2ElevateInformation.html
+++ b/src/lwc/rd2ElevateInformation/rd2ElevateInformation.html
@@ -105,54 +105,52 @@
                                 </div>
                             </lightning-layout-item>
                         </template>
-                        <template if:true={isElevateCustomer}>
-                            <template if:true={showLastFourACH}>
-                                <lightning-layout-item size="4" flexibility="auto"
-                                                    class="slds-dl_horizontal__label slds-text-color_weak" slds-m-bottom_x-small>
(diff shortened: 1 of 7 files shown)


src/lwc/rd2ElevateInformation/rd2ElevateInformation.html
```diff
@@ -105,7 +105,7 @@ <h2 slot="title" class="slds-p-top_xx-small" data-qa-locator={qaLocatorHeader}>
                                 </div>
                             </lightning-layout-item>
                         </template>
-                        <template if:true={isEditEnabled}>
+                        <template if:true={isElevateCustomer}>
```
andrewyu-salesforce, line 108: I think this conditional render can be completely remove since we already checking if the record is a Elevate record above in line 58

andrewyu-salesforce (commented):
@lparrott Everything looks great. Just a small suggestion on removing one of the conditional rendering .

src/lwc/rd2ElevateInformation/rd2ElevateInformation.html
lparrott: Just pushed an update with this change

andrewyu-salesforce (approved):

PR #6348, opened 2021-04-01
Feature/232  ptewson show auth expired message on bge load

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/labels/CustomLabels.labels
```diff
@@ -8634,6 +8634,22 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <value>Payment Authorization Expired</value>
+    </labels>
+    <labels>
+        <fullName>gePaymentAuthExpiredWarningText</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Warning about processing batch when authorizations have expired</shortDescription>
+        <value>One or more card authorizations expired. To ensure these payments are charged when processing the batch, reenter the payment information for donations with Expired Authorization Payment Status.</value>
+    </labels>
```
ptewson-sfdo, line 8652: @melissabarber labels for review!

melissabarber: Look splendiferous! Thanks.

PS. We may need to update the end of the warning message depending on what we settle on for indicating expired authorizations.

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -76,6 +78,8 @@ describe('c-ge-batch-gift-entry-header', () => {
             getGiftBatchTotalsBy.mockResolvedValue(APEX_GIFT_BATCH_TOTALS_BY_SUCCESS);
```
psadie-sfdo, line 78: These `mockResolvedValue` can be removed since we're no longer importing that apex method in this component.

Instead, we'll need to populate the public `batchTotals` property with the values needed to verify that the elements this jest tests expects gets rendered appropriately.

Something like below perhaps
```
element.batchTotals = {
    hasValuesGreaterThanZero: true,
    totalGifts: 20,
    processedGifts: 10,
    failedPayments: 0,
    failedGifts: 5,
    expiredPayments: 0
}
```

psadie-sfdo: above keys are wrong
```
element.batchTotals = {
                hasValuesGreaterThanZero: true,
                totalGiftsCount: 20,
                processedGiftsCount: 10,
                failedPaymentsCount: 0,
                failedGiftsCount: 5,
                hasPaymentsWithExpiredAuthorizations: false
            }
```

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -51,9 +30,18 @@ describe('c-ge-batch-gift-entry-header', () => {
...
 
+            element.batchTotals = {
+                hasValuesGreaterThanZero: true,
+                hasPaymentsWithExpiredAuthorizations: false,
+                totalGiftsCount: 20,
+                processedGiftsCount: 10,
+                failedPaymentsCount: 0,
+                failedGiftsCount: 5,
+                expiredPaymentsCount: 0
+            }
```
psadie-sfdo, line 43: This block isn't strictly necessary for this test since we're confirming that the expected title is rendered per the record that's returned by the getRecord wire adapter. `batchTotals` is optional for this test and I would recommend we remove it.

```suggestion
```

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -64,35 +52,73 @@ describe('c-ge-batch-gift-entry-header', () => {
...
```
psadie-sfdo, line 65: Similarly here. Testing that the batch related buttons are rendered.

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -64,35 +52,73 @@ describe('c-ge-batch-gift-entry-header', () => {
...
             const element = setup();
 
+            element.batchTotals = {
+                hasValuesGreaterThanZero: false,
+                hasPaymentsWithExpiredAuthorizations: false,
+                totalGiftsCount: 20,
+                processedGiftsCount: 0,
+                failedPaymentsCount: 0,
+                failedGiftsCount: 0,
+                expiredPaymentsCount: 0
```
psadie-sfdo, line 101:
```suggestion
                hasValuesGreaterThanZero: false,
                hasPaymentsWithExpiredAuthorizations: false,
                totalGiftsCount: 0,
                processedGiftsCount: 0,
                failedPaymentsCount: 0,
                failedGiftsCount: 0,
                expiredPaymentsCount: 0
```

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -64,35 +52,73 @@ describe('c-ge-batch-gift-entry-header', () => {
...
+            }
+
             await flushPromises();
             const headerDetailRows = element.shadowRoot.querySelectorAll('c-util-page-header-detail-row');
             expect(headerDetailRows.length).toBe(0);
         });
 
         it('renders detail blocks for records processed and records failed with correct record counts on load', async () => {
```
psadie-sfdo, line 109: I think we could duplicate this jest test to add coverage for the Elevate payment-related page header detail blocks.

src/lwc/geBatchGiftEntryHeader/geBatchGiftEntryHeader.js
```diff
@@ -1,10 +1,8 @@
 import { LightningElement, api, wire, track } from 'lwc';
```
psadie-sfdo, line 1: Remove unused import.

```suggestion
import { LightningElement, api, wire } from 'lwc';
```

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -272,7 +272,7 @@ export default class GeFormRenderer extends LightningElement{
             });
         registerListener('paymentError', this.handleAsyncWidgetError, this);
         registerListener('doNotChargeState', this.handleDisableElevateWidgetState, this);
-        registerListener('geDonationMatchingEvent', this.handleChangeSelectedDonation, this);
+        registerListener('geModalCloseEvent', this.handleChangeSelectedDonation, this);
```
psadie-sfdo, line 275: 🙏

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -65,6 +71,37 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+                },
+                componentProperties: {
+                    'severity': 'warning',
+                    'title': this.CUSTOM_LABELS.gePaymentAuthExpiredHeader,
+                    'message': this.CUSTOM_LABELS.gePaymentAuthExpiredWarningText,
+                    'buttonText': this.CUSTOM_LABELS.commonOkay
+                },
+            };
+            this.dispatchEvent(new CustomEvent('togglemodal', { detail }));
+            this._hasPaymentsWithExpiredAuthorizations = true;
```
psadie-sfdo, line 100: Is this to guard against the modal popping up multiple times?

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -51,9 +30,18 @@ describe('c-ge-batch-gift-entry-header', () => {
 
     describe('getRecord @wire data', () => {
         it('renders user record details', async () => {
```
psadie-sfdo, line 32:
```suggestion
        it('renders the data import batch record name as the title', async () => {
```

psadie-sfdo (changes requested):
Thanks for all this work Patrick! Looking really good! The comments below are from my read-through. Currently deploying and will do a bit of manual testing as well

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
ptewson-sfdo: Yes - and definitely welcome feedback for this approach. Without this in place, you would see the modal pop up every time you added a gift and there was an expired auth present, since this function is called on the geBatchGiftEntryTableChangeEvent. I believe that the intent is that you only see this notification once-per-page-load, but this feels a little bit clunky and maybe there's a better way to accomplish the goal?

psadie-sfdo: Gotcha, thank you for clarifying! I think this is fine. My only suggestion would be to update the name to be specific to the modal display. Maybe along the lines of `_hasShownExpiredAuthorizationWarning` or something?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -65,6 +71,37 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+
+    disconnectedCallback() {
+        unregisterListener('geBatchGiftEntryTableChangeEvent', this.retrieveBatchTotals, this);
+    }
+
+    async retrieveBatchTotals() {
+        this.batchTotals = await BatchTotals(this.batchId);
+
+        if(this.batchTotals.hasPaymentsWithExpiredAuthorizations &&
+            !this._hasPaymentsWithExpiredAuthorizations) {
```
voduyemi, line 86: Can we extract this gaurd condition into its own function?

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -65,6 +71,37 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+            const detail = {
```
voduyemi, line 87: This can also be extracted into its own function, we could call it `buildModalDetail`

src/lwc/geModalPrompt/geModalPrompt.js
```diff
@@ -0,0 +1,40 @@
...
+    @api message;
+    @api buttonText;
+
+    handleCloseModal() {
+        fireEvent(this.pageRef, 'geModalCloseEvent', {});
+    }
+
+    get titleSectionComputedClass() {
+        let baseClass = ['slds-box', 'slds-theme_alert-texture', 'slds-box_extension'];
+
```
voduyemi, line 17: I think this getter can be simplified to 
        ```
        if (isEmpty(this.severity)) {
            baseClass.push('slds-theme_default');
        }
        baseClass.push($`slds-theme_`+ this.severity);
        return baseClass.join(' ');
        ```

voduyemi (changes requested):
Looks good Patrick! Just had some comments around extracting some if statements and object creation. Also refactoring a function.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
ptewson-sfdo: @voduyemi Do you think it's better to build the detail in the function and then call the dispatch event here, or should the function be something like `displayExpiredAuthorizationWarningModal` and handle the dispatch of the togglemodal event as well?

src/lwc/geModalPrompt/geModalPrompt.js
ptewson-sfdo: @voduyemi I was thinking similarly, but that doesn't do as much to enforce the vocabulary as the individual else-if checks. But maybe there's a combination approach that avoids all the else-ifs but checks against the supported list of LWC classes!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
voduyemi: I think I like the idea of naming it `displayExpiredAuthorizationWarningModal` and handling and dispatching in that function

src/lwc/geModalPrompt/geModalPrompt.js
voduyemi: Very true, we can have an array of severity levels and just check to see if the string passed in matches anything in the list before we push to `baseClass`, if not we push the default theme.

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
ptewson-sfdo: @psadie-sfdo - I might be mis-reading this, but I believe the elevate-specific header details aren't yet visible in this WI? That story is still in the backlog I believe!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -351,6 +385,23 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
         }));
     }
 
+    displayExpiredAuthorizationWarningModal() {
```
ptewson-sfdo, line 388: @voduyemi - I went with this approach for displaying the modal - not just building the modal detail but also firing the togglemodal event within this method. Let me know if you would prefer to see this broken down with this function just returning the modal detail structure!

src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
psadie-sfdo: 🤦‍♂️ sorry, that's right! Resolved!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -197,6 +219,11 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
         return format(geBatchGiftsHeader, [this.batchName]);
     }
 
+    get shouldDisplayExpiredAuthorizationWarning() {
+        return this.batchTotals.hasPaymentsWithExpiredAuthorizations 
+            && !this._hasDisplayedExpiredAuthorizationWarning;
+    }
```
psadie-sfdo, line 225: No need for a getter here I think. Just a normal function.
```suggestion
    shouldDisplayExpiredAuthorizationWarning() {
        return this.batchTotals.hasPaymentsWithExpiredAuthorizations 
            && !this._hasDisplayedExpiredAuthorizationWarning;
    }
```

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -65,6 +71,22 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+
+    disconnectedCallback() {
+        unregisterListener('geBatchGiftEntryTableChangeEvent', this.retrieveBatchTotals, this);
+    }
+
+    async retrieveBatchTotals() {
+        this.batchTotals = await BatchTotals(this.batchId);
+
+        if(this.shouldDisplayExpiredAuthorizationWarning) {
```
psadie-sfdo, line 85: Please see comment for lines 222 - 225.
```suggestion
        if (this.shouldDisplayExpiredAuthorizationWarning()) {
```

psadie-sfdo (changes requested):
One final change request from my perspective and possibly a jest test for `geGiftEntryFormApp` to check that the `toggleModal` event is dispatched/called if we can squeeze it in.

Everything else looks great! Thanks Patrick!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
voduyemi: This works for me. Thanks!

psadie-sfdo (approved):
Looks good! Thanks Patrick!

voduyemi (approved):
Looks good @ptewson-sfdo !

src/lwc/geModalPrompt/__tests__/geModalPrompt.test.js
```diff
@@ -0,0 +1,66 @@
+import { createElement } from 'lwc';
+import GeModalPrompt from 'c/geModalPrompt';
+
+describe('c-ge-modal-prompt', () => {
```
psadie-sfdo, line 4: Nioce! 👍

ptewson-sfdo: It is a humble thing, but I am very proud!

PR #6351, opened 2021-04-01
Display Elevate widget in BGE

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.html
+++ b/src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.html
@@ -1,7 +1,7 @@
 <template>
     <article class='slds-card slds-card_extension slds-m-vertical_small'>
-        <div if:true={ready}>
-            <div if:true={hasData}>
+        <template if:true={ready}>
+            <template if:true={hasData}>
                 <div if:false={isLoaded}
                      class="slds-is-relative">
                     <lightning-spinner alternative-text="Loading..."
@@ -69,14 +69,14 @@
                                          data-qa-locator="datatable Batch Gifts">
                     </lightning-datatable>
                 </div>
-            </div>
-            <div if:false={hasData}>
+            </template>
+            <template if:false={hasData}>
                 <c-util-illustration title="Looks like you haven't entered any gifts yet!"
                                      message="Enter a new gift and it will appear here."
                                      size="small"
                                      variant="lake-mountain">
                 </c-util-illustration>
-            </div>
-        </div>
+            </template>
+        </template>
     </article>
 </template>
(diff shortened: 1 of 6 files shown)


src/lwc/geBatchGiftEntryTable/geBatchGiftEntryTable.html
```diff
@@ -27,58 +27,56 @@ <h1 class="slds-text-title_bold">{title}</h1>
...
+                                         enable-infinite-loading
+                                         onloadmore={loadMoreData}
+                                         data-qa-locator="datatable Batch Gifts">
                     </lightning-datatable>
                 </div>
-            </div>
-            <div if:false={hasData}>
+            </template>
+            <template if:false={hasData}>
                 <c-util-illustration title="Looks like you haven't entered any gifts yet!"
```
voduyemi, line 74: I know this isn't part of your WI? Do you know why we're not using a custom label here?

src/lwc/geFormService/geFormService.js
```diff
@@ -42,6 +43,24 @@ class GeFormService {
     fieldTargetMappings;
     donationFieldTemplateLabel;
 
+    getPaymentTransactionStatusEnums = async () => {
```
voduyemi, line 46: Love the move here to `geFormService`. I'm just a little worried about the `handleError` function here, In `geFormTokenizeCard`, the `handleError` is tailored toward payment/transaction errors and is a banner inside of the widget, but if we start using this we start popping up toast error messages instead.

Also, side note, should we be looking to move this handleError function to a more generic class? We keep importing `utilTemplateBuilder` everywhere.

voduyemi (commented):
Looks good to me, just had a few comments around custom labels and reshuffling of methods.

src/lwc/geFormService/geFormService.js
```diff
@@ -42,6 +43,24 @@ class GeFormService {
...
+        try {
+            return Object.freeze(
+                JSON.parse(await getPaymentTransactionStatusValues())
```
dfuller114, line 49: Adding this comment from our discussion this morning for posterity. What do you think about creating an ElevateTransactionStatus javascript class that freezes an object containing hard coded values? I know Javascript and enums don't quite work the same way, but I think my motivation for doing so would moreso be to maintain consistency in matching types between client and server. I'm thinking ideally if we see a type in Apex we know is being used in the UI then we can know there's a corresponding javascript class for that type.

dfuller114 (commented):
Everything looks good to me. Added the comment about ElevateTransactionStatus from our conversation earlier to finalize on that. This isn't a blocker though.

src/lwc/geFormService/geFormService.js
dfuller114: @kenlewis-sfdo curious to get your thoughts on this.

psadie-sfdo: @dfuller114 @kenlewis-sfdo  - I merged this into the parent branch. Please drop comments into that branch for the full PR review. Thank you both!

PR #6354, opened 2021-04-02
geFormRenderer jest tests

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH0000009AUEYA2/view
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


tests/__mocks__/psElevateTokenHandler.js
```diff
@@ -0,0 +1,254 @@
...
+            paymentMethod: paymentMethod
+        };
+        this.sendIframeMessage(iframe, message, this._visualforceOrigin);
+
+        return mountPromise;
+    }
+
+
+    sendIframeMessage(iframe, message, targetOrigin) {
+        const reply = mockGetIframeReply(iframe, message, targetOrigin);
```
bdvorachek, line 251: Here's where we insert the mock for getting a message back from the iframe. Happens synchronously. Would like to be able to provide a mock for the iframe itself instead eventually.

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -238,19 +234,20 @@ class psElevateTokenHandler {
...
         return mountPromise;
     }
-}
 
-/**
- * A new instance of the Elevate tokenization handler
- */
-export default new psElevateTokenHandler();
+
+    sendIframeMessage(iframe, message, targetOrigin) {
```
bdvorachek, line 247: New method extracted out from psElevateTokenHandler.js, this method is overridden at the bottom of the mocked version of this class. [See here](https://github.com/SalesforceFoundation/NPSP/pull/6354/files#r606489039)

src/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -238,19 +234,20 @@ class psElevateTokenHandler {
...
+        this.sendIframeMessage(iframe, message, this._visualforceOrigin);
 
         return mountPromise;
     }
-}
 
-/**
- * A new instance of the Elevate tokenization handler
- */
-export default new psElevateTokenHandler();
```
bdvorachek, line 256: Intentionally converted this away from a singleton, any time this component is referenced we can safely `new` it up. Nobody shares references to it.

jest.config.js
```diff
@@ -2,8 +2,16 @@ const { jestConfig } = require('@salesforce/sfdx-lwc-jest/config');
 module.exports = {
     ...jestConfig,
     moduleNameMapper: {
+        "^c/psElevateTokenHandler$": "<rootDir>/tests/__mocks__/psElevateTokenHandler",
```
bdvorachek, line 5: This allows us to have a mocked psElevateTokenHandler inside tests. The whole class is replaced with this implementation, which is as close to actual as possible minus using the iframe's postMessage.

tests/__mocks__/psElevateTokenHandler.js
```diff
@@ -0,0 +1,257 @@
...
+        fireEvent(null, eventName, payload);
+    }
+
+    /***
+     * @description Listens for a message from the Visualforce iframe.
+     * Rejects any messages from an unknown origin.
+     */
+    registerPostMessageListener(component) {
+        const self = this;
+        this._componentRef = component;
```
bdvorachek, line 103: This is different from the original implementation. Since we're not using `window.onmessage` we need a reference to the original component in order to call `handleMessage` on it.

psadie-sfdo (approved):
Looking pretty good! Ran through and was able to tokenized and capture a payment in SGE. 

Not a blocker, but could we prettier all the new JS files?

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -48,7 +48,7 @@ export default class rd2ElevateCreditCardForm extends LightningElement {
     _paymentMethod = undefined;
     _nextDonationDate = undefined;
     _isEditPayment = false;
-    
+
```
dfuller114, line 51: I didn't see any changes in this file other than whitespace. Did we mean this file to be in the PR?

dfuller114 (changes requested):
just one quick comment on whether an rd2 file is needed here but everything else looks good to me!

dfuller114 (approved):
Looks good to me. Thanks Ben!

src/lwc/geFormRenderer/__tests__/data/dataImportObjectInfo.json
```diff
@@ -0,0 +1,5200 @@
+{
```
ptewson-sfdo, line 1: How do you generate this - just building a DI out on the platform and saving off the serialization? I'm curious how much ongoing maintenance we might have to do, if we add fields to DI in the future?

sokosaurus (approved):
GREAT WORK!

src/lwc/geFormRenderer/__tests__/data/dataImportObjectInfo.json
bdvorachek: We'd need to update the dataImportMock whenever we update the Template Mock to include a new field, because we use this mock to check for CRUD access to fields on the template.

PR #6367, opened 2021-04-12
BGE Header Updates - Elevate

sokosaurus:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geBatchGiftEntryHeader/geBatchGiftEntryHeader.js
```diff
@@ -17,6 +18,8 @@ export default class GeBatchGiftEntryHeader extends LightningElement {
     @api batchTotals = {};
     @api isPermissionError;
 
+    isElevateCustomer;
```
voduyemi, line 21: Can we create a getter that returns the value of `isElevateCustomer` so that's referenced in the markup?

voduyemi (commented):
Looks good to me, I just have a comment around using a getter. I was thinking of how we're checking for elevate customers in this component, Do we think we can retrieve this information as we retrieve the `batchTotals` from the backend. I am just a little weary of our use of connected callbacks in child components. It was called out as a code smell when we met with the LWC team. Also I think this should be a bigger discussion with the team.

src/lwc/geBatchGiftEntryHeader/geBatchGiftEntryHeader.js
psadie-sfdo: An alternative would be to piggyback off of the parent component's existing connectedCallback and pass the `isElevateCustomer` value into the header component via a public prop.

psadie-sfdo (approved):
Looks good to me! No change requests! Thanks for the updates David!

voduyemi (approved):
Looks good @sokosaurus. Thanks!

ptewson-sfdo (approved):
Looks good from my end!

PR #6371, opened 2021-04-13
GE: Handle Auth and Tokenize failures in BGE

ptewson-sfdo:
This work introduces handling and display of key errors as part of the Batch Gift Credit Card payment flow in Gift Entry, including tokenization errors, Data Import save, and Authorization failures

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -711,7 +712,8 @@ export default class GeFormRenderer extends LightningElement{
 
                 dataImportFromFormState = this.saveableFormState();
             } catch (ex) {
-                this.handleAsyncWidgetError(ex);
+                const errors = [{ message: buildErrorMessage(ex) }];
+                this.handlePurchaseCallValidationErrors(errors);
```
ptewson-sfdo, line 716: Not sure if we need to rename this function to be more generic or if I should be calling something else - this calls through to handleAsyncWidgetError() but does some additional transformation and checking and label replacement that is likely useful here. 

I'm also wondering if the call to `this.handleAsyncWidgetError(ex);` on line 673 of handleSave() should also do something similar - I think I've seen that come through as "unknown error" when there's an exception thrown, and using handlePurchaseXCallValidationErrors() or something similar would massage that in a way that gets it to display as we intend, I believe.

voduyemi (approved):
Looks good to me. Thanks! @ptewson-sfdo

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Yeah, I think that makes sense. We should probably also just pull that entire try/catch block for tokenizing inside handleSave() into its own method entirely.

psadie-sfdo:
Looks good!

psadie-sfdo (approved):
Looks good! Thanks, Patrick!

Reminder to drop a link to the related work item.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -691,27 +693,33 @@ export default class GeFormRenderer extends LightningElement{
     
                 const currentCaptureGroup = new ElevateCaptureGroup(this.latestCaptureGroupId);
                 const authorizedGift = await currentCaptureGroup.add(tokenizedGift);
+                if(authorizedGift.status === this.PAYMENT_TRANSACTION_STATUS_ENUM.AUTHORIZED) {
```
psadie-sfdo, line 696: 👍
```suggestion
                if (authorizedGift.status === this.PAYMENT_TRANSACTION_STATUS_ENUM.AUTHORIZED) {
```

psadie-sfdo (approved):
Still looks good after a quick smoke test

PR #6379, opened 2021-04-23
W-8985203 - RD GAU Allocations and Other Allocation Improvements

callen-sfdo:
- Added ability for GAUs to be linked to recurring donations and be cascaded down to donations and payments through the use of the ALLO_AllocationsService
- Disabled Allocations trigger during Data Import processing when there are GAU Allocations to be created.  This prevents incorrect defaults from being assigned before the GAUs are created in the additional object service
- Added basic validation check to ensure that total percent for the GAU allocations on the DI do not exceed 100%.  If they are over 100, an error will be added and the DI record will be failed.
- Added  validation check to ensure that there were no existing allocations on the donation listed in DonationImported__c and no allocations on payments for that donation.  If there are any, the DI record will be failed and an error message added.
-  Modified unit tests for Additional Objects since updating GAUs via BDI will no longer work due to the above validations.
-  Added data import field mapping to allow GAU Allocations to be mapped onto Recurring donations.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/BDI_CustObjMappingGAUAllocation.cls
+++ b/src/classes/BDI_CustObjMappingGAUAllocation.cls
@@ -44,6 +44,7 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
 
         String PAYMENT_FIELDNAME = SObjectType.Allocation__c.fields.Payment__c.Name;
         String OPPORTUNITY_FIELDNAME = SObjectType.Allocation__c.fields.Opportunity__c.Name;
+        String RECURRING_DONATION_FIELDNAME = SObjectType.Allocation__c.fields.Recurring_Donation__c.Name;
 
         String sourceObjName = SObjectType.DataImport__c.getName();
 
@@ -57,6 +58,7 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
 
             Object paymentIdValue;
             Object opportunityIdValue;
+            Object recurringDonationIdValue;
 
             for (BDI_FieldMapping fieldMapping : objWrap.fieldMappings) {
                 String sourceFieldName = fieldMapping.Source_Field_API_Name;
@@ -76,15 +78,19 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
                             paymentIdValue = value;
                         } else if (fieldMapping.Target_Field_API_Name == OPPORTUNITY_FIELDNAME) {
                             opportunityIdValue = value;
+                        } else if (fieldMapping.Target_Field_API_Name == RECURRING_DONATION_FIELDNAME) {
+                            recurringDonationIdValue = value;
                         } else {
                             castAndCopyField(objWrap, sourceFieldName, sourceFieldDescribe, targetFieldName, targetFieldDescribe);
                         }
                     }
                 }
             }
-
+            System.debug('CustObjMappingGAUAllocation: ' + settings.Payment_Allocations_Enabled__c);
             // If the payment is specified and payment allocations are enabled, always use that first.
-            if ( paymentIdValue != null && settings.Payment_Allocations_Enabled__c) {
+            if ( recurringDonationIdValue != null ) {
+                objWrap.sObj.put(RECURRING_DONATION_FIELDNAME,recurringDonationIdValue);
+            } else if ( paymentIdValue != null && settings.Payment_Allocations_Enabled__c) {
                 objWrap.sObj.put(PAYMENT_FIELDNAME,paymentIdValue);
             } else if ( opportunityIdValue != null ) {
                 objWrap.sObj.put(OPPORTUNITY_FIELDNAME,opportunityIdValue);
--- a/src/classes/BDI_CustObjMappingGAUAllocation.cls
+++ b/src/classes/BDI_CustObjMappingGAUAllocation.cls
@@ -86,7 +86,7 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
                     }
                 }
             }
-            System.debug('CustObjMappingGAUAllocation: ' + settings.Payment_Allocations_Enabled__c);
+
             // If the payment is specified and payment allocations are enabled, always use that first.
             if ( recurringDonationIdValue != null ) {
                 objWrap.sObj.put(RECURRING_DONATION_FIELDNAME,recurringDonationIdValue);
--- a/src/classes/BDI_GAUAllocationsUtil.cls
+++ b/src/classes/BDI_GAUAllocationsUtil.cls
@@ -1,0 +1,252 @@
+/*
+    Copyright (c) 2021 Salesforce.org
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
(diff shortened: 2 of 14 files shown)


src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -76,15 +78,19 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
...
+                        } else if (fieldMapping.Target_Field_API_Name == RECURRING_DONATION_FIELDNAME) {
+                            recurringDonationIdValue = value;
                         } else {
                             castAndCopyField(objWrap, sourceFieldName, sourceFieldDescribe, targetFieldName, targetFieldDescribe);
                         }
                     }
                 }
             }
-
+            System.debug('CustObjMappingGAUAllocation: ' + settings.Payment_Allocations_Enabled__c);
```
grivero, line 89: do you want to leave this for  future testing purposes?

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,252 @@
...
+                } else {
+                    disByOpptId.put(di.DonationImported__c,new DataImport__c[]{di});
+                }
+            }
+        }
+
+        // If there are disByOpptId then query existing Allocations
+        if (!disByOpptId.isEmpty()) {
+            Set<Id> opptIds = disByOpptId.keySet();
+            String queryString = ' SELECT Id, ' +
```
grivero, line 130: why are we using query strings instead of fixed queries?

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,252 @@
...
+                    disByOpptId.get(di.DonationImported__c).add(di);
+                } else {
+                    disByOpptId.put(di.DonationImported__c,new DataImport__c[]{di});
+                }
+            }
+        }
+
+        // If there are disByOpptId then query existing Allocations
+        if (!disByOpptId.isEmpty()) {
+            Set<Id> opptIds = disByOpptId.keySet();
```
grivero, line 129: checkmarks usually requires to check for != null every set/list that is part of where clause.

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,252 @@
...
+                    'Opportunity__c,' +
+                    'Payment__c,' +
+                    'Payment__r.npe01__Opportunity__c '+
+                    'FROM Allocation__c WHERE Opportunity__c IN: opptIds';
+
+            if (alloSettings.Payment_Allocations_Enabled__c) {
+                queryString += ' OR Payment__r.npe01__Opportunity__c IN: opptIds';
+            }
+
+            Allocation__c[] existingAllos = Database.query(queryString);
```
grivero, line 141: We should check FLS

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,252 @@
...
+                    } else if (di.DonationImported__c != null) {
+                        statusFieldName = SObjectType.DataImport__c.fields.DonationImportStatus__c.Name;
+                    }
+
+                    bdi.LogBDIError(di, errorObj.Full_Message__c, statusFieldName);
+                }
+            }
+            // Delete any allocations that were created for an oppt that had related allocation errors.
+            if (allocForDelete.size() > 0) {
+                if (UTIL_Permissions.canDelete(String.valueOf(Allocation__c.getSObjectType()), false)) {
```
grivero, line 246: 👍

src/classes/ERR_RecordError.cls
```diff
@@ -43,8 +43,8 @@ public class ERR_RecordError {
 
     private String context;
 
-    private Map<ID, Error__c> insRecordErrors = new Map<Id, Error__c>();
-    private Map<ID, Error__c> updRecordErrors = new Map<Id, Error__c>();
+    public Map<ID, Error__c> insRecordErrors = new Map<Id, Error__c>();
```
grivero, line 46: if this change was for testing purposes maybe we can use @TestVisible flag. 
If for some reason there's any security concern we could also create get/set methods to avoid changing visibility

src/labels/CustomLabels.labels
```diff
@@ -3890,6 +3890,30 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>error message if the donation matched opp(s), but behavior required no match</shortDescription>
         <value>The Donation information matched against one or more Opportunities, but the setting required no matches, so the NPSP Data Importer did not import it. The Donation Possible Matches field contains the Salesforce Ids of the matched Opportunities.</value>
     </labels>
+    <labels>
+        <fullName>bdiErrorExistingDonationAllocations</fullName>
+        <categories>Data Importer</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>bdiErrorExistingDonationAllocations</shortDescription>
+        <value>There are existing allocations on the donation which may conflict with the GAU allocations entered.</value>
```
grivero, line 3899: @sam-knox - Could you review these 3 labels?

grivero (commented):
Very good overall, I left some comments/questions. 
Remember to check for FLS when performing queries and CRUD when performing DML operations. 
Sam Knox probably wants to review those labels too (In case he haven't already)

src/labels/CustomLabels.labels
callen-sfdo: I mentioned sam on the ticket, and explained the use cases, these are placeholder until he has reviewed.

src/classes/ERR_RecordError.cls
callen-sfdo: @grivero This is necessary so I can extract the errors from the instance of the ALLO_AllocationService class when working with it in the BDI_GAUAllocationsUtil code.  There isn't a particular reason these needed to be private, and I ran it by John brandolini who originally wrote the code and he didn't see an issue with it.

src/classes/BDI_GAUAllocationsUtil.cls
callen-sfdo: @grivero This was necessary because the last part of the query is conditional on whether payment allocations is enabled.

src/classes/BDI_CustObjMappingGAUAllocation.cls
callen-sfdo: @grivero I'll take it out

src/labels/CustomLabels.labels
sam-knox: @callen-sfdo - I posted some questions on the ticket but can post here too. Does this text accurately capture the errors being reported here?

1: "The provided GAU Allocations exceed 100 percent." 

2: "The existing GAU Allocations on the donation Opportunity record and the provided GAU Allocations exceed 100 percent." 

3: "The existing GAU Allocations on the donation Payment record and the provided GAU Allocations exceed 100 percent."

In cases 2 and 3, the errors occur only when the provided allocations exceed 100%, right? So if there was an existing Allocation on the donation at 50% and the DI had another at 50%, we shouldn't see the error?

src/classes/BDI_GAUAllocationsUtil.cls
callen-sfdo: @grivero Based on checking that the disByOpptId isn't empty then the keyset would have to be not null as well.  Lets see if it complains.

src/labels/CustomLabels.labels
callen-sfdo: @sam-knox for 2 and 3 there is an issue whenever there are other existing allocations on the donation regardless of percent.

grivero (approved):

src/labels/CustomLabels.labels
sam-knox: @callen-sfdo - Ok thanks, then I'd like to revise messages 2 and 3 in a future work item. 

For the customer, I suppose they'd have to remove the existing allocations on either the opp/payment or from the DI and re-process.

PR #6380, opened 2021-04-23
RD2: Adding Last Elevate Event Version to new RDs and checking version on Update

lparrott:
We are now adding the Elevate Event Version to an Elevate RD when it is created, and comparing the Event version when updates are sent from Elevate.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_ElevateIntegrationService_TEST.cls
+++ b/src/classes/RD2_ElevateIntegrationService_TEST.cls
@@ -632,10 +632,12 @@ public class RD2_ElevateIntegrationService_TEST {
 
         npe03__Recurring_Donation__c newRd = getElevateRecurringDonationBaseBuilder()
             .withStatusClosed()
+            .withLastElevateVersionPlayed(2)
             .build();
-        npe03__Recurring_Donation__c oldRd = newRd.clone();
-        newRd.LastElevateVersionPlayed__c = 2;
-        oldRd.LastElevateVersionPlayed__c = 1;
+        npe03__Recurring_Donation__c oldRd = getElevateRecurringDonationBaseBuilder()
+            .withStatusClosed()
+            .withLastElevateVersionPlayed(1)
+            .build();
 
         System.assertEquals(
             true,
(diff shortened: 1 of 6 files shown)


src/classes/RD2_ElevateIntegrationService_TEST.cls
```diff
@@ -654,8 +654,8 @@ public class RD2_ElevateIntegrationService_TEST {
             .withStatusClosed()
             .build();
         npe03__Recurring_Donation__c oldRd = newRd.clone();
-        newRd.LastElevateEventPlayed__c = 'event-v1';
-        oldRd.LastElevateEventPlayed__c = 'event-v1';
+        newRd.LastElevateVersionPlayed__c = 1;
```
npsp-reedestockton, line 657: Given your code, I know it will pass, but as defense against future code modifications, it might be good to write a unit test for the case where the previous version is null.

npsp-reedestockton (changes requested):

src/classes/RD2_ElevateIntegrationService_TEST.cls
```diff
@@ -634,8 +634,8 @@ public class RD2_ElevateIntegrationService_TEST {
             .withStatusClosed()
             .build();
         npe03__Recurring_Donation__c oldRd = newRd.clone();
-        newRd.LastElevateEventPlayed__c = 'event-v2';
-        oldRd.LastElevateEventPlayed__c = 'event-v1';
+        newRd.LastElevateVersionPlayed__c = 2;
```
andrewyu-salesforce, line 637: I think it's worth it to add new a method `withLastElevateVersionPlayed(Integer)` in TEST_RecurringDonationBuilder to update this field

andrewyu-salesforce (changes requested):
@lparrott The change looks great. A small recommendation on updating LastElevateVersionPlayed__c

src/classes/RD2_ElevateIntegrationService_TEST.cls
lparrott: Good idea, I just pushed an update with this change.

src/classes/RD2_ElevateIntegrationService_TEST.cls
lparrott: Great find, I pushed an update to test these scenarios.

npsp-reedestockton (approved):

npsp-reedestockton:
Nice @lparrott!!  Looks good to me!

npsp-reedestockton:
@lparrott @mattwashington 

At this point, the only time we check LastElevateVersionPlayed__c is when we are cancelling a recurring donation.  I don't know whether it's worth clarifying that in our description or not, but thought it was worth mentioning.

andrewyu-salesforce (approved):

PR #6386, opened 2021-04-28
[RD2] Support for Elevate ACH Commitments

bdvorachek:
WI - https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000094WViIAM/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

- We've added ACH as an available payment method for new Recurring Donations.
# New Metadata
- Label - RD2_NextACHPaymentDonationDateInfo
- Label - RD2_EntryFormSaveACHMessage
- LWC - FormattedTextWithDate
# Deleted Metadata


src/classes/PS_CommitmentRequest.cls
```diff
@@ -330,6 +345,12 @@ public inherited sharing class PS_CommitmentRequest {
         return accounts.isEmpty() ? null : accounts[0];
     }
 
+    public with sharing class AchData {
```
bdvorachek, line 348: I'd like to combine this with the Gift Entry version of AchData eventually. I think there's an opportunity for having a class whose sole purpose is containing the classes that represent the messages we send to Elevate.

src/classes/PS_CommitmentRequest.cls
```diff
@@ -416,6 +437,38 @@ public inherited sharing class PS_CommitmentRequest {
             return this;
         }
 
+        public RequestBody withPaymentMethodType(String paymentMethodType) {
+            switch on paymentMethodType {
+                when 'ACH' {
```
bdvorachek, line 442: These strings represent the RD2 payment method field's picklist values.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -9,19 +9,24 @@ import spinnerAltText from '@salesforce/label/c.geAssistiveSpinner';
...
 /***
 * @description Event name fired when the Elevate credit card widget is displayed or hidden
 * on the RD2 entry form
 */
 const WIDGET_EVENT_NAME = 'rd2ElevateCreditCardForm';
 
-const TOKENIZE_CREDIT_CARD_EVENT_ACTION = 'createToken';
+const CREATE_TOKEN_ACTION = 'createToken';
+const CREATE_ACH_TOKEN_ACTION = 'createAchToken';
+const ELEVATE_PAYMENT_METHODS = ['ACH', 'Credit Card'];
```
bdvorachek, line 29: These are RD2 Payment Method picklist values again.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -38,31 +43,57 @@ export default class rd2ElevateCreditCardForm extends LightningElement {
...
 
     set paymentMethod(value) {
+        this.notifyIframePaymentMethodChanged(value);
         this._paymentMethod = value;
     }
 
+    notifyIframePaymentMethodChanged(newValue) {
+        const iframe = this.selectIframe();
+        if (isNull(iframe)) {
+            return;
```
bdvorachek, line 83: No frame? No Problem. If the frame isn't loaded, it should get the correct value when it loads so we don't need to notify.

src/lwc/rd2EntryForm/__tests__/data/recurringDonationObjectInfo.json
```diff
@@ -0,0 +1,1677 @@
+{
```
bdvorachek, line 1: Some of the test data in this PR could get trimmed down a bit. There's a lot of describe info that we don't need.

src/lwc/rd2EntryForm/rd2EntryForm.html
```diff
@@ -141,9 +151,15 @@ <h3 class="slds-section__title slds-theme_shade">
                                             is-edit-mode={isEdit}
                                             card-exp-date={cardExpDate}
                                             card-last-four-label={cardLastFourLabel}
+                                            ach-last-four-label={achLastFourLabel}
                                             card-last-four={cardLastFour}
+                                            ach-last-four={achLastFour}
                                             is-edit-payment={isEdit}
```
bdvorachek, line 157: We have isEdit bound twice here (151 and 157) - I'd like to try to de-duplicate it but when I did it broke the rd2EditPaymentInformationModal component, nothing rendered.

andrewyu-salesforce: The isEditMode is the toggle between editing and displaying the edit information. isEditPayment is the option to opt out and not use Elevate. I think we can have a better naming for isEditPayment to make it less confusing

andrewyu-salesforce (commented):
@bdvorachek Sorry for the late review. I went through the function testing and notice the Loading Text is not updated when the User is submitting the ACH information. This [line](https://github.com/SalesforceFoundation/NPSP/blob/5f27bf770d9aa0a401ad645f55a4ccd631461a0a/src/lwc/rd2EntryForm/rd2EntryForm.js#L540) should be set to saving ACH when the user is using ACH data

andrewyu-salesforce (approved):
looks great

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -370,22 +423,30 @@ export default class rd2EntryForm extends LightningElement {
...
+        const currencySupported = this.isCurrencySupported();
+        const countrySupported = this.isCountrySupported();
         this.isElevateWidgetEnabled = this.isElevateEditWidgetEnabled
             || (this.isElevateCustomer === true
             && !this.isEdit
-            && paymentMethod === PAYMENT_METHOD_CREDIT_CARD
-            && (this.scheduleComponent && this.scheduleComponent.getRecurringType() === RECURRING_TYPE_OPEN)
-            && this.isCurrencySupported()
-            && this.isCountrySupported());
+            && isValidPaymentMethod
```
lparrott, line 438: Another great refactor.

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -930,7 +1006,7 @@ export default class rd2EntryForm extends LightningElement {
     }
 
     /**
-     * @description Close the modal when Escape key is pressed
+     * @description Close the modal when Escsape key is pressed
```
lparrott, line 1009: Typo?

lparrott (approved):
Great work! I'm going to run through a few more scenarios, but I see no reason to hold up QE.

src/classes/PS_CommitmentRequest.cls
bdvorachek: I think we can do constants, but then we'd need to lose the switch statement because switch conditions need to be  known at run time, and static final vars don't qualify sadly. I'll move these strings into RD_Constants, and I think they'll be called `RD2_PAYMENT_METHOD_ACH` and `RD2_PAYMENT_METHOD_CREDIT_CARD`, and then this method will be an `if` statement rather than a `switch`.

lparrott: Interesting, I didn't know that about `switch`, I'm fine either way!

PR #6388, opened 2021-04-29
Feature/232  bge alert expired auths when batch open

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -389,23 +403,50 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
     }
 
+    displayExpiredAuthorizationWarningModalForProcessAndDryRun() {
+        this.displayModalPrompt ({
+                'variant': 'warning',
+                'title': this.CUSTOM_LABELS.gePaymentAuthExpiredHeader,
+                'message': this.CUSTOM_LABELS.gePaymentAuthExpiredWarningText,
+                'buttons': 
+                    [{
+                        label: 'Proceed Anyway',
```
ptewson-sfdo, line 438: Note: This needs to be a custom label

src/labels/CustomLabels.labels
```diff
@@ -8662,6 +8662,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Failure message if an invalid base64 string was provided</shortDescription>
         <value>The base64 format is invalid. The number of characters must be a multiple of 4 and can only contain these characters: [a-z A-Z 0-9 + / =]</value>
     </labels>
+    <labels>
+        <fullName>geProceedAnyway</fullName>
```
melissabarber, line 8666:
```suggestion
        <fullName>geProcessAnyway</fullName>
```

src/labels/CustomLabels.labels
```diff
@@ -8662,6 +8662,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Proceed Anyway</shortDescription>
```
melissabarber, line 8670:
```suggestion
        <shortDescription>Process Anyway</shortDescription>
```

src/labels/CustomLabels.labels
```diff
@@ -8662,6 +8662,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+        <value>Proceed Anyway</value>
```
melissabarber, line 8671:
```suggestion
        <value>Process Anyway</value>
```

melissabarber (commented):
@ptewson-sfdo added some suggested updates for the button.

src/labels/CustomLabels.labels
ptewson-sfdo: @melissabarber - the mockup from Lauren says "Proceed" and not "Process" - just want to make sure we want to go with "process"?

https://salesforce.invisionapp.com/share/CBZUHFUYFH7#/screens/448085214_Error_State_-_Token_Expired_Revisits

melissabarber: Yep, decided to go with '"Process" after discussing with the docs UI text committee. @laurenchen-sf please update the mockup.

melissabarber: Thanks for checking @ptewson-sfdo!

voduyemi (commented):
This looks good to me! @ptewson-sfdo. I think we need to coordinate on the `handleProcessBatch` function since I'm also adding a call to the server at that point for capturing the credit cards.

src/lwc/geModalPrompt/geModalPrompt.js
```diff
@@ -7,7 +7,19 @@ export default class geModalPrompt extends LightningElement {
     @api variant = '';
     @api title;
     @api message;
-    @api buttonText;
+    @api buttons = [];
+
+    @track enrichedButtons = [];
+
+    connectedCallback() {
```
kenlewis-sfdo, line 14: Would it be possible to pass in the result of "enrichedButtons" through the buttons api property? So, enrich them in a parent component and then just pass them to this component? 

Not saying that this _has_ to change, but could the use of connectedCallback possibly be avoided that way?

kenlewis-sfdo (approved):
One question to ponder, but this looks great to me! Thanks, Patrick!

src/lwc/geModalPrompt/geModalPrompt.js
ptewson-sfdo: Yes, it definitely could! The intent here was to define a default button behavior of "modal close", just to make the geModalPrompt component a bit more light weight in the simplest use case - single "okay" button that just dismisses the modal. This allows you to define a more basic prompt modal without having to pass in your button action, like in like 447 of the form app.

However, I was honestly on the fence about this - if you have a preference to pass in the actions in all cases and remove this connected callback, that would nudge me in the direction of that and it would be a small enough change to make.

kenlewis-sfdo: I see, yeah if its not too big of a lift, I think we should avoid use of the connectedCallback.  Maybe the idea of a simple default button could be added later as a slot or something...

kenlewis-sfdo (approved):
Awesome!

PR #6389, opened 2021-04-29
Bulk Credit Card Capture in BGE

voduyemi:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/ElevateCaptureGroupResponse.cls
```diff
@@ -40,6 +41,10 @@ public with sharing class ElevateCaptureGroupResponse {
         return id;
     }
 
+    public String status() {
+        return status;
+    }
```
kenlewis-sfdo, line 46: This getter is where we should translate the response Status IF, it actually needs to be "stored" in our domain.

If this value means something to Gift Entry Users, what does it mean?  What can they do now that they couldn't do before?

I'm guessing we need _some_ value here to prevent these records from being prematurely processed by BDI, but the value should be meaningful for our domain.

src/classes/ElevateCaptureGroupService.cls
```diff
@@ -48,6 +48,37 @@ public virtual with sharing class ElevateCaptureGroupService {
         return new ElevateAuthorizedGift(tokenizedGift, authorizedGiftResponse);
     }
 
+    public ElevateCaptureGroup sendCreateRequest() {
```
kenlewis-sfdo, line 51:
```suggestion
    public ElevateCaptureGroup getCaptureGroup() {
```

kenlewis-sfdo: I think we should stick closely to our already established language re capturing and capture groups etc

src/classes/ElevateCaptureGroupService.cls
```diff
@@ -48,6 +48,37 @@ public virtual with sharing class ElevateCaptureGroupService {
...
+    public HttpRequest addRequest(ElevateTokenizedGift tokenizedGift, String groupId) {
+        return new PS_Request.Builder()
+                .withMethod(UTIL_Http.Method.POST)
+                .withCaptureGroupId(groupId)
+                .withEndpoint(PS_Request.ElevateEndpoint.ADD_TO_CAPTURE_GROUP)
+                .withBody(JSON.serialize(tokenizedGift))
+                .build();
+    }
+
+    public List<ElevateCaptureGroupResponse> sendChargeRequest(List<String> groupIds) {
```
kenlewis-sfdo, line 69:
```suggestion
      public List<ElevateCaptureGroupResponse> capture(List<String> groupIds) {
```

src/classes/ElevateCaptureGroupService.cls
```diff
@@ -81,20 +103,19 @@ public virtual with sharing class ElevateCaptureGroupService {
                !captureGroupResponse.errors().isEmpty();
     }
 
-    public HttpRequest addRequest(ElevateTokenizedGift tokenizedGift, String groupId) {
+    private HttpRequest createRequest() {
```
kenlewis-sfdo, line 106:
```suggestion
    private HttpRequest captureRequest(String groupId) {
```

src/classes/ElevateCaptureGroupService.cls
```diff
@@ -81,20 +103,19 @@ public virtual with sharing class ElevateCaptureGroupService {
...
-                .withCaptureGroupId(groupId)
-                .withEndpoint(PS_Request.ElevateEndpoint.ADD_TO_CAPTURE_GROUP)
-                .withBody(JSON.serialize(tokenizedGift))
+                .withEndpoint(PS_Request.ElevateEndpoint.CREATE_CAPTURE_GROUP)
+                .withBody(createRequestBody())
                 .build();
     }
 
-    private HttpRequest createRequest() {
+    private HttpRequest chargeRequest(String groupId) {
```
kenlewis-sfdo, line 114:
```suggestion
        private HttpRequest captureRequest(String groupId) {
```

src/classes/GE_GiftEntryController.cls
```diff
@@ -210,6 +210,16 @@ public with sharing class GE_GiftEntryController {
         return captureGroup;
     }
 
+    @AuraEnabled
+    public static void processPaymentsFor(Id batchId) {
+       try {
+           NPSP_Batch batch = new NPSP_Batch(batchId);
+           batch.processPayments();
+       } catch (AuraHandledException ex) {
+           UTIL_AuraEnabledCommon.throwAuraHandledException(ex.getMessage());
```
kenlewis-sfdo, line 219:
```suggestion
           throw new AuraHandledException(ex.getMessage());
```

kenlewis-sfdo: Let's stop using `UTIL_AuraEnabledCommon.throwAuraHandledException`

src/classes/NPSP_Batch.cls
```diff
@@ -0,0 +1,94 @@
...
+* @author Salesforce.org
+* @date 2021
+* @description Batch entity class in NPSP.
+*/
+
+public inherited sharing class NPSP_Batch {
+
+    private Id batchId;
+    private List<DataImport__c> dataImports;
+    private List<String> groupIds = new List<String>();
```
kenlewis-sfdo, line 40:
```suggestion
    private Set<String> captureGroupIds = new Set<String>();
```

src/classes/NPSP_Batch.cls
```diff
@@ -0,0 +1,94 @@
...
+
+
+    private ElevateCaptureGroupService captureGroupService {
```
kenlewis-sfdo, line 43:
```suggestion
     @TestVisible
    private ElevateCaptureGroupService captureGroupService {
```

src/classes/NPSP_Batch.cls
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private Boolean isValidCaptureResponse (ElevateCaptureGroupResponse response, DataImport__c dataImport) {
+        return response.id.equals(dataImport.Payment_Elevate_Batch_ID__c) && response.errors == null;
+    }
+
+    private void setElevateBatchIds() {
+        // Build list of elevate batch Ids here
+        dataImports = [
+                SELECT Payment_Elevate_Batch_ID__c, Id
```
kenlewis-sfdo, line 72:
```suggestion
                             SELECT Payment_Elevate_Batch_ID__c
```

src/classes/NPSP_Batch.cls
```diff
@@ -0,0 +1,94 @@
...
+                FROM DataImport__c
+                WHERE NPSP_Data_Import_Batch__c =: batchId
+                AND Payment_Status__c=: ElevateTransactionStatus.AUTHORIZED.name()
+        ];
+
+        for (DataImport__c dataImport : dataImports) {
+            groupIds.add(dataImport?.Payment_Elevate_Batch_ID__c);
```
kenlewis-sfdo, line 79:
```suggestion
            captureGroupIds.add(dataImport.Payment_Elevate_Batch_ID__c);
```

src/classes/NPSP_Batch.cls
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+        update dataImports;
+    }
+
+    private Boolean isValidCaptureResponse (ElevateCaptureGroupResponse response, DataImport__c dataImport) {
+        return response.id.equals(dataImport.Payment_Elevate_Batch_ID__c) && response.errors == null;
+    }
+
+    private void setElevateBatchIds() {
```
kenlewis-sfdo, line 69: I would suggest:

1. Move the query into a selector
2. Make the selector injectable
3. Establish a "Gifts" domain class
4. Add a filter method to Gifts to return an instance with the Authorized Gifts only
5. Rename the method "captureGroupIds"

src/classes/GE_GiftEntryController.cls
```diff
@@ -210,6 +210,16 @@ public with sharing class GE_GiftEntryController {
...
     }
 
+    @AuraEnabled
+    public static void processPaymentsFor(Id batchId) {
+       try {
+           NPSP_Batch batch = new NPSP_Batch(batchId);
+           batch.processPayments();
+       } catch (AuraHandledException ex) {
+          throw new AuraHandledException(ex.getMessage());
+       }
```
kenlewis-sfdo, line 220: This should catch a regular Exception and then throw a new AuraHandled

} catch (Exception ex) {
       }

src/classes/GE_GiftEntryController_TEST.cls
```diff
@@ -78,6 +78,31 @@ private with sharing class GE_GiftEntryController_TEST {
...
+            batch.processPayments();
+        } catch (AuraHandledException ex) {
+            auraException = ex;
+        }
+        Test.stopTest();
+        List<DataImport__c> updatedDataImports = [
+                SELECT Payment_Status__c FROM DataImport__c WHERE Id =: dataImports[0].Id
+        ];
+        System.assertEquals(ElevateTransactionStatus.PAYMENTPROCESSING.name(),updatedDataImports[0].Payment_Status__c);
+    }
```
kenlewis-sfdo, line 104: I would put this test in an NPSP_Batch_TEST file, since it is really testing the `batch.processPayments();` method.

src/classes/UTIL_Http_TEST.cls
```diff
@@ -307,4 +307,13 @@ public with sharing class UTIL_Http_TEST {
...
     }
 
+    public static void mockChargeCaptureGroup(String responseBody){
+        CalloutMock callout = new CalloutMock()
+                .withStatusCode(UTIL_Http.STATUS_CODE_OK)
+                .withStatus(UTIL_Http.STATUS_OK)
+                .withBody(responseBody);
+
+        Test.setMock(HttpCalloutMock.class, callout);
+    }
```
kenlewis-sfdo, line 317: I would move this mock into the NPSP_Batch_TEST file also, since that is the only place it is used currently and it will make that test easier to read and have one less dependency for that class (higher cohesion, looser coupling).

kenlewis-sfdo (approved):
A couple comments about moving things around, but overall this is looking really great, @voduyemi !  Thanks for this work and thank you so much for working through the feedback and getting this PR into really good form!

ptewson-sfdo (approved):
Not much to add on to what's already been said, nice work here Victor!

src/classes/ElevateCaptureGroupService.cls
```diff
@@ -48,6 +48,37 @@ public virtual with sharing class ElevateCaptureGroupService {
...
+                .withCaptureGroupId(groupId)
+                .withEndpoint(PS_Request.ElevateEndpoint.ADD_TO_CAPTURE_GROUP)
+                .withBody(JSON.serialize(tokenizedGift))
+                .build();
+    }
+
+    public List<ElevateCaptureGroupResponse> capture(Set<String> captureGroupIds) {
+        List<ElevateCaptureGroupResponse> elevateCaptureGroupResponses = new List<ElevateCaptureGroupResponse>();
+        for (String elevateBatchId : captureGroupIds) {
+            UTIL_Http.Response response = new UTIL_Http.RequestService().sendRequest(
```
sokosaurus, line 72: This is a probably a veryyyyy minor edge case but I figured I'd raise it just in case. Given that we could be making multiple callouts to the elevate "capture" endpoint when a BGE batch contains multiple capture groups, is there any concern that we might hit governor limits?

sokosaurus (approved):

sokosaurus:
Really great work here!

PR #6390, opened 2021-04-30
CRLP Account Soft Credit fix

npsp-reedestockton:
We corrected a bug that caused Account Soft Credit batch jobs to run twice - once in NonSkew mode and once in Skew mode.
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupBatch_SVC.cls
+++ b/src/classes/CRLP_RollupBatch_SVC.cls
@@ -91,7 +91,7 @@ public class CRLP_RollupBatch_SVC {
                 queryFilter = '(' + softCreditField + ' >= ' + maxRelatedOppsForNonLDVMode + ')';
             }
 
-        } else if (CRLP_Rollup_SVC.isOppContactRoleSoftCreditRollup(jobType) ||
+        } else if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
             jobType == CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
 
             // TODO -- How to get Accounts that require using this method for Soft Credits??
--- a/src/classes/CRLP_RollupBatch_SVC.cls
+++ b/src/classes/CRLP_RollupBatch_SVC.cls
@@ -91,7 +91,9 @@ public class CRLP_RollupBatch_SVC {
                 queryFilter = '(' + softCreditField + ' >= ' + maxRelatedOppsForNonLDVMode + ')';
             }
 
-        } else if (CRLP_Rollup_SVC.isOppContactRoleSoftCreditRollup(jobType)) {
+        } else if (CRLP_Rollup_SVC.isOppContactRoleSoftCreditRollup(jobType) ||
+            jobType == CRLP_RollupProcessingOptions.RollupType.AccountSoftCredit) {
+
             // TODO -- How to get Accounts that require using this method for Soft Credits??
             // For now just use the NumberOfClosedOpportunities field as a proxy for the number of SoftCredits
             String softCreditField = parentRelationshipField + 'npo02__NumberOfClosedOpps__c';
--- a/src/classes/CRLP_RollupBatch_SVC.cls
+++ b/src/classes/CRLP_RollupBatch_SVC.cls
@@ -127,10 +127,9 @@ public class CRLP_RollupBatch_SVC {
      * @return True or False
      */
     public static Boolean doesRecordNeedSkewRollupMethod(Id recordId, CRLP_RollupProcessingOptions.RollupType jobType) {
-        if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
+        if (jobType != CRLP_RollupProcessingOptions.RollupType.GAU &&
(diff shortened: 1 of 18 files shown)


src/classes/CRLP_RollupBatch_SVC.cls
```diff
@@ -125,10 +127,9 @@ public class CRLP_RollupBatch_SVC {
      * @return True or False
      */
     public static Boolean doesRecordNeedSkewRollupMethod(Id recordId, CRLP_RollupProcessingOptions.RollupType jobType) {
-        if (jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.AccountContactSoftCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
+        if (jobType != CRLP_RollupProcessingOptions.RollupType.GAU &&
```
lparrott, line 130: Is there any fear of a RollupType being added that we wouldn't want included here?

npsp-reedestockton: @lparrott Rollup types are not only fairly stable, but also require substantial infrastructure to implement.  Any newly implemented rollup type will need integration in multiple locations and this is one of them.  I would expect anyone working on a new rollup type to understand the implications of how it should be implemented.  Flagging how records will be selected for inclusion is fundamental.

npsp-reedestockton:
@force2b This PR is in draft status.  I am adding a new test class to validate our query results, but the logic for the work item should be complete.  It's a small PR, so if you have a few minutes to take a quick look, I would appreciate it.

src/classes/CRLP_RollupBatch_SVC.cls
npsp-reedestockton: @lparrott To your point, though, I intentionally added AccountSoftCredit rollup types to this selection.  I saw no good reason for their exclusion, but would like my thinking confirmed by @force2b.

force2b: So, this went from explicit list to just not RD or GAU, meaning that the code in the if block runs for all other rollup types? That makes sense since the new `CustomizableRollups_UseSkewMode__c` field only exists on the Account and Contact objects - and all the other rollup types are for those two objects.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -523,6 +522,11 @@ public inherited sharing class CRLP_RollupProcessor {
         if (mode == CRLP_RollupProcessingOptions.BatchJobMode.SkewMode && this.useForceSkewModeField && !isForceSkewModeTrue) {
             updatedRecord.put(this.qualifiedSkewField, true);
             needsUpdate = true;
+
+        // Reset any flags erroneously set for NonSkewMode records
+        } else if (mode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode && this.useForceSkewModeField && isForceSkewModeTrue) {
```
force2b, line 527: Will Account record with the ForceSkewMode flag set to true still be included in the NonSkew batch job for Account Hard Credit? If so, then this logic will run and reset the flag. However, if that logic is using the flag to build the queryLocator on the batch job then those will continue to run in the SkewMode job and not be picked up here.

npsp-reedestockton: Account records with the ForceSkewMode flag will not be picked up to run in skew mode by the nightly run on the basis of the ForceSkewMode flag.  If they qualify on the basis of threshold settings, they will run in skew mode and don't need the flag reset.  Account records which should qualify for nonskew batch runs on the basis of the threshold settings will be picked up to run in nonskew mode by the nightly jobs and the flag will be reset.  The ForceSkewMode flag will become the basis for selection in our nightly jobs in the next work item.  Sound right?

force2b: Yep. Sounds perfect. Thank you for confirming.

npsp-reedestockton:
@lparrott @force2b Moving out of draft status. Would appreciate another look.

@mattwashington Would you take a look at the PR description?

Thanks all!!!

npsp-reedestockton:
Tracking [W-9162816](https://gus.lightning.force.com/a07AH000000BMZwYAO) and [W-9061202](https://gus.lightning.force.com/a07AH0000009FtHYAU)

src/classes/CRLP_RollupBatch_TEST.cls
```diff
@@ -0,0 +1,177 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Customizable Rollups Operations Services
+* @description Unit Tests to validate skew vs nonskew mode query selection thresholds
+*/
+@IsTest(isParallel=true)
+private class CRLP_RollupBatch_TEST {
```
lparrott, line 37: This is a creative way to test these changes, thanks for adding this new test class 👍

lparrott (approved):

src/classes/CRLP_RollupBatch_TEST.cls
npsp-reedestockton: @kenlewis-sfdo persuaded me that this was the proper approach to these tests.  

Given the limited scope of the testing gap, I'm really happy with the way this turned out.  Thanks, @kenlewis-sfdo!

force2b (approved):
Very nice work!

mattwashington:
Looks good @npsp-reedestockton. Nice work!

PR #6392, opened 2021-05-02
Project Hummingbird: Integration Tests to Unit Tests

kenlewis-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000ACCWYA4/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/ContactBuilder.cls
```diff
@@ -0,0 +1,74 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Used to create test Contacts in NPSP.
+*/
+@IsTest
+public class ContactBuilder {
```
voduyemi, line 36: Do we think the name of this class should have the word 'Test' in it somewhere since it's only for test classes? `ContactBuilder` at first glance eludes to a factory class for building valid contact records across NPSP.

src/classes/HouseholdNamingService.cls
```diff
@@ -39,6 +39,28 @@ public without sharing class HouseholdNamingService {
...
+            if (selector == null) {
+                selector = new ContactSelector();
+            }
+            return selector;
+        }
+        set;
+    }
+
+    @TestVisible
+    private UnitOfWork unitOfWorkInst {
```
voduyemi, line 54:
```suggestion
    private UnitOfWork unitOfWorkInstance {
```

src/classes/HouseholdService.cls
```diff
@@ -54,8 +55,20 @@ public inherited sharing class HouseholdService {
         }
         set;
     }
+
     private static final String RECORD_TYPE_ID = 'RecordTypeId';
 
+    @TestVisible
+    private UnitOfWork unitOfWorkInst {
```
voduyemi, line 62:
```suggestion
    private UnitOfWork unitOfWorkInstance {
```

src/classes/UnitOfWork.cls
```diff
@@ -0,0 +1,78 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Unit of Work in NPSP.
+*/
+public inherited sharing class UnitOfWork {
```
voduyemi, line 35: Do we think we need a test class for this class?

src/classes/fflib_IDGenerator.cls
```diff
@@ -0,0 +1,43 @@
...
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+public with sharing class fflib_IDGenerator
```
voduyemi, line 26: Do we need a test class for this class?

voduyemi (commented):
Thanks @kenlewis-sfdo for all this awesome work. I have some comments around refactoring, renaming and adding test classes.

src/classes/UnitOfWork.cls
kenlewis-sfdo, line 35: Yes, Victor, good call!

As demonstrated by this work, NPSP needs a [Unit of Work](https://trailhead.salesforce.com/en/content/learn/modules/apex_patterns_sl/apex_patterns_sl_learn_uow_principles) so that Domains can be unit-tested.   I'd prefer to bring in the [Unit of Work from fflib](https://github.com/wimvelzeboer/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls) along with its [full test suite](https://github.com/wimvelzeboer/fflib-apex-common/blob/master/sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls) rather than building this one out further.  Curious to hear the teams thoughts.

src/classes/fflib_IDGenerator.cls
kenlewis-sfdo: Added!

src/classes/UnitOfWork.cls
bdvorachek: I'd like to see the fflib version of UoW get pulled in too, already built and has test coverage even for the stuff we wouldn't be using quite yet. 👍

src/classes/HouseholdNamingService.cls
```diff
@@ -98,9 +131,24 @@ public without sharing class HouseholdNamingService {
...
+            List<Contact> members = membersByHouseholdId.get(idFor(household));
+            if (numberOfHouseholdMembersIsIncorrect(household, members)) {
+                refreshNumberOfHouseholdMembers(household, members);
+                unitOfWorkInst.registerDirty(new List<SObject>{household});
+            }
+        }
+    }
+
+    private void refreshNumberOfHouseholdMembers(SObject household, List<Contact> members) {
+        household.put(UTIL_Namespace.StrTokenNSPrefix('Number_of_Household_Members__c'),
```
ptewson-sfdo, line 143: (Non Blocking!) Not sure I love this use of inline string for field name, but I can see how `SObjectType.<SObjectName>.fields.Number_Of_Househodl_Members__c.Name` is difficult here

src/classes/NPSP_Address_TEST.cls
```diff
@@ -0,0 +1,10 @@
+@IsTest
+private class NPSP_Address_TEST {
+    //Todo: test copyFromSObject
+    @IsTest
+    static void testBehavior() {
```
ptewson-sfdo, line 5: Is this test necessary? Accomplishing anything?

src/classes/NPSP_Contact.cls
```diff
@@ -117,16 +118,74 @@ public inherited sharing class NPSP_Contact {
         return contact.AccountId;
     }
 
+    public Id householdId() {
```
ptewson-sfdo, line 121: Should this be "legacyHousholdId" or anything to differentiate it from a household account id? or is that nomenclature not used everywhere?

src/classes/UnitOfWork.cls
ptewson-sfdo: (Blocking!!!!) I think the expectation is clear that we need test coverage for anything we are shipping to customers, even if temporary. If we aren't ready to bring in the fflib UoW infrastructure before me merge this to the feature branch, I think we need a test class here (even if relatively basic).

ptewson-sfdo (commented):
Thanks for all this great work Ken - only a couple of comments I would consider blocking, the rest is more for possible consideration as this work evolves.

src/classes/ContactBuilder.cls
kenlewis-sfdo: As I mentioned in our design huddle meeting yesterday, I don't think there's really any needed in production code to"build" a contact. Usually you are getting contacts from triggers or controllers (and sometimes from something like a form). The class is marked @isTest, and "builder" is a fairly common naming convention for this pattern so I think sticking with ContactBuilder would be good.

src/classes/HouseholdNamingService.cls
kenlewis-sfdo: Yeah, I don't love this either, but hesitant to change it at the moment.  I think we need a follow up WI to define the interface IHouseholds that NPSP_Households and NPSP_HouseholdAccounts will implement.  Then we can define behavior like this on the interface and have each implementation contain its own "numberOfHouseholdMembers(Integer count)" method implementation, then removing this usage of UTIL_Namespace and the string field reference.

src/classes/UnitOfWork.cls
kenlewis-sfdo: This class/code is definitely covered, it just doesn't have its own explicit test class.  My opinion is that we shouldn't spend time writing new dedicated tests for this utility.  It is covered by implicit tests.  And if/when we are able to use fflib UoW, then NPSP will have a full, robust UoW with its own dedicated tests.

src/classes/NPSP_Contact.cls
kenlewis-sfdo: I think this should stay as is because NPSP does need an IHousehold/s interface.  Then NPSP_Household and NPSP_HouseholdAccount will both have a "householdId()" method, each of which will return its respective (Account or Household__c) Id.

src/classes/AddressService.cls
```diff
@@ -114,16 +116,70 @@ public inherited sharing class AddressService {
...
+    * @param strFieldAddrType an optional Address Type field on sobjDst to copy to
+    * @return void
     */
-    public String getMultilineStreet(Address__c addr) {
-        Boolean isOneLineStreetAddress = String.isBlank(addr.MailingStreet2__c);
-        if (isOneLineStreetAddress) {
-            return addr.MailingStreet__c;
+    public void copyOntoSObject(IAddress anAddress, SObject sobjDst,
+            String strFieldPrefix, String strFieldAddrType) {
+        Address__c addr = (Address__c) anAddress.getRecord();
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yes, @ptewson-sfdo spot on!  This should definitely be a follow up item also.  This method body should really be an implementation detail on each of the two entities Contact and Household. When you give them each an Address, they know how to set their fields internally, instead of here in this AddressService class.  But I am going to reserve this one for a follow up unless I hear otherwise!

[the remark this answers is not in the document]
kenlewis-sfdo: And yes @voduyemi, as it is this method would be better named something like buildContactAndAccountAddresses and have a guard that only accepts the two supported SObjects.  But, ultimately it should be broken up and relocated inside the entities themselves, and then it won't be a method that supports multiple SObjects.

src/classes/AddressService.cls
```diff
@@ -74,25 +74,27 @@ public inherited sharing class AddressService {
     * @param addr Address
     * @return boolean. true if any of the Address fields on the Contact are different from this Address record
     */
-    public Boolean isSObjectAddressDifferent(SObject sObj, Address__c addr) {
+    //Todo: includes lat and long, wBilling, compares HHAcct not HH?
+    public Boolean isSObjectAddressDifferent(SObject sObj, IAddress other) {
+        Address__c addr = (Address__c) other.getRecord();
         if (sObj == null || addr == null) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yeah, this is another method that should be further refactored/relocated.  A follow up WI here should refactor so that Addresses are compared using the entities' query methods, i.e., `contact.mailingAddress.isEqualTo(household.address())`
This line should be refactored along with that work.

src/classes/NPSP_Address.cls
```diff
@@ -476,16 +449,75 @@ public inherited sharing class NPSP_Address {
...
+    }
+
+    /*******************************************************************************************************
+    * @description utility to compare two address objects address fields
+    * @param addrNew the new version of the Address
+    * @param addrOld the old version of the Address
+    * @param includeAddressType whether to include comparing the AddressType field
+    * @return boolean. true if any of the fields fields have changed on the Address
+    */
+    public Boolean isAddressChanged(Address__c addrOld, Boolean includeAddressType) {
```
bdvorachek, line 508: Boolean in a function argument here that we may be able to refactor. There doesn't appear to be anywhere that passes `true` to that argument - would that be code we could remove possibly?

src/classes/ContactBuilder.cls
bdvorachek: The names for test classes will matter much less once we're on SFDX project structure too, I imagine we want to pull those into their own tests folder separate from prod code.

src/classes/NPSP_Address.cls
kenlewis-sfdo: Yep, great call, Ben!  It looks like we might be able to adjust the one call site for this method and have it use the isEqualTo method instead, which is a dupe of this method.  I'll take a look!

src/classes/HH_Households_TEST.cls
```diff
@@ -635,17 +717,12 @@ public with sharing class HH_Households_TEST {
...
-        String street = '2345 Elm St.';
-        String city = 'Seattle';
-        String state = 'Washington';
-        String zip = '88777';
-        String country = 'United States';
         String lastname = 'SmithForTest99';
 
         List<Contact> cons = new List<contact>();
-        Integer conCount = 200;
+        Integer conCount = 1;
```
lparrott, line 725: Can this still be considered `testBulk` if there's only 1 Contact 😉 I also worry that by reducing the number of Contacts, we're changing what code gets covered by the test method (but I haven't confirmed that...)

src/classes/HouseholdNamingService.cls
```diff
@@ -39,6 +39,39 @@ public without sharing class HouseholdNamingService {
 
     private HouseholdSettings settings = new HouseholdSettings();
 
+    @TestVisible
```
bdvorachek, line 42: Some future planning here, non-blocking.

Can we eventually have dependency injection without using `@TestVisible`? These dependencies could be passed into the class via a constructor, but some other class would need to manage that - and I think that'd be an [Application class](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_Application.cls). When `HouseholdNamingService` would need an instance of `ContactSelector`, it'd grab whatever one is given to it by the Application Class. Does that seem viable in the long-run?

src/classes/HH_Households_TEST.cls
kenlewis-sfdo: Yep, definitely a leftover from testing/dev.  Thanks for catching!  Didn't mean to make that change permanent.  Fixed!

src/classes/HouseholdNamingService.cls
kenlewis-sfdo: Yep, exactly! Once we have a framework like [Force-DI](https://github.com/apex-enterprise-patterns/force-di) in place, then the injection of these class member collaborators become one-liners `(forceDI.getInstance(Collaborator.class))`, and the Application is configured for run-time injection using custom metadata types.  At that point, the domain can be configured at run time in a test by telling the Application to build an instance using mock collaborators instead of real ones, etc.

src/classes/IAddress.cls
```diff
@@ -0,0 +1,58 @@
...
+* @author Salesforce.org
+* @date 2021
+* @description Defines common functionality between addresses used in NPSP.
+*/
+public interface IAddress {
+
+    //Todo: refactor until getRecord() is no longer necessary, then remove
+    SObject getRecord();
+    //Todo: refactor until getType() can be removed.  An Entity that knows one or more
+    // addresses should be storint
```
bdvorachek, line 40: Typo here

src/classes/UnitOfWork.cls
ptewson-sfdo: Sold! Thanks for laying that out for me

ptewson-sfdo (approved):
Thanks for responding to feedback so quickly! This looks good!

src/classes/IAddress.cls
kenlewis-sfdo: 👍🏽

src/classes/NPSP_Address.cls
kenlewis-sfdo: Was able to remove this method entirely.  Thanks, Ben!

src/classes/HouseholdService_TEST.cls
```diff
@@ -0,0 +1,113 @@
...
+        service.createHouseholdsFor(
+                new ContactsInLegacyHouseholds(
+                        ContactBuilder.dummyContactList()));
+
+        //Then
+        System.assert(mock.householdsToBeInserted.size() == 1);
+        System.assert(mock.saveCalled);
+    }
+
+    public class UnitOfWorkMock implements StubProvider {
```
force2b, line 75: Does this need to be public?

src/classes/HouseholdService_TEST.cls
```diff
@@ -0,0 +1,113 @@
...
+/**
+* @author Salesforce.org
+* @date 2021
+* @description HouseholdService tests in NPSP.
+*/
+@IsTest
+private class HouseholdService_TEST {
+
+    @IsTest
+    static void insertHouseholdsForContacts_insertsAHouseholdForDummyContact_whenHHSettingsAll
```
force2b, line 39: Same with test method naming and a separator for helper methods.

src/classes/IAddress.cls
```diff
@@ -0,0 +1,57 @@
...
+    String getType();
+
+    String street();
+    String street2();
+    String multilineStreet();
+
+    String city();
+    String state();
+    String postalCode();
+    String country();
```
force2b, line 49: Since this is a common interface, I think it should also include CountryCode and StateCode even though those 2 only exist when enabled in the org. Though, I suppose that this might not work as well since the `Address__c` object doesn't implement these 2 fields.

src/classes/IAddress.cls
```diff
@@ -0,0 +1,57 @@
...
+*/
+public interface IAddress {
+
+    //Todo: refactor until getRecord() is no longer necessary, then remove
+    SObject getRecord();
+
+    String getType();
+
+    String street();
+    String street2();
```
force2b, line 43: Could there be more than 2 lines? Maybe this should be `List<String> streetLines`?

src/classes/NPSP_Address.cls
```diff
@@ -313,17 +276,17 @@ public inherited sharing class NPSP_Address {
...
+    public void copyFrom(IAddress other) {
+        addr.MailingStreet__c = other.street();
+        addr.MailingStreet2__c = other.street2();
+        addr.MailingCity__c = other.city();
+        addr.MailingState__c = other.state();
+        addr.MailingPostalCode__c = other.postalCode();
+        addr.MailingCountry__c = other.country();
+        addr.Geolocation__Latitude__s = other.latitude();
+        addr.Geolocation__Longitude__s = other.longitude();
+        if (other.getType() != null)
```
force2b, line 288: curly brackets needed here since new code

src/classes/NPSP_Contact.cls
force2b: @kenlewis-sfdo To @ptewson-sfdo's point though, the method name is `householdId()`. If Account represents a household in the HouseholdAccount Model, then this method should return the AccountId, but it doesn't. I think the method name needs to be different, and the method (and some others here) probably need comments documenting behavior so it's clear how to use these and when.

src/classes/NPSP_Contact_TEST.cls
```diff
@@ -0,0 +1,58 @@
...
+/**
+* @author Salesforce.org
+* @date 2021
+* @description NPSP_Contact tests.
+*/
+@IsTest
+private class NPSP_Contact_TEST {
+
+    @IsTest
+    static void givenAContactWithAddressInfo_thenHasMailingAddress() {
```
force2b, line 39: Use should naming syntax here too.

src/classes/OrgConfig.cls
```diff
@@ -46,7 +46,7 @@ public inherited sharing class OrgConfig {
     /*******************************************************************************************************
     * @description map from State name (UPPER CASE) to State Code
     */
-    private Map<String, String> validStatesByLabel = new Map<String, String>();
+    private static Map<String, String> validStatesByLabel = new Map<String, String>();
```
force2b, line 49: I'm curious why the switch to static, and why not all methods use static.

src/classes/fflib_IDGenerator.cls
```diff
@@ -0,0 +1,43 @@
...
+{
+    private static Integer fakeIdCount = 0;
+    private static final String ID_PATTERN = '000000000000';
+
+    /**
+     * Generate a fake Salesforce Id for the given SObjectType
+     */
+    public static Id generate(Schema.SObjectType sobjectType)
```
force2b, line 34: This is a duplicate of what already exists in [UTIL_UnitTestData_TEST.mockIds()](https://github.com/SalesforceFoundation/NPSP/blob/1b4c3af5674ac0a5df375b3c12f65d5bb490bd58/src/classes/UTIL_UnitTestData_TEST.cls#L1003). I don't think we should implement a new class that duplicates the behavior unless all existing uses of mockIds() are also updated. It might be better for your new tests to just use the existing method.

kenlewis-sfdo: Yep! Thanks, will update!

src/classes/OrgConfig.cls
kenlewis-sfdo: Instance methods for mocking, static fields since not all instances of this class need their own version of this map.

src/classes/AddressService.cls
```diff
@@ -74,29 +74,29 @@ public inherited sharing class AddressService {
...
-                        (isContact &&
-                                (!equalsCaseSensitive((String) sObj.get('npe01__Primary_Address_Type__c'), addr.Address_Type__c) ||
-                                        !equalsCaseSensitive((String) sObj.get('Current_Address__c'), addr.Id))
-                        )
-        );
+        Boolean isDifferent = isDifferentIncludingLatLong(sObj, prefix, other);
         return isDifferent;
     }
 
+    private Boolean isDifferentIncludingLatLong(SObject sObj, String prefix, IAddress other) {
```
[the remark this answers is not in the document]
kenlewis-sfdo: The implementation in the service method is not yet using the I address interface it's pulling the field directly from the S objects. This should definitely be fixed but I've tracked it here at the follow-up work item: https://salesforce.quip.com/6KXhAwNldybY

src/classes/ContactSelector.cls
```diff
@@ -0,0 +1,90 @@
...
+
+        return Database.Query(householdMembersQueryFor(queryHHIds));
+    }
+
+    private String householdMembersQueryFor(Set<Id> queryHHIds) {
+        //get all of the contacts and their associated households
+        //return them in the order of household and then household naming order
+        //we use dynamic soql so we can include all contact fields, since custom naming may refer to any field.
+        String strSoql = getQueryForContactNamingFields();
+        String strHHId = UTIL_Namespace.StrTokenNSPrefix('HHId__c');
```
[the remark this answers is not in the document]
kenlewis-sfdo: Yep, seems odd to me too. Added for follow up! https://salesforce.quip.com/6KXhAwNldybY

src/classes/Contacts.cls
```diff
@@ -496,16 +491,31 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+            if (isAnAccountOfTypeHousehold(accountFor(npspContact))) {
+                contactsInHouseholdAccounts.add(npspContact);
             }
-            contacts.add(con);
         }
-        return new Contacts(contacts);
+        return contactsInHouseholdAccounts;
+    }
+
+    private List<NPSP_Contact> thatHaveAccountsAsNpspContacts;
```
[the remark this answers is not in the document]
kenlewis-sfdo: Renamed! Hopefully better now.

src/classes/ContactsInLegacyHouseholds_TEST.cls
```diff
@@ -0,0 +1,135 @@
...
+                    updateHouseholdNamesCalled = true;
+                    contactsQueuedForHouseholdNameUpdate =
+                            (ContactsInLegacyHouseholds) listOfArgs[0];
+                }
+            }
+            return null;
+        }
+    }
+
+    @IsTest
```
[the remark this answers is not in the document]
kenlewis-sfdo: Incomplete test method idea.  Removed!

src/classes/ContactsInLegacyHouseholds_TEST.cls
```diff
@@ -0,0 +1,135 @@
...
+/**
+* @author Salesforce.org
+* @date 2021
+* @description ContactsInLegacyHouseholds tests in NPSP.
+*/
+@IsTest
+private class ContactsInLegacyHouseholds_TEST {
+
+    @IsTest
+    static void whenOnAfterInsert_thenCreateHouseholdsCalled() {
```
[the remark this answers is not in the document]
kenlewis-sfdo: I dont disagree, but lets standardize and document before changing all of these test method names in the interest of efficiency.  Follow up: https://salesforce.quip.com/6KXhAwNldybY

src/classes/ContactsInLegacyHouseholds_TEST.cls
```diff
@@ -0,0 +1,135 @@
...
+        contactsInLegacyHouseholds.onAfterUpdate();
+
+        //Then
+        System.assert(householdServiceMock.updateHouseholdNamesCalled);
+        Contact contactInQueue =
+                (Contact) householdServiceMock.contactsQueuedForHouseholdNameUpdate.getRecords()[0];
+        System.assert(isDummyContact(contactInQueue), 'The dummy Contact should be queued for ' +
+                'Household renaming since its Title changed.');
+    }
+
```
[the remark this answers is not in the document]
kenlewis-sfdo: As a team we've shied away from these as hard to maintain.  Added as follow up to finalize: https://salesforce.quip.com/6KXhAwNldybY

src/classes/NPSP_Address.cls
kenlewis-sfdo: Postponing as follow up

src/classes/NPSP_Contact_TEST.cls
kenlewis-sfdo: Postponing as follow up

src/classes/fflib_IDGenerator.cls
kenlewis-sfdo: done!

src/classes/NPSP_Contact.cls
kenlewis-sfdo: I think this method on this class should keep this name for now.  It should go away entirely eventually.  https://salesforce.enterprise.slack.com/files/WUC2LP0KH/F022PRUFZR7/ihousehold.png

src/classes/HouseholdService_TEST.cls
kenlewis-sfdo: updated!

src/classes/IAddress.cls
kenlewis-sfdo: Added as follow up! https://salesforce.quip.com/6KXhAwNldybY#RFCACAFsFro

src/classes/ContactsInLegacyHouseholds_TEST.cls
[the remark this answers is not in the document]
force2b: Interesting. That was pretty standard for all NGOhana work. We can reconcile later.

force2b (approved):

src/classes/NPSP_Address.cls
jesslopez-sf: @kenlewis-sfdo - would be curious to hear if the refactored work would eliminate this Bug we recently created that does use isAddressChanged() with a value of true during one scenario:
#6416

PR #6404, opened 2021-05-12
Gift Entry: add ability to delete authorized gifts from gift entry batches

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# Features Intended for Future Release

# Features for Elevate Customers

# New Metadata

# Deleted Metadata

PR #6396, opened 2021-05-04
Gift Entry: update elevate widget display

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# Features Intended for Future Release

# Features for Elevate Customers

# New Metadata

# Deleted Metadata

PR #6411, opened 2021-05-14
Validations when editing an authorized gift in BGE

sokosaurus:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -731,11 +739,24 @@ export default class GeFormRenderer extends LightningElement{
                 this.handleElevateAPIErrors(errors);
                 return;
             }
+        }  else if (this.isGiftAuthorized()) {
+            try {
+                await validateAuthorizedGiftEdit({'dataImport': dataImportFromFormState});
```
voduyemi, line 744: Didn't know we could make imperative apex calls with params in `''`. Interesting!

voduyemi (commented):
Thanks @sokosaurus. I left some comments are refactoring. Looking good so far. Let me know what you think!

src/classes/GiftBatchService.cls
```diff
@@ -39,4 +39,29 @@ public with sharing class GiftBatchService {
         GiftBatchGroupedTotals groupedTotals = batches.getGroupedTotals();
         return groupedTotals.totalsFor(batchId);
     }
+
+    public Boolean comparePaymentFieldsForChange(DataImport__c updatedDataImport, GiftSelector selector) {
+        List<String> paymentFields = new List<String>{
```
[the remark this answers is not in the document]
sokosaurus: Thanks for calling this out! In a namespaced org, will the value returned from `String.valueOf(Object.field)` have the namespace prefixed to the API name of the field?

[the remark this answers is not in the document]
voduyemi: Yup I believe so.

src/labels/CustomLabels.labels
```diff
@@ -6871,6 +6871,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Common error when a total doesn't match Donation Amount.</shortDescription>
         <value>Total amount doesn't match the {0}</value>
     </labels>
+    <labels>
+        <fullName>geErrorBatchGiftEntryReauthRequired</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when payment attributes changed on authorized gift</shortDescription>
+        <value>To update donor, date or amount, reenter card details to reauthorized the payment.</value>
```
sokosaurus, line 6880: @melissabarber - FYI, here's a new label that I'll be using for an error message here.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -731,11 +739,24 @@ export default class GeFormRenderer extends LightningElement{
                 this.handleElevateAPIErrors(errors);
                 return;
             }
+        }  else if (this.isGiftAuthorized()) {
```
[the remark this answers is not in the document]
sokosaurus: Done! Please let me know if you think any additional refactoring is necessary here.

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -140,6 +140,11 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
         return this._displayState === 'criticalError';
     }
 
+    get shouldDisplayEditPaymentInformation() {
+        return this.isReadOnly  
+            && (this.paymentStatus() === 'EXPIRED' || this.paymentStatus() === 'AUTHORIZED');
```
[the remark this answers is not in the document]
sokosaurus: Good call. Done!

src/lwc/geFormRenderer/geFormRenderer.js
sokosaurus: Thanks for drawing attention here. I went ahead and removed the `''` to align with how we're passing parameters into other apex methods within BGE.

src/classes/GiftBatchService.cls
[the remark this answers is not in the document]
sokosaurus: Ok, thanks. I'm now using `String.valueOf(Object.field)` and have moved the List of fields to a static constant in this class. This list will also be referenced by `getGiftById()` on `GiftSelector.cls`.

I'm not sure if this is the right approach or place for it, so please let me know if you have any additionl thoughts.

src/labels/CustomLabels.labels
```diff
@@ -6879,6 +6879,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Common error when a total doesn't match Donation Amount.</shortDescription>
         <value>Total amount doesn't match the {0}</value>
     </labels>
+    <labels>
+        <fullName>geErrorBatchGiftEntryReauthRequired</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when payment attributes changed on authorized gift</shortDescription>
```
melissabarber, line 6887:
```suggestion
        <shortDescription>Instructions to reenter payment details when updating authorized details</shortDescription>
```

src/labels/CustomLabels.labels
melissabarber: Thanks! added one suggestion for the shortDescription other than that, all good to go.

src/classes/GE_PaymentServices.cls
```diff
@@ -100,8 +100,8 @@ public with sharing class GE_PaymentServices {
      * @description Returns Elevate Product Id to payments services tokenize card iframe
      * @return String
      */
-    public String getProductId() {
-        return configService.getProductId();
+    public String getClientId() {
```
ptewson-sfdo, line 103: @sokosaurus This update was originally part of Victor's work item and I *believe* should also be reflected in the parent branch of this PR. I think this is maybe only showing up as a diff here as a weird glitch in github - 

https://github.com/SalesforceFoundation/NPSP/blob/feature/232/src/classes/GE_PaymentServices.cls#L103

I'm pointing that out only because it's throwing me to see this in the diff, but I'm going to ignore it as something goofy github is doing unless you have any further insight!

src/classes/Gift.cls
```diff
@@ -0,0 +1,54 @@
...
+    
+    @TestVisible
+    private DataImport__c dataImport;
+    
+    public Gift(DataImport__c dataImport) {
+        this.dataImport = dataImport;
+    }
+
+    public Boolean isGiftReauthorizedRequired() {
+        if (dataImport.Payment_Status__c != ElevateTransactionStatus.AUTHORIZED.name()) { return false; }
```
ptewson-sfdo, line 46:
```suggestion
        if (dataImport.Payment_Status__c != ElevateTransactionStatus.AUTHORIZED.name()) {
            return false; 
        }
```

src/classes/Gift.cls
```diff
@@ -0,0 +1,54 @@
...
+        return 
+            new GiftBatchService().comparePaymentFieldsForChange(
```
ptewson-sfdo, line 48: Should we have a GiftBatchService.getInstance() method?

src/classes/Gift.cls
```diff
@@ -0,0 +1,54 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Gift entity class in NPSP.
+*/
+
+public with sharing class Gift {
```
ptewson-sfdo, line 36: (NON-BLOCKING) From a Domain perspective, is "Gift" different than "Donation"? It feels a little like the name Gift here feels like a cheat to constrain this concept to Gift Entry as opposed to incorporating the concept into the broader NPSP domain - but there definitely might be good reason behind it. CC @kenlewis-sfdo

src/classes/GiftBatchService.cls
```diff
@@ -34,9 +34,36 @@
 */
 public with sharing class GiftBatchService {
 
+    public static final List<String> PAYMENT_FIELDS = new List<String>{
```
ptewson-sfdo, line 37: (NON-BLOCKING) I'm struggling with the name "PAYMENT FIELDS" but not coming up with a better option - but there's maybe a distinction between fields that map to payment but aren't sent to elevate, which would fall under the category of "payment fields" but wouldn't belong in this actual list. "PAYMENT_INTEGRATION_FIELDS"? 

This is non-blocking but if we can think of a better way to represent the concept here that would be tops.

src/classes/GiftBatchService.cls
```diff
@@ -34,9 +34,36 @@
...
         GiftBatchGroupedTotals groupedTotals = batches.getGroupedTotals();
         return groupedTotals.totalsFor(batchId);
     }
+
+    public Boolean comparePaymentFieldsForChange(DataImport__c updatedDataImport, GiftSelector selector) {
+        DataImport__c databaseDataImport = selector.getGiftById(updatedDataImport.Id);
+        Map<String, Object> updatedDataImportFields = updatedDataImport.getPopulatedFieldsAsMap();
+
+        for (String field : PAYMENT_FIELDS) {
+          Object newValue = updatedDataImport.get(field);
```
ptewson-sfdo, line 58: This file should be converted to 4-space tabs!

src/classes/GiftSelector.cls
```diff
@@ -43,4 +43,13 @@ public with sharing class GiftSelector {
                 AND Payment_Status__c=: ElevateTransactionStatus.AUTHORIZED.name()
         ];
     }
+
+    public DataImport__c getGiftById(Id giftId) {
```
ptewson-sfdo, line 47: If this is called "getGiftById" should it return an instance of gift instead of a DataImport__c? Or does this class actually maintain that boundary so it's appropriate?

src/classes/Gift_Test.cls
```diff
@@ -0,0 +1,55 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+@isTest
+private class Gift_Test {
+    
+    @isTest
+    static void validateThatGiftNeedsToBeReauthorized() {
+        DataImport__c dataImport = new DataImport__c(
```
ptewson-sfdo, line 36: (NON-BLOCKING) Would it be possible to build this test without DML? Perhaps by mocking GiftService?

ptewson-sfdo (commented):
Looks pretty good overall, a few comments/questions/suggestions!

src/classes/Gift.cls
```diff
@@ -0,0 +1,80 @@
...
+
+    public Boolean isGiftReauthorizedRequired() {
+      if (dataImport.Payment_Status__c != ElevateTransactionStatus.AUTHORIZED.name()) { 
+        return false; 
+      }
+
+      return this.hasPaymentFieldsChanged();
+    }
+
+    public void populateAvailableFields() {
```
psadie-sfdo, line 55: This should be private

src/classes/Gift.cls
```diff
@@ -0,0 +1,80 @@
...
+      for (String field : dataImportFields.keySet()) {
+        if (GiftBatchService.PAYMENT_FIELDS.contains(field)) {
+          fields.add(field);
+        }
+      }
+
+      this.availableFields = fields;
+    }
+
+    public Boolean hasPaymentFieldsChanged() {
```
psadie-sfdo, line 68: This should be private

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1004,6 +1050,7 @@ export default class GeFormRenderer extends LightningElement{
     reset() {
         this.resetFormState();
         this._openedGiftId = null;
+        this.clearErrors();
```
psadie-sfdo, line 1053: +++ doing a similar thing in #6404

src/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -140,12 +140,27 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
...
 
+    get showCancelButton() {
+        return this.isEdit 
+            && !this.isExpiredTransaction
+            && !this.isPaymentStatusAuthorized();
+    }
+
     get isExpiredTransaction() {
-        return this.paymentStatus() === 'EXPIRED';
+        return this.paymentStatus() === this.paymentTransactionStatusValues.EXPIRED;
```
psadie-sfdo, line 159: 👍

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -325,7 +327,13 @@ export default class GeFormRenderer extends LightningElement{
         });
     }
 
+    handleWidgetStateChange(changeEvent) {
+        this._isElevateWidgetInDisabledState = (changeEvent.state === 'readOnly');
```
psadie-sfdo, line 331: I wouldn't consider this a blocker and we can keep this as-is.

I'm just a bit uneasy about "spilling" the tokenize widget's internal display state values to the form renderer. But if we keep this, this dispatch logic should be in the 'readOnly' onEnter method.

psadie-sfdo (commented):
Just a few comments. I'll take another pass once all the latest updates are in. Thanks David!

src/classes/GiftBatchService.cls
sokosaurus: I hear what you're saying and at the same time, couldn't think of a better name myself - so, I went ahead and updated the list to use your suggested name of `PAYMENT_INTEGRATION_FIELDS`.

src/lwc/geFormRenderer/geFormRenderer.js
sokosaurus: Thanks for raising your concern here. I now get that ideally the widget's internal display state values should stay well, internal.

Ultimately I need a way for geFormRender to know when the widget is in a read-only state as we're using that for one of the requirements to determine whether or not we should tokenize. I believe there are mechanisms in place already that _should_ be doing this, but it did no appear to be working. Perhaps this is just a band-aid for now, but ideally - it'd be great to connect with you and figure out what the ideal path forward is. If this is a non-blocker, maybe we can look into getting this correctly developed in #6396 ?

src/classes/GiftSelector.cls
sokosaurus: That's a good question that I feel like the other developers would be able to better answer. I know for me, I using the naming already established where `getAuthorizedGiftsFor()` on the class returns a `List<DataImport__c>`. Please let me know if you think any adjustments to the naming is necessary here.

sokosaurus:
@kenlewis-sfdo @psadie-sfdo @voduyemi @ptewson-sfdo - Thank you all for the great feedback you've sent my way for this PR. I _believe_ I've made all the changes discussed/requested, so I'd really appreciate another review when you have a moment. Thanks!

src/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: I'm with ya. The existing mechanism is tied to the "Do not enter payment" behavior.

The event name is "doNotChargeState" (bad name). Prior to the last round of changes in this component, that one event had dual responsibilities for setting the "isElevateWidgetDisabled" property in the renderer for the "Do not enter payment" user action and for when the widget went into read-only mode.

I think ideally we rename that event and additionally start firing it from the `DISPLAY_DEFINITIONS.readOnly.onEnter` and `DISPLAY_DEFINITIONS.readOnly.onExit` as needed.

Let me know what you think

src/classes/Gift.cls
```diff
@@ -0,0 +1,89 @@
...
+        }
+        set;
+    }
+    
+    public Gift(DataImport__c dataImport) {
+        this.dataImport = dataImport;
+        this.populateAvailableFields();
+    }
+
+    public Boolean isGiftReauthorizedRequired() {
```
kenlewis-sfdo, line 56: This should probably just be named "isAuthorized()".  Reauth required is dependent on context.

src/classes/Gift_Test.cls
```diff
@@ -0,0 +1,55 @@
...
+            Payment_Status__c = ElevateTransactionStatus.AUTHORIZED.name(),
+            Payment_Elevate_Batch_ID__c = 'test-valid-group-id',
+            Donation_Amount__c = 3.00
+        );
+
+        insert dataImport;
```
kenlewis-sfdo, line 42: Agree that this probably doesnt need to be inserted.

src/classes/Gift_Test.cls
```diff
@@ -0,0 +1,55 @@
...
+
+        Test.startTest();
+
+        dataImport.Donation_Amount__c = 5.00;
+        Gift authorizedGift = new Gift(dataImport);
+        Boolean needsReauthorization = authorizedGift.isGiftReauthorizedRequired();
+
+        Test.stopTest();
+
+        System.assertEquals(dataImport.Id, authorizedGift.dataImport.Id);
```
kenlewis-sfdo, line 52: I'd remove this line from the test and just have this test use in-memory DI and assert the isAuthorized() method.

src/classes/GiftSelector.cls
```diff
@@ -43,4 +43,16 @@ public with sharing class GiftSelector {
                 AND Payment_Status__c=: ElevateTransactionStatus.AUTHORIZED.name()
         ];
     }
+
+    public DataImport__c getGiftById(Id giftId, List<String> fieldsToQuery) {
```
kenlewis-sfdo, line 47: Looks great!

kenlewis-sfdo (commented):
Looking really good, David! Left a few comments, let me know if you'd like to discuss anything.

src/lwc/geFormRenderer/geFormRenderer.js
ptewson-sfdo: @psadie-sfdo @sokosaurus Are we going to have conflicts/overlaps to untangle between this PR and #6404 do you think?

sokosaurus: I've tried to shuffle some things around to minimize conflicts but I wouldn't be surprised if there were a few remaining things that we'd need to ultimately untangle here.

src/classes/Gift_Test.cls
sokosaurus: Given that this work item has been kicked to the next sprint, I'm going to push to get this test class updated before we merge. If I run out of time, then I'll make sure to log this as a tech debt item so we can revisit it at a future date. LMK if you have any concerns.

src/classes/Gift_Test.cls
bdvorachek: This'll let us remove the `@TestVisible` on `dataImport` in Gift.cls too.

src/classes/Gift.cls
sokosaurus: Feel free to push back against this... just wondering if `isAuthorized()` makes the most sense for this method in its current structure. 

To me, a gift being authorized is a state and preventing an edit of an authorized gift is a constraint. One is certainly dependent on the other. My concern is that if we move to `isAuthorized`, we lose resolution to one of those items. A gift can be authorized. At the same time, we may or may not want to prevent that authorized gift from being edited depending on the scenario.

I feel like what you're proposing tracks against the notion of a given "type" being more than just a data structure, but also having behavior and constraints baked into it. Perhaps this is a larger conversation on the Gift entity as a whole? Happy to chat about this more.

ptewson-sfdo: What about something like "allowsChange()" or something better but still includes the specific context in the name? Or maybe even "requiresReauthorizationBeforeChange()" or something?

sokosaurus: `requiresReauthorizationBeforeChange()` feels like the winner to me. Let me go ahead and swap out the methods name right now and then we can always adjust to something else anytime thereafter if need be.

PR #6422, opened 2021-05-18
Customizable Rollups: Update queries to use Skew Mode flag

npsp-reedestockton:
We updated our Customizable Rollup queries to use the Skew Mode flag on Account and Contact to determine whether or not to use Skew Mode for most Customizable Rollups jobs.
# Critical Changes

# Changes
 - When Force Skew Mode is selected on an Account or Contact, Customizable Rollups uses Skew Mode to roll up data for those records.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/CRLP_RollupBatch_SVC.cls
+++ b/src/classes/CRLP_RollupBatch_SVC.cls
@@ -60,51 +60,19 @@ public class CRLP_RollupBatch_SVC {
      * @return A portion of a WHERE clause without WHERE or AND; or an empty string
      */
     public static String getSkewWhereClause(CRLP_RollupProcessingOptions.RollupType jobType, CRLP_RollupProcessingOptions.BatchJobMode jobMode,
-            String parentRelationshipField) {
+        String parentRelationshipField) {
         String queryFilter = '';
         Integer maxRelatedOppsForNonLDVMode = CRLP_Rollup_SVC.getMaxRelatedOppsForNonSkewMode();
 
-        // If the Summary Object is the Account, then filter on Accounts that have at least a single
-        // Opportunity attached. This is helpful to reduce the overall query size.
-        // To handle a scenario where an attached Opportunity was deleted, but the record not recalculated
-        // also include any records where the TotalGifts or TotalMemberships fields are not zero.
-        // If the field is null, a query with "< #" will not work. "= null" must be explicitly included in the query.
-        if (jobType == CRLP_RollupProcessingOptions.RollupType.ContactHardCredit ||
-                jobType == CRLP_RollupProcessingOptions.RollupType.AccountHardCredit) {
+        if (jobType != CRLP_RollupProcessingOptions.RollupType.GAU &&
+            jobType != CRLP_RollupProcessingOptions.RollupType.RecurringDonations
+            ) {
 
             if (jobMode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode) {
-                queryFilter = '(' + parentRelationshipField + 'npo02__NumberOfClosedOpps__c = null OR ' + parentRelationshipField +
-                    'npo02__NumberOfClosedOpps__c < ' + maxRelatedOppsForNonLDVMode + ') ' +
-                    'AND (' + parentRelationshipField + 'npo02__NumberOfMembershipOpps__c = null OR ' + parentRelationshipField +
-                    'npo02__NumberOfMembershipOpps__c < ' + maxRelatedOppsForNonLDVMode + ')';
+                queryFilter = '(' + parentRelationshipField + UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c') + ' = false)';
             } else {
-                queryFilter = '(' + parentRelationshipField + 'npo02__NumberOfClosedOpps__c >= ' + maxRelatedOppsForNonLDVMode +
-                        ' OR ' + parentRelationshipField + 'npo02__NumberOfMembershipOpps__c >= ' + maxRelatedOppsForNonLDVMode + ')';
+                queryFilter = '(' + parentRelationshipField + UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c') + ' = true)';
             }
-
-        } else if (jobType == CRLP_RollupProcessingOptions.RollupType.ContactSoftCredit) {
-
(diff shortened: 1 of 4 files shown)


npsp-reedestockton:
Tracking [W-9103417](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000ACroYAG/view)

npsp-reedestockton:
@lparrott @force2b I'd appreciate your review if you have the time.

@sam-knox I'd appreciate it if you'd take a look at the PR description.

dfuller114 (changes requested):
great work Reede! just a couple of refactor suggestions.

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -518,15 +511,10 @@ public inherited sharing class CRLP_RollupProcessor {
...
-            updatedRecord.put(this.qualifiedSkewField, true);
-            needsUpdate = true;
-
-        // Reset any flags erroneously set for NonSkewMode records
-        } else if (mode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode && this.useForceSkewModeField && isForceSkewModeTrue) {
-            updatedRecord.put(this.qualifiedSkewField, false);
-            needsUpdate = true;
+        // Check whether skew mode threshold has been exceeded
+        initForceSkewModeValues(parent.getSObjectType());
+        if (mode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode && this.useForceSkewModeField) {
```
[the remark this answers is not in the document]
npsp-reedestockton: The class variable useForceSkewModeField was introduced in response to feedback from @force2b to avoid making the call repeatedly since it will be the same for every call in a class instance.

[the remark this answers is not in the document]
force2b: Yes. This method is called in a loop, so the goal was to only initialize the var a single time given that it's retrieving the sobject token for 2 object types. When rolling up opps for millions of contacts or account this can really add up. I'm not opposed to a different pattern either, but overall code should avoid repeating logic in a loop when it can be done a single time [for the batch scope].

src/classes/CRLP_RollupBatch_SVC.cls
```diff
@@ -60,52 +60,20 @@ public class CRLP_RollupBatch_SVC {
...
+        if (jobType != CRLP_RollupProcessingOptions.RollupType.GAU &&
+            jobType != CRLP_RollupProcessingOptions.RollupType.RecurringDonations
+            ) {
 
-            // TODO -- How to get Accounts that require using this method for Soft Credits??
-            // For now just use the NumberOfClosedOpportunities field as a proxy for the number of SoftCredits
-            String softCreditField = parentRelationshipField + 'npo02__NumberOfClosedOpps__c';
             if (jobMode == CRLP_RollupProcessingOptions.BatchJobMode.NonSkewMode) {
-                queryFilter = '(' + softCreditField + ' = null OR ' + softCreditField + ' < ' + maxRelatedOppsForNonLDVMode + ')';
+                queryFilter = '(' + parentRelationshipField + UTIL_Namespace.StrTokenNSPrefix('CustomizableRollups_UseSkewMode__c') + ' = false)';
```
lparrott, line 72: This is so much cleaner!

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -547,6 +535,30 @@ public inherited sharing class CRLP_RollupProcessor {
         }
     }
 
+    private Boolean isSkewModeNeeded(SObject updatedRecord) {
```
force2b, line 538: Block comments needed on these.

lparrott (approved):
I can't speak to the functionality, but the code looks good 👍

src/classes/CRLP_RollupProcessor.cls
```diff
@@ -547,6 +535,30 @@ public inherited sharing class CRLP_RollupProcessor {
...
+    }
+
+    private Boolean isRollupFieldAtThreshold(SObject summaryObject) {
+        Integer maxRelatedOppsForNonLDVMode = CRLP_Rollup_SVC.getMaxRelatedOppsForNonSkewMode();
+
+        if ((Decimal) summaryObject.get('npo02__NumberOfClosedOpps__c') >= maxRelatedOppsForNonLDVMode ||
+            (Decimal) summaryObject.get('npo02__NumberOfMembershipOpps__c') >= maxRelatedOppsForNonLDVMode) {
+            return true;
+
+        } else if (summaryObject.getSObjectType() == Contact.getSObjectType() &&
```
force2b, line 553: This check for sobject type, which would be in a loop, falls into the category as why `useForceSkewModeField` is only instantiated once. The determination of whether this is for a Contact could be made earlier - once.  There's a couple of ways to do that too, so whatever feels best. 

I did do a quick test too. Calling `getSObjectType()` in a loop takes around 3x more milliseconds than just checking a boolean var in a. loop. Not a perfect test, but I wanted to make sure that the call to `getSobjectType()` was still relatively slow. Though, if my calculations are accurate, a rollup of 1M Contacts with a batch size of 200 would save around 80 seconds over the of the batch job in this one operation. I suppose combined with the previous checks for object for `useForceSkewMode` that would double to 160 seconds - nearly 3 minutes. Not a gigantic savings by any stretch. It's good data to have, and maybe 3 minutes over 1M Contacts is not worth the extra work to use a getter to limit the calls. So, I'll leave it up to the team to decide. FYI @dfuller114

force2b (commented):
This looks really good overall. Made a couple of suggestions, and added along comment about perf optimization which in the end may not matter.

sam-knox:
@npsp-reedestockton - Does this PR represent the completion of this feature? Is there any other pending work before customers can use this?

npsp-reedestockton:
@sam-knox This PR makes the sequence fully usable by our customers.  We have further plans to make Customizable Rollups self-healing in certain circumstances, but the change to our query structure is complete with this work item.

npsp-reedestockton:
@force2b @dfuller114 I've implemented most of your code review suggestions.  I did not eliminate the needsUpdate boolean for the reasons explained earlier.

npsp-reedestockton:
@force2b @dfuller114 I've reworked a few things to simplify/clarify.  Hope this helps.

force2b (approved):
Great work!

dfuller114 (approved):
looks good! thanks Reede!

PR #6425, opened 2021-05-19
W-9251993 - Fixing where DI GAU allocations are placed

callen-sfdo:
- Previously we attached DI GAU Allocations to the payment level if payment allocations were enabled since we relied on two-way sync to populate the GAU allocations on the Donation level.  Now that we are using Allo_AllocationsService it is no longer necessary, and the DI GAU allocations should be attached to either Donation or Recurring donation depending on which is present.
- Added a unit test to specifically test that partial GAU allocations will be correctly handled with the new logic.
- Also cleaned up a few typos and other things I spotted.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


grivero (approved):
If the definitions of labels and custom metadata was already pushed in a different PR, then everything looks good.

PR #6436, opened 2021-05-24
Recurring donation creation in BDI and BDI GAU Allocation overhaul

callen-sfdo:
This branch contains all the previously QA'ed work from the existing feature/RDInBDI branch, and I'm looking to merge it into feature/232 since we are targeting the Summer release.  This work pre-dates the feature branch strategy, and was not mature enough until recently to merge in, which is why it is coming now as one PR.  Going forward all related work will be merged into 232 on a story by story basis. 

The major changes are:

- New fields on the Data Import object for Recurring Donations, and new Object Groups / Field mappings mapping those fields to the existing Recurring Donation object.
- New logic in BDI to treat Recurring Donations as a "core" object (i.e. not an additional object ) that has its own logic class that will execute after contacts and accounts are inserted and immediately before BDI_Donations in the main BDI_DataImportService process method
- Some changes to other BDI code to allow Recurring donations to have either Contact1 or Account1 as a parent.
- An overhaul to how GAU allocations are handled in BDI in order to support RD GAU Allocations, and also to allow GAUs on the DI to override Default / Campaign GAU for new Donations/Payments.
-- Now if there are GAU allocations on the DI record we are going to disable the normal allocation triggers, insert the core objects (ie Contact/Account/Recurring Donation/Donation/Campaign, then insert the GAU Allocations at either the Recurring Donation or Donation level, then call ALLO_AllocationsService to roll down the GAU allocations appropriately, then re-enable the allocation triggers.  More detail on this can be found in the below linked Lucidchart.
- As part of the BDI GAU overhaul we are now checking and throwing an error if there are any existing GAU Allocations on either the Donation or Payment (previously this would just allow invalid GAU Allocation combinations to be created).  
- Also as part of the BDI GAU overhaul We will also be checking to make sure that the allocation percentages for GAU Allocation 1 & 2 do not exceed 100%.

https://lucid.app/lucidchart/invitations/accept/inv_6092942c-d19a-415e-bb2b-eb940d859653

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/BDI_CustObjMappingGAUAllocation.cls
```diff
@@ -73,9 +72,12 @@ public with sharing class BDI_CustObjMappingGAUAllocation extends BDI_ObjectMapp
                     if (objWrap.sObj.Id == null || targetFieldDescribe.isUpdateable()) {
                         // If the target field mapping is one of the key parent objects, then extract the values for later logic
                         if (fieldMapping.Target_Field_API_Name == PAYMENT_FIELDNAME) {
-                            paymentIdValue = value;
+                            // Skip payment lookup since we will not directly link a GAU Allocation to Payment
+                            continue;
```
lparrott, line 76: If a current customer has a mapping like this, will they be informed with an error that it no longer works? I guess it will probably reach a point where we don't have payment/object information.

lparrott (approved):

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,269 @@
...
+     * @param dataImports he data import records to examine for GAU Allocation Information
+     * @return Map<Id, DataImport__c> the DI records with GAU information in a map.
+     */
+    public static Map<Id, DataImport__c> getDIsWithGAUtoProcess(DataImport__c[] dataImports) {
+        Map<Id, DataImport__c> diWithGAUToProcess = new Map<Id, DataImport__c>();
+        // Loop through and determine which records have Additional Object JSON containing GAUAllocation objects or
+        // the DI GAU fields filled out
+        for (DataImport__c di : dataImports) {
+            if (di.Additional_Object_JSON__c != null) {
+                if (di.Additional_Object_JSON__c.contains('GAU_Allocation')) {
```
ptewson-sfdo, line 54: What's actually being checked here? Can we avoid that string literal by using the sObject Name from Describe or something else that avoids this hardcode?

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,269 @@
...
+            if (di.GAU_Allocation_1_Percent__c != null) {
+                totalPercent += di.GAU_Allocation_1_Percent__c;
+            }
+            if (di.GAU_Allocation_2_Percent__c != null) {
+                totalPercent += di.GAU_Allocation_2_Percent__c;
+            }
+
+            // if the total percent is over 100 then fail the record and log an error.
+            if (totalPercent > 100) {
+                bdi.LogBDIError(di,label.bdiErrorGAUAllocationOver100,'GAU_Allocation_1_Import_Status__c');
```
ptewson-sfdo, line 91: SObjectType.Allocation__c.fields.GAU_Allocation_1_Import_Status__c.Name

src/classes/BDI_GAUAllocationsUtil.cls
```diff
@@ -0,0 +1,269 @@
...
+                        'Payment__r.npe01__Opportunity__c '+
+                        'FROM Allocation__c WHERE Opportunity__c IN: opptIds';
+
+                if (alloSettings.Payment_Allocations_Enabled__c) {
+                    queryString += ' OR Payment__r.npe01__Opportunity__c IN: opptIds';
+                }
+
+                Allocation__c[] existingAllos = Database.query(queryString);
+
+                if (existingAllos != null){
```
ptewson-sfdo, line 159:
```suggestion
                if (existingAllos != null) {
```

src/classes/BDI_GAUAllocationsUtil.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>50.0</apiVersion>
```
ptewson-sfdo, line 3:
```suggestion
    <apiVersion>51.0</apiVersion>
```

src/classes/BDI_GAUAllocationsUtil_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>50.0</apiVersion>
```
ptewson-sfdo, line 3:
```suggestion
    <apiVersion>51.0</apiVersion>
```

src/classes/BDI_RecurringDonations.cls-meta.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
ptewson-sfdo, line 3:
```suggestion
    <apiVersion>51.0</apiVersion>
```

src/classes/BDI_RecurringDonations_TEST.cls-meta.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
ptewson-sfdo, line 3:
```suggestion
    <apiVersion>51.0</apiVersion>
```

src/classes/BDI_RecurringDonations.cls
```diff
@@ -0,0 +1,293 @@
...
+    * @description constructor for the BDI Recurring Donations helper
+    * @param bdi The Batch Data Import Service the helper should use
+    * @return void
+    */ 
+    public BDI_RecurringDonations(BDI_DataImportService bdi) {
+        this.bdi = bdi;
+    }
+
+
+    private final String IMPORT_STATUS_FIELD = UTIL_Namespace.StrTokenNSPrefix('RecurringDonationImportStatus__c');
```
ptewson-sfdo, line 62: SObjectType.DataImport__c.fields.RecurringDonationImportStatus__c.Name

ptewson-sfdo (commented):
I know this has cleared QA, but we really should be avoiding string literals for field names whenever possible, and definitely should be giving an updated API version (51) for any new Apex classes, as well as updating when we make alterations to pre-existing classes when we can!

ptewson-sfdo (approved):

src/classes/BDI_CustObjMappingGAUAllocation.cls
callen-sfdo: Theoretically yes, but it is an extremely unlikely scenario that they would be modifying mappings on the version of the GAU allocation that we ship with the product to map a different value into the payment field. Any changes to mapping for the lookups on GAU Allocation would likely result in failures before or after this change.

src/classes/BDI_GAUAllocationsUtil.cls
callen-sfdo: We're checking the additional object JSON, which could theoretically contain many different object mapping types to see if it contains a GAU allocation, which would require additional validation and processing later in the logic.

src/classes/BDI_GAUAllocationsUtil.cls
callen-sfdo: I'll change this in the ticket I'm currently working on.

PR #6440, opened 2021-05-25
Feature/232  change capture group to elevate batch

ptewson-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/GE_GiftEntryController.cls
```diff
@@ -42,12 +42,12 @@ public with sharing class GE_GiftEntryController {
     private static GiftBatchService giftBatchService = new GiftBatchService();
 
     @TestVisible
-    private static ElevateCaptureGroupService captureGroupService {
+    private static ElevateBatchService elevateBatchService {
```
ptewson-sfdo, line 45: @kenlewis-sfdo With this terminology shift, there are a couple places that feel like we are bleeding the elevate batch concept - which is specific to the Elevate adaptor - way too high up into our application. These references in GE_GiftEntryController feel like one instance, but curious to see if you see any good opportunities or approaches to improve this here!

src/classes/PS_Request_TEST.cls
```diff
@@ -147,9 +147,9 @@ private with sharing class PS_Request_TEST {
                 || endpoint == PS_Request.ElevateEndpoint.COMMITMENT_CANCEL
                 ) {
                 roles = PS_Request.ROLE_PAYMENT_COMMITMENTS;
-            } else if (endpoint == PS_Request.ElevateEndpoint.ADD_TO_CAPTURE_GROUP
-                || endpoint == PS_Request.ElevateEndpoint.CREATE_CAPTURE_GROUP
-                    || endpoint == PS_Request.ElevateEndpoint.CHARGE_CAPTURE_GROUP) {
+            } else if (endpoint == PS_Request.ElevateEndpoint.ADD_TO_ELEVATE_BATCH
+                || endpoint == PS_Request.ElevateEndpoint.CREATE_ELEVATE_BATCH
+                    || endpoint == PS_Request.ElevateEndpoint.CHARGE_ELEVATE_BATCH) {
```
psadie-sfdo, line 152: alignment seems off here

src/classes/GE_GiftEntryController.cls
psadie-sfdo: 2 cents incoming, I think this is ok. Referencing the elevate batch service in the controller is more ideal than directly using the elevate batch domain and/or entity imo.

Something to discuss maybe is whether or not Elevate batches actually belong to a Gift Batch and so should these calls to the elevate batch service actually be behind the gift batch/gift batch service?

src/classes/ElevateBatch.cls
```diff
@@ -27,38 +27,38 @@
...
-    public ElevateCaptureGroup(ElevateCaptureGroupResponse captureGroupResponse) {
-        this.groupId = captureGroupResponse.id();
+    public ElevateBatch(ElevateBatchResponse elevateBatchResponse) {
+        this.elevateBatchId = elevateBatchResponse.id();
     }
 
-    public ElevateCaptureGroup create() {
-        return captureGroupService.getCaptureGroup();
+    public ElevateBatch create() {
+        return elevateBatchService.getelevateBatch();
```
psadie-sfdo, line 54:
```suggestion
        return elevateBatchService.getElevateBatch();
```

psadie-sfdo (changes requested):

src/classes/GE_GiftEntryController.cls
kenlewis-sfdo: I created this diagram to help us in that discussion - I think this is what we should refactor toward as this work continues to evolve: https://lucid.app/lucidchart/invitations/accept/inv_1d610dfc-bcb1-4ee9-8833-db0f6e52e552

kenlewis-sfdo (commented):
Patrick, this looks really good!  Removing the CaptureGroup interface is a good call, and overall I think this work will really help reduce translation, and sets us up for continued pushing of the Elevate code into the adapter layer.  I'll look again for any ideas for further refactor, but I think this is pretty solid as is and would support it moving forward as is here.

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -710,13 +710,13 @@ export default class GeFormRenderer extends LightningElement{
             try {
                 this.loadingText = this.CUSTOM_LABELS.geAuthorizingCreditCard;
     
-                const currentCaptureGroup = new ElevateCaptureGroup(this.latestCaptureGroupId);
-                const authorizedGift = await currentCaptureGroup.add(tokenizedGift);
+                const currentElevateBatch = new ElevateBatch(this.latestElevateBatchId);
```
sokosaurus, line 713: We are no longer passing an Id value into the constructor for ElevateCaptureGroup/ElevateBatch as per #6404, which has yet to be merged. Just calling it out here in the event that there are merge conflicts and also, to ensure that updated logic doesn't get inadvertently overwritten.

src/classes/GE_GiftEntryController.cls
ptewson-sfdo: Thanks for sharing (and creating) that diagram, @kenlewis-sfdo! If I'm correctly understanding the vision, we would eventually want to avoid putting anything specific to Elevate into the UI layer - and so the use of the ElevateAuthorizedGift and ElevateTokenizedGift in the AuraEnabled methods of this controller would have to revisited?

kenlewis-sfdo: Yep, definitely.  The controller should be making calls to an NPSP service layer (something like "GiftEntryService") to add the Gift to a Gift Batch.  The Service layer should instantiate and coordinate among NPSP Domain classes to get work done.  

Once the GiftBatches are configured to use an instance of IPaymentServices (defining an `add` and a `process` method), then the *implementation* will use ElevateBatches etc.

So yes, ElevateTokenizedGift and ElevateAuthorizedGift should both be renamed and/or removed from the controller.

kenlewis-sfdo (approved):
Looks great, Patrick.  This is great and continues to push in the direction of ideal design.  Thank you!

PR #6449, opened 2021-05-27
BGE Process Flow Improvement

voduyemi:
# Critical Changes
WI: [W-8769060](https://gus.lightning.force.com/a07B00000094FNTIA2)

Implemented the new Batch Gift Entry Processing Flow. By clicking on **Process Batch** or **Process Batch & Payments**, we kick off batch processing in the Gift Entry app while we periodically poll for the status of the gifts in the batch to determine when batch processing is done.

Custom Labels
- geBatchProcessingSuccess
- geProcessingBatch

Picklist Value
- Dataimport__c.Status__c : Processing

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.html
```diff
@@ -7,38 +7,54 @@
...
-            loading-text={loadingText}
-            batch-currency-iso-code={batchCurrencyIsoCode}>
-    </c-ge-form-renderer>
+    <div if:true={isBatchProcessing} class='spinner-container'>
+        <div class='slds-card'>
+            <lightning-spinner alternative-text={CUSTOM_LABELS.messageLoading}
+                               variant='brand'
+                               size='large'>
+            </lightning-spinner>
+            <p class='loading-text slds-text-heading_medium'>Processing Batch...</p>
```
voduyemi, line 21: Custom label to be added soon!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -289,55 +291,70 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
-            true
+            })
+        }, 5000);
+    }
+
+    handleProcessedBatch() {
+        this.collapseForm();
+        showToast(
+            this.CUSTOM_LABELS.PageMessagesConfirm,
+            this.batchName + ' was processed.',
```
voduyemi, line 338: Custom label coming soon!

melissabarber:
@voduyemi Reviewed labels. They look good to go! CC: @psadie-sfdo @sokosaurus

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.html
```diff
@@ -7,38 +7,54 @@
...
-            batch-id={recordId}
-            onsubmit={handleSubmit}
-            onsectionsretrieved={handleSectionsRetrieved}
-            onpermissionerror={handlePermissionErrors}
-            ontogglemodal={handleToggleModal}
-            onnavigate={handleNavigateEvent}
-            loading-text={loadingText}
-            batch-currency-iso-code={batchCurrencyIsoCode}>
-    </c-ge-form-renderer>
+    <div if:true={isBatchProcessing} class='spinner-container'>
```
psadie-sfdo, line 15: Not a blocker. I prefer having these `if:true` directives on their own `template` ala `<template if:true={isBatchProcessing}>` personally. I think it's easier to tell at a glance which elements are included in the conditional rendering instead of having the directive on an element that may also have classes and other attributes.

psadie-sfdo (changes requested):
This is looking pretty good with a first pass. Spinning up an org to run through some dev tests, but wanted to drop these initials comments right away.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -253,7 +254,7 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
         recordId: '$recordId',
         fields: [
             BATCH_ID_FIELD,
-            BATCH_NAME
+            BATCH_NAME,
```
sokosaurus, line 257: This is really minor but the comma at the end of line 257 should be removed

```suggestion
            BATCH_NAME
```

sokosaurus (changes requested):
This is looking REALLY good @voduyemi! Awesome work here. I'm going to follow @psadie-sfdo's lead and run this PR through some dev tests in my own scratch org. I'll let you folks know if I come across anything interesting.

src/classes/GiftBatch.cls
```diff
@@ -97,6 +97,12 @@ public inherited sharing class GiftBatch {
         gifts.failGiftsByCaptureIdWithReason(failReasonByCaptureGroupId);
     }
 
+    public void processGifts() {
+        gifts.giftsToProcessingStatus();
+        captureAuthorizedGifts();
+        BDI_DataImport_API.processDataImportBatches(new List<Id>{batchId});
+    }
```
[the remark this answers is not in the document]
voduyemi: Execution will be halted and we should catch the exception in `GE_GiftEntryController` and an `AuraHandledException` should be bubbled up to the client-side.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.html
voduyemi: Updated!

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -289,55 +293,72 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
     }
 
+    collapseForm() {
+        const form = this.template.querySelector('c-ge-form-renderer');
+        form.collapse();
+    }
+
     async refreshBatchTotals() {
-        if(this.isElevateCustomer) {
+        if (this.isElevateCustomer) {
```
ptewson-sfdo, line 355: With this change to move the batch totals outside of this condition - which totally makes sense - I think we might drop this condition entirely? setting `_hasDisplayedExpiredAuthorizationWarning `to false shouldn't really have any impact to non-elevate customers and I think we can simplify by avoiding the` if (this.isElevateCustomer)` call entirely?

src/classes/Gifts.cls
```diff
@@ -54,8 +54,24 @@ public with sharing class Gifts {
         return giftsToReturn;
     }
 
+    public void giftsToProcessingStatus() {
+        giftsToUpdate.clear();
+        for (DataImport__c gift : dataImports) {
+            if (isProcessableGift(gift)) {
+                gift.Status__c = BDI_DataImport_API.bdiProcessing;
```
ptewson-sfdo, line 61: @pedrodalcol @lparrott @callen-sfdo - just want to give the heads up that we are introducing the "Processing" status for DI here! Should not impact anyone as this PR are only sets the value within the context of Gift Entry

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -279,6 +282,7 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
             } catch (error) {
                 handleError(error);
             } finally {
+                this._isBatchProcessing = true;
                 if (this.shouldDisplayExpiredAuthorizationWarning()) {
                     this.displayExpiredAuthorizationWarningModalForProcessAndDryRun(
                         async () => {
```
ptewson-sfdo, line 288: Now that we aren't navigating away from this page on batch process, I think we need to explicitly close this modal - so this should be: 

```
async () => {
    await this.processBatch();
    this.dispatchEvent(new CustomEvent('closemodal')); 
}
```

ptewson-sfdo (commented):
This looks really close! The only comment I might consider blocking is related to closing the "auths expired" modal when you hit process batch

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
voduyemi: Updated!

src/classes/GiftBatchService.cls
```diff
@@ -50,10 +50,9 @@ public with sharing class GiftBatchService {
         return groupedTotals.totalsFor(batchId);
     }
 
-    public void processPaymentsFor(Id batchId) {
+    public void processGiftsFor(Id batchId) {
         GiftBatch giftBatch = new GiftBatch(batchId);
-        giftBatch.captureAuthorizedGifts();
-        giftBatch.updateRecords();
+        giftBatch.processGifts();
```
kenlewis-sfdo, line 55: 👍🏽

src/classes/GiftBatchStatusGroup.cls
```diff
@@ -56,6 +56,10 @@ public inherited sharing class GiftBatchStatusGroup {
         return status == BDI_DataImport_API.bdiFailed;
     }
 
+    public Boolean isProcessing() {
+        return status == BDI_DataImport_API.bdiProcessing;
+    }
```
kenlewis-sfdo, line 61: "isProcessing" seems like a state of the GiftBatch, while this class to me feels like it represents a subset of the Gifts in a batch that all have the same Status.  I'd expect to ask the Batch whether it is processing: "giftBatch.isProcessing()"

There are also now both status and paymentStatus fields on this class, which both seem like properties of the GiftBatch.  

@psadie-sfdo how do you think this class is looking?

src/lwc/geGiftEntryFormApp/__tests__/geGiftEntryFormApp.test.js
```diff
@@ -0,0 +1,70 @@
...
+);
+
+describe('c-ge-gift-entry-form-app', () => {
+
+    afterEach(  () => {
+       clearDOM();
+       jest.clearAllMocks();
+    });
+
+    it('should render processing batch spinner if batch is still processing', async  () => {
```
kenlewis-sfdo, line 32: This looks great!  What would you think about adding tests to verify that the spinner starts when it is supposed to (process batch clicked and Batch has pmts to process), and stops when it is supposed to (last Gift is done processing)?

src/classes/GiftBatchStatusGroup.cls
psadie-sfdo: Yeah, it's clearer to me now that the various "totals" belong to the GiftBatch. I think I'm still ok with GiftBatchStatusGroup having the responsibility of consuming an AggregateResult and determining if the results are in various statuses/payment statuses. GiftBatchTotals then has the responsibility of holding the organized results for a GiftBatch.

_Maybe a tiny detail, but "isProcessing", "isImported", etc in GiftBatchStatusGroup should probably be "hasProcessing" or "hasImported" i.e. "this aggregate result has gifts that are processing" and "this aggregate result has gifts that are imported"._

Ultimately, I think GiftBatchTotals needs to move under and be a member of GiftBatch. Where we're directly referencing GiftBatchTotals in GiftBatchService we should be instantiating a GiftBatch and asking it for its totals.

voduyemi: Well said guys! I think we should try to make this change but in a different WI. Adding this to my doc.

src/lwc/geGiftEntryFormApp/__tests__/geGiftEntryFormApp.test.js
voduyemi: I would love to add the tests to cover the spinner not showing after processing, but it requires me to mock all the child components(`geFormRenderer` and `geBatchGiftEntryTable`) in `geGiftEntryFormApp` which is a lot to do right now. So I thought I thought I get the ball rolling with this one.

src/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -279,64 +282,89 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
-                if (this.shouldDisplayExpiredAuthorizationWarning()) {
-                    this.displayExpiredAuthorizationWarningModalForProcessAndDryRun(
-                        async () => {
-                            await this.processBatch();
-                        } 
-                    );
-                } else {
-                    await this.processBatch();
-                }
+                await this.startBatchProcessing();
```
ptewson-sfdo, line 285: :tada:

ptewson-sfdo (approved):
This looks good from my end! Nice work, Victor!

sokosaurus (approved):
Really, really solid work here Victor! Looks great to me!

psadie-sfdo (approved):

PR #6450, opened 2021-05-28
[RD] Hide disable button, fix payment method picklist's control of elevate widget

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000N2enYAC/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -235,6 +235,15 @@ export default class rd2EntryForm extends LightningElement {
...
      */
     handleElevateWidgetDisplayState(event) {
+
+        const shouldResetPaymentMethodInEditMode = this.isEdit
+            && event.isDisabled
+            && this._paymentMethod !== this.existingPaymentMethod
+            && this.isCommitmentEdit;
+
+        if(shouldResetPaymentMethodInEditMode) {
+            this.resetPaymentMethod();
```
lparrott, line 245: Great catch here, sometimes I hate how many routes there are through forms 😫

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -463,10 +472,17 @@ export default class rd2EntryForm extends LightningElement {
         const period = getFieldValue(this.record, FIELD_INSTALLMENT_PERIOD);
         const campaignId = getFieldValue(this.record, FIELD_CAMPAIGN);
 
-        return amount !== Number(allFields[FIELD_AMOUNT.fieldApiName])
-            || frequency !== Number(allFields[FIELD_INSTALLMENT_FREQUENCY.fieldApiName])
-            || period !== allFields[FIELD_INSTALLMENT_PERIOD.fieldApiName]
-            || campaignId !== allFields[FIELD_CAMPAIGN.fieldApiName];
+
+        const hasFrequencyField = allFields.hasOwnProperty(FIELD_INSTALLMENT_FREQUENCY.fieldApiName);
```
lparrott, line 476: Interesting, was there a situation that this field wasn't available? If so, my bad!

src/lwc/rd2EntryForm/__tests__/rd2EntryForm.test.js
```diff
@@ -133,6 +133,7 @@ describe('c-rd2-entry-form', () => {
             await flushPromises();
 
             const elevateWidget = controller.elevateWidget();
+            expect(controller.disableElevateButton()).toBeTruthy();
```
lparrott, line 136: These are great additions to the test! We'll have to set these conditions up as part of the design next time, I know this got a little out of hand.

lparrott (approved):
Great work! I'm glad we found these edge cases before the release.

src/lwc/rd2EntryForm/rd2EntryForm.js
bdvorachek: The frequency field is only present when we're using Advanced scheduling.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.html
```diff
@@ -23,26 +23,30 @@ <h1 class='slds-text-body_regular' data-qa-locator='heading'>{labels.elevateWidg
...
+                                                    onclick={handleUserCancelledWidget}
+                                                    class='slds-button slds-button_base slds-float_right hide-card-button'
+                                                    data-qa-locator='button Cancel Update Payment Information'>
+                                                </lightning-button>
+                                            </template>
                                         </template>
                                     </template>
-                                    <template if:false={isEditPayment}>
+                                    <template if:false={isEditMode}>
                                         <template if:false={isDisabled}>
```
dfuller114, line 39: Is it possible for the form to be in edit mode and be disabled at the same time? Was wondering if we still needed this isDisabled check here.

bdvorachek: Yep! This happens when you click Edit on a non-elevate RD2 record that is a "Credit Card" or "ACH" payment, the widget loads in Edit mode, but in the Disabled state where you can click "alternatively, enter payment details now".

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -83,12 +85,15 @@ export default class rd2ElevateCreditCardForm extends LightningElement {
     set paymentMethod(value) {
         this.notifyIframePaymentMethodChanged(value);
         this._paymentMethod = value;
+        this.enableWidgetOnPaymentMethodChange = true;
     }
 
     notifyIframePaymentMethodChanged(newValue) {
         const iframe = this.selectIframe();
         if(this.shouldNotifyIframe(newValue)) {
```
dfuller114, line 93: I think having this `shouldNotifyIframe` boolean logic inside of the notify function could lead to confusion since `notifyIframePaymentMethodChanged()` indicates that the notification is going to happen without condition. It might be better to wrap `notifyIframePaymentMethodChanged()` inside of `if (this.shouldNotifyIframe()` in the `paymentMethod` setter so that it reads a little clearer. Think we can run into issues down the road when we have conditionals inside functions when both are at different levels of abstraction since in terms of the call flow, checking if we need to send the notification comes before sending it and is not actually a part of sending it.

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
```diff
@@ -277,6 +299,11 @@ export default class rd2ElevateCreditCardForm extends LightningElement {
         }
     }
 
+    paymentMethodEdited(newPaymentMethod) {
```
dfuller114, line 302: could we change this function name to `isPaymentMethodChanged()`? We may want to come down on whether we want to use "edit" or "change" language in the codebase in regard to data changes to avoid any confusion. I think "change" is more often used for LWC UI/component state, vs "changed" is more used for field/data value changes.

src/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -235,6 +235,15 @@ export default class rd2EntryForm extends LightningElement {
      * @param event
      */
     handleElevateWidgetDisplayState(event) {
+
+        const shouldResetPaymentMethodInEditMode = this.isEdit
```
dfuller114, line 239: could we extract a function from this constant instead of having the block inside the `handle` function?

dfuller114 (changes requested):
just a couple of requested changes but overall looks good!

src/lwc/rd2ElevateCreditCardForm/rd2ElevateCreditCardForm.js
bdvorachek: That makes sense, that whole function is wrapped in a conditional minus a single line.

dfuller114 (approved):
Looks good!

PR #6471, opened 2021-06-05
[RD2] Fix duplicate Elevate payment on same day update

lparrott:
- Fix for bug that charges donor twice if Elevate Commitment is edited on day of payment.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/RD2_ScheduleService.cls
+++ b/src/classes/RD2_ScheduleService.cls
@@ -531,7 +531,7 @@ public without sharing class RD2_ScheduleService {
     public Date getNextDonationDateGreaterThan(RecurringDonationSchedule__c schedule, Date referenceDate) {
         Date firstValid = getFirstPossibleDonationDateFromStart(schedule);
 
-        if (firstValid > referenceDate) {
+        if (firstValid >= referenceDate) {
             return firstValid;
         }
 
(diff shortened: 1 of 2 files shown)


src/classes/RD2_ScheduleService.cls
```diff
@@ -531,7 +531,7 @@ public without sharing class RD2_ScheduleService {
     public Date getNextDonationDateGreaterThan(RecurringDonationSchedule__c schedule, Date referenceDate) {
         Date firstValid = getFirstPossibleDonationDateFromStart(schedule);
 
-        if (firstValid >= referenceDate) {
+        if (firstValid > referenceDate) {
```
npsp-reedestockton, line 534: I have serious concerns about this.  At the very least it requires rigorous testing.

npsp-reedestockton: I just tested this and it seems to work.  The only concern I had was around the first payment date, and that works as expected, so I think this is okay.

npsp-reedestockton (approved):

src/classes/RD2_ScheduleService.cls
lparrott: I had the same concern, it's honestly a little weird that it works, maybe we need a better understanding of `firstOccurrenceOnTimestamp` (that's where this Date ends up)

andrewyu-salesforce (approved):
Everything works with the new change on both same and future date update.

PR #6473, opened 2021-06-07
W-8436131 NPSP Error - Fill the Related Record ID

mldyang:
Part 1 of 2 for adding Related Record Id to NPSP Error Log: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008iywTIAQ/view
Currently, npsp.ERR_Handler_API only takes in Exception + String to process errors. In order to create NPSP Error records with Related Record Id for DML Errors, we need a way to pass in Database.Results. 

(Part 2 of 2 will be done in Elevate NPSP Connector repo to utilize the global method created here)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


mldyang:
Hi @lparrott - added you as a reviewer as this is part of the work requested by your team, so wanted to give you a heads up on what we're doing.
fyi @force2b - new global method in case you have any concerns. Thank you!

src/classes/ERR_Handler_API.cls
```diff
@@ -88,9 +88,27 @@ global class ERR_Handler_API {
...
+
+    /******************************************************************************************************
+     * @description Creates error records from DML results, stores in database (if enabled) and sends error
+     *              notification (if enabled)
+     * @param dmlResult The results of a single DML operation performed using the Database class.
+     * @param records The records the DML operation was performed on.
+     * @param context The context the errors were raised on
+     */
+
+    global static void processDMLError(List<Object> dmlResult, List<SObject> records, String context) {
```
force2b, line 105: Since this accepts a collection of SObjects, I recommend a method name of `processDMLErrors()`.

src/classes/ERR_Handler_API.cls
```diff
@@ -88,9 +88,27 @@ global class ERR_Handler_API {
...
+
+        ERR_Handler.Errors errors = ERR_Handler.getErrors(dmlResult, records);
```
force2b, line 107: This implementation will end up in `ERR_Handler.getErrors(List<Object> dmlResults, List<SObject> sObjects, Boolean displayErrors)` where displayErrors is true. That will cause the code to issue `ApexPages.addMessage()` if there is an error and a current apex page. Is that the intent, or could this method include a 3rd parameter of false to prevent errors from being rendered in the UI?

mldyang: Thank you @force2b! I updated the PR based on both feedback items you mentioned.

force2b (changes requested):
Would you mind adding a unit test for the new method?

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
+/*
+    Copyright (c) 2014, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2021, Salesforce.org
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description test class for Error Handling API
+* @group ErrorHandling
+*/
+@isTest
```
force2b, line 36:
```suggestion
@isTest(IsParallel=true)
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+private class ERR_Handler_API_TEST {
+
+    private static final String TEST_ACCOUNT_NAME = 'ERR_Handler_API Acct';
+    private static final String ERR_CONTEXT = 'ERR_Handler_API_TEST';
+
+    @TestSetup
+    static void setupInitalAccount() {
+        Account account = new Account(Name = TEST_ACCOUNT_NAME);
```
force2b, line 44: Minor difference, but it's useful to use the test data factory. You also won't need `TEST_ACCOUNT_NAME` and the queries in the test methods can just select the first Account since there will always only be a single Account record in the system.
```suggestion
        Account account = UTIL_UnitTestData_TEST.createMultipleTestAccounts(1, null)[0];
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+        acct.Name = null;
+        List<Account> accounts = new List<Account>{acct};
+
+        Test.startTest();
+        List<Database.SaveResult> saveResults = Database.update(accounts, false);
+        ERR_Handler_API.processDMLErrors(saveResults, accounts, ERR_CONTEXT, true);
+        Test.stopTest();
+
+        // query Error__c record result for asserts
+        List<Error__c> errors = new List<Error__c>([SELECT Id, Related_Record_ID__c, Context_Type__c FROM Error__c]);
```
force2b, line 67: This can be replaced by the existing data factory logic for errors:

Add the following to the top of the class:
`private static final TEST_SObjectGateway.ErrorGateway errorGateway = new TEST_SObjectGateway.ErrorGateway();`

And then this line can be
```suggestion
        List<Error__c> errors = errorGateway.getRecords();
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+
+        // assert error is created
+        System.assertEquals(1, errors.size());
```
force2b, line 70:
```suggestion
        System.assertEquals(1, errors.size(), 'There should be only a single Error record created');
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+        // assert error fields match
+        System.assertEquals(acct.Id, errors[0].Related_Record_ID__c);
+        System.assertEquals(ERR_CONTEXT, errors[0].Context_Type__c);
+        // assert page error is displayed
+        List<ApexPages.Message> pageMessages = ApexPages.getMessages();
+        System.assertEquals(1, pageMessages.size(), 'Error message should be displayed on the page: ' + pageMessages);
```
force2b, line 76: This logic can be replace by a call to the factory logic:
```suggestion
        // assert page error is displayed
        UTIL_UnitTestData_TEST.assertPageHasMessage(errors[0].Full_Message__c);
```

src/classes/ERR_Handler_API_TEST.cls
```diff
@@ -0,0 +1,106 @@
...
+
+    }
+
+    /**
+      * @description Verifies DML error is created and the error is displayed on a page
+      */
+    @IsTest
+    private static void shouldCreateErrorFromDMLButDoNotDisplayOnPage() {
```
force2b, line 84: Same changes needed in this method as the previous method.

force2b (changes requested):
Thanks for creating the test class @mldyang. I requested some changes to implement existing factory methods rather than recreating everything in this class. Would you mind adding an additional test method for `shouldNotCreateErrorFromDmlIfNoErrorsGenerated()`?

mldyang:
Thank you @force2b. I did not realize there are existing code I could utilize here. Thanks for all your feedback!

force2b:
@mldyang Did you see that there are [build errors](https://metaci.sfdc.sh/builds/497243) due to a metadata failure?

force2b (approved):
Nice work @mldyang

lparrott (approved):

mldyang:
No longer needed as the goal was accomplished by utilizing an existing callable class (`CallableDispatchService_ERR`). Full details in WI [here](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008iywTIAQ/view)

PR #6487, opened 2021-06-14
Allow user to determine date range for matched Opportunities for Yearly RDs

dfuller114:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


dfuller114:
This work is for the `232.1` release and will be held until 7/14 when the 232 merge freeze is lifted

src/classes/RD2_OpportunityMatcher.cls
```diff
@@ -216,10 +216,12 @@ public inherited sharing class RD2_OpportunityMatcher {
             matched.attachInstallment(installment);
             return;
         }
-        
+
         //if the Installment period is not monthly, do not attempt to match by +/- match days range
         //and add the installment without matching Opp
-        if (installment.installmentPeriod != RD2_Constants.INSTALLMENT_PERIOD_MONTHLY) {
+        if (installment.installmentPeriod != RD2_Constants.INSTALLMENT_PERIOD_MONTHLY &&
+            installment.installmentPeriod != RD2_Constants.INSTALLMENT_PERIOD_YEARLY) {
```
bdvorachek, line 223: Can extract this as a method on the Installment class.

dfuller114: @bdvorachek, thinking through this one a little more, I think moving this to the Installment class might be a little tricky. The reason is that this boolean's result is determining matching behavior that belongs to this Matcher class. I was thinking this boolean condition could determine a certain type of Installment so that the Installment class could know what type it is. However, that would only work if that type is useful or valid outside of needing it for matching behavior. That being said, I'm not sure that it would belong in the Installment class unless there's a way it could be decoupled from matching behavior

bdvorachek: Would this be better as a method on this class that accepts an installment as an argument then? I think there's still value in giving this conditional a name.

dfuller114: Just pushed up a change to extract that conditional logic into a method

npsp-reedestockton (approved):
Looks good @dfuller114!  Nice work!

bdvorachek (approved):

PR #6489, opened 2021-06-16
Gift Entry Design: Establish GiftBatchView, Decouple from old BGE

kenlewis-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/GiftBatch.cls
+++ b/src/classes/GiftBatch.cls
@@ -142,12 +142,22 @@ public inherited sharing class GiftBatch implements IGiftBatch {
 
     public Money total() {
         //todo: use giftBatch Totals (and place in Selector layer)
-        String model = BGE_DataImportBatchEntry_CTRL.getDataImportModel(batchId);
-        BGE_DataImportBatchEntry_CTRL.DataImportModel typedModel =
-                (BGE_DataImportBatchEntry_CTRL.DataImportModel)
-                        JSON.deserialize(model,
-                BGE_DataImportBatchEntry_CTRL.DataImportModel.class);
-        return new Money(typedModel.totalRowAmount, 'US');
+        return new Money(dataImportModel().totalRowAmount, 'US');
     }
 
+    public Integer totalCount() {
+        //todo: use giftBatch Totals (and place in Selector layer)
+        return dataImportModel().totalCountOfRows;
+    }
+
+    private BGE_DataImportBatchEntry_CTRL.DataImportModel typedModel;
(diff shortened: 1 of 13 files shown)


src/classes/GiftBatch.cls
```diff
@@ -138,4 +139,15 @@ public inherited sharing class GiftBatch {
...
+
+    public Money total() {
+        //todo: use giftBatch Totals (and place in Selector layer)
+        String model = BGE_DataImportBatchEntry_CTRL.getDataImportModel(batchId);
+        BGE_DataImportBatchEntry_CTRL.DataImportModel typedModel =
+                (BGE_DataImportBatchEntry_CTRL.DataImportModel)
+                        JSON.deserialize(model,
+                BGE_DataImportBatchEntry_CTRL.DataImportModel.class);
+        return new Money(typedModel.totalRowAmount, 'US');
+    }
```
kenlewis-sfdo, line 151: This isn't pretty, but it is a way to begin to install the ideal model while still using the same underlying source code while putting characterization tests in place.  Once the characterization tests are in place, then this method can be safely refactored to simplify and/or use another mechanism to retrieve the total, like GIftBatchTotals.

src/classes/GiftBatch.cls
```diff
@@ -138,4 +139,25 @@ public inherited sharing class GiftBatch {
...
+        return new Money(dataImportModel().totalRowAmount, 'US');
+    }
+
+    public Integer totalCount() {
+        //todo: use giftBatch Totals (and place in Selector layer)
+        return dataImportModel().totalCountOfRows;
+    }
+
+    private BGE_DataImportBatchEntry_CTRL.DataImportModel typedModel;
+    private BGE_DataImportBatchEntry_CTRL.DataImportModel dataImportModel() {
```
kenlewis-sfdo, line 154: The refactoring work will eventually remove this dependency on the old BGE controller.

src/classes/IGiftBatch.cls
```diff
@@ -0,0 +1,13 @@
+public interface IGiftBatch {
```
voduyemi, line 1: At what point do we expect to marry the GiftBatchTotals class with this interface since this interface should also include and define a `GiftBatchTotals` which makes the `Money` type internal to it?

voduyemi: Looks like you answered my question in the comment at the top!

PR #6509, opened 2021-06-24
[RD2] Send productMetadata updates non-destructively

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000NXxiYAG/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/UTIL_Http_TEST.cls
```diff
@@ -307,13 +306,4 @@ public with sharing class UTIL_Http_TEST {
         Test.setMock(HttpCalloutMock.class, callout);
     }
 
-    public static void mockChargeElevateBatch(String responseBody){
-        CalloutMock callout = new CalloutMock()
-                .withStatusCode(UTIL_Http.STATUS_CODE_OK)
```
lparrott, line 312: I like the move to a single method, were these Status values just never being checked?

src/classes/PS_CommitmentRequest.cls
```diff
@@ -239,20 +265,40 @@ public inherited sharing class PS_CommitmentRequest {
...
+
+        Map<String, Object> commitment = (Map<String, Object>) responseBody.get('commitment');
+        Map<String, Object> productMetadata = (Map<String, Object>) commitment.get('productMetadata');
+
+        return productMetadata;
+    }
+
+    private Map<String, Object> getElevateCommitment(String commitmentId) {
+        HttpRequest commitmentGetRequest = buildGetRequest(commitmentId);
+        UTIL_Http.Response response = requestService.sendRequest(commitmentGetRequest);
```
lparrott, line 296: Not sure where to add it, but I'd feel better if this was in a try/catch 😬

lparrott (commented):
Looking great, you even got in some good cleanup and refactoring 👍 Just had one comment about a potential try/catch block.

src/classes/PS_CommitmentRequest.cls
bdvorachek: The same try/catch that's in the update call would catch any errors from this callout, which I think is what we'd want. If the GET request times out, we'd want the same handling as a PUT/POST timeout. 
https://github.com/SalesforceFoundation/NPSP/blob/41a6bcd22c75e9c720cbfb4cf6b6d04ca1c1792b/src/classes/RD2_EntryFormController.cls#L278

src/classes/UTIL_Http_TEST.cls
bdvorachek: These status values are the default values in the constructor.

src/classes/PS_CommitmentRequest.cls
lparrott: Oh, excellent, I should've looked further upstream 👍

lparrott (approved):

npsp-reedestockton (approved):
This is excellent, @bdvorachek!!!  Very clean code!  It should be very easy to extend this as needed.

dfuller114 (approved):
looks good! thanks Ben!

Issue #6221, opened 2021-02-09
Gift Entry: Add New Payment or Update This Opportunity Cause Script-Thrown Exception If NPSP Data Import Batch field is included in Opportunity or Payment Field Mappings


jesslopez-sf:
When a customer reviews potential matching donations for a Contact or Account using the Review Donations Modal in Gift Entry and they click either the _Add New Payment_ or _Update this Opportunity_ link,  the transaction fails and the customer sees a "Script-Thrown Exception" error. 

<img width="821" alt="Screen Shot 2021-02-09 at 5 49 30 PM" src="https://user-images.githubusercontent.com/29731456/107440239-4e6f7a00-6b01-11eb-9a2b-1ed7b6fb00b7.png">

### Root Cause
We call the selectedDonationOrPaymentRecord() method in geFormRenderer.js. We then call setSelectedDonationInFormState() which next calls resetDonationAndPaymentImportedFields(), which wipes the fields associated with the Opportunity/Payment field mappings. And then we update the form state. If the NPSP Data Import Batch (NPSP_Data_Import_Batch__c ) field is included within the Opportunity or Payment field mappings, we clear that field. This causes issues with SOQL queries later on. Specifically, the Exception is ultimately occurring in the BGE_DataImportBatchEntry_CTRL class's runSingleDryRun() method. When we call getDataImportRecordsByIds() on line 293, 0 records are returned because the Data Import record is no longer associated with the batch. Because 0 records are returned in getDataImportRecordsByIds(), this leads to a "List Index Out of Bounds: 0" Error which triggers the catch(){} and throws the AuraHandledException (ultimately surfaced to customer as “Script-thrown Exception”.

### Steps to Repeat
1. Create Lookup Relationship field to NPSP Data Import Batch object on Opportunity object.
2. In NPSP Settings, enable Advanced Mapping and Gift Entry
3. In Advanced Mapping settings, go to the Opportunity Object Mapping and add a Field Mapping for NPSP Data Import Batch field on NPSP Data Import object to the new Opportunity Lookup Field for NPSP Data Import Batch. 
4. Create a Contact
5. Create an Opportunity for that Contact and set the Stage to Prospecting (or anything other than Closed Won or Closed Lost). 
6. Go to the Gift Entry Tab. 
7. Clone the default Template and make sure you add the NPSP Data Import Batch field to the form (this isn’t required, but it will show you that it’s initially set and that we clear it…).
8. Create a new Batch using the new Template. 
[...]

PR #6517, opened 2021-06-29
Gift Entry - Script Error on Gift Save Fix

sokosaurus:
# Critical Changes

Issue

When a customer reviews potential matching donations for a Contact or Account using the Review Donations Modal in Gift Entry and they click either the Add New Payment or Update this Opportunity link, the transaction fails and the customer sees a "Script-Thrown Exception" error.

Github Issue: #6221

Resolution

There are certain fields within form state that we should only expect to be set once, likely when a gift (new or existing) is initially loaded in geFormRenderer. In fact, we can get into trouble when these fields are changed during the lifecycle of a gift edit.

In our case here (based on the details provided above), the lookup field that lives on the Data Import record and looks up to the Data Import Batch record gets cleared out after a user runs through the “Review Donations” process. We can address this by introducing the notion of immutability to Form State.

Certain fields should only be set ONCE in form state (when creating or editing a gift) and we can build out functionality to block any subsequent update to that field. At the same time, we’ll make sure that this field value can be reset when form state is initialized/re-initialized.

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/lwc/geFormRenderer/geFormRenderer.js
+++ b/src/lwc/geFormRenderer/geFormRenderer.js
@@ -135,6 +135,10 @@ const DONATION_DONOR_TYPE_ENUM = Object.freeze({
     CONTACT1: 'Contact1'
 });
 
+const FORM_STATE_IMMUTABLE_FIELDS_BY_APINAME = [
+    NPSP_DATA_IMPORT_BATCH_FIELD.fieldApiName
+];
+
 const ACH_CONSENT_MESSAGE = 'true';
 const EXPANDABLE_SECTION_CONTAINER = 'expandableSectionContainer';
 
@@ -1693,6 +1697,8 @@ export default class GeFormRenderer extends LightningElement{
      * @param fields An object with key-value pairs.
      */
     updateFormState(fields) {
+        fields = this.removeFieldsNotUpdatableInFormState(fields);
+
         Object.assign(this.formState, fields);
         if (fields.hasOwnProperty(apiNameFor(DONATION_RECORD_TYPE_NAME))) {
             this.updateFormStateForDonationRecordType(fields);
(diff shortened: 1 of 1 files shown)


sokosaurus:
Replaces #6515

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -135,6 +135,10 @@ const DONATION_DONOR_TYPE_ENUM = Object.freeze({
     CONTACT1: 'Contact1'
 });
 
+const FORM_STATE_IMMUTABLE_FIELDS_BY_APINAME = [
```
psadie-sfdo, line 138:
```suggestion
const FORM_STATE_IMMUTABLE_FIELDS_BY_API_NAME = [
```

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1706,6 +1712,20 @@ export default class GeFormRenderer extends LightningElement{
...
+            if (this.isFormStateFieldNotUpdatable(fieldsToUpdate, immutableField)) {
+                delete fieldsToUpdate[immutableField];
+            }
+        });
+
+        return fieldsToUpdate;
+    }
+
+    isFormStateFieldNotUpdatable(fields, field) {
+        return fields.hasOwnProperty(field) && this.formState[field];
```
psadie-sfdo, line 1726: Just to confirm, the second condition `this.formState[field]` means that we _would_ allow a field, defined as immutable in the `FORM_STATE_IMMUTABLE_FIELDS_BY_API_NAME` constant, to be added to formState _if_ the field hasn't been added to formState prior to this form state update attempt?

`isFormStateFieldNotUpdatable` would return false if the field in question is not in formState right and so would not be removed from `fieldsToUpdate`?

psadie-sfdo (commented):
Couple of questions!

src/lwc/geFormRenderer/geFormRenderer.js
sokosaurus: Your understanding here is 100% correct @psadie-sfdo. We want to make sure that the field value can be set to a non-null value once in form state but block any subsequent update -- until form state is re-initialized.

psadie-sfdo: Ahhh, gotcha gotcha. Need to allow that first one when the form initializes. Makes sense and thanks! Deploying the changes to a scratch org now!

src/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -135,6 +135,10 @@ const DONATION_DONOR_TYPE_ENUM = Object.freeze({
     CONTACT1: 'Contact1'
 });
 
+const FORM_STATE_IMMUTABLE_FIELDS_BY_API_NAME = [
```
ptewson-sfdo, line 138: This is minor, but the "by api name" here implies a map to me. Maybe FORM_STATE_IMMUTABLE_FIELDS_API_NAMES?

sokosaurus: Sounds great to me. Thanks for the feedback here. Made the change!

psadie-sfdo (approved):
Looks good! Thanks!!

ptewson-sfdo (approved):
Looks good!

PR #6529, opened 2021-07-07
[wip] Github Actions experimental branch

jstvz:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -17,8 +17,7 @@ jobs:
           if [ $GITHUB_BASE_REF ]; then
             # Pull Request
           git fetch origin $GITHUB_BASE_REF:$GITHUB_BASE_REF --depth=1
-            echo "RUN: git diff-tree $GITHUB_BASE_REF..$GITHUB_HEAD_REF --patch-with-raw"
-            export DIFF=$( git diff-tree $GITHUB_BASE_REF..$GITHUB_HEAD_REF --patch-with-raw | grep 'SfdoLogUtils.log' )
+            echo "RUN: git diff-tree $GITHUB_BASE_REF..HEAD --patch-with-raw"
             echo "Diff between origin/$GITHUB_BASE_REF and $GITHUB_SHA"
           fi
           echo "$DIFF"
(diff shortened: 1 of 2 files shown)


force2b:
@jstvz  I don't think the action ran because there were no classes committed. Feel free to merge this into my branch and work there; or I can modify a class file in this branch. Either is good for me.

jstvz:
@force2b I've tried a couple of approaches here, none of which appear to work correctly. The crux of the problem is that, by default, we've got a... strange git context, where not every ref is present. The last few commits here attempted to work around that by being more explicit about which refs to compare, but unfortunately the diffs don't match expectations. 

There may be a better approach here that makes use of the built-in diff url ([example](https://patch-diff.githubusercontent.com/raw/SalesforceFoundation/NPSP/pull/6529.diff)) to check for the strings we care about. Alternatively, a full-fledged local action offers more control, but at the risk of being overkill for this use case.

Thanks for the nerd snipe, btw.

force2b:
Thanks @jstvz. So, the issue is the `git fetch` statement vs. the actual diff?

jstvz:
> Thanks @jstvz. So, the issue is the `git fetch` statement vs. the actual diff?

Yeah, I think that's the case. For example, on a fresh clone:
```console
$ git branch --all | rg 234
  remotes/origin/feature/234
  remotes/origin/feature/234__gift-entry-soft-credits
  remotes/origin/feature/234__perftest-insert-contacts
  remotes/origin/feature/234__rdChangeHistoryMetadata
  remotes/origin/feature/234__undeliverableAddressToAccountAndContacts
$ git diff-tree master..feature/234 --patch-with-raw
fatal: ambiguous argument 'master..feature/234': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

The solution, I think, is that we need to pass a `refspec` to `git fetch` to resolve the reference:
```console
$ git fetch origin feature/234:feature/234
From github.com:SalesforceFoundation/NPSP
 * [new branch]          feature/234 -> feature/234
$ git diff-tree master..feature/234 --patch-with-raw
[...no diff...]
```

I can't get the workflow to fire to confirm, but that's my best guess.

force2b:
@jstvz I think that worked! https://github.com/SalesforceFoundation/NPSP/pull/6292/checks?check_run_id=3032462005

PR #6530, opened 2021-07-07
Project Hummingbird: Move Account Creation Logic into Domains

kenlewis-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000Q6jMYAS/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/src/classes/Contacts.cls
+++ b/src/classes/Contacts.cls
@@ -810,6 +810,74 @@ public inherited sharing class Contacts extends fflib_SObjects implements IConta
         return oldMap.get(con.Id);
     }
 
+    public List<Id> getHouseholdIdsToRename(Contacts contacts) {
+        List<Id> householdIdsToRename = new List<Id>();
+        for (Contact contactRecord : (List<Contact>) getRecords()) {
+            Contact oldContact = contacts.oldVersionOf(contactRecord);
+            if (needsHHAccountRename(contactRecord, oldContact)) {
+                householdIdsToRename.add(contactRecord.AccountId);
+            }
+        }
+        return householdIdsToRename;
+    }
+
+    /*******************************************************************************************************
+    * @description returns whether changes in the contact require a rename of a HH Account
+    * @param contactRecord the new version of the Contact
+    * @param oldContact the old version of the Contact
+    * @return Boolean
+    */
+    private Boolean needsHHAccountRename(Contact contactRecord, Contact oldContact) {
+        if (contactRecord.npe01__Organization_Type__c != CAO_Constants.HH_ACCOUNT_TYPE) {
+            return false;
+        }
+
+        if (contactRecord != null && oldContact == null) {
+            return true;
+        }
+
+        contactRecord.FirstName = contactRecord.FirstName == null ? '' : contactRecord.FirstName;
+        contactRecord.LastName = contactRecord.LastName == null ? '' : contactRecord.LastName;
+
+        // check all household naming fields for changes
+        for(String fieldName : new HouseholdNamingService().getHouseholdNamingContactFields()) {
+            String currentName;
+            String oldName;
+
+            //separate cross object references, i.e. account.name
+            List<String> splitFields = (fieldName.split('\\.',0));
+            if (splitFields.size() > 1) {
+                SObject sObj = contactRecord.getSObject(splitFields[0]);
+                if (sObj != null) {
+                    currentName = sObj.Id;
+                }
+                sObj = oldContact.getSObject(splitFields[0]);
+                if (sObj != null) {
+                    oldName = sObj.Id;
+                }
+            } else {
+                currentName = String.valueOf(contactRecord.get(fieldName));
+                oldName = String.valueOf(oldContact.get(fieldName));
+            }
+
+            // make sure to use case sensitive comparison
+            if (currentName == null) {
+                currentName = '';
+            }
+            if (oldName == null) {
+                oldName = '';
+            }
+            if (!currentName.equals(oldName)) {
+                return true;
+            }
+        }
+        if (contactRecord.npo02__Household_Naming_Order__c != oldContact.npo02__Household_Naming_Order__c ||
+            contactRecord.npo02__Naming_Exclusions__c != oldContact.npo02__Naming_Exclusions__c) {
+            return true;
+        }
+        return false;
+    }
+
     /*******************************************************************************************************
      * @description kicks off Opportunity Rollups after contacts have been merged, but only if NOT using
      * the Household Account model
--- a/src/classes/Contacts.cls
+++ b/src/classes/Contacts.cls
@@ -1054,7 +1054,7 @@ public inherited sharing class Contacts extends fflib_SObjects implements IConta
      ********************************************************************************************************/
     @TestVisible
     private static void rollupContactsAndHouseholdsAfterMerge(Set<Id> winningContacts) {
-        if (contactAndOrgSettings.isHouseholdAccountModel()) {
+        if (contactAndOrgSettings.isHouseholdAccountModelEnabled()) {
             return;
         }
 
--- a/src/classes/Contacts.cls
+++ b/src/classes/Contacts.cls
@@ -138,10 +138,6 @@ public inherited sharing class Contacts extends fflib_SObjects implements IConta
         }
     }
 
-    public void onBeforeInsert_ACCT_IndividualAccounts_TDTM() {
-        new HouseholdsApp().handleContactsBeforeInsert(this);
-    }
-
     public void onBeforeUpdate() {
         syncNamingExclusionsBeforeUpdate();
     }
@@ -190,10 +186,6 @@ public inherited sharing class Contacts extends fflib_SObjects implements IConta
 
         // update the Latest date fields on any address overrides
         Addresses.updateAddrIsOverride(isOverrideByAddressId);
-    }
-
-    public void onBeforeUpdate_ACCT_IndividualAccounts_TDTM() {
-        Households.handleContactsBeforeUpdate(this);
     }
 
     public void onAfterUpdate() {
--- a/src/classes/Contacts.cls
+++ b/src/classes/Contacts.cls
@@ -512,6 +512,55 @@ public inherited sharing class Contacts extends fflib_SObjects implements IConta
         return new Contacts(contactsInHouseholdAccounts);
     }
 
+    public List<Contact> getOneToOneContactsToUpdate() {
+        List<Contact> oneToOneContactsToUpdate = new List<Contact>();
+        for (Contact contactRecord : (List<Contact>) getRecords()) {
+            Contact oldContact = oldMap.get(contactRecord.Id);
+
+            if (needsOne2OneAccountUpdate(contactRecord, oldContact)) {
+                oneToOneContactsToUpdate.add(contactRecord);
+            }
+        }
+        return oneToOneContactsToUpdate;
+    }
+
+    /*******************************************************************************************************
+    * @description returns whether changes in the contact require an update to the 1:1 account
+    * @param contactRecord the new version of the Contact
+    * @param oldContact the old version of the Contact
+    * @return Boolean
+    */
+    private Boolean needsOne2OneAccountUpdate(Contact contactRecord, Contact oldContact) {
+        if (contactRecord.npe01__Organization_Type__c != CAO_Constants.ONE_TO_ONE_ORGANIZATION_TYPE) {
+            return false;
+        }
+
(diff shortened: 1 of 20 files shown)


src/classes/Contacts.cls
```diff
@@ -156,15 +192,50 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+
     public void onAfterUpdate() {
         isUpdate = true;
         createAddressesForContacts();
     }
 
+    public void onAfterUpdate_ACCT_IndividualAccounts_TDTM(
+            TDTM_Runnable.DmlWrapper dmlWrapper) {
+        /**
+         * @description In the after trigger context, additional fields from the account
```
lparrott, line 207: Maybe it's just me, but having descriptions inside methods feels weird 🤷‍♂️ could make them block comments, or move them above the method name?

src/classes/Contacts.cls
```diff
@@ -491,6 +625,135 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+        if(contactRecord != null && oldContact == null) {
+            return true;
+        }
+
+        String newFirst = contactRecord.FirstName == null ? '' : contactRecord.FirstName;
+        String oldFirst = oldContact.FirstName == null ? '' : oldContact.FirstName;
+        String newLast = contactRecord.LastName == null ? '' : contactRecord.LastName;
+        String oldLast = oldContact.LastName == null ? '' : oldContact.LastName;
+
+        return (contactRecord.AccountId != oldContact.AccountId ||
```
lparrott, line 740: Out of scope if we're just trying to reorganize, but this could just be a loop over a list of field names.

src/classes/HouseholdAccounts.cls
```diff
@@ -98,4 +157,362 @@ public inherited sharing class HouseholdAccounts {
...
+    * @description sets the Primary Account for each Account in the list of Accounts
+    * @param accountIds the list of Account (Id's) that need updating.
+    */
+    private static void resetPrimaryContactForAccount(List<Id>
+            accountIds, TDTM_Runnable.DmlWrapper dmlWrapper) {
+
+        // for each account, we need a set of their contacts.
+        Map<Id, Set<Id>> contactIdsByAccountId = new Map<Id, Set<Id>>();
+
+        for (Contact contactRecord : [
```
lparrott, line 226: Any chance we already have these Contacts queried? Feels risky to run a query (2 actually...) each time this method is called.

src/classes/HouseholdAccounts.cls
```diff
@@ -98,4 +157,362 @@ public inherited sharing class HouseholdAccounts {
...
+            if (contactIds == null) {
+                contactIds = new Set<Id>();
+                contactIdsByAccountId.put(contactRecord.AccountId, contactIds);
+            }
+            contactIds.add(contactRecord.Id);
+        }
+
+        // now go thru each account and update their primary contact if the current one isn't valid
+        List<Account> accounts = new List<Account>();
+        for (Account accountRecord : [SELECT Id, npe01__One2OneContact__c FROM Account WHERE Id IN :accountIds]) {
```
lparrott, line 241: I feel like these records should already be available?

src/classes/HouseholdAccounts.cls
```diff
@@ -98,4 +157,362 @@ public inherited sharing class HouseholdAccounts {
...
+        List<Contact> contacts = new List<Contact>();
+
+        for (Contact contactRecord : contactsWithoutAccounts) {
+            //track which contacts are being inserted vs. updated
+            if (contactRecord.Id != null) {
+                contactIds.add(contactRecord.Id);
+            }
+        }
+        // get all the Accounts that are connected to the existing Contacts
+        for (Account accountRecord : [
```
lparrott, line 296: More queries, maybe this is a larger problem.

src/classes/Households.cls
```diff
@@ -187,4 +458,53 @@ public inherited sharing class Households extends fflib_SObjects {
...
+
+    private static List<Contact> contactsWithoutAccounts(
+            Contacts contactsInstance) {
+        List<Contact> contactsWithoutAccounts = new List<Contact>();
+
+        if (CAO_Constants.isHHAccountModel()) {
+            contactsWithoutAccounts.addAll(contactsInstance.contactsWithoutAccounts());
+        }
+
+        if (CAO_Constants.isOneToOne()) {
```
lparrott, line 502: Super small thing, could this be an `else if`, since you can only have one Account model active?

lparrott (commented):
Looking great, I just have a few comments after a quick run through.

src/classes/HouseholdAccounts.cls
bdvorachek: I think we can combine these queries, my only reservation was "What happens when one of the Accounts has no Contacts?". When looping through the second query, contactIds would be null for any Account that had zero contacts, and that looks to be OK. 

We'll still need a second loop here, but we wouldn't need to query Accounts. We can gather all of the AccountIds in the first loop w/ their `npe01__One2OneContact__c`, and add those to the Accounts to update if the Contact Id in `npe01__One2OneContact__c` does not match with any of the Contacts that are currently part of the given Household.

src/classes/Households.cls
```diff
@@ -130,18 +141,278 @@ public inherited sharing class Households extends fflib_SObjects {
...
+     */
+    //todo: fix
+    public static Boolean needsOneToOneAccount(Contact contactRecord, Account accountRecord) {
+        return contactRecord.npe01__Private__c != true
+                && (contactRecord.AccountId == null
+                || (accountRecord != null
+                && accountRecord.npe01__SYSTEM_AccountType__c == CAO_Constants.ONE_TO_ONE_ORGANIZATION_TYPE));
+    }
+
+    public static List<Contact> contactsWithoutAccounts(Contacts contactsInstance, Map<Id, Contact> oldContactById) {
```
bdvorachek, line 202:
```suggestion
    public static List<Contact> contactsWithoutAccounts(Contacts contactsInstance) {
```

`oldContactById` is equivalent to `contactInstance.oldMap` here I think, so we could drop the second argument.

src/classes/Households.cls
```diff
@@ -130,18 +141,278 @@ public inherited sharing class Households extends fflib_SObjects {
...
+        List<Contact> contactsWithoutAccounts2 = new List<Contact>();
+        Map<Id, Account> accountsByIds = contactsInstance.accountById;
+        for (Contact contactRecord : (List<Contact>) contactsInstance.getRecords()) {
+            Account account = accountsByIds.get(contactRecord.AccountId);
+            if (needsOneToOneAccount(contactRecord,
+                    account)) {
+
+                Contact oldContact = oldContactById.get(contactRecord.Id);
```
bdvorachek, line 210:
```suggestion
                Contact oldContact = contactsInstance.oldMap.get(contactRecord.Id);
```

src/classes/Contacts.cls
kenlewis-sfdo: Agreed.  Do you think this is a comment that is useful enough to keep? Or should it just be removed?  Not sure that it is worth keeping.

src/classes/HouseholdAccounts.cls
kenlewis-sfdo: This method was fully lifted from ACCT_IndividualAccounts_TDTM, and HouseholdAccounts is evolving from its original incarnation as a way to lazy load a Households' members.  

Now that we have these methods in the same domain, we should absolutely be able to take advantage and query what is needed up front at instantiation, so this type of multiple query pattern can be avoided.

I'll take a look today and see if this seems to make sense to do now.

src/classes/Contacts.cls
```diff
@@ -491,6 +616,135 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+                householdIdsToRename.add(contactRecord.AccountId);
+            }
+        }
+        return householdIdsToRename;
+    }
+
+    public void resetDeprecatedFields() {
+        for (Contact contactRecord : (List<Contact>) getRecords()) {
+            // always clear this deprecated field to avoid confusing users.
+            contactRecord.npe01__SystemAccountProcessor__c = null;
```
sokosaurus, line 649: At some point we may want to consider moving this into a class-level constant variable (or somewhere else) so that we can easily track/maintain a potentially growing list of deprecated fields that need to be nulled out here.

src/classes/Contacts.cls
```diff
@@ -491,6 +616,135 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+        }
+    }
+
+    public void keepPrivateContactsPrivate() {
+        for (Contact contactRecord : (List<Contact>) getRecords()) {
+            // for contacts marked private after they have been created, break the relationship from their account (any model)
+            if (contactRecord.AccountId != null && contactRecord.npe01__Private__c == true) {
```
sokosaurus, line 656: Just a small cleanup item (for the future) - we could consider dropping `== true` from this conditional.

src/classes/Contacts.cls
```diff
@@ -174,6 +245,60 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+
+    /**
+     * @description Determines whether or not Contacts belong to the BucketAccount, when
+     * Contacts are being inserted.
+     */
+    public Contacts bucketContacts_insertContext() {
+        return new Contacts(contactsWithoutAccounts());
+    }
+
+    public Contacts contactsAndOneToOneContacts_insertContext() {
```
sokosaurus, line 278: In our variable and method names that are related to the 1:1 model, we use the naming convention of either `OneToOne` or `One2One`. For searchability (and perhaps other benefits), do you think it makes sense to align on a singular naming convention here?

src/classes/HouseholdAccounts.cls
```diff
@@ -98,4 +157,362 @@ public inherited sharing class HouseholdAccounts {
...
+    */
+    public static void rollupAccounts(List<Id> accountIds) {
+        new RLLP_OppRollup().rollupAccounts(new Set<Id>(accountIds));
+    }
+
+    /*******************************************************************************************************
+    * @description Inserts a new 1:1 or HH Account for an Individual Contact
+    * @param contactsWithoutAccounts the list of Contacts that need updating.
+    */
+    public static void createOneToOneAndHouseholdAccountsFor(List<Contact>
```
sokosaurus, line 279: Probably out of scope for this initial effort, but flagging this as a long running method that we should consider refactoring.

sokosaurus (changes requested):
Took a pass at this PR. Feel free to push back on anything that doesn't feel actionable in this iteration of the work. Looking really good @kenlewis-sfdo!

src/classes/Contacts.cls
kenlewis-sfdo: Yes, good call.  Let's do that when we have another field to add here.

src/classes/Contacts.cls
kenlewis-sfdo: Yep.  Should prob standardize on OneToOne, oneToOne, One-to-one, etc.

src/classes/HouseholdAccounts.cls
kenlewis-sfdo: 👍🏽

src/classes/HouseholdAccounts.cls
kenlewis-sfdo: I really like these ideas about combining queries in this class and getting it instantiated with the right set of records.  But going to hold off on that for a follow up next WI: https://salesforce.quip.com/n2vzAQOiRbT7#UMTACAR9Raq

src/classes/HouseholdAccounts.cls
kenlewis-sfdo: Added to design doc for further discussion!: https://salesforce.quip.com/n2vzAQOiRbT7#UMTACAjM6Uc

src/classes/HouseholdAccounts.cls
kenlewis-sfdo: I like the plan, Ben, adding it to the list of items to follow up on on this class: https://salesforce.quip.com/n2vzAQOiRbT7#UMTACAd3ibZ

src/classes/Contacts.cls
lparrott: I agree, just delete them (I see 3 areas that are like this)

lparrott (approved):
Looking good, approving with the assumption that we'll tackle the feedback in a future PR 👍

sokosaurus (approved):
Following @lparrott and approving. Really great work here!

PR #6537, opened 2021-07-08
Gift Entry: Soft Credits

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/classes/Gift.cls
```diff
@@ -68,14 +68,14 @@ public with sharing class Gift {
...
 
-    public DataImport__c dataImportRecord() {
-        return this.dataImport;
-    }
-
     public SoftCredits softCredits() {
         return this.softCredits;
     }
 
+    public Map<String, Object> fieldsWithValues() {
```
kenlewis-sfdo, line 75: What do you think about naming this 'fields' also for consistency?

psadie-sfdo: Agree!

src/classes/GE_GiftEntryController.cls
```diff
@@ -40,6 +40,7 @@ public with sharing class GE_GiftEntryController {
     public static final String PURCHASE_CALL_TIMEOUT_MESSAGE = System.Label.geErrorRequestTimedout;
     private static final String BOOLEAN_DATA_TYPE = 'BOOLEAN';
     private static GiftBatchService giftBatchService = new GiftBatchService();
+    private static GiftService giftService = new GiftService();
```
kenlewis-sfdo, line 43: This is the pattern we are currently following to allow for injection:

src/classes/GiftSelector.cls
```diff
@@ -53,5 +68,73 @@ public with sharing class GiftSelector {
...
+            .withFrom(DataImport__c.SObjectType)
+            .withSelectFields(queryFields(fieldsToQuery))
+            .withWhere('Id =: giftId')
+            .withSecurityEnforced()
+            .build();
+
+        return Database.query(queryString);
+    }
+
+    private List<String> elevateRelatedFieldApiNames() {
```
kenlewis-sfdo, line 83: 'elevateFields' ?

src/classes/GE_GiftEntryController.cls
```diff
@@ -52,6 +69,24 @@ public with sharing class GE_GiftEntryController {
...
+    public static GiftBatchView getGiftBatchView(Id giftBatchId){
+        try {
+            return giftBatchService.viewModelFrom(giftBatchId);
+        } catch (AuraHandledException e) {
+            throw new AuraHandledException(e.getMessage());
+        }
+    }
+
+    @AuraEnabled
+    public static GiftView getGiftView(Id giftId){
```
ptewson-sfdo, line 82: It feels a bit strange to call the param "giftId" when the Id, in truth, belongs to a DataImport? But not sure if that feeling stands up to further questioning.

src/classes/GE_GiftEntryController.cls
```diff
@@ -39,7 +39,24 @@ public with sharing class GE_GiftEntryController {
 
     public static final String PURCHASE_CALL_TIMEOUT_MESSAGE = System.Label.geErrorRequestTimedout;
     private static final String BOOLEAN_DATA_TYPE = 'BOOLEAN';
-    private static GiftBatchService giftBatchService = new GiftBatchService();
+    private static GiftBatchService giftBatchService {
```
kenlewis-sfdo, line 42:
```suggestion
    @TestVisible
    private static GiftBatchService giftBatchService {
```

src/classes/GE_GiftEntryController.cls
```diff
@@ -39,7 +39,24 @@ public with sharing class GE_GiftEntryController {
...
+        get {
+            if (giftBatchService == null) {
+                giftBatchService = new GiftBatchService();
+            }
+            return giftBatchService;
+        }
+        set;
+    }
+    private static GiftService giftService {
```
kenlewis-sfdo, line 51:
```suggestion
    @TestVisible
    private static GiftService giftService {
```

src/classes/GE_GiftEntryController.cls
kenlewis-sfdo: We could use a value object called "GiftBatchId"

src/classes/Gift.cls
```diff
@@ -47,10 +48,16 @@ public with sharing class Gift {
         }
         set;
     }
-    
+
+    public Gift(Id giftId) {
+        this.dataImport = giftSelector.getGiftById(giftId, new GiftEntryFormTemplate());
```
kenlewis-sfdo, line 53: It seems a Gift Entry Gift (or maybe just a Batch) is dependent on a Template, since it determines its View state.  What would you think about renaming  `GiftEntryFormTemplate` to `GiftTemplate` ?

Our language could be "The Form uses the Batch's Gift Template (or the default Gift Template in SGE) to display Gifts on the Form."

It also makes more sense for the GiftSelector to get its fields from a GiftTemplate than a "Form"Template, which is tied to a specific UI.

So, I think the Batch should know about the Template (to configure the Selector), but then the Batch should configure the Selector and pass that into the constructor of the Gifts instance.

src/classes/GiftBatch.cls
```diff
@@ -138,4 +119,37 @@ public inherited sharing class GiftBatch {
     public void updateRecords() {
         gifts.updateRecords();
     }
+
+    public Id id() {
+        return this.batchId;
+    }
+
+    public Gifts gifts() {
+        if (this.gifts == null) {
```
kenlewis-sfdo, line 128: This null check shouldn't be necessary since the var is set in the only constructor and there is no public setter.

src/classes/GE_GiftEntryController.cls
psadie-sfdo: I'll update these to use GiftId and GiftBatchId value objects!

src/classes/GiftBatch.cls
```diff
@@ -93,35 +66,29 @@ public inherited sharing class GiftBatch {
...
     public GiftBatch(Id batchId) {
-        this.batchId = batchId;
-        this.batch = giftBatchSelector.selectGiftBatchById(this.batchId);
+        this.giftBatchId = new GiftBatchId(batchId);
+        this.batch = giftBatchSelector.selectGiftBatchBy(this.giftBatchId);
+        this.giftTemplate = new GiftTemplate(this.batch.Form_Template__r.Template_JSON__c);
+        this.gifts = new Gifts(this.giftBatchId, new GiftSelector(giftTemplate));
     }
 
     public Boolean hasElevateWidget() {
```
ptewson-sfdo, line 85: I'm a little surprised to see this in the domain class, though I certainly wouldn't dig my heels in over this. It seems like the presence of the elevate widget is more of a "view-like" behavior?

src/classes/GiftTemplate.cls
```diff
@@ -0,0 +1,67 @@
+public with sharing class GiftTemplate {
```
ptewson-sfdo, line 1: It feels a bit strange to have both the GiftTemplate class and the GE_Template class. I guess GE_Template is pretty tightly coupled to the implementation and JSON structure we are storing it with, but for whatever reason this caught my attention. 

Did you end up resolving the issue with persisting permissions errors in the template JSON?

ptewson-sfdo (commented):
I wanted to get these comments in your hands before EOD! Looking good!

src/classes/GiftBatch.cls
psadie-sfdo: You're absolutely right! This doesn't quite belong here. We need to clean this out once Gift Entry has been completely decoupled from the legacy BGE classes.

This is in place currently because a legacy BGE utility class needs to check the batch's template for the elevate widget. It needs to do that so that we can display a couple of Elevate-related Data Import fields in the BGE datatable.

I'm hoping we can perform this removal during the Gift Entry refactor work item next sprint.

src/classes/GiftTemplate.cls
psadie-sfdo: It does to me as well, feels a bit incomplete. The primary reason I created this separately was so that whenever we work with form templates in Gift Entry specifically, we'd have a class that encompassed all the Gift Entry specific behavior related to form templates. Definitely not super tied to this particular implementation of that so I'd be happy to move this to GE_Template if we want.

And no, unfortunately, I haven't gotten around to resolving the issue with persisting perms error in the template json.

src/classes/GiftBatch.cls
psadie-sfdo: Taking another look at this... I actually did deliver this method in the new GiftTemplate class as well. What do you think about replacing the `GiftBatch.hasElevateWidget()` method with a `GiftBatch.template()`?

We could then update the legacy BGE_BatchGiftEntryUtil to do something like `giftBatch.template().hasElevateWidget()`.

ptewson-sfdo (commented):
Only additional feedback is for the test using the equality of the query string! I'm wrapped on review as of this round

ptewson-sfdo (approved):
Looks great!

PR #6538, opened 2021-07-08
[RD] Adding new SObject and Custom Setting for tracking Change History

lparrott:
# Critical Changes

# Changes
 - We added a Recurring Donation Change History object to track amount, schedule, or campaign updates.

# Issues Closed

# Community Ideas Delivered

# New Metadata

- Objects:
  - RecurringDonationChangeHistory__c

- Fields
  - RecurringDonationChangeHistory__c.Change_Type__c
  - RecurringDonationChangeHistory__c.Effective_Date__c
  - RecurringDonationChangeHistory__c.New_Amount__c
  - RecurringDonationChangeHistory__c.New_Campaign__c
  - RecurringDonationChangeHistory__c.New_Payment_Method__c
  - RecurringDonationChangeHistory__c.New_Schedule__c
  - RecurringDonationChangeHistory__c.New_Status__c
  - RecurringDonationChangeHistory__c.New_Status_Reason__c
  - RecurringDonationChangeHistory__c.New_Yearly_Value__c
  - RecurringDonationChangeHistory__c.Previous_Amount__c
  - RecurringDonationChangeHistory__c.Previous_Campaign__c
  - RecurringDonationChangeHistory__c.Previous_Payment_Method__c
  - RecurringDonationChangeHistory__c.Previous_Schedule__c
  - RecurringDonationChangeHistory__c.Previous_Status__c
  - RecurringDonationChangeHistory__c.Previous_Status_Reason__c
  - RecurringDonationChangeHistory__c.Previous_Yearly_Value__c
  - RecurringDonationChangeHistory__c.Recurring_Donation__c
  - npe03__Recurring_Donations_Settings__c.EnableChangeHistory__c

# Deleted Metadata


src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -34,6 +34,16 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>EnableChangeHistory__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>When Enabled, NPSP creates a Recurring Donation Change History record each time amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```
sam-knox, line 40:
```suggestion
        <description>When enabled, NPSP creates a Recurring Donation Change History record each time amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```

src/objects/npe03__Recurring_Donations_Settings__c.object
```diff
@@ -34,6 +34,16 @@
...
+        <externalId>false</externalId>
+        <inlineHelpText>When Enabled, NPSP creates a Recurring Donation Change History record each time amount, schedule, or campaign fields of a Recurring Donation are updated.</inlineHelpText>
```
sam-knox, line 42:
```suggestion
        <inlineHelpText>When enabled, NPSP creates a Recurring Donation Change History record each time amount, schedule, or campaign fields of a Recurring Donation are updated.</inlineHelpText>
```

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+        <type>MasterDetail</type>
+        <writeRequiresMasterRead>false</writeRequiresMasterRead>
+    </fields>
+    <label>Recurring Donation Change History</label>
+    <nameField>
+        <displayFormat>RDCH-{0000000}</displayFormat>
+        <label>Recurring Donation Change History ID</label>
+        <type>AutoNumber</type>
+    </nameField>
+    <pluralLabel>Recurring Donation Change History</pluralLabel>
```
sam-knox, line 385:
```suggestion
    <pluralLabel>Recurring Donation Change Histories</pluralLabel>
```

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+    </actionOverrides>
+    <actionOverrides>
+        <actionName>View</actionName>
+        <formFactor>Small</formFactor>
+        <type>Default</type>
+    </actionOverrides>
+    <allowInChatterGroups>false</allowInChatterGroups>
+    <compactLayoutAssignment>SYSTEM</compactLayoutAssignment>
+    <deploymentStatus>Deployed</deploymentStatus>
+    <description>When Enable Change History in Recurring Donation Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```
sam-knox, line 146:
```suggestion
    <description>When Enable Change History in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```

We probably don't want to point people to Custom Settings, but NPSP Settings instead.

src/objects/RecurringDonationChangeHistory__c.object
lparrott: Ha, I had that originally, but thought it looked funny in the Related List 🤷‍♂️

sam-knox (changes requested):
Just a couple of minor changes requested here. Looking great!

src/objects/RecurringDonationChangeHistory__c.object
sam-knox: You know, I think you're right actually. The standard field history tracking uses the singular version too. Please revert my suggestion! @lparrott

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+    </actionOverrides>
+    <actionOverrides>
+        <actionName>View</actionName>
+        <formFactor>Small</formFactor>
+        <type>Default</type>
+    </actionOverrides>
+    <allowInChatterGroups>false</allowInChatterGroups>
+    <compactLayoutAssignment>SYSTEM</compactLayoutAssignment>
+    <deploymentStatus>Deployed</deploymentStatus>
+    <description>When Enable Change History in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```
sam-knox, line 146:
```suggestion
    <description>When Enable Recurring Donation Change History in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```

Sorry, just realized that the setting label is different than what we had here.

sam-knox (changes requested):
Caught one small change we need, then this is all set!

src/objects/RecurringDonationChangeHistory__c.object
jamesbsfdc: @lparrott Just wanted to check on a few of the fields on the History object: 

- New Payment Method / Previous Payment Method fields
- New Schedule / Previous Schedule fields
- New Status / Previous Status fields
- New Status Reason / Previous Status Reason fields

The equivalent fields on the RD object are picklist fields and the values in the picklists and NPSP ships translations of the picklist values for the the languages we support. How are these text fields going to work for users of the languages we support, or multinational orgs that may have users in more than one language using Recurring Donations? Or are we saying the values in these history fields will be English only?

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+    <enableFeeds>false</enableFeeds>
+    <enableHistory>false</enableHistory>
+    <enableLicensing>false</enableLicensing>
+    <enableReports>true</enableReports>
+    <enableSearch>false</enableSearch>
+    <enableSharing>true</enableSharing>
+    <enableStreamingApi>true</enableStreamingApi>
+    <externalSharingModel>ControlledByParent</externalSharingModel>
+    <fields>
+        <fullName>Change_Type__c</fullName>
```
force2b, line 158: There was a point where we decided all new fields and objects needed to use Camel Case vs. Snake (so `ChangeType__c` in this case). All other RD fields follow this pattern. Has that changed?

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+        <inlineHelpText>The new Payment Method for the related Recurring Donation.</inlineHelpText>
+        <label>New Payment Method</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
+        <fullName>New_Schedule__c</fullName>
```
force2b, line 231: This prevent customers from reporting on changes from Monthly to Yearly or every 1 month to every 3 months, etc. Mainly I'm thinking about the report aspect of this. As a text field that concatenates the schedule into a single string, it won't be parseable for reports or queries.

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+    <fields>
+        <fullName>New_Status__c</fullName>
+        <description>The new Status for the related Recurring Donation.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The new Status for the related Recurring Donation.</inlineHelpText>
+        <label>New Status</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
```
force2b, line 263: We should be aware that as a plain text field this will contain the Status value in the language of the User that made the change (assuming the code retrieves the Label for the picklist api name). It won't dynamically translate like other picklist values. Making this a picklist field (as well as StatusReason and any others) adds a lot of complexity as well, so keeping with a plain text field isn't bad. But it's good to document the translation limitation.

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -0,0 +1,389 @@
...
+        <relationshipOrder>0</relationshipOrder>
+        <reparentableMasterDetail>false</reparentableMasterDetail>
+        <trackTrending>false</trackTrending>
+        <type>MasterDetail</type>
+        <writeRequiresMasterRead>false</writeRequiresMasterRead>
+    </fields>
+    <label>Recurring Donation Change History</label>
+    <nameField>
+        <displayFormat>RDCH-{0000000}</displayFormat>
+        <label>Recurring Donation Change History ID</label>
```
force2b, line 382: I generally think we should avoid the word "Id" for auto number fields since there's already an Id on the object.
```suggestion
        <label>Recurring Donation Change History Num</label>
```

src/objects/RecurringDonationChangeHistory__c.object
lparrott: Nope, I'm just new(ish) here, I'll update them!

src/objects/RecurringDonationChangeHistory__c.object
lparrott: We've been back and forth on this a lot, I'm fine separating them at this point due to the confusion and questions the concatenated string has caused.

src/objects/RecurringDonationChangeHistory__c.object
lparrott: There is no consistency with our existing auto number fields, so I went with one that made sense. Great callout though, having 2 "Id" fields is confusing.

src/objects/RecurringDonationChangeHistory__c.object
lparrott: @jamesbsfdc The field values in RD Change History can either be the picklist selection's Label, which will be translated to the current User's language (I believe), or the picklist selection's API name, which will be uniform for all Users, but will most likely be in English (at least, for the options we provide by default). We haven't implemented this yet, does one option sound better than the other?

Having the History fields be **matching** picklists would be ideal, but then we run into issues with keeping them in sync and up to date with each other.

sam-knox (approved):

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -143,7 +143,7 @@
     <allowInChatterGroups>false</allowInChatterGroups>
     <compactLayoutAssignment>SYSTEM</compactLayoutAssignment>
     <deploymentStatus>Deployed</deploymentStatus>
-    <description>When Enable Change History in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
+    <description>When Enable Recurring Donation Change History in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```
force2b, line 146:
```suggestion
    <description>When &quot;Enable Recurring Donation Change History&quot; in NPSP Settings is enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.</description>
```

src/objects/RecurringDonationChangeHistory__c.object
```diff
@@ -288,45 +300,57 @@
...
+        <description>The previous Installment Frequency for the related Recurring Donation.</description>
         <externalId>false</externalId>
-        <inlineHelpText>The previous Payment Method for the related Recurring Donation.</inlineHelpText>
-        <label>Previous Payment Method</label>
+        <inlineHelpText>The previous Installment Frequency for the related Recurring Donation.</inlineHelpText>
+        <label>Previous Installment Frequency</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
```
force2b, line 325: This is a numeric field in the RecurringDonation and RecurringDonationSchedule objects.

src/package.xml
```diff
@@ -1465,23 +1465,25 @@
...
-        <members>RecurringDonationChangeHistory__c.Previous_Schedule__c</members>
-        <members>RecurringDonationChangeHistory__c.Previous_Status_Reason__c</members>
-        <members>RecurringDonationChangeHistory__c.Previous_Status__c</members>
-        <members>RecurringDonationChangeHistory__c.Previous_Yearly_Value__c</members>
-        <members>RecurringDonationChangeHistory__c.Recurring_Donation__c</members>
+        <members>RecurringDonationChangeHistory__c.ChangeType__c</members>
+        <members>RecurringDonationChangeHistory__c.EffectiveDate__c</members>
+        <members>RecurringDonationChangeHistory__c.NewAmount__c</members>
+        <members>RecurringDonationChangeHistory__c.NewCampaign__c</members>
+        <members>RecurringDonationChangeHistory__c.NewInstallmentFrequency__c</members>
```
force2b, line 1472: I think this object is missing the DayOfMonth field. That's part of the schedue.

src/objects/RecurringDonationChangeHistory__c.object
jamesbsfdc: Thanks for the detail @lparrott! Sounds like both come with tradeoffs for international audiences. 

I guess using the user's language rather than API values sounds like the least bad option if they have to be text values. This would work for the customers whose org is using one language across the board for all users (be this English or any other language). 

The issue is for customers that may have English and Spanish users. If they have users with visibility to all change history data they will see English and Spanish values across the data, which may also make reporting and challenging.  I thinking of international nonprofits here who have gone for a one-org approach rather than an org by country approach. They may have shared resource teams (e.g a data analysis team) who review data across the business (country) units.

Interested to hear what others think. Think this may be important to document @samknox  @mpusto ?

src/package.xml
lparrott: We decided not to track that field, this object is too big already (part of the reason we were going to concatenate schedule into a single field)

src/objects/RecurringDonationChangeHistory__c.object
lparrott: Good catch, that's what I get for adding a new field late on a Friday 😬

dfuller114 (approved):
looks good!

rponti-sforg:
QE has completed testing on this PR and corresponding WI.  Tests have passed based on WI Description, ACs, Technical Design document, and PR comments.

For further WI and testing details see:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000QA8PYAW/view
https://salesforce.quip.com/I3YbADPFIh45

Proceeding with merge of PR.  Setting WI status to 'Pending Release'.  WI status will be set to 'Close' after successful merge.

PR #6555, opened 2021-07-21
Gift Entry: Use view models

psadie-sfdo:
[W-9547734](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000QAJXYA4/view)

We're making necessary changes to various LWCs within Gift Entry to use the view models we introduced in #6537. The responsibility of several LWCs will be reduced and much of the work that's the responsibility of a Gift Batch or Gift moved to the geGiftEntryFormApp component and the geGiftBatch and geGift modules.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


src/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -108,4 +121,56 @@ describe('c-ge-batch-gift-entry-header', () => {
             expect(failedGifts).toBe('5');
         });
     });
+
+    describe('event handling and dispatching', () => {
+        it('should dispatch expected custom event when dry run button is clicked', async () => {
+            const batchHeader = setup();
+            const dispatchEventSpy = jest.spyOn(batchHeader, 'dispatchEvent');
```
bdvorachek, line 128: This is a bit different pattern than what I've gone with for testing customEvents. Since we have a reference to the batchHeader component, we can add event listeners to it directly in the test:

```
const handler = jest.fn();
batchHeader.addEventListener('batchdryrun', handler);
expect(handler).toHaveBeenCalled();
```

psadie-sfdo: Updated!

src/classes/GE_GiftEntryController.cls
```diff
@@ -141,6 +141,15 @@ public with sharing class GE_GiftEntryController {
         return giftBatchService.getGiftBatchTotalsMapFor((Id) batchId);
     }
 
+    @AuraEnabled
+    public static GiftBatchView updateGiftBatchWith(DataImportBatch__c newGiftBatchValues){
+        try {
+            return giftBatchService.store(newGiftBatchValues);
```
kenlewis-sfdo, line 147: Store/save should be an operation of the domain instead of the service here.

.gitignore
```diff
@@ -44,6 +44,6 @@ datasets/dev_org/test_data.db
 .cci
 
 # LWC
-/src/lwc/jsconfig.json
-/src/lwc/.eslintrc.json
+force-app/main/default/lwc/jsconfig.json
```
bdvorachek, line 47: I think we're including this jsconfig.json in the repo now, so it could be removed from here.

force-app/main/default/classes/Gifts.cls
```diff
@@ -59,6 +59,15 @@ public with sharing class Gifts {
         setElevateBatchIds(dataImports);
     }
 
+    public Gift findBy(GiftId giftId) {
+        for (DataImport__c dataImport : dataImports) {
```
ptewson-sfdo, line 63: Would it be worth storing dataImports internally as a Map<Id,DataImport__c> so that we don't have to do this iteration?

force-app/main/default/classes/InboundGiftBatchDTO.cls
```diff
@@ -0,0 +1,71 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description DTO class.
```
ptewson-sfdo, line 33: Very minor, but this description could be slightly more verbose!

ptewson-sfdo: Also I will probably figure this out through further review, but ... inbound to what?

force-app/main/default/classes/UTIL_Query.cls
```diff
@@ -179,6 +181,11 @@ public with sharing class UTIL_Query {
         return this;
     }
 
+    public UTIL_Query withOffset(Integer offset) {
```
ptewson-sfdo, line 184: Yes! Great improvement!

ptewson-sfdo (commented):
Prince - this looks and feels really great. I had a small handful of comments, but nothing major and I think the structure is inline with all the team discussions to this point. Awesome to see this coming together!

force-app/main/default/classes/InboundGiftBatchDTO.cls
psadie-sfdo: Ha! Great question, I'll update the description. This should say something closer to "DTO class for inbound data changes to the Gift Batch from the Gift Entry UI".

sokosaurus (approved):
Totally blown away by the scope of changes and refactors that you've implemented here. Amazing work and I'm excited to be able to build (and build faster) on top of all these improvements.

force-app/main/default/lwc/geGiftEntryFormApp/geGiftEntryFormApp.js
```diff
@@ -406,49 +417,56 @@ export default class GeGiftEntryFormApp extends NavigationMixin(LightningElement
...
+        this.gift = new Gift();
+        this.giftInView = this.gift.state();
+        // fireEvent(this, 'formRendererReset', {});
     }
 
-    handleCountChanged(event) {
-        this.count = event.detail.value;
-    }
+    async handleDelete(event) {
+        // TODO: maybe throw spinner while record is being deleted?
```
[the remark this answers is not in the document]
psadie-sfdo: That sounds good to me! I was really hoping to evaluate the way we handle "loading states" in between actions that are async in this work. Another set of eyes there would be awesome!

Besides this gift delete, the other ones that are top of mind for me are the 'Batch Dry Run' and 'Process Batch'. I think the loading state for 'Batch Dry Run' isn't ideal - makes the form look like it's saving a gift. 'Process Batch' on the other hand has this noticeable delay between click and the loading state.

ptewson-sfdo (approved):
Looks good!

PR #6563, opened 2021-07-22
Marking Contact's Address as 'undeliverable' Updates Parent Address Undeliverable Field

dfuller114:
# Critical Changes

# Changes
 - When a user selects or deselects the Undeliverable Mailing Address field on a Contact record, the Undeliverable Mailing Address field on the related Address record is updated with the same value.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/Addresses.cls
```diff
@@ -165,210 +166,293 @@ public inherited sharing class Addresses extends fflib_SObjects {
...
+        HouseholdInfo householdInfo = householdInfoByAccountId.get(accountId);
 
-            // update HH Account's Billing address fields
-            NPSP_Address npspAddress = new NPSP_Address(addr);
+        return new NPSP_Address(householdInfo.seasonalAddress != null ?
+               householdInfo.seasonalAddress : householdInfo.defaultAddress);
+    }
 
-            NPSP_HouseholdAccount npspHouseholdAccount = new NPSP_HouseholdAccount(accountsById.get(accId));
+    private static void updateAddressInfoForAccountsAndContacts(Set<Id> accountIds,
```
bdvorachek, line 247: This method feels like it has quite a few parameters. Are these able to be consolidated, or maybe a class that could be extracted?

force-app/main/default/classes/Addresses.cls
```diff
@@ -165,210 +166,293 @@ public inherited sharing class Addresses extends fflib_SObjects {
...
+        List<Contact> contactsToUpdate = new List<Contact>();
 
-            // finally, if no default found, create an empty default
-            if (hhinfo.addrDefault == null) {
-                Address__c addrEmpty = new Address__c();
-                addrEmpty.Household_Account__c = accId;
-                hhinfo.addrDefault = addrEmpty;
-            }
+        // now go through each HH, and do all the work!
+        updateAddressInfoForAccountsAndContacts(
```
bdvorachek, line 201: One possible idea for shortening the signature of this method, can we have it return an instance of a class that contains `accountsToUpdate`, `addresseesToUpdate` and `contactsToUpdate`?

dfuller114:
There are a number of changes here to the `Addresses` domain class to refactor the previous `refreshHHAddress()` method which was several hundred lines long. This is why there are a number of new methods that will appear in the diff for that class.

force-app/main/default/classes/ADDR_Contacts_TEST.cls
```diff
@@ -0,0 +1,150 @@
...
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
```
lparrott, line 29: Let's add the author, description, etc. comment block here, and describe the types of tests in this class.

force-app/main/default/classes/ADDR_Contacts_TEST.cls
```diff
@@ -0,0 +1,150 @@
+/*
+    Copyright (c) 2014, Salesforce.org
```
lparrott, line 2: 2021! 😉

force-app/main/default/classes/Addresses.cls
lparrott: Something like `AddressAndRelationships`

lparrott (commented):
Nice work on the refactor, it was a little tricky to follow in some places, but the new logic is working great, I just suggested a minor fix 👍

force-app/main/default/classes/Addresses.cls
```diff
@@ -485,69 +569,69 @@ public inherited sharing class Addresses extends fflib_SObjects {
...
     }
 
-    public Map<Id, Address__c> defaultAndSeasonalAddressesByHouseholdAccountId() {
-        Map<Id, Address__c> defaultAndSeasonalAddressesByAcctId = new Map<Id, Address__c>();
-        for (Address__c addr : (List<Address__c>) getRecords()) {
-            NPSP_Address npspAddress = new NPSP_Address(addr);
-            if (isDefault(addr) || npspAddress.isSeasonalAddr()) {
-                defaultAndSeasonalAddressesByAcctId.put(householdAccountIdFor(addr), addr);
+
+    private Boolean shouldSyncAddressChangesToContactOverrides(NPSP_Address npspAddress) {
```
bdvorachek, line 621: Can this instead be an instance method on NPSP_Address? I think `hasChangesToSync()` would be an appropriate name for it over there.

force-app/main/default/classes/Addresses.cls
```diff
@@ -485,69 +569,69 @@ public inherited sharing class Addresses extends fflib_SObjects {
...
-            i++;
         }
-        return defaultAndSeasonalWithChanges;
+        return addressesByAccountId;
     }
 
-    private Boolean isUndeliverableStatusChanged(Address__c addr, Address__c addrOld) {
-        return addr.Undeliverable__c != addrOld.Undeliverable__c;
-    }
+    private Boolean shouldSyncAddressToHouseholdAccountAndContacts(Address__c address, Address__c oldAddress) {
```
bdvorachek, line 585: I think most of this is also able to be an instance method on NPSP_Address as well. I see you've done a few refactors like this in this PR already, with functions that use the old/new version of an Address.  The `isSeasonalChanged(Address__c addr, Address__c oldAddress)` function would also get moved over to that Entity class.

npsp-reedestockton (changes requested):
The refactoring you have done here makes a tremendous difference in the readability of the code.  Great work!  I've made a couple of nitpick comments.

force-app/main/default/classes/Addresses.cls
```diff
@@ -165,210 +166,293 @@ public inherited sharing class Addresses extends fflib_SObjects {
...
+        Address__c defaultAddress = new Address__c();
+        defaultAddress.Household_Account__c = accountId;
+        householdInfo.defaultAddress = defaultAddress;
+    }
+
+    private static Boolean householdHasNoDefaultAddress(HouseholdInfo householdInfo) {
+        return householdInfo.defaultAddress == null;
+    }
+
+    private static void setHouseholdAddressesToDefaultAddress(Map<Id, Address__c> defaultSeasonalAddressesByAccountId, Id accountId, HouseholdInfo householdInfo, Map<Id, List<Address__c>> allAddressesByAccountId) {
```
npsp-reedestockton, line 386: There are some really long lines in this file.  Can we go to multiple lines for those?

force-app/main/default/classes/Addresses.cls
dfuller114: great recommendation!

force-app/main/default/classes/Addresses.cls
dfuller114: great callout. updated!

npsp-reedestockton (approved):

lparrott (approved):

dfuller114:
@sam-knox Updated the `Changes` section. Let me know if there are any changes in the description we need to make.

bdvorachek (approved):

sam-knox:
@dfuller114 - Thanks for the ping to review the change notes. I updated slightly. Am I correct in saying that if you check OR uncheck the field on the Contact, then it triggers the update to the related Address?

dfuller114:
> @dfuller114 - Thanks for the ping to review the change notes. I updated slightly. Am I correct in saying that if you check OR uncheck the field on the Contact, then it triggers the update to the related Address?

@sam-knox yeah that's correct

PR #6568, opened 2021-07-22
[RD2] Recurring Donation Change History

lparrott:
WI: [W-9547597](https://gus.lightning.force.com/a07AH000000QA8tYAG)

# Critical Changes

# Changes

- We added an Enable Recurring Donation Change History Setting in NPSP Settings -> Recurring Donations. When enabled, NPSP creates a Recurring Donation Change History record each time the amount, schedule, or campaign fields of a Recurring Donation are updated.

# Issues Closed

# Community Ideas Delivered

# New Metadata

Classes
- RD2_ChangeHistoryService.cls
- RD2_ChangeHistoryService_TEST.cls

# Deleted Metadata


force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -166,6 +166,16 @@ public inherited sharing class RD2_RecurringDonation {
...
     }
 
+    /**
+    * @description Returns the annual value of this Recurring Donation based on the amount and schedule
+    * @return Decimal The annual value of this Recurring Donation
+    */
+    public Decimal getAnnualValue(){
+        Decimal yearlyFrequency = RD_Constants.PERIOD_TO_YEARLY_FREQUENCY.get(rd.npe03__Installment_Period__c);
+        Decimal installmentFrequency = rd.InstallmentFrequency__c;
+        return rd.npe03__Amount__c * (yearlyFrequency / installmentFrequency);
```
lparrott, line 176: This value gets weird in some edge cases (ex. donation every 3 weeks, or 17.33 donations per year), so we could take the Floor of the interval to get a consistent number, BUT the value provided here will be technically correct over the long term, since some years will have an extra donation.

force-app/main/default/classes/RD_Constants.cls
```diff
@@ -61,4 +61,13 @@ public class RD_Constants {
     public static final String INSTALLMENT_PERIOD_DAILY = 'Daily';
     public static final String INSTALLMENT_PERIOD_QUARTERLY = 'Quarterly';
 
+    public static final Map<String, Integer> PERIOD_TO_YEARLY_FREQUENCY = new Map<String, Integer>{
+      INSTALLMENT_PERIOD_YEARLY => 1,
+      INSTALLMENT_PERIOD_QUARTERLY => 4,
+      INSTALLMENT_PERIOD_MONTHLY => 12,
+      INSTALLMENT_PERIOD_FIRST_AND_FIFTEENTH => 24,
+      INSTALLMENT_PERIOD_WEEKLY => 52,
+      INSTALLMENT_PERIOD_DAILY => 365
```
lparrott, line 70: I don't want to hear about leap years, y'all 😆

force-app/main/default/objects/npe03__Recurring_Donations_Settings__c/fields/EnableChangeHistory__c.field-meta.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
```
lparrott, line 2: This should have been part of feature/234, but I guess I missed it during the sfdx conversion 😬

force-app/main/default/pages/STG_PanelRD.page
```diff
@@ -75,6 +75,21 @@
                     </div>
                 </div>
             </div>
+            <div class="slds-form-element" style="{!IF(isRD2Enabled, 'display:block', 'display:none')}">
```
lparrott, line 78: The location of this setting can move, we haven't discussed that yet.

force-app/main/default/classes/RD2_RecurringDonation.cls
force2b: This does not count pausd months or future schedule changes.  What about using the CurrentYearValue or NextYearValue custom fields?

lparrott: There's some complexity with the timing of when those fields get updated, it's during a "second" update phase which does not trigger the TDTM class again. I'll discuss with Mary, but I believe the intention with this new field is less about "actual value" and more a quick calculation of what this donation will bring in during a normal year.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Recurring Donations
+* @description All services specific to Recurring Donation Change History
+*
+*/
+public without sharing class RD2_ChangeHistoryService {
```
force2b, line 37: I think this specific class should be set to either `with sharing` or `inherited sharing`. What's needed though, is the ability to instantiate and insert data into an object the User doesn't have permission to. At the moment, I'm pretty sure this is only executed in a trigger context - which means it's already executing in System permissions (which is all the more reason to stick with `inherited sharing`). 

It is very possible that at some point this class may be called from another non-trigger entry point. In that case, the sharing context does become more important. My recommendation is to use an inner class in this class explicitly defined as `without sharing` that isolates the operations that will always required escalated perms. This way everything else runs _normally_ and only the very specific operations that need the extra perms always run in that mode. Plus, it helps self-document the operations and reasoning so that this class doesn't get called out by the inevitable security review.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+        get {
+            if (isChangeHistoryEnabled == null) {
+                isChangeHistoryEnabled = new RD2_Settings().isChangeHistoryEnabled();
+            }
+            return isChangeHistoryEnabled;
+        }
+        set;
+    }
+
+    SObjectField rdAmountField = npe03__Recurring_Donation__c.npe03__Amount__c;
```
force2b, line 52: If these are class vars, can they be defined as `private final`?

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+            RecurringDonationChangeHistory__c.PreviousCampaign__c)
+    };
+
+    /**
+     * @description Returns true if any of the key fields used to create Change History has been changed.
+     * @param rd Changed Recurring Donation
+     * @param oldRd Old Recurring Donation
+     * @return Boolean
+     */
+    public void handleHistoryFieldChanged(RD2_RecurringDonation rd, RD2_RecurringDonation oldRd) {
```
force2b, line 96:
```suggestion
    public void handleHistoryFieldChanged(RD2_RecurringDonation updatedRd, RD2_RecurringDonation originalRd) {
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+        RecurringDonationChangeHistory__c rdch;
```
force2b, line 97:
```suggestion
        RecurringDonationChangeHistory__c changeHistory;
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+
+        for(SObjectField rdField : recurringDonationToChangeHistoryFields.keySet()){
+
+            if(rd.fieldChanged(rdField, oldRd)){
+                if(rdch == null){
+                    rdch = returnChangeHistory(rd, oldRd);
+                }
```
force2b, line 104:
```suggestion
            if (rd.fieldChanged(rdField, oldRd)) {
                if (rdch == null) {
                    rdch = returnChangeHistory(rd, oldRd);
                }
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * @description Returns true if any of the key fields used to create Change History has been changed.
+     * @param rd Changed Recurring Donation
+     * @param oldRd Old Recurring Donation
+     * @return Boolean
+     */
+    public void handleHistoryFieldChanged(RD2_RecurringDonation rd, RD2_RecurringDonation oldRd) {
+        RecurringDonationChangeHistory__c rdch;
+
+        for(SObjectField rdField : recurringDonationToChangeHistoryFields.keySet()){
```
force2b, line 99:
```suggestion
        for (SObjectField rdField : recurringDonationToChangeHistoryFields.keySet()) {
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+        ChangeHistoryFieldSet(SObjectField newField, SObjectField previousField) {
+            this.newField = newField;
+            this.previousField = previousField;
+        }
+
+        SObjectField newField;
+        SObjectField previousField;
+    }
+
+    private Map<SObjectField, ChangeHistoryFieldSet> recurringDonationToChangeHistoryFields = new Map<SObjectField, ChangeHistoryFieldSet>{
```
force2b, line 73: Minor, but I think the longer var is hard to read and the shorter one gets across the same meaning
```suggestion
    private Map<SObjectField, ChangeHistoryFieldSet> rd2ChangeHistoryFields = new Map<SObjectField, ChangeHistoryFieldSet>{
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+* @group Recurring Donations
+* @description All services specific to Recurring Donation Change History
+*
+*/
+public without sharing class RD2_ChangeHistoryService {
+
+    /***
+    * @description Determines if Change History is enabled for the org
+    */
+    public Boolean isChangeHistoryEnabled {
```
force2b, line 42: Where is this used?

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,136 @@
...
+            changeHistoryList.add(rdch);
+        }
+
+    }
+
+    public List<RecurringDonationChangeHistory__c> getChangeHistoryRecords(){
+        return changeHistoryList;
+    }
+
+    private RecurringDonationChangeHistory__c returnChangeHistory(RD2_RecurringDonation rd,
```
force2b, line 123:
```suggestion
    private RecurringDonationChangeHistory__c initChangeHistoryRecord(RD2_RecurringDonation rd, 
```

force-app/main/default/classes/RD2_ChangeHistoryService.cls
lparrott: This is a great approach, I'll work on this today. I'll also write a unit test leveraging the Standard User profile to make sure this works as expected.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
lparrott: In the TDTM class, if the setting is not enabled, we return from `createChangeHistoryOnUpdate`

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -70,67 +101,70 @@ public without sharing class RD2_ChangeHistoryService {
...
-     */
-    public void handleHistoryFieldChanged(RD2_RecurringDonation rd, RD2_RecurringDonation oldRd) {
-        RecurringDonationChangeHistory__c rdch;
-
-        for(SObjectField rdField : recurringDonationToChangeHistoryFields.keySet()){
-
-            if(rd.fieldChanged(rdField, oldRd)){
-                if(rdch == null){
-                    rdch = returnChangeHistory(rd, oldRd);
+    private without sharing class ChangeHistoryCreator {
```
force2b, line 104: Nice. My only comment here is to add apexdoc comments to the class and methods - being sure to add something about why this inner class uses `without sharing`.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -70,67 +101,70 @@ public without sharing class RD2_ChangeHistoryService {
...
+                );
         }
 
-    }
-
-    public List<RecurringDonationChangeHistory__c> getChangeHistoryRecords(){
-        return changeHistoryList;
-    }
+        public void insertChangeHistoryRecords(List<RecurringDonationChangeHistory__c> changeHistoryList){
+            insert changeHistoryList;
```
force2b, line 165: Any thoughts on error handling here? For example, should this use allOrNone (see the UTIL_DMLService) and/or should it log insert failures if they do happen?

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
```diff
@@ -34,7 +34,7 @@
 * @description All logic specific to Enhanced Recurring Donation Change History
 *
 */
-@IsTest(IsParallel=true)
+@IsTest(IsParallel=false)
```
force2b, line 37: You don't need to specify `isParallel=false` since that's the default behavior.

force2b (commented):
Looking good. Just a couple of small notes

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,170 @@
...
+        RecurringDonationChangeHistory__c changeHistory;
+
+        changeHistory = changeHistoryCreator.setChangedFields(changeHistory, updatedRd, originalRd);
+
+        if (changeHistory != null) {
+            changeHistoryList.add(changeHistory);
+        }
+    }
+
+    public List<RecurringDonationChangeHistory__c> getChangeHistoryRecords(){
```
dfuller114, line 84: This method seems to only be called from within this class or from test methods. Can we change the private with @TestVisible annotation?

force-app/main/default/classes/RD2_ChangeHistoryService.cls
dfuller114: Not usually a proponent of inner classes, but I see why we are using this one and I like that we have a discrete reason for doing so. Agree with @force2b about documentation on the without sharing.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -0,0 +1,170 @@
...
+    private List<RecurringDonationChangeHistory__c> changeHistoryList = 
+        new List<RecurringDonationChangeHistory__c>();
+
+    /**
+     * @description Returns true if any of the key fields used to create Change History has been changed.
+     * @param updatedRd Changed Recurring Donation
+     * @param originalRd Old Recurring Donation
+     * @return Boolean
+     */
+    public void handleHistoryFieldChanged(RD2_RecurringDonation updatedRd, RD2_RecurringDonation originalRd) {
```
dfuller114, line 74: I think this method should eventually move into a RD2_ChangeHistories aggregate class that would return a list of change history entity records (i.e. List<RD2_ChangeHistory>) that have been changed since it's beyond the service's responbility to determine which change history records have changed. And the RD2_ChangeHistory entity would have a isFieldChanged() method on it.

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
```diff
@@ -0,0 +1,380 @@
...
+    private static final String PAYMENT_CREDIT_CARD = 'Credit Card';
+    private static final String PAYMENT_CHECK = 'Check';
+    private static final Decimal RD_AMOUNT = 100;
+    private static final Decimal RD_NEW_AMOUNT = 200;
+    private static final String RD_PERIOD = RD_Constants.INSTALLMENT_PERIOD_MONTHLY;
+    private static final String RD_NEW_PERIOD = RD_Constants.INSTALLMENT_PERIOD_WEEKLY;
+    private static final Decimal RD_NEW_FREQUENCY = 2;
+
+    /****
+    * @description Creates data required for unit tests
```
dfuller114, line 52: I don't think we need this comment since the behavior of @TestSetup is already publicly documented

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
```diff
@@ -0,0 +1,380 @@
...
+        configureRecurringDonationChangeHistoryEnabled();
+
+        npe03__Recurring_Donation__c rd = getRecurringDonationBuilder(getContact())
+            .withDayOfMonth('20')
+            .withCalculateNextDonationDate()
+            .build();
+
+        RD2_ChangeHistoryService rdchService = new RD2_ChangeHistoryService();
+
+        npe03__Recurring_Donation__c rdUpdate = rd.clone();
```
dfuller114, line 162: Love the use of clone() here to avoid the update DML

force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -531,4 +541,25 @@ public inherited sharing class RD2_RecurringDonation {
...
+    }
+
+    /**
+    * @description Return true if the value of the provided field is different between this
+    * Recurring Donation and the oldRd that was passed in
+    * @param rdField SObjectField to get the value for
+    * @param oldRd RD2_RecurringDonation The Recurring Donation to compare to
+    * @return Boolean
+    */
+    public Boolean fieldChanged(SObjectField rdField, RD2_RecurringDonation oldRd){
```
dfuller114, line 561: could we add the "is" prefix here to clarify that this is a boolean?

force-app/main/default/classes/RD2_Settings.cls
```diff
@@ -83,6 +83,13 @@ public with sharing class RD2_Settings {
         private set;
     }
 
+    /***
+    * @description Indicates if Recurring Donation Change History is Enabled
+    */
+    public Boolean isChangeHistoryEnabled() {
```
dfuller114, line 89: 👍🏻

dfuller114 (changes requested):
looking good! Added a few comments but nothing major. The comment about the aggregate class isn't required for this review.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
lparrott: Good question, error handling is something we lost by not using the TDTM wrapper to insert. Using allOrNone makes sense to me, an error here means something very unexpected happened, so we should prevent further operations.

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
lparrott: You know, I thought that, but I saw this in a different Test class, so **then** I thought something had changed 🤷‍♂️

force-app/main/default/classes/RD2_ChangeHistoryService.cls
lparrott: It isn't the Change History that's changing though, it's the Recurring Donation. I do like the approach you laid out though, we could move that direction with future WIs in this feature area.

force-app/main/default/classes/RD2_RecurringDonation.cls
lparrott: Definitely, great catch.

force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -172,8 +172,8 @@ public inherited sharing class RD2_RecurringDonation {
     */
     public Decimal getAnnualValue(){
         Decimal yearlyFrequency = RD_Constants.PERIOD_TO_YEARLY_FREQUENCY.get(rd.npe03__Installment_Period__c);
-        Decimal installmentFrequency = rd.InstallmentFrequency__c;
-        return rd.npe03__Amount__c * (yearlyFrequency / installmentFrequency);
+        Decimal annualValueToRound = rd.npe03__Amount__c * (yearlyFrequency / rd.InstallmentFrequency__c);
```
force2b, line 175: Technically this doesn't calculate an accurate YearlyValue because it doesn't account for future changes to the schedule such a Pause or just a changed future amount or period. However, it works as a general proxy to determine if the value of the RD went up or down based on a change to something. That noted, it might be useful to update the description on the method to explain usage given that it's not calculating the same value as the "Current Yearly Value" or "Next Yearly Value" fields on the RD.

lparrott: Sure, I'll make sure the fields have an accurate description as well.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
dfuller114: yeah that's right. good call

dfuller114 (approved):
Looks good! Thanks Luke! Looking forward to seeing this in action.

PR #6570, opened 2021-07-23
SFDX: Remove Duplicate Files

force2b:
Converting to SFDX format left a bunch of duplicate files without the `-meta.xml` extension. This PR removes them from the repository to both avoid confusion and to allow SFDX commands to work without error when using the beta source plugin.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


lparrott:
@force2b This is on me, I did a `git mv` on some folders thinking it would preserve the history, but it looks like it just copied over the old versions, thanks for cleaning it up!

lparrott (approved):

PR #6573, opened 2021-07-27
RD2: Hide Upgrade Page When RD2 Upgrade Is Completed More Than 7 Days

andrewyu-salesforce:
As a user who is already on Enhanced Recurring Donations, I do not need to see a section to upgrade to Recurring Donations.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/STG_Panel.cls
```diff
@@ -110,6 +119,61 @@ public with sharing virtual class STG_Panel {
...
+        RD2_EnablementDelegate_CTRL.EnablementState enablementState = RD2_EnablementDelegate_CTRL.getEnablementState();
+        if (enablementState.isMetaConfirmed == false) {
+            return true;
+        }
+
+        npe03__Recurring_Donations_Settings__c rdSettings = UTIL_CustomSettingsFacade.getRecurringDonationsSettings();            
+        Map<String, UTIL_BatchJobService.JobSummary> summaryByType = new RD2_DataMigrationEnablement.BatchJob().getSummaryByType(rdSettings);
+        
+        return isMirgrationCompleteInThePast7Days(summaryByType)
+            || isSimpleEnablmentEnableInThePast7Days(summaryByType);
```
andrewyu-salesforce, line 144: The way to determine if the upgrade has completed for more than 7 days is by checking if migration batch has ran in the last 7 days/cannot find the apex job from Id (Store in Enablement State), or nightly batch job has ran 7 times if going through simple upgrade.

force-app/main/default/classes/STG_SettingsManager_TEST.cls
```diff
@@ -1909,6 +1910,107 @@ private with sharing class STG_SettingsManager_TEST {
...
+    
+        System.assertEquals(true, isDisplayingRD2UpgradePage,
+            'Should not display RD2 Upgrade page if user has upgrade RD through Simple Upgrade and has not pass 7 days');
+    }
+
+    /**
+    * @description Verifies when RD2 is enabled through simple upgrade and 7 days has pass, the RD2 Upgrade page should be hidden 
+    */
+    @isTest
+    private static void shouldNotDisplayRD2UpgradePageWhenUserFinishRD2SetupWithSimpleUpgrade7DaysAgo() {
```
npsp-reedestockton, line 1999: I think all references to the number of days should be changed for the simple upgrade path.  It's occurrences we are looking for, not days.

force-app/main/default/classes/UTIL_BatchJobService.cls
```diff
@@ -154,6 +154,21 @@ public without sharing class UTIL_BatchJobService {
         return getCompletedDaysBetween(batchJob.CompletedDate, batchJob.Status);
     }
 
+    /***
+    * @description Determines number of days passed since the batch job has completed
+    * @param batchId Batch job Id
+    * @return Integer
+    */
+    public Integer getCompleteDaysBetween(Id batchId) {
```
npsp-reedestockton, line 162: I assume the minor difference in spelling is because Id and String don't work as different signatures.  If that's the case, I think we should use a method name that is more different.  This one is confusing since it almost seems like a misspelling.

npsp-reedestockton (changes requested):
Nice work @andrewyu-salesforce.  My comments are mostly about naming.  The logic seems solid.

force-app/main/default/classes/STG_Panel.cls
```diff
@@ -110,6 +120,58 @@ public with sharing virtual class STG_Panel {
...
+        if (migrationSummary == null) {
+            return false;
+        }
+
+        Integer daysDifference = batchJobService.getCompletedDaysByBatchId(migrationSummary.batchId);
+        return daysDifference <= 7;
+    }
+
+    /**
+    * @description Check if RD Enhancement is completed in simple version in the past 7 days
```
npsp-reedestockton, line 164: Comment refers to number of days instead of occurrences

force-app/main/default/classes/UTIL_BatchJobService.cls
```diff
@@ -154,6 +154,21 @@ public without sharing class UTIL_BatchJobService {
...
+    /***
+    * @description Determines number of days passed since the batch job has completed
+    * @param batchId Batch job Id
+    * @return Integer
+    */
+    public Integer getCompletedDaysByBatchId(Id batchId) {
+        AsyncApexJob batchJob = getAsyncApexJob(batchId);
+
+        if (batchJob == null) {
+            return MAX_DAYS_EXCEEDED;
```
lparrott, line 166: I get why we're returning a large number here (anything greater than 7 would work, right?), but I find it confusing that this method returns `MAX_DAYS_EXCEEDED` for `null`, while `getCompletedDaysByClassName` returns 0 🤷‍♂️

force-app/main/default/classes/STG_SettingsManager_TEST.cls
```diff
@@ -1909,6 +1910,181 @@ private with sharing class STG_SettingsManager_TEST {
...
+        for (Integer i = 0; i < 6; i++) {
+            Database.executeBatch(new RD2_OpportunityEvaluation_BATCH(), 1);
+        }
+
+        Test.startTest();
+        Boolean isDisplayingRD2UpgradePage = new STG_Panel().isDisplayingRD2UpgradePage;
+        Test.stopTest();
+    
+        System.assertEquals(true, isDisplayingRD2UpgradePage,
+            'Should not display RD2 Upgrade page if user has upgrade RD through Simple Upgrade and has not pass 7 days');
```
npsp-reedestockton, line 2066: Message should refer to occurrences, not days

force-app/main/default/classes/STG_SettingsManager_TEST.cls
```diff
@@ -1909,6 +1910,181 @@ private with sharing class STG_SettingsManager_TEST {
...
+    }
+
+    /**
+    * @description Verifies when RD2 is enabled through simple upgrade and 7 days has pass, the RD2 Upgrade page should be hidden
```
npsp-reedestockton, line 2070: Occurrences not days

force-app/main/default/classes/UTIL_BatchJobService.cls
```diff
@@ -144,7 +144,7 @@ public without sharing class UTIL_BatchJobService {
     * @param className Batch job class name
     * @return Integer
     */
-    public Integer getCompletedDaysBetween(String className) {
+    public Integer getCompletedDaysByClassName(String className) {
```
npsp-reedestockton, line 147: I really like the way you changed this and getCompletedDaysByBatchId()

npsp-reedestockton (changes requested):
Nice work!!!  Made a few minor comments, mostly related to comments and messages that still reflect days for the simple upgrade path.

lparrott (approved):
Looks good, I manually updated some code to test the hide/show of the Setting, my batch job will be old enough to do a real test later this week where I can confirm one of the logic paths.

force-app/main/default/classes/UTIL_BatchJobService.cls
andrewyu-salesforce: change to return -1 when batch job does not exist.

npsp-reedestockton (approved):

andrewyu-salesforce:
[W-7780247](https://gus.lightning.force.com/lightning/_classic/%2Fa07B0000008EHrIIAW)

PR #6581, opened 2021-07-30
RD2 Change History, moving logic to Opp Eval so it runs from Pause Modal

lparrott:
- Fixing an issue where RD2 Change History doesn't get created when Pausing through the custom Pause Modal.
- [W-9680206](https://gus.lightning.force.com/a07AH000000SeCcYAK)

# Critical Changes

# Changes
 -NPSP creates a Recurring Donation Change History record when a user pauses the Recurring Donation.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -351,6 +366,8 @@ public inherited sharing class RD2_OpportunityEvaluationService {
         List<npe03__Recurring_Donation__c> closeOppRds = new List<npe03__Recurring_Donation__c>();
         List<RD2_RecurringDonation> rdRecords = new List<RD2_RecurringDonation>();
 
+        setOldRdListIfNull();
```
lparrott, line 369: This turned out to be the fix when using the Pause modal window, since the actual update happens in `handleOutdatedRecurringDonations` further in the code (rather than in a trigger), I needed a way to compare the changed RDs to their original form.

force-app/main/default/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -364,41 +363,14 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
         }
 
         if (triggerAction == TDTM_Runnable.Action.AfterInsert && shouldEvaluateOppsSynchronouslyOnInsert(rds.size() > 1)) {
-            new RD2_QueueableService.EvaluateInstallmentOpportunities(rdIds, rdIdsWhereScheduleChanged)
+            new RD2_QueueableService.EvaluateInstallmentOpportunities(rdIds, rdIdsWhereScheduleChanged, oldRds)
```
lparrott, line 366: This was the best way to have the Change History created in a single place. In the Trigger context, we provide the "old list" for comparison, and in the non-Trigger context, we create our own "old list" by cloning the RDs before updates occur (see my other note)

bdvorachek (commented):
I'm not able to drop this review directly on the lines in question, but RD2_ChangeHistoryService is no longer used [in RD2_RecurringDonations_TDTM](https://github.com/SalesforceFoundation/NPSP/pull/6581/files#diff-524d3370f28cf599d2231b0f5240be1c28e40e15cd87d8768228d2793d5a7589R75-R83) and can be removed.

npsp-reedestockton:
This looks really good, Luke.  The only thing I have a small question about is locking in to the Opportunity Evaluation service.  As long as the only changes we want to track have implications for Opportunities, that works fine, but if we ever want to track other changes, I think we will need to move or augment that logic.

npsp-reedestockton:
I'm curious about the choice to not allow partial success on insert of history records.  

The pattern for other inserts/updates spawned by RD2_OpportunityEvaluationService is to allow partial success and handle errors.

lparrott:
> I'm curious about the choice to not allow partial success on insert of history records.
> 
> The pattern for other inserts/updates spawned by RD2_OpportunityEvaluationService is to allow partial success and handle errors.

I went back and forth on this, but ultimately decided that we don't want to "lose data" or "miss changes" since we have no way of recovering if an RD is successfully updated, and we didn't capture the Change History. If possible, we could also fail the related RD change, but I'm not sure if that's an option.

npsp-reedestockton:
> > I'm curious about the choice to not allow partial success on insert of history records.
> > The pattern for other inserts/updates spawned by RD2_OpportunityEvaluationService is to allow partial success and handle errors.
> 
> I went back and forth on this, but ultimately decided that we don't want to "lose data" or "miss changes" since we have no way of recovering if an RD is successfully updated, and we didn't capture the Change History. If possible, we could also fail the related RD change, but I'm not sure if that's an option.

Since we're in RD2_OpportunityEvaluationService now, I think the ability to tie anything to an RD update is gone.  We'll be in a queueable unless we've hit governor limits, so how we will handle updates to ensure maximum visibility/recoverability is something to think about.

npsp-reedestockton (approved):

bdvorachek (approved):

PR #6588, opened 2021-08-03
Gift Entry - Remove Gift From Elevate

sokosaurus:
# Critical Changes

# Changes

- If an authorized credit card is removed from Salesforce.org Elevate, but the subsequent change to the NPSP Data Import Record fails, we now log an error in the NSPSP "Errors" object.

# Issues Closed

# Community Ideas Delivered

# New Metadata

Custom Labels:

- geErrorElevateDelete
- geErrorRecordFailAfterElevateDelete
- geElevateDeleteErrorLog
- geErrorElevateUpdate
- geElevateUpdateErrorLog

# Deleted Metadata


force-app/main/default/lwc/geGift/geGift.js
```diff
@@ -45,6 +45,15 @@ class Gift {
             softCredits: { ...this._softCredits }
         }
     }
+
+    isRemovableFromElevateBatch() {
+        // PLACEHOLDER; validate that Status = 'Authorized' AND ...
+        return true;
+    }
+
+    isGiftAuthorized() {
```
psadie-sfdo, line 54: Could we rename this to `isAuthorized`?

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -731,6 +736,14 @@ export default class GeFormRenderer extends LightningElement{
             if (!canUpdate) { return; }
         }
 
+        const recordId = this.formState['Id'];
+        if (recordId && this.selectedPaymentMethod() !== PAYMENT_METHOD_CREDIT_CARD) {
+            const gift = this.giftBatch.findGiftBy(recordId);
```
psadie-sfdo, line 741: I think an alternative to instantiating a geGiftBatch to find the individual gift would be to instantiate the geGift directly using giftInView.

psadie-sfdo (changes requested):
Just a couple of early comments!

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
sokosaurus: That makes a lot of sense to me. Would it be ideal to instantiate the geGift when we initially load the individual gift into form state?

force-app/main/default/lwc/geGift/geGift.js
sokosaurus: Absolutely! I had made that change myself but just hadn't committed yet. Glad to see we're aligned.

force-app/main/default/classes/Gift.cls
```diff
@@ -82,6 +82,10 @@ public with sharing class Gift {
         delete this.dataImport;
     }
 
+    public void undeleteRecord() {
```
kenlewis-sfdo, line 85: Is this still needed?

force-app/main/default/classes/GiftBatch.cls
```diff
@@ -140,6 +140,13 @@ public inherited sharing class GiftBatch implements IGiftBatch {
         memberGift.deleteRecord();
     }
 
+    public void undeleteGiftBy(GiftId giftId) {
```
kenlewis-sfdo, line 143: Still needed?

force-app/main/default/classes/ElevateAuthorizedGift.cls
```diff
@@ -64,6 +64,8 @@ public with sharing class ElevateAuthorizedGift {
         this.gatewayTransactionId = authorizedGiftResponse.gatewayTransactionId();
     }
 
+    public ElevateAuthorizedGift() {}
```
kenlewis-sfdo, line 67: Is this empty constructor needed?  Couldn't find any usages.

force-app/main/default/classes/ElevateAuthorizedGift.cls
```diff
@@ -29,12 +29,12 @@
...
 public with sharing class ElevateAuthorizedGift {
     @AuraEnabled public ElevateTokenizedGift tokenizedGift;
-    @AuraEnabled public String paymentId;
+    @AuraEnabled public String paymentId {get; set;}
     @AuraEnabled public DateTime authExpiration;
     @AuraEnabled public String status;
     @AuraEnabled public String paymentMethod;
     @AuraEnabled public String originalTransactionId;
-    @AuraEnabled public String elevateBatchId;
+    @AuraEnabled public String elevateBatchId {get; set;}
```
kenlewis-sfdo, line 37: What is the reason for adding the getter/setters to these two fields?

force-app/main/default/lwc/geElevateBatch/geElevateBatch.js
```diff
@@ -0,0 +1,51 @@
...
+
+    constructor() {
+        this.elevateBatchId = null;
+    }
+    
+    async add(tokenizedGift) {
+        return await this.performAdd(tokenizedGift, true);
+    }
+
+    async performAdd(tokenizedGift, retryOnFailure) {
```
kenlewis-sfdo, line 17: Is the retryOnFailure param ever used as false?  I think it should be removed and the single method should just be called "add"

kenlewis-sfdo (commented):
Looking pretty good so far, David, thanks for all of this work!  I left a few initial comments.  Deploying to test a bit also.

force-app/main/default/classes/Gift.cls
sokosaurus: Nope, this can and will be removed shortly. Thought I had found all references to undelete, so thanks for surfacing this.

force-app/main/default/classes/ElevateAuthorizedGift.cls
sokosaurus: During the `remove()` call on ElevateBatch.js, we pass an ElevateAuthorizedGift to `removeFromElevateBatch()` on GE_GiftEntryController.cls.

The server was expecting a no-argument constructor on ElevateAuthorizedGift in order for the deserialization to occur. Without this constructor, the method call was failing. I may be missing something here so please let me know if you have any recommendations.

force-app/main/default/classes/ElevateAuthorizedGift.cls
sokosaurus: Same with above, this was related to the deserialization of an ElevateAuthorizedGift. Without the getter/setters, elevateBatchId and elevatePaymentId would not be set when passing the object from LWC to the server.

force-app/main/default/lwc/geElevateBatch/geElevateBatch.js
sokosaurus: The `add()` method will attempt to add a gift to an Elevate Batch. If there is a failure on the first attempt, `performAdd()` is recursively called to try and add the gift to a _new_ batch. This second attempt is initiated on line 29 of the class where retryOnFailure is set to false. 

We do this because the existing elevate batch Id may no longer be valid, and without a way to know this proactively, we retry after the first failure, this time on a new elevate batch. Open to making any additional adjustments here if you see fit.

kenlewis-sfdo: Ok.  I think the issue is that the retry logic is inside the method (and the class/type).  I would expect the "add" method to return either success or failure, and the client/user of the object to then determine if they want to retry, which would involve requesting/creating a different ElevateBatch and then trying to call "add" on it.      It seems odd to have a Batch create a different Batch and call methods on it.  

A change here is not required for this WI, but we should revisit later.

kenlewis-sfdo:
@sokosaurus see my comment here about the bullets in the PR:  https://salesforce.quip.com/hjpOAUGvctSz#temp:C:AWccks6nb6h60002305ekem4ds46

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -7072,6 +7072,46 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
         <shortDescription>Failure msg when transaction is processed but record isn&apos;t created in Salesforce</shortDescription>
         <value>Elevate successfully processed the payment, but NPSP didn&apos;t create the record for the transaction in Salesforce. Please correct the errors and resubmit the form to create the Salesforce record.</value>
+    </labels>
+        <labels>
+        <fullName>geElevateDeleteErrorLog</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error log entry for gift removed from Elevate by not removed from Gift Entry.</shortDescription>
+        <value>Gift deleted from {0}, but Gift Entry cannot delete the gift. Remove the gift&apos;s NPSP Data Import record from the batch in NPSP Data Import Batches.</value>
```
melissabarber, line 7082:
```suggestion
        <value>Gift deleted from {0}, but Gift Entry cannot delete the gift. Remove the gift's NPSP Data Import record with ID: {1} from the batch with ID {2} in NPSP Data Import Batches.</value>
```

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -7072,6 +7072,46 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+    </labels>
+    <labels>
+        <fullName>geElevateUpdateErrorLog</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error log entry for gift updated in Elevate by not updated in Gift Entry.</shortDescription>
+        <value>Gift deleted from {0}, but Gift Entry cannot delete the gift. Remove the gift&apos;s NPSP Data Import record from the batch in NPSP Data Import Batches.</value>
```
melissabarber, line 7090:
```suggestion
        <value>Gift updated in {0}, but Gift Entry cannot update the gift. Update the gift's NPSP Data Import record with ID: {1} from the batch with ID {2} in NPSP Data Import Batches.</value>
```

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -7072,6 +7072,46 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+    </labels>
+    <labels>
+        <fullName>geErrorElevateDelete</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when gift deleted in Gift Entry but not deleted from Elevate batch.</shortDescription>
+        <value>Cannot delete gift in {0}. Payment can be charged if you process the batch. Refresh and try again.</value>
```
melissabarber, line 7098:
```suggestion
        <value>Cannot delete gift in {0}. We don't recommend processing this batch, as the payment for this gift can still be charged. Refresh and try deleting again.</value>
```

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -7072,6 +7072,46 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
...
+    </labels>
+    <labels>
+        <fullName>geErrorElevateUpdate</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Error when gift updated in Gift Entry but not fixed in Elevate batch.</shortDescription>
+        <value>Unable to update because cannot update gift in {0}. Refresh and try updating the gift again.</value>
```
melissabarber, line 7106:
```suggestion
        <value>Cannot update gift in {0}. Refresh and try updating the gift again.</value>
```

melissabarber:
@sokosaurus Reviewed the labels and added changes based on my final conversation with the UI text team. Once those suggestions are committed, the labels in this PR are good to go from my end. Thanks!

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -731,8 +734,15 @@ export default class GeFormRenderer extends LightningElement{
             if (!canUpdate) { return; }
         }
 
+        const removeResult = await this.handleRemoveFromElevateBatch(tokenizedGift);
+        if (removeResult.hasError) { return; }
+
         const hasSaved = await this.saveDataImport(this.saveableFormState());
         if (!hasSaved) {
+            if (removeResult.hasProcessed) {
+                this.handleLogError('error', 'save flow'); // PLACEHOLDER
```
psadie-sfdo, line 743: Do these placeholders need to be updated?

psadie-sfdo (changes requested):
A couple of questions and a format request! Also, I think we can leave out the geElevateBatch from the metadata changes in the PR description since it's not a user-facing component.

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
sokosaurus: Absolutely! That should have been resolved in my 2nd to last commit.

sokosaurus:
> @sokosaurus Reviewed the labels and added changes based on my final conversation with the UI text team. Once those suggestions are committed, the labels in this PR are good to go from my end. Thanks!

Thanks for all the support here @melissabarber! I've pushed in your latest label updated but please let me know if anything looks off.

melissabarber:
> Thanks for all the support here @melissabarber! I've pushed in your latest label updated but please let me know if anything looks off.

Updated labels look good!

force-app/main/default/lwc/geGift/geGift.js
```diff
@@ -26,6 +34,10 @@ class Gift {
         delete this._fields[field];
     }
 
+    getFieldValue(field) {
+        return this._fields[field];
+    }
```
psadie-sfdo, line 39: Not a blocker!

Taking another look at this method and its current usage, I think I'd prefer specific individual methods like gift.isPaymentAuthorized(), gift.isPaymentPending(), and gift.paymentId() with the current isAuthorized method being called isPaymentRemovable or something.

Probably inevitable that we'd need a method like getFieldValue, but just a thought in this case.

psadie-sfdo (approved):
Latest changes are looking good! Left one non-blocking comment. Thanks for all the work on this David!

PR #6589, opened 2021-08-03
Marking Account's Address as 'undeliverable' reflects the same value on the Address record

dfuller114:
# Critical Changes

# Changes
 - When a user selects or deselects the Undeliverable Billing Address field on an Account record, the Undeliverable Address field on the related Address record is updated with the same value.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


dfuller114:
In this PR, the first two commits are to check for Advancement in our new integration test classes. If we don't do this, our tests for address management will fail in an org where GEM is installed because GEM uses EDA's address management framework when addresses are created/updated/deleted.

lparrott (approved):
Nice job with this, the change itself works great and there's some good code cleanup in here as well. I just had a few comments, nothing stopping this from moving to QE.

force-app/main/default/classes/HouseholdsApp.cls
```diff
@@ -58,57 +58,45 @@ public inherited sharing class HouseholdsApp {
...
-    * @param listAcc a list of Accounts
-    * @param dmlWrapper to hold the Addresses that need creating
-    ********************************************************************************************************/
-    private void createAddrFromAcc(List<Account> listAcc,
-            TDTM_Runnable.DmlWrapper dmlWrapper) {
-        List<Address__c> listAddr = new List<Address__c>();
-        for (Account acc : listAcc) {
-            NPSP_Address npspAddress = new NPSP_Address(acc);
-            listAddr.add(npspAddress.getRecord());
+    private List<Account> getNewAccountsWithAddressesToCreate(Accounts accounts) {
```
npsp-reedestockton, line 80: This is basically duplicated by Accounts.newAccountsWithAddressesToCreate().  Unless this was created to coordinate with the Household refactoring, shouldn't one of them go away?

npsp-reedestockton (changes requested):

force-app/main/default/classes/HouseholdsApp.cls
dfuller114: good call. This method wasn't actually being used anyway

npsp-reedestockton (approved):

PR #6590, opened 2021-08-03
Feature/234  soft credits behavior

psadie-sfdo:
[W-9547750](https://gus.my.salesforce.com/a07AH000000QAKkYAO)

The Soft Credits field bundle in Gift Entry will now be able to add, remove, and edit unprocessed soft credit entries.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/Gift.cls
+++ b/force-app/main/default/classes/Gift.cls
@@ -82,6 +82,10 @@ public with sharing class Gift {
         delete this.dataImport;
     }
 
+    public void addSoftCredits(SoftCredits incomingSoftCredits) {
+        this.softCredits.addAll(incomingSoftCredits.all());
+    }
+
     private void populateAvailableFields() {
         Map<String, Object> dataImportFields = this.dataImport.getPopulatedFieldsAsMap();
         List<String> fields = new List<String>();
--- a/force-app/main/default/classes/GiftDTO.cls
+++ b/force-app/main/default/classes/GiftDTO.cls
@@ -37,9 +37,16 @@ public with sharing class GiftDTO {
     public Map<String, Object> fields;
     public SoftCreditsDTO softCreditsDTO;
 
+    public GiftDTO() {}
+
     public GiftDTO(Gift gift) {
         this.fields = gift.fields();
         this.softCreditsDTO = new SoftCreditsDTO(gift.softCredits());
+    }
+
+    public GiftDTO(Map<String, Object> genericObject) {
+        this.fields = (Map<String, Object>) genericObject.get('fields');
+        this.softCreditsDTO = (SoftCreditsDTO) genericObject.get('softCreditsDTO');
     }
 
     public Gift asGift() {
@@ -59,7 +66,4 @@ public with sharing class GiftDTO {
 
         return gift;
     }
-
-    @TestVisible
-    private GiftDTO() {}
 }
--- a/force-app/main/default/classes/GiftDTO.cls
+++ b/force-app/main/default/classes/GiftDTO.cls
@@ -41,4 +41,25 @@ public with sharing class GiftDTO {
         this.fields = gift.fields();
         this.softCreditsDTO = new SoftCreditsDTO(gift.softCredits());
     }
+
+    public Gift asGift() {
+        DataImport__c dataImport = new DataImport__c();
+
+        if (fields != null) {
+            for (String key : fields.keySet()) {
+                dataImport.put(key, fields.get(key));
+            }
+        }
+
+        Gift gift = new Gift(dataImport);
+
+        if (softCreditsDTO != null) {
+            gift.addSoftCredits(softCreditsDTO.asSoftCredits());
+        }
+
+        return gift;
+    }
+
+    @TestVisible
+    private GiftDTO() {}
 }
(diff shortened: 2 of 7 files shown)


force-app/main/default/classes/BDI_ObjectWrapper.cls
```diff
@@ -141,11 +141,15 @@ public with sharing class BDI_ObjectWrapper {
...
-            if (bdiMSAdv.importedRecordFieldNames.contains(sourceFieldAPIName.toLowerCase())) {
-                value = dataImport.get(sourceFieldAPIName);
+            Boolean isAnObjectMappingImportField = bdiMSAdv.importedRecordFieldNames.contains(sourceFieldAPIName.toLowerCase());
+            if (isAnObjectMappingImportField) {
+                Object valueFromDynamicSource = dynamicSource.sourceObj.get(sourceFieldAPIName);
+                Object valueFromDataImport = dataImport.get(sourceFieldAPIName);
+                value = valueFromDynamicSource != null ? valueFromDynamicSource : valueFromDataImport;
             } else {
                 value = dynamicSource.sourceObj.get(sourceFieldAPIName);
             }
```
psadie-sfdo, line 155: @callen-sfdo - tagging ya for a review of the change we chatted about

force-app/main/default/classes/GiftService.cls
```diff
@@ -42,4 +42,9 @@ public with sharing class GiftService {
         GiftDTO giftDTO = new GiftDTO(gift);
         return new GiftView(giftDTO);
     }
+
+    public void save(Gift gift) {
```
sokosaurus, line 46: It looks like this method only applies to SGe but the word "save" feels like it could apply to both SGE and BGE (to me at least). Would it make sense to change the name to be more SGE-specific? Thinking something like `saveAndProcess()` or even `saveSingleGift()` to mirror the calling function?

force-app/main/default/classes/AdditionalObjectJSON.cls
```diff
@@ -35,17 +35,50 @@
...
+    // Currently need to clear and re-add soft credits so we can preserve allocations
+    // in the additional object json
+    public void clearSoftCreditsFromDynamicSourceGroup() {
+        List<String> softCreditObjectMappingNames =
+            objectMappingNamesFor(SObjectType.OpportunityContactRole.getName());
+        for (String key : this.dynamicSourceGroup.dynamicSourceByObjMappingDevName.keySet()) {
+            BDI_DynamicSource dynamicSource =
+                this.dynamicSourceGroup.dynamicSourceByObjMappingDevName.get(key);
+
+            if (softCreditObjectMappingNames.contains(dynamicSource.objectMappingTemplateDevName)) {
```
sokosaurus, line 74: 👍

ptewson-sfdo (commented):
This is looking good! Left a few comments/questions

sokosaurus (changes requested):
Just a small comment from me for this pass. Great work here!

force-app/main/default/lwc/geFormRenderer/geFormRendererHelper.js
```diff
@@ -31,15 +31,15 @@ export function flatten(obj) {
...
+        return additionalObjectJson;
     }
 
     const additionalObjects = JSON.parse(additionalObjectJson);
 
     if (isEmptyObject(additionalObjects) ||
         !additionalObjects.hasOwnProperty('dynamicSourceByObjMappingDevName')) {
 
-        return;
+        return additionalObjectJson;
```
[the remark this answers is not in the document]
psadie-sfdo: Great questions!

1. Yep, this will need to be cleared out once we've got allocations as a type as well and only need to perform such transformations in apex.

2. When we return early in this method on line 42, it means that the additional object json has already been transformed to the format that the allocations widget can consume. I can make an update to return a better-named variable for this particular scenario. Thanks for catching!

force-app/main/default/classes/AdditionalObjectJSON.cls
```diff
@@ -87,18 +160,104 @@ public inherited sharing class AdditionalObjectJSON {
...
+            if (objectMapping.Object_API_Name == objectApiName) {
+                objectMappingNames.add(objectMapping.DeveloperName);
+            }
         }
 
         return objectMappingNames;
     }
+
+    private BDI_ObjectMapping objectMappingFor(String objectAPIName) {
+        for (BDI_ObjectMapping objectMapping : mappingService.objectMappingByDevName.values()) {
```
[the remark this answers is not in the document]
psadie-sfdo: I think that's a good idea, but I'd like to hold off on that change for this PR.

force-app/main/default/lwc/geGift/geSoftCredits.js
```diff
@@ -0,0 +1,68 @@
+
+class SoftCredits {
+    _all = [];
```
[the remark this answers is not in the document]
psadie-sfdo: This actually all-that-aren't-processed. Renamed!

force-app/main/default/classes/GiftService.cls
psadie-sfdo: Updated!

ptewson-sfdo (approved):
Looks good! Nice work!

sokosaurus (approved):
Approved! Great work.

voduyemi (approved):
Looks good @psadie-sfdo. Thanks for making the suggested changes.

PR #6595, opened 2021-08-05
RD Change History Sidebar component

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000QA98YAG/view

# Critical Changes

# Changes
* We added a component for viewing Recurring Donation Change History on Recurring Donation Record Page layouts. To use the component, you must have Recurring Donation Change History enabled in Recurring Donation Settings. (and have access to the Apex controller class)

# Issues Closed

# Community Ideas Delivered

# New Metadata
* Lightning Web Components
  * rd2ChangeHistory
  * rd2ChangeEntry
  * rd2ChangeEntryItem
  * rd2ChangeEntryValue
* Classes
  * RD2_ChangeHistorySelector
  * RD2_ChangeHistory
  * RD2_ChangeHistory_TEST
  * RD2_ChangeHistoryController
  * RD2_ChangeHistoryView
  * RD2_ChangeView
  * RD2_FieldChangeView
 

# Deleted Metadata


force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.html
```diff
@@ -0,0 +1,17 @@
+<template>
+    <template if:true={isReady}>
```
dfuller114: instead of using isReady here, could we just check that changeHistory is populated? i.e. `if:true={changeHistory}`?

force-app/main/default/lwc/rd2ChangeEntry/rd2ChangeEntry.html
```diff
@@ -0,0 +1,22 @@
+<template>
+    <lightning-layout vertical-align="center">
+        <lightning-layout-item size="4" class="slds-p-around_x-small">
+            <b><lightning-formatted-date-time value={changeEntry.changeDate}></lightning-formatted-date-time></b>
+        </lightning-layout-item>
+        <lightning-layout-item size="8">
+            <lightning-badge label={changeEntry.changeType} class={badgeClass}></lightning-badge>
+        </lightning-layout-item>
+    </lightning-layout>
+    <template for:each={changeEntry.fields} for:item='item'>
```
dfuller114, line 10: Could we sync the naming here with fields and items? since the `for:item` here is already "item" and the child component name is `rd2ChangeEntryItem` it might make the most sense to change the `changeEntry` property from `fields` to something like `entryItems` and then the for:item could also be changed to `entryItem` so that it all matches.

force-app/main/default/lwc/rd2ChangeEntryValue/rd2ChangeEntryValue.js
```diff
@@ -0,0 +1,34 @@
+import { LightningElement, api } from 'lwc';
+import { NavigationMixin } from 'lightning/navigation';
+
+export default class Rd2ChangeEntryValue extends NavigationMixin(LightningElement) {
+    @api value;
+    @api idValue;
```
dfuller114, line 6: Could we rename this variable to make it more clear what it's used for since it's a public property? Something like entryRecordId?

dfuller114 (commented):
Couple of preliminary comments

force-app/main/default/classes/RD2_ChangeHistoryController.cls
```diff
@@ -0,0 +1,12 @@
+public with sharing class RD2_ChangeHistoryController {
+
+    @AuraEnabled
+    public static RD2_ChangeHistoryView getChangeHistory(Id recurringDonationId, Integer recordLimit, String filter) {
+        RD2_ChangeHistoryService service = new RD2_ChangeHistoryService();
```
dfuller114, line 5: Since this service is a collaborator for this class, could we use the lazy load singleton pattern we have used elsewhere to make it easier to mock the service directly in test classes in the future? Example below:

`@TestVisible
    private static RD2_ChangeHistoryService changeHistoryService {
        get {
            if (changeHistoryService == null) {
                changeHistoryService = new RD2_ChangeHistoryService();
            }
            return changeHistoryService;
        }
        set;
    }`

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -52,6 +52,46 @@ public inherited sharing class RD2_ChangeHistoryService {
...
+    public RD2_ChangeHistoryView getChangeHistoryView(Id recurringDonationId, Integer recordLimit, String filter) {
+        Integer limitPlusOne = recordLimit + 1;
+        changeHistorySelector.withRecordLimit(limitPlusOne);
+
+        if (filter != null) {
+            changeHistorySelector.withChangeTypeFilter(filter);
+        }
+
+        List<RecurringDonationChangeHistory__c> changes = changeHistorySelector.getChangeHistories(recurringDonationId);
+        System.debug(JSON.serializePretty(changes));
```
dfuller114, line 75: we can remove debug?

force-app/main/default/classes/RD2_ChangeView.cls
```diff
@@ -0,0 +1,7 @@
+public with sharing class RD2_ChangeView {
```
dfuller114, line 1: Could we rename this to `RD2_ChangeHistoryView`? I was trying to think if there would be a more broad use for this class outside of the change history but since we have `statusReason` here that narrows the scope to change history so would be good to have that in the class name for clarity I think.

force-app/main/default/classes/RD2_ChangeHistoryView.cls
```diff
@@ -0,0 +1,9 @@
+public with sharing class RD2_ChangeHistoryView {
```
dfuller114, line 1: Do we need a `ChangeHistoryView` and `ChangeView` class? Still thinking through it. In this case I'm not sure the extra specificity adds to the clarity since it's not readily apparent what the difference would be between the two classes.

force-app/main/default/classes/RD2_ChangeHistoryController.cls
bdvorachek: I think that's a little bit more than we'd need there, every method in this file should be `@AuraEnabled public static`, and no two methods in this file should get called in the same transaction.

force-app/main/default/lwc/rd2ChangeEntryValue/rd2ChangeEntryValue.js
bdvorachek: How about `recordIdValue` since entry is already in the name of the component?

force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.html
bdvorachek: Ended up having to use a more complicated conditional here due to `slots` not working when wrapped with `<template if:true` . `hasMoreChanges` is when the app is ready and there are additional records - so the "View More" link should get rendered, and `hasChanges` for when there aren't more changes beyond the number of records rendered.

force-app/main/default/classes/RD2_ChangeHistoryView.cls
bdvorachek: `RD2_ChangeHistoryView` is the top-level message class that contains the whole View for the app. Each `RD2_ChangeView` in this class represents one `RecurringDonationChangeHistory__c` record. We needed a spot to house flags for whether or not the RD2 Change History Setting is enabled, and whether or not to show the "View More" link.

force-app/main/default/classes/RD2_ChangeHistory.cls
```diff
@@ -0,0 +1,210 @@
...
+            view.statusReason = this.record.NewStatusReason__c;
+        }
+
+        view.changeDate = this.record.EffectiveDate__c;
+        view.fields = buildFieldChangeViews();
+
+        return view;
+    }
+
+    private RD2_FieldChangeView getCampaignChange() {
```
lparrott, line 24: Not a blocker, but I would like to pursue a more generic approach to checking for changes and permissions (we chatted about this during our dev call)

force-app/main/default/classes/RD2_ChangeHistorySelector.cls
```diff
@@ -0,0 +1,61 @@
...
+    private List<String> changeTypesFilter;
+
+    private List<String> fields = new List<String>{
+            String.valueOf(RecurringDonationChangeHistory__c.ChangeType__c),
+            String.valueOf(RecurringDonationChangeHistory__c.EffectiveDate__c),
+            String.valueOf(RecurringDonationChangeHistory__c.Id),
+            String.valueOf(RecurringDonationChangeHistory__c.CreatedDate),
+            String.valueOf(RecurringDonationChangeHistory__c.NewAmount__c),
+            String.valueOf(RecurringDonationChangeHistory__c.NewCampaign__c),
+            'NewCampaign__r.Name',
```
lparrott, line 13: This still worries me, but we can test it after the 234 merge 😬

force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.js-meta.xml
```diff
@@ -0,0 +1,19 @@
...
+    <description>Recurring Donation Change History</description>
+    <isExposed>true</isExposed>
+    <masterLabel>Recurring Donation Change History</masterLabel>
+    <targets>
+        <target>lightning__RecordPage</target>
+    </targets>
+    <targetConfigs>
+        <targetConfig targets="lightning__RecordPage">
+            <property name="recordLimit" label="Record Limit" type="Integer" default="3" min="1"/>
+            <property name="changeTypeFilter" label="Change Type Filter" type="String"/>
```
lparrott, line 13: Looks like we can (and should!) use a Custom Label to define help text here.

force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.js
```diff
@@ -0,0 +1,87 @@
...
+        commonViewMore,
+        rdchDisabled,
+        rdchChangeHistory,
+        rdchNoRecords
+    };
+
+    async connectedCallback() {
+        const { recordId, changeTypeFilter, recordLimit } = this;
+        try {
+            this.changeHistory = await getChangeHistory({ recurringDonationId: recordId, changeTypeFilter, recordLimit });
```
lparrott, line 40: I think the order is wrong here? `changeTypeFilter` should be after `recordLimit`

lparrott (changes requested):
Looking great, just had some feedback on the component properties.

force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.js
bdvorachek: Arguments for imperative Apex calls have to be an object, and the shorthand makes it look like it's a list of params here. I'll re-format this for clarity.

lparrott: Hmm, well I couldn't get Change Type filtering to work and thought maybe this was the culprit.

lparrott: Hey I was right, updating to `await getChangeHistory({ recurringDonationId: recordId, recordLimit: recordLimit, filter: changeTypeFilter})` fixes is (you're right though, order doesn't matter)

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -52,6 +52,49 @@ public inherited sharing class RD2_ChangeHistoryService {
...
+        List<RecurringDonationChangeHistory__c> changes = changeHistorySelector.getChangeHistories(recurringDonationId);
+        RD2_ChangeHistoryView changeHistoryView = buildView(changes, recordLimit);
+
+        return changeHistoryView;
+    }
+
+    private RD2_ChangeHistoryView buildView(List<RecurringDonationChangeHistory__c> changes, Integer recordLimit) {
+        RD2_ChangeHistoryView changeHistoryView = new RD2_ChangeHistoryView();
+
+        changeHistoryView.settingEnabled = this.isChangeHistoryEnabled;
```
andrewyu-salesforce, line 87: I think we should also check if RD2 is enabled or not to create a universal experience. Currently the other 2 components has the validation check for it, but for Change History, we only check if the Change History is enabled or not.

Screen Shot 2021-08-10 at 3 14 13 PM

andrewyu-salesforce (commented):
@bdvorachek great work on the component. Just a small feedback on the RD2 validation check

force-app/main/default/classes/RD2_ChangeHistory.cls
bdvorachek: I gave this a run-through on the code I currently have. I ended up with a long method that invokes `new FieldChangeViewDefinition(field, field, label, fieldType, displayRule);` 7 times to build up the list of definitions for the fields that we want to render. Feels a bit like too much is going on in one spot with that change.

force-app/main/default/lwc/rd2ChangeHistory/rd2ChangeHistory.js
bdvorachek: Aha, thanks for spotting that! I changed the Controller args to match what's in the LWC, and reformatted that call so it is a bit easier to read.

andrewyu-salesforce (approved):
@bdvorachek the change looks great.

PR #6601, opened 2021-08-07
Fix GAU CRLP Reset to 0 If Subsequent GAUs Have Allocations Only Tied to Non-Opportunities

### LINKED ISSUE #6204: General Accounting Unit Customizable Rollups Reset to 0 If Subsequent GAUs Have Allocations Only Tied to Non-Opportunities
### Issue
Running the Customizable Rollups batch job for General Accounting Units (CRLP_GAU_BATCH) via the nightly scheduled job or the Bulk Data Process may zero out the rollup fields on a General Accounting Unit depending on the order of how it falls within the batch execution and what types of Allocations are associated with the GAUs following it. These records do calculate properly if a customer clicks the **Recalculate Rollups** button on an affected record.

### Root Cause
Customizable Rollups for GAUs are always processed in Skew Mode via the Skew Dispatcher (which gathers GAUs, 300 at a time by default) and CRLP_GAU_BATCH (which then processes those 300 GAUs’ Allocations, 1000 at a time), which work in tandem until there are no more GAUs to process. When CRLP_GAU_BATCH executes over the Allocations, there’s no guarantee that all of a GAU’s Allocations are processed in the same execution. They may span multiple executions of the batch. So, we save the rollup values for the first and lastId processed in a batch execution and commit any updates in the finish() method of the batch, where all of their Allocations should have been processed at that point.

When we call [saveFirstAndLastParentIdData()](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CRLP_Batch_Base_Skew.cls#L452-L469) in [CRLP_Batch_Base_Skew,](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CRLP_Batch_Base_Skew.cls#L199) we pass in the batch execution’s firstParentId and lastParentId and then we get the lastParentIdProcessed from CRLP_Processor.getLastParentIdProcessed(). If the last GAU Id in the batch had Allocations tied to Opportunities, the lastParentId should match the lastParentIdProcessed. In that case, we update the stateful values for the first/last GAU and commit the data in the finish() method. The issue occurs when the last GAU in the batch only has Allocations that are tied to non-Opportunities (i.e. Campaigns, Recurring Donations, Payments). In that case, we don’t set the lastParentIdProcessed to that final GAU. [So, we hit the 3rd if() block in saveFirstAndLastParentIdData()](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CRLP_Batch_Base_Skew.cls#L461-L463):
```
if (lastParentIdProcessed != lastParentId && lastParentIdProcessed != null && 	processor.getRollupDefsForParent(lastParentIdProcessed) != null) {
            updateStatefulCollections(processor, lastParentIdProcessed);
}
```

[processor.getRollupDefsForParent()](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CRLP_RollupProcessor.cls#L470-L477) does not return null because it looks for that Id in rollupsDefsByParentId. If it doesn’t find it, it [instantiates a whole new List<CRLP_Rollup>](https://github.com/SalesforceFoundation/NPSP/blob/master/src/classes/CRLP_RollupProcessor.cls#L476) for that ID (i.e. all 0s and blanks). So the stateful collection for that middle GAU now includes all 0s/blanks. When we then go to commit that data in the finish() method, it is committed with all 0s.

### Steps to Repeat 
I will try to create a Snowfakery file that does all of this for you… but batch order is not guaranteed.
1. Create 10 General Accounting Units
2. Create 10 Accounts
3. Create a Campaign
4. Go to NPSP Settings and enable Customizable Rollups.
5. In NPSP Settings —> Bulk Data Processes —> Batch Process Settings, set the GAU Batch Size to 10 (mainly so that we don’t have to create thousands of records as test data).
6. This is where things get tricky… order is not guaranteed in a batch job. You need to create enough Opportunities (Closed Won) that have GAU Allocations of 100% the Opportunity Amount that one batch execution of CRLP_GAU has at least 3 GAUs being processed. To determine which GAUs are being captured in which batch execution, run the Rollup Allocations Batch in NPSP Settings and then you’ll need to download the debug logs, open them in VS Code and search for the following string: [440]|parentId|"a0d [note, the 3 characters at the end of that string represent the first 3 characters of the GAU object. I’ve seen it vary widely in orgs, so verify what yours is before doing the search]
7. Confirm all rollup values are correct on your GAUs. Once you’ve created enough Allocations so that the batch has at least 3 GAUs, switch the Allocations for the GAU immediately following that middle GAU so that they are no longer associated with Opportunities and are instead associated with Campaigns.
8. For good measure, set all the values of the middle GAU to a negative number (i.e. -1 or -$1). Leave the dates blank.
9. Go to NPSP Settings —> Bulk Data Processes —> Rollup Allocations Batch. 
10. When the job finishes, navigate to the GAU record for the middle GAU. Its rollup values will be reset to 0. (Clicking Recalculate Rollups will bring the correct values back). 

### Workaround
Clicking on the **Recalculate Rollups** button of an affected General Accounting Unit will reset the rollup fields to their correct values. However, they may be reset to 0s or blanks when the nightly job runs.

npsp-reedestockton:
We fixed an issue that caused a General Account Unit rollup record to reset to 0 if the next General Accounting Unit rollup record processed had no allocatons tied to Opportunities
# Critical Changes

# Changes

# Issues Closed
Fixes #6204 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/CRLP_Batch_Base_Skew.cls
+++ b/force-app/main/default/classes/CRLP_Batch_Base_Skew.cls
@@ -145,6 +145,9 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
                 parentKeyField = parentKeyField.split('\\.')[1];
             }
 
+            // If there is any post-query filtering that needs to be applied, do it here
+            detailRecords = applyPostQueryLocatorFilters(detailRecords);
+
             Id firstParentId, lastParentId;
             if (parentObjectField == null) {
                 firstParentId = (Id)detailRecords[0].get(parentKeyField);
@@ -159,9 +162,6 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
             // Get all parent records to be processed by this batch iteration
             List<Id> parentIds = getUniqueParentIds(this.summaryObjectType, detailRecords,
                 (parentObjectField != null ? parentObjectField + '.' : '') + parentKeyField);
-
-            // If there is any post-query filtering that needs to be applied, do it here
-            detailRecords = applyPostQueryLocatorFilters(detailRecords);
 
             // Map of Payment Child details by parent Opportunity (this job only handles Opportunity or Allocation
             // as the primary queryLocator).


npsp-reedestockton:
Tracks [W-8818353](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000095CUJIA2/view)

npsp-reedestockton:
@lparrott @andrewyu-salesforce Would appreciate your review on this PR.
@sam-knox Would appreciate your review of the PR description.

Thanks all!

lparrott (approved):
Nice job tracking down this bug! Is it possible to write (or update an existing) Unit test for this scenario?

andrewyu-salesforce (approved):

PR #6603, opened 2021-08-09
Add isActive Service Record Check in Elevate Enabled Check

andrewyu-salesforce:
Implement the isActive record validation in Apex to check if Elevate is registered or not
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

 - We added an additional check to the Elevate integration when a request is sent to deregister Elevate from Salesforce.

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/PS_IntegrationServiceConfig.cls
+++ b/force-app/main/default/classes/PS_IntegrationServiceConfig.cls
@@ -76,6 +76,22 @@ public with sharing class PS_IntegrationServiceConfig {
     };
 
     /**
+    * @description Returns the Elevate integration service enablement status
+    * @return Boolean
+    */
+    public Boolean isIntegrationEnabled() {
+        return isEnabled;
+    }
+
+    /**
+    * @description Determines if the current user has the appropriate permissions to modify Elevate records
+    * @return Boolean
+    */
+    public Boolean hasIntegrationPermissions() {
+        return hasPermissions;
+    }
+
+    /**
     * @description Determines if the Elevate integration service is configured and enabled.
     * The Elevate Payment Services integration is considered enabled when all required
     * service records are present with value and isActive record is either missing or true
@@ -120,14 +136,6 @@ public with sharing class PS_IntegrationServiceConfig {
     }
 
     /**
-    * @description Returns the Elevate integration service enablement status
-    * @return Boolean
-    */
-    public Boolean isIntegrationEnabled() {
-        return isEnabled;
-    }
-
-    /**
     * @description Determines if the current user has the appropriate permissions to modify Elevate records
     * @return Boolean
     */
@@ -144,13 +152,6 @@ public with sharing class PS_IntegrationServiceConfig {
         set;
     }
 
-    /**
-    * @description Determines if the current user has the appropriate permissions to modify Elevate records
-    * @return Boolean
-    */
-    public Boolean hasIntegrationPermissions() {
-        return hasPermissions;
-    }
 
     /**
     * @description Returns the same instance of the Configuration inner class
--- a/force-app/main/default/classes/PS_IntegrationServiceConfig.cls
+++ b/force-app/main/default/classes/PS_IntegrationServiceConfig.cls
@@ -53,7 +53,7 @@ public with sharing class PS_IntegrationServiceConfig {
     public static final String CLIENTID = 'clientId';
 
     //Determine if Elevate connection is active
-    public static final String isActive = 'isActive';
+    public static final String IS_ACTIVE = 'isActive';
 
     // Elevate record view types
     public enum VIEW_TYPE {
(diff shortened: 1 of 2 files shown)


force-app/main/default/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -72,28 +75,64 @@ public with sharing class PS_IntegrationServiceConfig {
...
-    * is at least a single record in the protected Payment Services Configuration object
+    * The Elevate Payment Services integration is considered enabled when all required
+    * service records are present with value and isActive record is either missing or true
     * @return Boolean
     */
     private Boolean isEnabled {
         get {
             if (isEnabled == null) {
-                return config.keyValueMap.keySet().containsAll(REQUIRED_CONFIG_KEYS);
+                return isIntegrationActive() && hasElevateServiceData();
```
psadie-sfdo, line 103: Is there additional value to having these two method calls in a private getter? Could we just call these two private methods directly in the public `isIntegrationEnabled`?

psadie-sfdo (changes requested):
Just one question and a minor typo. Could we add a link to the work item in the PR description as well?

Otherwise, this is looking good to me!

andrewyu-salesforce:
[WI-9419790](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000O5rrYAC/view)

force-app/main/default/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -74,26 +77,48 @@ public with sharing class PS_IntegrationServiceConfig {
...
+    }
+
+    /**
+    * @description Determine if isActive service record exist. If the record exist,
+    * using the record to determine if Elevate is registered. Else assume the key
+    * is active
+    * @return Boolean
+    */
+    private Boolean isIntegrationActive() {
+        if (config.keyValueMap.containsKey(isActive)) {
```
lparrott, line 103: I'm glad you took this approach, we should only care about this key/value if a customer deactivated Elevate.

force-app/main/default/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -52,6 +52,9 @@ public with sharing class PS_IntegrationServiceConfig {
     public static final String ELEVATE_SDK = 'elevateSDK';
     public static final String CLIENTID = 'clientId';
 
+    //Determine if Elevate connection is active
+    public static final String isActive = 'isActive';
```
psadie-sfdo, line 56: Should this be `IS_ACTIVE`?

force-app/main/default/classes/PS_IntegrationServiceConfig.cls
psadie-sfdo: What do y'all think about pushing this `config.keyValueMap.containsKey(isActive)` as a query method on config ala `config.hasIsActiveFlag()` and further can do `config.getIsActive()` instead of `Boolean.valueOf(config.keyValueMap.get(isActive))`?

andrewyu-salesforce:
I don't think configuration class should know the context of what is stored in the Payment Service Info. I move the checking logic to the Service class at the bottom. There is a inner class specifically for retrieving different Key Value pair

force-app/main/default/classes/PS_IntegrationServiceConfig.cls
```diff
@@ -78,11 +78,11 @@ public with sharing class PS_IntegrationServiceConfig {
     /**
     * @description Determines if the Elevate integration service is configured and enabled.
     * The Elevate Payment Services integration is considered enabled when all required
-    * service records are present with value and isActive record is either missing or true
+    * service records are present with value and Integration is Active
     * @return Boolean
     */
     public Boolean isIntegrationEnabled() {
-        return isIntegrationActive() && hasElevateServiceData();
+        return isIntegrationActive() && ElevateConfigService.hasAllRequiredServiceData();
```
andrewyu-salesforce, line 85: What do you think of also moving `hasAllRequiredServiceData()` to the Service class and directly call here. I can have the same pattern as `isIntegrationActive()` and call the ConfigService in the inner function. However, I don't see the value for `hasAllRequiredServiceData()` since the function name is pretty self explanatory in this context

psadie-sfdo: I'd say the only value at that point would be consistency. I'd say either we should go ahead and follow the same pattern as `isIntegrationActive` and call the service in the inner function or call the service directly here.

lparrott (approved):

psadie-sfdo (approved):
Looks good! Thanks!

bdvorachek (approved):

andrewyu-salesforce:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000O5rrYAC/view

PR #6620, opened 2021-08-17
Update Recurring Donation and Data Import End Date Help Text

andrewyu-salesforce:
- Update DataImport__c.Recurring_Donation_End_Date__c description and help text
- Update npe03__Recurring_Donation__c.EndDate__c description and help text

# Critical Changes

# Changes

 - To make the field's purpose clearer, we updated the help text for End Date on the Recurring Donation and NPSP Data Import objects. Existing customers will need to make this update manually. Ask your system administrator to update the help text to: "The date on which this Recurring Donation was closed."

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/objects/npe03__Recurring_Donation__c/fields/EndDate__c.field-meta.xml
+++ b/force-app/main/default/objects/npe03__Recurring_Donation__c/fields/EndDate__c.field-meta.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
     <fullName>EndDate__c</fullName>
-    <description>After this date, NPSP no longer creates installments for this Recurring Donation and updates Status to Closed.</description>
+    <description>The date on which this Recurring Donation was closed.</description>
     <externalId>false</externalId>
-    <inlineHelpText>After this date, NPSP no longer creates installments for this Recurring Donation and updates Status to Closed.</inlineHelpText>
+    <inlineHelpText>The date on which this Recurring Donation was closed.</inlineHelpText>
     <label>End Date</label>
     <required>false</required>
     <trackTrending>false</trackTrending>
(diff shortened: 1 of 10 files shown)


lparrott (approved):

force-app/main/default/objects/npe03__Recurring_Donation__c/fields/EndDate__c.field-meta.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
     <fullName>EndDate__c</fullName>
-    <description>After this date, NPSP no longer creates installments for this Recurring Donation and updates Status to Closed.</description>
+    <description>The date on which this Recurring Donation was closed.</description>
     <externalId>false</externalId>
-    <inlineHelpText>After this date, NPSP no longer creates installments for this Recurring Donation and updates Status to Closed.</inlineHelpText>
+    <inlineHelpText>The date on which this Recurring Donation was closed.</inlineHelpText>
```
force2b, line 6: I suspect everyone knows this already, but just in case. InlineHelpText changes to existing fields will not be deployed to any existing customer orgs that already have this field. That would include orgs with or without RD2 enabled. There's really no way to update orgs with RD2 enabled already, but if you wanted to make sure this change was applied to orgs when they enable RD2, the InlineHelpText for this field could be added to the RD2 [metadata](https://github.com/SalesforceFoundation/NPSP/blob/feature/andrewyu/update-rd-end-date-help-text/unpackaged/config/rd2_post_config/objects/npe03__Recurring_Donation__c.object). There are some other fields with help text changes already in that file.

Another important note is that the TSO is still a static org, which means this change won't be applied there either. Someone will need to make the change manually in that org, at least until we're actually building the TSO dynamically from the metadata/repo.

andrewyu-salesforce: Description and Inline Help Text update is added to the RD2 metadata file

sam-knox (approved):
Looks good!

npsp-reedestockton (approved):

PR #6621, opened 2021-08-18
Adding Change Type to RD2 Entry Form for Change History

lparrott:
WI: [W-9547606 ](https://gus.lightning.force.com/a07AH000000QA9cYAG)

# Critical Changes

# Changes

- If Recurring Donation Change History is Enabled, a Change Type field will now be available on the Enhanced Recurring Donation Form.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -265,7 +284,7 @@ public inherited sharing class RD2_RecurringDonation {
 
             if (rd.Status__c == RD2_Constants.STATUS_PAUSED) {
                 rd.Status__c = RD2_Constants.STATUS_ACTIVE;
-                rd.ClosedReason__c = '';
+                rd.ClosedReason__c = null;
```
lparrott, line 287: This change makes it easier to compare the "old" value to the "new" value, since blank picklists are stored as `null`, and not empty strings. I don't see this causing any regression issues!

force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -144,6 +149,20 @@ public inherited sharing class RD2_RecurringDonation {
         return this;
     }
 
+    /**
+    * @description Stores the Change Type in the object, then sets it to null on the Recurring Donation
+    * @return RD2_RecurringDonation This record instance
+    */
+    public RD2_RecurringDonation storeAndClearChangeType() {
```
lparrott, line 156: My reason for doing this is to have the Change Type cleared before the actual Recurring Donation update, but still have the value itself available in the logic for Change History records.

force-app/main/default/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -449,6 +449,7 @@ public inherited sharing class RD2_OpportunityEvaluationService {
             //when Opportunities were evaluated so no need to redo the work
             if (rdRecord
                 .reviseYearValues(yearValueProcessor)
+                .storeAndClearChangeType()
```
andrewyu-salesforce, line 452: Notes: This may affect performance because the DML update over here is separate from the original trigger update.

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
```diff
@@ -77,6 +78,12 @@ public with sharing class RD2_ChangeHistoryService_TEST {
 
         rdchList = getRecurringDonationChangeHistory(rd.Id);
         System.assertEquals(0, rdchList.size(), 'There should be no Change History with the Setting not Enabled');
+
+        // Make sure that the Change Type is cleared, even if it wasn't used
+        Id rdId = rd.Id;
+        List<npe03__Recurring_Donation__c> rdList = 
+            [SELECT Id, ChangeType__c FROM npe03__Recurring_Donation__c WHERE Id = :rdId];
```
andrewyu-salesforce, line 85: We have a rdGatewat class that is used in Test method to query RD. `private static final TEST_SObjectGateway.RecurringDonationGateway rdGateway = new TEST_SObjectGateway.RecurringDonationGateway();` Using this class can reduce the amount of RD query being written in this test class

andrewyu-salesforce (commented):
@lparrott Great work. Have some feedback on nulling out the Change Type.

force-app/main/default/classes/RD2_ChangeHistoryService.cls
```diff
@@ -179,65 +182,116 @@ public inherited sharing class RD2_ChangeHistoryService {
...
+            String changeType = updatedRd.getChangeType();
+            if (changeType != null) {
+                return changeType;
+            }
+
+            Decimal newAnnualValue = updatedRd.getAnnualValue();
+            Decimal previousAnnualValue = originalRd.getAnnualValue();
+
+            if (newAnnualValue > previousAnnualValue) {
+                changeType = RD2_ChangeHistoryService.CHANGE_TYPE_UPGRADE;
```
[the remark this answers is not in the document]
lparrott: Great suggestion, done!

force-app/main/default/classes/RD2_ChangeHistoryService_TEST.cls
lparrott: Great reminder, I just updated them.

PR #6623, opened 2021-08-18
Implement a Compliance GH Action to ensure that instrumentation usage is reviwed

force2b:
- Check for any new or changed apex code that references the "sfdoLogUtil" class directly. This should be blocked entirely.
- Check for any new or changed apex code that references the SfdoInstrumentationService classes. This should add a new reviewer to the PR.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -1,0 +1,33 @@
+name: Compliance_Checking
+on:
+  pull_request:
+    paths:    
+      - '**.cls'
+      - '!unpackaged/**.cls'
+jobs:
+  CoreLoggingCompliance:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Check for References to the SfdoLogging class (not permitted for direct use)
+        id: diff
+        run: |
+          if [ $GITHUB_BASE_REF ]; then
+            git fetch origin $GITHUB_HEAD_REF:$GITHUB_HEAD_REF
+            export DIFF=$( git diff-tree origin/$GITHUB_BASE_REF..$GITHUB_HEAD_REF --patch-with-raw | grep 'SfdoLogUtils.log' )
+          fi
+
+          # echo "$DIFF"
+          # Escape newlines (replace \n with %0A)
+          echo "::set-output name=diff::$( echo "$DIFF" | sed ':a;N;$!ba;s/\n/%0A/g' )"
+
+          if [ $DIFF ]; then
+            echo "::error::Invalid direct use of the `SfdoLogUtils` core class. Splunk Instrumentation MUST use the `sfdoInstrumentationService` class."
+            # Labels are probably not supported for PR's natively. Might need to import an external solution (security first!)
+            # labels:
+            #   - "invalid"
+            #   - "Do Not Merge"
+          fi
+      


force2b:
@jstvz Thank you very much for all of your help on this. I think this is working pretty much as expected. Would you mind reviewing before I merge into the parent branch?

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: Check for previously assigned Instrumentation Label
+        id: label_check
+        run: |
+          echo "::set-output name=prlabels::${{ join(github.event.pull_request.labels.*.name, ',') }}"
+
+      - name: Check for references to the SfdoLogUtils or SfdoInstrumentationServices Apex classes
+        # If the Instrumentation label is already assigned to the PR, all checks are skipped to avoid redundant PR comments and alerts
+        id: get_pr_diff
+        if: contains(steps.label_check.outputs.prlabels, 'Instrumentation') == false
```
jstvz, line 35: You can use contains directly here:
```suggestion
        if: !contains(github.event.issue.labels.*.name, 'Instrumentation')
```

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check for previously assigned Instrumentation Label
+        id: label_check
+        run: |
+          echo "::set-output name=prlabels::${{ join(github.event.pull_request.labels.*.name, ',') }}"
+
```
jstvz, line 31: Not needed if you only need it for the step output (see above):
```suggestion
```

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Check for references to the SfdoLogUtils or SfdoInstrumentationServices Apex classes
+        # If the Instrumentation label is already assigned to the PR, all checks are skipped to avoid redundant PR comments and alerts
+        id: get_pr_diff
+        if: contains(steps.label_check.outputs.prlabels, 'Instrumentation') == false
+        run: |
+          echo "Did not find label 'Instrumentation' in '${{steps.label_check.outputs.prlabels}}'"
```
jstvz, line 37: Ah, I see, you're using the output here. You should still remove that step, but you can use `toJSON(value)` to get pretty printed JSON output here instead.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              assignees: [ "force2b" ]
+            })
+            github.pulls.requestReviewers({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number,
+              reviewers: [ "jstvz" ]
```
jstvz, line 74: I don't think you want me as a reviewer.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: [ "Instrumentation" ]
+            })
+            github.issues.addAssignees({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              assignees: [ "force2b" ]
```
jstvz, line 68: Suggested `TODO`: You can improve maintainability here by moving `force2b` to a workflow-level environment variable.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+            })
+            github.pulls.requestReviewers({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number,
+              reviewers: [ "jstvz" ]
```
jstvz, line 105: Same here.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+            })
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: ':no_entry: @force2b This Pull Request references the "SfdoLogUtils" class which should not be called directly by package code :no_entry:'
+            })
+            core.setFailed('This Pull Request may be calling an SfdoLogUtils.log() method which should not be called directly by package code')
```
jstvz, line 113: 👍🏾

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+        id: assign_reviewers_if_core_class
+        if: contains(steps.get_pr_diff.outputs.diff, 'SfdoLogUtils.log(')
+        uses: SalesforceFoundation/github-script@v4
+        with:
+          script: |
+            github.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: [ "Instrumentation" ]
```
jstvz, line 93: What happens if the label already exists?

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,114 @@
...
+# @author       Michael Smith
+# @date         2021-08-23
+# @description  As SFDO Managed Packages start using the newly introduced Instrumentation Services to 'log' 
+#               usage data to Splunk and Argus, this Compliance Check action does two things:
+#               1. Ensures that Packaged Apex Classes do not make any direct reference to the core SfdoLogUtils class.
+#                  Any code references to this class will cause the Compliance build to fail.
+#               2. At least in the short term, adds an Instrumentation reviewer to any PR that in any way references or
+#                  or uses the SfdoInstrumentationService. This is primarily meant as a short term action to aid in
+#                  adoption of the new instrumentation services.
+#####################################################################################################################
```
jstvz, line 11: That's handy.

jstvz (changes requested):
Suggested some changes. Blocking change is not assigning me as a reviewer. 😏

.github/workflows/compliance.yml
force2b: @jstvz  I could not get `!contains()` to not throw an error. For some reason it just didn't recognize that syntax.

.github/workflows/compliance.yml
force2b: I ended up switching to `join()` instead of `toJSON` to get the array into a single line string. Apparenty that was required for it to work with `contains()`.

.github/workflows/compliance.yml
force2b: @jstvz  I was wondering about that.  Is the right syntax something like:
`echo "::set-output name=reviewer::force2b"`?

What would be even more interesting is to dynamically select the reviewer/assignee based on the owner of the PR. If it's owned by _me_, I can't be set as the reviewer (which is why you're assigned below). So maybe the line above can dynamically change the reviewer based on the current PR owner.  Not quite sure how to do that yet, but I'm somewhat assuming that the owner is in the context possibly.

.github/workflows/compliance.yml
force2b: Nothing. It just leaves it alone

.github/workflows/compliance.yml
jstvz: Annoying. Did you try `!(contains(...))`?

Not a blocker.

.github/workflows/compliance.yml
jstvz: About `contains`: the [docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains) say "If search is an array, this function returns true if the item is an element in the array." 

Did using `github.event.pull_request.labels.*.name` directly not work?

.github/workflows/compliance.yml
jstvz: > I was wondering about that. Is the right syntax something like:
echo "::set-output name=reviewer::force2b"?

I'd use [`env`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#env) here. You could access it with `process.env.VAR` in the script.

> What would be even more interesting is to dynamically select the reviewer/assignee based on the owner of the PR. 

I see a couple of options:

1. Request review from a team... that may or may not work if you're the only member.
2. Add some logic to select the user based on the value of `context.payload.pull_request.user.login`

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,140 @@
...
+    env:
+      DEFAULT_REVIEWER: force2b
+      ALTERNATE_REVIEWER: lparrott
+      ASSIGNED_LABEL: Instrumentation
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
jstvz, line 34: We need to find a more efficient way to get the references we need. Fetching the entire history for HEAD is pretty inefficient: it looks like your workflow runs are averaging around 20-25s, but 18s of that is spent on this step. 18s will add up very quickly in NPSP.

I'm okay with leaving this for now, but we'll want to revisit this before merging into 234. Once you have a known good workflow you can try tweaking this value to save time. We may be able to sidestep it by retrieving and parsing the [PR diff url](https://github.com/SalesforceFoundation/NPSP/pull/6623.diff) (`github.event.pull_request.diff_url`), for example.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,140 @@
...
+name: Instrumentation_Service_Compliance_Verification
+on:
+  pull_request:
+    # Filter the job to only execute if CLS files are in the PR
+    paths:    
+      - '**.cls'
+      - '!unpackaged/**.cls'
+
+jobs:
+  Code_Check:
```
jstvz, line 21: I think moving your workflow name (`Instrumentation_Service_Compliance_Verification`) here would be more descriptive. You can set `name` to be a quoted string (e.g. `"Compliance Checks"`).

jstvz (approved):
Nicely done, @force2b. Left some comments, none of which are blockers here, but will need to be addressed before we merge to the release branch.

An aside, ICYMI: it's possible to [squash and merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits) a PR to tidy up the history a bit.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,140 @@
...
+
+    # Constant vars
+    # These should be adjusted per repository
+    env:
+      DEFAULT_REVIEWER: force2b
+      ALTERNATE_REVIEWER: lparrott
```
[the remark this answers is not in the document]
force2b: Agreed.

.github/workflows/compliance.yml
force2b: Good point. I did notice that was taking some time. Maybe once this is working fully, a incremental check would be sufficient - just checking the latest commit contents rather than the full diff?

force2b:
> An aside, ICYMI: it's possible to [squash and merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits) a PR to tidy up the history a bit.

Thanks. I was also thinking about squashing this for that exact reason.

.github/workflows/compliance.yml
force2b: I see both of those return the same result, but I'm not sure a) what the difference is; or if one is faster than the other. Also, should I change or remove `fetch-depth`? I don't think there's a way to test that at the command line though.

PR #6626, opened 2021-08-23
Project Hummingbird: New Contact Flow Organization

kenlewis-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000Tw2vYAC/view

These changes clarify and organize the code exercised during the insertion of a new Contact in NPSP.  

It also establishes top-level `tdtm` and `test` directories. 

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
@kenlewis-sfdo I'd prefer that all FeatureParameters remain in a single directory and not be split up by "feature". Telemetry itself is a feature. Thanks.

kenlewis-sfdo:
> @kenlewis-sfdo I'd prefer that all FeatureParameters remain in a single directory and not be split up by "feature". Telemetry itself is a feature. Thanks.

Thanks for the feedback, @force2b !  I actually see Telemetry as a platform feature, not an application feature.  And I see NPSP as a package that has sub-modules that each use that feature in their own way, independently.  So eventually I see NPSP having a core "module" and feature modules, like Gift Entry, that each may or may not be using FeatureParameters.  It will be better to group all of the features that a module uses under its parent directory.  I don't want a tdtm folder that has some of its components off in another directory.  Plus I checked the EDA repo and they are doing it this way as well.  How strongly do you feel about it?

force-app/main/adapter/in/sobjects/contact/ContactService.cls
```diff
@@ -208,71 +208,88 @@ public with sharing class ContactService {
...
         Map<Id, Account> accountByPrimaryContactId = new Map<Id, Account>();
         List<Contact> contacts = new List<Contact>();
 
-        for (Contact contactRecord : contactsWithoutAccounts) {
-            //track which contacts are being inserted vs. updated
-            if (contactRecord.Id != null) {
-                contactIds.add(contactRecord.Id);
-            }
-        }
+        //track which contacts are being inserted vs. updated
```
voduyemi, line 214: In a `BeforeInsert` context, this list `contactIds` will be empty, and if it's empty we shouldn't be trying to query for the households that the contacts(household members) belong in. It seems that this flow only benefits Legacy HouseHolds since it's called in its `BeforeUpdate` context.

force-app/main/adapter/in/sobjects/contact/ContactService.cls
```diff
@@ -307,4 +324,32 @@ public with sharing class ContactService {
...
+        List<Id> contactIds = new List<Id>();
+        for (Contact contactRecord : contactsWithoutAccounts) {
+            if (contactRecord.Id != null) {
+                contactIds.add(contactRecord.Id);
+            }
+        }
+        return contactIds;
+    }
+
+    private List<Account> getAccounts(List<Id> contactIds) {
```
voduyemi, line 348: I think we should move this into an `HouseHoldSelector` class. Or better still we let the `ContactAdapter` collect this data since it's already collaborating with the selector already.

force-app/main/adapter/in/sobjects/contact/ContactAddress.cls
```diff
@@ -0,0 +1,97 @@
...
+    private Blob AddressBlob;
+    private String MD5;
+
+    /*******************************************************************************************************
+    * @description Compares a CONV_Account_Conversion_BATCH.ContactAddress object with the current object
+    * @param ca A ContactAddress object
+    * @return Boolean
+    */
+    public Boolean isSameAddress(ContactAddress ca){
+        if(this.MailingStreet == ca.MailingStreet &&
```
voduyemi, line 37: Can we add braces to the `if` statements in this class?

voduyemi (commented):
Thanks @kenlewis-sfdo for all this work. Left some comments. Let me know what you think!

voduyemi (approved):
Looks good! @kenlewis-sfdo

PR #6631, opened 2021-08-24
Refactor the BDI Performance Tests 

force2b:
- Refactor the BDI Performance Tests:
  - Remove all previous tests
  - Enable CRLP in the org before running tests
  - Two new tests, each creating and processing 10K DataImport records
  - The first test has a 100% match to an existing Contact
  - The second test has no match to an existing Contact.
- https://salesforce.quip.com/cBESAkiiPf2h

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


datasets/bdi_benchmark/BDI-With-Matching.recipe.yml
```diff
@@ -0,0 +1,42 @@
...
+  friends:
+      - object: Contact  # Contact for matching
+        fields:
+            FirstName:
+              fake: first_name
+            Email:
+              fake: free_email
+            Phone:
+              fake: phone_number
+            LastName: ${{DataImport__c.Contact1_Lastname__c}}
```
prescod, line 41: Double-checking: last-name alone is sufficient to generate the match?

robot/Cumulus/perftests/BDI_Tests.robot
```diff
@@ -1,138 +1,142 @@
...
-BGE/BDI Import - CO - 1000000 / 250 - 7.5 Objs/DI - 0.75 New Acc 0.25 Mtchd Acc 0.25 New Con 0.25 Mtchd Con 1CO1 0.5CO2 1CO3 1Payment 1Allocation 0.5ASC 1 Opp
-    [Setup]     Setup For Test    1000000    Data Import Field Mapping
-    [Teardown]     Validate Data      1000000
-    [Tags]    ultra-long     advanced-mapping
-    Batch Data Import   250
+## =============================================================================================
+## INDIVIDUAL TEST CASES
+## =============================================================================================
+
+BDI - Advanced Mapping with Full Match to Existing Contact
```
prescod, line 132: I have historically put numbers in the test name so we don't accidentally record apples and oranges in MetaCI. If we decide next month to test 500 rather than 250, I'd like the test records in the MetaCI DB to have a new name.

Ultimately I'll leave it up to you, but that's why I don't do it this way.

force2b: Good suggestion

datasets/bdi_benchmark/BDI-With-Matching.recipe.yml
force2b: I seem to remember that worked, but to be honest it's been a little while. I'll spin up a new scratch org to re-test and confirm.

robot/Cumulus/perftests/BDI_Tests.robot
```diff
@@ -1,138 +1,142 @@
...
+## =============================================================================================
+## Data Creation Keywords:
+## =============================================================================================
 Generate Data
-    [Arguments]    ${count}
-    ${count} =  Convert To Integer	${count}
+    [Arguments]    ${data-matching}
+    ${count} =  Convert To Integer      ${DataImportRecordCount}
+
+    ${recipe} =     Set Variable if    '${data-matching}'=='Full-Match'       ${RecipeWithMatches}
```
force2b, line 47: @prescod  Just re-tested. The full match test is using the wrong recipe file, which means this line is not returning the expected value. I'll try to figure out why, unless you see something obvious.

force2b: @prescod  I added the `ELSE`, but it hasn't helped.  This is what I see from the  output below:
> Generating 10 Records for 'Full-Match' Test Using 'datasets/bdi_benchmark/BDI-without-matching.recipe.yml'

Any thoughts?

prescod (approved):

codexmarcel (approved):

PR #6632, opened 2021-08-25
Project Hummingbird:  New Contact Flow Trigger Consolidation

### LINKED ISSUE #6522: Contact Primary Address Type Changes to "Mailing" When Switched to "None" Value
### Issue:

This change was introduced with NPSP 3.201. When a customer switches the Contact's Primary Address Type (npe01__Primary_Address_Type__c) field to "None" (usually if clearing out  mailing address fields), the Primary Address Type value switches to "Mailing". We previously had an [open Github Issue](https://github.com/SalesforceFoundation/NPSP/issues/5266) where this value reverted back to "Home", but the behavior has changed.

### 
Steps to Repeat:

1. Create a Contact with First/Last Name, Primary Address Type = Home, Mailing Address = 123 1st Avenue, New York, NY 10000. Save.
2. Edit the Contact - switch the Primary Address Type to "None", clear out mailing fields, clear out Current Address. Save.
3. The Primary Address Type switches to "Mailing" and the "Mailing" picklist value is added to the Inactive Values list for that field.

### 
Root Cause:
Coming soon... haven't dug into debug logs yet.

### LINKED ISSUE #6646: Number of Household Members Field Is Not Recalculating When Last Contact is Deleted From Account
### Summary
If a customer deletes the last Contact from an Account (either in the UI and selecting the “Delete Contact and Leave Account Empty” option) or via Apex/API, the Number of Household Members is not being reset to 0. (The Account Name is also not being reset to Anonymous Household).

There are no Unit Tests that assert that Number of Household Members is 0 when there are 0 Contacts under an Account. I replicated this in both the latest version of NPSP (3.206) as well as a release 20 releases back (3.186), so this is not a recent regression. 

### Root Cause
We are not calling the HouseholdNamingService in AfterDelete context. We call processAccountsAfterDelete() in [ACCT_IndividualAccounts_TDTM](https://github.com/SalesforceFoundation/NPSP/blob/main/force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls#L321-L345) (AfterDelete)

* this.Contacts is null since newlist is null
* oldContactById is populated
* deleted Contact gets passed to processAccountsAfterDelete()
* Query for Contacts under that Account (0 found)
* needsAccountUpdate boolean = (if the contact list under that Account is not empty or the Account has Opportunities) && it’s a Household Account, set needsAccountUpdate to true. In our case, it’s false.
    `07:23:05.797 (3640867949)|VARIABLE_ASSIGNMENT|[336]|needsAccountUpdate|false`
    Because it’s false, we don’t add it to the householdIdsToUpdate List and when we call updateHouseholds(), there’s nothing in that list.

If you delete a Contact from the Manage Household page (when it’s the only Contact in the Household Account) and click Save, the Account is renamed to Anonymous Household and the Number of Household Members _is_ correctly set to 0 due to code in HH_ContainerHelper.js.

### Steps to Repeat
1. Create a Contact (First Name = Single; Last Name = Contact)
2. Delete that Contact. When given the option to also delete the Household Account as well, choose the option to “Delete Contact and Leave Empty Account”. This is essentially the same as a customer deleting one or more Contacts via Apex or API.
3. Confirm Contact is deleted and that the Account is still called “Contact Household”. It also still shows Number of Household Members (npsp__Number_of_Household_Members__c) as 1 instead of 0.

kenlewis-sfdo:
Current, performance related WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000ACFVYA4/view

Previous WI related to Organization of the new Contact Flow: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000Tw2vYAC/view

**Summary:**
The changes in this PR make inserting a new Contact more efficient by reducing the number of DML statements and trigger handler executions needed to create a new Household (when using the default Household Account model).  

When creating the new Household, it is set with the appropriate Household Name, Greetings and Number of Members before insert.

Addresses created from the Contact record are inserted before the Contact is inserted and are now mapped from the Contact rather than from the Account, which means the Contact's Primary Address Type field is now included and should be set correctly on the new Household Address.  

Old Household Ids are also submitted for renaming + recalc of Number of Households after a Household Member switches Households.

**Technical Notes:**
Currently, ACCT_IndividualAccounts_TDTM sets the Name and Greetings and # of HH Members after the Contact is inserted and ADDR_Account_TDTM and ADDR_Addresses_TDTM create the new Household Address and set the Current_Address__c value on the new Contact record.  

In this PR the Household Account is set correctly before insert and ADDR_Account_TDTM and ADDR_Addresses_TDTM are disabled temporarily by ACCT_IndividualAccounts_TDTM since it inserts the new Addresses and sets the Contact’s Current Address value.

If you have customized anything about these trigger handlers in your org (Load Order,  Synchronous/Asynchronous, Active/Inactive), or have custom logic also running when new Household Contacts are inserted, you should thoroughly test these changes in a Sandbox.

# Critical Changes

# Changes

# Issues Closed

Fixes #6646
Fixes #6522

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/objects/Address__c/fields/Address_Type__c.field-meta.xml
```diff
@@ -12,7 +12,7 @@
             <sorted>false</sorted>
             <value>
                 <fullName>Home</fullName>
-                <default>true</default>
+                <default>false</default>
```
voduyemi, line 15: Is there a reason for this change?

force-app/main/default/classes/HouseholdNamingService.cls
```diff
@@ -149,8 +151,17 @@ public without sharing class HouseholdNamingService {
     }
 
     private Boolean numberOfHouseholdMembersIsIncorrect(SObject household, List<Contact> householdMembers) {
-        if (householdMembers != null) {
-            return household.get('Number_of_Household_Members__c') != householdMembers.size();
+        Object currentNumberOfHouseholdMembersValue = household.get('Number_of_Household_Members__c');
```
voduyemi, line 154: Is this added logic our way of guarding against not calculating the current number of houshold members?

force-app/main/service/HouseholdService.cls
```diff
@@ -281,4 +303,241 @@ public inherited sharing class HouseholdService {
         households.updateHouseholds(dmlWrapper);
     }
 
+    /*******************************************************************************************************
+    * @description Inserts a new 1:1 or HH Account for an Individual Contact
+    * @param contacts the list of Contacts that need updating.
+    */
+    public void createOneToOneAndHouseholdAccountsFor(List<Contact> contacts){
```
voduyemi, line 310: Non-blocking comment: Like we've discussed, we probably would want to break up this method.

force-app/tdtm/triggerHandlers/account/ACCT_Accounts_TDTM.cls
```diff
@@ -61,7 +61,7 @@ public class ACCT_Accounts_TDTM extends TDTM_Runnable {
             when BeforeUpdate {
                 if (!isContactBeforeInsertContext()) {
                     AccountAdapter accountsAdapter = new AccountAdapter(listNew, listOld);
-                    accountsAdapter.onBeforeUpdate();
+                    accountsAdapter.onBeforeUpdate_ACCT_Accounts_TDTM();
```
voduyemi, line 64: I know we named these methods to differentiate the action that triggered these updates but I would love to see us use the `onBeforeUpdate` method name and maybe fork the behaviour we want in the `AccountAdapter` class since we already have guard methods on the `onBeforeUpdate_ADDR_Account_TDTM` method to determine if we run its containing logic or not.

voduyemi (commented):
Thanks @kenlewis-sfdo  for these changes. I left some non-blocking comments. Let me know what you think!

force-app/main/default/objects/Address__c/fields/Address_Type__c.field-meta.xml
kenlewis-sfdo: Fixing a bug related to this workflow: https://github.com/SalesforceFoundation/NPSP/pull/6632/commits/4eb9f1acda44638477e5d28c1c92eae79a411ad0

force-app/main/default/classes/HouseholdNamingService.cls
kenlewis-sfdo: Yep! And with this change it now runs a recalc if members is null and Numb of HH Members is not null or zero, which recalcs the Number on the old Household when a Contact switches Households.

force-app/tdtm/triggerHandlers/account/ACCT_Accounts_TDTM.cls
kenlewis-sfdo: When there is only a single trigger handler per object, then the logic will be able to be combined into a single onBefore... method.  For now these are named this way to clarify that there are multiple handlers running in the same context on the same objects.

force-app/main/service/HouseholdService.cls
kenlewis-sfdo: I started and clarified things a bit, but this method will probably be revisited and further refined.

voduyemi (approved):
Looks good Ken! Thanks for addressing my comments!

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
```diff
@@ -103,10 +114,98 @@ public inherited sharing class ContactAdapter extends fflib_SObjects {
...
+        Boolean isExistingOneToOne = contactRecord.AccountId != null
+            && accountRecord.npe01__One2OneContact__c != contactRecord.Id
+            && accountRecord.npe01__One2OneContact__c != null
+            && contactRecord.AccountId != oldContact.AccountId;
+
+        if (!isExistingOneToOne) {
+            return false;
+        }
+
+        Boolean isDeletedOldAccount = [SELECT COUNT() FROM Account WHERE Id =: oldContact.AccountId] == 0;
```
force2b, line 187: This a potential to exceed governor limits if this method is called in a loop, such as if deleting multiple Contacts.

force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -1011,6 +1013,9 @@ public inherited sharing class Contacts extends fflib_SObjects {
             contactRecord.npo02__Naming_Exclusions__c != oldContact.npo02__Naming_Exclusions__c) {
             return true;
         }
+        if (contactRecord.AccountId != null && contactRecord.AccountId != oldContact.AccountId) {
+return true;
```
force2b, line 1017: I think indentation is off here

force-app/main/default/classes/HouseholdNamingService.cls
```diff
@@ -163,12 +175,23 @@ public without sharing class HouseholdNamingService {
                     membersByHouseholdId.get(idFor(household)) == null ?
                             new List<Contact>() :
                             membersByHouseholdId.get(idFor(household));
-            if (settings.isAdvancedHouseholdNaming()) {
+            HouseholdName householdName = new HouseholdName(
```
force2b, line 178: This adds complexity to the method. Can the inner loop logic, or part of it, be abstracted out into another method to be easier to read/debug/test?

force-app/main/default/classes/NPSP_Address.cls
```diff
@@ -87,7 +87,7 @@ public inherited sharing class NPSP_Address implements IAddress {
         )) {
             copyFromSObject(con, 'Mailing', con.npe01__Primary_Address_Type__c);
         } else {
-            copyFromSObject(con, 'Mailing', 'Mailing');
+            copyFromSObject(con, 'Mailing', null);
```
force2b, line 90: Is the intent for `null` to not update the Address type? If I'm not mistaken though, this field was always populated with a default address type - or is that code being removed because it wasn't the original legacy behavior?

force-app/main/default/objects/Address__c/fields/Address_Type__c.field-meta.xml
force2b: Does this change introduce a legacy behavior regression though? I'm pretty sure this was always the original default value. If we change it so there is no longer a default value, doesn't that impact what customers expect to happen?

force2b (changes requested):
Nice work on this. I like how the classes are starting to take shape.

force-app/main/default/classes/NPSP_Address.cls
kenlewis-sfdo: These two changes together fix this bug: https://github.com/SalesforceFoundation/NPSP/issues/6522

Before, this field would be set to Home or Mailing even when there is no Address info on the Contact.  Now a blank Primary Address Type on the Contact is respected.  New installs will see blank respected, existing customers will see the default picklist value of Home since it can't be push upgraded.  I ran this by @jesslopez-sf who thought the changes are an improvement and move more toward expected behavior.

force2b: Thank you for the explanation. Makes perfect sense. I forgot that the default value change won't be applied to existing customers.

force-app/main/default/classes/HouseholdNamingService.cls
```diff
@@ -196,6 +186,24 @@ public without sharing class HouseholdNamingService {
...
+                || informalGreeting != null
+                && !informalGreeting.equals(String.valueOf(
+                householdName.asInformalGreeting()));
+    }
+
+    private List<Contact> getHouseholdMembers(Map<Id, List<Contact>> membersByHouseholdId, SObject household) {
+        List<Contact> householdMembers =
+                membersByHouseholdId.get(idFor(household)) == null ?
+                        new List<Contact>() :
+                        membersByHouseholdId.get(idFor(household));
```
force2b, line 203: This logic can be expensive because it issues two `get()`'s against the map. How about:
```
List<Contact> householdMembers = membersByHouseholdId.get(idFor(household));
if (householdMembers == null) {
    householdMembers = new List<Contact>();
}
```

force2b (changes requested):

force2b (approved):
Very nice work!!

PR #6648, opened 2021-09-07
Ensure Contact Merge page behavior is as expected

dfuller114:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -389,83 +389,104 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
+        for (Contact contact : contacts) {
+            Address__c newAddressFromContact = contactAddressesByContact.get(contact);
+            Address__c existingAddressFromContact = existingAddressesByAddress.get(newAddressFromContact);
 
-                    //reset the Primary Address Type only; resetting other fields on Contact will create a new Address
-                    con.npe01__Primary_Address_Type__c = con.npe01__Primary_Address_Type__c == null
-                            ? addrDup.Address_Type__c
-                            : con.npe01__Primary_Address_Type__c;
+            // if found a match
+            if (contactAddressHasAddressMatch(existingAddressFromContact)) {
```
bdvorachek, line 431: I do think that `existingAddressFromContact != null` might be ok to have here instead of a function call

force-app/test/ADDR_Contacts_TEST.cls
```diff
@@ -47,27 +47,124 @@ public with sharing class ADDR_Contacts_TEST {
...
+            return;
+        }
+
+        List<Account> testAccountWithBillingAddress = createTestAccountsWithBillingAddress(1);
+
+        NPSP_Contact npspContact = new NPSP_Contact(UTIL_UnitTestData_TEST.getContact());
+        npspContact.getRecord().AccountId = testAccountWithBillingAddress[0].Id;
+        npspContact.getRecord().is_Address_Override__c = true;
+        npspContact.setUndeliverableMailingAddress(true);
+        System.debug(npspContact);
```
bdvorachek, line 87: Debug statement that can be deleted here.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -627,8 +627,16 @@ public with sharing class CON_ContactMerge_CTRL {
 
         // Sort the custom fields list by name before creating the list of FieldRow for merge UI
         List<String> customFieldList = new List<String>();
-        customFieldList.addAll(customFieldMap.keySet());
+        String undeliverableFieldName = String.valueOf(Contact.Undeliverable_Address__c).toLowerCase();
```
bdvorachek, line 630: This block of code could be extracted to a function called `buildSortedCustomFieldList`.

force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -389,83 +389,104 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
-
-            Map<Contact, Address__c> mapConAddr = new Map<Contact, Address__c>();
-            List<Contact> listCon = new List<Contact>();
-
-            for (Contact con : listConAll) {
-                listCon.add(con);
-                NPSP_Address npspAddressFromContact = new NPSP_Address(con);
-                mapConAddr.put(con, npspAddressFromContact.getRecord());
-            }
+    public void createOrUpdateContactAddress(List<Contact> contacts) {
```
npsp-reedestockton, line 394: This will require coordination with @kenlewis-sfdo

force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -389,83 +389,104 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
-                TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.ADDR_hasRunContactOnBefore, true);
-                for (Contact con : mapConAddrInsertNow.keySet()) {
-                    Address__c addr = mapConAddrInsertNow.get(con);
-                    con.Current_Address__c = addr.Id;
-                }
+        insertContactAddresses(contactAddressesToInsertByContact);
+    }
+
+    private void insertContactAddresses(Map<Contact, Address__c> contactAddressesToInsertByContact) {
+        if (contactAddressesToInsertByContact.size() > 0) {
```
npsp-reedestockton, line 450: Might want to consider returning immediately if test fails rather than having entire body of method inside the if block.

npsp-reedestockton (changes requested):
Looks very good.  Left a couple of minor suggestions.

force-app/main/adapter/in/sobjects/contact/Contacts.cls
dfuller114: updated!

force-app/main/adapter/in/sobjects/contact/Contacts.cls
dfuller114: I think the reason it may be good to keep is that it helps clarify what the null check is doing by wrapping behavior around it. Ideally, this would be a query on an entity to make it a little more straightforward but I think this at least moves us in that direction.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
dfuller114: good callout

npsp-reedestockton (approved):
Great, work @dfuller114!!

dfuller114:
Added a Do Not Merge label to the PR. We are going to wait to see if we will merge #6632 from @kenlewis-sfdo. If that PR will merge this sprint, that PR will then be merged locally into this branch, merge conflicts resolved, and then pushed back up to this branch.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -720,6 +713,17 @@ public with sharing class CON_ContactMerge_CTRL {
 
     }
 
+    private void buildSortedCustomFieldList(Map<String, DescribeFieldResult> customFieldMap, String undeliverableFieldName, List<String> customFieldList) {
```
bdvorachek, line 716: I think a better signature for this would be `private static List<String> buildSortedCustomFieldList(Map<String, DescribeFieldResult> customFieldMap)`. Then on line 629 we could have:

`List<String> customFieldList = buildSortedCustomFieldList(customFieldMap);`

force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -389,81 +391,103 @@ public inherited sharing class Contacts extends fflib_SObjects {
...
-                // prevent the After Insert/Update Contact logic from executing when it's eventually called.
-                TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.flag.ADDR_hasRunContactOnBefore, true);
-                for (Contact con : mapConAddrInsertNow.keySet()) {
-                    Address__c addr = mapConAddrInsertNow.get(con);
-                    con.Current_Address__c = addr.Id;
-                }
+            // no match found, and its an override just for this contact
+            else if (contact.is_Address_Override__c) {
+                // put it on the list of addresss to create now
+                newAddressFromContact.Undeliverable__c = contact.Undeliverable_Address__c;
```
kenlewis-sfdo, line 443: Can this be moved to the NPSP_Address constructor so that the undeliverable value is already set inside newAddressFromContact and doesn't need to be explicitly set here?

bdvorachek (approved):

PR #6654, opened 2021-09-14
Update Newly Created Household Contact with Existing Address Undeliverable Status

dfuller114:
Whenever a new contact is created with no mailing address specified, the Contact Undeliverable Mailing Address is updated with the value of the Undeliverable Household Billing Address.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


dfuller114:
Since we are keeping refactoring to a minimum on this work, the bulk of the code changes is adding additional test scenarios. We also fixed a regression issue with this PR that I've called out in the design [doc](https://salesforce.quip.com/6vjjAJZykN0m)

npsp-reedestockton (approved):

force-app/test/ADDR_Contacts_TEST.cls
```diff
@@ -345,7 +327,61 @@ public with sharing class ADDR_Contacts_TEST {
...
+        Address__c updatedAddress = [SELECT Undeliverable__c
+                                     FROM Address__c
+                                     WHERE Id = :testAddress.Id];
+        Account updatedAccount = [SELECT Undeliverable_Address__c
+                                  FROM Account
+                                  WHERE Id = :UTIL_UnitTestData_TEST.listAccT[0].Id];
+
+        System.assertEquals(updatedContact.Current_Address__c, testAddress.Id, 'The contact should have the expected ' +
+                'current address lookup assigned.');
+        System.assertEquals(updatedAddress.Undeliverable__c, updatedContact.Undeliverable_Address__c,
```
lparrott, line 377: Couldn't this potentially pass if the Address was set to "Deliverable" by mistake? Put another way, is it worth hard coding the expected value to `true` here?

force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -539,7 +539,7 @@ public inherited sharing class Contacts extends fflib_SObjects {
         for (Contact contact : listCon) {
             NPSP_Address npspAddress = new NPSP_Address(addressesByContact.get(contact));
             NPSP_Address npspMatchingAddress = new NPSP_Address(addressMatchesByAddress.get(npspAddress.getRecord()));
-            NPSP_Contact npspContact = new NPSP_Contact(contact);
+            NPSP_Contact npspContact = new NPSP_Contact(contact, oldVersionOf(contact));
```
lparrott, line 542: This doesn't seem to hurt anything, but are we using the old Contact anywhere?

lparrott (commented):
Looks great, just had a few questions before I approve.

dfuller114:
@lparrott just pushed an update to fix the test class that was failing

force-app/test/ADDR_Contacts_TEST.cls
dfuller114: I was going back and forth on that and agree it's better safe than sorry to add an additional like with the explicit true value

force-app/main/adapter/in/sobjects/contact/Contacts.cls
dfuller114: Yeah we actually need that old contact version, particularly when we are checking if the undeliverable status on the contact record has changed in the `isUndeliverableStatusChanged()` method of `NPSP_Contact`. The absence of this parameter in the constructor is what caused the regression I called out in the design doc.

lparrott (approved):

dfuller114:
@sam-knox tagging you for any necessary changes you see in the PR description

bdvorachek (approved):

PR #6658, opened 2021-09-20
CRLP: GAU Rollups not updated if the last Allocation record for a GAU has been deleted

### LINKED ISSUE #4255: CRLP: GAU Rollups not updated if the last Allocation record for a GAU has been deleted
GAU Rollups execute only in Skew Mode where the main object queried is the Allocation object. As a result, if there are no Allocation records for a GAU that had previously had Allocations records that were rolled up, the GAU will not be updated to change the rollup totals to zero/null after the last Allocation is deleted when the next nightly job is executed or when the [Run Batch] button is used in NPSP Settings.

This can affect other Skew Mode rollups as well, though this is unlikely. For example, it would extremely rare for a single Contact to go from having more than 250 related Soft Credit Opportunities to having zero attached Soft Credit Opportunities at once (in a single day).

npsp-reedestockton:
Fixed an issue that could cause a General Allocation Unit record to retain previously calculated values if all related Allocation__c records are deleted.
# Critical Changes

# Changes

# Issues Closed
Fixes #4255 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


npsp-reedestockton:
@force2b @lparrott Would appreciate your feedback on this draft PR.  Still working on unit tests.

npsp-reedestockton:
Tracking [W-8276496](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008hQRkIAM/view)

npsp-reedestockton:
@force2b @lparrott On line 276 in the finish method of CRLP_Batch_Base_Skew, we can restrict to GAU jobType if it seems appropriate.  I have not done so because theoretically, this same error can occur in other run types.  Happy to hear your thoughts on this issue.

force-app/main/default/classes/CRLP_ClearSkewQueueable.cls
```diff
@@ -0,0 +1,119 @@
...
+        } catch (Exception ex) {
+            ERR_Handler.processError(ex, ERR_Handler_API.Context.CRLP + '.' + this.jobType.name());
+        }
+    }
+
+    /**
+     * @description Clear summary records for records whose associated details have been deleted after a skew mode run.
+     */
+    @TestVisible
+    private void clearRollups() {
```
force2b, line 70:
```suggestion
    private void clearRollupFieldValues() {
```

force-app/main/default/classes/CRLP_ClearSkewQueueable.cls
```diff
@@ -0,0 +1,119 @@
...
+        List<SObject> updatedRecords = new List<SObject>();
+        for (SObject parent : parents) {
+            SObject updatedRecord = processor.withSummaryRecord(parent).completeRollupForSingleSummaryRecord(parent);
+            if (updatedRecord != null) {
+                updatedRecords.add(updatedRecord);
+            }
+        }
+
+        if (!updatedRecords.isEmpty()) {
+            CRLP_RollupProcessor_SVC.updateChangedRecordsWithLogging(clearSkewModeFlag(updatedRecords), jobType);
```
force2b, line 91:
```suggestion
            CRLP_RollupProcessor_SVC.updateChangedRecordsWithLogging(
                clearSkewModeFlag(updatedRecords), 
                jobType
            );
```

force-app/main/default/classes/CRLP_ClearSkewQueueable.cls
```diff
@@ -0,0 +1,119 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Skew mode queueable job to zero out summary records whose related detail records have been deleted.
+*/
+public class CRLP_ClearSkewQueueable implements System.Queueable {
```
force2b, line 37:
```suggestion
public class CRLP_ResetRollupFieldsQueueable implements System.Queueable {
```

force-app/main/default/classes/CRLP_ClearSkewQueueable.cls
```diff
@@ -0,0 +1,119 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Rollups
+* @group-content ../../ApexDocContent/Rollups2.htm
+* @description Skew mode queueable job to zero out summary records whose related detail records have been deleted.
```
force2b, line 35: Technically this doesn't actually require Skew. It just knows that it needs to clear out rollup field values regardless of the origin of the all.

force-app/main/default/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -278,6 +293,21 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
 
     }
 
+    /**
+     * @description Identify and return unhandled record Ids from all batches.
+     * @return List<Id> unhandledIds
+     */
+    @TestVisible
+    private List<Id> getUnhandledRecordIds() {
```
force2b, line 301:
```suggestion
    private List<Id> getUnhandledParentRecordIds() {
```

force2b (commented):
Looks very nice. Made some name change recommendations.

force2b:
@npsp-reedestockton 
> On line 276 in the finish method of CRLP_Batch_Base_Skew, we can restrict to GAU jobType if it seems appropriate. I have not done so because theoretically, this same error can occur in other run types. Happy to hear your thoughts on this issue.

Though I think this is a good idea overall, we'd want to thoroughly test it with Account and Contact hard and soft credit rollups. That could be part of this work, or it could be another work item focused on expanding the usage of this and the additional testing.

npsp-reedestockton:
> @npsp-reedestockton
> 
> > On line 276 in the finish method of CRLP_Batch_Base_Skew, we can restrict to GAU jobType if it seems appropriate. I have not done so because theoretically, this same error can occur in other run types. Happy to hear your thoughts on this issue.
> 
> Though I think this is a good idea overall, we'd want to thoroughly test it with Account and Contact hard and soft credit rollups. That could be part of this work, or it could be another work item focused on expanding the usage of this and the additional testing.

Perhaps it makes sense to just see how our work progresses.  We know how to easily ensure this work only applies to GAU jobTypes.  If we're not comfortable we have tested it thoroughly, we can just add the restriction and spin up a second work item for the remaining testing.

npsp-reedestockton:
@force2b @lparrott Restricted jobType to GAU and added the ability to reset appropriate records when GAU rollups without specifying GAU recordIds.  

I still believe we should cover all jobTypes in a future work item, but handling runs without input recordIds was too complex to handle here.  That's the reason for restricting to GAU runs.

I don't anticipate further changes to functionality unless you find something that should be addressed.

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable.cls
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return records;
+    }
+
+    /**
+     * @description Query all GAU records with children.
+     * @return Set<Id>
+     */
+    private Set<Id> getGAUsWithChildren() {
+        String gauWithNamespace = UTIL_Namespace.StrTokenNSPrefix('General_Accounting_Unit__c');
```
force2b, line 159: This query was timing out for me in an LDV Sandbox that has 7M Allocation records. The Query Explain gave it a query cost of 4.95541, which is too high.

force2b (changes requested):
The base logic looks good, but overall I think this class should only handle the scenario where Id's are passed. The larger goal is to only clear out the specific records that need to be cleared. Since it's called at the end of the Skew batch with only the Id's to clear, it should always have what it needs.

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable.cls
npsp-reedestockton: Sounds good.  I'll make that change and prohibit a run with null recordIds.  Thanks!!!

npsp-reedestockton: @force2b @lparrott null recordIds is now prohibited before queueable is enqueued in CRLP_Batch_Base_Skew.  jobType is restricted to GAU there as well.  

I've left the logic to handle other run types in CRLP_ResetRollupFieldsQueueable, but given where we are in the sprint, it is unrealistic to test other scenarios.  Saving that low priority fix for a second work item.

The costly query attempting to deal with null recordIds has been removed.

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable_TEST.cls
```diff
@@ -0,0 +1,83 @@
...
+        gauBatch.handledParentRecordIds.addAll(gauBatch.recordIds);
+
+        Test.startTest();
+        gauBatch.clearUnhandledRollupTotals();
+        Test.stopTest();
+
+        List<General_Accounting_Unit__c> generalAccountingUnits =
+        [SELECT Id, Total_Allocations__c, Total_Number_of_Allocations__c FROM General_Accounting_Unit__c];
+
+        System.assertNotEquals(0, generalAccountingUnits[0].Total_Allocations__c,
```
lparrott, line 78: This is fine for our scenario, but in general I'd like to be more explicit in tests, to ensure an unwanted change didn't occur (assert equal to 90 in this case)

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable.cls
```diff
@@ -0,0 +1,134 @@
...
+        List<SObject> parents = Database.query(query);
+        return parents;
+    }
+
+    /**
+     * @description Unconditionally set skew mode flag false.
+     * @return List<SObject>
+     */
+    @TestVisible
+    private List<SObject> clearSkewModeFlag(List<SObject> updatedRecords) {
```
lparrott, line 126: Since it's new functionality, can we add a small Unit test for this? Or, since we're only calling this on GAUs now, we could just remove it...

force-app/main/default/classes/CRLP_Batch_Base_Skew.cls
```diff
@@ -278,6 +290,35 @@ public abstract class CRLP_Batch_Base_Skew extends CRLP_Batch_Base {
 
     }
 
+    /**
+     * @description Test conditions and enqueue queueable job to clear Rollup totals.
+     */
+    @TestVisible
+    private void clearUnhandledRollupTotals() {
+        if (handledParentRecordIds.size() != recordIds.size()) {
+            if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs()) {
```
lparrott, line 299: If we're over the limit for whatever reason, does this functionality just not run?

lparrott (commented):
Thanks for adding Unit tests, that helped me better understand this scenario. I just had a few comments and questions, looks great over all!

force-app/main/default/classes/CRLP_Batch_Base_Skew.cls
npsp-reedestockton: > If we're over the limit for whatever reason, does this functionality just not run?

Yes, that's correct.  I don't believe it is safe to do anything else.  That said, we should never be over the limit unless someone adds a queueable invocation in the future.

npsp-reedestockton:
@force2b @lparrott Thanks for the suggestions.  Requested changes have been made.

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable_TEST.cls
```diff
@@ -134,4 +138,44 @@ private class CRLP_ResetRollupFieldsQueueable_TEST {
...
+        String filterGroupId2 = CMT_UnitTestData_TEST.getNewRecordId();
+
+        // Extremely simple Filter Rule for IsWon only
+        String filterGroupsJSON = '[' +
+            CMT_UnitTestData_TEST.createFilterGroupRecord(filterGroupId1, 'TestFilterGroup1-IsWon+Paid') + ',' +
+            CMT_UnitTestData_TEST.createFilterGroupRecord(filterGroupId2, 'TestFilterGroup2-IsWon+WrittenOff') +
+            ']';
+        String filterRulesJSON = '[' +
+            CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId1, 'Group1.Rule1', 'Opportunity', 'IsWon', 'Equals', 'True') + ',' +
+            CMT_UnitTestData_TEST.createFilterRuleRecord(null, filterGroupId1, 'Group1.Rule2', 'npe01__OppPayment__c', 'npe01__Paid__c', 'Equals', 'True') + ',' +
```
force2b, line 158: The filter rules against the Payment Object aren't necessary here. In fact, you could just have zero filter groups/rules for the purposes of this test class.

npsp-reedestockton: Cool.  I'll pull them.  Thanks!!

npsp-reedestockton: Rules and filters have been removed.  Also switched to run in parallel.

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable_TEST.cls
```diff
@@ -0,0 +1,160 @@
...
+
+        System.assertEquals(amount, generalAccountingUnits[0].Total_Allocations__c,
+            'Total Allocations should retain previous value');
+        System.assertEquals(count, generalAccountingUnits[0].Total_Number_of_Allocations__c,
+            'Total Number of Allocations should retain previous value');
+    }
+
+    @IsTest
+    private static void shouldClearSkewModeFlagForContact() {
+        List<Contact> contacts = new List<Contact>{new Contact(CustomizableRollups_UseSkewMode__c = true)};
```
force2b, line 96: The `mockRollupCMTValues()` method will need to be called in all test methods well because the `clearUnhandledRollupTotals()`  method will use the rollup services to build a query that retrieves all target fields from the rollup definitions - and then clear those fields. If there are no rollup definitions, then the call to the `clearUnhandledRollupTotals()` method won't actually do anything.

force2b (changes requested):
One more minor change

force-app/main/default/classes/CRLP_ResetRollupFieldsQueueable_TEST.cls
npsp-reedestockton: ahh...  totally makes sense.  thanks!!

force2b (approved):
Awesome!!

lparrott (approved):

PR #6660, opened 2021-09-20
[Gift Entry] geFormSection - Use form state / template instead of getFieldValueAndLabel

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000RBiyYAG/view

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1002,26 +990,26 @@ export default class GeFormRenderer extends LightningElement{
...
         // get data import record helper
-        const di_record = this.getDataImportHelper(fieldWrapper);
+        const dataImportHelper = this.getDataImportHelper();
 
         // donation donor validation depending on selection and field presence
-        let isError = (di_record.donationDonorValue === DONATION_DONOR.isAccount1) ?
-            di_record.isAccount1ImportedEmpty && di_record.isAccount1NameEmpty :
-            di_record.donationDonorValue === DONATION_DONOR.isContact1 &&
-            di_record.isContact1ImportedEmpty && di_record.isContact1LastNameEmpty;
+        let isError = (dataImportHelper.donationDonorValue === DONATION_DONOR.isAccount1) ?
```
psadie-sfdo, line 999: Can this 'let' be a 'const'?

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1002,26 +990,26 @@ export default class GeFormRenderer extends LightningElement{
      * @param sectionsList - Array, all sections
      * @returns {boolean} - true if error message was generated, false if otherwise
      */
-    getDonorTypeValidationError(fieldWrapper, sectionsList) {
+    getDonorTypeValidationError(sectionsList) {
```
dfuller114, line 993: Do we need this additional function? Could we move the boolean block for `isError` in this function into `isDonorTypeInvalid()` since that seems to be the intention of `isDonorTypeInvalid()`.  We should probably then extract the `if (isError)` block into its own function since it has doing something else other than checking the validity of the donor type and seems to have the responsibility of displaying the error message. That function could be named something like `displayDonorTypeError()`. Then, that function could be called in the `isFormValid()` function before we return the `false` value. This way we aren't also calling a display function inside a boolean function whose responsibility is only to check the validity.

Thoughts?

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1073,17 +1061,17 @@ export default class GeFormRenderer extends LightningElement{
 
     /**
      * highlight geForm fields on lSections using sError as message
-     * @param diRecord, Object - helper obj
+     * @param dataImportHelper, Object - helper obj
      * @param lSections, Array of geFormSection
      * @param sError, String to set on setCustomValidity
      */
-    highlightValidationErrorFields(diRecord, lSections, sError) {
+    highlightValidationErrorFields(dataImportHelper, lSections, sError) {
```
dfuller114, line 1068: Could we rename `lSections` to `sections` and `sError` to `errorMessage`?

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1002,26 +990,26 @@ export default class GeFormRenderer extends LightningElement{
...
 
         // donation donor validation depending on selection and field presence
-        let isError = (di_record.donationDonorValue === DONATION_DONOR.isAccount1) ?
-            di_record.isAccount1ImportedEmpty && di_record.isAccount1NameEmpty :
-            di_record.donationDonorValue === DONATION_DONOR.isContact1 &&
-            di_record.isContact1ImportedEmpty && di_record.isContact1LastNameEmpty;
+        let isError = (dataImportHelper.donationDonorValue === DONATION_DONOR.isAccount1) ?
+            dataImportHelper.isAccount1ImportedEmpty && dataImportHelper.isAccount1NameEmpty :
+            dataImportHelper.donationDonorValue === DONATION_DONOR.isContact1 &&
+            dataImportHelper.isContact1ImportedEmpty && dataImportHelper.isContact1LastNameEmpty;
```
psadie-sfdo, line 1002: What do you think of breaking this up like so?
```suggestion
        const isAccountDonor = dataImportHelper.donationDonorValue === DONATION_DONOR.isAccount1;
        const areEmptyAccountFields = dataImportHelper.isAccount1ImportedEmpty && dataImportHelper.isAccount1NameEmpty;
        const isContactDonor = dataImportHelper.donationDonorValue === DONATION_DONOR.isContact1;
        const areEmptyContactFields = dataImportHelper.isContact1ImportedEmpty && dataImportHelper.isContact1LastNameEmpty;

        const isError = 
            isAccountDonor
            ? areEmptyAccountFields
            : isContactDonor && areEmptyContactFields;
```

dfuller114 (changes requested):
great job, Ben. had just a couple comments.

force-app/main/default/lwc/geGiftEntryFormApp/__tests__/geGiftEntryFormApp.test.js
```diff
@@ -90,6 +97,27 @@ jest.mock(
     { virtual: true }
 );
 
+// mock labels so that we can assert the values being replaced in the string
```
psadie-sfdo, line 100: 👍

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
psadie-sfdo: Agree with extracting the `if (isError) { ... }` block.

psadie-sfdo (commented):
Just a few comments beyond the DM regarding the inline error behavior change.

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1097,25 +1085,27 @@ export default class GeFormRenderer extends LightningElement{
      * @param fieldWrapper, Array of fields with Values and Labels
      * @returns Object, helper object to minimize length of if statements and improve code legibility
      */
-    getDataImportHelper(fieldWrapper) {
+    getDataImportHelper() {
```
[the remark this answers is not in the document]
bdvorachek: Yep, that'll help with clarity here!

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -975,80 +975,65 @@ export default class GeFormRenderer extends LightningElement{
...
         if (isError) {
-            // highlight validation fields
-            this.highlightValidationErrorFields(di_record, sectionsList, ' ');
-            // set page error
-            this.hasPageLevelError = true;
-            this.pageLevelErrorMessageList = [ {
-                index: 0,
-                errorMessage: this.getDonationDonorErrorLabel(di_record, fieldWrapper)
-            } ];
+            this.displayDonorTypeError(dataImportHelper, sectionsList);
```
dfuller114, line 999: I don't think we should call this function inside of `isDonorTypeInvalid()` since this is doing more than performing validation. I would think we would call this around line 950 in the `isFormValid()` function if `isDonorTypeInvalid()` returns true. Thoughts?

dfuller114 (changes requested):
Changes look good. Had one additional comment about moving the newly extracted function call out of `isDonorTypeValid()`

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
bdvorachek: I see what you meant originally now. If `this.isDonorTypeInvalid` returns false, then we can immediately apply the validation errors in the `isFormValid` function.

dfuller114 (approved):
looks good! Thanks Ben!

psadie-sfdo (approved):
Changes look good! Approved!

dfuller114 (approved):
recent change to reset the field validity looks good to me! Thanks Ben!

PR #6661, opened 2021-09-21
[Gift Entry] Soft Credits component performance improvements

psadie-sfdo:
[W-9547800](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000QAOcYAO/view)

We're implementing a workaround to a known LockerService bug that causes a performance drop and sometimes stalls the page when arrays are passed to grand-level child components from the container component through public properties.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/lwc/geFormWidgetSoftCredit/__tests__/geFormWidgetSoftCredit.test.js
```diff
@@ -1,17 +1,21 @@
...
-const DUMMY_PROCESSED_SOFT_CREDITS = [
+let DUMMY_PROCESSED_SOFT_CREDITS = [
     { Role: 'Donor', ContactId: '4', Id: '1' },
     { Role: 'Household Member', ContactId: '5', Id: '2' },
 ];
 
 describe('c-form-widget-soft-credit', () => {
+
+    DUMMY_SOFT_CREDITS = JSON.stringify(DUMMY_SOFT_CREDITS);
+    DUMMY_PROCESSED_SOFT_CREDITS = JSON.stringify(DUMMY_PROCESSED_SOFT_CREDITS);
```
bdvorachek, line 17: I think we should probably avoid re-assignment here, we can rename either the original variable or the serialized one.

force-app/main/default/lwc/geFormWidgetSoftCredit/geFormWidgetSoftCredit.js
```diff
@@ -25,47 +25,22 @@ export default class GeFormWidgetSoftCredit extends LightningElement {
...
-
-    validate() {
-        // validate
-    }
-
-    get hasAlert() {
-        return false;
+    get softCredits() {
+        if (this.giftInView?.softCredits) {
+            return JSON.parse(this.giftInView.softCredits);
```
bdvorachek, line 30: 👍  - I'm glad this works as well as it does performance-wise but look forward to when we don't have to work around it.

force-app/main/default/lwc/geGift/geSoftCredits.js
```diff
@@ -35,6 +38,9 @@ class SoftCredits {
     }
 
     addProcessedSoftCredits(processedSoftCredits) {
+        if (typeof processedSoftCredits === 'string') {
+            processedSoftCredits = JSON.parse(processedSoftCredits);
```
bdvorachek, line 42: I think we should try to avoid reassignment in this spot and in the constructor too. 

`this._processedSoftCredits = parseIfString(processedSoftCredits);` possibly?

force-app/main/default/lwc/geGift/geSoftCredits.js
```diff
@@ -35,6 +38,9 @@ class SoftCredits {
...
+        }
         this._processedSoftCredits = [ ...processedSoftCredits ];
```
bdvorachek, line 44: I don't think we need to use the spread operator here, this could be written as `this._processedSoftCredits = processedSoftCredits;`

force-app/main/default/lwc/geFormWidgetSoftCredit/geFormWidgetSoftCredit.js
psadie-sfdo: 1000% agree

force-app/main/default/lwc/geGift/__tests__/geGift.test.js
```diff
@@ -14,10 +14,11 @@ describe('ge-gift', () => {
 
     it('should initialize with expected properties', async () => {
         const gift = new Gift(mockGiftView);
+        const giftState = gift.state();
 
-        expect(gift.state().fields).toEqual(DUMMY_GIFT_VIEW);
-        expect(gift.state().softCredits.length).toBe(0);
-        expect(gift.state().processedSoftCredits.length).toBe(0);
+        expect(giftState.fields).toEqual(DUMMY_GIFT_VIEW);
+        expect(JSON.parse(giftState.softCredits).length).toBe(0);
```
kenlewis-sfdo, line 20: Should/could there be a getter that does the parsing before returning?  "giftState.softCredits()?"

psadie-sfdo: The 'gift' object on line 16 actually has that ".softCredits()" method. That returns a "softCredits" js object with soft credits-related behavior.

My goal in this assertion is to confirm that we're able to unpack the stringified version after instantiating geGift.js and retrieving that gift's state object. Now that I'm looking at it again, it'd probably be good to have a test that has some expected number of soft credits in addition to this zero check.

kenlewis-sfdo (approved):
PR looks great Prince, thanks for these improvements!!  Doing some addt'l testing and will log any findings in Quip.

voduyemi (approved):
Thanks @psadie-sfdo. These changes look good to me!

PR #6677, opened 2021-09-29
[Gift Entry] Check field validity on value change 

psadie-sfdo:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/lwc/geFormField/geFormField.js
+++ b/force-app/main/default/lwc/geFormField/geFormField.js
@@ -376,8 +376,20 @@ export default class GeFormField extends LightningElement {
     }
 
     set formState(formState) {
-        this._formState = formState;
-        this._recordTypeId = this.recordTypeId();
+        const newValue = formState[this.sourceFieldAPIName];
+        const previousValue = this.formState && this.formState[this.sourceFieldAPIName];
+
+        new Promise((resolve,reject) => {
+            this._formState = formState;
+            this._recordTypeId = this.recordTypeId();
+            resolve();
+        })
+        .finally(() => {
+            const isValueChanged = newValue !== previousValue;
+            if (isValueChanged) {
+                this.isValid();
+            }
+        });
     }
 
     get valueFromFormState() {


force-app/main/default/lwc/geFormField/geFormField.js
```diff
@@ -376,8 +376,20 @@ export default class GeFormField extends LightningElement {
     }
 
     set formState(formState) {
```
kenlewis-sfdo, line 378: Could a jest test be added to test this?

psadie-sfdo: Yeah! Pushed up a new commit. Had to extract the block to a helper file so we can spy on it and confirm that it was called in the jest test.

force-app/main/default/lwc/geFormField/geFormFieldHelper.js
```diff
@@ -0,0 +1,17 @@
+import { isNotEmpty } from "c/utilCommon";
+
+const validityCheck = (formField) => {
+    let fieldIsValid = formField.checkFieldValidity();
+
+    if (formField.element !== null && formField.element.required) {
+        return isNotEmpty(formField.value)
+            && formField.value !== formField.CUSTOM_LABELS.commonLabelNone
```
kenlewis-sfdo, line 8: Not sure if it matters here, since this method is primarily about calling the checkFieldValidity method, but does this mean that this method returns false when someone changes a picklist back to the --None-- option?  Is that considered an invalid option for picklists?

force-app/main/default/lwc/geFormField/geFormField.js
```diff
@@ -106,16 +107,7 @@ export default class GeFormField extends LightningElement {
...
-        let fieldIsValid = this.checkFieldValidity();
-
-        if (this.element !== null && this.element.required) {
-            return isNotEmpty(this.value)
-                && this.value !== this.CUSTOM_LABELS.commonLabelNone
-                && fieldIsValid;
-        }
-
-        return fieldIsValid;
+        return validityCheck(this);
```
kenlewis-sfdo, line 110: Is the isValid method used externally somewhere?  Wondering if it needs its @api tag.

psadie-sfdo: Yeah, unfortunately, this is still used in the formSection component.

force-app/main/default/lwc/geFormField/geFormFieldHelper.js
psadie-sfdo: Yeah, '--None--' would be considered an invalid option for a required picklist. I believe we have another bug work item from a while back related to applying the custom validity/inline error for this.

Currently the user would receive a top-level error message saying something like, 'The following fields are required: Payment Method' for a required Payment Method picklist where the value is '--None--' but the field itself doesn't have an inline error.

Issue #6580, opened 2021-07-30
Gift Entry Save of New Contact with Donation Queries All Open Opportunities Without a Primary Contact


jesslopez-sf:
### Summary:
When a customer enters a new Contact with Donation information into Gift Entry and clicks the Save and New button, we do a Dry Run on the data while creating the Data Import record. If the Contact information does not match any existing Contacts, we still add the Contact Imported value (which is null in this case) to a Set of Ids called setConId. Later in the transaction, we check for matching donations in the BDI_MatchDonations class. If the size of setConId (and an Account and Opportunity Set) are all equal to 0, we bail out of the method. Otherwise, we query for all open Opportunities and their Payments where the Primary Contact is in the set of Contact Ids (in this case, null). This can lead to governor limits being hit in customer orgs where they have over 50,000 Opportunities with no Primary Contact (many times, Organization Account gifts). 

### Root Cause
When we call getPotentialDonationMatches() in BDI_MatchDonations, we add a null value to the setConId on lines 444-446:
https://github.com/SalesforceFoundation/NPSP/blob/1139b796ed082e2d1872385fa4c8fb831e8adb84/force-app/main/default/classes/BDI_MatchDonations.cls#L444-L446
In the next code block, we check to see if that set size is 0 (along with setOppId and setAccId) and if so, we return out of the function.
https://github.com/SalesforceFoundation/NPSP/blob/1139b796ed082e2d1872385fa4c8fb831e8adb84/force-app/main/default/classes/BDI_MatchDonations.cls#L453-L456
In this case, the null value results in the setConId size being 1, so we don’t return out of the function and instead continue on and run a SOQL query on all open Opportunities with Payments where the Primary Contact Id is in setConId (in this case, null).

I added some System.debug() lines to BDI_MatchDonations prior to the previous code block to prove my theory that setConId’s size was not 0. It resolve to 1 and ran a query which pulled in 10,000 Opportunities in my org.

```
// bail out if no donations to import!
        System.debug('setConId size: '+ setConId.size());
        System.debug('setCondId contents: ' + setConId);
        if (setOppId.size() == 0 && setAccId.size() == 0 && setConId.size() == 0) {
            return null;
        }
```

Debug Log Lines:
```
10:31:08.156 (1196572709)|USER_DEBUG|[454]|DEBUG|setConId size: 1
10:31:08.156 (1196644688)|USER_DEBUG|[455]|DEBUG|setCondId contents: {null}
[...]

PR #6681, opened 2021-09-30
Fixed query bug for matching donations

voduyemi:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/BDI_MatchDonations.cls
+++ b/force-app/main/default/classes/BDI_MatchDonations.cls
@@ -430,32 +430,80 @@ public with sharing class BDI_MatchDonations implements BDI_IMatchDonations {
     * @param listDI The DataImport records we are trying to match
     * @return list<Opportunity> The list of potential matching Opps.  null if none.
     */
-    private list<Opportunity> getPotentialDonationMatches(list<DataImport__c> listDI) {
+    private List<Opportunity> getPotentialDonationMatches(List<DataImport__c> listDI) {
         
         // collect all the contacts and accounts who have donations to import
-        set<Id> setConId = new set<Id>();
-        set<Id> setAccId = new set<Id>();
-        set<Id> setOppId = new set<Id>();
-        
-        for (DataImport__c di : listDI) {
+        Set<Id> contactIds = new Set<Id>();
+        Set<Id> accountIds = new Set<Id>();
+        Set<Id> opportunityIds = new Set<Id>();
+        populateQueryFilterSetsFrom(listDI, opportunityIds, accountIds, contactIds);
+
+        // bail out if no donations to import!
+        if (opportunityIds.size() == 0 && accountIds.size() == 0 && contactIds.size() == 0) {
+            return null;
+        }
+
+        BDI_PerfLogger.PerfLog pl = BDI_DataImportService.perfLogger.newPerfLog('SOQL existing Opps', BDI_PerfLogger.QUERY);
+        List<Opportunity> listOpp = Database.query(buildPotentialDonationMatchingQuery(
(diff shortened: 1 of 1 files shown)


force-app/main/default/classes/BDI_MatchDonations.cls
```diff
@@ -430,94 +430,105 @@ public with sharing class BDI_MatchDonations implements BDI_IMatchDonations {
...
+        Set<Id> setOppId,
+        Set<Id> setConId,
+        Set<Id> setAccId) {
+        for (DataImport__c di : dataImports) {
+            Contact con = bdi.contactFromDi(di, 1);
+            Account acc = bdi.AccountFromDi(di, 1);
+            // handle already been given the match!
+            if (di.DonationImported__c != null) {
+                setOppId.add(di.DonationImported__c);
+            } else if (di.Donation_Donor__c == 'Contact1' && con.Id != null) {
```
voduyemi, line 498: Hey @callen-sfdo and @pedrodalcol, I just made this change to prevent the query built for potential donation matching from filtering on null account or contact ids on dry-run through BDI. I wanted to put it on your team's radar so you are aware of this change. This change fixes this issue #6580. Open to all feedback!

force-app/main/default/classes/BDI_MatchDonations.cls
```diff
@@ -430,94 +430,105 @@ public with sharing class BDI_MatchDonations implements BDI_IMatchDonations {
...
             }
             strSoql += ' ' + UTIL_Namespace.StrTokenNSPrefix('Primary_Contact__c') + ' IN :setConId ';
             cFilter++;
         }
-        strSoql += ') ';
-        BDI_PerfLogger.PerfLog pl = BDI_DataImportService.perfLogger.newPerfLog('SOQL existing Opps', BDI_PerfLogger.QUERY);
-        list<Opportunity> listOpp = database.query(strSoql);
-        pl.stop();
-        return listOpp;           
+        strSoql += ') LIMIT 50000';
```
kenlewis-sfdo, line 483: What do you think about making this method @TestVisible and asserting that the generated query contains `LIMIT 50000`?

force-app/main/default/classes/BDI_MatchDonations.cls
```diff
@@ -513,7 +513,7 @@ public with sharing class BDI_MatchDonations implements BDI_IMatchDonations {
             strSoql += ' ' + UTIL_Namespace.StrTokenNSPrefix('Primary_Contact__c') + ' IN :setConId ';
             cFilter++;
         }
-        strSoql += ') ';
+        strSoql += ') LIMIT 50000 ';
```
kenlewis-sfdo, line 516: How about extracting lines 490-516 into a "buildQuery" method, so that it can be made @TestVisible and enable a test assertion that the LIMIT is now included in the query string?

kenlewis-sfdo (changes requested):

force-app/main/default/classes/BDI_MatchDonations.cls
voduyemi: Done!

force-app/main/default/classes/BDI_MatchDonations.cls
callen-sfdo: @voduyemi Hi Victor, thanks for looping us in on this, I just reviewed the change and I don't think it should cause any problems for us.  We actually skip most of the donation matching logic for elevate payment related DIs anyway.  cc @pedrodalcol

kenlewis-sfdo (approved):
Looks Good @voduyemi !  Thanks for adding the test!

PR #6682, opened 2021-09-30
BDI Performance Test Fix

force2b:
When deleting the default Entitlement record, also delete the related… account otherwise the check for number of records will fail

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b:
@prescod I had to make one more fix to the BDI Performance Tests since it's continued to fail. Would you mind doing a quick review/approve on this one?

PR #6365, opened 2021-04-12
Retire legacy BGE UI

psadie-sfdo:
Work Item: [W-8871417](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000096EDKIA2/view)
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# Features Intended for Future Release

# Features for Elevate Customers

# New Metadata

# Deleted Metadata

PR #6693, opened 2021-10-14
Make Gift Entry UI Display for User Locales that use Commas

### LINKED ISSUE #6125: Batch Gift Entry and Gift Entry - Unable to load Batch for Users with certain Locales
**Issue**
Users are able to create batches and access the "Batch Gift Entry" and "Gift Entry" Tabs; however, upon trying to enter a batch to edit it, or add additional Opportunities, the loading stalls and the Batch Gift Entry or Gift Entry components are not loaded.

<br>

**Cause**
When a User is attempting to access with Batch Gift Entry or Gift Entry, and the locale that is set for the user record is of a country that uses "decimal commas" (100,00) as opposed to "decimal points" (100,00), the field `npsp__Batch_Gift_Entry_Version__c` interprets `2` as `2,00` instead of the expected `2.00`. This causes an issue as the value the can be entered in the "Set Component Visibility" section of a Lightning Record Page can only be a "decimal point" number, or whole number. Attempts to enter a "decimal comma" number result in an error. 

image

Even if a whole number such as `2` is entered on the Lightning Record Page's "Set Component Visibility" section, the system will interpret the value to 2 decimal places, as the `npsp__Batch_Gift_Entry_Version__c` field is `double (4, 2)` and if the User's locale is from a country that uses "decimal commas" then Batch Gift Entry and Gift Entry attempt to intepret the `npsp__Batch_Gift_Entry_Version__c` in that locale, as it checks to see if:

1. `npsp__GiftBatch__c` = TRUE
2. `npsp__Batch_Gift_Entry_Version__c`
 a.  for Batch Gift Entry: `less than 2.0`
 b.  for Git Entry: `greater than or equal to 2.0`
 
 This correlates to the https://github.com/SalesforceFoundation/NPSP/blob/c413295aa556bc0a367daedb47bfe7f816599070/src/classes/BGE_BatchGiftEntry_UTIL.cls#L84
 
 Where the integer referenced includes `.0` which will not provide a valid argument in a "decimal comma" locale.
 
 <br>

**Steps to Reproduce**
1. Setup whichever entry product you are testing against in your org:
a.  Gift Entry - https://powerofus.force.com/s/article/NPSP-Configure-Gift-Entry
b.  Batch Gift Entry (Legacy) - https://powerofus.force.com/s/article/NPSP-Configure-Batch-Gift-Entry
2. Create a new User record
3. Change that newly created User record's locale to "Swedish (Sweden)"
a.  Note: You do **not** need to change language, currency, or timezone for the User record, only the locale
4. Login as that User
5. Attempt to access a batch record from the entry product that you enabled
6. Note that the page will not load and a blank screen persists without the respective page components loading

<br>

**Workaround**
For the affected users, change the locale on their User record to that of a country that utilizes a "decimal point" rather than a "decimal comma".

klewis-sfdc:
Remove References to the Batch Gift Entry Version field to choose UI

Old BGE and newer Gift Entry both set this field to true to distinguish
GiftBatches created via the application UI vs other sources. The Flexi-
page determined which components to display using this field, which is
set upon Batch creation. However, locales using commas break the Flexi-
page filter and the UI does not display.
(https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008lb7IIAQ/view)

Now that the older application is retired (#6365) the UI no longer 
needs to be dynamic to render two different GiftBatch UIs. 
It can simply use the Gift Entry UI for Gift Batches and the Standard UI
for non-Gift Batches.

Supersedes #6488 which was closed to update the branch name.

# Critical Changes

# Changes

# Issues Closed
Fixes #6125

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


csurname:
Hey @kenlewis-sfdo Just checking, is this one still on track to potentially be patched in 238? Only asking as we've had another case about it

jamesbsfdc:
Hey @kenlewis-sfdo  Just checking, is this one still on track to potentially be patched in 238? Only asking as we've had another case about it? 

Ignore the earlier message - I originally commented from a personal GitHub account!

PR #6718, opened 2021-10-27
SFDO Instrumentation Wrapper Services

force2b:
- Implement wrapper service classes with Enums to use for all Splunk and Argus logging from within NPSP.
- Mock (as unmanaged) the Core Splunk logging class (`SfdoLogUtils`) for instrumentation to allow for compilation.
- The actual Core class (of the same name) will only exist the context of the "npsp" namespace. To support development in an org without a namespace, the mocked class in this PR provides a simple stub interface to reflect the core class and methods. It's sole purpose is to  allow the package Apex to compile when there is no defined namespace in the org. (i.e., deploy the package code into the org).
- The `SfdoLogUtils` apex class mocked in this PR is a stub only and will not log anything.
- The `SfdoLogUtils` apex class in this PR should **never** be deployed to the packaging org.
- Compliance Workflow Action:
   - Check for any new or changed apex code that references the core sfdoLogUtil class directly instead of using the SfdoInstrumentationService classes. If found, the action will fail, assign the Instrumentation label and the PR will be assigned for review with a comment.
   - Check for any new or changed apex code that references the SfdoInstrumentationService classes. If found, the action will pass, but assign the Instrumentation label, assign the PR to a specific reviewer, and add a comment.
- Work Item: [W-9776327](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000UTO0YAO/view)

# Critical Changes

# Changes

- Implement an Instrumentation Service for NPSP to support collecting aggregate data to improve the understanding of product feature usage. More information can be found at [NPSP Feature Telemetry](https://powerofus.force.com/s/article/NPSP-Feature-Telemetry).

# Issues Closed

# Community Ideas Delivered

# New Metadata

Apex Classes:
- SfdoInstrumentationService
- SfdoInstrumentationService_TEST
- SfdoInstrumentationEnum

# Deleted Metadata


.github/workflows/compliance.yml
```diff
@@ -0,0 +1,140 @@
...
+
+            # Find references to "Sfdo" in the force-app/ folder structure. Filters for the specific class references in later steps.
+            export DIFF=$( git diff-tree origin/$GITHUB_BASE_REF..$GITHUB_HEAD_REF --patch-with-raw -- force-app/** | grep 'Sfdo' )
+          fi
+
+          # Assign the "git diff-tree" output to an internal workflow var to use later
+          # Escape newlines (replace \n with %0A) so it's a single line of text
+          echo "::set-output name=diff::$( echo "$DIFF" | sed ':a;N;$!ba;s/\n/%0A/g' )"
+
+          # echo "Diff Output = ${{ steps.get_pr_diff.outputs.diff }}"
```
jstvz, line 76: Appears unused:
```suggestion
```

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Get PR Owner
+        id: pr_owner
+        run: |
+          echo "::set-output name=owner::${{ github.event.pull_request.user.login }}"
+
+      - name: Set Reviewer
+        id: set_reviewer
+        run: |
+          if [${{ steps.pr_owner.outputs.owner != env.DEFAULT_REVIEWER}}]
```
jstvz, line 44: This won't quite work, since it'll be treated as a command. I suggest:
```suggestion
          if [[ "${{ steps.pr_owner.outputs.owner != env.DEFAULT_REVIEWER}}" == "true" ]]
```

(Comparing `owner` and env var directly can also work.)

jstvz (changes requested):
Couple of tweaks needed, but looks good overall.

.github/workflows/compliance.yml
force2b: @jstvz  I think this was working though. There were points where I tested changes and Luke was assigned as the reviewer when I was the PR owner.

jstvz: Ah, I see the problem. I'm seeing this in the [log](https://github.com/SalesforceFoundation/NPSP/runs/4216032180?check_suite_focus=true#step:4:13):
```
/home/runner/work/_temp/5b944cf8-50e2-4dfe-b4e3-12a048928657.sh: line 1: [false]: command not found
```

Because the test-command returns a non-zero exit code when `command not found`, the so the else clause is always executed. For example:
```console
bash-3.2$ if [false]; then echo 'false'; else echo 'true';fi
bash: [false]: command not found
true
```

force-app/main/instrumentation/classes/SfdoInstrumentationEnum.cls
```diff
@@ -0,0 +1,130 @@
...
+        Affiliations,
+        SoftCreditAutomation,
+        MatchingGifts,
+        PartialSoftCredits,
+        CustomizableRollups,
+        LegacyRollups,
+        PaymentServices,
+        LeadConversion,
+        RecordMerge
+    }
```
npsp-reedestockton, line 68: @force2b Where would things like CallableAPI and Elevate setup/usage fall into this list?  Maybe no need to think about it at this point since it can be easily modified.

force2b: We can add features to this list at any point. The CallableApi stuff I think we'd just put under the related feature, unless we wanted to explicitly instrument usage of the Callable Api's. In that case, then maybe we'd add a new Feature for it or maybe just one for Api in general. Elevate would probably get added as well - especially if we wanted to capture data on outbound api calls for authentication/charging/etc.

.github/workflows/compliance.yml
force2b: Thanks. Just to confirm - the line does fail, but it still works; which means I should make that change?

jstvz: Yes. This is a "stopped clock"; you could transpose the values for default and alternate reviewers and the alternate would still be selected.

cumulusci.yml
```diff
@@ -836,6 +845,13 @@ flows:
                     api_names: QA,Gift_Entry
 
 
+    deploy_unmanaged:
+        description: Override to the deploy_unmanaged flow to force a mock apex class to be deployed that will prevent compilation errors when the packaged code references file-based apex in the core stack for instrumentation purposes.
+        steps:
+            2.1:
+                when: not org_config.namespace
+                task: deploy_core_mock_classes
```
andrewyu-salesforce, line 853: The condition is correct and skip the mock class deployment. However, in the general flow `step 1) depenedices` `task 2: deploy_pre`, all files in the `unpakcaged/pre` are being deployed. This result `SfdoLogUtils.cls` being deployed regardless of the namespace.

andrewyu-salesforce (approved):

dfuller114 (approved):

force2b:
Replaced by #6792

PR #6727, opened 2021-11-01
Check for Field Permission When Adding Undeliverable Address to Contact Merge Page

dfuller114:
Add field-level security checking before adding the undeliverable address field to the contact merge page.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/CON_ContactMerge_CTRL.cls
+++ b/force-app/main/default/classes/CON_ContactMerge_CTRL.cls
@@ -712,16 +712,20 @@ public with sharing class CON_ContactMerge_CTRL {
 
     private List<String> buildSortedCustomFieldList(Map<String, DescribeFieldResult> customFieldMap) {
         List<String> customFields = new List<String>();
-        String undeliverableFieldName = String.valueOf(Contact.Undeliverable_Address__c).toLowerCase();
+        DescribeFieldResult undeliverableFieldDescribeResult = Contact.Undeliverable_Address__c.getDescribe();
 
         for (String fieldName : customFieldMap.keySet()) {
-            if (fieldName == undeliverableFieldName) {
+            if (fieldName == undeliverableFieldDescribeResult.getName()) {
                 continue;
             }
             customFields.add(fieldName);
         }
+
         customFields.sort();
-        customFields.add(0, undeliverableFieldName);
+        if (UTIL_Permissions.canRead(undeliverableFieldDescribeResult, false) &&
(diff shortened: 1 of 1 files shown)


lparrott (approved):
I confirmed that the Contact Merge page now loads, but without the Undeliverable Address row, if the User does not have Update access to that field. Nice work!

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -712,16 +712,20 @@ public with sharing class CON_ContactMerge_CTRL {
 
     private List<String> buildSortedCustomFieldList(Map<String, DescribeFieldResult> customFieldMap) {
         List<String> customFields = new List<String>();
-        String undeliverableFieldName = String.valueOf(Contact.Undeliverable_Address__c).toLowerCase();
+        DescribeFieldResult undeliverableFieldDescribeResult = Contact.Undeliverable_Address__c.getDescribe();
```
force2b, line 715: I think what I'm confused about with this is why it's necessary to check for this specific field? Couldn't there be other fields that a given user doesn't have perms to that would then throw the same exception?

npsp-reedestockton (approved):
Looks good to me, Daniel.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
dfuller114: The reason we are having to handle this field uniquely is because we had a product requirement to display the `Undeliverable_Address__c` field at the top of the custom fields section of the page so that it's underneath the other contact address info. This page uses the order of the fields returned from the object getMap() describe call so the order is already predetermined. That's what we are adding the undeliverable field first to `customFields` in `buildSortedCustomFieldList()`. However, what we missed is that we have existing logic on line 568 that only places the describe fields from contact in the collection returned to the page if the user has read and update access, but we were not doing that for Undeliverable_Address__c when we added it separately. We were just added the field to the `customFields` list without checking for access first.

So the change here is to keep the behavior consistent with the rest of the custom fields in the page so that if the user doesn't have access to the field, it won't display on the page.

force2b: Thank you for the explanation. Makes sense.

force2b (approved):

sam-knox:
Do we need a doc update to Contact Merge config that says you need FLS on the Undeliverable field? If so, just the Contact one, or all three?

PR #6743, opened 2021-11-16
W-10114645 - Donation History dataset

JoseRossiSF:
# Work Item
[W-10114645](https://gus.lightning.force.com/a07EE00000LKpruYAD)

# Changes
Adding records to dev_org test_data.sql dataset in order to include over seventy Opportunities related to Hard Credits, Soft Credits, and Partial Soft Credits for the Contact "Alice Patterson".

The goal was to create a dataset that can be used by Dev and QA in the development of Donation History Table community feature.


force2b:
@JoseRossiSF I'm curious - is there a reason you didn't create a Snowfakery recipe to generate the test data? Plus, if this data is very specific to Donation History testing, it should be a new recipe vs. modifying one that could be used by other teams.

JoseRossiSF:
> @JoseRossiSF I'm curious - is there a reason you didn't create a Snowfakery recipe to generate the test data? Plus, if this data is very specific to Donation History testing, it should be a new recipe vs. modifying one that could be used by other teams.

@force2b Manually creating the records seemed like a more straightforward way to accomplish our goal, taking in consideration the records we required were quite specific and not too numerous.
Regarding why I am modifying an existing Dataset instead of creating a new one, that was originally my first approach, but was advised that, given there's plenty of existing datasets, it would be better to simply add to an existing one the records which we need. If you believe a different Dataset should be created instead, let me know and I will make the required changes.

dcaballeroSf (approved):
LGTM

mcasco (approved):
LGTM

trigger_status.yml
```diff
@@ -0,0 +1,4 @@
+Opportunity:CRLP_Rollup_TDTM: true
```
force2b, line 1: This file needs to be removed from the PR

force2b (changes requested):
One change needed, but overall I do feel that the better solution here would have been to create a new snowfakery script that built data exactly as your product needs it. The SQL file modified here is pretty old and has never been updated or brought up to date with any of the more recent changes to NPSP.

trigger_status.yml
JoseRossiSF: Agreed, thanks for pointing it out

force2b (approved):
Overall this will probably work ... at least for this year. Because the dates are hardcoded into the SQL file the data is very dated. This is another primary advantage of Snowfakery. The dates be automatically created as relative to the current date.

psadie-sfdo (approved):

PR #6749, opened 2021-11-22
Fix for Manage HH Page when Undeliverable Address permissions are missing

### LINKED ISSUE #6733: Manage Household Page Fails to Load If Customer Does Not Have Access to New Undeliverable Address Fields
### Summary
Customers are unable to view the **_Manage Households_** page after receiving the NPSP 3.209 update (released to sandboxes on November 4 and Production on November 16). This is due to a field access issue on 3 new fields introduced in the update. The page shows a never-ending spinner, with an Account error vaguely visible in the background.

### Workaround
While we address this within the product, please grant access to the following fields:

- **Account object**: Read and Edit access to Undeliverable Billing Address (npsp__Undeliverable_Address__c)
- **Contact object**: Read and Edit access to Undeliverable Mailing Address (npsp__Undeliverable_Address__c)
- **Address object:** Read access to Undeliverable (npsp__Undeliverable__c)

### Screenshot
Screen Shot 2021-11-05 at 11 51 44 AM

lparrott:
# Critical Changes

# Changes

- We fixed an issue with the Manage Household page where the page wouldn't load without permissions to Undeliverable Address fields.

# Issues Closed

Fixes #6733 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
+++ b/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
@@ -476,7 +476,10 @@
         con.MailingState = addr.MailingState__c;
         con.MailingPostalCode = addr.MailingPostalCode__c;
         con.MailingCountry = addr.MailingCountry__c;
-        con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+        // If Undeliverable Address information was passed from Apex for both objects, show the status
+        if(con.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
+            con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+        }
     },
 
     /*******************************************************************************************************
--- a/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
+++ b/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
@@ -442,9 +442,7 @@
             hh.BillingState = addr.MailingState__c;
             hh.BillingPostalCode = addr.MailingPostalCode__c;
             hh.BillingCountry = addr.MailingCountry__c;
-            if(hh.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
-                hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
-            }
+            hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
         } else {
             hh.npo02__MailingStreet__c = addr.MailingStreet__c;
             if (addr.MailingStreet2__c)
--- a/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
+++ b/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
@@ -440,7 +440,9 @@
             hh.BillingState = addr.MailingState__c;
             hh.BillingPostalCode = addr.MailingPostalCode__c;
             hh.BillingCountry = addr.MailingCountry__c;
-            hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+            if(hh.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
+                hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+            }
         } else {
             hh.npo02__MailingStreet__c = addr.MailingStreet__c;
             if (addr.MailingStreet2__c)
@@ -474,7 +476,9 @@
         con.MailingState = addr.MailingState__c;
         con.MailingPostalCode = addr.MailingPostalCode__c;
         con.MailingCountry = addr.MailingCountry__c;
-        con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+        if(con.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
+            con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+        }
     },
 
     /*******************************************************************************************************
--- a/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
+++ b/force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
@@ -478,9 +478,7 @@
         con.MailingState = addr.MailingState__c;
         con.MailingPostalCode = addr.MailingPostalCode__c;
         con.MailingCountry = addr.MailingCountry__c;
-        if(con.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
-            con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
-        }
+        con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
     },
 
     /*******************************************************************************************************
--- a/force-app/main/default/classes/HH_Container_LCTRL.cls
+++ b/force-app/main/default/classes/HH_Container_LCTRL.cls
@@ -325,7 +325,7 @@ public with sharing class HH_Container_LCTRL {
     * @return void
     */
     private static void addContactUndeliverableAddressField(Set<String> fieldNames) {   
-        if (canReadContactUndeliverable) {
+        if (canUpdateAddressUndeliverable && canReadContactUndeliverable) {
             fieldNames.add(undeliverableAddressField);
         }
     }
--- a/force-app/main/default/classes/HH_Container_LCTRL.cls
+++ b/force-app/main/default/classes/HH_Container_LCTRL.cls
@@ -448,8 +448,9 @@ public with sharing class HH_Container_LCTRL {
             UTIL_Describe.copyObjectFLS(strObject, hh, hhCopy, new List<String>(householdFieldsToEdit));
 
             // Update Undeliverable field on Account to ensure correct syncing
-            hhCopy.put(UTIL_Namespace.StrTokenNSPrefix('Undeliverable_Address__c'), selectedAddressIsUndeliverable);
-            
+            if (isHHAccount) {
+                hhCopy.put(UTIL_Namespace.StrTokenNSPrefix('Undeliverable_Address__c'), selectedAddressIsUndeliverable);
+            }
             // we've already saved the data for the contacts in the household.
             // now we only need to save the account fields an update, and we don't want
             // any of the normal account trigger work to run again.  specifically,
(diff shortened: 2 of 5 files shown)


force-app/main/default/classes/HH_Container_LCTRL.cls
```diff
@@ -514,12 +535,16 @@ public with sharing class HH_Container_LCTRL {
...
+            UTIL_Describe.getFieldLabel(String.valueOf(Contact.SObjectType), String.valueOf(Contact.Undeliverable_Address__c)) : '';
+        String undeliverableLabel = canReadAddressUndeliverable ?
+            UTIL_Describe.getFieldLabel(String.valueOf(Address__c.SObjectType), String.valueOf(Address__c.Undeliverable__c)) : '';
+
         Map<String, String> fieldLabelsBySObject = new Map<String, String> {
-            String.valueOf(Contact.Undeliverable_Address__c) => UTIL_Describe.getFieldLabel(
-                    String.valueOf(Contact.SObjectType), String.valueOf(Contact.Undeliverable_Address__c)),
-            String.valueOf(Address__c.Undeliverable__c) => UTIL_Describe.getFieldLabel(
-                    String.valueOf(Address__c.SObjectType), String.valueOf(Address__c.Undeliverable__c))
+            'Undeliverable_Address__c' => undeliverableAddressLabel,
```
lparrott, line 544: I changed this to avoid additional permissions checking in the front end, the Label will just be blank in cases where permissions are missing.

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -440,7 +440,9 @@
             hh.BillingState = addr.MailingState__c;
             hh.BillingPostalCode = addr.MailingPostalCode__c;
             hh.BillingCountry = addr.MailingCountry__c;
-            hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+            if(hh.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
+                hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
```
force2b, line 444: Does this need to be set to a default value, or is leaving the value _as is_ the goal?

force-app/main/default/classes/HH_Container_LCTRL.cls
```diff
@@ -615,8 +640,10 @@ public with sharing class HH_Container_LCTRL {
                 address.MailingState__c = (String)household.get('BillingState');
                 address.MailingPostalCode__c = (String)household.get('BillingPostalCode');
                 address.MailingCountry__c = (String)household.get('BillingCountry');
-                address.Undeliverable__c = (Boolean)household.get(
-                    UTIL_Namespace.StrTokenNSPrefix('Undeliverable_Address__c'));
+                if (canReadAddressUndeliverable && canReadAccountUndeliverable) {
```
force2b, line 643: Does this check (and the ones below) also need for `canUpdate`?

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
lparrott: In the case where we don't have permissions, this attribute won't exist on the object, so not changing it felt like the best approach.

force-app/main/default/classes/HH_Container_LCTRL.cls
lparrott: During my testing, I didn't see a case where the User can explicitly set the Undeliverable checkbox through this page. If changing an Address causes a record to be set as Undeliverable, that code should run without sharing via the TDTM trigger, not in the Manage Household logic.

force2b: Interesting. If the page doesn't let the user modify the field value, why does the controller allow updating the field value?

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
force2b: Makes sesnse.

force-app/main/default/classes/HH_Container_LCTRL.cls
lparrott: Ha, I walked away and thought the same thing. Since the Apex is updating values, the User will need that Perm as well. I can't speak to the "why", maybe some of this can be rewritten now that we're allowing use of the form without Undeliverable Address permissions, this first pass was just to get it working.

force2b (commented):
What do you think about adding a Unit test for the new condition?

force-app/main/default/classes/HH_Container_LCTRL.cls
```diff
@@ -38,7 +38,15 @@ public with sharing class HH_Container_LCTRL {
 
     /** @description our exception object for Field Level & Object Security errors. */
     private class permsException extends Exception {}
-    
+
+    private static Boolean canReadContactUndeliverable =
```
npsp-reedestockton, line 42: Don't we need to add namespace for the field name on these calls to canRead?

force-app/main/default/classes/HH_Container_LCTRL.cls
```diff
@@ -125,8 +133,10 @@ public with sharing class HH_Container_LCTRL {
         if (isHousehold) {
             householdFieldsToEdit.addAll(new Set<String>{
                 'BillingStreet', 'BillingCity', 'BillingState', 'BillingPostalCode',
-                 String.valueOf(Account.Undeliverable_Address__c), 'BillingCountry',
-                'BillingLatitude', 'BillingLongitude'});
+                'BillingCountry', 'BillingLatitude', 'BillingLongitude'});
+            if (canReadAccountUndeliverable) {
+                householdFieldsToEdit.add(String.valueOf(Account.Undeliverable_Address__c));
```
npsp-reedestockton, line 138: won't this fail in a namespaced org?

force-app/main/default/classes/HH_Container_LCTRL.cls
```diff
@@ -514,12 +536,16 @@ public with sharing class HH_Container_LCTRL {
 
     @AuraEnabled
     public static Map<String, String> getFieldLabels() {
+        String undeliverableAddressLabel = canReadContactUndeliverable ?
```
npsp-reedestockton, line 539: Don't we need namespace here as well?

npsp-reedestockton (changes requested):

force-app/main/default/classes/HH_Container_LCTRL.cls
lparrott: Since it's using a direct reference to the Field Describe, I think it's fine! This is how Daniel did it on the Contact Merge page, no issues there.

force-app/main/default/classes/HH_Container_LCTRL.cls
lparrott: In this case, we're stripping the namespace in the Aura component anyways, so this should work fine.

force-app/main/default/classes/HH_Container_LCTRL.cls
lparrott: It shouldn't, we're referencing the Object.Fieldname directly instead of storing it as a String.

npsp-reedestockton (approved):

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -474,7 +476,9 @@
         con.MailingState = addr.MailingState__c;
         con.MailingPostalCode = addr.MailingPostalCode__c;
         con.MailingCountry = addr.MailingCountry__c;
-        con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+        if(con.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
+            con.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
```
bdvorachek, line 480: In the event that a user doesn't have access/permissions to the Undeliverable field, its value wouldn't get copied here. Wanted to double check to make sure that we're ok with that here?

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
```diff
@@ -440,7 +440,9 @@
             hh.BillingState = addr.MailingState__c;
             hh.BillingPostalCode = addr.MailingPostalCode__c;
             hh.BillingCountry = addr.MailingCountry__c;
-            hh.Undeliverable_Address__c = addr.Undeliverable__c === undefined ? false : addr.Undeliverable__c;
+            if(hh.hasOwnProperty('Undeliverable_Address__c') && addr.hasOwnProperty('Undeliverable__c')){
```
force2b, line 443: This being javascript, I have a feeling that a namespace is needed on the field names in quotes. I remember bumping into this kind of issue in the CRLP Config pages. It's worth testing though, if there's a way to do that.

force-app/main/default/classes/HH_Container_LCTRL.cls
force2b: Apex is smart enough to handle the namespace dynamically 99.99% of the time.

dfuller114:
@lparrott I just pushed https://github.com/SalesforceFoundation/NPSP/pull/6749/commits/e6a06dbf68c78e8c8ba32192b33d49a40647e476 to address the spinner not clearing when the permissions error is thrown. FYI @bdvorachek I believed you flagged this when we first looked at this. This should now at least keep the page from being inaccessible whenever the user doesn't correct FLS.

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
lparrott: Normally yes, but we use a method `removePrefixFromObjectFields` when loading the household (and the other sobjects) that should allow us to safely ignore the namespace.

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
lparrott: Yeah, I'm questioning this now, but the update call on Contact will fail is this field gets changed without permissions, so we'd need to include Update access to the Undeliverable field on Contact in our permissions check.

force-app/main/default/classes/HH_Container_TEST.cls
```diff
@@ -662,6 +662,53 @@ private class HH_Container_TEST {
...
+        List<FieldPermissions> fieldPermissions = [
+            SELECT Id, PermissionsRead, PermissionsEdit
+            FROM FieldPermissions
+            WHERE Parent.ProfileId = :adminProfileId
+            AND (Field = :addressUndeliverableField
+                OR Field =: contactUndeliverableField
+                OR Field = : accountUndeliverableField
+            )
+        ];
+        for(FieldPermissions fieldPermission : fieldPermissions){
```
force2b, line 687: Is it worth testing some other combinations, such as:
- Edit Access to the Address.Undeliverable field, but only Read to the Account/Contact fields
- Read Access to the Address.Undeliverable field, but no to the Account/Contact fields
- No Access to the Address.Undeliverable field, but Edit to the Account/Contact fields

There's obviously a lot of permutations. The question is how many of them we want to test to be sure this isn't going to still generate unexpected errors for customers.

lparrott: I added one additional test that retains Account.Undeliverable access, since I needed to query that field as part of the test, I'm not sure testing every combination is worth it here, unless we find issues.

bdvorachek: We're more concerned with data integrity than unexpected errors in this situation. The problem we were seeing is that the Undeliverable field would get set to an unexpected value if you didn't have permissions to any of the fields, because we relied on field visibility to drive the automated updates.

bdvorachek (approved):

force-app/main/default/aura/HH_Container/HH_ContainerHelper.js
kenlewis-sfdo:
```suggestion
                hh.Undeliverable_Address__c = !!addr.Undeliverable__c;
```

dfuller114 (approved):
looks good. thanks Luke!

PR #6758, opened 2021-11-30
Saving a Recurring Donation Should Redirect to the Newly Created Record

dfuller114:
Any time a new recurring donation is created using the "New" button, we will redirect the user to the newly created recurring donation record. 

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/aura/RD2_EntryForm/RD2_EntryFormHelper.js
```diff
@@ -1,17 +1,4 @@
...
-    /**
-     * @description: Create the modal when initializing the component
-     */
-    handleShowModal: function (component, event, helper) {
-        let recordId = component.get('v.recordId');
-        let parentId;
-
-        if (!recordId) {
-            parentId = helper.getParentId();
-            component.set('v.parentId', parentId);
```
bdvorachek, line 11: I think we need to keep this assignment statement (and the inContextOfRef code), as it is responsible for setting the Contact/Account lookup in the RD2 Entry Form Modal when opening it from a Related List.

bdvorachek (changes requested):

force-app/main/default/aura/RD2_EntryForm/RD2_EntryFormHelper.js
dfuller114: reverted the commit

force-app/main/default/aura/RD2_EntryForm/RD2_EntryFormHelper.js
```diff
@@ -23,7 +23,7 @@
      * @description: Redirect the page to either parent or RD record
      */
     redirectToPage: function (component, event) {
-        let navigateToId = event.getParams('detail').recordId;
+        let navigateToId = event.getParams('detail').recordId || component.get('v.parentId');
```
dfuller114, line 26: Flipping this logic will allow us to redirect back to the parent when there is no `recordId` such as when the user clicks the `cancel` button

bdvorachek (approved):

andrewyu-salesforce (approved):

PR #6760, opened 2021-12-01
W-10061529 - Donation History List Condition Class

mcasco:
# Critical Changes
List all donations. What donations should display is defined by a set of conditions:
- Opportunity are Closed/won
- Includes Hard Credit, Soft Credit and Partial Soft Credit data
- We should never show open or Closed/lost opportunities

# Changes


mcasco:
[W-10061529](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000I3OpsYAF/view)

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+        List<DonationHistory> hardCredit = verifyHardCreditCondition(donations);
+        donationHistory.addAll(hardCredit);
+
+        List<DonationHistory> softCredit = verifySoftCreditCondition(donations);
+        donationHistory.addAll(softCredit);
+
+        List<Id> donationsId = filterVerifiedIds(donationHistory, donations);
+        List<DonationHistory> partialSoftCredit = verifyPartialSoftCreditConsition(donationsId, donations);
+        donationHistory.addAll(partialSoftCredit);
+        System.debug(donationHistory);
```
JoseRossiSF, line 33: please remove System.debug

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+        List<DonationHistory> softCreditList = new List<DonationHistory>();
+        for(OpportunityContactRole opp: donations){
+            if (softCreditRoles.contains(opp.Role)) {
+                DonationHistory softCredit = new DonationHistory();
+                softCredit.id = opp.Id;
+                softCredit.name = opp.Opportunity.Name;
+                softCredit.closeDate = opp.Opportunity.CloseDate;
+                softCredit.role = opp.Role;
+                softCredit.amount = opp.Opportunity.Amount;
+                softCredit.creditType = HARD_CREDIT;
```
JoseRossiSF, line 67: I am confused as if to this should be `softCredit.creditType = SOFT_CREDIT;`

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+        Set<Id> resultIds = (new Map<Id,Partial_Soft_Credit__c>(partialSoftCredit)).keySet();
+        List<DonationHistory> partialSoftCreditList = new List<DonationHistory>();
+        for(OpportunityContactRole opp: donations){
+            if (resultIds.contains(opp.id)) {
+                DonationHistory newPartialSoftCredit = new DonationHistory();
+                newPartialSoftCredit.id = opp.Id;
+                newPartialSoftCredit.name = opp.Opportunity.Name;
+                newPartialSoftCredit.closeDate = opp.Opportunity.CloseDate;
+                newPartialSoftCredit.role = opp.Role;
+                newPartialSoftCredit.creditType = HARD_CREDIT;
```
JoseRossiSF, line 103: Likewise, should this not be `newPartialSoftCredit.creditType = PARTIAL_SOFT_CREDIT;` ?

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+        for(OpportunityContactRole opp: donations){
+            if (!donationHistoryList.contains(opp.id)) {
+                donationsId.add(opp.Id);
+            }
+        }
+        return donationsId;
+    }
+
+    public static List<DonationHistory> verifyPartialSoftCreditConsition(List<Id> donationsId, List<OpportunityContactRole> donations){
+        System.debug(donationsId);
```
JoseRossiSF, line 90: remove System.debug

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+        }
+        for(OpportunityContactRole opp: donations){
+            if (!donationHistoryList.contains(opp.id)) {
+                donationsId.add(opp.Id);
+            }
+        }
+        return donationsId;
+    }
+
+    public static List<DonationHistory> verifyPartialSoftCreditConsition(List<Id> donationsId, List<OpportunityContactRole> donations){
```
JoseRossiSF, line 89: typo, method should be named _verifyPartialSoftCreditCondition_

JoseRossiSF (changes requested):
Looks good overall. However, please look into my comments to see if modifications are needed.
Also, add Apex Doc comments

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,111 @@
...
+                softCreditList.add(softCredit);
+            }
+        }
+        return softCreditList;
+    }
+
+    public static List<Id> filterVerifiedIds(List<DonationHistory> donationHistory, List<OpportunityContactRole> donations){
+        List<Id> donationsId = new List<Id>();
+        List<Id> donationHistoryList = new List<Id>();
+        //Set<String> donationHistorySet = (new Map<String, DonationHistory>(donationHistory)).keySet();
```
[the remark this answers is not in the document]
mcasco: removed

force-app/main/default/classes/ListDonation.cls
mcasco: fixed

force-app/main/default/classes/ListDonation.cls
mcasco: removed

force-app/main/default/classes/ListDonation.cls
mcasco: fixed

force-app/main/default/classes/ListDonation.cls
mcasco: fixed

force-app/main/default/classes/ListDonation.cls
mcasco: removed

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,144 @@
...
+        public Decimal amount { get; set; }
+        //public String paymentMethod { get; set; }
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    public static void getDonationHistory(){
```
JoseRossiSF, line 27: Should this method be using "@AuraEnabled" ?

JoseRossiSF (changes requested):
Besides the getDonationHistory being AuraEnabled, please note Apex Doc comments for the methods are missing the parameter and return clarification. Please add these and it should be good to go

force-app/main/default/classes/ListDonation.cls
mcasco: yes, added

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    @AuraEnabled
+    public static List<DonationHistory> getDonationHistory(Id contactId){
+        List<OpportunityContactRole> donations = [SELECT Id, Opportunity.Name, Opportunity.CloseDate, Opportunity.Amount, Contact.Email, Contact.Name, Role, isPrimary FROM OpportunityContactRole WHERE Opportunity.StageName = 'Closed Won' AND ContactId =: contactId order by Opportunity.CloseDate desc Limit 50 ];
```
bdvorachek, line 30: This query should probably use `Opportunity.IsWon = TRUE` instead of `Opportunity.StageName = 'Closed Won'`, since there could be more than one stage in a customer's org that is set up to be a Closed/Won stage.

force-app/main/default/classes/ListDonation_TEST.cls
```diff
@@ -0,0 +1,53 @@
...
+		insert opportunitySoftCredit;
+		insert opportunityPartialSoftCredit;
+
+		
+		OpportunityContactRole oppCntRoleSoftCredit = new OpportunityContactRole(Role='Soft Credit', OpportunityId=opportunitySoftCredit.Id, ContactId=mockDefaultContact.Id, isPrimary=false);
+		OpportunityContactRole oppCntRolePartialSoftCredit = new OpportunityContactRole(Role='Soft Credit', OpportunityId=opportunityPartialSoftCredit.Id, ContactId=mockDefaultContact.Id, isPrimary=false);
+		insert oppCntRoleSoftCredit;
+		insert oppCntRolePartialSoftCredit;
+        
+        Partial_Soft_Credit__c partialSoftCredit = new Partial_Soft_Credit__c
```
bdvorachek, line 31: Some of the formatting here gained a bit of whitespace, can we get this Partial Soft Credit constructor to be reformatted to not be shifted so far to the right?

force-app/main/default/classes/ListDonation_TEST.cls
```diff
@@ -0,0 +1,53 @@
...
+
+        Account mockDefaultAccount = new Account(Name = 'Default Account');
+        insert mockDefaultAccount;
+
+		Contact mockDefaultContact = new Contact(FirstName='Alice', LastName='Patterson', AccountId=mockDefaultAccount.Id);
+		Contact secondaryContact = new Contact(FirstName='Josh', LastName='Patterson', AccountId=mockDefaultAccount.Id);
+		insert mockDefaultContact;
+		insert secondaryContact;
+
+		Opportunity opportunityHardCredit = new Opportunity(Name='HardCredit1', AccountId=mockDefaultAccount.Id, Primary_Contact__c=mockDefaultContact.Id, Amount=400, StageName='Closed Won', CloseDate=myDateTime);
```
bdvorachek, line 18:
```suggestion
		Opportunity opportunityHardCredit = new Opportunity(
		    Name='HardCredit1', 
		    AccountId=mockDefaultAccount.Id, 
		    Primary_Contact__c=mockDefaultContact.Id, 
		    Amount=400, 
		    StageName='Closed Won', 
		    CloseDate=myDateTime
		);
```

Suggestion for readability on these opportunities that are being constructed here.

force-app/main/default/classes/ListDonation_TEST.cls
```diff
@@ -0,0 +1,53 @@
...
+
+		Test.stopTest();
+		
+		List<ListDonation.DonationHistory> donationHistory = ListDonation.getDonationHistory(mockDefaultContact.Id);
+		List<ListDonation.DonationHistory> filteredDonationHistory = new List<ListDonation.DonationHistory>();
+        
+		
+		for (ListDonation.DonationHistory donation : donationHistory) {
+        	filteredDonationHistory.add(donation);
+			System.debug('donation: '+donation);
```
bdvorachek, line 46: Make sure we trim this debug statement out of the final version of this.

bdvorachek (changes requested):

force-app/main/default/classes/ListDonation.cls
psadie-sfdo: Additionally, could we add line breaks to this and uppercase 'order by' and 'desc'?

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,166 @@
+public with sharing class ListDonation {
```
psadie-sfdo, line 1: Copyright block needed for this class.

force-app/main/default/classes/ListDonation_TEST.cls
```diff
@@ -0,0 +1,53 @@
+@IsTest
+public class ListDonation_TEST {
+
+    @IsTest
+    public static void testGetDonationHistory() {
```
psadie-sfdo, line 5: Could fix the tabbing in this file? Some lines like line 8 are tabbed one or two too far.

psadie-sfdo (changes requested):

JoseRossiSF (approved):
LGTM

force-app/main/default/lwc/donationHistory/donationHistory.js
```diff
@@ -1,5 +1,5 @@
 import { LightningElement, api } from 'lwc';
 
 export default class DonationHistory extends LightningElement {
-    @api recordId;
+    @api recordId = '0035400000l70bkAAA';
```
bdvorachek, line 4: Hardcoded ID here that should probably be removed.

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -26,26 +27,32 @@ export default class DonationHistoryTable extends LightningElement {
...
+            this.data = data;
+        }
+    }
+
     
     
     totalNumberOfRows = MOCK_DATA.length;
     
     // eslint-disable-next-line @lwc/lwc/no-async-await
     async connectedCallback() {
```
bdvorachek, line 55: We can remove this function entirely if there's nothing in the body of it.

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -26,26 +27,32 @@ export default class DonationHistoryTable extends LightningElement {
...
+            { label: donorLabel, fieldName: 'name', type: 'text' },
             { label: commonAmount, fieldName: 'amount', type: 'currency', },
             { label: this.paymentMethodLabel, fieldName: 'paymentMethod', type: 'text', },
         ];
     }
 
+    @wire(getDonationHistory, { contactId: '$contactId' })
+    getDonation({ data, error }) {
+        if (data) {
+            console.log(JSON.stringify(data));
```
bdvorachek, line 45: This console.log can be removed.

bdvorachek (changes requested):

force-app/main/default/classes/ListDonation_TEST.cls
psadie-sfdo: The indentation in this file still has a mixture of tabs and spaces. Let's update it. We generally use 4 spaces = 1 tab.

psadie-sfdo (changes requested):
Just one request left for me related to the ListDonation_TEST file. Thanks!

bdvorachek (approved):

JoseRossiSF (approved):

force-app/main/default/classes/ListDonation.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
force2b, line 3: All new classes and components should use api v53.

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,215 @@
...
+    * @description method to return Donation History list.
+    */
+    @AuraEnabled(cacheable=true)
+    public static List<DonationHistory> getDonationHistory(Id contactId, Integer offset){
+        List<OpportunityContactRole> donations = [SELECT Id, Opportunity.Name, Opportunity.CloseDate, Opportunity.Amount, Contact.Email, Contact.Name, Role, isPrimary 
+                                                    FROM OpportunityContactRole 
+                                                    WHERE Opportunity.IsWon = true 
+                                                    AND ContactId =:contactId 
+                                                    ORDER BY Opportunity.CloseDate 
+                                                    DESC LIMIT 50 OFFSET :offset];
```
force2b, line 70: Please keep the [max of 2000 rows](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_offset.htm) in mind when using `OFFSET`. This may require a "known limitation" statement that the page cannot show more than 2000 gifts or some extra logic to prevent a crash. It's probably unlikely that a single donor would have 2000 related OCR records, but it's not impossible either. Would be good to alert Ali as well.

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,215 @@
...
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify soft credit condition.
+    */
+    public static List<DonationHistory> verifySoftCreditCondition(List<OpportunityContactRole> donations){
+        //with the list of donations verify if the Role match the filter rule value on the NPSP_ClosedWon_Opps_SC filter group
+        Filter_Rule__mdt roles = [SELECT Constant__c FROM Filter_Rule__mdt WHERE QualifiedApiName = 'NPSP_OppsSoftCreditRoles'];
+        List<String> softCreditRoles = roles.Constant__c.split(';');
+        List<DonationHistory> softCreditList = new List<DonationHistory>();
+        for(OpportunityContactRole opp: donations){
```
force2b, line 137:
```suggestion
        for(OpportunityContactRole ocr: donations){
```

force-app/main/default/classes/ListDonation_TEST.cls
```diff
@@ -0,0 +1,86 @@
...
+
+		Partial_Soft_Credit__c partialSoftCredit = new Partial_Soft_Credit__c (
+			Amount__c=50,
+			Contact__c=mockDefaultContact.Id,
+			Opportunity__c=opportunityPartialSoftCredit.Id,
+			Contact_Role_ID__c=oppCntRolePartialSoftCredit.Id
+		);
+		insert partialSoftCredit;
+
+		Test.stopTest();
```
force2b, line 76: `Test.startTest()` and `Test.stopTest()` should be used to wrap around the actual calls to logic being tested. In this method though, it's used to wrap the data setup - which isn't actually testing anything. Instead, these should be used around the two lines below that actually make calls to the classes to be tested.

force2b (changes requested):

JoseRossiSF (approved):
LGTM

force-app/main/default/classes/ListDonation.cls
```diff
@@ -0,0 +1,215 @@
+/*
```
force2b, line 1: I'm unsure about this class name. It's a generic name, starting with "list", but the implementation is specific to this component. It also has a lot of `public static` methods. Are all of these methods called externally? Is there a reason it's all static versus an instance service class called by a controller for the LWC (one that would have the @AuraEnabled methods). Basically, I think the architecture for the class structure should be reviewed and standarized. @bdvorachek @psadie-sfdo What do you think?

Also, as a new class, the copyright should be 2021, and as noted earlier all new classes, components, etc. should be at api v53.

force2b (changes requested):

force-app/main/default/classes/ListDonation.cls
bdvorachek: A better name here would be `DonationHistoryController`. Any class that has @AuraEnabled methods is a controller, and ideally it only has @AuraEnabled methods.

Inside of any @AuraEnabled static method in DonationHistoryController, we would spin up an instance of DonationHistoryService, probably providing the Id of the Contact as a constructor argument. All of the non-@AuraEnabled methods would get converted from static to being instance methods inside of DonationHistoryService. Example of how getDonationHistory would look in this controller class:

```
public static List<DonationHistoryView> getDonationHistory(Id contactId) {
    DonationHistoryService service = new DonationHistoryService(contactId);
    try {
        return service.getDonationHistory();
    } catch(Exception ex) {
    }
}
```

bdvorachek (changes requested):

JoseRossiSF (approved):

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,87 @@
+@IsTest
```
force2b, line 1: All classes require the same copyright and info headers

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,87 @@
...
+public class DonationHistoryController_TEST {
+	@IsTest
```
force2b, line 3: The indentation is off in this class. It should match all other classes (4 spaces).

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,87 @@
...
+		insert oppCntRolePartialSoftCredit;
+
+		Partial_Soft_Credit__c partialSoftCredit = new Partial_Soft_Credit__c (
+			Amount__c=50,
+			Contact__c=mockDefaultContact.Id,
+			Opportunity__c=opportunityPartialSoftCredit.Id,
+			Contact_Role_ID__c=oppCntRolePartialSoftCredit.Id
+		);
+		insert partialSoftCredit;
+		Test.stopTest();
```
force2b, line 75: Test.startTest() and Test.stopTest() should only be used to wrap around the code actually being tested below, not the data setup logic.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,201 @@
...
+        return result;
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return total number of record
+    */
+    public static List<AggregateResult> getTotalRecords(Id contactId){
+        List<AggregateResult> totalRecords = [SELECT COUNT(Id)
```
force2b, line 63: I highly suggest testing this in a **large** organization - one with 20 million Opportunity records and at least the same number of OCR records.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,201 @@
...
+                DonationHistoryController.DonationHistory softCredit = new DonationHistoryController.DonationHistory();
+                softCredit.id = ocr.Id;
+                softCredit.name = ocr.Opportunity.Name;
+                softCredit.isPrimary = ocr.isPrimary;
+                softCredit.closeDate = ocr.Opportunity.CloseDate;
+                softCredit.role = ocr.Role;
+                softCredit.amount = ocr.Opportunity.Amount;
+                softCredit.creditType = SOFT_CREDIT;
+                softCredit.paymentMethod = 'Cash';
+                softCreditList.add(softCredit);
```
force2b, line 134: This logic is nearly identical the logic on lines 98-107 above. These can be combined into a single method, possibly as the constructor for `DonationHistory`, and then only change the values that require changing (`CreditType`).

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,201 @@
...
+                newPartialSoftCredit.id = ocr.Id;
+                newPartialSoftCredit.name = ocr.Opportunity.Name;
+                newPartialSoftCredit.isPrimary = ocr.isPrimary;
+                newPartialSoftCredit.closeDate = ocr.Opportunity.CloseDate;
+                newPartialSoftCredit.role = ocr.Role;
+                newPartialSoftCredit.creditType = PARTIAL_SOFT_CREDIT;
+                newPartialSoftCredit.paymentMethod = 'Cash';
+                //TODO: change for Amount__c get on partialSoftCredit query
+                newPartialSoftCredit.amount = ocr.Opportunity.Amount;
+                partialSoftCreditList.add(newPartialSoftCredit);
```
force2b, line 179: Same duplicate logic as noted previously.

force2b (changes requested):

force-app/main/service/DonationHistoryService.cls
mcasco: Constructor was added for the class but the logic that validates the types of credits is different for each case, so it was not modified.

force-app/main/service/DonationHistoryService.cls
mcasco: Constructor was added for the class but the logic that validates the types of credits is different for each case, so it was not modified.

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: unnecessary Test.startTest() and Test.stopTest() removed!

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: indentation fixed for this file!

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: missing copyright added!

force-app/main/service/DonationHistoryService.cls
mcasco: I agree that we should consider this case.
I also think we should discuss it a bit better considering the implementation and scope of the ticket.
Would you suggest generating more records in the datasets?

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -1,87 +1,120 @@
...
+*/
 @IsTest
 public class DonationHistoryController_TEST {
-	@IsTest
-	public static void testGetDonationHistory() {
-		Test.startTest();
-		
-		Date myDateTime = Date.newInstance(2020, 2, 2);
+    @IsTest
+    public static void testGetDonationHistory() {
```
force2b, line 39: Rather than a generic method name, I recommend something more specific that defines exactly what is being tested. For example, `shouldReturnExpectedDonationHistoryResult()`.

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Donation
+* @description Class for test list donation history controller.
+*/
+@IsTest
+public class DonationHistoryController_TEST {
+    @IsTest
+    public static void testGetDonationHistory() {
```
force2b, line 39: Also, test classes and methods can be `private`.

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+        
+        Date myDateTime = Date.newInstance(2020, 2, 2);
+
+        Account mockDefaultAccount = new Account(Name = 'Default Account');
```
force2b, line 43: Please review the `UTIL_UnitTestData_TEST` class for pre-existing test data factory methods.

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+            Name='PartialSoftCredit1', 
+            AccountId=mockDefaultAccount.Id, 
+            Primary_Contact__c=secondaryContact.Id, 
+            Amount=150, 
+            StageName='Closed Won', 
+            CloseDate=myDateTime
+        );
+        insert opportunityHardCredit;
+        insert opportunitySoftCredit;
+        insert opportunityPartialSoftCredit;
```
force2b, line 85: This should be done in a single DML statement, not 3.

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+            isPrimary=false
+        );
+        OpportunityContactRole oppCntRolePartialSoftCredit = new OpportunityContactRole(
+            Role='Soft Credit', 
+            OpportunityId=opportunityPartialSoftCredit.Id, 
+            ContactId=mockDefaultContact.Id, 
+            isPrimary=false
+        );
+        insert oppCntRoleSoftCredit;
+        insert oppCntRolePartialSoftCredit;
```
force2b, line 100: Single DML

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+        
+        Test.startTest();
+        List<DonationHistoryController.DonationHistory> donationHistory = DonationHistoryController.getDonationHistory(mockDefaultContact.Id);
+        List<DonationHistoryController.DonationHistory> filteredDonationHistory = new List<DonationHistoryController.DonationHistory>();
+        Test.stopTest();
+
+        for (DonationHistoryController.DonationHistory donation : donationHistory) {
+            filteredDonationHistory.add(donation);
+        }
+        System.assertEquals(3, donationHistory.size());
```
force2b, line 118: I'm not sure I understand the objective here. The logic takes the result in the `donationHistory` collection and adds it to the `filteredDonationHistory` collection (in a loop that doesn't add any filtering or conditions) and then checks the size of the resulting `filteredDonationHistory`. Unit Tests need to be more explicit - testing very specific behaviors. Many unit tests won't even require DML because they can be passed mocked data to work with. In this case, I'm not sure this is actually testing anything at all other than that there are 3 records - and the test data setup is only setting up 3 records. 

There's no negative test methods either. For example, are there Opportunities or types of Soft Credits that should not be retrieved by the `getDonationHistory()` method? If so, then a new test method is needed to verify that the logic is _only_ pulling in the expected data and not data that should be filtered out.

force-app/main/service/DonationHistoryService.cls
force2b: Yes, and work with @psadie-sfdo and @bdvorachek to get an org with larger data sets that a query such as this can be tested against.

force2b (changes requested):

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: For this case we generate different opportunities and each one with a different configuration to be able to have all the types of credits, I did not find a method that allows us to make these configurations in the class, what exactly would you be suggesting in this case?

force2b: There's certainly methods for generating Accounts and Contacts though.

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: Validations against the data returned by the method were missing, as well as more test data to validate that it is being filtered correctly, which was the initial idea for this test.
more data was added to corroborate the filtering and also more assertions for the returned data

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: Yes I saw them, but we have methods like `getContact` that return a contact but without an ID (since it does not make the dml) which we need to reference in the creation of opportunities, also `createAccountContactTestData` which looks very useful but at the same time it does not return the accounts and contacts generated, it is also `oppsForContactListByRecTypeId` but do not reference the account id in the opportunity, since we need certain references, I just wanted to clarify what would be the suggestion to implement the methods of that class or in any case which one

mcasco: I'm sorry if this is generating confusion but since I am not aware of the resources, maybe I need a hand, thanks in advance @force2b

bdvorachek: >it does not return the accounts and contacts

This is a great observation! I think that `createAccountContactTestData` can be updated to return either the Contacts or Accounts that it creates. If I had to choose, I'd say that method should return a `List<Contact>`. Then you can build out the base data like this:

```
List<Contact> contacts = UTIL_UnitTestData_TEST.createAccountContactTestData(CAO_Constants.HH_ACCOUNT_TYPE, 1, 2, 0);
List<Opportunity> opportunities = UTIL_UnitTestData_TEST.oppsForContactList(contacts, null, 'Closed Won', myDateTime, 250, null, null);
insert opportunities;

```

There's no UTIL_UnitTestData_TEST method for creating OpportunityContactRoles, yet. I think one could be created here. A function with this signature should work:

`public static List<OpportunityContactRole> createRolesForOpportunities(List<Id> opportunityIds, Id contactId, String role)`

force-app/main/service/DonationHistoryService.cls
bdvorachek: @mcasco @force2b - I think this function can be removed, it's not being used anywhere that I can see. Are there future plans to need something like this for Donation History?

force-app/main/default/classes/DonationHistoryController_TEST.cls
force2b: Very good suggestions @bdvorachek. Overall though, I think what's missing in this test class is a solid set of test scenarios (see [yesterday's comment](https://github.com/SalesforceFoundation/NPSP/pull/6760#discussion_r773126218)). This test class has one test method that tests only one possibility. There are way more scenarios to test. Plus, I'm questioning whether DML is even needed to fully test this logic. The service class could have the selector results injected with mock data. That would open this test class up to being able to test multiple scenarios very quickly.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+    }
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifySoftCreditCondition(List<OpportunityContactRole> donations){
+        //with the list of donations verify if the Role match the filter rule value on the NPSP_ClosedWon_Opps_SC filter group
+        Filter_Rule__mdt roles = [SELECT Constant__c FROM Filter_Rule__mdt WHERE QualifiedApiName = 'NPSP_OppsSoftCreditRoles'];
```
force2b, line 112: One more extremely important point - in a unit test context, there will be zero records in the `Filter_Rule__mdt` object **and** it's not possible to insert records into that object. 

See the `CMT_FilterRule_TEST.mockFilterCMTValues()` method for a pattern on how to mock data in that object. Once you follow those steps, you'll need to modify this method use the existing `CMT_FilterRule_SEL` class to retrieve the mocked records - maybe with the `getFilterRules()` method and then loop through the collection to find the record with the defined qualified APi name.

Also, I recommend moving 'NPSP_OppsSoftCreditRoles' to a constant in this class.

force2b (changes requested):

force-app/main/service/DonationHistoryService.cls
mcasco: After checking and debugging the content of `Filter_Rule__mdt`, I can confirm it returns data in the test context. 
Given this facts do you still think the suggested changes above are required?
I will put NPSP_OppsSoftCreditRoles in a constant!

force2b: You're testing that in your personal scratch org where customizable rollups is already enabled and thus that object has data. That is not the case in a new scratch org, which is exactly how our automated builds function. In that context, the object will have zero records and this will fail.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+    static final String SOFT_CREDIT = 'Soft Credit';
+    static final String PARTIAL_SOFT_CREDIT = 'Partial Soft Credit';
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    public List<DonationHistoryController.DonationHistory> getDonationHistory(Id contactId){
+        List<OpportunityContactRole> donations = [SELECT Id, Opportunity.Name, Opportunity.CloseDate, Opportunity.Amount, Contact.Email, Contact.Name, Role, isPrimary
```
bdvorachek, line 47: This line is a bit long, can we do one field per line so it doesn't wrap?

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+                                AND ContactId =:contactId];
+        return totalRecords;
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify each condition to list Donation.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyCreditCondition(List<OpportunityContactRole> donations){
```
bdvorachek, line 75: The name of this function could be a little more descriptive - `buildDonationHistories` is my suggestion. Maybe `buildHistoriesFromOCRs`?

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+
+        return donationHistory;
+    }
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify hard credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyHardCreditCondition(List<OpportunityContactRole> donations){
```
bdvorachek, line 94: The name of this function could describe it better, "verify" isn't what we're doing here. I think a better name here would be `buildHardCreditHistoryFromOCRs`.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify partial soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyPartialSoftCreditCondition(List<Id> donationsId, List<OpportunityContactRole> donations){
+        //search on PartialSoftCredit object for ids filtered before usin IN()
+        //only ids with not type of credit defined before
+        List<Id> partialSoftCreditIds = new List<Id>();
+        List<Partial_Soft_Credit__c> partialSoftCredit = [SELECT id, Amount__c, Contact_Role_ID__c FROM Partial_Soft_Credit__c WHERE Contact_Role_ID__c IN :donationsId];
```
bdvorachek, line 146: Let's take this query, and the one for OpportunityContactRole, and break them out into a DonationHistorySelector class as instance methods. We should also add `WITH SECURITY_ENFORCED` to those queries. There's a few more details in our slack channel on how this would work, and how we can mock data from that selector class to greatly speed up unit test performance.

force-app/main/service/DonationHistoryService.cls
bdvorachek: One more note on these, instead of doing String.split(';') on the Constant__c field, you can construct an instance of a ` new CMT_FilterRule(Filter_Rule__mdt filterRule)`. Then inside of the loop, you can call `isIncluded(SObject)` on it, and pass in the OCR you want to evaluate against that rule.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify each condition to list Donation.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyCreditCondition(List<OpportunityContactRole> donations){
+        List<DonationHistoryController.DonationHistory> donationHistory = new List<DonationHistoryController.DonationHistory>();
+        List<DonationHistoryController.DonationHistory> hardCredit = verifyHardCreditCondition(donations);
+        List<DonationHistoryController.DonationHistory> softCredit = verifySoftCreditCondition(donations);
+        List<Id> donationsId = filterVerifiedIds(softCredit);
```
bdvorachek, line 79: These are actually OCR Ids in this list. I think a better name for this variable would be `softCreditIds`.

force-app/main/service/DonationHistoryService.cls
psadie-sfdo, line 75: What do we think about changing the name `donations` to just `credits` here? It feels more accurate to me.
```suggestion
    private static List<DonationHistoryController.DonationHistory> verifyCreditCondition(List<OpportunityContactRole> credits){
```

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify each condition to list Donation.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyCreditCondition(List<OpportunityContactRole> donations){
+        List<DonationHistoryController.DonationHistory> donationHistory = new List<DonationHistoryController.DonationHistory>();
+        List<DonationHistoryController.DonationHistory> hardCredit = verifyHardCreditCondition(donations);
```
psadie-sfdo, line 77: We should pluralize the names for these lists.

force-app/main/service/DonationHistoryService.cls
psadie-sfdo, line 94: Similar comment to the above with regards to renaming `donations` to `credits`.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+        }
+        return hardCreditList;
+    }
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifySoftCreditCondition(List<OpportunityContactRole> donations){
```
psadie-sfdo, line 110: Similar comment to the above with regards to renaming `donations` to `credits`.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+        }
+        return softCreditList;
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Filter for ids already classified as Soft Credit.
+    */
+    private static List<Id> filterVerifiedIds(List<DonationHistoryController.DonationHistory> softCredit){
```
psadie-sfdo, line 129:
```suggestion
    private static List<Id> filterVerifiedIds(List<DonationHistoryController.DonationHistory> softCredits){
```

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+        List<Id> donationsId = new List<Id>();
```
psadie-sfdo, line 130: This should probably either be `softCreditIds` or `donationHistoryIds`.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,176 @@
...
+        }
+        return donationsId;
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify partial soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyPartialSoftCreditCondition(List<Id> donationsId, List<OpportunityContactRole> donations){
```
psadie-sfdo, line 142: Similar comment to the above with regards to renaming `donations` to `credits`.
Similar situation with `donationsId` here. I believe this should be `softCreditIds`. Using `donationIds` here would, at a glance, lead me to think of Opportunities.

psadie-sfdo (changes requested):
A handful of comments. Please update all list variable names to be plural and let me know if you agree/disagree regarding the usage of the word donations vs softCredits when referring to OCRs.

force-app/main/service/DonationHistoryService.cls
mcasco: When testing this implementation with the instance of `CMT_FilterRule`, it throws an error when passing as a parameter the ocr record to `isIncluded(SObject)` method. Since that should be resolved in the `CMT_FilterRule` class, for this time we keep the split(';')

force-app/main/service/DonationHistoryService.cls
mcasco: Done

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: done

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: Done

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: added test cases to check permissions and no results

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: done

force-app/main/service/DonationHistoryService.cls
mcasco: Done

force-app/main/service/DonationHistoryService.cls
mcasco: Done

force-app/main/default/classes/DonationHistoryController.cls
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    @AuraEnabled(cacheable=true)
+    public static List<DonationHistory> getDonationHistory(Id contactId){
+        DonationHistoryService service = new DonationHIstoryService();
```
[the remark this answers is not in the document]
mcasco: Done

force-app/main/service/DonationHistoryService.cls
mcasco: Done

JoseRossiSF (approved):
LGTM

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,191 @@
...
+        } else {
+            throw new UTIL_Permissions.InsufficientPermissionException(Label.commonInsufficientPermissions);
+        }
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify each condition to list Donation.
+    */
```
bdvorachek, line 91: Generally we don't write ApexDoc comments for private methods. The Function name, Signature, and Return type here are pretty descriptive on their own.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,191 @@
...
+        });
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    public List<DonationHistoryController.DonationHistory> getDonationHistory(Id contactId){
+        if(hasAccessToDonationHistory()){
```
bdvorachek, line 78: Our styling for if statements includes spaces in a few spots:
```suggestion
        if (hasAccessToDonationHistory()) {
```

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -0,0 +1,75 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description Donation History Service class in NPSP.
```
bdvorachek, line 33:
```suggestion
* @description Donation History Selector class.
```

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -0,0 +1,75 @@
...
+            ORDER BY Opportunity.CloseDate 
+            DESC LIMIT 2000
+        ];
+    }
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Partial_Soft_Credit__c data.
+    */
+    public List<Partial_Soft_Credit__c> getPartialSoftCredit(List<Id> donationIds){
```
bdvorachek, line 64: `donationIds` suggests that these are Opportunity Ids, but I think we should make it clearer that these are OCR Ids:
```suggestion
    public List<Partial_Soft_Credit__c> getPartialSoftCredit(List<Id> opportunityContactRoleIds){
```

force-app/main/selector/DonationHistorySelector.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
bdvorachek, line 3:
```suggestion
    <apiVersion>53.0</apiVersion>
```

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -0,0 +1,75 @@
...
+* @date 2021
+* @description Donation History Service class in NPSP.
+*/
+public inherited sharing class DonationHistorySelector {
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return OpportunityContactRole data.
+    */
+    public List<OpportunityContactRole> getOpportunityContactRoleList(Id contactId) {
```
bdvorachek, line 41: One more suggestion for naming in this method, since we're querying for Won Opportunities and not all OCRs associated with the contact:

`getWonOpportunityContactRolesByContactId(Id contactId)`

force-app/main/service/DonationHistoryService.cls
mcasco: done

force-app/main/service/DonationHistoryService.cls
mcasco: spaces added

force-app/main/selector/DonationHistorySelector.cls-meta.xml
mcasco: done

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,191 @@
...
+    }
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> buildSoftCreditHistoryFromOCRs(List<OpportunityContactRole> donations){
+        //with the list of donations verify if the Role match the filter rule value on the NPSP_ClosedWon_Opps_SC filter group 
+        Filter_Rule__mdt roles = CMT_FilterRule_SEL.getFilterRuleByDeveloperName(NPSP_CREDIT_ROLES);
```
bdvorachek, line 129: Another naming suggestion here, since the object is a Filter Rule we can name it after the type that it is.
```suggestion
        Filter_Rule__mdt filterRule = CMT_FilterRule_SEL.getFilterRuleByDeveloperName(NPSP_CREDIT_ROLES);
```

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,191 @@
...
+    * @author Salesforce.org
+    * @date 2021
+    * @description Verify partial soft credit condition.
+    */
+    private static List<DonationHistoryController.DonationHistory> verifyPartialSoftCreditCondition(List<Id> softCreditIds, List<OpportunityContactRole> credits){
+        List<Id> partialSoftCreditIds = new List<Id>();
+        List<Partial_Soft_Credit__c> partialSoftCredit = new DonationHistorySelector().getPartialSoftCredit(softCreditIds);
+        for(Partial_Soft_Credit__c ps: partialSoftCredit){
+            partialSoftCreditIds.add(ps.Contact_Role_ID__c);
+        }
```
bdvorachek, line 164: I think this is where we can apply an instance of the CMT_FilterRule class.

```suggestion
        Filter_Rule__mdt ruleCmt = CMT_FilterRule_SEL.getFilterRuleByDeveloperName(NPSP_CREDIT_ROLES);
        CMT_FilterRule rule = new CMT_FilterRule(ruleCmt);
        for(Partial_Soft_Credit__c ps: partialSoftCredit) {
            if(rule.isIncluded(ps)) {
                partialSoftCreditIds.add(ps.Contact_Role_ID__c);
            }
        }
```

Then, when the partialSoftCreditIds variable is referenced further down, it will only contain Partial Soft Credits with a valid Role. The Role on the OCR and on the Role Name field on Partial Soft Credit should be the same value. With this change, we should not need to do any filtering on the Role/Role Name field in `buildSoftCreditHistoryFromOCRs`

bdvorachek (changes requested):
One more comment on error handling - This component will only work if Customizable Rollups is enabled in the given org. Additionally, if the `NPSP_OppsSoftCreditRoles` Filter_Rule__mdt is renamed or deleted this component will no longer work. Two options that I think make sense for this scenario:

- Display an error message indicating that CRLP must be enabled, and the NPSP_OppsSoftCreditRoles rule must exist for the component to function.
- Don't display any soft credit information in the widget if CRLP is not enabled, or the NPSP_OppsSoftCreditRoles rule does not exist, but do display all of the other donation information.

force-app/main/selector/DonationHistorySelector.cls
mcasco: done

force-app/main/selector/DonationHistorySelector.cls
mcasco: done

force-app/main/selector/DonationHistorySelector.cls
mcasco: changed

force-app/main/service/DonationHistoryService.cls
mcasco: changed

force-app/main/service/DonationHistoryService.cls
mcasco: instance of CMT_FilterRule added!

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,160 @@
...
+            Partial_Soft_Credit__c.Contact_Role_ID__c
+        });
+    }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History list.
+    */
+    public List<DonationHistoryController.DonationHistory> getDonationHistory(Id contactId){
```
psadie-sfdo, line 72:
```suggestion
    public List<DonationHistoryController.DonationHistory> getDonationHistory(Id contactId) {
```

Let's add a space before every method's opening brace.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -0,0 +1,160 @@
...
+        if (hasAccessToDonationHistory()) {
+            List<OpportunityContactRole> credits = new DonationHistorySelector().getWonOpportunityContactRolesByContactId(contactId);
```
psadie-sfdo, line 74:
```suggestion
            List<OpportunityContactRole> credits =
                new DonationHistorySelector().getWonOpportunityContactRolesByContactId(contactId);
```

Non-blocking but ideally break up lines longer than 120 characters for better readability.

psadie-sfdo (changes requested):
Hey folks, looks like there are two donation history unit tests failing. Below are a couple more general change requests. Thanks!

force-app/main/service/DonationHistoryService.cls
mcasco: spaces added!

force-app/main/service/DonationHistoryService.cls
mcasco: break up added for lines longer than 120 characters!

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -1,17 +1,25 @@
...
 import PAYMENT_FIELD from '@salesforce/schema/DataImport__c.Payment_Method__c';
 import donationHistoryDatatableAriaLabel from '@salesforce/label/c.donationHistoryDatatableAriaLabel';
 import RD2_ScheduleVisualizerColumnDate from '@salesforce/label/c.RD2_ScheduleVisualizerColumnDate';
+import getDonationHistory from '@salesforce/apex/DonationHistoryController.getDonationHistory';
 import commonAmount from '@salesforce/label/c.commonAmount';
 import donationHistoryDonorLabel from '@salesforce/label/c.donationHistoryDonorLabel';
 export default class DonationHistoryTable extends LightningElement {
     @api contactId;
 
+    recordsToLoad = 50;
```
bdvorachek, line 13: I think the records to load variable here is truly a constant, so lets bring it outside of the class and give it an UPPER_SNAKE_CASE name:

```suggestion
const RECORDS_TO_LOAD = 50;
export default class DonationHistoryTable extends LightningElement {
    @api contactId;
```

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -26,26 +34,28 @@ export default class DonationHistoryTable extends LightningElement {
...
     
     // eslint-disable-next-line @lwc/lwc/no-async-await
     async connectedCallback() {
-        const data = MOCK_DATA.slice(0, 50);
-        this.data = data;
+        getDonationHistory({contactId: this.contactId})
+        .then(data => {
+            if (data) {
+                this.allData = data;
+                this.data = data.slice(0, this.recordsToLoad);
```
bdvorachek, line 55: See above, this variable gets lifted out of the class body and into a `const`
```suggestion
                this.data = data.slice(0, RECORDS_TO_LOAD);
```

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -54,17 +64,17 @@ export default class DonationHistoryTable extends LightningElement {
...
-        //TODO: add wired service
-        //Display a spinner to signal that data is being loaded
         event.target.isLoading = true;
         if (this.data.length >= this.totalNumberOfRows) {
             event.target.enableInfiniteLoading = false;
             event.target.isLoading = false;
             return;
         }
+        const current = this.data.length;
+        const offset = current + this.recordsToLoad;
```
bdvorachek, line 74:
```suggestion
        const offset = current + RECORDS_TO_LOAD;
```

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: changed to RECORDS_TO_LOAD constant

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: changed

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: changed

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -0,0 +1,231 @@
...
+        insert recordsOcr;
+
+        Partial_Soft_Credit__c partialSoftCredit = new Partial_Soft_Credit__c (
+            Amount__c=50,
+            Contact__c=listConSrc[0].Id,
+            Opportunity__c=opportunityPartialSoftCredit.Id,
+            Contact_Role_ID__c=oppCntRolePartialSoftCredit.Id
+        );
+        insert partialSoftCredit;
+        generateFilterRules();
```
bdvorachek, line 144: generateFilterRules() won't work here because it is setting a static variable on another class, which gets cleared between each test run. Instead, it'll need to be called inside of each test method in this class.

bdvorachek (changes requested):
Looking good! Few more changes to request:

I sent a snippet of `DonationHistoryService#buildPartialSoftCredits` to you in Slack. Filtering CMTs there does not have the effect that I thought it did when I originally made that recommendation, apologies! We can remove the CMT_FilterRule stuff in that method and just gather up all of the matching Partial Soft Credit OCR records.

To fix the failing tests here, we'll need to call `generateFilterRules()` in each test, as static variables get unset between each test run even if you're setting them in the TestSetup method.

And one additional question here - Should the result be displayed in some sort of order, like newest donation at the top and then older ones below it?

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: the method was added for each test as required, and yes the results have to be delivered in order but we have a ticket where the implementation of the infinite scroll and the order will be evaluated, since we perform the filtration after obtaining the data it can generate a variation in how the data is displayed

bdvorachek (approved):

bdvorachek:
Re-started 2gp feature test since I think those failures are not related to this WI, but approved this as long as that build ends up passing.

psadie-sfdo (approved):

dcaballeroSf (approved):

sam-knox:
@jencrabtree - Can you write the release note for this feature? No rush at this point.

PR #6761, opened 2021-12-02
Feature/callen/238  fix duplicate oppts when creating r dwith campaign

callen-sfdo:
This fixes an issue where if a Recurring Donation is created in BDI with a campaign defined, but no initial Donation defined it still creates an initial donation.  This was occurring because the code that checks to see if any Opportunity Object mapping fields were populated was including the DonationCampaignImported__c field (Which is used by both RD and Oppts) in the populated field check that determines whether or not the Opportunity attempts to be created.

We are creating a field mapping because Donation Campaign Source was missed when we originally converted the help text mappings to the custom metadata "advanced" mappings.  It has been getting created automatically during advanced mapping enablement so it hasn't directly caused problems, but the fact that it is missing meant that the unit tests for Recurring Donation didn't catch this bug earlier. Adding it means that the default mapping is more accurate and we can now catch regressions for this bug.  It does not need to be added to any install scripts because an equivalent mapping has already been created for all orgs during enablement.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/BDI_DataImportService.cls
+++ b/force-app/main/default/classes/BDI_DataImportService.cls
@@ -2110,9 +2110,36 @@ global with sharing class BDI_DataImportService {
     */
     public static Boolean anyFieldsPopulatedForObjectMapping(DataImport__c di,
                                                             Map<String,String> diFieldToTargetFieldMap) {
+        return anyFieldsPopulatedForObjectMapping(di, diFieldToTargetFieldMap, new String[]{});
+    }
+
+    /*******************************************************************************************************
+    * @description loops through fields provided in Map and determines if any of them have a value.  If at least
+    * one field does, then it returns true.  This variant of the method also allows a list of fields to ignore to be 
+    * passed in since some fields may be mapped to two different object mappings, and others may always have a value.
+    * @param di the data import record to check
+    * @param diFieldToTargetFieldMap the mapping of DI field to target field that will be evaluated.
+    * @param fieldsToIgnore List of fields that should be ignored when checking for values, usually because these fields
+    * are shared with another object mapping, or are always populated.
+    * @return Boolean whether any fields were populated
+    */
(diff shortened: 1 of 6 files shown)


grivero (commented):
Changes looks good overall but shouldn't we have unit tests to go over the changes in BDI_Donations and BDI_RecurringDonations?

psadie-sfdo (changes requested):
One small request and I think this branch needs to be renamed to something like `feature/238__callenFixDuplicateOpptsWhenCreatingRDwithCampaign`

force-app/main/default/classes/BDI_DataImportService.cls
```diff
@@ -2110,9 +2110,37 @@ global with sharing class BDI_DataImportService {
...
+            // Force all fields to lower case so they can match correctly
+            for (Integer i = 0; i < fieldsToIgnore.size(); i++) {
+                fieldsToIgnore[i] = fieldsToIgnore[i].toLowerCase();
+            }
+        }
+
         for (String diField : diFieldToTargetFieldMap.keySet()){
-            if (di.get(diField) != null) {
+            // Set the is populated flag to true if the field should not be ignored, and it does have a value in the DI
+            Boolean shouldAddPopulatedField = !fieldsToIgnore.contains(diField.toLowerCase()) && di.get(diField) != null
```
psadie-sfdo, line 2142: Sorry, previous suggestion didn't have the closing ;
```suggestion
            Boolean shouldAddPopulatedField = !fieldsToIgnore.contains(diField.toLowerCase()) && di.get(diField) != null;
```

psadie-sfdo (changes requested):

callen-sfdo:
Changed branch name and opened new PR: https://github.com/SalesforceFoundation/NPSP/pull/6768

PR #6765, opened 2021-12-06
Gift Entry: a11y improvements

psadie-sfdo:
[W-7965470](https://gus.lightning.force.com/a07B0000008UEhIIAW)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
+++ b/force-app/main/default/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
@@ -29,6 +29,17 @@ export default class geTemplateBuilderFormField extends LightningElement {
     hasPermission = hasViewSetup;
     shouldRender = true;
 
+    @api
+    focusOnDownButton() {
+        const downButton = this.template.querySelector('[data-id="down"]');
+        downButton.focus();
+    }
+
+    @api
+    focusOnUpButton() {
+        const upButton = this.template.querySelector('[data-id="up"]');
+        upButton.focus();
+    }
 
     /*******************************************************************************
     * @description Retrieves the target object's describe data. Used to get the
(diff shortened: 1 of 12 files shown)


force-app/main/default/lwc/geTemplateBuilderFormField/geTemplateBuilderFormField.js
```diff
@@ -29,6 +29,17 @@ export default class geTemplateBuilderFormField extends LightningElement {
     hasPermission = hasViewSetup;
     shouldRender = true;
 
+    @api
```
kenlewis-sfdo, line 32: Is it correct that the component hierarchy requires that these methods be api enabled?

force-app/main/default/lwc/geTemplateBuilderFormSection/geTemplateBuilderFormSection.js
```diff
@@ -16,6 +18,29 @@ export default class GeTemplateBuilderFormSection extends LightningElement {
     @api formSection;
     @api sourceObjectFieldsDescribe;
 
+    @api
```
kenlewis-sfdo, line 21: These also.  Is it that App needs to call the methods on section and section needs to call them on fields?  Just curious - did you look at or consider what a pub sub solution would look like?

psadie-sfdo: Yeah, absolutely on point there. I opted to go down the public method route as opposed to pubsub for three reasons.

1. We already have to query the DOM regardless of the communication method in order to set focus.
2. I didn't want to expand usage of our custom pubsub solution unless there was no other way to reduce to amount of code we'll need to retire when we move onto LMS. 
3. We're not locked into these public methods since neither of the involved components are exposed.

kenlewis-sfdo: OK, thanks for the details! Very helpful.

kenlewis-sfdo (approved):
Section solution in the last commit looks great. Great work, Prince! Thank you.

PR #6766, opened 2021-12-07
Updating API version to 53

lparrott:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


bdvorachek:
Some NPM Packages that we can bump as part of this:
@sa11y/jest 0.5.0 -> 3.1.0
jest 26.6.3 -> 27.4.3

You can check these by running `npm outdated` in the root directory. To update these packages, you can do `npm i jest@latest` and `npm i @sa11y/jest@latest`. That should update the package.json file which we can then commit into the repo.

force2b:
@lparrott FYI - There are some new classes added in #6767 that are at API v51. Since that's an emergency fix, I wasn't going to request those be changed in that PR, but once it's merged and the new classes bubble up to this branch, those new classes will need to be updated.

force-app/main/default/classes/AN_AutoNumberService.cls
```diff
@@ -202,6 +202,9 @@ public class AN_AutoNumberService {
         }
 
         public String getFormattedNumber(Integer num) {
+            // A change in API version 51 and later causes this to fail, so we're keeping 
+            // this Class at 50 for now.
+            // {000} is no longer a valid pattern for the first match when using String.format
```
force2b, line 207: Hmm. Technically `{000}` should have never been supported. In a format string the number zero represents the first parameter in the list sent to `String.format()`. It's very different than the format of an AutoNumber field. It looks like the the number of zeros is used to control the result of `getNumberLength()`, effectively trying to simulate how an auto-number field works.

So, maybe the fix here (to allow us to move this to api 53) is to first call `getNumberLength()` and then replace the format pattern in the string to just `{0}`. That would allow for both the simulation and format logic to work:
* `'ABCZZZ-{000}'` ==> `'ABCZZZ-{0}'`

lparrott: Works for me, I'll get this change (as well as the new Classes) in today.

lparrott: Alright, these changes are in, tests pass, but I haven't done any actual testing with the AN_AutoNumberService yet

force-app/main/default/classes/AN_AutoNumberService.cls
```diff
@@ -202,9 +202,11 @@ public class AN_AutoNumberService {
         }
 
         public String getFormattedNumber(Integer num) {
-            return String.format(displayFormat,
+            Integer numberLength = getNumberLength();
+            String singleDigitDisplayFormat = displayFormat.replace('{000}', '{0}');
```
force2b, line 206: This will fix unit tests, but it wouldn't handle customers who might choose a different format such as "{00000}". You can make it more dynamic by using something like:
```
String singleDigit = '0';
String currentDigitFormat = '{'  + singleDigit.repeat(numberLength) + '}';
String singleDigitDisplayFormat = displayFormat.replace(currentDigitFormat, '{0}');
```

lparrott: I was afraid of something like this, it felt too specific, fix incoming.

dfuller114 (approved):
Looks good. Thanks Luke!

bdvorachek (approved):

PR #6767, opened 2021-12-07
Handle Contact Deletion from Non-Household Accounts 

### LINKED ISSUE #6755: Deleting a contact from a non-Household account renames the Account Name and populates the Primary Contact field
### Background

NPSP 3.209 included some refactor work for accounts and contacts which introduced this new bug when contacts are deleted. 

### Issue Summary

**Scenario 1:** 
If an organisational account (i.e any account that is not a household account) has more than one contact directly associated with it, and one of the contacts in the account is deleted the account name of the organisation account will be renamed to "Anonymous Household" and the Primary Contact lookup field on the account will be populated with an existing contact in the account.

**Scenario 2:**
If customers are using the Individual Account Model (NPSP Settings --> People --> Account Model), their Contacts are associated with 1 Account, typically named "Individual". If a customer deletes one Contact from this Account, the Account is renamed to "Anonymous Household".

**Scenario 3:**
Customers may also encounter this bug when merging contacts, where the contacts being merged are not both household contacts. If using the Individual (Bucket) Account Model please also see issue: https://github.com/SalesforceFoundation/NPSP/issues/6772

**Note:** The organisation or individual account will not be renamed when you delete the last contact in the account

### Steps to Reproduce

**Scenario 1:**
1. Create a new account "Test Organization" ensuring the Account Record Type is not the household account record type i.e Record Type = Organization
2. Create a business contact "Test Contact 1" for this account i.e ensure the Account Name field on the contact is "Test Organization"
3. Create a 2nd business contact "Test Contact 2" for this account as per step 2
4. Delete either "Test Contact 1" or "Test Contact 2"
5. Confirm the Account Name of "Test Organization" has been changed to "Anonymous Household" and the Primary Contact lookup field has been populated with a link to the remaining contact

**Scenario 2:**
1. Go to NPSP Settings --> People --> Account Model and switch to the Individual Account Model.
2. Create 3 Contacts and leave their Account blank. They will be associated with an Account named "Individual."
3. Delete one of the Contacts. The Account is renamed to "Anonymous Household."

**Scenario 3:**
1. Go to NPSP Settings --> People --> Account Model and switch to the Individual Account Model.
2. Create 3 Contacts and leave their Account blank. They will be associated with an Account named "Individual."
3. Merge 2 of these contacts. The Account is renamed to "Anonymous Household."

### Workarounds
**Scenario 1:** 
Do not directly delete the contact. First, clear the Account Name value in the Contact you want to delete, which will result in a household account being created for the contact, then delete the contact.

**Scenario 2:**
Do not directly delete the contact. Create a new Organisation Account. Edit the Contact and switch its Account value to that organisational Account. Save and then delete the Contact.

**Scenario 3:**
Unfortunately there is no viable workaround for this scenario that is less onerous than simply renaming the Anonymous Account back to it's original name

kenlewis-sfdo:
[W-10227258](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000SO8pDYAT/view)

# Critical Changes

# Changes

# Issues Closed
Fixes #6755

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/CON_ContactMerge_TEST.cls
+++ b/force-app/main/default/classes/CON_ContactMerge_TEST.cls
@@ -784,7 +784,9 @@ public class CON_ContactMerge_TEST {
         selectMasterRecord(controller, con.id);
 
         // Act: WHEN one Individual Bucket Contact is merged into another Bucket Contact
+        Test.startTest();
         controller.mergeContacts();
+        Test.stopTest();
 
         // Assert: THEN
         //second contact should be deleted
--- a/force-app/main/default/classes/CON_ContactMerge_TEST.cls
+++ b/force-app/main/default/classes/CON_ContactMerge_TEST.cls
@@ -736,29 +736,30 @@ public class CON_ContactMerge_TEST {
         //skip the test if Advancement is installed
         if(ADV_PackageInfo_SVC.useAdv()) return;
 
-        merge2Individuals(CAO_Constants.ONE_TO_ONE_PROCESSOR);
+        merge2IndividualBucketContacts(CAO_Constants.ONE_TO_ONE_PROCESSOR);
     }
     static testMethod void merge2IndividualsHHAccount(){
         //skip the test if Advancement is installed
         if(ADV_PackageInfo_SVC.useAdv()) return;
 
-        merge2Individuals(CAO_Constants.HH_ACCOUNT_PROCESSOR);
+        merge2IndividualBucketContacts(CAO_Constants.HH_ACCOUNT_PROCESSOR);
     }
     //merge two contacts in the Individual Account
-    //the organization shouldn't be deleted
-    static void merge2Individuals(string strProcessor) {
-
-        npe01__Contacts_and_Orgs_Settings__c contactSettingsForTests = UTIL_CustomSettingsFacade.getContactsSettingsForTests(new npe01__Contacts_and_Orgs_Settings__c (npe01__Account_Processor__c = CAO_Constants.BUCKET_PROCESSOR));
-
+    //the organization shouldn't be deleted or renamed
+    static void merge2IndividualBucketContacts(string strProcessor) {
+        // Arrange: GIVEN two Individual Bucket Contacts exist and are to be merged on the
+        // Contact Merge Visualforce page
         Contact con = UTIL_UnitTestData_TEST.getContact();
-
         Contact con2 = UTIL_UnitTestData_TEST.getContact();
         con2.LastName = '2Contact_forTests_Merge';
+        Contact[] contacts = new Contact[]{
+                con, con2
+        };
+        insertAsIndividualBucketContacts(contacts);
 
-        insert new Contact[]{con, con2};
-
-        con = [Select id, accountId, FirstName, LastName, MailingStreet from Contact where Id = :con.id];
-        con2 = [Select id, accountId, FirstName, LastName, MailingStreet from Contact where Id = :con2.id];
+        Map<Id, Contact> queriedBucketContacts = getContacts(new Set<Id>{con.Id, con2.Id});
+        con = queriedBucketContacts.get(con.Id);
+        con2 = queriedBucketContacts.get(con2.Id);
 
         Test.setCurrentPageReference(new PageReference('Page.CON_ContactMerge'));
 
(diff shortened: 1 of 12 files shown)


force-app/main/default/classes/IndividualBucketAccount.cls
```diff
@@ -0,0 +1,63 @@
...
+
+    private IndividualBucketAccount(Account anAccount){
+        this.record = anAccount;
+    }
+
+    public String name() {
+        return record.Name;
+    }
+
+    private static List<Account> getBucketAccounts() {
```
dfuller114, line 53: Would there only ever be one bucket account in an org?

force-app/main/default/classes/IndividualBucketAccount.cls
```diff
@@ -0,0 +1,63 @@
...
+        Account bucketAccount = (bucketAccounts.size() == 1) ? bucketAccounts.get(0) : null;
+        return new IndividualBucketAccount(bucketAccount);
+    }
+
+    private IndividualBucketAccount(Account anAccount){
+        this.record = anAccount;
+    }
+
+    public String name() {
+        return record.Name;
```
dfuller114, line 50: If the `anAccount` parameter is null from the constructor after calling `getInstance()`, this could throw a null pointer. We should use `record?.Name` here.

force-app/main/default/classes/IndividualBucketAccountTests_TEST.cls
```diff
@@ -0,0 +1,62 @@
...
+
+    @IsTest
+    static void delete_one_contact_from_bucket_account_should_not_rename_account() {
+        // Arrange: GIVEN the App is configured to use the Individual Bucket Account
+        UTIL_CustomSettingsFacade.getContactsSettingsForTests(
+                new npe01__Contacts_And_Orgs_Settings__c
+                        (npe01__Account_Processor__c = CAO_Constants.BUCKET_PROCESSOR));
+
+        // Arrange: AND the Bucket Account has multiple child Contacts
+        Contact orgContact1 = new Contact(LastName = 'testCon1');
```
dfuller114, line 46: Does the bucket account automatically get created when we create the first contact? Noticed that we didn't insert an account here first

dfuller114 (commented):
Looks good Ken! Just added a few comments.

force-app/main/default/classes/IndividualBucketAccount.cls
bdvorachek: If they're using the bucket account model, that should be the case. It does create data skew problems, which is why we don't recommend it. Anyone who has more than one bucket is probably doing some kind of workaround due to skew.

force-app/main/default/classes/IndividualBucketAccount.cls
```diff
@@ -0,0 +1,63 @@
...
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @description The Individual Bucket Account Entity in NPSP.
+*/
+public inherited sharing class IndividualBucketAccount {
```
bdvorachek, line 35: Should we annotate this as a test class for now since it isn't used outside of tests?

bdvorachek (approved):

force-app/main/default/classes/IndividualBucketAccount.cls
kenlewis-sfdo: I'm inclined to keep it as opposed to annotating as test and then later un-annotating when needed.  Reason is that the Individual Bucket Account is an Entity in NPSP and it should be easy to get and work with in Apex.  It is only being used in Tests currently, but my preference would be to not annotate as test only.  Curious to hear others' thoughts if anyone else has an opinion one way or the other.

force-app/main/default/classes/IndividualBucketAccountTests_TEST.cls
kenlewis-sfdo: Yep! via LegacyHouseholds#attachToBucketAccount

I'll add to an [upcoming WI](https://gus.lightning.force.com/lightning/r/a07EE00000TxulVYAR/view) that this method should begin using this new IndividualBucketAccount entity instead of doing the on-the-fly insertion of the bucket account inside this method: https://github.com/SalesforceFoundation/NPSP/blob/c105514d755197a5d3565fedf5f866ac24856040/force-app/main/domain/LegacyHouseholds.cls#L181

force-app/main/default/classes/IndividualBucketAccount.cls
dfuller114: I agree we shouldn't annotate unless we only ever intend it to be used in test classes.

force-app/main/default/classes/IndividualBucketAccount.cls
kenlewis-sfdo: Good call, Daniel. After the next iteration of re-factor I think this won't be necessary because this entity should actually be created and inserted if it doesn't already exist when it is referenced. But for the time being there is a slight chance that this class could be used in the name property could be referenced when the account is null.  I added a test and the safe navigator. We can modify this test in the future in line with the re-factoring to use this class when creating new contacts and attaching them to the bucket account.

npsp-reedestockton (approved):

dfuller114 (approved):
thanks Ken!

force2b (changes requested):
I think this looks good, but I'm recommending a change to the unit tests that would help ensure it catches any async operations.

force2b (approved):

force-app/main/default/classes/CON_ContactMerge_TEST.cls
```diff
@@ -782,18 +783,44 @@ public class CON_ContactMerge_TEST {
 
         selectMasterRecord(controller, con.id);
 
-        //merge them
+        // Act: WHEN one Individual Bucket Contact is merged into another Bucket Contact
```
force2b, line 786: Do you think this should be wrapped in a Test.Start/Stop?

kenlewis-sfdo: I'll add it.  I don't think it hurts, but I also don't think it is totally necessary since the offending behavior is observable (with the bug in place) without it.  It would be valuable if there was a specific use case or configuration where we _knew_ that the offending operation ran async.

force2b: Good point about the issue being reproducible in this without start/stop.

force2b (approved):

psadie-sfdo (approved):
Looks good, listed test scenarios all pass! Thanks, Ken!

PR #6768, opened 2021-12-10
W-10034496 - Fix Duplicate Oppts when creating RDs in BDI

callen-sfdo:
This fixes an issue where if a Recurring Donation is created in BDI with a campaign defined, but no initial Donation defined it still creates an initial donation. This was occurring because the code that checks to see if any Opportunity Object mapping fields were populated was including the DonationCampaignImported__c field (Which is used by both RD and Oppts) in the populated field check that determines whether or not the Opportunity attempts to be created.

We are creating a field mapping because Donation Campaign Source was missed when we originally converted the help text mappings to the custom metadata "advanced" mappings. It has been getting created automatically during advanced mapping enablement so it hasn't directly caused problems, but the fact that it is missing meant that the unit tests for Recurring Donation didn't catch this bug earlier. Adding it means that the default mapping is more accurate and we can now catch regressions for this bug. It does not need to be added to any install scripts because an equivalent mapping has already been created for all orgs during enablement.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/BDI_SettingsUI_TEST.cls
+++ b/force-app/main/default/classes/BDI_SettingsUI_TEST.cls
@@ -76,14 +76,15 @@ private with sharing class BDI_SettingsUI_TEST {
         Boolean containsImportedField = false;
 
         for (SelectOption option : ctrl.listSODonationFields) {
-            if (option.getValue() == UTIL_Namespace.alignSchemaNSWithEnvironment(
-                    'npsp__donationcampaignimported__c')) {
+
+            String paymentOriginIdField = SObjectType.DataImport__c.fields.Payment_Origin_Id__c.Name.toLowerCase();
+            if (option.getValue() == paymentOriginIdField) {
                 containsImportedField = true;
                 break;
             }
         }
         Test.startTest();
-        System.assert(containsImportedField, 'Help Text Mapping Enabled');
+        System.assert(!containsImportedField, 'Help Text Mapping Enabled');
         Test.stopTest();
     }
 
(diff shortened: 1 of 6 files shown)


force-app/main/default/classes/BDI_SettingsUI_TEST.cls
```diff
@@ -76,14 +76,15 @@ private with sharing class BDI_SettingsUI_TEST {
...
+
+            String paymentOriginIdField = SObjectType.DataImport__c.fields.Payment_Origin_Id__c.Name.toLowerCase();
+            if (option.getValue() == paymentOriginIdField) {
                 containsImportedField = true;
                 break;
             }
         }
         Test.startTest();
-        System.assert(containsImportedField, 'Help Text Mapping Enabled');
+        System.assert(!containsImportedField, 'Help Text Mapping Enabled');
```
force2b, line 87: @callen-sfdo  Will this fail if the language is not english?

callen-sfdo: @force2b No, these are just asserts not assertEquals, so its not doing any text comparison here.

force2b: @callen-sfdo If I'm not mistaken, this text refers to the value of a [label](https://github.com/SalesforceFoundation/NPSP/blob/main/force-app/main/default/labels/CustomLabels.labels-meta.xml#L12426). When this test runs as the Spanish user in the automated Non-English build, it will fail because that label will have been [translated](https://github.com/SalesforceFoundation/NPSP/blob/main/force-app/main/default/translations/es.translation-meta.xml#L6330) into Spanish.

callen-sfdo: @force2b Sorry, thought you were referring to the assert in particular.  In the comparison we are using .getValue() on the select option which gets the api value of the picklist option, and its comparing it to the api name of the payment origin field. So there should never be any translation concerns.

force2b: @callen-sfdo Sorry. I completely misread the assert and thought it was basically `assertEquals`. The value of "Help Text Mapping Enabled" is the message displayed if the assert fails. I thought this was actually a label value, since there actually is a label with that exact value. So feel free to ignore this entirely.

grivero (approved):

PR #6775, opened 2021-12-15
Unauthorized user can view Filter Groups on /aura?r=1&npsp.CRLP_RollupUI_SVC.setupFilterGroupDetail=1 - NPSP

dfuller114:
Users can no longer make calls into `AuraEnabled` methods in the `CRLP_RollupUI_SVC` class unless the user is an admin.

[W-8821156](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000095FuTIAU/view)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -173,6 +182,8 @@ public with sharing class CRLP_RollupUI_SVC {
     */
     @AuraEnabled
     public static String setupRollupDetail(String rollupId, List<String> targetObjectNames, List<String> detailObjectNames) {
+        if (!STG_Panel.runningUserIsAdmin()) throwAuraPermsException();
```
bdvorachek, line 185: Style-wise, I think we prefer to have if statements w/ curly brackets. I know that'd expand these statements out to three lines from one, and there's a small handful of them in this file. What if we moved this check & throw into a function of its own?

```suggestion
        throwIfRunningUserNotAdmin();
```

```public static void throwIfRunningUserNotAdmin() {
    if(!STG_Panel.runningUserIsAdmin()) {
        throwAuraPermsException();
    }
}```

File: force-app/main/default/classes/CRLP_RollupUI_SVC.cls [hunk 2]
```diff
@@ -428,9 +440,8 @@ public with sharing class CRLP_RollupUI_SVC {
     * @param detailObjectNames list of detail or objects to rollup to query
     * @return a map of strings to string-string maps. Map<String,Map<String,String>>
     */
-
-    @AuraEnabled
```
bdvorachek, line 432: Thanks for spotting all of these, they definitely don't need to be `@AuraEnabled` or even `public`!

force-app/main/default/classes/CRLP_RollupUI_SVC.cls
force2b: Great suggestion @bdvorachek. I wonder if we should consider putting that method in the `STG_PanelUserRollup_CTRL` or `STG_SettingsService` class?

force2b: @dfuller114 Or maybe even the `UTIL_Permissions` class?

dfuller114: @bdvorachek @force2b I like the idea of having this be its own method.
I would like this to not be tied to NPSP Settings since this level of `AuraEnabled` security is not related to NPSP Settings UI. I'm also not sure if this would fit in `UTIL_Permissions` since that class is intended for Object CRUD and FLS checking and this is more API focused permission checking with a client. What about having it in `USER_UserService` where we determine if a user is an admin in the first place?

I can see the user service being refactored in the future to handle entities of different user types such as an `AdminUser` entity that is constructed by `CRLP_RollupUI_SVC` at the top of each of these methods and if that constructor fails because the user is not an admin, the constructor itself could throw the error or something like that. All that to say, it seems to make since to push this logic down as far as we can near where an admin user is determined to allow for refactors like that in the future.

Thoughts?

force2b: @dfuller114 In this case, the security might be specific to NPSP Settings since we're calling the `runningUserIsAdmin()` method. (CRLP Configuration is accessible through NPSP settings, even though the classes don't start with STG). So, maybe using STG_SettingsService or SETG_PanelUserController could work. Other implementations of this kind of security check, on other pages, might check for different perms and possibly through a different exception.

dfuller114: `STG_PanelUserRollup_CTRL` makes sense since it is already performing other user permissions operations related to the user's interaction with configuring rollups. Just pushed up those changes.

force-app/main/default/classes/STG_PanelUserRollup_CTRL.cls
```diff
@@ -381,6 +381,19 @@ public with sharing class STG_PanelUserRollup_CTRL extends STG_Panel {
         ApexPages.addMessage(myMsg);
     }
 
+    public static void throwAuraExceptionIfRunningUserNotAdmin() {
```
force2b, line 384: I like the method abstraction, however this particular class is for a legacy feature (not part of CRLP). If all (or at least almost all) NPSP Settings pages require the permissions as defined in `runningUserIsAdmin()`, couldn't this method just exist in the same `STG_Panel` class? I can easily see this being used by other Aura/LWC components that are referenced in NPSP Settings, such as Advanced Field Mapping, RD2 enablement, and RD2 status mapping. All would have at least this same perm requirement.

bdvorachek: I agree with this, since it'll probably be used in many of the classes that extends STG_Panel or are related to STG_Panel in some way, that's a decent spot for it.

force2b (approved):

bdvorachek (approved):
All tests passing, I appreciate getting coverage on all of the tests that have had the `throwAuraExceptionIfRunningUserNotAdmin()` added, but think that in the future some of these would qualify as duplicates/equivalent class partitions since they all end up testing that particular method.

PR #6781, opened 2021-12-16
Prevent unauthorized contact delete and check account delete access.

npsp-reedestockton:
# Critical Changes

# Changes
We addressed a potential security vulnerability in our delete Contact behavior.

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Label
- commonAccessErrorMessage

# Deleted Metadata


npsp-reedestockton:
Tracking [W-8818978](https://gus.lightning.force.com/a07B00000095DFOIA2)

npsp-reedestockton:
@lparrott @bdvorachek Would appreciate your comments.  Thanks!!

npsp-reedestockton:
@kenlewis-sfdo Would love your comments on the handling of the Selector class and mocking.

force-app/main/default/classes/CON_DeleteContactOverride_TEST.cls
```diff
@@ -551,4 +640,82 @@ private class CON_DeleteContactOverride_TEST {
...
+        {
+            this.contactId = contactId;
+            this.canDeleteContact = canDeleteContact;
+            this.accountId = accountId;
+            this.canDeleteAccount = canDeleteAccount;
+            this.isAccountWithSingleContact = isHouseholdWithSingleContact;
+        }
+
+        private Contact queryContact() {
+            Contact contactDetails = (Contact) JSON.deserialize('{' +
```
kenlewis-sfdo, line 668: Is there a reason JSON.deserialize is used here as opposed to instantiating a new Contact instance and assigning the field values directly?

force-app/main/default/classes/CON_DeleteContactOverrideSelector.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
lparrott, line 3: Not a big deal for a few classes, but let's start using 53.0 as our default!

lparrott (commented):
@npsp-reedestockton lots of good improvements here, I just had a few comments.

force-app/main/default/pages/CON_DeleteContactOverride.page
```diff
@@ -1,58 +1,83 @@
 <apex:page standardController="Contact" extensions="CON_DeleteContactOverride_CTRL" action="{!processDelete}" id="delContactPage">
     <apex:slds />
+    <apex:includeLightning />
+
+    <script>
+        if ({!NOT(hasMinimumAccess)}) {
+           $Lightning.use("{!namespace}" + ":RD2_EnablementApp", function() {
```
bdvorachek, line 7: It seems a little weird to have a reference to the RD2_EnablementApp here in the Contact Delete Override page. I suppose it makes sense to not introduce another Aura Application when we already have one that has utilIllustration declared as a dependency.

force-app/main/default/classes/CON_DeleteContactOverride_TEST.cls
```diff
@@ -551,4 +640,82 @@ private class CON_DeleteContactOverride_TEST {
...
+
+        private Id generate(Schema.SObjectType sObjectType) {
+            String keyPrefix = UTIL_Describe.getObjectDescribe(sObjectType).getKeyPrefix();
+            fakeIdCount++;
+
+            String fakeIdPrefix = ID_PATTERN.substring(0, ID_PATTERN.length() - String.valueOf(fakeIdCount).length());
+
+            return Id.valueOf(keyPrefix + fakeIdPrefix + fakeIdCount);
+        }
+    }
```
bdvorachek, line 720: Let's pop this out into its own class file. Once that's done, we should be able to delete the one in RelationshipsTreeGrid_TEST.cls too. Eventually, we may replace it with the fflib Id Generator class.

force-app/main/default/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -427,6 +455,18 @@ public with sharing class CON_DeleteContactOverride_CTRL {
         }
     }
 
+    /**
+    * @description NPSP namespace: an empty string if unmanaged, or 'npsp' if managed
+    * @return String
+    */
+    public String getNamespace() {
```
bdvorachek, line 462: We have a few instances of this exact function across our codebase. I don't think we need to refactor them all right here, but we could move this into UTIL_Namespace and call it `getNamespaceLWC()`.

force-app/main/default/classes/CON_DeleteContactOverride_TEST.cls
bdvorachek: You should be able to assign an Account SObject directly to that field on the Contact record like this:

`contactDetails.Account = new Account(Name = ACCOUNT_NAME, npe01__SYSTEMIsIndividual__c = true);`

force-app/main/default/pages/CON_DeleteContactOverride.page
bdvorachek: I'd say let's address in a different WI if need be. It's not a huge deal, but is still sorta awkward.

force-app/main/default/classes/CON_DeleteContactOverride_TEST.cls
npsp-reedestockton: @kenlewis-sfdo @bdvorachek It turns out that the actual problem with abandoning the JSON.deserialize approach is Contact.Name.  That field is not writeable and I have to have it to mock the way the confirmation dialog will behave.  Consequently, I believe we need to leave this as JSON.

force-app/main/default/classes/CON_DeleteContactOverride_CTRL.cls
npsp-reedestockton: I started to add this to UTIL_Namespace and discovered there is already a method we can use instead - getComponentNamespace().

npsp-reedestockton:
@npsp-reedestockton @bdvorachek @dfuller114 @kenlewis-sfdo 

Assuming we are leaving the use of RD2_EnablementApp in place until we spin up a new work item to replace with a more appropriately named utility app, I believe I have addressed all of the feedback.

Thank you very much for the great suggestions!!

npsp-reedestockton:
@lparrott I believe all the feedback has been addressed now.  Thanks very much!!!

lparrott (approved):
Thanks for the quick turn around on the feedback!

force-app/main/default/classes/TEST_IDGenerator.cls
```diff
@@ -0,0 +1,53 @@
...
+* https://github.com/apex-enterprise-patterns/fflib-apex-mocks/blob/master/sfdx-source/apex-mocks/main/classes/fflib_IDGenerator.cls
+*/
+
+@IsTest
+public with sharing class TEST_IDGenerator {
+
+    private Integer fakeIdCount = 0;
+    private final String ID_PATTERN = '000000000000';
+
+    public Id generate(Schema.SObjectType sObjectType) {
```
kenlewis-sfdo, line 45: This new class + method shouldn't be needed - see `UTIL_UnitTestData_TEST.mockId`

force-app/main/default/classes/CON_DeleteContactOverrideSelector.cls
```diff
@@ -0,0 +1,133 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    /*********************************************************************************************************
+    * @description Check if the running user has minimum access to delete the record
+    * @return Boolean
+    */
+    public Boolean canDeleteRecord(Id recordId) {
```
kenlewis-sfdo, line 68: The permissions checks should not be in the selector class.  In order to maintain separation of concerns they should be split out into their own class.

force-app/main/default/classes/CON_DeleteContactOverrideSelector.cls
```diff
@@ -0,0 +1,133 @@
...
+        UserRecordAccess userAccess = [
+            SELECT RecordId, HasDeleteAccess
+            FROM UserRecordAccess
```
kenlewis-sfdo, line 71: Is there a reason for querying the UserRecordAccess object directly for this record instead of checking for Delete access on the SObject?  Since the class is 'with sharing', shouldn't SObject Delete access be sufficient here?

force-app/main/default/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -84,56 +101,90 @@ public with sharing class CON_DeleteContactOverride_CTRL {
...
+        hasMinimumAccess = contactOverrideSelector.canDeleteContactRecord(contactId);
+        if (!hasMinimumAccess) {
+            return;
+        }
+
+        contactRecord = contactOverrideSelector.queryContact(contactId);
+
+        isHouseholdAccount = false;
+        isOnlyContactInHousehold = false;
+        hasAccountDeleteAccess = contactOverrideSelector.canDeleteAccountRecord(contactRecord.AccountId);
```
kenlewis-sfdo, line 143: Running canDeleteContactRecord and then this method canDeleteAccountRecord queries the same table twice in the same method.  Could these queries be combined?  Also, see the other question about querying the UserRecordAccess table vs. checking SObject delete access when the ctrl is with sharing.

force-app/main/default/classes/CON_DeleteContactOverride_CTRL.cls
```diff
@@ -315,21 +371,10 @@ public with sharing class CON_DeleteContactOverride_CTRL {
...
-                SELECT Name, AccountId, Primary_Contact__c, Primary_Contact__r.AccountId, IsWon, IsClosed
-                FROM Opportunity
-                WHERE Primary_Contact__c = :contactToDelete.Id
-            ];
-            relatedRecurringDonations = [
-                SELECT Name, npe03__Contact__c
-                FROM npe03__Recurring_Donation__c
-                WHERE npe03__Contact__c = :contactToDelete.Id
-            ];
+            CON_DeleteContactOverrideSelector contactOverrideSelector = new CON_DeleteContactOverrideSelector();
```
kenlewis-sfdo, line 374: Instead of new'ing up additional instances of the selector class in these inner classes, the instance in the parent controller could be passed in as a second param to the constructor.  

`new ContactCascadeDelete(contactRecord, contactOverrideSelector);`

force-app/main/default/classes/CON_DeleteContactOverrideSelector.cls
npsp-reedestockton: SObject access check is sufficient if our only goal is to prevent the deletion, but it is a very poor user experience to display a user friendly message about inadequate permissions for a failing SObject access check and present the user with a default apex error on delete in the case of shared access failure.  That's what happens here if we do not check UserRecordAccess instead.  @jamesbsfdc called out the poor user experience issue for us during Sprint Review.

npsp-reedestockton:
@sam-knox Would you please take a look at the new label here?

@lparrott @dfuller114 @bdvorachek @kenlewis-sfdo Feedback has been implemented.  Very much appreciate the extremely helpful suggestions!!

bdvorachek (approved):

kenlewis-sfdo (approved):
Looks good, Reede! Thank you for making those updates!

PR #6786, opened 2021-12-22
Removing deprecated button exclusion to fix 236 deployment

lparrott:
I can't fully test this because I'm seeing "invalid cross reference id" when trying to deploy Apex in the pre-release org, but I believe that has to do with the org and not our code, since I can't manually create an Apex class either!

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


kenlewis-sfdo (approved):
I got the error message when deploying the main branch to a pre-release org, and then was able to deploy successfully to a pre-release org with this fix in place.  Thanks for the update, Luke!

PR #6791, opened 2022-01-06
Relationships Viewer - Handle Platform Encryption on Name Field

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Sprint__c/a0lEE0000001M6lYAE/view

Fixes an issue in relationship viewer where we're not able to sort the returned results by Name in an org with platform encryption enabled.
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/UTIL_Describe.cls
+++ b/force-app/main/default/classes/UTIL_Describe.cls
@@ -799,6 +799,10 @@ public class UTIL_Describe {
         return objectName;
     }
 
+    public Boolean isFieldSortable(Schema.SObjectField field) {
+        return field.getDescribe().isSortable();
+    }
+
     /** @description our exception object for Field Level & Object Security errors. */
     private class permsException extends Exception {}
 
--- a/force-app/main/default/classes/UTIL_Query.cls
+++ b/force-app/main/default/classes/UTIL_Query.cls
@@ -34,6 +34,17 @@
 * @description SOQL Builder
 */
 public with sharing class UTIL_Query {
+    @TestVisible
+    private UTIL_Describe describe {
+        get {
+            if(describe == null) {
+                describe = new UTIL_Describe();
+            }
+            return describe;
+        }
+        set;
+    }
+
     /** @description Exception raised by the Query builder */
     public class UTIL_QueryException extends Exception {}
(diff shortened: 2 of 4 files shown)


force-app/main/default/classes/UTIL_Query.cls
```diff
@@ -110,6 +121,20 @@ public with sharing class UTIL_Query {
...
     }
 
+    /**
+     * @description Registers an ORDER BY field only if that field is unencrypted and sortable.
+     * @param field A reference to the field being sorted on
+     * @param orderByClause The order by clause to add to the query
+     *
+     * @return UTIL_Query The instance of the Query builder
+     */
+    public UTIL_Query withOrderByEncryptable(Schema.SObjectField field, String orderByClause) {
```
dfuller114, line 131: I don't think we need the "encryptable" piece of the method name here. We could just have withOrderBy overloaded since the signature is different and we can just know to use this signature if we want to check if the field is sortable first since there are other instances where a field could not be sortable other than encryption such as long text area.

force-app/main/default/classes/UTIL_Describe.cls
```diff
@@ -799,6 +799,10 @@ public class UTIL_Describe {
         return objectName;
     }
 
+    public Boolean isFieldSortable(Schema.SObjectField field) {
+        return field.getDescribe().isSortable();
```
lparrott, line 803: I assume this takes into account the encryption status of the field/org? I don't see that referenced anywhere in the docs.

lparrott (approved):
Great fix! As long as my assumption about isSortable is correct, this looks good to go.

dfuller114 (changes requested):
looks great! had just one comment on a method name.

force-app/main/default/classes/UTIL_Describe.cls
bdvorachek: It does, if the field is encrypted then it won't be sortable. Other fields can return false for this always too, like Long or Rich Text.

dfuller114 (approved):

PR #6805, opened 2022-01-13
RD2 - Improve state management for Donor Fields

bdvorachek:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000cEORCYA4/view
Consolidate donor-specific state into a state object similar to Gift Entry’s formState

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_QueryService.cls
+++ b/force-app/main/default/classes/RD2_QueryService.cls
@@ -117,7 +117,7 @@ public inherited sharing class RD2_QueryService {
         Set<String> queryFields = new Set<String>();
 
         queryFields.addAll(rdFields);
-        queryFields.addAll((new RD2_CustomFieldMapper()).getRdFieldNames());
+        queryFields.addAll(customFieldMapper.getRdFieldNames());
         queryFields.addAll((new RD2_NamingService()).nameFormatter.fields);
         queryFields.add(getOpportunitySubQuery());
         queryFields.add(getScheduleSubQuery());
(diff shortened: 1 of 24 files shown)


force-app/main/default/classes/UTIL_Describe.cls
```diff
@@ -245,26 +247,32 @@ public class UTIL_Describe {
...
-                ret = ((string)(salesforceId)).startsWith(prefix);
+            String prefix = objectDescribes.get(objectName).getKeyPrefix();
+            if (prefix != null) {
+                ret = ((String) (salesforceId)).startsWith(prefix);
+            }
         }
         return ret;
     }
 
+    public static String objectNameFromId(Id objectId) {
```
lparrott, line 272: I'm surprised we didn't have this! Good addition.

force-app/main/default/classes/UTIL_Describe.cls
```diff
@@ -576,41 +591,29 @@ public class UTIL_Describe {
...
         }
     }
 
-    /*******************************************************************************************************
-    * @description utility to return a comma separated string of field names from a field set
-    * @param strObject the name of the object
-    * @param strFS the name of the object's field set
-    * @return a comma separated string of field names in the field set
-    */
-    public static string strFieldNamesFromFieldSet(string strObject, string strFS) {
```
lparrott, line 593: Heh, no references to this, huh?

lparrott (commented):
Solid work! I'm going to move on to testing and report back if I find anything.

force-app/main/default/classes/RD2_QueryService.cls
```diff
@@ -155,10 +162,12 @@ public inherited sharing class RD2_QueryService {
...
             .withFrom(npe03__Recurring_Donation__c.SObjectType)
             .withSelectFields(queryFields)
-            .withWhere('Id IN :rdIds')
+            .withWhere('Id = :recurringDonationId')
+            .withLimit(1)
             .build();
 
-        return Database.query(soql);
+        List<npe03__Recurring_Donation__c> results = Database.query(soql);
+        return results[0];
```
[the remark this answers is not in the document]
bdvorachek: It does, and there were test failures in this that happened as a result of doing this in tests that did not insert a recurring donation as part of test data setup, see [shouldReturnRD2CustomFieldsWithoutNPSPNameSpaceAndExcludedFields](https://github.com/SalesforceFoundation/NPSP/pull/6805/files#diff-d6e1da34d77650d75ecc81e48d283ffc98879c8b03448e76fe5d7f135574a337L212). My thinking was that any time we're querying on `Id`, that we can say that getting a result from it is an invariant/guaranteed.

This query only runs when we're trying to get the current database values for an RD that is being updated during a save operation.

force-app/main/default/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -158,56 +165,56 @@ export default class rd2EntryForm extends LightningElement {
...
     }
 
     /***
-    * @description Get settings required to enable or disable fields and populate their values
-    */
-    connectedCallback() {
+     * @description Get settings required to enable or disable fields and populate their values
+     */
+    async connectedCallback() {
         getRecurringSettings({ parentId: this.parentId })
```
andrewyu-salesforce, line 218: Since `getInitialView` is implemented, I think `getRecurringSettings` can be removed and reference the new function.

bdvorachek: Eventually we can decommission it, but at the moment `getRecurringSettings` still provides data for a few fields on rd2EntryForm. `customFields`, `isChangeLogEnabled`, `periodToYearlyFrequencyMap` to name a few. I can probably have it fully pulled out of here w/ the next WI that targets the Schedule Component.

force-app/main/default/classes/UTIL_Describe.cls
bdvorachek: Yep, I looked high and low for references when I saw it was greyed out in my IDE.

andrewyu-salesforce (approved):
The refactoring looks great.

lparrott (approved):

PR #6806, opened 2022-01-13
W-10269485 Fix Required Field Missing Error

kenlewis-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000Vizh8YAB/view

# Critical Changes

# Changes
 - We fixed an issue where some users were receiving a "Required fields are missing:Household_Account__c" or similar error when creating Contacts.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/adapter/in/sobjects/contact/Contacts.cls
```diff
@@ -336,4 +336,14 @@ public inherited sharing class Contacts extends fflib_SObjects {
        return getIdFieldValues(Contact.Id);
     }
 
+    public Contacts contactsWithNonNullAccountIds() {
+        List<Contact> contacts = new List<Contact>();
+        for (Contact c :(List<Contact>)  getRecords()) {
```
voduyemi, line 341: Can we replace the variable `c` with something more descriptive, like `contactWithAccount`

force-app/test/ContactInsertTests_TEST.cls
```diff
@@ -0,0 +1,41 @@
...
+            insert dupeContact;
+        } catch (Exception e) {
+            exceptionMessage = e.getMessage();
+        }
+
+        // Assert
+        if (exceptionMessage != null) {
+            // If it fails, assert that it is not due to the required fields missing error
+            System.assertEquals(
+                    false,
```
voduyemi, line 29: Not sure if we can prevent this, but could there be a chance that the error message for missing required fields might change in the future? If it does, this `assert` will fail

force-app/test/ContactInsertTests_TEST.cls
```diff
@@ -0,0 +1,41 @@
...
+                    exceptionMessage.contains(
+                            'Required fields are missing:Household_Account__c'
+                    ),
+                    'An address should not be created for a new Contact until ' +
+                            'its Household Account has been created.'
+            );
+        } else {
```
voduyemi, line 36: Since we would not be acting on this `else` block, we should remove it.

voduyemi (commented):
Good work Ken, testing this checks out. Left some comments around variable naming and the tests you included.

force-app/test/ContactInsertTests_TEST.cls
kenlewis-sfdo: In this case I don't think the assert would fail, it just wouldn't flag for us a regression. If the assert message is changed by the platform, then this test wouldn't fail even if somebody change the code back to reintroduce the regression. However I think that is a risk that is acceptable in order to give us a test that will flag a regression if a change is made to go back and attempt to insert addresses without account ids here.  See this section in the design doc: https://salesforce.quip.com/HO0zAflm4oyr#temp:C:PGB20c4f7a7cbcf9468e5a5a3967

force-app/test/ContactInsertTests_TEST.cls
kenlewis-sfdo: The else block is important here because other exception types could be caught but the test should still pass in that case. I added detail to the comment to explain why the test should pass when it reaches that point. Let me know if you think this is helpful.

voduyemi (approved):
Looks really good! Thanks Ken!

Shaws-WD:
We too are experiencing this error which is not solved by the [posted solution](https://help.salesforce.com/s/articleView?id=000363717&type=1)  and it's solution is critical to us.

We experience it when adding contacts via Salesforce REST API

jamesbsfdc:
@Shaws-WD  Please can you log a support case if you have not already done so as we do not provide support via GitHub. 

If you already have a support case can you post the case number? 

Thanks

PR #6810, opened 2022-01-13
W-7780834 BDI Donation Stage Defaults to "Opportunity Stage When Fully Paid" value

mldyang:
Fixed the issue where BDI Donation Stage ignores Opportunity Stage When Fully Paid value. Now, if "Opportunity Stage When Fully Paid" setting has a value, the default Donation Stage will be set to that value. 
(Note that DI's Donation Stage, if different than Opportunity Stage When Fully Paid value, will take precedence)
For reference, here is a table summarizing expected results https://salesforce.quip.com/hW4eA11a5zFd

# Critical Changes

# Changes
BDI Donation Stage will be defaulted to Opportunity Stage When Fully Paid value if one is available.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/BDI_Donations.cls
```diff
@@ -61,22 +61,40 @@ public class BDI_Donations {
...
-                if (stages.size() > 0) {
-                    defaultClosedWonStage = stages[0].MasterLabel;
+        get {
+            if (defaultClosedWonStage == null) {
+                // contact setting value takes priority
+                if (contactSettings.Payments_Auto_Close_Stage_Name__c != null) {
+                    defaultClosedWonStage = contactSettings.Payments_Auto_Close_Stage_Name__c;
                 } else {
-                    defaultClosedWonStage = '';
+                    // if no field is specified in contact setting, then query for the active isWon value and use that
```
callen-sfdo, line 86:
```suggestion
                    // if no stage name is specified in contact setting, then query for the active isWon value and use that
```

force-app/main/default/classes/BDI_Donations_TEST.cls
```diff
@@ -3092,6 +3092,80 @@ public with sharing class BDI_Donations_TEST {
         }
     }
 
+    /**
+     * @description If DI.Donation_Stage__c field has a value, use it for donation status, regardless of the status
+     *              in Contact Setting Payments_Auto_Close_Stage_Name__c
+     */
+    @IsTest
+    static void donationShouldUseDIDonationStage() {
```
callen-sfdo, line 3100: These two tests are good, but I would consider combining them into one unit test with two data import records to reduce the total number of DML statements and trigger run time (just concerned about total test execution time). Also you could consider just running these synchronously with BDI_DataImport_API.processDataImportRecords instead of batch execution which should save a little more processing time but not effect the outcome.

callen-sfdo (changes requested):
Looks good, just a few small things.

force-app/main/default/classes/BDI_Donations.cls
```diff
@@ -61,22 +61,40 @@ public class BDI_Donations {
...
+                    List<OpportunityStage> stages = [
+                            SELECT MasterLabel
+                            FROM OpportunityStage
+                            WHERE IsActive = TRUE
+                            AND IsWon = TRUE
+                    ];
+                    if (stages.size() > 0) {
```
grivero, line 93: We could replace these 5 lines of code with the following:
`defaultClosedWonStage = stages.size()>0 ? stages[0].MasterLabel : '';`

mldyang: Cool. Implemented, thanks!

PR #6811, opened 2022-01-14
Fix Gift Entry limit exception

### LINKED ISSUE #6757: Gift Entry: Users receiving a "System.LimitException: npsp:Too many SOQL queries: 101" error when processing a batch
Updated: 3 Dec 2021

### Background

Gift Entry allows a user to enter a new gift or to update an existing gift.  NPSP 3.209 introduced some new functionality to allow users to use a Soft Credit bundle within Gift Entry and also included some refactoring of existing Gift Entry code. 

### Issue Summary

One of the SOQL queries being executed when a user processes a Gift Entry batch is querying all NPSP Data Import records associated with the batch that have a value in the DonationImported__c field on NPSP Data Import. For each row returned we then query the associated opportunity and contact roles. Cumulatively the number of queries being executed will exceed the total number of SOQL queries governor limit if the number of records in the batch with DonationImported__c exceeds 45. 

https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_apexgov.htm

What this means in practice is you cannot process a batch if the batch contains 45 gifts that have either been processed or will update update existing gifts. 

Here are some examples: 

- Batch A contains 45 unprocessed new gifts. The user should be able to process these gifts however if after processing the 45 gifts they create a new gift in the batch and attempt to process the batch again they will receive the SOQL error.

- Batch B contains 45 unprocessed existing gifts that correspond to existing open opportunities (i.e the user wants these opportunities to be updated to indicate payment has been received). The user should be able to process these gifts however if after processing the 45 gifts they create a new gift or add a gift that should update an existing open opportunity and attempt to process the batch again they will receive the SOQL error. 

- Batch C contains 100 unprocessed new gifts. The user should be able to process the batch however if after processing the gifts they add any new gifts to the batch they will be unable to process the batch again. 

- Batch D contains 100 unprocessed existing gifts that correspond to existing open opportunities (i.e the user wants these opportunities to be updated to indicate payment has been received). The user will not be able to process the batch. They will need to reduce the batch to 45 or fewer before they can process the batch. 

### Steps to Reproduce

1. Create a new Gift Entry batch
2. Import and process 45 records within the batch such that all records have a status of Imported
3. Create a new gift within the batch and attempt to process this
4. Confirm you receive a "npsp:Too many SOQL queries: 101" error

### Workaround
Process records via a new batch rather than an existing one if the batch already contains 45 or more records with value in the DonationImported__c field on NPSP Data Import.

For example: If you want to add 100 or more records to a batch that already contains 45 processed records, create a new batch, add the 100 records to the new batch and process these in bulk. Also see the examples above for scenarios.

Another option is to add gifts to a spreadsheet and use NPSP Data Import to process the records rather than Gift Entry
https://powerofus.force.com/s/article/NPSP-Configure-Data-Importer-Options
https://trailhead.salesforce.com/content/learn/projects/import-your-data-using-npsp-data-importer

psadie-sfdo:
[W-10244877](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000UDD6FYAX/view)

# Critical Changes

# Changes

# Issues Closed
- Fixes #6757 
# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/Donations_TEST.cls-meta.xml
+++ b/force-app/main/default/classes/Donations_TEST.cls-meta.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
-    <apiVersion>52.0</apiVersion>
+    <apiVersion>53.0</apiVersion>
     <status>Active</status>
 </ApexClass>
--- a/force-app/main/default/classes/Donations_TEST.cls-meta.xml
+++ b/force-app/main/default/classes/Donations_TEST.cls-meta.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
+    <status>Active</status>
+</ApexClass>
--- a/force-app/main/default/classes/Gift_Test.cls
+++ b/force-app/main/default/classes/Gift_Test.cls
@@ -32,6 +32,32 @@
 private class Gift_Test {
 
     @isTest
+    static void shouldPopulateDonationInGift(){
+        // Arrange
+        Id dummyDataImportId = UTIL_UnitTestData_TEST.mockId(DataImport__c.SObjectType);
+        Id dummyOpportunityId = UTIL_UnitTestData_TEST.mockId(Opportunity.SObjectType);
+        DataImport__c dummyDataImport = new DataImport__c(
+            Id = dummyDataImportId,
+            Donation_Amount__c = 3.00,
+            Donation_Date__c = Date.today()
+        );
+        Opportunity dummyOpportunity = new Opportunity(
+            Id = dummyOpportunityId,
+            Amount = 100.00,
+            StageName = 'Open'
+        );
+        Donation dummyDonation = new Donation(dummyOpportunity);
+        Gift gift = new Gift(dummyDataImport, dummyDonation);
+
+        // Act
+        Donation retrievedDonation = gift.donation();
+
+        // Assert
+        System.assertEquals(dummyOpportunityId, retrievedDonation.opportunity().Id,
+            'Should return the donation associated with the gift.');
+    }
+
+    @isTest
     static void validateThatGiftNeedsToBeReauthorized() {
         DataImport__c dataImport = new DataImport__c(
             Payment_Status__c = ElevateTransactionStatus.AUTHORIZED.name(),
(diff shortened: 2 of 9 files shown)


force-app/main/default/classes/Gifts_TEST.cls
```diff
@@ -241,14 +338,57 @@ private class Gifts_TEST {
         return failReasonByElevateBatchId;
     }
 
+    public class DonationSelectorMock implements StubProvider {
+
+        public Integer numberOfQueriesRan = 0;
```
kenlewis-sfdo, line 343: This is actually asserting the number of times called for the getdonation or getdonations methods (not directly testing the # of queries called).  Should probably clarify that distinction in the tests.

kenlewis-sfdo (approved):
Looks great Prince.  Thanks for these changes!

force-app/main/default/classes/Donations_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
force2b, line 3: I believe this should be at 53 for new classes.

force-app/main/default/classes/Gift_Test.cls
```diff
@@ -31,6 +31,32 @@
...
+        Id dummyOpportunityId = UTIL_UnitTestData_TEST.mockId(Opportunity.SObjectType);
+        DataImport__c dummyDataImport = new DataImport__c(
+            Id = dummyDataImportId,
+            Donation_Amount__c = 3.00,
+            Donation_Date__c = Date.today()
+        );
+        Opportunity dummyOpportunity = new Opportunity(
+            Id = dummyOpportunityId,
+            Amount = 100.00,
+            StageName = 'Open'
```
force2b, line 47: I don't this stage exists anywhere. Probably not a big issue if you're not inserting for the test, but if you need to there is `UTIL_UnitTestData_TEST.getOpenStage()`.

force2b (commented):
Looks very good. One requested change and a comment. I'm assuming the main change is in the `force-app/main/default/classes/DonationSelector.cls` file where it changes the query for a single gift to now support a collection of opp ids to query for.

force2b:
Also, I like the use of the stub providers to inject a mocked selector class.

force2b (approved):
Great work on this!

PR #6813, opened 2022-01-18
W-10098593: Donation History - Amount and Payment Method

JoseRossiSF:
# Work Item
[W-10098593](https://gus.lightning.force.com/a07EE00000K5HBHYA3)

# Changes
Modified query to bring Payment records alongside Opportunity records. Component can now tell whether Payments are active in org or not and show Payment information accordingly.

PR #6814, opened 2022-01-18
[W-10373214] Donation History infinite scroll

mcasco:
# Critical Changes

Add infinite scroll to Donation history table component

# Changes

Add method `getTotalNumberOfRecords` and change query on `getDonationHistory` to implement infinite scroll on LWC component `donationHistoryTable`


mcasco:
[W-10373214](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000cAcBEYA0/view)

JoseRossiSF (approved):
LGTM

force-app/main/service/DonationHistoryService.cls
```diff
@@ -84,9 +85,19 @@ public with sharing class DonationHistoryService {
...
         }
     }
+    
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return Donation History Number of Records for infinite scroll.
+    * @param contactId contact id which is going to be requested.
+    */
+    public List<AggregateResult> getTotalNumberOfRecords() {
```
dcaballeroSf, line 100: I think this service should return the number(Integer) and not the AggregateResult.

force-app/main/default/classes/DonationHistoryController_TEST.cls
```diff
@@ -137,7 +137,7 @@ public class DonationHistoryController_TEST {
             Test.startTest();
             System.runAs(u){
                 List<DonationHistoryController.DonationHistoryView> donationHistory = 
-                DonationHistoryController.getDonationHistory(testContact.Id); 
+                DonationHistoryController.getDonationHistory(testContact.Id, 0); 
             }
 	        Test.stopTest();
         } catch(Exception e){
```
dcaballeroSf, line 143: dont we need a new test for the new method?

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -48,12 +48,17 @@ export default class DonationHistoryTable extends LightningElement {
     
     // eslint-disable-next-line @lwc/lwc/no-async-await
     async connectedCallback() {
-        getDonationHistory({contactId: this.contactId})
+        getTotalNumberOfRecords({contactId: this.contactId})
+        .then(data => {
+            if(data){
+                this.totalNumberOfRows = data[0].expr0;
```
dcaballeroSf, line 54: I think this has to be managed by the service in the backend

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -52,6 +52,21 @@ public inherited sharing class DonationHistorySelector {
...
+            OFFSET :offset
         ];
     }
+
+    /** 
+    * @author Salesforce.org
+    * @date 2021
+    * @description method to return total number of record
+    */
+    public List<AggregateResult> getTotalNumberOfRecords(Id contactId){
```
dcaballeroSf, line 64: add white space between ) and {

dcaballeroSf (changes requested):
Some changes to do

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -5,13 +5,13 @@ import PAYMENT_FIELD from '@salesforce/schema/DataImport__c.Payment_Method__c';
 import donationHistoryDatatableAriaLabel from '@salesforce/label/c.donationHistoryDatatableAriaLabel';
 import RD2_ScheduleVisualizerColumnDate from '@salesforce/label/c.RD2_ScheduleVisualizerColumnDate';
 import getDonationHistory from '@salesforce/apex/DonationHistoryController.getDonationHistory';
+import getTotalNumberOfRecords from '@salesforce/apex/DonationHistoryController.getTotalNumberOfRecords';
```
bdvorachek, line 8: I had another suggestion in [PR #6813](https://github.com/SalesforceFoundation/NPSP/pull/6813/files#r787308992) on how we should be trying to retrieve additional information from the back-end that I think is worth implementing here too. Lets include the `totalNumberOfRecords` as part of the response returned by `getDonationHistory`.

Instead of returning a `List<DonationHistoryView>` from `getDonationHistory`, we'd return a `DonationHistoryAppView` that looks like this:

```
public class DonationHistoryAppView {
  @AuraEnabled List<DonationHistoryView> donations;
  @AuraEnabled Integer totalNumberOfRecords;
}
```

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
bdvorachek: I left a bit of feedback above on this, but I think that `getDonationHistory()` should return as much of the data needed for the initial load of the app as possible, and the total number of records is a good candidate for including in that response.

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
```diff
@@ -65,17 +70,19 @@ export default class DonationHistoryTable extends LightningElement {
...
-        //Appends new data to the end of the table
-        const newData = currentData.concat(this.allData.slice(current, offset));
-        this.data = newData;
-        event.target.isLoading = false;
+        this.tableElement = event.target;
+        getDonationHistory({contactId: this.contactId, offset: this.data.length})
+        .then(data => {
+                if (this.data.length >= this.totalNumberOfRows) {
+                    this.tableElement.isLoading = false;
+                    this.tableElement.enableInfiniteLoading = false;
```
bdvorachek, line 78: I see that the docs for lightning-datatable suggest using `event.target`, but I think it's better to bind properties in the template like this:

```suggestion
                    this.isLoading = false;
                    this.enableInfiniteLoading = false;
```

Then in the template add: `enable-infinite-loading=(enableInfiniteLoading}`

bdvorachek (changes requested):

force-app/main/service/DonationHistoryService.cls
mcasco: done

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: when implementing this the component does not take the assigned values and apparently the reference is not working, both for isLoading and for enableInfiniteLoading

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: response changed!

force-app/main/default/lwc/donationHistoryTable/donationHistoryTable.js
mcasco: done

force-app/main/default/classes/DonationHistoryController_TEST.cls
mcasco: change the implementation of the test so that it can correctly evaluate the new response with the data of the number of records and the list of donations

bdvorachek (approved):

psadie-sfdo (approved):

dcaballeroSf (approved):

jofsky (approved):

PR #6815, opened 2022-01-18
[ProductHealth -- Trust] Improve Contact Merge Apex Tests Performance

dfuller114:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


dfuller114:
This PR is still very much in flight but wanted to get up some of the changes made to setup for mocking as much as we can.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
```diff
@@ -336,6 +333,7 @@ public with sharing class CON_ContactMerge_CTRL {
     /***********************************************************************************************
     * @description The rows to merge.
     */
+    @TestVisible
```
lparrott, line 336: Should these be switched to private if they're marked TestVisible?

force-app/main/default/classes/CON_ContactMerge.cls
```diff
@@ -36,6 +36,28 @@
 * @group-content ../../ApexDocContent/ContactMerge.htm
 */
 public class CON_ContactMerge {
+
+    @TestVisible
+    private ContactSelector contactSelector {
```
kenlewis-sfdo, line 41: The class names AccountSelector and ContactSelector should be reserved for potential use at the adapter or domain layers. Any queries factored out of these classes related to contact merging should be added to the ContactMergeSelector that is also introduced here in this PR.

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
kenlewis-sfdo: Agreed.  If they stay public then they don't need the test visible. If they're accessed only in test then they should be made private and test visible.

dfuller114:
> Please put all of the queries pulled out of the controller into the new ContactMergeSelector class. In the long run, we want controller methods to be calling service methods instead of querying directly. A custom selector just for this controller is a good interim step that enables mocking for methods in this controller. But the contact selector and account selector classes should be reserved for primary use with domains and services not for use with individual controllers.

@kenlewis-sfdo just pushed these changes up!

force-app/main/default/classes/CON_ContactMerge_CTRL.cls
dfuller114: The reason this has the annotation still even though it's public is because the `set` property is set to `private` so it functionally operates like a private property still and cannot be changed publicly without that annotation.

dfuller114: I didn't want to make the entire property `private` since that would likely interfere with performing a `get` from the VF page.

lparrott (approved):
Looks great, I'm moving this one onto QA, I'll keep you posted!

force-app/main/selector/ContactSelector.cls
```diff
@@ -167,6 +168,18 @@ public inherited sharing class ContactSelector {
                 .build();
     }
 
+    @TestVisible
```
kenlewis-sfdo, line 171: Are the changes in this file still need it? If so can they be moved to the contact merge selector?

force-app/main/default/classes/CON_ContactMerge.cls
```diff
@@ -36,6 +36,17 @@
 * @group-content ../../ApexDocContent/ContactMerge.htm
 */
 public class CON_ContactMerge {
```
kenlewis-sfdo, line 38: What do you think about renaming this class? "ContactMergeService"

force-app/main/default/classes/OrgConfig.cls
```diff
@@ -61,6 +61,11 @@ public inherited sharing class OrgConfig {
         return validCountriesByCode;
     }
 
+    //todo: add test coverage
+    public Boolean isPersonAccountsEnabled() {
```
kenlewis-sfdo, line 65: If this method is going to be added here then I think both the selector and the "contact merge" class should both reference this same method.

force-app/main/default/classes/CON_ContactMerge.cls
```diff
@@ -79,23 +90,16 @@ public class CON_ContactMerge {
     
         Id winningAccountId = null;
         Boolean winnerAccountIsOneToOne = false;
-        Contact winner = [SELECT AccountId,Account.npe01__SYSTEM_AccountType__c FROM Contact WHERE Id =:winningContact.Id];
+        Contact winner = contactMergeSelector.getContactsWithAccountFields(new List<Contact>{winningContact})[0];
 
         // detect PersonAccounts
-        boolean isPAEnabled = UTIL_Describe.isValidField('Account', 'isPersonAccount');
-        
+        Boolean isPAEnabled = UTIL_Describe.isValidField('Account', 'isPersonAccount');
```
kenlewis-sfdo, line 96: Here is the other reference to checking whether a person accounts are enabled.

force-app/main/default/classes/OrgConfig.cls
```diff
@@ -61,6 +61,11 @@ public inherited sharing class OrgConfig {
...
+        return UTIL_Describe.isValidField(String.valueOf(Account.SObjectType), 'isPersonAccount');
```
kenlewis-sfdo, line 66: Another reference to update to use the OrgConfig method instead.

force-app/main/default/classes/UnitOfWork.cls
```diff
@@ -41,12 +41,14 @@ public inherited sharing class UnitOfWork {
 
     public Database.SaveResult[] save() {
         List<Database.SaveResult> saveResults = new List<Database.SaveResult>();
+
```
kenlewis-sfdo, line 44: If there are no changes to this file could it be reverted so that these diff lines go away?

force-app/main/default/classes/CON_ContactMerge_TEST2.cls
```diff
@@ -50,59 +50,61 @@ private class CON_ContactMerge_TEST2 {
...
-        
+
         // make sure addresses created
         if (strProcessor == CAO_Constants.HH_ACCOUNT_PROCESSOR) {
             system.assertEquals(3, [select count() from Address__c]);
         }
 
-        list<Contact> listCon = [Select id, accountId, FirstName, LastName, MailingStreet, MailingCity from Contact order by Id];
+        list<Contact> listCon = [Select id, accountId, FirstName, LastName, MailingStreet, MailingCity, Primary_Affiliation__c from
+                Contact order by Id];
```
kenlewis-sfdo, line 92: What is the reason for adding an additional field to the query here?

force-app/main/default/classes/CON_ContactMerge_TEST2.cls
```diff
@@ -143,29 +145,33 @@ private class CON_ContactMerge_TEST2 {
...
-        
-        con = [Select id, accountId, FirstName, LastName, MailingStreet from Contact where Id = :con.id];
-        con2 = [Select id, accountId, FirstName, LastName, MailingStreet from Contact where Id = :con2.id];
-        
-        CON_ContactMerge_CTRL controller = setupPageController(new List<Contact> { con, con2 }, 'test');
-        
+
+        con = [Select id, accountId, FirstName, LastName, MailingStreet, MailingCity, Primary_Affiliation__c from Contact where Id =
+                :con.id];
+        con2 = [Select id, accountId, FirstName, LastName, MailingStreet, MailingCity, Primary_Affiliation__c from Contact where Id =
```
kenlewis-sfdo, line 169: Here also. What prompted the change to add a field to the query?

force-app/main/default/classes/CON_ContactMerge_TEST2.cls
```diff
@@ -839,7 +836,7 @@ private class CON_ContactMerge_TEST2 {
             insert objectsToInsert;
             
             List<Contact> contactsToMerge = [
-                SELECT Id, AccountId, FirstName, LastName, MailingStreet, Description , Primary_Affiliation__c
+                SELECT Id, AccountId, FirstName, LastName, MailingStreet, Description , Primary_Affiliation__c, MailingCity
```
kenlewis-sfdo, line 839: And here. Just curious to learn a little bit more about why these query changes were needed.

lparrott (approved):

kenlewis-sfdo (approved):
Looks great, Daniel!  Thanks for making those last few changes.  Good work!

PR #6816, opened 2022-01-18
Mark Household All Members Deceased

npsp-reedestockton:
Added All Members Deceased flag to Account and aggregated the Deceased flag on member households after contact insert, contact delete, contact addition to, and contact removal from a household.

# Critical Changes

# Changes
 - We introduced new feature to help you manage deceased Contacts. When all Contacts within a Household are marked as deceased, NPSP selects the All Members Deceased checkbox on the Household Account. Use this field in report and list view filters to remove deceased members from your mailing lists.

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Field
- Account.All_Members_Deceased__c

# Deleted Metadata

### DIFF
--- a/force-app/main/default/layouts/Account-Household Layout.layout-meta.xml
+++ b/force-app/main/default/layouts/Account-Household Layout.layout-meta.xml
@@ -29,6 +29,10 @@
                 <behavior>Edit</behavior>
                 <field>npo02__Informal_Greeting__c</field>
             </layoutItems>
+            <layoutItems>
+                <behavior>Edit</behavior>
+                <field>All_Members_Deceased__c</field>
+            </layoutItems>
         </layoutColumns>
         <layoutColumns>
             <layoutItems>
--- a/force-app/main/default/layouts/Account-Household Layout.layout-meta.xml
+++ b/force-app/main/default/layouts/Account-Household Layout.layout-meta.xml
@@ -29,10 +29,6 @@
                 <behavior>Edit</behavior>
                 <field>npo02__Informal_Greeting__c</field>
             </layoutItems>
-            <layoutItems>
-                <behavior>Edit</behavior>
-                <field>All_Members_Deceased__c</field>
-            </layoutItems>
         </layoutColumns>
         <layoutColumns>
             <layoutItems>
--- a/force-app/main/default/objects/Account/fields/All_Members_Deceased__c.field-meta.xml
+++ b/force-app/main/default/objects/Account/fields/All_Members_Deceased__c.field-meta.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>All_Members_Deceased__c</fullName>
+    <defaultValue>false</defaultValue>
+    <description>All members of this household are deceased.</description>
+    <externalId>false</externalId>
+    <inlineHelpText>All members of this household are deceased. This field should not be directly updated. It is updated automatically as a result of updates to the Deceased checkbox on member Contacts.</inlineHelpText>
+    <label>All Household Members Deceased</label>
+    <trackFeedHistory>false</trackFeedHistory>
+    <type>Checkbox</type>
+</CustomField>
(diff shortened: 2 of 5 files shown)


npsp-reedestockton:
Tracking [W-1034272](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000ZSLWtYAP/view)

force-app/main/default/objects/Account/fields/All_Members_Deceased__c.field-meta.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>All_Members_Deceased__c</fullName>
+    <defaultValue>false</defaultValue>
+    <description>All members of this household are deceased.</description>
```
sam-knox, line 5:
```suggestion
    <description>If selected, all members of this household are deceased. Do not update this field manually. It's updated automatically by NPSP as a result of updates to the Deceased checkbox on household member Contacts.</description>
```

npsp-reedestockton:
@lparrott @bdvorachek Would appreciate your review comments.  Thanks!

npsp-reedestockton:
@sam-knox I've made the changes you requested.  Thank you for your review!!

force-app/main/service/HouseholdNamingService.cls
```diff
@@ -124,6 +124,11 @@ public without sharing class HouseholdNamingService {
             setHouseholdNameFieldValues(householdsOrAccounts, membersByHouseholdId);
         }
         setNumberOfHouseholdMembers(householdsOrAccounts, membersByHouseholdId);
+
+        if (!householdsOrAccounts.isEmpty() && householdsOrAccounts?.get(0).getSObjectType() == Account.SObjectType) {
```
bdvorachek, line 128:
```suggestion
        if (isListOfAccountIds(householdOrAccountIds)) {
```

force-app/main/service/HouseholdNamingService.cls
```diff
@@ -168,6 +173,32 @@ public without sharing class HouseholdNamingService {
         }
     }
 
+    public void setAllMembersDeceasedFlag(List<SObject> householdsOrAccounts, Map<Id, List<Contact>> membersByHouseholdId) {
```
bdvorachek, line 176: I think we should make it clear this method only works for Accounts, and do the cast to `List<Account>` when calling this method.
```suggestion
    public void setAllMembersDeceasedFlag(List<Account> accounts, Map<Id, List<Contact>> membersByHouseholdId) {
```

force-app/main/service/HouseholdNamingService.cls
npsp-reedestockton: Not sure we can eliminate this part - householdsOrAccounts?.get(0).getSObjectType() == Account.SObjectType.  Household object Ids get to this point, so we need to ensure we are only processing Accounts.

npsp-reedestockton: I initially misread your comment.  I see now that I just need to extract the logic.  Good call!!

force-app/main/default/layouts/Account-Household Layout.layout-meta.xml
```diff
@@ -29,6 +29,10 @@
                 <behavior>Edit</behavior>
                 <field>npo02__Informal_Greeting__c</field>
             </layoutItems>
+            <layoutItems>
+                <behavior>Edit</behavior>
+                <field>All_Members_Deceased__c</field>
```
lparrott, line 34: We aren't including this in our default Contact Layout, so I think it's worth a quick discussion if we want it add here (and exactly where to put it in the Layout!)

npsp-reedestockton: It might be good to pull it from this PR and add it in either of the next 2 sprints if we decide we want it.

npsp-reedestockton:
@lparrott @bdvorachek I believe all the feedback has been implemented now.  Would appreciate another look.  Thanks!

sam-knox:
> 

You bet! I was just going through the open PRs and decided to add my label review.

lparrott (approved):
Looks good! One thing to consider for the next WI is a Unit test to make sure the All Members Deceased checkbox gets set correctly even if the User doesn't have permission to that field.

bdvorachek (approved):

PR #6824, opened 2022-01-25
Removing "total" from RD Migration JSON to free up space in field

### LINKED ISSUE #6808: Recurring Donations: Customers with greater than 99,999 Recurring Donations will receive a "STRING_TOO_LONG" error when they upgrade to Enhanced Recurring Donations
### Background

During the Enhanced Recurring Donations upgrade process we store the results of the Dry Run and the Data Migration jobs as JSON in the "Enhanced RD Data Migration State" field on the Recurring Donations custom setting. The character limit for this field is 255 characters.

### Issue Summary

If a customer who has more than 99,999 Recurring Donations tries to upgrade they will receive an error when the Data Migration completes as the "Enhanced RD Data Migration State" field cannot hold the JSON data associated with both the Dry Run and Data Migration job results.

e.g for a customer with 100,000 Recurring Donations the following sample would total up to 258 characters

{"run":{"total":100000,"processed":100000,"failed":0,"completedDate":"2022-01-13T15:26:27.078Z","batchId":"70754000032i2O0AAI"},"dryRun":{"total":100000,"processed":100000,"failed":0,"completedDate":"2022-01-13T15:18:07.525Z","batchId":"70754000032i0g7AAA"}}

### Steps to Reproduce

1. Create a scratch org that is using Legacy Recurring Donations
2. Insert 100,000 Recurring Donations
3. Run the Dry Run validation job, wait for it to complete and then run the Data Migration job
4. Confirm you receive a STRING_TOO_LONG error on the Recurring Donation upgrade page in NPSP Settings when the Data Migration job completes, and an associated error is created in the NPSP Error log object

### Workaround

The error message can be ignored as it only relates to the update of the custom setting value.

lparrott:
- We removed the "total" count of processed records from the Enhanced Recurring Donation migration JSON in custom settings, since it can be calculated using the failed and processed amounts.

# Critical Changes

# Changes

# Issues Closed

Fixes #6808 

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/lwc/batchProgress/batchProgress.js
```diff
@@ -96,6 +98,13 @@ export default class BatchProgress extends LightningElement {
         }, this.pollingTimeout, self);
     }
 
+    /***
+    * @description Sets the Total Records value based on Processed and Failed
+    */
+    setTotalRecords() {
+        this.batchJob.summary.total = this.batchJob.summary.processed + this.batchJob.summary.failed;
+    }
```
bdvorachek, line 106: I think I'd change this up a tiny bit to add that total property on to the object before assigning it to `this.batchJob`. Here's a suggestion below:

```suggestion
    setTotalRecords(data) {
        if(!isNull(data)) {
           data.summary.total = data.summary.processed + data.summary.failed;
           return data;
        }
        return null;
    }
```

Then with that, we could do something like:

`this.batchJob = setTotalRecords(JSON.parse(data));`

dfuller114 (approved):
looks good!

npsp-reedestockton (approved):
Great solution - simple and effective!

PR #6831, opened 2022-02-01
[ProductHealth -- Release Advantage] Expose EnableAdvancedDataMapping and EnableGiftEntry as Callable Api

dfuller114:
# Critical Changes

# Changes

Add new Cumulus CI tasks to the cumulus.yml to enable Advanced Mapping and Gift Entry via the Apex Callable API.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/AdvancedMappingEnablementService_TEST.cls
```diff
@@ -0,0 +1,78 @@
+@IsTest
```
force2b, line 1: I think this class still needs the copyright header

force-app/main/default/classes/GiftEntryEnablementService.cls
```diff
@@ -0,0 +1,47 @@
+/*
+    Copyright (c) 2020, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2022, Salesforce.org
```

force-app/main/default/classes/GiftEntryEnablementService.cls
```diff
@@ -0,0 +1,47 @@
...
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
```
force2b, line 32:
```suggestion
* @date 2022
```

force-app/main/default/classes/AdvancedMappingEnablementService.cls
```diff
@@ -0,0 +1,63 @@
+/*
+    Copyright (c) 2019, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2022, Salesforce.org
```

force-app/main/default/classes/AdvancedMappingEnablementService.cls
```diff
@@ -0,0 +1,63 @@
...
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+    * @author: Salesforce.org
+    * @date: 2021
```
force2b, line 32:
```suggestion
    * @date: 2022
```

force-app/main/default/classes/GiftEntryEnablementService.cls
```diff
@@ -0,0 +1,47 @@
...
+* @description Enablement Service for the Gift Entry feature
+*/
+public with sharing class GiftEntryEnablementService {
+
+    public Boolean isEnabled() {
+        return UTIL_CustomSettingsFacade.getGiftEntrySettings().Enable_Gift_Entry__c;
+    }
+
+    public void enable() {
```
force2b, line 41: If I'm not mistaken, GiftEntry cannot be enabled unless Advanced Mapping has been enabled. If so, we should have a check for it and throw and exception if this is called before Advanced Mapping has been enabled.

force-app/main/default/classes/GiftEntryEnablement_TEST.cls
```diff
@@ -0,0 +1,42 @@
+/*
+    Copyright (c) 2020, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2022, Salesforce.org
```

cumulusci.yml
```diff
@@ -938,6 +938,71 @@ flows:
             2:
                 task: update_admin_profile
 
+    enable_gift_entry_managed:
```
force2b, line 941: Can you also add these to the `documentation/automation.md` file?

force2b (changes requested):
Looking pretty good. Made a few minor suggestions

force-app/main/default/classes/Callable_API.cls
```diff
@@ -159,7 +182,25 @@ global with sharing class Callable_API implements System.Callable {
                 // It returns the Job ID of the Scheduled Job.
                 return CallableApiApexDelegationService.handleApexScheduleJob(params);
 
-            } when else {
+            } when 'settings.enableadvancedmapping' {
+                if (advancedMappingEnablementService.isEnabled()) {
+                    return null;
+                }
++
```
lparrott, line 189: Rogue + sign?

cumulusci.yml
```diff
@@ -938,6 +938,71 @@ flows:
...
+        group: NPSP
+        steps:
+            1:
+                flow: enable_advanced_mapping_managed
+            2:
+                task: execute_anon
+                options:
+                    managed: true
+                    unmanaged: false
+                    apex: >
```
lparrott, line 952: We might as well create a task for this Apex since we're calling it twice.

dfuller114: @lparrott I thought about this a little more, and I think it doesn't make as much sense to split this out into a task since gift entry shouldn't ever be enabled without first enabling advanced mapping, and splitting this out into its own task would imply it can be executed in isolation. Thoughts?

lparrott: Good point, there might be a way to write the managed version as an "overload" of the unmanaged, like you did with `enable_advanced_mapping_managed` and avoid using the apex block twice, but I wouldn't spend much time on that if it isn't straightforward!

lparrott (approved):

force-app/main/default/classes/Callable_API_TEST.cls
```diff
@@ -117,6 +119,53 @@ private with sharing class Callable_API_TEST {
...
+        Test.startTest();
+            callableApi.call(ADVANCED_MAPPING_ACTION, null);
+        Test.stopTest();
+
+        System.assert(!mappingServiceMock.enableAdvancedMappingCalled, 'The enable advanced mapping service method ' +
+                'should not be called.');
+    }
+
+    @IsTest
+    private static void shouldCallEnableGiftEntryOnAction() {
```
bdvorachek, line 156: I see you got the already enabled scenario covered for Advanced Mapping, but it'd be nice to have it for Gift Entry too. Great test coverage otherwise!

cumulusci.yml
bdvorachek: How about moving this apex into a file and using the `path:` param on these two tasks? The scripts directory has a few of these already.

bdvorachek (changes requested):

force-app/main/default/classes/GiftEntryEnablementService.cls
dfuller114: good point

force-app/main/default/classes/GiftEntryEnablementService.cls
```diff
@@ -33,15 +33,32 @@
...
+        set;
+    }
 
     public Boolean isEnabled() {
         return UTIL_CustomSettingsFacade.getGiftEntrySettings().Enable_Gift_Entry__c;
     }
 
     public void enable() {
+        if (!advancedMappingEnablementService.isEnabled()) {
+            throw new GiftEntryEnablementException('Advanced mapping is not enabled. First enable advanced mapping ' +
```
force2b, line 53: This should be a custom label instead of hardcoded text.

force2b (changes requested):
Thanks for making the changes. One small change requested to use a custom label for the error.

force-app/main/default/classes/Callable_API_TEST.cls
```diff
@@ -117,6 +119,68 @@ private with sharing class Callable_API_TEST {
...
+    @IsTest
+    private static void shouldNotCallEnableGiftEntryIfAlreadyEnabled() {
+        GiftEntryEnablementServiceMock giftEntryEnablementServiceMock = new GiftEntryEnablementServiceMock();
+        giftEntryEnablementServiceMock.setEnabled();
+
+        Callable_API callableApi = new Callable_API();
+        callableApi.giftEntryEnablementService = stubFor(giftEntryEnablementServiceMock);
+
+        Test.startTest();
+            callableApi.call(ADVANCED_MAPPING_ACTION, null);
```
force2b, line 177: Do you want to verify that the call throws the exception (and wouldn't this fail since an exception is thrown)?

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -8802,6 +8802,14 @@ If preferred, you can disable Gift Entry and instead use older Batch Gift Entry
         <shortDescription>Page-level error message body when Advanced Mapping and Gift Entry are off</shortDescription>
         <value>This feature requires both Advanced Mapping and Gift Entry. Please enable them in NPSP Settings.</value>
     </labels>
+    <labels>
+        <fullName>geServiceLevelErrorAdvancedMapping</fullName>
+        <categories>Gift Entry</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Service-level error message when advanced mapping is not enabled.</shortDescription>
+        <value>Advanced mapping is not enabled. First enable advanced mapping before gift entry can be enabled.</value>
```
force2b, line 8811: @sam-knox Would you mind doing a quick review on this label?

force-app/main/default/classes/Callable_API_TEST.cls
dfuller114: @force2b In these tests, we are only verifying that the method is called since the primary responsibility of the Callable_API is to call the methods. And since the method is stubbed, it's just returning a "called" `boolean` so it wouldn't fail. Then we have the underlying service test class to actually test the method execution.

force-app/main/default/classes/Callable_API_TEST.cls
dfuller114: added!

jstvz (approved):
CODEOWNERS :ok:

force-app/main/default/classes/Callable_API_TEST.cls
force2b: Ah. Very good pont.

force2b (approved):

bdvorachek (approved):

cumulusci.yml
```diff
@@ -938,6 +938,40 @@ flows:
             2:
                 task: update_admin_profile
 
+    enable_gift_entry:
+        description: Enable the Gift Entry feature (unmanaged code only).
```
bdvorachek, line 942: I think we can drop the part about this only working for unmanaged code in both of these flows.
```suggestion
        description: Enable the Gift Entry feature.
```

force-app/main/default/labels/CustomLabels.labels-meta.xml
sam-knox, line 8811:
```suggestion
        <value>You must enable Advanced Mapping before you can enable Gift Entry.</value>
```

sam-knox (changes requested):
Ok, I added my updated error message text.

sam-knox (approved):

PR #6839, opened 2022-02-07
[CAB] RD2: TDTM Caching Old List Resulting RD2 Schedule Update Fail

andrewyu-salesforce:
[W-10518958](https://github.com/SalesforceFoundation/NPSP/pull/6839#issue-1126570053)

# Critical Changes
Fixes the issue during RD2 bulk update with All or None set to false, 1 invalid record will cause other RD schedules to not update.

# Changes

# Issues Closed
[Issue](https://trailblazer.salesforce.com/issues_view?id=a1p4V0000024G3nQAE&title=changes-to-recurring-donation-fields-are-not-being-applied-to-recurring-donation-schedules-or-installment-opportunities)

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -40,6 +40,8 @@ public class TDTM_TriggerHandler {
     
     private static Boolean defaultRecordsInserted = false;
     
+    private static Boolean preventCachingFalseTransaction = false;
```
andrewyu-salesforce, line 43: End up having to do a static variable to preserve the flag from Before to After Context. The `addError` is called in Before Context. However, the caching is happening in the after context.

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -351,37 +375,41 @@ public class TDTM_TriggerHandler {
...
         }
 
         // now cache objects for future Update calls
         if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-            mapDescObjToMapOldForUpdates.put(describeObj, new Map<Id, SObject>(newList));
+            if (!preventCachingFalseTransaction) {
+                mapDescObjToMapOldForUpdates.put(describeObj, new Map<Id, SObject>(newList));
+            }
+
+            preventCachingFalseTransaction = false;
```
andrewyu-salesforce, line 409: Only cache the new list if there caching flag is set to true.

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -279,6 +281,28 @@ public class TDTM_TriggerHandler {
...
+        processDMLAllOrNothing(dmlWrapper, true);
+        processDMLError(dmlWrapper);
+    }
+
+    public static void processDMLError(TDTM_Runnable.DmlWrapper dmlWrapper) {
+        if (dmlWrapper.objectsWithError?.isEmpty()) {
+            return;
+        }
+
+        preventCachingFalseTransaction = true;
```
andrewyu-salesforce, line 300: The prevent caching flag will only be set in the `processDMLWithErrors` and will not interfered with original `processDML` method

force-app/tdtm/classes/TDTM_TriggerHandler.cls
bdvorachek: I think we can improve the name for this variable a bit, couple of ideas:

* skipCachingNewList
* isTransactionRollingBack

force-app/tdtm/classes/TDTM_TriggerHandler.cls
lparrott: I like that you left this open for use by other objects (if we were to change their `processDML` calls) 👍

lparrott (commented):
Looks great, I just had a comment on our new test scenario.

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -332,7 +355,7 @@ public class TDTM_TriggerHandler {
     /*******************************************************************************************************
     * @description a cache of SObject maps for each object type, used by FixupOldList()
     */
-    private static Map<Schema.DescribeSobjectResult, Map<Id, SObject>> mapDescObjToMapOldForUpdates = new Map<Schema.DescribeSobjectResult, Map<Id, SObject>>();
+    public static Map<Schema.DescribeSobjectResult, Map<Id, SObject>> mapDescObjToMapOldForUpdates = new Map<Schema.DescribeSobjectResult, Map<Id, SObject>>();
```
bdvorachek, line 358: I think this can remain private.

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -40,6 +40,9 @@ public class TDTM_TriggerHandler {
     
     private static Boolean defaultRecordsInserted = false;
     
+    private static Boolean useDefaultCachingStrategry = true;
```
bdvorachek, line 43: Typo here, `useDefaultCachingStrategy`

reedestockton, line 43: useDefaultCachingStrategy

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -377,11 +399,27 @@ public class TDTM_TriggerHandler {
 
         // now cache objects for future Update calls
         if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-            mapDescObjToMapOldForUpdates.put(describeObj, new Map<Id, SObject>(newList));
+            if (useDefaultCachingStrategry
```
bdvorachek, line 402: Can we extract this conditional into "shouldCacheNewList"?

reedestockton, line 402: I'm a big fan of this solution.  Really nice work!!!

force-app/tdtm/classes/TDTM_Runnable.cls
```diff
@@ -195,6 +196,81 @@ global abstract class TDTM_Runnable {
         }
     }
     
+    /**
```
reedestockton, line 199: I like the way you've handled this.  It ought to make it easier for us to generalize this solution if we need to.

force-app/main/default/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -108,6 +108,7 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
         List<SObject> rds, List<SObject> oldRds,
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult
     ) {
+        TDTM_TriggerHandler.enableOldListCachingEvaluation(triggerAction, rds?.size());
```
bdvorachek, line 111: I feel like TDTM_TriggerHandler should already know the current triggerAction / size. Is there a way for us to not pass these two variables back up to it?

reedestockton (approved):
Minor spelling issue, but  I'm approving.  

I really like what you've done here.  Excellent work.

force-app/tdtm/classes/TDTM_TriggerHandler.cls
```diff
@@ -377,11 +399,27 @@ public class TDTM_TriggerHandler {
...
+
             //UTIL_Debug.debug(LoggingLevel.WARN, '**** cached newList: ' + newList);
         }
 
         return oldList;
     }
 
+    public static void enableOldListCachingEvaluation(TDTM_Runnable.Action triggerAction, Integer recordCount) {
+        if (triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
+            useDefaultCachingStrategry = false;
```
bdvorachek, line 417: I think this variable might need to be a Map<SObjectType, Boolean>, we don't want to disable the default caching strategy for all SObject Types, only RDs.

bdvorachek (changes requested):
Great work on including an exact repro for this bug as a test in this pr!

force-app/main/default/classes/RD2_ChangeLogService_TEST.cls
```diff
@@ -206,6 +206,44 @@ public with sharing class RD2_ChangeLogService_TEST {
...
+
+        rdList[0].npe03__Amount__c = RD_NEW_AMOUNT;
+        rdList[1].npe03__Amount__c = 0;
+
+        Test.startTest();
+        Database.update(rdList, false);
+        Test.stopTest();
+
+        // Check first change
+        List<RecurringDonationChangeLog__c> changeLogs = getRecurringDonationChangeLog(goodRd.Id);
```
lparrott, line 236: Can we also check that the badRd does **not** create a Change Log?

npsp-reedestockton (approved):
Well, I guess I might as well approve with both of my github Ids...

force-app/main/default/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -127,12 +128,14 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
...
         // evaluation service is called so that the full schedule is available to the
         // service to determine what Opportunities to insert or modify
-        TDTM_TriggerHandler.processDML(dmlActions);
+        TDTM_TriggerHandler.processDMLWithErrors(dmlActions);
         dmlActions = null;
 
         findMatchingElevateOpps(rds, triggerAction);
 
-        evaluateOpportunities(rds, oldRds, triggerAction);
+        if (TDTM_TriggerHandler.noErrorInCurrentTransaction(objResult, rds)) {
```
lparrott, line 136: Is this just here to prevent calling `evaluateOpportunities` twice? We still want to run this for the successful records.

andrewyu-salesforce:
@jamesbsfdc Hey James, this is the fix for the RD2 Schedules not being updated correctly. We found out that the bug is originating from TDTM and it will occur to any trigger that is comparing new and old records. When performing any DML Update with All or None set to false and one of the record fail to update, the succeed record will get inconsistent result. In RD2 case both Change Log and Schedules will not get created. For Undeliverable Address, the checkbox will not sync up correctly. There are probably more places where the bug exist.

Currently we gate the fix behind a static variable so only RD2 gets the fix. The reason behind it is because the fix will affect everything TDTM. We can remove the gated variable once we are confident the change is good.

force-app/main/default/classes/ErrorRecord.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
bdvorachek, line 3:
```suggestion
    <apiVersion>53.0</apiVersion>
```

dfuller114 (approved):
great work Andrew!

jamesbsfdc:
Thanks so much for all your work on this team!!

bdvorachek (approved):

lparrott:
@davidmreed We're looking to merge this for the upcoming patch release, a release WI has also been created!

davidmreed:
@lparrott Go ahead and merge at your leisure.

PR #6854, opened 2022-02-14
RD Status Automation Label Updates

sam-knox:
[W-10699321](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000oTeVBYA0/view)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -2786,7 +2786,7 @@
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Status Automation introduction</shortDescription>
-        <value>These settings automate changing the Recurring Donation Status when donors miss installment payments.&lt;br/&gt;&lt;br/&gt;For example, given a monthly Recurring Donation started on January 1st, and Number of Days for Lapsed is set to 10 days, the Recurring Donation becomes Lapsed on February 10th if no installment payment is received. If Number of Days for Closed is set to 20 days, the Recurring Donation becomes Closed on February 20th.&lt;br/&gt;&lt;br/&gt;Elevate Recurring Donations are not included in this automation.</value>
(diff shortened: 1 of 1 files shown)


force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -2786,7 +2786,7 @@
...
+        <value>These settings automate changing the Recurring Donation Status when donors miss installment payments.&lt;br/&gt;&lt;br/&gt;For example, given a monthly Recurring Donation started on January 1st, and Number of Days for Lapsed is set to 10 days, the Recurring Donation becomes Lapsed on February 10th if no installment payment is received. If Number of Days for Closed is set to 20 days, the Recurring Donation becomes Closed on February 20th.&lt;br/&gt;&lt;br/&gt;Elevate Recurring Donations are not included in this automation.</value>
```
andrewyu-salesforce, line 2789: The RD will become Lapsed after 10 days have passed, which is on February 12th. (2/1 + 10 days + 1 days). The same applies to the Closed RD scenario.

sam-knox: Ok, right. Thanks for clarifying. So we don't start counting the 10 days until after the 1st of the month has passed (Feb 2nd in this case). Which is why the status change is on the 12th. I'll re-write this and post a new version for review.

sam-knox: @andrewyu-salesforce - I may reword this a bit, but is the following technically correct?

For example, given a monthly Recurring Donation started on January 1st, and Number of Days for Lapsed is set to 10 days, the Recurring Donation becomes Lapsed on February 12th if no installment payment is received. If Number of Days for Closed is set to 20 days, the Recurring Donation becomes Closed on February 22th.

To explain further, the lapsed Status update occurs on the 12th because Salesforce begins counting Number of Days for Lapsed on February 2nd (the day after the expected installment date). Then, after 10 days, the Status update occurs on February 12th.

andrewyu-salesforce: That's correct, 2/12 and 2/22 are the date to mark RD as Lapsed/Closed. I did the math wrong in the first reply. We give one extra day because we want to give the whole threshold days before changing the RD status.

sam-knox: Ok, thanks for confirming my understanding. I'll make another label update and ping you to review and merge.

The special case you mentioned makes sense to me. It seems in line with how other RDs would be treated.

lparrott (approved):

sam-knox:
@andrewyu-salesforce - You're listed as a reviewer, but I see Luke already did a review. Not sure you still need to. Also, do I need to follow the Rel Eng release instructions for this PR? Even if it's just labels?

andrewyu-salesforce:
I believe we do not need to follow the instruction for the label change. The plan is to merge this PR into my current working [PR](https://github.com/SalesforceFoundation/NPSP/pull/6843) and both can be merged into 238 together.

andrewyu-salesforce (approved):

sam-knox:
Thank you!

PR #6863, opened 2022-02-23
[Gift Entry] Display messages for RD in GE

psadie-sfdo:
[W-10343091](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000ZTGGXYA5/view)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/GiftBatchForQueueable_TEST.cls
```diff
@@ -0,0 +1,439 @@
...
+        // Act
+        dummyGiftBatchForQueueable.captureElevateBatches();
+
+        // Assert
+        System.assertEquals(true, elevateBatchServiceMock.enqueuePaymentCaptureJobForCalled);
+        System.assertEquals(true, giftsMock.authorizedToPaymentProcessingStatusWasCalled);
+    }
+
+    @isTest
+    static void shouldNotAttemptToCaptureElevatePaymentsWhenServiceAvailable() {
```
kenlewis-sfdo, line 69: Are the names of this method and the first one in this test class switched? Looks like this one is when service is unavailable and the first one is when the service is available.

force-app/main/default/classes/GiftBatchForQueueable_TEST.cls
```diff
@@ -0,0 +1,439 @@
...
+                .withProcessingGifts(processingGifts)
+                .withFailedGifts(failedGifts)
+                .withDryRunValidatedGifts(dryRunValidatedGifts);
+        dummyGiftBatchForQueueable.giftsSelector = stubFor(selectorMock);
+
+        // Act
+        dummyGiftBatchForQueueable.chunkGiftsThatCanBeProcessed();
+
+        // Assert
+        System.assertEquals(14, dummyGiftBatchForQueueable.chunkedIds.size(),
```
kenlewis-sfdo, line 128: Is 14 the correct number here because the batch is defaulted to a processing size of 50? If so cannot be explicitly called out here or somewhere in this test? Maybe with just a comment?

force-app/main/default/classes/GiftBatchForQueueable.cls
```diff
@@ -0,0 +1,148 @@
...
+
+    public GiftBatchForQueueable(GiftBatchId giftBatchId) {
+        this.giftBatchId = giftBatchId;
+    }
+
+    public GiftBatchId id() {
+        return giftBatchId;
+    }
+
+    public Boolean hasChunksToProcess() {
```
kenlewis-sfdo, line 93: There are a lot of public methods on this class.  Would it make sense to move any of them into the class where they are used and make them private?

kenlewis-sfdo (approved):
Great work, @psadie-sfdo - thank you!

PR #6869, opened 2022-03-01
[ProductHealth -- Release Advantage] Move the TSO (Trialforce Source Org) to GitHub

lparrott:
[W-9561751](https://gus.lightning.force.com/a07AH000000QRKfYAO)
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


bdvorachek (approved):
When we change this out to use the beta package, make sure we get the template id out of orgs/trial.json.

cumulusci.yml
```diff
@@ -876,6 +876,26 @@ flows:
...
+            2:
+                flow: install_2gp_commit #task: install_managed
+            3:
+                task: deploy_post
+                options:
+                    unmanaged: False
+            4:
+                flow: config_trial
+            5:
+                flow: enable_rd2_managed
```
dfuller114, line 893: @lparrott I think we can just use `enable_rd2` here. I ran it against a managed org and it went through fine. It doesn't appear that the `managed` property is needed to warrant a separate flow for this. If we could just use `enable_rd2` that would help avoid confusion around why that flow has a `managed` suffix and the others don't.

dfuller114 (changes requested):
@lparrott I think we can just use enable_rd2 here. I ran it against a managed org and it went through fine. It doesn't appear that the managed property is needed to warrant a separate flow for this. If we could just use enable_rd2 that would help avoid confusion around why that flow has a managed suffix and the others don't.

dfuller114 (approved):

dfuller114:
@lparrott not sure why, but you disappeared from the review and i don't seem to be able to add you again.

unpackaged/config/trial/standardValueSetTranslations/ContactRole-de.standardValueSetTranslation
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<StandardValueSetTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
```
bdvorachek, line 2: Maybe these standardValueSetTranslations should live in the trial_translations folder? There's some duplicated metadata between the two.

cumulusci.yml
```diff
@@ -1027,25 +1059,6 @@ flows:
                 options:
                     class_name: RD2_DataMigration_BATCH
 
-    enable_rd2_managed:
```
andrewyu-salesforce, line 1030: The description (line 1008)for `enable_rd2` flow should be modified to indicate it works for both managed and un-managed code.

andrewyu-salesforce (commented):
@dfuller114 minor feedback on the description of the `enable_rd2` flow

andrewyu-salesforce (approved):

lparrott:
Approved! (Since this started as my PR, I can't actually approve it 😉)

PR #6883, opened 2022-03-11
RD2: Status Automation Update Elevate Commitment Status

andrewyu-salesforce:
[WI-1034876](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000ZS5f2YAD/view)
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/PS_Request.cls
```diff
@@ -74,6 +75,7 @@ public inherited sharing class PS_Request {
     public static String ENDPOINT_PURCHASE = '/v1/payments/verified/purchase';
     public static String ENDPOINT_COMMITMENT = '/v1/payments/verified/commitments';
     public static String ENDPOINT_COMMITMENT_CANCEL = '/v1/payments/verified/commitments/{0}/cancel';
+    public static String ENDPOINT_COMMITMENT_UPDATE_BULK = '/v1/payments/internal/commitments/update/bulk';
```
andrewyu-salesforce, line 78: Using Internal endpoint for now. Will switch to Verified endpoint once it's completed.

lparrott (commented):
Looks good! Just had a few questions/comments.

force-app/main/default/classes/RD2_OpportunityEvaluation_BATCH.cls
```diff
@@ -225,6 +229,24 @@ public without sharing class RD2_OpportunityEvaluation_BATCH
...
+        Decimal configValue = UTIL_CustomSettingsFacade.getRecurringDonationsSettings().Recurring_Donation_Batch_Size__c;
+
+        if (configValue == null) {
+            configValue = 50;
+        }
+
+        if (RD2_ElevateIntegrationService.isIntegrationEnabled() &&
+            (rd2Settings.isStatusAutomationForClosedEnabled() || rd2Settings.isStatusAutomationForLapsedEnabled())
+        ) {
+            Decimal maxSize = RD2_UpdateCommitmentBulkService.MAXIMUM_API_CALL_PER_TRANSACTION
```
andrewyu-salesforce, line 242: @lparrott I made some small adjustments to increase code readability. BatchSize is only called from `Database.executeBatch({{your-batch}}, batchSize)` to determine the record count of individual transactions in a batch. In this case, I am limiting the max record count to 250 records (5 * 50) when the Bulk API can be performed. The reasons for 250 records are the following:
-  The maximum record count per API callout is 50 records
-  The recommended timeout per API callout is set to 20 seconds and the government limit on API timeout per transaction is 120 seconds. 5 API callout will lend at 100 seconds on API timeout. I give a 20 seconds buffer room just in case.

Therefore, the bulk commitment API MAXIMUM_API_CALL_PER_TRANSACTION * REQUEST_SIZE get us 250 records.

force-app/main/default/classes/RD2_UpdateCommitmentBulkService.cls
```diff
@@ -0,0 +1,156 @@
...
+/**
+* @author Salesforce.org
+* @date 2022
+* @group Recurring Donations
+* @description Update Elevate Commitment In Bulk
+*/
+public with sharing class RD2_UpdateCommitmentBulkService {
+    public static final Integer MAXIMUM_API_CALL_PER_TRANSACTION = 5;
+    public static final Integer REQUEST_SIZE = 50;
+    private Integer apiCallOutCounter = MAXIMUM_API_CALL_PER_TRANSACTION;
```
andrewyu-salesforce, line 39: apiCalloutCounter tracks how many leftover API callouts we can perform in a single transaction. It starts with `MAXIMUM_API_CALL_PER_TRANSACTION`, which is 5 in this case, and `processCommitments()` function will reduce the callout counter each time a callout is made.

force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -319,18 +319,28 @@ public inherited sharing class RD2_RecurringDonation {
     public RD2_RecurringDonation statusAutomationReviseStatus(RD2_ScheduleService scheduleService) {
```
dfuller114, line 319: I think as a general principle, if we can push out deterministic if/else logic from behavioral methods like this, particularly in entity classes, we should. In this case, the schedule service parameter is only used to determine if the RD should be marked as lapsed or closed. So we could have this conditional logic in the Opportunity Evaluation Service where we call this method and keep the service dependencies out of this class. There are a few things I think that gets us:
1. Clarity that when a client calls this method, it will do what it says it will do
2. If the conditional logic around when this method would be called needs to change with different clients then the client can determine when to call it.
3. Improved readability 

Also, I think we could even split up the two code blocks into two separate methods like `setStatusToClosed()` and `setStatusToLapsed()`

force-app/main/default/classes/RD2_UpdateCommitmentBulkService.cls
```diff
@@ -0,0 +1,159 @@
...
+        return response.statusCode == UTIL_Http.STATUS_CODE_CREATED
+            || response.statusCode == UTIL_Http.STATUS_CODE_OK;
+    }
+
+    public class RequestBody {
+        @TestVisible
+        List<Commitment> commitments;
+    }
+
+    public class Commitment {
```
dfuller114, line 145: Could we extract this out into its own class? I can see this being used again that we could continue building additional behavior around as its own entity

dfuller114 (changes requested):
looks great, Andrew. just had a couple of comments.

lparrott (approved):

PR #6889, opened 2022-03-17
[Mark HH] Batch job to synchronize All Household Members Deceased flag

npsp-reedestockton:
Added a batch job to synchronize the Account All Household Members Deceased status with the statuses of all Household members.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom labels
- stgBtnRunHHDeceasedBatch
- stgHelpDeceasedBatch
- stgHelpHHData
- stgHHDataRefreshTitle
- stgLabelDeceasedBatch

# Deleted Metadata


npsp-reedestockton:
Tracking [W-10737644](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000rD45tYAC/view)

npsp-reedestockton:
@lparrott @bdvorachek Would appreciate your review.

@sam-knox These changes should reflect your suggestions, including renaming "Refresh Household Names" to "Refresh Household Data"

force-app/main/default/classes/DeceasedBatch_TEST.cls
```diff
@@ -0,0 +1,238 @@
...
+            testContact.Id = UTIL_UnitTestData_TEST.mockId(Contact.SObjectType);
+            testContact.Account = testAccount;
+            testContact.AccountId = testAccount.Id;
+            testContact.Deceased__c = false;
+        }
+
+        return contacts;
+    }
+
+    private static void setContactsDeceased(List<Contact> contacts, Boolean accountMatch, Boolean mixed) {
```
lparrott, line 224: This could use a description comment, walking through it I get what you're doing, but it's a little confusing!
Alternatively, you could setup a few abstracted helpers that set the Booleans, `setAllContactsDeceased`,  `setSomeContactsDeceased` and `matchContactsDeceasedToHouseholds` or something like that.

force-app/main/default/pages/STG_PanelDeceasedBatch.page
```diff
@@ -0,0 +1,49 @@
+<apex:page controller="STG_PanelDeceasedBatch_CTRL" docType="html-5.0" standardStylesheets="false" >
```
lparrott, line 1: Not more Visualforce! 😉  This is a simple enough page that I think it's appropriate for our timeline, but my goodness we need to update these Settings...

lparrott (commented):
Looking good, just had a few comments. On to QA!

force-app/main/default/classes/DeceasedBatch.cls
```diff
@@ -0,0 +1,146 @@
...
+*/
+public class DeceasedBatch implements Database.Batchable<sObject>, Database.Stateful {
+
+    @TestVisible
+    Set<Id> highLowProcessedIds = new Set<Id>();
+
+    public Database.QueryLocator start(Database.BatchableContext info) {
+        final String HH_ACCOUNT_TYPE = CAO_Constants.HH_ACCOUNT_TYPE;
+        return Database.getQueryLocator(
+            'SELECT AccountId, Account.All_Members_Deceased__c ' +
```
lparrott, line 44: I just realized this misses cases where the Household is marked Deceased, but there are no Deceased Contacts, could we make it `(Deceased__c = TRUE OR Account.All_Members_Deceased__c = TRUE)` in the WHERE clause?

npsp-reedestockton: I think there are some other areas that will need adjustment to deal with this as well.  I need to think about the best approach, but it is definitely an issue and there should also be a unit test to deal with it.

force-app/main/default/classes/DeceasedBatch.cls
```diff
@@ -0,0 +1,146 @@
...
+        Map<Id, Boolean> allMembersDeceasedById = getAllMembersDeceasedById(scope);
+        List<Contact> contacts = getContacts(allMembersDeceasedById);
+        if (allMembersDeceasedById.isEmpty() || contacts.isEmpty()) {
+            return;
+        }
+
+        List<Account> allMembersDeceasedUpdates = getAllMembersDeceasedUpdates(contacts, allMembersDeceasedById);
+        updateModifiedAccounts(allMembersDeceasedUpdates);
+
+        saveHighAndLowProcessedIds(contacts);
```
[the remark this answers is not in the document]
npsp-reedestockton: correct.   I don't really think it would be much of an issue, but to the extent possible I think it is a scenario we should avoid.

force-app/main/default/pages/STG_PanelDeceasedBatch.page
npsp-reedestockton: We briefly discussed lwc, but I'm not sure that makes sense with every one of the other pages handled this way...

force-app/main/default/classes/DeceasedBatch.cls
npsp-reedestockton: This needs to be handled in the finish method.  The queryLocator and the query in the execute method are both driven by Contacts.  This has a completely different set of needs.

npsp-reedestockton:
> Reede - I think what we wanted was to combine both Recalculate Deceased Household Accounts and Refresh All Household Names into a single settings page called Refresh Household Data.
> 
> @bdvorachek

That makes sense.  Unfortunately, that requires a different approach and more discussion.  Everything other than this comment is directed toward a different approach.

npsp-reedestockton:
> 

Actually...  that might be easier than I indicated.  I'll take a look at that.

sam-knox:
@npsp-reedestockton - I'm ok if we go with two separate pages. Based on timeline, that might be best. If you do decide to keep it that way please change `stgHHNameRefreshTitle` back to Refresh Household Names.

force-app/main/default/classes/DeceasedBatch.cls
```diff
@@ -0,0 +1,167 @@
...
+        Database.update(allMembersDeceasedUpdates, false);
+    }
+
+    @TestVisible
+    private void saveHighAndLowProcessedIds(List<Contact> contacts) {
+        highLowProcessedIds.add(contacts[0].AccountId);
+        highLowProcessedIds.add(contacts[contacts.size() - 1].AccountId);
+    }
+
+    public void finish(Database.BatchableContext info) {
```
npsp-reedestockton, line 146: This may not be something we care about enough to handle.  Is this worth doing???

force-app/main/default/pages/STG_PanelRenameHouseholds.page
```diff
@@ -17,26 +17,67 @@
         }
     </script>
     <apex:form id="form" styleClass="slds-m-around_x-large">
-        <c:STG_PageHeader sectionLabel="{!$Label.stgNavBulkProcesses}" pageLabel="{!$Label.stgHHNameRefreshTitle}" />
+        <c:STG_PageHeader sectionLabel="{!$Label.stgNavBulkProcesses}" pageLabel="{!$Label.stgHHDataRefreshTitle}" />
```
andrewyu-salesforce, line 20: This label does not exist in the Custom Label metadata

sam-knox: @npsp-reedestockton - Is this a label you need me to write?

npsp-reedestockton: @sam-knox @andrewyu-salesforce That's text you already provided @sam-knox.  "Refresh Household Data".  In flipping things around with the pages I neglected to add it - it was previously a modification of an existing label.

force-app/main/default/classes/DeceasedBatch.cls
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        List<Account> allMembersDeceasedUpdates = getAllMembersDeceasedUpdates(contacts, allMembersDeceasedById);
+        updateModifiedAccounts(allMembersDeceasedUpdates);
+
+        saveHighAndLowProcessedIds(contacts);
+    }
+
+    @TestVisible
+    private Map<Id, Boolean> getAllMembersDeceasedById(List<Contact> scope) {
```
dfuller114, line 67: test

sam-knox (changes requested):
I added a suggestion for a second sentence to the intro text at the top explaining that you need to refresh to run the second batch on the page. Otherwise good to go here!

force-app/main/default/classes/DeceasedBatch.cls
```diff
@@ -0,0 +1,148 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2022
+* @group Settings
+* @description Batch Class to recalculate the All Household Members Deceased flag on Household Accounts.
+*/
+public class DeceasedBatch implements Database.Batchable<sObject> {
```
npsp-reedestockton, line 36: Removed stateful interface and stateful variables.  After discussing with @dfuller114, we decided that controlling for overlap was unnecessary.  This would be useful if there was a possibility for row locking as a result of async processes, but since we are turning off triggers for our update, there is no such risk here.  Simple is better in this case.

force-app/main/default/classes/DeceasedBatch.cls
lparrott: This working well now, marking as resolved!

force-app/main/default/classes/DeceasedBatch.cls
lparrott: Ehhh, the Deceased flag gets unchecked when the last Contact gets removed (must be in the Trigger), so this is definitely an edge case. That being said, it's already here, other than one extra query it can't really hurt.

sam-knox (approved):

lparrott (approved):

PR #6899, opened 2022-03-23
Remove API connection to sfdo-status

davidmreed:
# Critical Changes

# Changes

- We removed the API connection in the legacy NPSP Getting Started page. Users who have this tab exposed will now see static links to our YouTube playlist and GitHub releases.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/RP_GitHubController.cls
```diff
@@ -34,100 +34,10 @@
 * @description Component that brings the latest release notes from Github using Salesforce org HEROKU APP.
 */
 public with sharing class RP_GitHubController {
-
-    //Object used to load all information related to github release notes
-    public GitHubWrapper gitHubData {get; set;}
-    public Boolean timeoutError {get; set;}
-    public Boolean remoteSiteSettingIsNotActive {get; set;}
+    // This class is slated for future deprecation and removal.
```
force2b, line 37: I noticed that you stripped down some classes as marked as "to be deleted", such as this one. But in other cases the classes are just deleted (`RP_GitHubClient`, `RP_GitHubTestMock`, etc.). Is there a reason some are deleted and others are not?

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -3216,27 +3216,6 @@
         <shortDescription>Resources Page - Product Name</shortDescription>
         <value>Nonprofit Success Pack (NPSP)</value>
     </labels>
-    <labels>
```
force2b, line 3219: I'm unsure about removing these. They'll still be left in the customers org, but just not as part of the package. An alternative is to change the `shortDescription` to prefix with "DEPRECATED" and add a `category` of "DEPRECATED" for all of the ones to be removed. Then in a subsequent release, we actually remove the labels and translations. This way if customers see the labels in their org they at least can easily know they're no longer used.

Do we need to manually delete these labels from the packaging org after they're removed from the package? Or maybe just manually remove from the release (or would the spider pick that up)? I do believe we'll need to manually delete any removed labels from the TSO, at least until we switch to a metadata driven TSO.

force2b (commented):
Overall this looks good, and it works in a scratch org. I had a couple of questions though. 

Also, for anything we delete in this PR they may need to be manually deleted from the packaging org after they're removed from the package (though that may not be possible). We'll definitely need to delete them from the TSO manually, at least until we switch to a metadata driven TSO.

force-app/main/default/labels/CustomLabels.labels-meta.xml
davidmreed: My understanding is that removing them from the source tree is essentially a no-op, except that they won't ever be part of the translation process again. CumulusCI cannot delete them from the packaging org because Custom Labels are not individually addressable. The package shouldn't actually change.

If you prefer to mark them deprecated I'm happy to do that.

force2b: Yeah. I see your point. I know there's a 2 release process involved in this, but I do think the best way to do it is to first mark everything as deprecated, release, and then remove them from the package. The former needed so customers know they're no longer used. The latter so that we don't have to pay for translations on them. Plus, adding a category of "Deprecated" will make them easier to delete from the Trial org - at least until we have a metadata based TSO.

davidmreed: It took me a silly number of commits to get the diff right on this - will squash! I changed the labels to Deprecated.

lparrott (approved):
I just deployed and tested, looks good!

PR #6914, opened 2022-04-05
W-10946402: Field type conversion failing in packaging org

grivero:
W-10946402: roll back custom field definition for `Opportunity.Notification_Recipient_Infomation__c` and create a fix for LWC component 

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


callen-sfdo (approved):

force2b (approved):

force-app/main/default/objects/Opportunity/fields/Elevate_Notification_Recipient_Info__c.field-meta.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>Elevate_Notification_Recipient_Info__c</fullName>
```
force2b, line 3: I'm a bit confused as to why this field starts with "Elevate_" and the Honoree Information field does not.  They both appear to be brand new for 238, and they both appear to be for the same related information. Shouldn't they start with the same field prefix and label?

grivero: Thanks for bringing that up @force2b, we could definitely do that, I'm trying to keep changes to minimum while we wait for further indications from docs team (@melissabarber)

grivero: The other suggestion I had for this field was: `Notification_Recipient_Contact_Info__c` 
Do you think/feel this is a better label to use? @force2b

mldyang (approved):

force-app/main/default/objects/Opportunity/fields/Elevate_Notification_Recipient_Info__c.field-meta.xml
force2b: Are these fields unique to Elevate? What have we done with other Elevate specific fields?

grivero: Yes, they are part of the Tribute's epic.
We haven't used Elevate_ as prefix for elevate's specific fields in the past

force2b: Interestingly, all the Elevate fields on the npe01_Payment object start with "Elevate_", but none of them on this object. Maybe these aren't exactly used the same way though, so probably no work to do on them. 

However, on this objects all the fields need to be consistent in naming. Since none of the other new fields added for this work have "Elevate" in the api name or label, this one should not either.

callen-sfdo: @force2b It all depends on whether it is elevate specific or not, most of the ones we created on the payment were elevate specific so we named them accordingly to avoid confusion.  This field would be elevate specific since the existing tribute notification recipient information field was not big enough.  The rest of the tribute fields are not elevate specific though so they would remain the same.

force2b: Thanks for that clarification @callen-sfdo. Excluding the two Tribute fields, it looks like there were 4 fields added that may all be related. If these are all Elevate specific, than I'd like to see the naming (api and label) for each also be consistent. If adding "Elevate_" to each is seriously complex, then just pull it off the new field; unless @lparrott feels differently.

- Elevate_Notification_Recipient_Info__c
- Honoree_Information__c
- Notification_Recipient_Email__c
- Notification_Recipient_Information__c

force2b (approved):

lparrott (approved):

PR #6921, opened 2022-04-07
W-10219787 Multi-Currency support 

maurimart:
# Critical Changes

Add currency conversion according to user
Add ISO codes instead of currency symbols

# Changes

Modification in the DonationHistoryController response to add the currency code of the current user


maurimart:
[W-10839969](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000sopHaYAI/view)

JoseRossiSF (approved):
LGTM

gerardorodriguezsf (approved):
LGTM

bdvorachek (changes requested):
Two unit tests to fix here, and a question about how we're displaying converted currency in the component.

force-app/main/default/classes/DonationHistoryController.cls
```diff
@@ -100,6 +101,7 @@ public with sharing class DonationHistoryController {
             donationHistories.donations = buildHistoryFromOpportunities(historyService.getDonationHistory(contactId, offset, filter), contactId, paymentsEnabled);
             donationHistories.totalNumberOfRecords = historyService.getTotalNumberOfRecords(contactId, filter);
             donationHistories.isPaymentsEnabled = paymentsEnabled;
+            donationHistories.currencyISOCode = UserInfo.getDefaultCurrency();
```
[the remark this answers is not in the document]
maurimart: Done!

JoseRossiSF (approved):

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -36,28 +36,27 @@ public inherited sharing class DonationHistorySelector {
...
-            ORDER BY CloseDate 
-            DESC LIMIT 50
-            OFFSET :offset
-        ];
+  public List<Opportunity> getWonOpportunitiesByContactId( Id contactId, Integer offset, String amount ) {       
+        return Database.query(
+            'SELECT Id, '+
+                    'Name, '+
+                    'CloseDate,' +
+                    + String.escapeSingleQuotes(amount) +
```
lparrott, line 44: Instead of passing `amount` as a parameter, could it be determined by a separate method that is called directly from here?

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -66,29 +65,29 @@ public inherited sharing class DonationHistorySelector {
...
-            ORDER BY Opportunity.CloseDate 
-            DESC LIMIT 50
-            OFFSET :offset
-        ];
+    public List<Opportunity> getWonOpportunitiesByContactIdAndYear(Id contactId, Integer offset, Integer year, String amount) {
+        return Database.query(
+            'SELECT  Id, '+
+                    'Name,' +
+                    'CloseDate, '+
+                    + String.escapeSingleQuotes(amount) +
```
lparrott, line 73: Same as above.

force-app/main/service/DonationHistoryService.cls
```diff
@@ -90,10 +90,11 @@ public with sharing class DonationHistoryService {
     public List<Opportunity> getDonationHistory(Id contactId, Integer offset, String filter) {
         if (hasAccessToDonationHistory()) {
             List<Opportunity> credits = null;
+            String amount = UTIL_Currency.getInstance().isMultiCurrencyOrganization() ? 'convertCurrency(Amount)' : 'Amount' ;
```
lparrott, line 93: This is similar to the method that could get called from within the query.

force-app/main/selector/DonationHistorySelector.cls
maurimart: Done

force-app/main/selector/DonationHistorySelector.cls
maurimart: Done

force-app/main/service/DonationHistoryService.cls
maurimart: Done

force-app/main/selector/DonationHistorySelector.cls
```diff
@@ -131,6 +131,13 @@ public inherited sharing class DonationHistorySelector {
         return totalRecords;
     }
 
+    /** 
+    * @description method to return Amount or converCurrency(Amount) field if org is multicurrency. 
+    */
+    public String isMultiCurrencyOrganization(){
```
lparrott, line 137: This method name sounds like it would return a Boolean, can we use `getOppAmountFieldForQuery` or something like that? Also, could it include the `String.escapeSingleQuotes` portion so we don't need to include that in the query itself? Thanks!

lparrott (approved):

bdvorachek (approved):

gerardorodriguezsf (approved):
LGTM

JoseRossiSF (approved):

PR #6924, opened 2022-04-12
Fixing regression where RD2 Change Type is not set without value change

lparrott:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/lwc/rd2Service/model.js
+++ b/force-app/main/default/lwc/rd2Service/model.js
@@ -7,6 +7,7 @@ import {
     SET_CONTACT_DETAILS,
     SET_CURRENCY,
     SET_ACCOUNT_DETAILS,
+    SET_CHANGE_TYPE,
     SET_DAY_OF_MONTH,
     SET_DONATION_AMOUNT,
     SET_DONOR_TYPE,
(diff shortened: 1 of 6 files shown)


force-app/main/default/lwc/rd2Service/model.js
```diff
@@ -134,7 +135,7 @@ const getDonationValue = (state) => {
 
 const getChangeType = (state) => {
     if (isRecurringTypeChanged(state) || isNewRecord(state)) {
-        return "";
+        return state.changeType;
```
lparrott, line 138: Doing this allows a manual change to the picklist to "stick" even if the overall Value changed, which is in line with our current release.

lparrott: Yup, I thought about this last night after you reached out, I'll make the change and do a bit more testing!

lparrott: Just pushed a fix, thanks for adding that Jest test 👍

bdvorachek (approved):

andrewyu-salesforce (approved):

lparrott:
Added to #6928

PR #6939, opened 2022-04-20
W-10004087 - W-10004095 /  GitHub actions

maurimart:
# Critical Changes

added cspell and eslint packages and configured github actions to run on every push

### DIFF
--- a/.github/workflows/cspell.yml
+++ b/.github/workflows/cspell.yml
@@ -1,0 +1,24 @@
+name: CSpell
+on: push
+jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
+      - name: Determine Node Version
+        id: node-version
+        run: |
+          echo "::set-output name=ver::$(node --version)"
+      - uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-${{ steps.node-version.outputs.ver }}-${{ hashFiles('package.json', 'yarn.lock') }}
+      - name: Install Packages
+        run: |
+          yarn install
+      - name: Run CSpell
+        run: |
+          npm run cspell:lwc
--- a/.github/workflows/cspell.yml
+++ b/.github/workflows/cspell.yml
@@ -1,7 +1,7 @@
 name: CSpell
 on: push
 jobs:
-  eslint:
+  cspell:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -4,16 +4,21 @@ jobs:
   eslint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
-      - run: npm ci # or yarn install
       - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
         with:
-          eslint-args: '--ignore-path=.gitignore --quiet'
-          extensions: 'js,jsx,ts,tsx'
-          annotations: true
+          node-version: '14'
+      - name: Determine Node Version
+        id: node-version
+        run: |
+          echo "::set-output name=ver::$(node --version)"
+      - uses: actions/cache@v2
+        with:
+          path: node_modules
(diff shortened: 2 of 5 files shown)


maurimart:
[W-10004087](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000EJxkKYAT/view)
[W-10004095](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000EKBbxYAH/view)

JoseRossiSF (approved):
LGTM for the moment. We might need to create a new ticket to address the checked issues in a future PR

gerardorodriguezsf (approved):

.github/workflows/cspell.yml
```diff
@@ -0,0 +1,24 @@
+name: CSpell
+on: push
+jobs:
+  cspell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
```
jstvz, line 10: Cache support is [built in](https://github.com/actions/setup-node/blob/25316bbc1f10ac9d8798711f44914b1cf3c4e954/docs/advanced-usage.md#caching-packages-dependencies) to `actions/setup-node@v2`
```suggestion
          node-version: '14'
          cache: 'yarn'
          cache-dependency-path: yarn.lock
```

.github/workflows/cspell.yml
```diff
@@ -0,0 +1,24 @@
+name: CSpell
+on: push
+jobs:
+  cspell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
```
jstvz, line 8: ±: Would you explain the pin to `@v2-beta`? It's a few years out of date, and I strongly recommend updating to v2 or v3 if possible.

.github/workflows/cspell.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          node-version: '14'
+      - name: Determine Node Version
+        id: node-version
+        run: |
+          echo "::set-output name=ver::$(node --version)"
+      - uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-${{ steps.node-version.outputs.ver }}-${{ hashFiles('package.json', 'yarn.lock') }}
```
jstvz, line 18: If we switch to `@v2`+ and use it's built-in cache you could remove these steps.

.github/workflows/eslint.yml
```diff
@@ -0,0 +1,24 @@
+name: ESLint
+on: push
+jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
```
jstvz, line 8: ±: Same question re: `@v2-beta`.

jstvz (changes requested):
Requested changes/queries are marked with ±. 

Since both `jobs` are triggered by the same event (`push`), I suggest combining them into a single workflow file:

```yaml
jobs:
  cspell:
    name: ...
    steps: ...
  eslint:
    name: ...
    steps: ...
```

They'll still run in parallel, but will be easier to maintain.

(TIL that eslint has Github check annotations built in!)

.github/workflows/cspell.yml
maurimart: Done!

.github/workflows/cspell.yml
maurimart: only for the reference of the previous actions, but it is a good suggestion, I changed it

.github/workflows/cspell.yml
maurimart: unnecessary steps removed!

.github/workflows/eslint.yml
maurimart: Done

jstvz (approved):
LGTM.

---

I don't know whether the team has additional work planned to resolve existing errors, but FYI: As configured, ESLint and CSpell are running against everything in the `force-app/main/default/lwc` directory, and is currently failing. It may be difficult for reviewers to recognize when a PR contains new errors. 

It's possible to use [lint-staged](https://github.com/okonet/lint-staged) to only lint files included in a PR (we have examples available in our [template repository](https://github.com/SFDO-Tooling/GitHubActions)). This approach, when combined with a pre-commit tool like [husky](https://github.com/typicode/husky) or [pre-commit](https://pre-commit.com/), allows the team to incrementally resolve lint errors without the noise of a permanently failing build.

Let us (RelEng) know if you have any questions or want to discuss what tools are available.

lparrott:
Can we delete this PR and branch since it was intended for the cx repository?

PR #6941, opened 2022-04-21
Removing Chatter references from Flexipages

lparrott:
- Removing Chatter references from Flexipages 
- Adding back managed rd2 flow to fix dependent package flows

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -1016,6 +1016,11 @@ flows:
                     field: "%%%NAMESPACE%%%Field_Mapping_Method__c"
                     value: "Data Import Field Mapping"
 
+    enable_rd2_managed:
+        steps:
+            1: 
+                flow: enable_rd2
+
     enable_rd2:
         description: 'Fully configures and enables a Scratch org with Enhanced Recurring Donations (unmanaged code only)'
         group: NPSP
--- a/unpackaged/config/trial/flexipages/NPSP_Account_Record_Page.flexipage
+++ b/unpackaged/config/trial/flexipages/NPSP_Account_Record_Page.flexipage
@@ -270,16 +270,6 @@
         <name>facet-bba4c430-44a0-423e-8e7d-a0a6e28ee13b</name>
         <type>Facet</type>
     </flexiPageRegions>
-    <flexiPageRegions>
-        <itemInstances>
-            <componentInstance>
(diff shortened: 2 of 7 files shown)


cumulusci.yml
```diff
@@ -1016,6 +1016,11 @@ flows:
                     field: "%%%NAMESPACE%%%Field_Mapping_Method__c"
                     value: "Data Import Field Mapping"
 
+    enable_rd2_managed:
```
force2b, line 1019: Is this supposed to be in here?

unpackaged/config/trial/flexipages/NPSP_Account_Record_Page.flexipage
```diff
@@ -270,16 +270,6 @@
         <name>facet-bba4c430-44a0-423e-8e7d-a0a6e28ee13b</name>
         <type>Facet</type>
     </flexiPageRegions>
-    <flexiPageRegions>
```
force2b, line 273: I don't think we need to - or should - remove these from the unpackaged metadata. This will be used to configure trial orgs, where we want the component to show. It will not be pushed to existing customers, which is where the current issue is.

force2b (changes requested):

unpackaged/config/trial/flexipages/NPSP_Account_Record_Page.flexipage
lparrott: That's fair, I can revert those, I got a little delete happy with this one.

cumulusci.yml
lparrott: Yes, we inadvertently broke flows in the Transport Layer and Elevate Connector that are using this (now unnecessary) flow, to avoid making changes to both of those projects I figured I'd just add this.

force2b (approved):

dfuller114:
@lparrott @force2b I see a number of `forceChatter` references in the unpackaged folder as well. are we concerned with those too?

force2b:
Thanks for adding that file @lparrott

force2b:
@davidmreed This should be ready for merging now.

dfuller114 (approved):

davidmreed (approved):

PR #6949, opened 2022-04-29
Fix for Queueable Context Issue

voduyemi:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_QueueableService.cls
+++ b/force-app/main/default/classes/RD2_QueueableService.cls
@@ -285,12 +285,20 @@ public class RD2_QueueableService {
     public static void enqueueOppNamingService(List<Opportunity> updatedOpps) {
         OpportunityNamingService service = new OpportunityNamingService(updatedOpps);
 
-        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs() && !Test.isRunningTest()) {
+        if (canQueueNamingService()) {
             System.enqueueJob(service);
 
         } else {
             service.refreshOppNames();
         }
+    }
+
+    private static Boolean canQueueNamingService() {
+        RD2_Settings settings = RD2_Settings.getInstance();
+        Boolean shouldRunSynchronously =
+            settings.firstInstallmentCreateMode == RD2_Constants.FirstInstallmentOppCreateOptions.Synchronous.name();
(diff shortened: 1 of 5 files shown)


force-app/main/default/classes/RD2_QueueableService.cls
```diff
@@ -285,14 +285,22 @@ public class RD2_QueueableService {
...
+        if (canQueueNamingService()) {
             System.enqueueJob(service);
 
         } else {
             service.refreshOppNames();
         }
     }
 
+    private static Boolean canQueueNamingService() {
+        RD2_Settings settings = RD2_Settings.getInstance();
```
andrewyu-salesforce, line 297: Over here it should be specifically for Gift Entry Mode Only. We only want the job to be synchronous when it's in gift entry mode, not when the created mode is set to `Synchronous`. I think you can introduce another static variable called `giftEntryMode` and in RD2's should run synchronously you can do a || operation to know if the job needs to be done synchronously.

force-app/main/default/classes/RD2_RecurringDonations_TDTM.cls
```diff
@@ -400,15 +400,20 @@ public class RD2_RecurringDonations_TDTM extends TDTM_Runnable {
             return;
         }
 
-        if (triggerAction == TDTM_Runnable.Action.AfterInsert && shouldEvaluateOppsSynchronouslyOnInsert(rds.size() > 1)) {
+        if (canQueueOpportunityEvaluationService(triggerAction, rds)) {
```
andrewyu-salesforce, line 403: nice refactor

andrewyu-salesforce (commented):
@voduyemi We need to be careful on using `firstInstallmentCreateMode` in oppNameService because we do not want the naming job to run synchronously every time the value is set to `firstInstallmentCreateMode` is set to `synchronous`. We only want it to run in synchronous when it's in gift entry context.

force-app/main/default/classes/RD2_QueueableService.cls
lparrott: Great catch, could set it during `disableRD2AsynchronousProcesses` as long as it's in the same transaction.

lparrott (approved):

andrewyu-salesforce (approved):

PR #6954, opened 2022-05-11
[Gift Entry] Stale permissions fix

voduyemi:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/lwc/geTemplateBuilder/geTemplateBuilder.js
+++ b/force-app/main/default/lwc/geTemplateBuilder/geTemplateBuilder.js
@@ -1360,6 +1360,9 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
             // create custom migrated field mapping sets yet.
             this.formTemplate.layout.fieldMappingSetDevName = DEFAULT_FIELD_MAPPING_SET;
 
+            delete this.formTemplate.permissionErrors;
+            delete this.formTemplate.permissionErrorType;
+
             const preppedFormTemplate = {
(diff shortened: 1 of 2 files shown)


force-app/main/default/lwc/geTemplateBuilder/geTemplateBuilder.js
```diff
@@ -1342,51 +1342,60 @@ export default class geTemplateBuilder extends NavigationMixin(LightningElement)
...
+        } else {
+            this.formTemplate.defaultBatchTableColumns
+                = this.selectedBatchTableColumnOptions;
+        }
+
+        // TODO: Currently hardcoded as we're not providing a way to
+        // create custom migrated field mapping sets yet.
+        this.formTemplate.layout.fieldMappingSetDevName
+            = DEFAULT_FIELD_MAPPING_SET;
+       this.stripPermissionsData();
```
lparrott, line 1383: Does it make more sense to call this at the start of `checkTabsValidity`? That way it's more connected to the actual permission checking logic. It seems sort of... hidden here? Looking at the code though, we may not have the `formTemplate` at that point.

voduyemi: It's actually called in this function `preppedFormTemplate`, this function prepares the template to be saved. The `checkTabsValidity` function gets called before we try to save the template.

lparrott (approved):

dfuller114 (approved):
looks good. thanks Victor!

PR #6962, opened 2022-05-24
Regression Fix: W-1179558 -- Missing State value when State & Country Picklists are enabled

force2b:
Work Item: [W-1179558](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000xRsWUYA0/view)

# Critical Changes

# Changes

- Fix a bug that caused the BillingState and BillingCountry fields on the Account to appear as null when State & Country Picklists are enabled, and a new Contact with an Address is created.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/NPSP_Address.cls
+++ b/force-app/main/default/classes/NPSP_Address.cls
@@ -341,6 +341,13 @@ public inherited sharing class NPSP_Address implements IAddress {
         address.Geolocation__Latitude__s = (Decimal) getSObjectField(sobjSrc, fieldPrefixSrc, 'Latitude');
         address.Geolocation__Longitude__s = (Decimal) getSObjectField(sobjSrc, fieldPrefixSrc, 'Longitude');
 
+        if (orgConfig.isStateCountryPicklistsEnabled()) {
+            if (address.MailingCountry__c == null && getSObjectField(sobjSrc, fieldPrefixSrc, 'CountryCode') != null) {
+                address.MailingState__c = (String) getSObjectField(sobjSrc, fieldPrefixSrc, 'StateCode');
+                address.MailingCountry__c = (String) getSObjectField(sobjSrc, fieldPrefixSrc, 'CountryCode');
+            }
+        }
+
         // if specified, also include Address Type
         if (addrType != null) {
             address.Address_Type__c = addrType;
--- a/force-app/main/default/classes/NPSP_Address.cls
+++ b/force-app/main/default/classes/NPSP_Address.cls
@@ -342,12 +342,8 @@ public inherited sharing class NPSP_Address implements IAddress {
         address.Geolocation__Longitude__s = (Decimal) getSObjectField(sobjSrc, fieldPrefixSrc, 'Longitude');
 
         if (orgConfig.isStateCountryPicklistsEnabled()) {
-            if (address.MailingCountry__c == null && getSObjectField(sobjSrc, fieldPrefixSrc, 'CountryCode') != null) {
-                address.MailingCountry__c = (String) getSObjectField(sobjSrc, fieldPrefixSrc, 'CountryCode');
-            }
-            if (getSObjectField(sobjSrc, fieldPrefixSrc, 'StateCode') != null) {
-                address.MailingState__c = (String) getSObjectField(sobjSrc, fieldPrefixSrc, 'StateCode');
-            }
+            populateAddressCountryFromCountryCode(sobjSrc, fieldPrefixSrc);
+            populateAddressStateFromStateCode(sobjSrc, fieldPrefixSrc);
         }
 
         // if specified, also include Address Type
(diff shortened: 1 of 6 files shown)


force-app/main/service/AddressService.cls
```diff
@@ -151,8 +155,18 @@ public inherited sharing class AddressService {
...
         sobjDst.put(strFieldPrefix + 'Longitude', addr.Geolocation__Longitude__s);
-        sobjDst.put(strFieldPrefix + 'State', addr.MailingState__c);
-        sobjDst.put(strFieldPrefix + 'Country', addr.MailingCountry__c);
+        String countryField = 'Country';
+        String stateField = 'State';
+        if (orgConfig.isStateCountryPicklistsEnabled()) {
+            if (addr.MailingCountry__c != null && addr.MailingCountry__c.length() <= 3) {
+                countryField = 'CountryCode';
+            }
+            if (addr.MailingState__c != null && addr.MailingState__c.length() == 2) {
```
lparrott, line 164: This is a replacement for relying on the Describe call to all State values, which lead to a bug when two countries have the same state code value... what else can we do to determine if the Address contains a Code vs a full name?

force-app/main/service/AddressService.cls
```diff
@@ -151,8 +155,18 @@ public inherited sharing class AddressService {
         sobjDst.put(strFieldPrefix + 'PostalCode', addr.MailingPostalCode__c);
         sobjDst.put(strFieldPrefix + 'Latitude', addr.Geolocation__Latitude__s);
         sobjDst.put(strFieldPrefix + 'Longitude', addr.Geolocation__Longitude__s);
-        sobjDst.put(strFieldPrefix + 'State', addr.MailingState__c);
-        sobjDst.put(strFieldPrefix + 'Country', addr.MailingCountry__c);
+        String countryField = 'Country';
+        String stateField = 'State';
+        if (orgConfig.isStateCountryPicklistsEnabled()) {
+            if (addr.MailingCountry__c != null && addr.MailingCountry__c.length() <= 3) {
```
force2b, line 161: Out of curiosity, why `<= 3`? I'm pretty sure country codes are always 2 characters ([link](https://help.salesforce.com/s/articleView?id=sf.admin_state_country_picklists_standard_countries.htm&type=5))

force-app/main/default/classes/NPSP_Address.cls
```diff
@@ -341,6 +341,15 @@ public inherited sharing class NPSP_Address implements IAddress {
         address.Geolocation__Latitude__s = (Decimal) getSObjectField(sobjSrc, fieldPrefixSrc, 'Latitude');
         address.Geolocation__Longitude__s = (Decimal) getSObjectField(sobjSrc, fieldPrefixSrc, 'Longitude');
 
+        if (orgConfig.isStateCountryPicklistsEnabled()) {
```
dfuller114, line 344: It seems a little strange that we allow the MailingState and MailingCountry on the address to potentially have a value copied to it in the above block, and then override that value with setting another value in this `if` block. And with the country code null check below, that would mean the country code wouldn't be copied to the address if it already has a value, even if it's not the country code

force-app/main/service/AddressService.cls
dfuller114, line 161: This may have been what we were doing before, but we are already checking in SCP are enabled at this point, so we should just check if the value in the address field is a valid state/country code or not and set it if it's a valid code. which I believe would require a describe on the state/country picklists. It's heavier-handed performance wise but more reliable than checking length

dfuller114 (commented):
added some initial comments

force-app/main/service/AddressService.cls
lparrott: Those are defaults, but 3 character codes also exist: [List of ISO 3166 country codes](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)

force-app/main/default/classes/NPSP_Address.cls
lparrott: Want to take another look now that we've reverted this logic?

force-app/main/service/AddressService.cls
```diff
@@ -63,6 +63,10 @@ public inherited sharing class AddressService {
         sobjDst.put(strFieldPrefixDst + 'Country', sobjSrc.get(strFieldPrefixSrc + 'Country'));
         sobjDst.put(strFieldPrefixDst + 'Latitude', sobjSrc.get(strFieldPrefixSrc + 'Latitude'));
         sobjDst.put(strFieldPrefixDst + 'Longitude', sobjSrc.get(strFieldPrefixSrc + 'Longitude'));
+        if (orgConfig.isStateCountryPicklistsEnabled()) {
+            sobjDst.put(strFieldPrefixDst + 'StateCode', sobjSrc.get(strFieldPrefixSrc + 'StateCode'));
```
npsp-reedestockton, line 67: Do we still want to do this after the code reversion???

lparrott: Yes, this method is for Contact <-> Account and Vice Versa, and we need as much information as possible when we eventually create NPSP Addresses.

npsp-reedestockton (approved):

lparrott (approved):

PR #6973, opened 2022-06-06
[W-10969589 - W-11055239] - Processed Payment Method

gerardorodriguezsf:
# Technical Design Review
[[Technical Design Review] W-10969589 - NPSP](https://salesforce.quip.com/Rs9uAhUVZMRp)

# Changes
- Updated Recurring Donations LWC and apex classes to show the processed payment data
- Updated apex classes to stop the method to get Recurring Donation records  and return an empty list if the RD2 feature is not enabled.

# Work Closed
[W-10969589](https://gus.lightning.force.com/a07EE00000uWnJnYAK)
[W-11055239](https://gus.lightning.force.com/a07EE00000vnCfHYAU)


force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -185,6 +189,7 @@ public with sharing class RD2_ETableController {
     * @author Salesforce.org
     * @date 2022
     * @description process Recurring Donation record to build the status string.
+    * @param recurringDonation recurring dontation record
```
JoseRossiSF, line 192: Minor typo: donation*

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -207,6 +212,32 @@ public with sharing class RD2_ETableController {
         return status;
     }
 
+    /** 
+    * @author Salesforce.org
+    * @date 2022
+    * @description process Recurring Donation record to build the status string.
```
JoseRossiSF, line 218: Shouldn't the description state that the "payment method" string is being built instead of the "status" string?

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -219,7 +250,13 @@ public with sharing class RD2_ETableController {
             npe03__Recurring_Donation__c.npe03__Installment_Period__c,
             npe03__Recurring_Donation__c.InstallmentFrequency__c,
             npe03__Recurring_Donation__c.Day_of_Month__c,
-            npe03__Recurring_Donation__c.Status__c
+            npe03__Recurring_Donation__c.Status__c,
+            npe03__Recurring_Donation__c.CommitmentId__c,
```
lparrott, line 254: Requiring these permissions for non-Elevate customers will cause issues. I would either solve this by not showing those fields if they don't have permission, or doing a separate permission check for Elevate orgs.

lparrott (changes requested):

force-app/main/default/classes/RD2_ETableController.cls
gerardorodriguezsf: Fixed

force-app/main/default/classes/RD2_ETableController.cls
gerardorodriguezsf: @lparrott We will work with both Elevate and Non-Elevate Enhanced Recurring Donations. I still not installed Elevate in my NPSP dev org, I only installed RD2 to it. I believe these fields are part of RD2. Do you have a list of fields to check for permissions only if Elevate is installed?

lparrott: The following fields are only used by customers that have Elevate, so other customers will most likely not have access to them: `CommitmentId__c, CardLast4__c, CardExpirationYear__c, CardExpirationMonth__c, ACH_Last_4__c`

gerardorodriguezsf: Based on [W-11055239](https://gus.lightning.force.com/a07EE00000vnCfHYAU) I updated the method to check if it has enabled the RD2 fields using the method RD2_EnablementService.isRecurringDonations2Enabled, in case it's not enabled it will just return an empty list, otherwise it will perform the current flow to check the permissions, etc.
So, if the RD2 is not enabled, it will not check these fields added when RD2 is enabled.

lparrott: That still won't work, customers can have RD2 enabled, but not Elevate. You could check if `CommitmentId__c` is **not** null, and if so the org must be Elevate connected, then change the permissions check based on that?

lparrott: I see the miscommunication now: just because a field exists in a Salesforce org, doesn't mean the current User has access to it, that's why we do Permission checks. With this check, you are requiring Users to have access to Elevate fields, which most of our customers will not have access to, preventing them from viewing the table.

gerardorodriguezsf: @lparrott Great! now I see it. I update some points:
- It will check Elevate field permissions only if GE_GiftEntryController.isElevateCustomer() is true.
- It will work with Elevate fields only if GE_GiftEntryController.isElevateCustomer() is true.

lparrott (approved):

PR #6976, opened 2022-06-07
[SFCA -- Trust/Security][Implementation] Address potential security vulnerabilities in NPSP (CRLP)

dfuller114:
We added a code notation to bypass the customized rollup service when the SFCA scanner is executed on the NPSP code base since we don't want all CRLP operations to execute in a system context regardless of user permissions.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/CRLP_RollupUI_SVC.cls
+++ b/force-app/main/default/classes/CRLP_RollupUI_SVC.cls
@@ -35,6 +35,9 @@
 * @description Lightning Component Server Controller for the Rollups UI page CRLP_Setup.
 */
 
+// Adding custom rule to avoid this class being scanned by the SFCA scanner since we will not enforce FLS for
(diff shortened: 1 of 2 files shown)


force-app/main/default/classes/CRLP_RollupUI_SVC.cls
```diff
@@ -35,6 +35,9 @@
 * @description Lightning Component Server Controller for the Rollups UI page CRLP_Setup.
 */
 
+// Adding custom rule to avoid this class being scanned by the SFCA scanner since we will not enforce FLS for
```
npsp-reedestockton, line 38: This is a useful comment.  Should it be included in CMT_FilterRule_SEL as well?

PR #6984, opened 2022-06-16
[W-11056833] - Added first and final donation date

gerardorodriguezsf:
# Technical Design Document
[[Technical Design Review] W-11056833 - NPSP](https://salesforce.quip.com/XN0BAYs42BnZ)

# Changes
Updated the TableView inner class from the RD2_ETableController class to have the Final Donation processed data, and check permissions for the npe03__Date_Established__c field. Updated the RD2_ERecurringDonationsSelector class to get the npe03__Date_Established__c field. Updated Recurring Donations LWC to expose First Donation and Final Donation (End Date) data.

# Work Closed
[W-11056833](https://gus.lightning.force.com/a07EE00000vnYAPYA2)


JoseRossiSF (approved):
LGTM

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -129,22 +131,38 @@ public with sharing class RD2_ETableController {
...
-                                                                                controller.getPaymentMethod(recDonation)));
+                                                                                controller.getPaymentMethod(recDonation),
+                                                                                null));
     
+                    if(recDonation.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED) {
+                        fixedRecurringDonationIds.add(recDonation.Id);
+                    }
+                }
+
+                if(!fixedRecurringDonationIds.isEmpty()) {
```
force2b, line 151: This method is getting a little large. Once it starts to have multiple if blocks, nested loops, etc, it's best to consider abstracting out portions of the method into new methods.

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -129,22 +131,38 @@ public with sharing class RD2_ETableController {
...
     private List<RD2_ETableController.TableView> getTableViews() {
-        List<RD2_ETableController.TableView> recurringDonations = new List<RD2_ETableController.TableView>();
+        Map<Id, RD2_ETableController.TableView> recurringDonations = new Map<Id,RD2_ETableController.TableView>();
         if(RD2_EnablementService.isRecurringDonations2Enabled) {
             if(controller.hasAccessToRecurringDonation()) {
+                Set<Id> fixedRecurringDonationIds = new Set<Id>();
+
                 for(npe03__Recurring_Donation__c recDonation : selector.getRecurringDonations()) {
-                    recurringDonations.add( new RD2_ETableController.TableView( recDonation,
+                    recurringDonations.put( recDonation.Id, new RD2_ETableController.TableView( recDonation,
```
voduyemi, line 140: Can we exclude the `finalDonation` parameter here since we're just setting a `null` value? We can also remove it from the `TableView` constructor. If we expect to add more optional parameters to the `TableView` constructor, then we should consider building the TableView object using the builder pattern. The `TEST_RecurringDonationBuilder` class is a good example.

gerardorodriguezsf: Thanks Victor, I updated the inner class constructor.

force-app/main/default/classes/RD2_ETableController.cls
gerardorodriguezsf: Thanks Michael, I moved this logic to a new method.

force2b (approved):

voduyemi (approved):

PR #6999, opened 2022-06-21
[Bug Bounty / H1] Unauthorized user can view/manage Donations - Opportunity Names

danielfuller0814:
We added an admin check on the Opportunity Names visualforce page so that page cannot be accessed explicitly outside of NPSP settings except by an admin.

# Critical Changes

# Changes
- We've improved security on the Opportunity Names visualforce page so that it can only be accessed outside of NPSP settings by an admin.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


voduyemi (approved):
Looks good to me! Thanks @danielfuller0814

npsp-reedestockton (commented):
Code looks fine, but nothing appears when accessing page through NPSP Settings menu without admin permissions.  Is that expected??

PR #7004, opened 2022-06-23
W-11007593 - BDI Contact Matching Using Duplicate Management

callen-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000v9JMIYA2/view

Implements the ability for customers to choose to pre-run their contact duplicate rules for matching contacts in Batch Data Import.
- New option in the BDI Settings UI to allow user to select to use Contact Duplicate rules for Matching
- New UTIL_DuplicateMgmt class that wraps the native DataCloud.findDuplicates method and allows more than 50 records to be passed in at a time
- Changes to BDI_ContactService that branch to new logic if the new option is selected. Also includes a new mapping method to create temp contact SObjects for mapping, logic for selecting a best match, and logic for updating the Data import with the correct contact id if a match is found

# Critical Changes

# Changes
- We've added a new Contact Duplicate Rules matching option for Batch Data Import that runs all active duplicate rules for the Contacts object that are managed in Setup | Duplicate Management.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -343,7 +393,25 @@ public with sharing class BDI_ContactService {
                 }
 
                 di.put(strCx + 'Imported__c', con.id);
-                di.put(strCx + 'ImportStatus__c', con.id != null ? bdi.statusMatchedExisting() : bdi.statusMatchedNew());
+
+                // Determine correct import status based on the type of matching and whether it matched on an existing
+                // record
+                String importStatus;
```
voduyemi, line 399: Do you think we can make this change a little cleaner?  we could extract this change into its own method and also extract these guard clauses that we use to determine the kind of contact into smaller methods so we improve readability.

force-app/main/default/classes/UTIL_DuplicateMgmt.cls
```diff
@@ -0,0 +1,157 @@
...
+                sObjBatch.add(records[nextObjToProcess]);
+                nextObjToProcess++;
+            }
+
+            objsRemaining = objsRemaining - objsToProcess;
+            // Call duplicate management logic
+            Map<SObject, MatchDetail[]> tempDupeResults = findDuplicates(sObjBatch);
+            
+            // Put results in main return map
+            for (SObject sObj : tempDupeResults.keySet()) {
```
voduyemi, line 102: Can we call the `putAll` operation on the main return map here? since we are not really doing any manipulation of the `tempDupeResults` map

force-app/main/default/classes/ComparisonResult.cls
```diff
@@ -0,0 +1,5 @@
+public class ComparisonResult {
```
voduyemi, line 1: We should remove this class since it is not being used

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -1146,6 +1258,15 @@ public with sharing class BDI_ContactService {
         } private set;
     }
 
+    /*******************************************************************************************************
+    * @description whether the Duplicate Management matching option is in the contact matching rules
+    */
+    private boolean isDuplicateManagement {
+        get {
+            return (bdi.diSettings.Contact_Matching_Rule__c == System.Label.bdiContactMatchDupeMgmt ? true : false);
```
voduyemi, line 1266:
```suggestion
            return bdi.diSettings.Contact_Matching_Rule__c == System.Label.bdiContactMatchDupeMgmt;
```

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -92,6 +121,26 @@ public with sharing class BDI_ContactService {
         set;
     }
 
+    private Map<String,String> mapDIHomeAddrToContact {
```
voduyemi, line 124: Can we add a unit test to validate this behaviour?

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -1168,4 +1289,117 @@ public with sharing class BDI_ContactService {
...
+
+        // Return if there were no sObjects added for dupe matching.
+        if (sObjsForDuplicateManagement == null || sObjsForDuplicateManagement.size() == 0) {
+            return;
+        }   
+
+        sObjToMatchDetails = dupeMgmtUtil.findDuplicatesBulk(new List<SObject>(sObjsForDuplicateManagement));
+
+        // Return if there were no dupe results
+        if (sObjToMatchDetails == null || sObjToMatchDetails.size() == 0) {
```
voduyemi, line 1323: `sObjToMatchDetails` is never going to be `null` because it is newed up in `UTIL_DuplicateMgmt`

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -1168,4 +1289,117 @@ public with sharing class BDI_ContactService {
         set;
     }
 
+    /*******************************************************************************************************
+    * @description updates DI records with salesforce ids if it finds a match using duplicate management rules
+    */
+    private void performDuplicateManagement() {
```
voduyemi, line 1295: This method can be further broken down into smaller methods to help with readability here. The population of the list of DIs for `Contact1` and `Contact2` can happen later in processing since we only want to create those lists if we found any duplicates

force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -4077,6 +4077,14 @@ To check your GAU Allocation settings, go to Donations | GAU Allocations.</value
         <shortDescription>bdiContactCustomIdError</shortDescription>
         <value>The Contact Custom Unique ID setting field {0} must have matching fields in the NPSP Data Import object whose API Names are {1} and {2}.</value>
     </labels>
+    <labels>
```
voduyemi, line 4080: Do we need to tag Erin here for a review?

voduyemi (changes requested):
Thanks! @callen-sfdo. I left some initial comments, let me know if you would like chat about them.

force-app/main/default/classes/UTIL_DuplicateMgmt.cls
```diff
@@ -0,0 +1,157 @@
...
+*/
+public with sharing class UTIL_DuplicateMgmt {
+
+    /**
+    * @description Returns map of duplicate matching details for the records passed in.  Note:
+    * Will throw error if more than 50 SObjects are passed in.
+    * @param List<sObject> list of Sobjects that duplicates will be located for
+    * @return Map<SObject, BDI_DuplicateService.MatchDetail[]>
+    */
+    public static Map<SObject, MatchDetail[]> findDuplicates(List<SObject> records) {
```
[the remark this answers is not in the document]
callen-sfdo: @voduyemi It doesn't necessarily need to be static although in general I make utility class methods static unless they need to be non-static.  I had to make the findDuplicatesBulk class non-static so that I could mock it correctly.  

For the 50 max limit, the findDuplicates core method already handles that error cleanly so there is no reason to add another layer of error messages/handling.  I actually removed one I had in there since it was redundant.

force-app/main/default/classes/ComparisonResult.cls
callen-sfdo: Yep, I will remove this, forgot to delete it.

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Yeah, I can include it in my existing test method that is creating contact Sobjs using this.

force-app/main/default/labels/CustomLabels.labels-meta.xml
ErinWiedemer: @callen-sfdo I sent you some questions in your WI Chatter feed. I need a little more info on this work and I'll have to pass all UI text through the UI text review team next week.

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -92,6 +121,26 @@ public with sharing class BDI_ContactService {
         set;
     }
 
+    private Map<String,String> mapDIHomeAddrToContact {
+        get {
+            mapDIHomeAddrToContact = new Map<String,String> {'Home_Street__c' => 'MailingStreet' ,
```
lparrott, line 126: Is there a risk of namespace issues here?

force-app/main/default/classes/BDI_ContactService.cls
lparrott, line 1266: Is there no longer a way to set Matching Rules per batch? I assume no, but I know there's a field on the DI Batch object.

callen-sfdo: There is, the batch would be converted into Data import settings overriding the default ones if it had different matching criteria.  But we should definitely test it with the new matching method.

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Yeah let me take another look

force-app/main/default/classes/UTIL_DuplicateMgmt.cls
callen-sfdo: Yeah that should work just as well.

force-app/main/default/classes/UTIL_DuplicateMgmt.cls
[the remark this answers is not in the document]
callen-sfdo: We might as well make it private since at most it would save a little processing and could be dangerous.

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -1168,4 +1289,117 @@ public with sharing class BDI_ContactService {
...
+            }
+
+            // Only add a contact for matching if there is information for it and there isn't already an Id specified
+            if (di.Contact2Imported__c == null && isContactSpecified(di, 'Contact2')) {
+                addSobjToDIMap(c2SObjectToDIs, di, 2);
+            }
+        }
+
+        // Return if there were no sObjects added for dupe matching.
+        if (sObjsForDuplicateManagement == null || sObjsForDuplicateManagement.size() == 0) {
```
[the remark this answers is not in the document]
callen-sfdo: Will update

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Made these changes.

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Made this update

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -959,6 +1013,49 @@ public with sharing class BDI_ContactService {
...
+        // temp object is created.
+        try {
+            for(String homeAddrField : mapDIHomeAddrToContact.keySet()) {
+                BDI_DataImportService.copyDIFieldToDestinationRecord(di, homeAddrField, tempCont, mapDIHomeAddrToContact);
+            }
+        } catch (Exception e) {
+            System.debug('Exception is: ' + e.getMessage() + ' ' + e.getStackTraceString());
+        }
+
+        System.debug('TempContis: ' + tempCont);
```
voduyemi, line 1050:
```suggestion

```
```suggestion

```

voduyemi (changes requested):
Hey @callen-sfdo, looking really good, just some little comments!

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Updated these to dynamically apply namespace

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -959,6 +1013,49 @@ public with sharing class BDI_ContactService {
...
+        }
+
+        // Map Home Address fields to contact for matching purposes.  Note that we are not checking to see if it copied
+        // anything since home address is not part of the normal contact mapping and shouldn't be used to determine if a
+        // temp object is created.
+        try {
+            for(String homeAddrField : mapDIHomeAddrToContact.keySet()) {
+                BDI_DataImportService.copyDIFieldToDestinationRecord(di, homeAddrField, tempCont, mapDIHomeAddrToContact);
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
callen-sfdo: I removed this, I had it in there temporarily and left it in by accident.

force-app/main/default/classes/BDI_ContactService.cls
```diff
@@ -1168,4 +1274,153 @@ public with sharing class BDI_ContactService {
...
+    * enabled and whether dry run was active.
+    * @param di the data import record from which the contact sobj will be generated
+    * @param con The contact 1/2 in context for updating import results.
+    * @param iCx whether the method should look at Contact 1 fields or Contact 2 fields.
+    * return String
+    */
+    private String determineImportStatus (DataImport__c di, Contact con, Integer iCx) {
+        // Determine correct import status based on the type of matching and whether it matched on an existing
+        // record
+        String importStatus;
```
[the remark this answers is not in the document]
callen-sfdo: Sounds good.

force-app/main/default/classes/BDI_ContactService.cls
callen-sfdo: Will do.

ErinWiedemer (approved):
I like the changes, @callen-sfdo.

voduyemi (approved):
Looks good to me, thanks @callen-sfdo

PR #7010, opened 2022-06-27
Attempt to fix NPSP push failures when encryption is enabled

lparrott:
[W-11140971](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000wy1T2YAI/view)
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


lparrott:
> This looks good, but how can we definitively test this to be sure it won't be a problem for future pushes?

Without knowing **which** fields customers have encrypted, it's impossible to know for sure 😩 This particular issue has come up with quite a few orgs though, so it's a start!

danielfuller0814:
> This looks good, but how can we definitively test this to be sure it won't be a problem for future pushes?
The only definitive way would be to not use DML for contacts so we wouldn't need to query them.

danielfuller0814 (approved):

voduyemi (approved):

PR #7012, opened 2022-06-28
NPSP Settings Admin Checks Using Visualforce Component

danielfuller0814:
This includes the new Visualforce component we are introducing to display insufficient permissions messaging in NPSP Settings as well as references to the component in several of the NPSP Settings pages

We are now also blocking the entire NPSP Settings page if the user is not an admin so that the user does not have to wait for the settings page to load only to find out that they cannot access all the pages within it.

# Critical Changes

# Changes
- We've improved security of NPSP Settings by restricting access to Admins only.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


danielfuller0814:
I want to call out that there were several VF pages that contained duplicated admin checks in the UI. Several of the pages contained calls to `UTIL_Permission` to perform object-level CRUD checks. This isn't needed since the `Modify All Data` permission is required to view any of these pages, and this system permission automatically grants CRUD for all standard and custom objects. I've removed these access checks anywhere they were used in the UI. For more info, see https://help.salesforce.com/s/articleView?id=000321298&type=1

force-app/main/default/pages/STG_PanelOppBatch.page
```diff
@@ -1,25 +1,8 @@
 <apex:page controller="STG_PanelOppBatch_CTRL" docType="html-5.0" standardStylesheets="false">
     <apex:includeLightning/>
```
npsp-reedestockton, line 2: I don't believe we need this now.

npsp-reedestockton (commented):
Nice work.  Just a few minor comments.

ErinWiedemer (approved):
LGTM, Daniel!

voduyemi (approved):
Looks good to me! Thanks @danielfuller0814

npsp-reedestockton (approved):

voduyemi (approved):
Looks good @danielfuller0814

PR #7019, opened 2022-07-05
[W-11090994] - Update status if lapsed payment

gerardorodriguezsf:
# Technical Design Review
[[Technical Design Review] W-11090994 - NPSP](https://salesforce.quip.com/FkHdAC6CHt6B)

# Changes
Updated the rd2EntryForm LWC to auto-update the Status value if this component is used from the Update Payment action from the Recurring Donations LWC.

# Work Closed
[W-11090994](https://gus.lightning.force.com/a07EE00000wFWsCYAW)


maurimart (approved):
LGTM

JoseRossiSF (approved):

force-app/main/default/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -274,6 +274,10 @@ export default class rd2EntryForm extends LightningElement {
                 this.rd2State.periodType = 'Advanced';
                 this.cssExperienceElevate = 'slds-hide'
             }
+            if(this.isPaymentModal && this.rd2State.recurringStatus === 'Lapsed') {
```
force2b, line 277: I suspect that "Lapsed" is the picklist api value you're checking for, so that should be fine.

However, what about the two "Active" values below? Are those rendered to the user as just plain text and not able to be translated? The RD.Status field uses picklist values that are translated. Maybe those can be re-used here or another solution is needed for non-english users. Unless you've tested this in a non-english language and confirm it works properly in the display.

gerardorodriguezsf: These are picklist api values and not labels, they are not rendered to the customer, just auto-update field values. Also this has been tested and it's working for non-english work.

lparrott (approved):

PR #7013, opened 2022-06-28
[W-11170848/W-11164006/W-11147296/W-11148437/W-11374554]  Elevate Experience Sites

gerardorodriguezsf:
# Technical Design Review
[[Technical Design Review] W-11170848/W-11164006/W-11147296/W-11148437/W-11374554 - NPSP](https://salesforce.quip.com/oT9uAQgBEAFZ)

# Changes
As part of this task, we first fixed the Elevate component currently used in the CRM to work with Elevate fields so that it can also work from the Digital Experience Site. Then part of the existing logic was reused to edit the Elevate Recurring Donation data and have it reflected in Elevate. It was updated the Elevate fields widget to be able to use different style when it is used from the Digital Experience Sites. Added Amount validation only for Experience Sites. Removed Elevate message from the spinner component only for Experience Sites.

# Work Closed
[W-11170848](https://gus.lightning.force.com/a07EE00000xN9VMYA0)
[W-11164006](https://gus.lightning.force.com/a07EE00000xJ8mIYAS)
[W-11147296](https://gus.lightning.force.com/a07EE00000x8xjUYAQ)
[W-11148437](https://gus.lightning.force.com/a07EE00000x8TqiYAE)
[W-11374554](https://gus.lightning.force.com/a07EE000010SVBvYAO)
[W-11402375](https://gus.lightning.force.com/a07EE000010zNmeYAE)
[W-11415744](https://gus.lightning.force.com/a07EE000011GgkbYAC)

PR #7020, opened 2022-07-05
[W-11379845] - Fix final donation date

gerardorodriguezsf:
# Technical Design Review
[[Technical Design Review] W-11379845 - NPSP](https://salesforce.quip.com/x8KIAjUPoJ8C)

# Changes
Updated the Final Donation Date logic updating the getFinalDonations apex method from apex class RD2_ETableController.

# Issues Closed
[W-11379845](https://gus.lightning.force.com/a07EE000010eZxqYAE)


voduyemi (commented):
Thanks @gerardorodriguezsf, is there a reason why the changes in this and this PR #7013 were duplicated?

gerardorodriguezsf:
> Thanks @gerardorodriguezsf, is there a reason why the changes in this and this PR #7013 were duplicated?

@voduyemi Sorry, I fixed the PR base.

gerardorodriguezsf:
> Thanks @gerardorodriguezsf, is there a reason why the changes in this and this PR #7013 were duplicated?

In the old implementation there was an internal custom logic to cache the recurring donation record, and there was a lot of logic to get the final donation date, now it's more clear and it's avoiding the custom cache issue.

force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
```diff
@@ -75,7 +75,8 @@ public inherited sharing class RD2_ERecurringDonationsSelector {
                                 + 'npe03__Date_Established__c, '
                                 + 'npe03__Next_Payment_Date__c, '
                                 + 'npe03__Total_Paid_Installments__c, '
-                                + 'Lastmodifieddate '
+                                + 'Lastmodifieddate, '
```
force2b, line 78:
```suggestion
                                + 'LastModifiedDate, '
```

force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
```diff
@@ -75,7 +75,8 @@ public inherited sharing class RD2_ERecurringDonationsSelector {
...
+                                + '(SELECT Active__c, Campaign__c, Campaign__r.Name, StartDate__c, EndDate__c, DayOfMonth__c, InstallmentAmount__c, InstallmentFrequency__c, InstallmentPeriod__c, PaymentMethod__c, RecurringDonation__c, IsPause__c, StatusReason__c FROM RecurringDonationSchedules__r) '
```
force2b, line 79: Note that `RecurringDonationSchedule__c` is a protected custom object. An end user will not be able to query that object in code that is using `with sharing`. This particular class uses `inherited sharing` (as it should), but it's possible this method gets called from a controller that is using `with sharing`. I'm pretty sure that will cause the query to fail because the customer cannot see that object. 

You may need to make a separate query against the schedule object using an inner class (within this class) where the inner class explicitly uses 'without sharing'. You'd have to refactor to use two queries rather than a single query with a sub-query.

I do not recommend modifying this class or the controller class to use `without sharing`.

force2b (changes requested):

force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
gerardorodriguezsf: @force2b Based on [[Technical Design Review] W-11379845 - NPSP](https://salesforce.quip.com/x8KIAjUPoJ8C) > Testing Considerations > Create Environment to test: The Admin will have to allow READ access to some objects to the partner profile, including Recurring Donation Schedules.
RD2_ETableController is public with sharing.
It's not failing the query and we are getting the information needed to process and get the final donation date.
It's working with 'with sharing'.

force2b:
This PR has 80 changed files and many of them are completed unrelated to this work?

npsp-reedestockton:
> It wasn't from our side, we only updated 2 files in this PR, but this pull request added the other files. what should we do?

@gerardorodriguezsf @force2b The merge you mentioned was a [standard merge into feature/240](https://github.com/SalesforceFoundation/NPSP/pull/7012).  The large number of files from the merged PR are apparently in feature/240__fix_final_donation_date, but not in feature/240__payment_with_elevate.  It might be a good idea to check in the RelEng to help understand why.

npsp-reedestockton:
@davidmreed @gerardorodriguezsf @force2b Not sure what is going on here, but something definitely seems amiss.  There are roughly 80 files identified as changed in this PR that were not originally a part of the PR, but that seems incorrect.

For example, check the changes flagged in STG_PanelRD.page.  The files appear to be identical in feature/240__payment_with_elevate and feature/240__fix_final_donation_date, yet they it is flagged as a change.  Both branches seem to have received the down merge from 240, but the comparison still seems to reflect the status before the down merge in feature/240__payment_with_elevate.

Recommendations??

PR #7026, opened 2022-07-06
[W-11379845] - New final donation date logic

gerardorodriguezsf:
# Technical Design Review
[[Technical Design Review] W-11379845 - NPSP](https://salesforce.quip.com/x8KIAjUPoJ8C)

# Changes
Updated the Final Donation Date logic updating the getFinalDonations apex method from apex class RD2_ETableController.

# Issues Closed
[W-11379845](https://gus.lightning.force.com/a07EE000010eZxqYAE)

### DIFF
--- a/force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
+++ b/force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
@@ -75,7 +75,8 @@ public inherited sharing class RD2_ERecurringDonationsSelector {
                                 + 'npe03__Date_Established__c, '
                                 + 'npe03__Next_Payment_Date__c, '
                                 + 'npe03__Total_Paid_Installments__c, '
-                                + 'Lastmodifieddate '
+                                + 'Lastmodifieddate ,'
+                                + '(SELECT Active__c, Campaign__c, Campaign__r.Name, StartDate__c, EndDate__c, DayOfMonth__c, InstallmentAmount__c, InstallmentFrequency__c, InstallmentPeriod__c, PaymentMethod__c, RecurringDonation__c, IsPause__c, StatusReason__c FROM RecurringDonationSchedules__r) '
                                 + 'FROM npe03__Recurring_Donation__c '
                                 + 'WHERE npe03__Contact__c = \'' + currentContactId + '\' ';
 
--- a/force-app/main/default/classes/RD2_ETableController.cls
+++ b/force-app/main/default/classes/RD2_ETableController.cls
@@ -156,7 +156,7 @@ public with sharing class RD2_ETableController {
         RD2_StatusMapper statusMapper = RD2_StatusMapper.getInstance();
         
         Map<Id, RD2_ETableController.TableView> recurringDonations = new Map<Id,RD2_ETableController.TableView>();
-        Set<Id> fixedRecurringDonationIds = new Set<Id>(); 
+        Set<npe03__Recurring_Donation__c> fixedRecurringDonations = new Set<npe03__Recurring_Donation__c>(); 
         for(npe03__Recurring_Donation__c recDonation : rdRecords) {
             recurringDonations.put( recDonation.Id, new RD2_ETableController.TableView( recDonation,
                                                                         controller.getFrequency(recDonation),
@@ -166,12 +166,12 @@ public with sharing class RD2_ETableController {
 
             if(recDonation.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED 
                 && statusMapper.getState(recDonation.Status__c) != RD2_Constants.STATUS_CLOSED) {
-                fixedRecurringDonationIds.add(recDonation.Id);
+                fixedRecurringDonations.add(recDonation);
             }
         }
 
-        if(!fixedRecurringDonationIds.isEmpty()) {
-            Map<ID, Date> mapRDFinalDate = getFinalDonations(fixedRecurringDonationIds);
+        if(!fixedRecurringDonations.isEmpty()) {
+            Map<ID, Date> mapRDFinalDate = getFinalDonations(fixedRecurringDonations);
             for(Id rdID : mapRDFinalDate.keySet()) {
                 RD2_ETableController.TableView tableView = recurringDonations.get(rdID);
                 tableView.finalDonation = mapRDFinalDate.get(rdID);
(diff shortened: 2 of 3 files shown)


force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
```diff
@@ -75,7 +75,8 @@ public inherited sharing class RD2_ERecurringDonationsSelector {
                                 + 'npe03__Date_Established__c, '
                                 + 'npe03__Next_Payment_Date__c, '
                                 + 'npe03__Total_Paid_Installments__c, '
-                                + 'Lastmodifieddate '
+                                + 'Lastmodifieddate ,'
+                                + '(SELECT Active__c, Campaign__c, Campaign__r.Name, StartDate__c, EndDate__c, DayOfMonth__c, InstallmentAmount__c, InstallmentFrequency__c, InstallmentPeriod__c, PaymentMethod__c, RecurringDonation__c, IsPause__c, StatusReason__c FROM RecurringDonationSchedules__r) '
```
force2b, line 79: Continuing this conversation from the original [PR thread](https://github.com/SalesforceFoundation/NPSP/pull/7020#discussion_r914173740).

force2b: Because the object is protected, an admin will never be able to assign any CRUD or FLS to the object or its fields because they cannot see the object anywhere in their org. It's like it doesn't exist (to them). 

It is in a MasterDetail relationship as you note, so technically if the user has READ on the RD object they will have READ on the RDSchedule object too. What may break this though is the use of `WITH_SECURITY_ENFORCED` on the query. That will enforce field perms, but it will never be possible for an Admin to grant field visibility to RDSchedule fields because they cannot see the object to do so. I'm almost positive we saw this elsewhere. I am nearly sure this will fail in a package context.

force2b (changes requested):

force-app/main/default/classes/RD2_ERecurringDonationsSelector.cls
gerardorodriguezsf: Ok, I get it. I'll work on this.

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -365,4 +365,32 @@ public with sharing class RD2_ETableController {
         
         return permissions.canRead(npe03__Recurring_Donation__c.SObjectType, fields);
     }
+
+    private without sharing class RDScheduleClass {
+        public Map<Id, List<RecurringDonationSchedule__c>> queryRows(Set<Id> recordIds) {
+            Map<Id, List<RecurringDonationSchedule__c>> mapRDSchedule = new Map<Id, List<RecurringDonationSchedule__c>>();
+            for(RecurringDonationSchedule__c schedule : [   SELECT Active__c,
```
force2b, line 372: This is going to fail in an org with more than 50K schedule records. The query must be filtered by either `RecurringDonation__c` or probably more usefully by `RecurringDonation__r.npe03__Contact__c`. In other words, this query must only retrieve the specific schedule records for the Donor (Contact) that it's being displayed for.

force2b (changes requested):

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -365,4 +365,35 @@ public with sharing class RD2_ETableController {
...
+                                                                InstallmentAmount__c, 
+                                                                InstallmentFrequency__c, 
+                                                                InstallmentPeriod__c, 
+                                                                PaymentMethod__c, 
+                                                                RecurringDonation__c, 
+                                                                IsPause__c, 
+                                                                StatusReason__c 
+                                                            FROM RecurringDonationSchedule__c
+                                                            WHERE RecurringDonation__c IN :recordIds
+                                                                AND RecurringDonation__r.npe03__Contact__c = :USER_UserService.getContactIdByUserId()
```
force2b, line 387: Why have a filter on both RecurringDonation and Contact? Isn't one sufficient - especially if you've already narrowed down the specific RD Id's you want the schedule records for?

force2b (changes requested):

force-app/main/default/classes/RD2_ETableController.cls
gerardorodriguezsf: I agree, I updated it.

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -365,4 +365,34 @@ public with sharing class RD2_ETableController {
...
+                                                                InstallmentAmount__c, 
+                                                                InstallmentFrequency__c, 
+                                                                InstallmentPeriod__c, 
+                                                                PaymentMethod__c, 
+                                                                RecurringDonation__c, 
+                                                                IsPause__c, 
+                                                                StatusReason__c 
+                                                            FROM RecurringDonationSchedule__c
+                                                            WHERE RecurringDonation__c IN :recordIds
+                                                            LIMIT 50000]) {
```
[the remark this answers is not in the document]
gerardorodriguezsf: Ok, I removed it. Thanks.

force-app/main/default/classes/RD2_ETableController.cls
```diff
@@ -303,7 +305,6 @@ public with sharing class RD2_ETableController {
         String nextDonation = '';
 		String nextDonationStringDate = String.valueOf(recurringDonation.npe03__Next_Payment_Date__c);
         if(recurringDonation.RecurringType__c == RD2_Constants.RECURRING_TYPE_FIXED && !String.isEmpty(nextDonationStringDate) ) {
-            //nextDonation = System.Label.RD2_PaidInstallments + ' ' + String.valueOf(recurringDonation.npe03__Total_Paid_Installments__c) +'/'+ String.valueOf(recurringDonation.npe03__Installments__c) + ', ' + System.Label.RD2_NextDonationDate + ' ' + String.valueOf(recurringDonation.npe03__Next_Payment_Date__c);
 			nextDonation = System.Label.RD2_PaidInstallments + ' ' + String.valueOf(recurringDonation.npe03__Total_Paid_Installments__c) +'/'+ String.valueOf(recurringDonation.npe03__Installments__c) + ', ' + System.Label.RD2_NextDonationDate + ' ' + (recurringDonation.npe03__Next_Payment_Date__c).month() + '/'+ (recurringDonation.npe03__Next_Payment_Date__c).day() + '/' + (recurringDonation.npe03__Next_Payment_Date__c).year();
```
force2b, line 308: This hardcodes formatting the date as mm/dd/yyyy. That is a US-centric date format. It should use the users current locale to properly format the date. Typically this is done in the LWC UI using the date component, but you maybe able to format it with date.format(). Testing needed for other locales however.

force2b (changes requested):

force-app/main/default/classes/RD2_ETableController.cls
force2b: Thank you for the update. Will keep an eye out for that PR.

force2b (approved):

maurimart (approved):

maurimart (approved):

maurimart (approved):

JoseRossiSF (approved):

PR #7029, opened 2022-07-07
W-10707489 - Ignore existing Default GAU Allocations when processing new GAU Allocations in BDI

callen-sfdo:
Currently in BDI if you have a DI with an existing Oppt or Payment entered that has an existing GAU Allocation linked to it, then BDI will intentionally throw an error since it cannot properly resize existing GAUs. This doesn't need to apply to Default GAU Allocations though since those can be automatically resized.

This ticket basically excludes the default GAU from the check for existing GAU Allocations since it is not relevant and will automatically resize if needed. This will prevent the Error from being thrown in a large number of cases in an org with Default GAU enabled.

Note: This ticket doesn't fix this issue if 'Payment Allocations' are enabled because in that scenario only the Opportunity Allocations are updated, and the payment allocations become out of sync with the Opportunity Allocations.  It will continue to throw the error in that case.

# Critical Changes

# Changes
- We've improved Bulk Data Import for orgs that have Default GAU enabled. Default GAU is now excluded from the check for existing GAU allocations, which reduces import errors.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/BDI_GAUAllocationsUtil.cls
+++ b/force-app/main/default/classes/BDI_GAUAllocationsUtil.cls
@@ -146,14 +146,30 @@ public with sharing class BDI_GAUAllocationsUtil {
                 && utilPerm.canRead(npe01__OppPayment__c.getSObjectType(),paymentFields)) {
 
                 String queryString = ' SELECT Id, ' +
-                        'General_Accounting_Unit__c, ' +
-                        'Opportunity__c,' +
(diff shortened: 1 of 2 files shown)


voduyemi (approved):
Looks good to me. Thanks @callen-sfdo

lparrott (approved):
Great work on this one! Code looks good and my testing all passed.

PR #7040, opened 2022-07-15
W-11064555 - Friendly Error message for Import Selected with Batch DI

callen-sfdo:
Previously if a user attempted to use the 'Import Selected Data Import Records' button on a selection that included a Data Import record that belonged to a batch they would receive an error that a the batch field was missing from the query.  

We contemplated adding the field to the query and allowing users to use the button with Batch DI records, but on further discussion with all the stakeholders we came to the decision that it is much safer to throw a friendly error that requests they use Gift Entry to process batch records (especially since Elevate Payments would not work if the DI record is processed outside gift entry).  Instead we are providing a friendly error message that appears immediately after they click the 'Import Selected Data Import Records' button and have a Batch DI selected.

Also, in this fix I am resolving an unrelated typo in a field label for Notification Recipient Email.

# Critical Changes

# Changes

# Issues Closed
- [Known Issue](https://trailblazer.salesforce.com/issues_view?id=a1p4V000001cQHrQAM): NPSP (Summer 22) - Utilizing "Import Selected Data Import Records" button to Import Data Import Records results in SOQL Error and Failure
# Community Ideas Delivered

# New Metadata

# Deleted Metadata


daniel-fuller (approved):
looks good!

npsp-reedestockton (approved):
One fairly trivial comment, but looks good!!

ErinWiedemer (approved):
LGTM!

PR #7047, opened 2022-07-20
[W-11455814] - Day of the Month field not being displayed

JoseRossiSF:
# Work Item
[W-11455814](https://gus.lightning.force.com/a07EE000011qOL0YAM)

# Issues Closed
Fixed issue in which a user changed the Installment Period for non-monthly Elevate recurring donations to "Monthly" yet the Day of Month field did not appear.

### DIFF
--- a/force-app/main/default/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
+++ b/force-app/main/default/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
@@ -33,6 +33,7 @@ export default class rd2EntryFormScheduleSection extends LightningElement {
     cssHideOnlyPaymentModal;
     cssHideOnlyAmountFrequencyModal;
     cssLastDay;
+    firstRendered = true;
     @track isMonthlyDonation = false;
     @api isElevateDonation = false;
     @api isInitiallyMonthlyDonation = false;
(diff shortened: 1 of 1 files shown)


maurimart (approved):
LGTM

force-app/main/default/lwc/rd2EntryFormScheduleSection/rd2EntryFormScheduleSection.js
```diff
@@ -135,6 +124,7 @@ export default class rd2EntryFormScheduleSection extends LightningElement {
                 this.rd2State.dayOfMonth = dd === 31 ? LAST_DAY_OF_MONTH : dd;
             }
             this.cssLastDay = !this.isMonthlyDonation || this.isPaymentModal ? 'slds-hide' : 'slds-p-right_small slds-p-left_small slds-size_12-of-12 slds-large-size_4-of-12 fixExperienceDayOfMonth';
+            this.firstRendered = false;
```
force2b, line 127: `firstRendered` as a var name is not explicitly tied to ExperienceSite users, so should it be set to false regardless of whether `experienceSite` is true to not?

JoseRossiSF: Hi @force2b, it should be false regardless of the component being rendered either on the CRM or on the ExperienceSite.
I have moved the variable assignment to outside the condition where it was being changed. Please let me know if you consider further changes are required.

force2b (approved):

PR #7049, opened 2022-07-20
[W-11455021] - Fixing ACH text for Experience Sites

gerardorodriguezsf:
# Changes
Removed spinner text related to Elevate when updating the payment method of an Elevate RD and the payment is ACH.

# Issues Closed
[W-11455021](https://gus.lightning.force.com/a07EE000011qCavYAE)

### DIFF
--- a/force-app/main/default/lwc/rd2EntryForm/rd2EntryForm.js
+++ b/force-app/main/default/lwc/rd2EntryForm/rd2EntryForm.js
@@ -78,6 +78,7 @@ import ACCOUNT_PRIMARY_CONTACT_LAST_NAME from "@salesforce/schema/Account.npe01_
 const ELEVATE_WIDGET_EVENT_NAME = "rd2ElevateCreditCardForm";
 const CREDIT_CARD = "Credit Card";
 const ACH = "ACH";
+const ELEVATE = "elevate";
 
 export default class rd2EntryForm extends LightningElement {
     customLabels = Object.freeze({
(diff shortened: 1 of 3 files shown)


JoseRossiSF (approved):

force-app/main/default/lwc/rd2EntryForm/rd2EntryForm.js
```diff
@@ -676,8 +677,8 @@ export default class rd2EntryForm extends LightningElement {
     async processCommitmentSubmit(allFields) {
         try {
             if (this.isElevateWidgetDisplayed()) {
-                this.loadingText = this.rd2Service.getPaymentProcessingMessage(this.rd2State.paymentMethod);
-
+                
+                this.loadingText = this.isExperienceSite && this.rd2Service.getPaymentProcessingMessage(this.rd2State.paymentMethod).includes(ELEVATE) ? '' : this.rd2Service.getPaymentProcessingMessage(this.rd2State.paymentMethod);
```
force2b, line 681: This feels fragile to me. The code is looking for the word "elevate" in text to determine if Elevate is enabled or not (I believe). Is there a better way to determine if Elevate is enabled or not? Assuming that donors cannot create a new Recurring Donation in this UI, then I believe the RD.CommitmentId__c field will have a value if it's an Elevate RD. Would that be a more reliable check to determine if Elevate is enabled?

gerardorodriguezsf: We should not show the word 'elevate' in the spinners if the component is used from the experience sites. i was checking the custom label value, but I agree it's a better idea to check the RD.CommitmentId__c. Let me apply the update.

force2b: @gerardorodriguezsf  Thanks. Is the intent to display nothing at all if "Elevate" is anywhere in the PaymentProcessing message?

gerardorodriguezsf: It' showing 'elevate' word only for ACH payments. I updated the condition to check the payment method.

force2b (approved):

PR #7068, opened 2022-08-01
NPSP 240 Localization - all languages

daniel-fuller:
# Critical Changes

# Changes
Translated labels for current language set.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

# Unaggregated Pull Requests
* Additional labels for 240 NPSP [[PR7070](https://github.com/SalesforceFoundation/NPSP/pull/7070)]


andrewyu-salesforce:
@daniel-fuller there are errors being thrown on the duplicate translation. I remembered running into the same issue before but I am not sure if it's caused by the script or not.

force-app/main/default/classes/BDI_Donations_TEST2.cls
```diff
@@ -127,17 +127,17 @@ public with sharing class BDI_Donations_TEST2 {
         }
 
         //Confirm that the DIs are Imported and that the correct status is in place for each Oppt/Payment
-        System.assertEquals(System.label.bdiImported,di1.Status__c);
+        System.assertEquals('Imported', di1.Status__c);
```
lparrott, line 130: Huh, is the DI logic hard coding the status to English?

force-app/main/default/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -682,11 +682,11 @@ public without sharing class UTIL_CustomSettingsFacade {
     }
 
     private static void configHouseholdNamingSettings(Household_Naming_Settings__c hns) {
-        hns.Household_Name_Format__c = '{!LastName} ' + label.npo02.DefaultHouseholdName;
+        hns.Household_Name_Format__c = '{!LastName} ' + 'Household';
```
lparrott, line 685: Interesting, so the Setting needs to be in English, but I assume the actual Naming logic uses translations?

lparrott (approved):
Full disclosure: I did *not* look at every file, but the deployment is working, so I assume all is well. Just had a few comments about Unit test changes.

andrewyu-salesforce (approved):

force-app/main/default/classes/BDI_Donations_TEST2.cls
daniel-fuller: This is comparing a field value that doesn't get translated anyway. So this assert was failing since the label was being translated but the `Status__c` field value was not even in a non-English org.

force-app/main/default/classes/UTIL_CustomSettingsFacade.cls
lparrott: After discussing with @daniel-fuller, we're going to revert this change and instead update the Unit test to work with non-English orgs.

daniel-fuller:
@davidjray Could I get a RelEng approval on this PR?

davidjray (approved):

PR #7070, opened 2022-08-02
Additional labels for 240 NPSP

dmkud:
# Critical Changes

# Changes

Additional labels from the latest update

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -652,11 +652,6 @@ tasks:
         options:
             output_style: salesforce
             elements:
-                # Strip all underlying field translations, leaving translations for new fields in these objects
-                - path: force-app/main/default/*ranslations/**/*ranslation-meta.xml
-                  xpath: /*/*[re:match(ns:name, '...+__..+')]
-                - path: force-app/tdtm/*ranslations/**/*ranslation-meta.xml
-                  xpath: /*/*[re:match(ns:name, '...+__..+')]
                 # Strip any underlying package object label translations
                 - path: force-app/main/default/objectTranslations/npe01**/**objectTranslation-meta.xml
                   xpath: //ns:layouts|//ns:nameFieldLabel|//ns:gender|//ns:caseValues|//ns:plural|//ns:startsWith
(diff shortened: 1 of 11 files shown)


daniel-fuller:
@lparrott One thing I noticed here is that our cleanup script is stripping out several of the `quickActions` elements that have been translated, even the new quick actions we've added for 240 like `Refund`. The newer `Paused` quick action is stripped as well. It looks like the part of the script that does this is in the screenshot below. @force2b, was there any reason why we would want to strip quick actions or do we just need to fix this in the script? The script stripped the following quick actions:

- npe03__Recurring_Donation__c.Pause
- npe03__Recurring_Donation__c.New_Opportunity
- npe01__oppPayment__c.Refund
- npe03__Recurring_Donation__c.Quick_Update
- npo02__Household__c.Update_Household
- Address__c.Update_Address
- npe03__Recurring_Donation__c.Close_Recurring_Donation
- npe01__OppPayment__c.Quick_Update

lparrott:
@daniel-fuller Those lines are removing translations with a double underscore in the name field, which I think is intended to target underlying packages, though I'm not sure why. 

It could be time to retire that part of the script, because we definitely want the Quick Action Label translations included!

daniel-fuller:
> @daniel-fuller Those lines are removing translations with a double underscore in the name field, which I think is intended to target underlying packages, though I'm not sure why.
> 
> It could be time to retire that part of the script, because we definitely want the Quick Action Label translations included!

@lparrott Agreed. I just won't push up those script changes. Those were the only changes the script made for these labels so this is ready for a final review to push into the parent localization branch

daniel-fuller (approved):

force2b:
I suspect removing that is fine. There was a metadata bug at one point where QuickAction labels were not translatable, so that may have been why it was stripping them out.

daniel-fuller:
@lparrott I added https://github.com/SalesforceFoundation/NPSP/pull/7070/commits/44abca055d31fc861dd6b331bd92b1a1fa8c944d to remove the `xpath` syntax that was stripping the quick actions. This should be ready for a final approval before we merge into the parent branch

lparrott (approved):

PR #7075, opened 2022-08-06
Use without sharing when querying protected payment services configuration object

daniel-fuller:
Now that the payment services config protected object is also being queried for community users who use the new constituent experience product, we found that sharing is being enforced when querying the protected payment services config records, causing the Elevate iframe not to display in the community portal. We are changing to without sharing instead to avoid sharing settings enforcement when querying this object.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/PaymentServicesConfigurationSelector.cls
+++ b/force-app/main/default/classes/PaymentServicesConfigurationSelector.cls
@@ -34,7 +34,7 @@
 * @description Selector class for Payment Services Configuration
 */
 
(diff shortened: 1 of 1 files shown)


daniel-fuller:
this is a duplicate of https://github.com/SalesforceFoundation/NPSP/pull/7073 which was already approved

daniel-fuller:
@lparrott I've tested this one and it's ready to merge. Just need one approving review on this duplicate PR.

lparrott (approved):

PR #7084, opened 2022-08-20
Support Adding Recurring Donations to an Elevate Batch in Batch Gift Entry

daniel-fuller:
# Critical Changes

# Changes
- Users can now add Recurring Donations to an Elevate batch in Batch Gift Entry.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


daniel-fuller:
Wanted to provide some explanation for renaming `ElevateAuthorizedGift` to `ElevateBatchItem`. `ElevateAuthorizedGift` was originally named as such because it contains authorization information for an Elevate gift that has been successfully tokenized and authorized.

This naming scheme is only relevant for **one-time credit card gifts* because only this gift item type is authorized upon adding the item to the Elevate batch. One-time ACH gifts, as well as Credit card or ACH commitments, are **NOT** authorized upon adding them to the batch. 

Since it is only the authorization info such as the `authorizedAt` property and one or two others that would differentiate an authorized gift from one that isn't, I don't believe it warrants creating a parent type that the authorized type inherits. I think we can achieve the most clarity if we move closer to the naming conventions used by the Elevate API since this object is meant to be used as a transitional non-domain object created through NPSP's interaction with Elevate.

Therefore, moving forward, ANY gift item added to an Elevate batch can simply be referred to as an "Elevate Batch Item" that will either have authorization info populated or not.

All feedback is welcome!

PR #7086, opened 2022-08-22
Update Next Recurring Donation Installment Amount

lparrott:
[W-11556857](https://gus.lightning.force.com/a07EE000013OCfXYAW)
[W-11556798](https://gus.lightning.force.com/a07EE000013O0YPYA0)
- We've added back-end methods and a UI to update the Amount of the next Installment of a Recurring Donation 
- Test steps can be found in the [Design Doc](https://salesforce.quip.com/5g9hA7TLw3yS#temp:C:aVE0204ea5524d74a7fa2581cbc9)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -159,6 +159,15 @@ public inherited sharing class RD2_RecurringDonation {
         return this;
     }
 
+    /**
+    * @description Returns True if the Next Payment Date has changed on an Elevate connected RD
+    * @return Boolean True if the Next Payment Date has changed on an Elevate connected RD
+    */
+    public Boolean needsNextPaymentScheduleCheck() {
+        return changedFields.contains(npe03__Recurring_Donation__c.npe03__Next_Payment_Date__c)
```
lparrott, line 167: Since RD2 disables the CRLP for this field, I think this is a safe way to "know" that a Payment has been Paid?

force-app/main/default/classes/PS_CommitmentRequest.cls
```diff
@@ -267,8 +267,11 @@ public inherited sharing class PS_CommitmentRequest {
             : UserInfo.getDefaultCurrency();
 
         if (isElevateScheduleFieldsChanged(rd, oldRd)) {
+            // If the schedule has a Next Payment Amount, we need to send it
+            List<RecurringDonationSchedule__c> rdSchedules = scheduleService.getExistingSchedules(rd);
+            Decimal nextPaymentAmount = rdSchedules.size() > 0 ? rdSchedules[0].NextPaymentAmount__c : null;
```
npsp-reedestockton, line 272: I'd be happier if we didn't have any implementation details in this class.  May getNextPaymentAmount in RD2_ScheduleService??

Also, is it safe to assume the amount is on the first active schedule? Will it be on both schedules for a 1st and 15th cadence?

force-app/main/default/classes/RD2_ScheduleService.cls
```diff
@@ -95,6 +106,9 @@ public without sharing class RD2_ScheduleService {
         schedule.Campaign__c = rd.npe03__Recurring_Donation_Campaign__c;
 
         schedule.InstallmentAmount__c = rd.npe03__Amount__c;
+        if(nextPaymentAmount != null) {
```
npsp-reedestockton, line 109: I think we end up with nextPaymentAmount on both schedules for a 1st and 15th cadence. Is that what we want?

force-app/main/default/classes/RD2_ScheduleService.cls
```diff
@@ -187,6 +203,7 @@ public without sharing class RD2_ScheduleService {
         List<RecurringDonationSchedule__c> newSchedules = buildNewSchedules(rd);
         for (RecurringDonationSchedule__c newSchedule : newSchedules) {
             newSchedule.StartDate__c = newStartDate;
+            newSchedule.NextPaymentAmount__c = nextPaymentAmount;
```
npsp-reedestockton, line 206: What's the reason for putting nextPaymentAmount on more than one schedule record?

force-app/main/default/classes/PS_CommitmentRequest.cls
lparrott: I didn't realize 1st and 15th used 2 Active schedules, I'll double check that they're in the correct sort order. I can create that method, good suggestion.

force-app/main/default/classes/RD2_ScheduleService.cls
lparrott: No reason, in my testing this was always a single schedule, but I didn't account for 1st and 15th. Are there other scenarios where there will be more than 1 new schedule created?

npsp-reedestockton: I believe that right now that's the only time we create more than one.

There can be multiple active schedules: 2 current if it's a 1st and 15th schedule + 2 future if the future schedule is a 1st and 15th + pause schedule. Currently, five active would be the cap I think, but there's no structural reason we would be prohibited from having more future schedules or more current schedules. Unlikely we'll do either of those things, but you never know.

force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -159,6 +159,15 @@ public inherited sharing class RD2_RecurringDonation {
...
+            && isElevateRecord();
```
andrewyu-salesforce, line 168: Is there a reason we aren't clearing the next payment amount for non-elevate RD?

force-app/main/default/classes/RD2_EntryFormController.cls
```diff
@@ -386,6 +386,22 @@ public with sharing class RD2_EntryFormController {
             return response;
         }
 
+        /**
+        * @description Handles validation and Elevate recurring commitment creation
+        * @param rd Recurring Donation
+        * @param nextPaymentAmount The Next Payment Amount to set on the Schedule
+        * @return UTIL_Http.Response Payments API response
+        */
+        public UTIL_Http.Response handleNextPaymentAmount(npe03__Recurring_Donation__c rd, Decimal nextPaymentAmount) {
```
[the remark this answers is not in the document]
lparrott: I was thinking I'd use the same Class since it has everything we need. Technically it's a type of RD Entry?

force-app/main/default/classes/RD2_OpportunityEvaluationService.cls
```diff
@@ -445,16 +447,27 @@ public inherited sharing class RD2_OpportunityEvaluationService {
             }
             else {
                 if (isNewOpportunityCandidate(rdRecord)) {
+                    // Prevent the Next Payment Amount from being used in later steps of this process
+                    rd.RecurringDonationSchedules__r[0].NextPaymentAmount__c = null;
```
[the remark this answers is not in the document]
lparrott: This is actually where I'm clearing NextPaymentAmount for non-Elevate RDs. My assumption is that if `isNewOpportunityCandidate` is true, a Payment has been made, which should have used the NextPaymentAmount, meaning we can now clear it.

force-app/main/default/classes/RD2_RecurringDonation.cls
lparrott: Since `isNewOpportunityCandidate` won't happen for Elevate RDs, I needed a different way to clear them. I'm unsure if this will work though, still need to do E2E testing.

force-app/main/default/classes/RD2_ScheduleService.cls
lparrott: I'm refactoring this a bit, but I'm still going to set NextPaymentAmount on all new Schedules. It shouldn't hurt anything if we clear them all after that Payment is paid.

force-app/main/default/classes/RD2_ScheduleService.cls
lparrott: On second thought, this shouldn't hurt anything as long as we clear both of them after the one-time payment is paid. Makes the logic a bit cleaner, no need to figure out which schedule to set 🤷

andrewyu-salesforce (approved):
Looks good

npsp-reedestockton (approved):
Looks good!

andrewyu-salesforce (commented):
After some investigation, the only thing we need to check is when `npe03__Next_Payment_Date__c` is changed regardless if it's Elevate record or not. The changed in field indicate the next installment is going to be created, which means it's also the time to reset the value.

When `NextPaymentAmount__c ` is modified from the UI, the UI can call `RD2_QueueableService.enqueueOppEvalService(rdId, true);` to force a schedule update and opportunity recalculation.

andrewyu-salesforce:
QE pass

PR #7089, opened 2022-08-25
Gift Entry Address issue of delayed spinner loading for process batch.

balsamhindi:
[W-11639879]( https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000014p6CvYAI/view)
We fixed an issue in which there was a delay in the **Processing Batch** spinner after selecting the **Process Batch** 
# Critical Changes

# Changes
- We've removed the **Processing Batch** spinner delay that was experienced by some customers when processing large batches.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/lwc/geBatchGiftEntryHeader/geBatchGiftEntryHeader.js
```diff
@@ -42,6 +42,7 @@ export default class GeBatchGiftEntryHeader extends LightningElement {
             case this.ACTIONS.PROCESS_BATCH:
                 this.dispatchEvent(new CustomEvent('processbatch'));
                 this.isGiftBatchProcessing = true;
+                this._isBatchProcessing = true;
```
balsamhindi, line 45: So i am not sure if this is the best place to place this but it basically allows for the spinner to be triggered immediately after a click event occurs with the Process Batch button

force-app/main/default/lwc/geBatchGiftEntryHeader/__tests__/geBatchGiftEntryHeader.test.js
```diff
@@ -142,6 +142,21 @@ describe('c-ge-batch-gift-entry-header', () => {
             expect(buttons[2].disabled).toBe(true);
         });
 
+        it('should load process batch spinner when process batch button is clicked', async () => {
+            const batchHeader = setupComponentWithDummy({});
+            const handler = jest.fn();
+
+            const buttons = batchHeader.shadowRoot.querySelectorAll('lightning-button');
```
balsamhindi, line 149: Also not super sure if this test is even necessary but added it just in case.

force-app/main/default/lwc/geBatchGiftEntryHeader/geBatchGiftEntryHeader.js
voduyemi: I believe adding this boolean here has no effect on the spinner that displays when the batch starts processing, the issue we are trying to solve here is we make an async call to the server and wait for a response before we display a spinner, the spinner is triggered in the `geGiftEntryFormApp`. To fix this, I think we need to add a seperate boolean flag that controls the spinner and will be triggered once the 'Process Batch' button is clicked but will flip its state to `false` based on the result of processing the batch.

voduyemi (changes requested):
Hey @balsamhindi , I added a comment regarding your change, let me know if you would like to chat about it when you get back.

balsamhindi:
> Hey @balsamhindi , I added a comment regarding your change, let me know if you would like to chat about it when you get back.

This makes sense Victor! I was a bit hesitant about setting that boolean in that location, I'm sure the improvement I noticed was probably more coincidental in that case. Will implement based on your suggestion and inform you as soon as I have any questions! Thank you.

balsamhindi:
@voduyemi I have addressed the changes you discussed, I want to make sure that I placed the changes in the correct location before updating the corresponding unit test, please let me know if that is what you were mentioning or if I'm totally off, thank you!

balsamhindi:
@voduyemi/@daniel-fuller This PR should now be ready for re-review, I've added a new test as well as updated some unit tests. Please review at your convenience, thank you!

voduyemi (approved):
Thanks, @balsamhindi! This looks good to me.

daniel-fuller (approved):

PR #7092, opened 2022-08-31
Removing Elevate Pause Schedule

voduyemi:
WI: [W-11565974](https://gus.lightning.force.com/a07EE000013aBqKYAU)

# Critical Changes

# Changes
Added the ability to remove Elevate pause schedules

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_PauseForm_TEST.cls
+++ b/force-app/main/default/classes/RD2_PauseForm_TEST.cls
@@ -153,7 +153,6 @@ public with sharing class RD2_PauseForm_TEST {
 
     @IsTest
     private static void shouldSavePauseWhenRDIsElevateRecord() {
-        RD2_EnablementService_TEST.setRecurringDonations2Enabled();
         RD2_ElevateIntegrationService_TEST.enableElevateUserPermissions();
         PS_IntegrationService.setConfiguration(PS_IntegrationServiceConfig_TEST.testConfig);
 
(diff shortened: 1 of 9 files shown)


force-app/main/default/classes/RD2_PauseForm_TEST.cls
```diff
@@ -188,6 +188,43 @@ public with sharing class RD2_PauseForm_TEST {
...
+            RD2_PauseForm_CTRL.savePause(JSON.serialize(pause));
+        } catch (Exception ex) {
+            auraEx = ex;
+        }
+
+        Test.stopTest();
+
+        RecurringDonationSchedule__c schedule = [SELECT Id FROM RecurringDonationSchedule__c][0];
+        System.assertEquals(null, auraEx, 'No exception was thrown as a result of the callout');
+        System.assertNotEquals(null, schedule, 'The Commitment response should be returned');
```
lparrott, line 225: We aren't actually checking if removing the Pause worked here. Is it possible to do a small integration test where the Schedule itself gets updated? I'd also like to test the new Callout and Response if possible, though I know that happens within `savePause`.

lparrott (commented):
Initial hands-on testing looks good, so my only concern is that we aren't testing the new API call logic in Unit tests (branching based on Pause dates being null or not)

npsp-reedestockton (approved):
Looks great!

lparrott (approved):

PR #7097, opened 2022-09-12
Ensure that a Gift Entry User can Successfully Process a Mix of Non-Elevate, One-time, and Recurring Elevate Donations in Batch Gift Entry

daniel-fuller:
[W-11562549](https://gus.lightning.force.com/a07EE000013UxmoYAC)

This work is to allow non-Elevate, one-time Elevate, and recurring Elevate donations to all be processed in a single data import batch in NPSP.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


daniel-fuller:
I looked through the test classes and don't see that we have introduced any new scenario for which we don't already have automation coverage. In particular, `GiftEntryProcessorQueue_TEST.shouldCreateRecurringDonationWhenGiftBatchHasRecurringGift()` is testing that a recurring donation can be successfully processed by that queueable class which would be the side of this work that doesn't involve callouts. Also, the callout response we receive from `capture` only contains status and batch id which we already have coverage for and have not introduced any new behavior there.

voduyemi (approved):
Looks good to me!

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -988,23 +990,23 @@ export default class GeFormRenderer extends LightningElement{
     }
 
     populateFormStateWithRDInfo(elevateBatchItem) {
-        const isSuccessful = elevateBatchItem.status === 'Active';
+        const isSuccessful = elevateBatchItem.status === COMMITMENT_INACTIVE_STATUS &&
```
lparrott, line 993: I just want to confirm that this will be the response even when the RD is processed right away. I assume it becomes 'Active' after the first payment is paid, which will happen async.

lparrott (approved):

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
daniel-fuller: yeah the "Active" status will be set by BDI

PR #7101, opened 2022-09-19
Elevate Partial Refund

andrewyu-salesforce:
[W-11563674](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000013X8jAYAS/view)
# Critical Changes

# Changes
- Partial Refunds can now be issued directly from NPSP.
# Issues Closed

# Community Ideas Delivered

-  We've allowed Partial Refunds to be initiated from NPSP.
# New Metadata

# Deleted Metadata


force-app/main/default/classes/PMT_ValidationService.cls
```diff
@@ -148,6 +148,10 @@ public inherited sharing class PMT_ValidationService {
     }
 
     private void validateRefund(npe01__OppPayment__c payment) {
+        if (String.isNotBlank(payment.Elevate_Payment_ID__c) && config.isIntegrationEnabled()) {
```
npsp-reedestockton, line 151: What happens here if Elevate_Payment_ID__c, but something is currently wrong with the Elevate config. Is the transaction permanently messed up?

force-app/main/default/classes/PMT_RefundController_TEST.cls
```diff
@@ -48,6 +48,36 @@ public with sharing class PMT_RefundController_TEST {
         }
     }
 
+    @IsTest
+    private static void verifyCalloutIsMadToRetrieveElevateRefundInformation() {
```
npsp-reedestockton, line 52: should add "e" to "Mad" - should be "verifyCalloutIsMadeToRetrieveElevateRefundInformation"

npsp-reedestockton (commented):
Looks good. A few small comments.

lparrott (approved):
Looks good!

force-app/main/default/classes/PMT_RefundController.cls
```diff
@@ -57,8 +57,12 @@ public with sharing class PMT_RefundController {
         refundService.withOriginalPaymentIds(new Set<Id>{paymentId});
         PMT_RefundService.RefundInfo refundInfo = refundService.getRefundInfoFor(paymentId);
 
-        refundView.paymentDate = refundInfo.originalPayment.npe01__Payment_Date__c;
-        refundView.remainingBalance = refundInfo.remainingBalance;
+        if (isElevatePayment(refundInfo.originalPayment.Elevate_Payment_ID__c)) {
```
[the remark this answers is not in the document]
andrewyu-salesforce: `refundService.getRefundInfoFor(paymentId);` did the query to retrieve the data. It will always query the record and assign to refundInfo. However, it's still good to have null check in place just in case. The safe navigation operator is added.

force-app/main/default/classes/PMT_ValidationService.cls
andrewyu-salesforce: The reason the validation for Elevate is removed is because of the following:
- Elevate refund will always be populated by the System, it will never be user input.
- We almost ran into dependency issues in the 240 release because the validation service is checking extra fields that the Connector package does not know before the 240 release and we cannot guarantee the update timing for both packages.

npsp-reedestockton (approved):
Looks great, @andrewyu-salesforce!

PR #7119, opened 2022-10-06
Refund: Convert Allocation and Support Trigger Action

andrewyu-salesforce:
[WI-11563652](https://gus.lightning.force.com/lightning/r/a07EE000013WzF5YAK/view)
# Critical Changes

# Changes
- Fixed amount opportunity Allocations will be converted to percentages when a partial refund is initiated
- Support Opportunity Amount and Allocations calculation when the refund is created from the trigger

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/PMT_Payment_TDTM.cls
+++ b/force-app/main/default/classes/PMT_Payment_TDTM.cls
@@ -100,10 +100,6 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
 
         validatePayments(newList, oldList, triggerAction);
 
-        if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-            handleAllocationsAndOpportunities(newList, oldList);
-        }
-
         if (UTIL_Currency.getInstance().isMultiCurrencyOrganization()) {
             return handleMultiCurrencyPayments(newlist, oldlist, triggerAction, dmlWrapper);
         }
@@ -317,41 +313,6 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
     ) {  
         PMT_ValidationService validationService = new PMT_ValidationService(newlist, oldList, triggerAction);
         List<ErrorRecord> errorRecords = validationService.validate().getErrors();
-
-        for (ErrorRecord error : errorRecords) {
-            if (error.hasError()) {
-                error.getRecord().addError(error.getFirstError());
-            }
-        }
-    }
-
-    private void handleAllocationsAndOpportunities(
-        List<npe01__OppPayment__c> newList,
-        List<npe01__OppPayment__c> oldList
-    ) {
-        List<npe01__OppPayment__c> refunds = new List<npe01__OppPayment__c>();
-        List<npe01__OppPayment__c> oldRefunds = new List<npe01__OppPayment__c>();
-        for (Integer i = 0; i < newList.size(); i++) {
-            if (newList[i].npe01__Payment_Amount__c < 0 && String.isBlank(newList[i].Elevate_Payment_ID__c)) {
-                refunds.add(newList[i]);
-                if ( oldList != null) {
-                    oldRefunds.add(oldList[i]);
-                }
-            }
-        }
-
-        if (refunds.isEmpty()) {
-            return;
-        }
-
-        PMT_RefundService refundService = new PMT_RefundService()
-            .withRefundRecords(refunds)
-            .withOldRefundRecords(oldRefunds);
-
-        refundService.adjustAllocationsAndOpportunities()
-            .updateAllocationsAndOpportunities();
-
-        List<ErrorRecord> errorRecords = refundService.getErrors();
 
         for (ErrorRecord error : errorRecords) {
             if (error.hasError()) {
--- a/force-app/main/default/classes/PMT_Payment_TDTM.cls
+++ b/force-app/main/default/classes/PMT_Payment_TDTM.cls
@@ -100,13 +100,14 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
 
         validatePayments(newList, oldList, triggerAction);
 
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            handleAllocationsAndOpportunities(newList, oldList);
+        }
+
         if (UTIL_Currency.getInstance().isMultiCurrencyOrganization()) {
             return handleMultiCurrencyPayments(newlist, oldlist, triggerAction, dmlWrapper);
         }
 
-        if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
-            handleAllocationsAndOpportunities(newList, oldList);
-        }
         return dmlWrapper;
     }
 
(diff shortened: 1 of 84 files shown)


force-app/main/default/classes/PMT_Payment_TDTM.cls
```diff
@@ -321,4 +323,39 @@ public class PMT_Payment_TDTM extends TDTM_Runnable {
...
     }
+
+    private void handleAllocationsAndOpportunities(
+        List<npe01__OppPayment__c> newList,
+        List<npe01__OppPayment__c> oldList
+    ) {
+        List<npe01__OppPayment__c> refunds = new List<npe01__OppPayment__c>();
+        List<npe01__OppPayment__c> oldRefunds = new List<npe01__OppPayment__c>();
+        for (Integer i = 0; i < newList.size(); i++) {
+            if (newList[i].npe01__Payment_Amount__c < 0 && String.isNotBlank(newList[i].Elevate_Payment_ID__c)) {
```
andrewyu-salesforce, line 334: Exclude Elevate payment because we don't want the opp to be updated twice if the Connector package version is outdated.

force-app/main/default/classes/PMT_RefundService.cls
```diff
@@ -190,20 +224,103 @@ public inherited sharing class PMT_RefundService {
         return response;
     }
 
-    public void processDML() {
-        Savepoint sp = Database.setSavepoint();
-        try {
-            if (!this.refundRecords.isEmpty()) {
-                insert this.refundRecords;
+    public npe01__OppPayment__c buildRefundRecord(npe01__OppPayment__c originalPayment, Decimal refundAmount) {
+        refundAmount = (refundAmount == null) ? originalPayment.npe01__Payment_Amount__c : refundAmount;
```
lparrott, line 228: I was going to propose a second method call for full refunds that doesn't take a refundAmount, but I think a description on this method should be enough. Having `null` represent a full refund just doesn't feel great to me.

force-app/main/default/classes/PMT_RefundService.cls
```diff
@@ -190,20 +224,103 @@ public inherited sharing class PMT_RefundService {
...
+                ? PMT_RefundService.FULL_REFUND
+                : PMT_RefundService.PARTIAL_REFUND
+        );
+        
+        if (isMultiCurrencyEnabled) {
+            refund.put(UTIL_Currency.CURRENCY_ISO_CODE_FIELD, originalPayment.get(UTIL_Currency.CURRENCY_ISO_CODE_FIELD));
+        }
+
+        if (this.refundRecords == null) {
+            this.refundRecords = new List<npe01__OppPayment__c>();
```
lparrott, line 243: I don't love this pattern, but if this is the only place were initializing the List is needed, this should be fine.

force-app/main/default/classes/PMT_RefundService.cls
```diff
@@ -218,57 +335,78 @@ public inherited sharing class PMT_RefundService {
...
+                allocation.Percent__c = calculatedPercentage.setScale(2, RoundingMode.HALF_UP);
+                toUpdate.add(allocation);
+            }
+
+            percentage += allocation.Percent__c;
         }
 
-        opportunity.Amount = opportunity.Amount - refundAmount;
+        if (percentage > 100) {
+            toUpdate[0].Percent__c = toUpdate[0].Percent__c - (percentage - 100);
```
lparrott, line 406: When would this happen, due to rounding? A comment would be helpful, this feels like an odd situation.

lparrott (commented):
First pass looks good, just have a few comments. Allocation work can be a nightmare, I'm glad you came up with an approach for this!

force-app/main/default/classes/PMT_RefundController.cls
```diff
@@ -134,16 +134,14 @@ public with sharing class PMT_RefundController {
             return;
         }
 
-        refundService.updateParentOpportunity(originalPayment, refundAmount);
```
lparrott, line 137: I just want to be sure: this Opportunity logic is now being run by the trigger that's called after `refundService.insertRefunds();`?

force-app/main/default/classes/PMT_RefundService.cls
```diff
@@ -218,57 +335,78 @@ public inherited sharing class PMT_RefundService {
...
-            );
+
+        List<Allocation__c> toUpdate = new List<Allocation__c>();
+
+        Decimal oppAmount = allocations[0].Opportunity__r.Amount;
+        Double percentage = 0.0;
+        for (Allocation__c allocation : allocations) {
+            if (allocation.Percent__c == null) {
+                Decimal calculatedPercentage = allocation.Amount__c / oppAmount * 100;
+                allocation.Percent__c = calculatedPercentage.setScale(2, RoundingMode.HALF_UP);
```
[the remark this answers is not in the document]
andrewyu-salesforce: there is no specific reason, I put a rounding there because it's easier to read. I can remove it and make the it goes all the way to 6th decimal place.

force-app/main/default/classes/PMT_RefundService.cls
andrewyu-salesforce: Correct, this is to handle the edge case of an edge case. If the allocations have some extreme weird setup and we don't want it to be off by 0.01%. I will add the comment in the address it

force-app/main/default/classes/PMT_RefundService.cls
andrewyu-salesforce: Initialize the list at class level now

force-app/main/default/classes/PMT_RefundService.cls
andrewyu-salesforce: I will remove the `null` check here. In the Controller, `refundAmount` will always have a value. This was a leftover from the previous WI

force-app/main/default/classes/PMT_RefundService.cls
```diff
@@ -218,57 +335,78 @@ public inherited sharing class PMT_RefundService {
...
 
-        if (this.refundRecords == null) {
-            this.refundRecords = new List<npe01__OppPayment__c>();
+        for (Integer i = 0; i < this.refundRecords.size(); i++) {
+            Opportunity parentOpp = this.opportunityMap.get(this.refundRecords[i].npe01__Opportunity__c);
+            parentOpp.Amount = (parentOpp.Amount == null) ? 0 : parentOpp.Amount;
+            parentOpp.Amount += this.refundRecords[i].npe01__Payment_Amount__c;
+
+            if (this.oldRefundRecords != null && !this.oldRefundRecords.isEmpty()) {
+                parentOpp.Amount -= this.oldRefundRecords[i].npe01__Payment_Amount__c;
```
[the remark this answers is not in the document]
andrewyu-salesforce: Like the helper method approach.

lparrott (approved):
Thanks for the quick turnaround on implementing feedback!

force-app/main/default/classes/PMT_Payment_TDTM.cls
daniel-fuller, line 334: @andrewyu-salesforce since we have `String.isNoBlank(Elevate_Payment_ID__c)` here, wouldn't this mean that the only refunds included in the following logic would be Elevate refunds? Thought we wanted this logic to only run for non-Elevate refunds for this work?

daniel-fuller (changes requested):
Looks good! Just had one question I added.

force-app/main/default/classes/PMT_Payment_TDTM.cls
andrewyu-salesforce: good catch!!

daniel-fuller (approved):

daniel-fuller:
we have JEST failure on this PR but it is not related to this work so we are merging and will address the JEST failure in a separate work stream

PR #7126, opened 2022-10-13
Clean Up Remaining POU Links in NPSP Repo

balsamhindi:
[W-11926218](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000019AMt2YAG/view) Clean up remaining POU NPSP References.
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/unpackaged/config/trial_tso/prompts/NPSPContactMergeGuidance.prompt
+++ b/unpackaged/config/trial_tso/prompts/NPSPContactMergeGuidance.prompt
@@ -3,7 +3,7 @@
     <masterLabel>NPSP Contact Merge Guidance</masterLabel>
     <promptVersions>
         <actionButtonLabel>Find and Merge Duplicates</actionButtonLabel>
-        <actionButtonLink>https://help.salesforce.com/s/articleView?id=sf.contacts_merge.htm&amp;type=5</actionButtonLink>
+        <actionButtonLink>https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&amp;type=5</actionButtonLink>
         <body>&lt;p&gt;Keeping your data clean and reliable is one of the most important ways to instill confidence in your system. &lt;/p&gt;&lt;p&gt;&lt;br&gt;&lt;/p&gt;&lt;p&gt;Use NPSP Duplicate rules to identify duplicate contacts. Then use NPSP Contact Merge to merge duplicates into a single record without losing data from any related households, activities or opportunities. &lt;/p&gt;</body>
         <customApplication>Nonprofit_Success_Pack</customApplication>
         <delayDays>1</delayDays>
--- a/unpackaged/config/trial_tso/prompts/NPSPContactMergeGuidance.prompt
+++ b/unpackaged/config/trial_tso/prompts/NPSPContactMergeGuidance.prompt
@@ -3,7 +3,7 @@
     <masterLabel>NPSP Contact Merge Guidance</masterLabel>
     <promptVersions>
         <actionButtonLabel>Find and Merge Duplicates</actionButtonLabel>
-        <actionButtonLink>https://powerofus.force.com/s/article/NPSP-Merging-Contacts</actionButtonLink>
+        <actionButtonLink>https://help.salesforce.com/s/articleView?id=sf.contacts_merge.htm&amp;type=5</actionButtonLink>
         <body>&lt;p&gt;Keeping your data clean and reliable is one of the most important ways to instill confidence in your system. &lt;/p&gt;&lt;p&gt;&lt;br&gt;&lt;/p&gt;&lt;p&gt;Use NPSP Duplicate rules to identify duplicate contacts. Then use NPSP Contact Merge to merge duplicates into a single record without losing data from any related households, activities or opportunities. &lt;/p&gt;</body>
         <customApplication>Nonprofit_Success_Pack</customApplication>
         <delayDays>1</delayDays>
--- a/unpackaged/config/trial_tso/prompts/ReportGuidance.prompt
+++ b/unpackaged/config/trial_tso/prompts/ReportGuidance.prompt
@@ -4,7 +4,7 @@
     <promptVersions>
         <actionButtonLabel>Learn About Reports</actionButtonLabel>
         <actionButtonLink>https://sfdc.co/QVnrG</actionButtonLink>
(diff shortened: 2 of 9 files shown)


balsamhindi:
We are back at it again! After discussion with @ErinWiedemer and observing that there were some powerofus links still left in the NPSP code, we decided to just get address everything in the NPSP repo and get it all out of the way now regardless of what the audit sheet specifies, feedback and general observations are much appreciated, thank you!

unpackaged/config/trial_tso/prompts/NPSPContactMergeGuidance.prompt
```diff
@@ -3,7 +3,7 @@
     <masterLabel>NPSP Contact Merge Guidance</masterLabel>
     <promptVersions>
         <actionButtonLabel>Find and Merge Duplicates</actionButtonLabel>
-        <actionButtonLink>https://powerofus.force.com/s/article/NPSP-Merging-Contacts</actionButtonLink>
+        <actionButtonLink>https://help.salesforce.com/s/articleView?id=sf.contacts_merge.htm&amp;type=5</actionButtonLink>
```
balsamhindi, line 6: I replaced this based on assumptions, I think a large part of this button is based on the link so I wasn't super sure if we wanted to remove it entirely, but can remove if seen as appropriate!

lparrott: This link is more specific to NPSP: https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&type=5

balsamhindi: Thank you for pointing that out, have made an update 👍

ErinWiedemer (approved):
Hi Balsam,
I checked the links in all of these files. They are all good, though I am not sure we need all of the extra code at the end of them. For example, in https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Automated_Soft_Credits.htm**&amp;type=5&quot;** I don't know if we need this or not: **&amp;type=5&quot;** Especially because the URL uses **&type=5** and not **&amp;type=5**. (Full URL is https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&type=5 though https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm works too. This is pure ignorance on my part, so it's probably not a problem. I just want to make sure the replacement URLs work :)

balsamhindi:
> Hi Balsam, I checked the links in all of these files. They are all good, though I am not sure we need all of the extra code at the end of them. For example, in https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Automated_Soft_Credits.htm**&amp;type=5"** I don't know if we need this or not: the bold portion or not. Especially because the URL uses **&type=5** and not **& amp; type=5**. (Full URL is https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&type=5 though https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm works too. This is pure ignorance on my part, so it's probably not a problem. I just want to make sure the replacement URLs work :)

@ErinWiedemer No pure ignorance at all, definitely a good question.  To answer the initial question, putting characters like `&` , `<`, or `>` in html by themselves will break the url as those are special characters that are already used for something else in html. So instead an `amp;` needs to be appended to those characters in order to escape them and allow them to be read properly as part of the string. So something like "https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&amp;type=5" actually ends up being displayed as "https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Merge_Contacts.htm&type=5". For the other part, yes I can definitely remove the `&type=5` if that is seen as preferred, I actually originally had them as `.htm` without the query string `&type=5` but I wasn't sure if that was significant as that's what the URL gets redirected to, whatever you prefer I can make the change 👍

ErinWiedemer:
@balsamhindi I definitely don't have a preference. I think it's fine to leave the `&type=5`.  Thanks for explaining the nomenclature. I wasn't sure about the use of the ampersand and "amp". So thanks for reassuring me that it works.

unpackaged/config/trial_tso/prompts/ReportGuidance.prompt
```diff
@@ -4,7 +4,7 @@
     <promptVersions>
         <actionButtonLabel>Learn About Reports</actionButtonLabel>
         <actionButtonLink>https://sfdc.co/QVnrG</actionButtonLink>
-        <body>&lt;p&gt;Bring your data to life by leveraging Reports. NPSP comes with a number of reports you can take advantage of immediately. &lt;/p&gt;&lt;p&gt;&lt;br&gt;&lt;/p&gt;&lt;p&gt;To get started, select ‘All Reports’ from the menu on the left to see which reports are available to you. Learn more about these specific reports in the &lt;a href=&quot;https://powerofus.force.com/s/article/NPSP-Reports-Workbook&quot; rel=&quot;noopener noreferrer&quot; target=&quot;_blank&quot;&gt;Power of Us Hub&lt;/a&gt;. &lt;/p&gt;</body>
+        <body>&lt;p&gt;Bring your data to life by leveraging Reports. NPSP comes with a number of reports you can take advantage of immediately. &lt;/p&gt;&lt;p&gt;&lt;br&gt;&lt;/p&gt;&lt;p&gt;To get started, select ‘All Reports’ from the menu on the left to see which reports are available to you.</body>
```
lparrott, line 7: Couldn't we link here for this one, or is this out of date and not worth using? https://help.salesforce.com/s/articleView?id=sfdo.NPSP_Reports_Workbook.htm&type=5

balsamhindi: That looks quite similar, but would wanna make sure to defer to @ErinWiedemer on this!

ErinWiedemer: HI @balsamhindi, I'm a little torn on this one.
I was just in a meeting yesterday with a communications writer about this Reports Workbook. It is out of date, but she's going to completely re-write it. So I think it is okay to add Luke's link. However, we're trying to get away from posting links in the product, so that if the links change, you don't have to go through all of this work again. 

Let's stick with the original plan and not include it. I think that we're going to develop a trail on NPSP reports, and that will be a lot more helpful than this link anyway.

lparrott (approved):

PR #7128, opened 2022-10-14
Fix for loading Elevate Widget in Read-Only mode for ACH gifts

lparrott:
- [W-11906042](https://gus.lightning.force.com/a07EE000018k7nEYAQ)
- Editing a previously entered ACH gift will now load the Elevate widget in read-only mode.
- When editing the ACH Account and Routing numbers, the option to Cancel the change will now be displayed.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -202,7 +208,6 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
     get showCancelButton() {
         return this.isEdit 
             && !this.isExpiredTransaction
-            && !this.isPaymentStatusAuthorized()
```
lparrott, line 205: I don't see a reason to hide the Cancel button if an unprocessed gift has this Status... shouldn't we allow customers to undo the activation of the widget?

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -1098,13 +1102,13 @@ export default class GeFormRenderer extends LightningElement{
...
 
+    isGiftPending() {
+        return this.formState[apiNameFor(PAYMENT_STATUS)] === this.PAYMENT_TRANSACTION_STATUS_ENUM.PENDING;
+    }
+
     isGiftExpired() {
         return this.formState[apiNameFor(PAYMENT_STATUS)] === this.PAYMENT_TRANSACTION_STATUS_ENUM.EXPIRED;
     }
-    
-    shouldNotNullPaymentFields() {
```
lparrott, line 1105: This method was not being used anywhere.

force-app/main/default/lwc/geFormWidget/geFormWidget.js
```diff
@@ -40,6 +41,7 @@ export default class GeFormWidget extends LightningElement {
         apiNameFor(DATA_IMPORT_DONATION_DONOR),
         apiNameFor(DATA_IMPORT_ACCOUNT_NAME),
         apiNameFor(DATA_IMPORT_PAYMENT_STATUS),
+        apiNameFor(DATA_IMPORT_STATUS),
```
lparrott, line 44: This was done to pass this value down to the Tokenize Card component, I don't think it will have any negative impact on existing functionality 🤞

force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -488,7 +488,7 @@ export default class GeFormRenderer extends LightningElement{
     }
 
     handleNullPaymentFieldsInFormState() {
-        if (this.shouldNullPaymentRelatedFields()) {
+        if (this.shouldNullPaymentRelatedFields() || this.shouldNullFormerAchPayment()) {
```
lparrott, line 491: This change was partially a guess based on the logic I'm seeing, I just want to make sure ACH has the same guard rails in place as Credit Card!

voduyemi: Don't you think we should have the logic in one function?

force-app/main/default/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -56,6 +56,9 @@ class psElevateTokenHandler {
         const sandboxEnhancedUrl =  `https://${domainInfo.orgDomain}--${namespace}.sandbox.vf.force.com`;
         const sandboxEnhancedLightning =  `https://${domainInfo.orgDomain}--${namespace}.sandbox.lightning.force.com`;
         const sandboxEnhancedCanon =  `https://${domainInfo.orgDomain}--${namespace}.sandbox.my.salesforce.com`;
+        const scratchEnhancedUrl =  `https://${domainInfo.orgDomain}--${namespace}.scratch.vf.force.com`;
```
voduyemi, line 59: I think we can remove this change since @balsamhindi has it covered in  #7131

force-app/main/default/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
voduyemi: The 'Don't Enter Payment Information' button and the payment method Combobox are what deactivates the widget

voduyemi (changes requested):
Hey @lparrott, I left some initial comments, and might have more once I start testing. Also, we have failing jest tests.

force-app/main/default/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
lparrott: Right, but I got into a situation where I could click "Edit Payment Information", and then had no way to cancel that change if I decided to keep the Payment Information the same.

force-app/main/default/lwc/psElevateTokenHandler/psElevateTokenHandler.js
lparrott: Sure, I needed it for my testing, but didn't need to commit it.

force-app/main/default/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
```diff
@@ -361,11 +382,24 @@ export default class geFormWidgetTokenizeCard extends LightningElement {
         this.display.transitionTo('readOnly');
     }
 
+    isDataImportStatusImported() {
+        return this._widgetDataFromState[apiNameFor(DATA_IMPORT_STATUS)] === GIFT_STATUSES.IMPORTED
+    }
+
+    isPaymentStatusInReadOnlyMode() {
+        return this.isPaymentStatusAuthorized() || this.isPaymentStatusPending();
```
daniel-fuller, line 390: It's a little confusing what we're trying to do with this function. If we are saying that AUTHORIZED and PEDNING are the two payment statuses that would render the widget in read-only mode, could we rename this to `isReadOnlyPaymentStatus()` ?

daniel-fuller: I see further down that we already have a `readOnlyStatuses()` function as well. So what is this function's purpose?

daniel-fuller (changes requested):

daniel-fuller:
@lparrott could we add a JEST test here for loading the ACH payment method and verifying that the widget is in readOnly mode? I assume we already have one for credit card but would be good to update that coverage now that we've added it for ACH

force-app/main/default/lwc/geFormWidgetTokenizeCard/geFormWidgetTokenizeCard.js
lparrott: Oh nice, I hadn't seen that, I think this could be replaced by a call to `hasReadOnlyStatus` because yes, the goal here is to load the widget in read-only mode for certain statuses.

daniel-fuller (approved):
looks good!

voduyemi (approved):
Looks good!

PR #7129, opened 2022-10-17
Pay first installment for offline recurring gifts

voduyemi:
WI: [W-11818484](https://gus.lightning.force.com/a07EE000017MEZSYA4)

# Critical Changes

# Changes
- When entering Recurring Donations in Gift Entry, users can indicate that the first installment is paid on the Batch Gift Entry form.
# Issues Closed

# Community Ideas Delivered

# New Metadata
Custom Label
- geRD2FirstInstallmentPaid

# Deleted Metadata


force-app/main/default/classes/GiftBatch.cls
```diff
@@ -46,6 +46,9 @@ public inherited sharing class GiftBatch implements IGiftBatch {
     @TestVisible
     private GiftTemplate giftTemplate;
 
+    private final Boolean shouldPayFirstInstallment;
+    private static final String AllowFirstInstallmentPaid = 'AllowFirstInstallment__f';
```
lparrott, line 50: From Victor: "It's almost like a fake field, stored in the batch defaults JSON string to be read in the backend to decide when a recurring gift should have it’s first installment paid.
Because I’m storing it in an existing JSON string with a defined structure, I had to follow the existing pattern"

force-app/main/default/classes/GiftBatch.cls
```diff
@@ -122,6 +130,32 @@ public inherited sharing class GiftBatch implements IGiftBatch {
...
+    }
+
+    private Boolean calculateShouldPayFirstInstallment() {
+        JSONParser parser = JSON.createParser(this.batch.Batch_Defaults__c);
+        String shouldPayFirstInstallment;
+
+        while (parser.nextToken() != null) {
+            if (parser.getCurrentToken() == JSONToken.FIELD_NAME) {
+                if (parser.getText() == AllowFirstInstallmentPaid) {
+                    parser.nextToken();
```
lparrott, line 144: Gonna need an explanation for this one, how do we know how many times to call it?

force-app/main/default/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -105,6 +109,36 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
...
+        if (!isChecked) {
+            this._allowFirstInstallment = false;
+            this._allowFirstInstallmentDisabled = true;
+        } else {
+            this._allowFirstInstallmentDisabled = false;
+        }
+
+    }
+
+    get allowFirstInstallment () {
```
lparrott, line 124: This looks intentional, but when I "Edit Batch Info" and deselect "Allow Recurring Donations", "First Installment Paid" remains checked. I see the extra logic here, but why is this different behavior from the Batch creation wizard?

force-app/main/default/classes/GiftBatch.cls
voduyemi: The string being parsed is `{"AllowFirstInstallment__f":{"objectApiName":"AllowFirstInstallment__o","fieldApiName":"AllowFirstInstallment__f","value":true}`. This structure is never going to change since we defined it in the front-end. The amount of times we call `parser.nextToken()` corresponds to the number of tokens that exist before the `value` token that holds the boolean we want.

force-app/main/default/lwc/geBatchWizard/geBatchWizard.js
voduyemi: Going to look into this, the idea here is, the `Allow Recurring Donations` checkbox controls the `First Installment Paid` checkbox, if it's unchecked, the other should be unchecked and disabled.

force-app/main/default/classes/GiftBatch.cls
lparrott: I know we try to avoid comments and let the code speak for itself, but I think this needs one, even just a line explaining the approach.

callen-sfdo: Should this have an inner loop that goes through the tokens until it hits the 'value' token.  That way its not as hardcoded to calling a specific number of times?

ErinWiedemer (approved):
LGTM, @voduyemi!

lparrott (approved):

callen-sfdo (approved):

PR #7130, opened 2022-10-17
W-11552158 - Adding 'Failing' as Recurring Donation Status

callen-sfdo:
- Added 'Failing' to the Status Picklist
- Added mapping for 'Failing' to 'Active' in the RD Status Mapper

# Critical Changes

# Changes
- We've added a status of "Failing' to the Recurring Donation status picklist, so that users can track and act upon failing Recurring Donations.
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


andrewyu-salesforce (approved):

lparrott (approved):

andrewyu-salesforce:
@callen-sfdo After testing the PR and creating the test suite, I noticed a couple of things that need to be changed:
1. Failing Status default mapping should be added to `RD2_StatusMApper_TEST.getDefaultStatusLabelByValue()` so it can be tested in the unit test
2. The Status should be added to the `rd2_post_config` file instead of where it is currently added because we don't want the field to show up in RD1.

andrewyu-salesforce (changes requested):
See comment aboved

andrewyu-salesforce (approved):

PR #7140, opened 2022-11-02
Implement fixes resulting from Lightning Web Security Compliance Spike

balsamhindi:
[W-11919343
](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE0000192sNPYAY/view)
Modified formatting for objects to be accepted by LWS requirements
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force-app/main/default/pages/GE_TokenizeCard.page
```diff
@@ -101,11 +102,11 @@
 
                 sfdo.mount(config)
                     .then(() => {
-                        const successMessage = JSON.stringify({ type: postMessageType, success: true });
+                        const successMessage = { type: postMessageType, success: true };
```
balsamhindi, line 105: @lparrott these are the changes that. were made for the credit card widget to load properly again, I've been debating writing a function that checks if an object is stringified and then parses it conditionally but I'm not sure if that is really worth it in this case, your feedback would be much appreciated here!

force-app/main/default/lwc/psElevateTokenHandler/psElevateTokenHandler.js
```diff
@@ -105,7 +105,7 @@ class psElevateTokenHandler {
                 if (typeof event.data === 'object') {
                     component.handleMessage(event.data);
                 } else {
-                    const message = JSON.parse(event.data);
+                    const message = event.data;
```
lparrott, line 108: I'm trying to follow the logic, does this change require an update to `handleMessage` in `geFormWidgetTokenizeCard.js`? That's one of the `component`s that get passed into this.

force-app/main/default/pages/GE_TokenizeCard.page
lparrott: We hopefully won't need that, as we should be consistent in how this gets called, but we may find a place that we missed during testing!

force-app/main/default/pages/GE_TokenizeCard.page
```diff
@@ -174,17 +175,17 @@
...
                  // Response contains the error, send it to lwc
-                 sendPostMessage(JSON.stringify({ type: postMessageType, ...resp }));
+                 sendPostMessage(({ type: postMessageType, ...resp }));
               }
            })
            .catch((err) => {
               // Send error to lwc
-              const errorString = JSON.stringify({ type: postMessageType, ...err });
+              const errorString = ({ type: postMessageType, ...err });
               sendPostMessage(errorString);
```
lparrott, line 189: Minor thing, but `sendPostMessage` below this should have its param description updated.

lparrott (commented):
Great approach! Just had a few comments before I start testing.

force-app/main/default/lwc/psElevateTokenHandler/psElevateTokenHandler.js
balsamhindi: That is a good point, I'm going to look into the logic that gets called there and see any changes that need to be made, the original issue appeared to have to do with how the `action` value in `event` was being unnecessarily stringified so when you tried to reference it, it was returning as `undefined` but I will look into the `geFormWidgetTokenizeCard` side as well.

balsamhindi: @lparrott So I took a loot at the `geFormWidgetTokenizeCard.js` logic, it looks like the `message` value that gets passed through there is already stringified and then gets passed through here and parsed. I would feel a bit more comfort going with this solution though if I can have ways to simulate errors-based events just to ensure those bodies are being parsed correctly. Would manually passing an error as an event be adequate in this case?

lparrott: Oh definitely, if you can test that and it comes through fine I think we're in good shape!

balsamhindi: @lparrott So I went ahead and manually passed in an error to test that it was parsing the error properly and it seemed to properly handle that. It would be a bit more comforting to be able to QA this in an error state as well but not sure how feasible that is.  I also tested ACH as an option and it appears to load fine: 
Screen Shot 2022-11-04 at 10 47 11 AM
I did notice when I selected `Paypal` this showed up:
Screen Shot 2022-11-04 at 10 49 00 AM
Unsure if that is expected behavior or a residual bug.

lparrott: Great! Causing an actual error would be tricky (unless there's a way to do that I haven't heard about), but this approach gives me confidence in the change. Paypal is not supported in Elevate, so selecting it will deactivate the widget.

lparrott (approved):
Nice work!

daniel-fuller (approved):

PR #7142, opened 2022-11-07
Pause API E2E fixes

voduyemi:
WI: [W-11575839](https://gus.lightning.force.com/a07EE000013lI2fYAE)
# Critical Changes

# Changes

- To pause elevate connected installments, we now have to supply intervals instead of installment start and end dates regardless of the number of selected installments
- The PATCH HTTP verb is now used for updating existing pauses
- Origin URLs fix for the elevate widget
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/GE_PaymentServices.cls
+++ b/force-app/main/default/classes/GE_PaymentServices.cls
@@ -153,8 +153,21 @@ public with sharing class GE_PaymentServices {
      * @return String JSON of visualforce URL
      */
     @AuraEnabled(cacheable=true)
-    public static String getVfURL(String namespace) {
-        return DomainCreator.getVisualforceHostname(namespace);
+    public static OriginUrls getOriginUrls(String namespace) {
+        return new OriginUrls(
+            DomainCreator.getVisualforceHostname(namespace),
+                DomainCreator.getLightningHostname()
+        );
+    }
+
+    public class OriginUrls {
+        @AuraEnabled public String visualForceOriginUrl;
+        @AuraEnabled public String lightningOriginUrl;
+
+        public OriginUrls(String visualForceOriginUrl, String lightningOriginUrl) {
+            this.visualForceOriginUrl = 'https://'+visualForceOriginUrl;
+            this.lightningOriginUrl = 'https://'+lightningOriginUrl;
+        }
     }
 
     /**
--- a/force-app/main/default/classes/PS_CommitmentRequest.cls
+++ b/force-app/main/default/classes/PS_CommitmentRequest.cls
@@ -703,12 +703,12 @@ public inherited sharing class PS_CommitmentRequest {
         }
 
         public RequestBody withStartTimestamp(Date startTimestamp) {
-            this.startTimestamp = getISOFormatDateTimeFor(startTimestamp);
+            this.startTimestamp = getISOFormatDateTimeFor(startTimestamp, false);
             return this;
         }
 
         public RequestBody withEndTimestamp(Date endTimestamp) {
-            this.endTimestamp = getISOFormatDateTimeFor(endTimestamp);
+            this.endTimestamp = getISOFormatDateTimeFor(endTimestamp, true);
             return this;
         }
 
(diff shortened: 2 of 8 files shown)


daniel-fuller (approved):
thanks for the work over a late night!

force-app/main/default/classes/RD2_CommitmentService.cls
```diff
@@ -47,7 +47,7 @@ public without sharing class RD2_CommitmentService {
             PS_CommitmentRequest.RequestBody requestBody = new PS_CommitmentRequest().getRequestBody(rd, oldRd, paymentMethodToken);
 
             response = sendRequest(rd.CommitmentId__c, JSON.serialize(requestBody),
-                PS_Request.ElevateEndpoint.COMMITMENT);
+                PS_Request.ElevateEndpoint.COMMITMENT, UTIL_Http.Method.POST);
```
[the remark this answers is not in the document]
andrewyu-salesforce: Reference to the above comment, this `POST` change will break the Elevate RD creation. There needs to be logic to determine if POST or PATCH is needed.

force-app/main/default/classes/PS_CommitmentRequest.cls
```diff
@@ -194,6 +194,17 @@ public inherited sharing class PS_CommitmentRequest {
             .withBody(jsonRequestBody)
             .build();
     }
+    public static HttpRequest buildRequest(String commitmentId, String jsonRequestBody,
```
andrewyu-salesforce, line 197: I think this method and the method above can be combined into 1 and move the post/patch method logic out to the R`D2_CommitmentService.handleCommitment()`. Similar to the work you did in the `handleCommitmentPause()`

andrewyu-salesforce (changes requested):

force-app/main/default/classes/RD2_CommitmentService.cls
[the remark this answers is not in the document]
voduyemi: Good catch! @andrewyu-salesforce I've made the update.

force-app/main/default/classes/PS_CommitmentRequest.cls
lparrott, line 197: The above version of `buildRequest` could call this one after determining the `method`, avoids repetition.

force-app/main/default/classes/PS_CommitmentRequest.cls
```diff
@@ -224,11 +236,14 @@ public inherited sharing class PS_CommitmentRequest {
             : getUpdateRequestBody(rd, oldRd, token);
     }
 
-    public RequestBody getPauseRequestBody(RecurringDonationSchedule__c schedule) {
-       return new RequestBody()
-           .withStartTimestamp(schedule.StartDate__c)
-           .withEndTimestamp(schedule.EndDate__c)
-           .withReason(ELEVATE_PAUSE_REASONS.get(schedule.StatusReason__c));
+    public PauseRequestBody getPauseRequestBody(RD2_ScheduleService.ElevatePauseSchedule schedule) {
```
lparrott, line 239: I really like having a separate `PauseRequestBody` 👍

force-app/main/default/classes/GE_PaymentServices.cls
```diff
@@ -153,8 +153,21 @@ public with sharing class GE_PaymentServices {
...
+            DomainCreator.getVisualforceHostname(namespace),
+                DomainCreator.getLightningHostname()
+        );
+    }
+
+    public class OriginUrls {
+        @AuraEnabled public String visualForceOriginUrl;
+        @AuraEnabled public String lightningOriginUrl;
+
+        public OriginUrls(String visualForceOriginUrl, String lightningOriginUrl) {
```
balsamhindi, line 167: This looks good to me, I like that it's an object of URLs instead of two separate functions, I honestly wish we could use this for the `buildLightningOrigin` lightningURL in `GE_tokenizecard` instead of having a separate method but I believe that can't be done if something is Auraenabled, correct? One thing to note (not a change needed to be made here of course) but if we want to return the url with the `https://` from the backend side for the sake of consistency  we should do the same for this method as well 
``` public String getLightningURL() {
        return DomainCreator.getLightningHostname();
    }
```

balsamhindi (approved):
Looks good to me, left one comment that isn't a required change but more of a note, thanks for taking care of the origin url change in here as well!

andrewyu-salesforce (approved):

PR #7143, opened 2022-11-07
Fix for colliding queueables in Gift Entry

voduyemi:
WI: [W-12013424](https://gus.lightning.force.com/a07EE00001BaWizYAF)
# Critical Changes

# Changes
This PR addresses the issue of the RD2OpportunityEval Queueable breaking the chaining mechanism of the GiftEntryProcessorQueue when recurring gifts are being processed via Gift Entry.

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_QueueableService.cls
+++ b/force-app/main/default/classes/RD2_QueueableService.cls
@@ -95,7 +95,7 @@ public class RD2_QueueableService {
 
         EvaluateInstallmentOpportunities service = new EvaluateInstallmentOpportunities(toProcess, rdIdsWhereScheduleChanged, oldRds);
 
-        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs()) {
+        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs() && canQueueNamingService()) {
             System.enqueueJob(service);
 
         } else {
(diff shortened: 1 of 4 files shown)


force-app/main/default/classes/RD2_QueueableService.cls
```diff
@@ -94,8 +94,9 @@ public class RD2_QueueableService {
         }
 
         EvaluateInstallmentOpportunities service = new EvaluateInstallmentOpportunities(toProcess, rdIdsWhereScheduleChanged, oldRds);
+        RD2_Settings settings = RD2_Settings.getInstance();
 
-        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs()) {
+        if (Limits.getQueueableJobs() < Limits.getLimitQueueableJobs() && !settings.isGiftEntryMode) {
```
lparrott, line 99: Is it worth renaming `isGiftEntryMode` to have something to do with this async logic? It looks like it's only being used to determine if we can enqueue a new job... 
Also, there's an unrelated reference to `isGiftEntryMode` in the `GE_GiftEntry` component, confused me for a second 😅

lparrott (approved):

force-app/main/default/classes/RD2_QueueableService.cls
andrewyu-salesforce, line 99: The logic on checking if it's in gift Entry mode should be moved to `RD2_RecurringDonations_TDTM` [line 403](https://github.com/SalesforceFoundation/NPSP/blob/db131092fef2093f645c2b394d9192903e469c49/force-app/main/default/classes/RD2_RecurringDonations_TDTM.cls#L403). `RD2_QueuableService` should only be used to determine if a queuable can be launched and should not have any business logic in there. RD2 TDTM has a method to determine if RDs should run in async mode.

andrewyu-salesforce (changes requested):
@voduyemi Some feedback on where the gift entry mode check should be

PR #7145, opened 2022-11-08
Change Payment Method/Other attribute value throwing error for elevate connected RD "Missing Authentication Token"

daniel-fuller:
Fix a bug causing a "Missing Authentication Token" error to be thrown when updating any schedule information for an existing Elevate-connected recurring donation

[W-11999572](https://gus.lightning.force.com/a07EE00001BMMjEYAX)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/PS_Request.cls
+++ b/force-app/main/default/classes/PS_Request.cls
@@ -274,9 +274,6 @@ public inherited sharing class PS_Request {
             } else if (endpoint == ElevateEndpoint.COMMITMENT) {
                 value = ENDPOINT_COMMITMENT;
 
-                if (String.isNotBlank(commitmentId)) {
-                    value += '/' + commitmentId;
-                }
             } else if (endpoint == ElevateEndpoint.COMMITMENT_PAUSE) {
                 value = String.format(
                     ENDPOINT_COMMITMENT_PAUSE,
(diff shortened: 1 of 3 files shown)


daniel-fuller:
The code block causing this bug was introduced in https://github.com/SalesforceFoundation/NPSP/commit/86a9f7fdf72d2e7871ffed48364fdf9f91bbb786 where we explicitly added the commitment id to the endpoint URL for commitments if it existed. @npsp-reedestockton I saw you introduced that change, and my assumption is that we previously needed the commitment id in the URL and the endpoint changed. Want to make sure there's nothing else we're missing here.

voduyemi (approved):

npsp-reedestockton:
> The code block causing this bug was introduced in [86a9f7f](https://github.com/SalesforceFoundation/NPSP/commit/86a9f7fdf72d2e7871ffed48364fdf9f91bbb786) where we explicitly added the commitment id to the endpoint URL for commitments if it existed. @npsp-reedestockton I saw you introduced that change, and my assumption is that we previously needed the commitment id in the URL and the endpoint changed. Want to make sure there's nothing else we're missing here.

@daniel-fuller This was _very_ early in our Elevate work. Cancel was one of the first things we did and that's when this was introduced. We weren't doing any patch calls at that point. I don't know how this will impact code added after that point, but you can verifty the reason for the addition is no longer valid by testing cancel.

force-app/main/default/classes/PS_Request.cls
```diff
@@ -274,9 +274,6 @@ public inherited sharing class PS_Request {
             } else if (endpoint == ElevateEndpoint.COMMITMENT) {
                 value = ENDPOINT_COMMITMENT;
 
-                if (String.isNotBlank(commitmentId)) {
-                    value += '/' + commitmentId;
```
lparrott, line 278: This **should** be safe, since COMMITMENT_CANCEL and COMMITMENT_PAUSE are both being set with separate logic.

lparrott (approved):

PR #7146, opened 2022-11-09
Fixing Elevate Commitment GET requests

lparrott:
[W-11999572](https://gus.lightning.force.com/a07EE00001BMMjEYAX)
- Fixed Commitment `GET` requests by including the Id in the endpoint.
- We previously removed this Id, but it should only be removed for `POST` and `PATCH` Commitment requests.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/PS_Request.cls
+++ b/force-app/main/default/classes/PS_Request.cls
@@ -41,6 +41,7 @@ public inherited sharing class PS_Request {
         REFUND,
         TRANSACTIONS,
         COMMITMENT,
+        COMMITMENT_GET,
         COMMITMENT_PAUSE,
         COMMITMENT_CANCEL,
         COMMITMENT_UPDATE_BULK,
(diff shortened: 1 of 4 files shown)


andrewyu-salesforce (approved):

daniel-fuller (approved):
I like that we are separating out a new GET enum. That will help clarify the difference between the two calls.

voduyemi (approved):

force-app/main/default/classes/PS_Request.cls
```diff
@@ -41,6 +41,7 @@ public inherited sharing class PS_Request {
         REFUND,
         TRANSACTIONS,
         COMMITMENT,
+        COMMITMENT_GET,
```
npsp-reedestockton, line 44: Nice approach!

npsp-reedestockton (approved):
Great!

PR #7147, opened 2022-11-09
Fix Display of Gateway Management Error and Correct Clone Behavior

npsp-reedestockton:
We fixed three errors and made a few changes to labels.

- We corrected the error display in Gateway Management if an Elevate connection is unavailable
- We now clear the default template Id on clone, permitting the Elevate bundle to behave properly
- During Gift Entry Batch creation, we now don't gather Elevate gateway information unless Gateway Assignment is enabled

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


daniel-fuller (approved):

lparrott (approved):
Good finds!

PR #7153, opened 2022-11-15
Fix for Pause Request to use PATCH on Updates

lparrott:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_ScheduleService.cls
+++ b/force-app/main/default/classes/RD2_ScheduleService.cls
@@ -1060,6 +1060,7 @@ public without sharing class RD2_ScheduleService {
             for (RecurringDonationSchedule__c schedule : donationSchedules) {
                 if (schedule.IsPause__c) {
                     shouldEdit = true;
+                    break;
                 }
             }
             return shouldEdit;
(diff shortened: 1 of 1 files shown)


force-app/main/default/classes/RD2_ScheduleService.cls
```diff
@@ -1056,10 +1056,13 @@ public without sharing class RD2_ScheduleService {
         }
 
         private Boolean shouldEditPause (List<RecurringDonationSchedule__c> donationSchedules) {
+            Boolean shouldEdit = false;
             for (RecurringDonationSchedule__c schedule : donationSchedules) {
-                return schedule.IsPause__c;
+                if (schedule.IsPause__c) {
+                    shouldEdit = true;
```
voduyemi, line 1062: I think we can return early here the moment we find a paused schedule instead of looping through all the schedules before returning a result.

voduyemi: Also thanks for jumping on this so quickly!

lparrott: Heh, I just thought of that too, I'll add a `break` since I prefer that to having multiple `return` statements.

voduyemi (approved):

PR #7165, opened 2023-01-13
W-12250612 - Fixed issue with BDI matching when using Dupe Mgmt

callen-sfdo:
WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001HsNBvYAN/view

This fixes incorrect internal matches with BDI matching when using the 'Contact Duplicate Rules' (aka Duplicate Management) option.  Essentially if you had two Data Import records with any one of the 4 pieces of contact information used for matching (first name, last name, email, phone) matching, then it would match the two records together incorrectly and only create one new contact.  It should only match internally like that if all the pieces of contact information match.

The cause of the issue was that because the Contact matching rule name didn't include the name of any of the fields used for matching, it was creating a key for matching for each field with just the field populated and all others blank.  This meant that if any of these contact fields matched, the whole contact would be matched (which is very problematic for things like first name).

The fix was to basically make sure that if Duplicate Management was enabled, then it should never include a blank value unless the contact field was actually blank.

Also updated the test class to test the partial contact match scenario and the exact internal duplicate match scenario.  Also added tests for non-dry run scenarios.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/BDI_ContactService.cls
+++ b/force-app/main/default/classes/BDI_ContactService.cls
@@ -355,12 +355,14 @@ public with sharing class BDI_ContactService {
         Boolean modelIsOneToOne = CAO_Constants.isOneToOne();
 
         for (DataImport__c di : bdi.listDI) {
+
             // skip di's that already have an error
             if (di.Status__c == BDI_DataImport_API.bdiFailed)
                 continue;
             List<String> listDiKey = listDIKeyCx(di, strCx);
+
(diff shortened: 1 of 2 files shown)


lparrott (approved):
Looks good, nice work on finding the bug!

daniel-fuller (approved):

PR #7250, opened 2024-09-13
W-14578178 Fix Contact Address Override Functionality

salesforce-suyash-more:
# Critical Changes
GUS [W-14578178](https://gus.lightning.force.com/a07EE00001fVTYKYA4)

To skip the update for contacts with address override, chages made in :

1. Addresses.cls
2. ContactAdapter.cls

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/.forceignore
+++ b/.forceignore
@@ -5,3 +5,6 @@
 # LWC Jest
 **/__tests__/**
 **/__mocks__/**
+**/tsconfig.json
+
+**/*.ts
--- a/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
+++ b/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
@@ -666,7 +666,7 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
             Address__c existingAddressFromContact = existingAddressesByAddress.get(newAddressFromContact);
 
             // Check if the contact has Address Override enabled
-            if (contact.is_Address_Override__c == true) {
+            if (contact.Is_Address_Override__c == true) {
                 // Create a new address record specific to this contact
                 Address__c newAddress = new Address__c();
                 newAddress.Household_Account__c = contact.AccountId; // Link to household/account
--- a/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
+++ b/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
@@ -168,9 +168,17 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
         // set the mailing address for contacts who specify an address lookup or override
         List<Contact> consNeedingAddressFieldsPopulated = consWhoseAddressChanged();
         if (consNeedingAddressFieldsPopulated.size() > 0) {
-            // to avoid pushing these address changes to the HH.
-            setRecursionFlags();
-            setContactAddressFromCurrentAddress(consNeedingAddressFieldsPopulated);
+            // Add null checks before setting address fields
+            for (Contact con : consNeedingAddressFieldsPopulated) {
+                // Only proceed if MailingStreet and MailingCity are not null
+                if (con.MailingStreet != null && con.MailingCity != null) {
+                    setRecursionFlags();
+                    setContactAddressFromCurrentAddress(consNeedingAddressFieldsPopulated);
+                } else {
+                    // Gracefully handle cases where address fields are null
+                    System.debug('Skipping address update for contact with null MailingStreet or MailingCity: ' + con.Id);
+                }
+            }
             deactivateAddrRecursionFlag();
         }
 
@@ -179,7 +187,6 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
                 consWhoseAccountChangedAndDontHaveAddressOverrides();
         addressRefreshQueue.addAll(consHavingAddressOverridesRemoved());
         if (addressRefreshQueue.size() > 0) {
-            // to avoid pushing these address changes to the HH.
             setRecursionFlags();
             HouseholdMembers householdMembersForAddressRefresh =
                     new HouseholdMembers(addressRefreshQueue);
@@ -187,7 +194,7 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
             deactivateAddrRecursionFlag();
         }
 
-        // create any new Default Address objects
+        // create new Default Address objects if required
         Set<Contact> addressCreationQueue = getAddressCreationQueueBeforeUpdate();
         createAddressFromContactAndSetCurrentAddressLookupId(new List<Contact>(addressCreationQueue));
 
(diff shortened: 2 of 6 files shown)


force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
```diff
@@ -665,21 +665,36 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
             Address__c newAddressFromContact = contactAddressesByContact.get(contact);
             Address__c existingAddressFromContact = existingAddressesByAddress.get(newAddressFromContact);
 
-            // if found a match
-            if (contactAddressHasAddressMatch(existingAddressFromContact)) {
+            // Check if the contact has Address Override enabled
+            if (contact.is_Address_Override__c == true) {
```
force2b, line 669: We typically capitalize the first character of a field name (not counting the namespace of course). 
```suggestion
            if (contact.Is_Address_Override__c == true) {
```

force-app/test/ADDR_Addresses_TEST.cls
```diff
@@ -911,14 +911,16 @@ public with sharing class ADDR_Addresses_TEST {
         Test.stopTest();
 
         Address__c updatedAddress = [SELECT Default_Address__c, Undeliverable__c FROM Address__c WHERE Id = :overrideAddressToCreate.Id];
-        Contact updatedContact = [SELECT Undeliverable_Address__c FROM Contact WHERE Id = :contact.Id];
+        Contact updatedContact = [SELECT Undeliverable_Address__c,is_Address_Override__c FROM Contact WHERE Id = :contact.Id];
```
force2b, line 914: Same here
```suggestion
        Contact updatedContact = [SELECT Undeliverable_Address__c, Is_Address_Override__c FROM Contact WHERE Id = :contact.Id];
```

force-app/main/domain/Addresses.cls
```diff
@@ -292,6 +292,10 @@ public inherited sharing class Addresses extends fflib_SObjects2 {
                 NPSP_Contact npspContact = new NPSP_Contact(contact);
 
                 Boolean shouldUpdateContact= false;
+                if (npspContact.hasAddressOverride()) {
```
force2b, line 295: There's logic below that possibly contradicts this.

On line 301-3, it will set IsAddressOverride to false if the Contact.CurrentAdress is null. So, if this IsAddressOverride is checked and there is no lookup to a CurrentAddress for the Contact, is the new intention that we'll leave the IsAddressOverride field as checked?

If that's true, I suspect some of the code below can be adjusted to support that new behavior. However, I'm also uncertain if that's what the exact intent is. 

Is there valid reason to lave Is_Address_Override__c set to true if there is no current address on the Contact? There could be; in that maybe the person just has no address for some reason. If true, then let's clean up the contradicting code below.

force2b (changes requested):
Just a couple of small comments

force-app/main/domain/Addresses.cls
salesforce-suyash-more: Thanks for the suggestion, @force2b. You're right. Since we already use
'continue' when the contact has an address override, the logic to set 'Is_Address_Override__c' to false when 'CurrentAddress' is null is redundant. I'll remove that part to avoid unnecessary processing and ensure the behavior aligns with the intention of skipping updates for override-enabled contacts.

.forceignore
```diff
@@ -4,4 +4,7 @@
 
 # LWC Jest
 **/__tests__/**
-**/__mocks__/**
\ No newline at end of file
+**/__mocks__/**
```
force2b, line 7: Did you mean to check in this file?

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
```diff
@@ -665,22 +665,56 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
...
+                                      String.isBlank(contact.MailingCountry);
+
+            // If all address fields are null, clear the Current_Address__c field and continue
+            if (isAddressFieldsNull) {
+                contact.Current_Address__c = null;
+                continue;  // Skip address creation for this contact
+            }
+            
+            // Check if the contact has Address Override enabled
+            if (contact.Is_Address_Override__c == true) {
```
force2b, line 682: Is there a possibility that this logic will create multiple Address records for the same address over time, basically each time the Contact record is edited to change any field value? The original logic appears to have first looked for a matching address before it inserted a new address. The new changes move that check to after the logic to insert a new address.

force-app/test/ACCT_AccountMerge_TEST.cls
```diff
@@ -197,7 +197,7 @@ private with sharing class ACCT_AccountMerge_TEST {
         System.assertEquals(cHH * cCon, listCon.size());
                     
         // verify Addresses are merged
-        System.assertEquals(cHH + 1, [SELECT count() FROM Address__c]);
+        System.assertEquals(cHH * 2, [SELECT count() FROM Address__c]);
```
force2b, line 200: Why check for double (2x) the number of records here; which I think would be 6 address records? The previous assert was checking for only 4 records. Does the new override cause more address records to be created?

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
salesforce-suyash-more: Thank you for bringing this up. The reason we're proceeding to create a new address when Address Override is enabled, without checking for an existing address, is to ensure that the contact’s address remains completely independent of any household address updates.
By generating a new address for each contact with Address Override enabled, we maintain distinct address records for the contact, which aligns with the intended behavior of keeping contact addresses separate from household changes. This prevents any potential conflicts between the contact’s address and the household’s address. For contacts without Address Override, we continue to match existing addresses before inserting new ones to ensure consistency.

force-app/test/ACCT_AccountMerge_TEST.cls
salesforce-suyash-more: @force2b thank you for your feedback. The assert now checks for double the number of address records because the new Address Override functionality creates a separate address for each contact with the override enabled. Previously, the contacts shared the household address, resulting in fewer records. With the override, each contact now gets its own distinct address, which correctly leads to a higher count. This behavior aligns with the goal of maintaining independent addresses for contacts with Address Override enabled.

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
force2b: That makes sense. My question though is whether every tiny change to the Contact will create a new Address record. To test:
- Create a Contact with an Address and the Override set
- Confirm that there is one Address record linked to that Contact
- Edit the Contact again to add or change the value in Title
- Is another Address record created for the same exact address?

If the answer to that last bullet is yes, we'll need to fix that. Let me know what you find.

salesforce-suyash-more: @force2b, I have tested the scenario where non-address fields like Title are updated, and I confirmed that no new address records are created in such cases. The new address creation is only triggered by changes in address-related fields, as intended.

force2b: Thank you for testing and confirming.

force2b (approved):

salesforce-suyash-more:
@force2b thank you for approving the changes, could you please merge this PR.

PR #7266, opened 2024-11-19
W-17216147 - 'First installment paid' checkbox not being set to true in Gift Entry

salesforce-suyash-more:
# Critical Changes

# Changes
GUS : [W-17216147](https://gus.lightning.force.com/a07EE000025JkNSYA0)

Modified files :
- geBatchWizard.js
- GiftBatch.cls
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/GiftBatch.cls
+++ b/force-app/main/default/classes/GiftBatch.cls
@@ -47,7 +47,7 @@ public inherited sharing class GiftBatch implements IGiftBatch {
     private GiftTemplate giftTemplate;
 
     private Boolean shouldPayFirstInstallment = false;
-    private static final String AllowFirstInstallmentPaid = Label.get('', 'geRD2FirstInstallmentPaid', UserInfo.getLanguage());
+    private static final String AllowFirstInstallmentPaid = Label.get(null, 'geRD2FirstInstallmentPaid', UserInfo.getLanguage());
 
     @TestVisible
     private GiftBatchSelector giftBatchSelector {
--- a/force-app/main/default/classes/GiftBatch.cls
+++ b/force-app/main/default/classes/GiftBatch.cls
@@ -47,7 +47,7 @@ public inherited sharing class GiftBatch implements IGiftBatch {
     private GiftTemplate giftTemplate;
 
     private Boolean shouldPayFirstInstallment = false;
-    private static final String AllowFirstInstallmentPaid = Label.get(null, 'geRD2FirstInstallmentPaid', UserInfo.getLanguage());
+    private static final String AllowFirstInstallmentPaid = 'AllowFirstInstallment__f';
 
     @TestVisible
     private GiftBatchSelector giftBatchSelector {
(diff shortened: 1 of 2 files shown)


force-app/main/default/classes/GiftBatch.cls
```diff
@@ -47,7 +47,7 @@ public inherited sharing class GiftBatch implements IGiftBatch {
     private GiftTemplate giftTemplate;
 
     private Boolean shouldPayFirstInstallment = false;
-    private static final String AllowFirstInstallmentPaid = 'AllowFirstInstallment__f';
+    private static final String AllowFirstInstallmentPaid = 'First installment paid';
```
lparrott, line 50: @salesforce-suyash-more Will this break when the User is set to a non-English language? It looks like the `geRD2FirstInstallmentPaid` Label has translations.

salesforce-suyash-more: @lparrott, your doubt about the translation issue was wise. I've made changes to ensure the FirstInstallmentPaid label is translation-compatible. Please review the latest updates.

lparrott: @salesforce-suyash-more Unfortunately this still won't work in an multi-language org since the Template (in `Batch_Defaults__c`) gets stored in the running User's language, and then it would potentially get read back as a different language. Is there a reason the `Batch_Defaults__c` use field Labels and not API names as the key values? That seems to be the issue, did we change that in the past release?

salesforce-suyash-more: Hi @lparrott , Let's consider a scenario where a user using the English language creates a Gift Batch, which stores the 'First installment paid' English label in 'Batch_Defaults'. However, when a user from a different language (e.g. German) accesses the same Gift Batch and edits the 'Batch Info', the translated label will be updated in 'Batch_Defaults'.

The reason for using field labels in 'Batch_Defaults__c' stems from the [bug fix](url) related to [W-14308598](url)﻿. This bug required replacing API names with field labels to distinguish between the ‘GAU Allocation 1’ and ‘GAU Allocation 2’ object group fields in the gift entry template. Since both object groups will have the same API names, it caused an issue of overriding form field values while setting defaults.
please refer to the [Quip document](url) for details of that bug.

lparrott: @salesforce-suyash-more Thanks for that info, I forgot about that bug fix.
Are you saying the scenario you mentioned will work? That was my concern, an English User creates the Batch which stores the Labels as English, then a German User accesses it and our logic fails because it won't find the German translation of "geRD2FirstInstallmentPaid".

salesforce-suyash-more: Right, @lparrott. The German user won't see the translated label in Batch_Defaults until they make an update in 'Batch Info'.

salesforce-suyash-more: @lparrott, I made some changes to the logic for populating the Batch_Defaults, ensuring that the fieldApiName for the 'First installment paid' checkbox is stored in Batch_Defaults to resolve the translation issues with the labels.

lparrott (approved):
Looks good!

PR #7278, opened 2025-02-05
@W-17760180 - Update to unit test to fix query for customer

lparrott:
Small change to a Unit test to fix a customer scenario where they want our tests to pass in their org.

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b (approved):
Interesting that they want us to fix the test to be deterministic, but technically they're right -- it should have been that way anyway. Thanks for fixing it.

daniel-fuller (approved):

PR #7281, opened 2025-03-06
W-17898954 - checkbox validation error on gift entry

salesforce-suyash-more:
GUS : [W-17898954](https://gus.lightning.force.com/a07EE00002AGI2nYAH)

# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/lwc/geBatchWizard/geBatchWizard.js
+++ b/force-app/main/default/lwc/geBatchWizard/geBatchWizard.js
@@ -68,6 +68,7 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
     @track hasInvalidBatchFields = false;
     @track missingBatchHeaderFieldLabels = [];
     @track missingRequiredFieldsMessage;
+    @track allowFirstInstallment = false;
 
     dataImportBatchFieldInfos;
     dataImportBatchInfo;
@@ -90,7 +91,6 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
         third: 2
     }
 
-    _allowFirstInstallment = false;
     _allowFirstInstallmentDisabled;
     _allowRecurringDonations = false;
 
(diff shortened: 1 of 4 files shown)


force-app/main/default/lwc/geFormRenderer/geFormRenderer.js
```diff
@@ -2651,11 +2651,12 @@ export default class GeFormRenderer extends LightningElement{
             ? this.CUSTOM_LABELS.geTextUpdating
             : this.CUSTOM_LABELS.geTextSaving;
         delete dataImportFromFormState[apiNameFor(PAYMENT_AUTHORIZE_TOKEN)];
+        const dataImportfieldsInfo = this.dataImportObjectInfo?.data?.fields || {};
         Object.keys(dataImportFromFormState).forEach((field)=>{
-            if(dataImportFromFormState[field] === undefined) {
```
lparrott, line 2655: @salesforce-suyash-more Don't we need to make sure Boolean fields are True or False? So if they are undefined or null, they should be set to False? Maybe that happens elsewhere.

andrewyu-salesforce: That's correct, I beleive it's better to set it to true/false instead of null since there is no concept empty on checkbox

daniel-fuller:
@force2b this is the PR you’ll need to pick up and see through to merge. We'll need to address any feedback and test this before merging. I can help with testing once the code changes are complete.

force-app/main/default/lwc/geBatchWizard/geBatchWizard.js
```diff
@@ -68,6 +68,7 @@ export default class geBatchWizard extends NavigationMixin(LightningElement) {
     @track hasInvalidBatchFields = false;
     @track missingBatchHeaderFieldLabels = [];
     @track missingRequiredFieldsMessage;
+    @track allowFirstInstallment = false;
```
lparrott, line 71: These changes fix the issues in Edit Batch Info where "First installment paid" is not disabled and/or unchecked when "Allow Recurring Donations" is changed.

PR #7285, opened 2025-03-27
Updating logic when overriding Contact Address to fix StateCode issue

lparrott:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
+++ b/force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
@@ -681,13 +681,7 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
             // Check if the contact has Address Override enabled
             if (contact.Is_Address_Override__c == true) {
                 // Create a new address record specific to this contact
-                Address__c newAddress = new Address__c();
-                newAddress.Household_Account__c = contact.AccountId; // Link to household/account
-                newAddress.MailingStreet__c = contact.MailingStreet;
-                newAddress.MailingCity__c = contact.MailingCity;
-                newAddress.MailingState__c = contact.MailingState;
-                newAddress.MailingPostalCode__c = contact.MailingPostalCode;
-                newAddress.MailingCountry__c = contact.MailingCountry;
+                Address__c newAddress = new NPSP_Address(contact).getRecord();
                 newAddress.Default_Address__c = false; // It's not a default address
     
                 // Add the new address to be inserted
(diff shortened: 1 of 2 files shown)


force-app/test/ADDR_Addresses_TEST.cls
```diff
@@ -2560,6 +2560,7 @@ public with sharing class ADDR_Addresses_TEST {
                 MailingStreet__c, MailingCity__c, MailingState__c,
                 MailingPostalCode__c, MailingCountry__c
             FROM Address__c
+            ORDER BY Name DESC
```
daniel-fuller, line 2563: @lparrott what was the reason for adding the order by here?

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
```diff
@@ -681,13 +681,7 @@ public inherited sharing class ContactAdapter extends fflib_SObjects2 {
...
             if (contact.Is_Address_Override__c == true) {
                 // Create a new address record specific to this contact
-                Address__c newAddress = new Address__c();
-                newAddress.Household_Account__c = contact.AccountId; // Link to household/account
-                newAddress.MailingStreet__c = contact.MailingStreet;
-                newAddress.MailingCity__c = contact.MailingCity;
-                newAddress.MailingState__c = contact.MailingState;
-                newAddress.MailingPostalCode__c = contact.MailingPostalCode;
-                newAddress.MailingCountry__c = contact.MailingCountry;
+                Address__c newAddress = new NPSP_Address(contact).getRecord();
```
daniel-fuller, line 684: @lparrott could you explain a little how this is preserving the country/state picklist codes? Is it because `NPSP_Address` uses that `copyFromSObject` which checks for state/country enabled?

daniel-fuller (commented):
Just a couple comments

force-app/test/ADDR_Addresses_TEST.cls
lparrott: Without this, one of the tests was failing. We originally were ordering by MalingState, but had to remove that for encryption reasons. This ensures the order of Addresses for testing purposes, but we could figure out a way to remove it.

force-app/main/adapter/in/sobjects/contact/ContactAdapter.cls
lparrott: That's exactly why!

force2b: Looks like it ends up in this method, which does respect State and Country Pick Lists: https://github.com/SalesforceFoundation/NPSP/blob/main/force-app/main/default/classes/NPSP_Address.cls#L334

force2b (approved):

PR #7286, opened 2025-04-10
@W-18224476 - Fix unpause issue with past Open Opportunities

lparrott:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


force2b (commented):
Can we add a unit test for this?

lparrott:
> Can we add a unit test for this?

That's a good idea, the test may end up being more specific to the changes made and not the actual scenario that we're fixing, but I can give it a try.

PR #7290, opened 2025-06-27
Manage Household Page: Auto name check box is getting checked automatically even if unchecked

daniel-fuller:
These code changes fix the following two issues:
1. Updating a household via the manage household page was clearing the `npo02__System_Custom_Naming__c` field on the account record. This means that DLRS jobs and other scheduled roll-ups were renaming household accounts that were manually changed at some point in the past. This was the regression issue.

2. When converting a lead via the NPSP convert logic, auto-naming was not working because when the naming logic was called twice in the same transaction, the npo02__System_Custom_Naming__c was being populated on the second call. I have introduced a flag to ensure that naming is only called once in a single transaction.

This change needs to go out in a patch, which is why I'm creating the PR against `main` instead of `feature/258`

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/domain/HouseholdNamingUserControlledFields.cls
+++ b/force-app/main/domain/HouseholdNamingUserControlledFields.cls
@@ -41,19 +41,30 @@ public inherited sharing class HouseholdNamingUserControlledFields {
     private SObject oldRecord;
 
     public HouseholdNamingUserControlledFields(String str) {
-        this.value = str == null ? ';' : str + ';';
+        // Ensure that double semicolons are not allowed when defining the user controlled naming fields.
+        // This will only add a new semicolon if the previous string does not already end with one.
+        this.value = str == null ? ';' : (str.endsWith(';') ? str : str + ';');
     }
 
     public HouseholdNamingUserControlledFields(SObject household, SObject oldRecord) {
-        this.value = userControlledNamingFieldsFor(household) == null ? ';' :
-                userControlledNamingFieldsFor(household) + ';';
+        // Ensure that double semicolons are not allowed when defining the user controlled naming fields.
+        // This will only add a new semicolon if the previous string does not already end with one.
+        String existingValue = userControlledNamingFieldsFor(household);
+        this.value = existingValue == null ? ';' : (existingValue.endsWith(';') ? existingValue : existingValue + ';');
         this.household = household;
         this.oldRecord = oldRecord;
         buildUserControlledNamingString();
     }
 
     public String asConcatenatedString() {
-        return this.value;
+        // Clean up any double semicolons that might have been added due to multiple executions in the same transaction.
+        // This is more for legacy cleanup if double semicolons were allowed previously that have broken custom naming.
+        String result = this.value.replaceAll(';;+', ';');
+        // Ensure it ends with exactly one semicolon if it doesn't already
+        if (!result.endsWith(';')) {
+            result += ';';
+        }
+        return result;
     }
 
     public Boolean isInformalGreetingControlledByUser() {
@@ -102,6 +113,8 @@ public inherited sharing class HouseholdNamingUserControlledFields {
             appendName();
         }
     }
+
+
 
     private Boolean isReplaceable(String fieldApiName) {
         String fieldValue = (String) household.get(fieldApiName);
--- a/force-app/main/service/HouseholdNamingService.cls
+++ b/force-app/main/service/HouseholdNamingService.cls
@@ -38,7 +38,6 @@
 public without sharing class HouseholdNamingService {
 
     private HouseholdSettings settings = new HouseholdSettings();
-    private static Boolean isNamingServiceExecuted = false;
 
     @TestVisible
     private ContactSelector contactSelector {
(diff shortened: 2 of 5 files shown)


force-app/main/service/HouseholdNamingService.cls
```diff
@@ -117,27 +131,43 @@ public without sharing class HouseholdNamingService {
     * @return void
     */
     public void updateHouseholdNameAndMemberCount(List<Id> householdOrAccountIds) {
-        if (isHouseholdNamingDisabled) {
+        // Increment total call counter for testing purposes
+        if (Test.isRunningTest()) {
```
force2b, line 135: Should this be after the line that returns if `isProcessingHouseholdNaming` is true? This way the counter only increments if the method actually executes fully?

force-app/main/service/HouseholdNamingService.cls
```diff
@@ -117,27 +131,43 @@ public without sharing class HouseholdNamingService {
...
+            testTotalCallCount++;
+        }
+        
+        if (isHouseholdNamingDisabled || isProcessingHouseholdNaming) {
```
force2b, line 139: In theory, using `isProcessingHouseholdNaming` is what recursion control (`TDTM_ProcessControl`) is supposed to do as well, but I wonder if it's not actually working as expected here.

force-app/test/HouseholdNamingUserControlledFields_TEST.cls
```diff
@@ -0,0 +1,142 @@
+/*
+    Copyright (c) 2021, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2025, Salesforce.org
```

force-app/test/HouseholdNamingUserControlledFields_TEST.cls
```diff
@@ -0,0 +1,142 @@
...
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
```
force2b, line 32:
```suggestion
* @date 2025
```

force-app/test/HouseholdNamingUserControlledFields_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>53.0</apiVersion>
```
force2b, line 3:
```suggestion
    <apiVersion>63.0</apiVersion>
```

force-app/main/domain/HouseholdNamingUserControlledFields.cls
```diff
@@ -41,19 +41,30 @@ public inherited sharing class HouseholdNamingUserControlledFields {
     private SObject oldRecord;
 
     public HouseholdNamingUserControlledFields(String str) {
-        this.value = str == null ? ';' : str + ';';
+        // Ensure that double semicolons are not allowed when defining the user controlled naming fields.
+        // This will only add a new semicolon if the previous string does not already end with one.
+        this.value = str == null ? ';' : (str.endsWith(';') ? str : str + ';');
```
force2b, line 46: Was this issue (double-semicolon) one of the root issues of the problems being addressed or was this just an observation?

force2b (changes requested):
A few small requested changes and questions. Also wondering if this was reproducible fully in a scratch org?

force-app/main/service/HouseholdNamingService.cls
daniel-fuller: This has been removed because it's not necessary

force-app/main/domain/HouseholdNamingUserControlledFields.cls
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2021 Salesforce.org
+    Copyright (c) 2025 Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2021 Salesforce.org
```

force-app/main/service/HouseholdNamingService.cls
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2011, Salesforce.org
+    Copyright (c) 2025, Salesforce.org
```
force2b, line 2:
```suggestion
    Copyright (c) 2011, Salesforce.org
```

lparrott (approved):

andrewyu-salesforce (approved):

PR #7291, opened 2025-07-01
@W-18492760 Fix Recursive Issue

andrewyu-salesforce:
# Critical Changes

# Changes
Revert change from #7250

Unit Test:
image
 
# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata


lparrott (approved):
Test scenario failed without this change, and passed with it. Also, since this is a revert of a different bug fix, I'm less worried about regressions. Approved!

force2b:
@andrewyu-salesforce I'm spinning up an org to test against, but a couple of questions:
- The PR being reverted was a P3 from Oct 2024. Reading Reede's comments on the [work item](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001fVTYKYA4/view), it looks like maybe that was a subjective bug. Are there any concerns with reverting that, and re-introducing the original bug?
- Are there any other tests that should be introduced with this fix to validate behaviors?

andrewyu-salesforce:
@force2b 
- I believe there is less concern with reverting the original bug fix, as it was based on opinion and no customers have actually reported an issue related to it. The new bug is significant and affect lots of customers. One of the customer is a ISV working with V4S.
- I tried to reproduce them in Apex test scenario but was not successful. The error requires creating a new Trigger on top of our existing NPSP default trigger. I reverted all the original tests that were modified in the original PR and make sure all original tests pass.

daniel-fuller (approved):

PR #7311, opened 2025-10-01
@W-19104143 Fix Pause API

andrewyu-salesforce:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_ApiService_TEST.cls
+++ b/force-app/main/default/classes/RD2_ApiService_TEST.cls
@@ -572,6 +572,10 @@ private class RD2_ApiService_TEST {
         npe03__Recurring_Donation__c rd = rdGateway.getRecords()[0];
 
         RD2_ScheduleService_TEST.createPauseSchedule(rd.Id, CURRENT_DATE.addMonths(1), CURRENT_DATE.addMonths(2));
+        
+        rd.Status__c = RD2_Constants.STATUS_PAUSED;
+        update rd;
+        rd = rdGateway.getRecords()[0];
 
         Map<Id, RD2_ApiService.PauseObject> pauseObjectsByIdIn =
             new Map<Id, RD2_ApiService.PauseObject>{
@@ -581,7 +585,11 @@ private class RD2_ApiService_TEST {
         RD2_ApiService apiService = new RD2_ApiService();
         String response = apiService.pause(pauseObjectsByIdIn);
         Test.stopTest();
-
+        
+        npe03__Recurring_Donation__c updatedRd = rdGateway.getRecords()[0];
+       
+        System.assertEquals(RD2_Constants.STATUS_ACTIVE, updatedRd.Status__c,
+        'The status should be active');
         Map<Id, RD2_ApiService.PauseObject> pauseObjectsByIdOut =
             (Map<Id, RD2_ApiService.PauseObject>) JSON.deserialize(response, Map<Id, RD2_ApiService.PauseObject>.class);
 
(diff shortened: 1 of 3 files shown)


force-app/main/default/classes/RD2_RecurringDonation.cls
```diff
@@ -474,13 +474,14 @@ public inherited sharing class RD2_RecurringDonation {
             : rd.npe03__Paid_Amount__c;
     }
 
-    public RecurringDonationSchedule__c getPauseSchedule() {
```
andrewyu-salesforce, line 477: this method is only being used by the Callable_API.

force2b (changes requested):
The way the API should work - I believe - is to allow the user to unpause an RD at any point. So if the pause doesn't start until December 2025, and they want to use this API to remove that pause, they should be able to. 

However, I think the change to use `isActivePause()` changes that behavior. It would prevent unpausing an RD when the Pause hasn't started yet. I don't believe that is our intent. Instead, I believe this should be something like `isActiveOrFuturePause()`.

At the same time, I would have expected some Unit Tests to catch that scenario too. Before changing the logic, can you add a new unit test to see if the API can be used to remove a future pause? It's a good test to have either way. Ideally that test will fail and changing the logic to `isActiveOrFuturePause()` will allow it to pass.

andrewyu-salesforce:
@force2b The changed has been implemented

force-app/main/default/classes/RD2_ApiService_TEST.cls
```diff
@@ -595,6 +603,97 @@ private class RD2_ApiService_TEST {
...
+            if (pauseSchedule.Id == pastSchedule.Id) {
+                System.assertEquals(CURRENT_DATE.addMonths(-3), pauseSchedule.StartDate__c,
+                    'The already passed scheduled pause schedule should stay unchanged');
+                System.assertEquals(CURRENT_DATE.addMonths(-2), pauseSchedule.EndDate__c,
+                    'The already passed pause schedule should stay unchanged');
+                continue;
+            }
+
+            System.assertEquals(CURRENT_DATE.addDays(-1), pauseSchedule.EndDate__c,
+            'The pause schedule should be deactivate');
```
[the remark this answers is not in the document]
andrewyu-salesforce: I will add the test case to cover the scenario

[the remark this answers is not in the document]
andrewyu-salesforce: test added

force-app/main/default/classes/RD2_ApiService_TEST.cls
```diff
@@ -694,6 +694,60 @@ private class RD2_ApiService_TEST {
...
+        String response = apiService.pause(pauseObjectsByIdIn);
+        Test.stopTest();
+        
+        npe03__Recurring_Donation__c updatedRd = rdGateway.getRecords()[0];
+       
+        System.assertEquals(RD2_Constants.STATUS_ACTIVE, updatedRd.Status__c,
+            'The status should be active');
+
+        List<RecurringDonationSchedule__c> pauseSchedules = getPauseSchedules(rd);
+        System.assertEquals(4, pauseSchedules.size(), 'There should be 5 pause schedules');
```
force2b, line 727: The assert is for 4, but the comment is for 5

force-app/main/default/classes/RD2_ApiService_TEST.cls
```diff
@@ -694,6 +694,60 @@ private class RD2_ApiService_TEST {
...
+            }
+
+            System.assertEquals(CURRENT_DATE.addDays(-1), pauseSchedule.EndDate__c,
+            'The pause schedule should be deactivate');
+        }
+       
+        Map<Id, RD2_ApiService.PauseObject> pauseObjectsByIdOut =
+            (Map<Id, RD2_ApiService.PauseObject>) JSON.deserialize(response, Map<Id, RD2_ApiService.PauseObject>.class);
+
+        System.assertEquals(true, pauseObjectsByIdOut.values()[0].isSuccess,
```
force2b, line 745: Do you think this should also validate that there is an active (unpaused) schedule with no end date?

andrewyu-salesforce: I think the active status assertion is sufficient, the active schedule is never modified in the first place when pausing/unpausing.

PR #7325, opened 2026-05-07
@W-19306508 Monthly RD Next Donation Date and Installment Fix

andrewyu-salesforce:
# Critical Changes

# Changes

# Issues Closed

# Community Ideas Delivered

# New Metadata

# Deleted Metadata

### DIFF
--- a/force-app/main/default/classes/RD2_ScheduleService.cls
+++ b/force-app/main/default/classes/RD2_ScheduleService.cls
@@ -524,29 +524,57 @@ public without sharing class RD2_ScheduleService {
     }
 
     /***
-    * @description Calculates the next donation date >= referenced date
+    * @description Calculates the next donation date that is on or after the given reference date.
+    *
+    * Algorithm:
+    * 1. Compute the first valid donation date from the schedule's start date (honoring DayOfMonth).
+    * 2. If that date already meets or exceeds the reference date, return it immediately.
+    * 3. Otherwise, use arithmetic to fast-forward close to the reference date in one step:
+    *    compute how many period-units (months/days/years) lie between the first valid date and
+    *    the reference date, then jump forward by the largest multiple of the installment frequency
+    *    that fits (integer division, so the result is <= referenceDate).
+    * 4. For monthly schedules, snap the result to the correct DayOfMonth (handles short months
+    *    and "Last_Day").
+    * 5. If the jump landed before the reference date (due to integer division truncation),
+    *    advance by one more installment period and re-snap for monthly.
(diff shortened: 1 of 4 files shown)


force-app/main/default/classes/RD2_ScheduleService.cls
```diff
@@ -543,24 +543,16 @@ public without sharing class RD2_ScheduleService {
...
-        if (schedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY) {
-            if (schedule.DayOfMonth__c == null) {
-                return adjusted;
-            }
-            Integer nextDayOfMonth;
-            if (schedule.DayOfMonth__c == RD2_Constants.DAY_OF_MONTH_LAST_DAY ||
-                Integer.valueOf(schedule.DayOfMonth__c) > Date.daysInMonth(adjusted.year(), adjusted.month()))
-            {
-                nextDayOfMonth = Date.daysInMonth(adjusted.year(),adjusted.month());
+            if (schedule.InstallmentPeriod__c == RD2_Constants.INSTALLMENT_PERIOD_MONTHLY && schedule.DayOfMonth__c != null) {
```
force2b, line 581: It's this particular `if` block that is duplicated in this method that's confused me. I think the reason is if the `referenceDate` is is before the calculated end of month date.

The method comment block has:
`Reference date used to calculate next donation date. It can be today or a future projected date.`

So if the `referenceDate` is today (5/12) and the initial calculated `adjustedDate` is 5/31 this won't run again.

But if the `referenceDate` where passed as 6/30 and the initial `adjustedDate` (line 547) came out to 5/31, then this block would run.  Then it would recalculate the `adjustedDate` again on line 551 and then again (if Monthly) on line 553, but passing in the previous calculation.

I will admit it's confusing. Do you think we could get Claude to write up a comment block on this method logic and maybe the one it calls?

andrewyu-salesforce: The `adjustToTargetDayOfMonth` happens after `addDateUnits` to change the monthly RD's installment date to the correct date. The calculation from line 539 to 544 do 1 calculation to shift the installment date from the start of the schedule to after the reference date, and due to rounding error, line 550 to 556 do the 1 last shift when `(unitsBetween / adjustedFrequency)` on line 544 doesn't return a whole number.

I will add the comment block to explain how the calculation works. It for sure is confusing without Claude's help.

andrewyu-salesforce: comment added

force2b (approved):

lparrott (approved):

